PR #7245, opened 2021-11-26
All langs: apply the explicit-this patch to all remaining code

erik-krogh:
The patch was written automatically by a tool.

### DIFF
--- a/java/ql/lib/semmle/code/java/security/CommandArguments.qll
+++ b/java/ql/lib/semmle/code/java/security/CommandArguments.qll
@@ -155,7 +155,7 @@ private class CommandArgArrayImmutableFirst extends CommandArgumentArray {
   Expr getFirstElement() {
     result = this.getAWrite(0)
     or
-    not exists(getAWrite(0)) and
+    not exists(this.getAWrite(0)) and
     result = firstElementOf(this.getDefiningExpr())
   }
 
--- a/java/ql/lib/semmle/code/java/security/CommandArguments.qll
+++ b/java/ql/lib/semmle/code/java/security/CommandArguments.qll
@@ -155,7 +155,7 @@ private class CommandArgArrayImmutableFirst extends CommandArgumentArray {
   Expr getFirstElement() {
     result = this.getAWrite(0)
     or
-    not exists(this.getAWrite(0)) and
+    not exists(getAWrite(0)) and
     result = firstElementOf(this.getDefiningExpr())
   }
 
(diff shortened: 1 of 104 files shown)


owen-mc:
❤️

esbena:
JS 👍🏻

java/ql/lib/semmle/code/java/security/CommandArguments.qll
```diff
@@ -155,7 +155,7 @@ private class CommandArgArrayImmutableFirst extends CommandArgumentArray {
   Expr getFirstElement() {
     result = this.getAWrite(0)
     or
-    not exists(this.getAWrite(0)) and
+    not exists(getAWrite(0)) and
```
esbena: Is this compiler difference being tracked somewhere?

hvitved (approved):
C# 👍 Could you please start DCA runs for all languages as well? We have seen in the past that adding explicit `this` qualifiers made performance worse.

java/ql/lib/semmle/code/java/security/CommandArguments.qll
erik-krogh: I don't think there is an issue tracking it.  
But it has been mentioned to the core team, and I think `alexet` is looking into removing the pointless negative recursion, but that is apparently tricky (see the core channel from October 18th).

MathiasVP (approved):
C++ 👍 (assuming DCA is happy)

aschackmull:
I think the dca run ought to have included more queries for Java to test more broadly as these changes are all over. As-is there are 3 projects that show minor slowdowns, which _may_ be significant, but I'm not entirely sure whether I ought to consider that suspicious or not. A 10% slowdown with the absolute slowdown being twice the stddev of 3 runs _seems_ like it might be significant. The two 25% slowdowns could have benefitted from a couple of reruns to get stddev values - as-is it's hard to say whether to read something into those.

erik-krogh:
> I think the dca run ought to have included more queries for Java to test more broadly as these changes are all over.

I've queued some retries to get some stddev values on those regressions.  

---- 

Evaluations for the other languages look OK (some improvements, some regressions, seems to be within margin of error).   

Except for `csharp`, which has some massive performance improvements.   
But those improvements are spurious and are caused by the second database missing a bunch of source-code (see the `source-archive-hashes` folder in the DCA report).   
I've started a new DCA experiment where the variants are not grouped.

aschackmull (approved):
The experiment I added with the full set of 179 queries looks better (just one bad row, but well within the stddev). So 👍  for Java.

yoff (approved):
Python 👍

PR #7246, opened 2021-11-26
Python: Support flow through `import *`

tausbn:
Adds results for `ModuleVariableNode::getARead` corresponding to reads
that go through (chains of) `import *`.

This required a bit of a change to _which_ module variables we define.
Previously, we only included variables that were accessed elsewhere in
the same file, but now we must ensure to also include variables that may
be accessed through `import *`.

---
Can be reviewed commit-by-commit, though be aware that the second commit not only moves stuff around, but also refactores and renames a bunch of predicates (so as to make them possible to expose publically). For that reason, it's probably best to review that commit as if it's an entirely new implementation. I apologise for the inconvenience.

For figuring out which module is referred to when writing `from ... import *`, I used the existing points-to machinery. It's possible this could be replaced with logic that doesn't depend on points-to, but in the short term this seemed like the easiest solution for a precise module resolution.

I don't think this needs a change note, but I'll be happy to add one if the reviewer(s) think otherwise.


python/ql/lib/semmle/python/dataflow/new/internal/ImportStar.qll
```diff
@@ -0,0 +1,80 @@
...
+    or
+    // Defined as a built-in
+    name = Builtins::getBuiltinName()
+    or
+    // Defined as a global in this module
+    globalNameDefinedInModule(name, n.getEnclosingModule())
+    or
+    // A non-built-in that still has file-specific meaning
+    name in ["__name__", "__package__"]
+  }
```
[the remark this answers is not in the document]
yoff: Alternatively, you can make `n` existentially quantified in the first branch and just take the string as the argument (that should give a smaller domain as `NameNode`s may refer to the same string).

python/ql/test/experimental/dataflow/import-star/trois.py
```diff
@@ -0,0 +1 @@
+foo = 5
```
yoff, line 1: Giving the variables the same value does not set us up for testing cross-talk... :-)

yoff (changes requested):
Overall looks fine. It was a bit hard to read as there are a number of similar predicates in `ImportStar.qll`

python/ql/lib/semmle/python/dataflow/new/internal/ImportStar.qll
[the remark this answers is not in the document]
tausbn: This is just straight-up an error on my part. Clearly there should be a relation between `n` and `name` in all cases.

The only thing that makes me slightly concerned about this is the use of the `x.foo() = y.foo()` idiom. I've been bitten by bad join orders on this sort of construction too many times in the past. Let's hope it does the join on the scope first. :slightly_smiling_face: 
(And test thoroughly, of course.)

python/ql/test/experimental/dataflow/import-star/trois.py
tausbn: I mean, the test is looking at the actual flow steps, but sure, I can change this to some other integer.

yoff: Indeed, but I am hoping that at some point we will also run the code..

tausbn:
Hm. The performance hit is a bit bigger than I expected. I'll investigate and see if there's something strange going on.

tausbn:
Performance looks good to me.

yoff (approved):
I wonder if going with the alternative of using just the string had made things easier, but I think the same three-way-join issue would still arise...

tausbn:
Yeah, I don't think it made a difference.

One thing that can still harm performance here is the fact that we calculate `isDefinedLocally` for each `Name` in a scope, when really all we want to know is that something with that `getId` is referenced in that scope.

Put differently, imagine we have 1000 `Name`s in a scope, all with the same `getId`. In that case, we will have 1000 entries in `isDefinedLocally` (possibly more if this is inside a deeply nested scope).

What we could do instead, then, would be to define `isDefinedLocally(Scope s, string name)`, and then on the outside check that our given name has the right scope and id.

I don't think we need to do this as things are right now, though, but it's something to keep in mind in case we encounter a really pathological codebase.

PR #7249, opened 2021-11-27
Fix ruby incorrect version in documentation

ShockwaveNN:
There is no `ruby 3.02` version - there is `ruby 3.0.2`


ShockwaveNN:
Also latest current stable of ruby - 3.0.3

There is no major changes, only security changes, so I think it's safe to change it to 3.0.3 in this doc if maintainers agree

aibaars (approved):
Thanks!

PR #7257, opened 2021-11-29
C#: Use .NET Core Nuget package stub i test

michaelnebel:
In this PR we use a combination of auto generated and (existing) hand written stubs as a replacement for dll's in the flow summaries test.
We suspect that the rows that have been removed from the flow summaries are due to the fact that the summaries are now created based on Microsoft.NETCore.App version 2.2.8, which is from 2019, where as some of the dll's that it is replacing might be newer.
On the other hand more summaries have been introduced than removed.
It would be great, if we could have a project for using NETCore 6.0 as the basis for our stubs.

As a side effect, some cleanup was needed in the hand written stub System.Web.cs as we otherwise would have an overlapping implementation of the class HttpResponse (as it exists in Microsoft.NETCore.App). This has a rippling effect into other test cases, which relies on System.Web.cs. These have also been cleaned up and refers now only to auto generated and manually created stubs (this required autogenerating more stubs).


hvitved:
> We suspect that the rows that have been removed from the flow summaries are due to the fact that the summaries are now created based on Microsoft.NETCore.App version 2.2.8, which is from 2019, where as some of the dll's that it is replacing might be newer.

Didn't we end up basing it on the .NET 5 SDK?

csharp/ql/test/resources/stubs/runtime.win-arm64.runtime.native.System.Data.SqlClient.sni/4.4.0/runtime.win-arm64.runtime.native.System.Data.SqlClient.sni.csproj
```diff
@@ -0,0 +1,12 @@
+<Project Sdk="Microsoft.NET.Sdk">
```
hvitved, line 1: Why are we getting just `.csproj` files, and no `.cs` files? Perhaps @tamasvajk knows?

tamasvajk: We can remove the folders that contain only a .csproj file, but we need to also remove the reference to these .csproj files.

We probably end up with these projects, because the original assemblies didn't contain any public types. As far as I remember, we're creating the folder structure based on the `project.assets.json` file (and not the extracted types), which includes these assemblies too.

michaelnebel:
> > We suspect that the rows that have been removed from the flow summaries are due to the fact that the summaries are now created based on Microsoft.NETCore.App version 2.2.8, which is from 2019, where as some of the dll's that it is replacing might be newer.
> 
> Didn't we end up basing it on the .NET 5 SDK?

The same stubs are generated for Microsoft.NETCore.App (module some comments mentioning NewtonSoft.JSon), when stubs are generated for NewtonSoft.JSon using net5.0 as the target framework compared to generating the stubs directly for Microsoft.NETCore.App version 2.2.8.

csharp/ql/test/resources/stubs/runtime.win-arm64.runtime.native.System.Data.SqlClient.sni/4.4.0/runtime.win-arm64.runtime.native.System.Data.SqlClient.sni.csproj
michaelnebel: Do you prefer that I manually go through the generated stubs and remove what is strictly not needed?

hvitved: In that case let's just leave them in.

hvitved (approved):

PR #7258, opened 2021-11-29
Java: Unsafe Hash Query

davidvenegasr:
This query finds Weak Hash Vulnerabilities. 

Why a new query?

I ran codeQL againts OWASP-Benchmark and noticed the score in the Weak Hashing category was zero, I managed to make a quick query based on the one used for Weak Cryptography. This new query detects now Hash Vulnerabilities with a precision equivalent to FindSecBugs or SonarQube with a True Positive rate of 70% on the OWASP Benchmark.   

Benchmark_v1 2_Scorecard_for_Weak_Hashing_Algorithm

I was wondering if there was any specific reason why this query wasn't implemented. If this query results interesting to the security lab I can put further work on it. 

I added two `.java` tests cases: a safe hash and an unsafe hash.

### DIFF
--- a/java/ql/src/experimental/Security/CWE/CWE-328/BrokenHashAlgorithm.qhelp
+++ b/java/ql/src/experimental/Security/CWE/CWE-328/BrokenHashAlgorithm.qhelp
@@ -1,0 +1,36 @@
+<!DOCTYPE qhelp PUBLIC
+>
+<qhelp>
+<overview>
+<p>Using weak cryptographic algorithms can allow and adversasy to adversary to reasonably determine the original input (preimage attack).
+Find another input that can produce the same hash (2nd preimage attack).
+Or find multiple inputs that evaluate to the same hash (birthday attack).</p>
+
+<p>Many cryptographic algorithms provided by cryptography libraries are known to be weak and should be avoided.</p>
+
+</overview>
+<recommendation>
+
+<p>Ensure that you use a strong, modern cryptographic HASH algorithm.</p>
+
+</recommendation>
+<example>
+
+<p>The following code shows an example of using a java <code>Cipher</code> to encrypt some data.
+When creating a <code>Cipher</code> instance, you must specify the encryption algorithm to use. The first
+example uses DES, which is an older algorithm that is now considered weak. The second example uses AES, which
+is a strong modern algorithm.</p>
+
+<sample src="UnsafeHash.java" />
+
+</example>
+<references>
+
+<li><a href="https://cwe.mitre.org/data/definitions/328.html">
+CWE-328</a>.</li>
+
+<!--  LocalWords:  CWE
+ -->
+
+</references>
+</qhelp>
--- a/java/ql/src/experimental/Security/CWE/CWE-328/SafeHash.java
+++ b/java/ql/src/experimental/Security/CWE/CWE-328/SafeHash.java
@@ -1,0 +1,123 @@
+/**
+ * OWASP Benchmark v1.2
+ *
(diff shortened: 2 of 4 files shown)


smowton:
Do you intend to apply to the security lab bounty program for this query?

davidvenegasr:
> Do you intend to apply to the security lab bounty program for this query?

I'm not sure if this query is complex enough to apply, but I guess I could try.  I will be openning an issue explaining in detail.

aschackmull:
This query looks like an exact (ignoring a few minor details) copy of `BrokenCryptoAlgorithm.ql`. Have you found differences between the results of this query and `BrokenCryptoAlgorithm.ql`? If so, what are those differences and why?

davidvenegasr:
Yes I said before this query was heavely base on the one used for Weak Cryptography (a.k.a. `BrokenCryptoAlgorithm.ql`) and I was surprised there was not one for weak Hash, maybe it will be better to extend  `BrokenCryptoAlgorithm.ql` to rather also detect Broken Hash. If that's something that you consider interesting I will be happy to do it. Also I could replace the deprecated `getRepresentedString()` in `BrokenCryptoAlgorithm.ql`.

And yes, there are different results! 

When I ran this new hash query against the OWASP Benchmark:

image

(I also found many other weak hash vulnerabilities alerts in other LGTM repositories).

When I ran the old `BrokenCryptoAlgorithm.ql` againts the Benchmark:

image

**This two images show that for the OWASP Java Benchmark results don't intersect.**

The main difference is in the algortihms it looks for:

```CodeQL
 //NEW Use of Weak Hash.ql
class BrokenHashLiteral extends StringLiteral {
  BrokenHashLiteral() {
    getValue().length() <  10 and 
    getValue().regexpMatch(getInsecureHashRegex())
  }
}

  string getInsecureHashRegex() {
    result = hashRegex(getAnInsecureHashAlgorithmName())
}
\\Which can be found in semmle.code.java.security.Encryption as
\\...
string getAnInsecureHashAlgorithmName() {
  result = "SHA1" or
  result = "MD5"
}
```
Versus the old one 
```CodeQL
//BrokenCryptoAlgorithm.q
private class ShortStringLiteral extends StringLiteral {
  ShortStringLiteral() { getRepresentedString().length() < 100 }
}

class BrokenAlgoLiteral extends ShortStringLiteral {
  BrokenAlgoLiteral() {
    getRepresentedString().regexpMatch(getInsecureAlgorithmRegex()) and
    // Exclude German and French sentences.
    not getRepresentedString().regexpMatch(".*\\p{IsLowercase} des \\p{IsLetter}.*")
  }
}

string getInsecureAlgorithmRegex() {
  result = algorithmRegex(insecureAlgorithmString(max(int i | exists(rankedInsecureAlgorithm(i)))))
}
// Where in the end is mostly looks the following known Algorithms
string getASecureAlgorithmName() {
  result =
    [
      "RSA", "SHA256", "SHA512", "CCM", "GCM", "AES(?![^a-zA-Z](ECB|CBC/PKCS[57]Padding))",
      "Blowfish", "ECIES"
    ]
}
```
I think this is my firts pull request ever to an OpenSource Project, so I would love to know your opinions on: if this is something usefull to detect with CodeQL and if you'll rather have me extend the old brokenAlgo query to also detect broken Hash instead of with a new query.

davidvenegasr:
Hi, any comments? maybe I should close this pull request and re submit as a modification to the original query to improve recall?

aschackmull:
The existing query already covers weak hashes, so I don't think this PR adds anything. The reason for it not being recognised in the benchmark was merely a case of a tag that was missing, which was fixed in https://github.com/github/codeql/pull/7318.

PR #7259, opened 2021-11-29
Python: Add more path-injection sinks from `os` and `tempfile` modules

RasmusWL:


yoff (approved):
Thanks for adding these. We could consider also flagging the setting of [`tempfile.tempdir`](https://docs.python.org/3/library/tempfile.html#tempfile.tempdir) since it is probably bad if a user is in control of that...

PR #7260, opened 2021-11-29
Data flow: Introduce `ParameterPosition` and `ArgumentPosition`

hvitved:
This PR generalizes argument-to-parameter matching in the shared data-flow library from simple integer position based matching to per-language defined matching.

## Why

For static languages, integer based matching has worked just fine, since each call knows the signature of the callee(s), and adjustments of argument positions can be made accordingly at the call sites. For example, if we have a C# call
```csharp
Foo(snd: x, fst: y)
```
which targets a `Foo(int fst, int snd)` method, but swaps the order using named arguments, then we know that `x` must have position `1` and `y` must have position `0`.  Similarly, for calls to methods with variadic arguments
```csharp
Bar(1, 2, 3, 4)
```
where the signature is `Bar(params int[] args)`, we can treat the call as syntactic sugar for
```csharp
Bar(new int[] {1, 2, 3, 4})
```
that is, `1..4` are no longer `ArgumentNode`s, but instead there is a synthesized array argument node with position `0`.

However, for dynamic languages we do not know the signature of the callee(s), so we may have a Ruby call
```rb
foo(1, 2, 3, 4)
```
that targets both a `foo(a, b, c, d)` function and a `foo(first, *mid, last)` function, so whether `2` and `3` should be wrapped in an array depends on the callee, and what gets passed into `last` depends on the number of arguments at the call sites.

## How

This PR introduces two new classes `ParameterPosition` and `ArgumentPosition`, and a predicate `parameterMatch(ParameterPosition ppos, ArgumentPosition apos)` to the interface of the shared data-flow library. Each `ParameterNode` has a `ParameterPosition`, each `ArgumentNode` has an `ArgumentPosition`, and an argument is compatible with a parameter as dictated by `parameterMatch`. (Note that `parameterMatch` does *not* rely on the call graph, we already have `viableCallable` for that.)

Defining `ParameterPosition = ArgumentPosition = int` and `parameterMatch(ParameterPosition pp, ArgumentPosition ap) { pp = ap }` yields the existing behaviour, and that is how all languages are currently implemented (except for C#, where I decided to use different IPA wrappers to catch type errors).

## Follow-up work

For the named argument example, assume we have two Ruby calls
```rb
foo(fst: a, snd: b)
foo(snd: c, fst: d)
```
both targeting `foo(fst: , snd: )`. We can assign to `fst`, `a`, and `d` the positions `Named(fst)`, assign to `snd`, `b`, and `c` the positions `Named(snd)`,  and have `parameterMatch` be the identity.

For the variadic arguments example, consider the Ruby call
```rb
foo(1, 2, 3, 4)
```
that targets `foo(a, b, c, d)` and `foo(first, *mid, last)`. Assuming we have `PositionalArg(x, y)` mean positional argument `x` of `y`, `PositionalParam(x)` mean positional parameter `x`, `SplatParam(x, y)` mean splat parameter at position `x` skipping the last `y` arguments,`LastParam(x)` mean the `x`th last parameter, and `mid_synth` be a synthesized parameter node, we can assign the following positions

Element | Position
--- | ---
`1` | `PositionalArg(0, 4)`
`2` | `PositionalArg(1, 4)`
`3` | `PositionalArg(2, 4)`
`4` | `PositionalArg(3, 4)`
`a` | `PositionalParam(0)`
`b` | `PositionalParam(1)`
`c` | `PositionalParam(2)`
`d` | `PositionalParam(3)`
`first` | `PositionalParam(0)`
`mid_synth` | `SplatParam(1, 1)`
`last` | `LastParam(0)`

and define

```ql
parameterMatch(ParameterPosition ppos, ArgumentPosition apos) {
  exists(int x, int y |
    apos = PositionalArg(x, y) |
    ppos = PositionalParam(x)
    or
    exists(int from, int except |
      ppos = SplatParam(from, except) and
      x >= from and
      y - x > except
    )
    or
    exists(int last |
      ppos = LastParam(last) and
      y - x - 1 = last
    )
  )
}
```

`mid` will then no longer be a `ParameterNode`, but instead we will have an array store-step from `mid_synth` to `mid`, so we are effectively moving the implicit array creation from the caller to the callee, but in the context of the callee we know exactly which arguments to include.

### DIFF
--- a/cpp/ql/lib/semmle/code/cpp/dataflow/internal/DataFlowImplCommon.qll
+++ b/cpp/ql/lib/semmle/code/cpp/dataflow/internal/DataFlowImplCommon.qll
@@ -391,14 +391,11 @@ private module Cached {
 
   /**
    * Holds if `p` is the parameter of a viable dispatch target of `call`,
-   * and `p` matches arguments at position `apos`.
+   * and `p` has position `ppos`.
    */
   pragma[nomagic]
-  private predicate viableParam(DataFlowCall call, ArgumentPosition apos, ParamNode p) {
-    exists(ParameterPosition ppos |
-      p.isParameterOf(viableCallableExt(call), ppos) and
-      parameterMatch(ppos, apos)
-    )
+  private predicate viableParam(DataFlowCall call, ParameterPosition ppos, ParamNode p) {
+    p.isParameterOf(viableCallableExt(call), ppos)
   }
 
   /**
@@ -407,9 +404,9 @@ private module Cached {
    */
   cached
   predicate viableParamArg(DataFlowCall call, ParamNode p, ArgNode arg) {
-    exists(ArgumentPosition pos |
-      viableParam(call, pos, p) and
-      arg.argumentOf(call, pos) and
+    exists(ParameterPosition ppos |
+      viableParam(call, ppos, p) and
+      argumentPositionMatch(call, arg, ppos) and
       compatibleTypes(getNodeDataFlowType(arg), getNodeDataFlowType(p))
     )
   }
--- a/cpp/ql/lib/semmle/code/cpp/dataflow/internal/DataFlowImplCommon.qll
+++ b/cpp/ql/lib/semmle/code/cpp/dataflow/internal/DataFlowImplCommon.qll
@@ -63,6 +63,18 @@ predicate accessPathCostLimits(int apLimit, int tupleLimit) {
 }
 
 /**
+ * Holds if `arg` is an argument of `call` with an argument position that matches
+ * parameter position `ppos`.
+ */
+pragma[noinline]
+predicate argumentPositionMatch(DataFlowCall call, ArgNode arg, ParameterPosition ppos) {
+  exists(ArgumentPosition apos |
+    arg.argumentOf(call, apos) and
+    parameterMatch(ppos, apos)
+  )
+}
+
+/**
  * Provides a simple data-flow analysis for resolving lambda calls. The analysis
  * currently excludes read-steps, store-steps, and flow-through.
  *
@@ -71,31 +83,27 @@ predicate accessPathCostLimits(int apLimit, int tupleLimit) {
  * calls. For this reason, we cannot reuse the code from `DataFlowImpl.qll` directly.
  */
 private module LambdaFlow {
-  private predicate viableParamNonLambda(DataFlowCall call, ArgumentPosition apos, ParamNode p) {
-    exists(ParameterPosition ppos |
-      p.isParameterOf(viableCallable(call), ppos) and
-      parameterMatch(ppos, apos)
-    )
+  pragma[noinline]
+  private predicate viableParamNonLambda(DataFlowCall call, ParameterPosition ppos, ParamNode p) {
+    p.isParameterOf(viableCallable(call), ppos)
   }
 
-  private predicate viableParamLambda(DataFlowCall call, ArgumentPosition apos, ParamNode p) {
-    exists(ParameterPosition ppos |
-      p.isParameterOf(viableCallableLambda(call, _), ppos) and
-      parameterMatch(ppos, apos)
-    )
+  pragma[noinline]
+  private predicate viableParamLambda(DataFlowCall call, ParameterPosition ppos, ParamNode p) {
+    p.isParameterOf(viableCallableLambda(call, _), ppos)
   }
 
   private predicate viableParamArgNonLambda(DataFlowCall call, ParamNode p, ArgNode arg) {
-    exists(ArgumentPosition pos |
-      viableParamNonLambda(call, pos, p) and
-      arg.argumentOf(call, pos)
+    exists(ParameterPosition ppos |
+      viableParamNonLambda(call, ppos, p) and
+      argumentPositionMatch(call, arg, ppos)
     )
   }
 
   private predicate viableParamArgLambda(DataFlowCall call, ParamNode p, ArgNode arg) {
-    exists(ArgumentPosition pos |
-      viableParamLambda(call, pos, p) and
-      arg.argumentOf(call, pos)
+    exists(ParameterPosition ppos |
+      viableParamLambda(call, ppos, p) and
+      argumentPositionMatch(call, arg, ppos)
     )
   }
 
--- a/cpp/ql/lib/semmle/code/cpp/dataflow/internal/DataFlowImplCommon.qll
+++ b/cpp/ql/lib/semmle/code/cpp/dataflow/internal/DataFlowImplCommon.qll
@@ -71,25 +71,31 @@ predicate accessPathCostLimits(int apLimit, int tupleLimit) {
  * calls. For this reason, we cannot reuse the code from `DataFlowImpl.qll` directly.
  */
 private module LambdaFlow {
-  private predicate viableParamNonLambda(DataFlowCall call, int i, ParamNode p) {
-    p.isParameterOf(viableCallable(call), i)
+  private predicate viableParamNonLambda(DataFlowCall call, ArgumentPosition apos, ParamNode p) {
+    exists(ParameterPosition ppos |
+      p.isParameterOf(viableCallable(call), ppos) and
+      parameterMatch(ppos, apos)
+    )
   }
 
-  private predicate viableParamLambda(DataFlowCall call, int i, ParamNode p) {
-    p.isParameterOf(viableCallableLambda(call, _), i)
+  private predicate viableParamLambda(DataFlowCall call, ArgumentPosition apos, ParamNode p) {
+    exists(ParameterPosition ppos |
+      p.isParameterOf(viableCallableLambda(call, _), ppos) and
+      parameterMatch(ppos, apos)
+    )
   }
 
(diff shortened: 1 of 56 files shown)


cpp/ql/lib/semmle/code/cpp/dataflow/internal/DataFlowImplCommon.qll
```diff
@@ -71,25 +71,31 @@ predicate accessPathCostLimits(int apLimit, int tupleLimit) {
  * calls. For this reason, we cannot reuse the code from `DataFlowImpl.qll` directly.
  */
 private module LambdaFlow {
-  private predicate viableParamNonLambda(DataFlowCall call, int i, ParamNode p) {
-    p.isParameterOf(viableCallable(call), i)
+  private predicate viableParamNonLambda(DataFlowCall call, ArgumentPosition apos, ParamNode p) {
+    exists(ParameterPosition ppos |
+      p.isParameterOf(viableCallable(call), ppos) and
+      parameterMatch(ppos, apos)
```
aschackmull, line 77: I'm not sure this is the right join-order, when `parameterMatch` isn't the identity.
I'd expect the following to do better:
```
private predicate argumentPosition(DataFlowCall call, ArgNode arg, ParameterPosition ppos) {
  exists(ArgumentPosition apos |
    arg.argumentOf(call, apos) and
    parameterMatch(ppos, apos)
  )
}
private predicate viableParamNonLambda(DataFlowCall call, ParameterPosition ppos, ParamNode p) {
  p.isParameterOf(viableCallable(call), ppos)
}
private predicate viableParamArgNonLambda(DataFlowCall call, ParamNode p, ArgNode arg) {
  exists(ParameterPosition ppos |
    viableParamNonLambda(call, ppos, p) and
    argumentPosition(call, arg, ppos)
  )
}
```

csharp/ql/lib/semmle/code/csharp/dataflow/internal/DataFlowPrivate.qll
```diff
@@ -1131,7 +1143,10 @@ private module ArgumentNodes {
     SummaryArgumentNode() { FlowSummaryImpl::Private::summaryArgumentNode(_, this, _) }
 
     override predicate argumentOf(DataFlowCall call, int pos) {
-      FlowSummaryImpl::Private::summaryArgumentNode(call, this, pos)
+      exists(ArgumentPosition apos |
+        FlowSummaryImpl::Private::summaryArgumentNode(call, this, apos) and
+        apos.getPosition() = pos
+      )
     }
```
yoff, line 1150: Should this still be using an `int` for `pos`?

cpp/ql/lib/semmle/code/cpp/dataflow/internal/DataFlowImplCommon.qll
hvitved: I agree that this i more safe; I have pushed updates.

csharp/ql/lib/semmle/code/csharp/dataflow/internal/DataFlowPrivate.qll
hvitved: No, this is purely an internal C# implementation predicate.

yoff (dismissed):
This feels good, I am looking forward to seeing how much of [ArgumentPassing](https://github.com/github/codeql/blob/main/python/ql/lib/semmle/python/dataflow/new/internal/DataFlowPrivate.qll#L323-L388) it will be able to replace :-)

hvitved:
> This feels good, I am looking forward to seeing how much of [ArgumentPassing](https://github.com/github/codeql/blob/main/python/ql/lib/semmle/python/dataflow/new/internal/DataFlowPrivate.qll#L323-L388) it will be able to replace :-)

Yeah, I was expecting that we should be able to improve that code. One of the key points is that we will no longer need the call graph (`connects` in `ArgumentPassing`, I believe) in order to map argument positions to parameter positions, which means we will be able to also handle the case where a call might target different functions with very different signatures (see examples in the PR description).

cpp/ql/lib/semmle/code/cpp/dataflow/internal/DataFlowImplCommon.qll
```diff
@@ -374,12 +390,15 @@ private module Cached {
...
   /**
-   * Holds if `p` is the `i`th parameter of a viable dispatch target of `call`.
-   * The instance parameter is considered to have index `-1`.
+   * Holds if `p` is the parameter of a viable dispatch target of `call`,
+   * and `p` matches arguments at position `apos`.
    */
   pragma[nomagic]
-  private predicate viableParam(DataFlowCall call, int i, ParamNode p) {
-    p.isParameterOf(viableCallableExt(call), i)
+  private predicate viableParam(DataFlowCall call, ArgumentPosition apos, ParamNode p) {
```
aschackmull, line 397: The same join-order rewrite applies here.

cpp/ql/lib/semmle/code/cpp/dataflow/internal/DataFlowImplCommon.qll
```diff
@@ -388,9 +407,9 @@ private module Cached {
    */
   cached
   predicate viableParamArg(DataFlowCall call, ParamNode p, ArgNode arg) {
-    exists(int i |
-      viableParam(call, i, p) and
-      arg.argumentOf(call, i) and
+    exists(ArgumentPosition pos |
+      viableParam(call, pos, p) and
+      arg.argumentOf(call, pos) and
```
aschackmull, line 412: You can reuse `argumentPositionMatch` here.

csharp/ql/lib/semmle/code/csharp/dataflow/internal/FlowSummaryImpl.qll
```diff
@@ -137,9 +143,13 @@ module Public {
     or
     noComponentSpecificCsv(sc) and
     (
-      exists(int i | sc = TParameterSummaryComponent(i) and result = "Parameter[" + i + "]")
+      exists(ArgumentPosition pos |
+        sc = TParameterSummaryComponent(pos) and result = "Parameter[" + pos + "]"
```
aschackmull, line 147: This mixes up the `toString` of `ArgumentPosition` with its csv representation, which seems slightly wrong. Perhaps `ArgumentPosition` ought to have a `toSpec` with suitable qldoc (and then the `toString` could alias that if need be) to clarify that its about more than the usual prettyprinting.
Same goes for `ParameterPosition`.
```suggestion
        sc = TParameterSummaryComponent(pos) and result = "Parameter[" + pos.toSpec() + "]"
```

csharp/ql/lib/semmle/code/csharp/dataflow/internal/FlowSummaryImplSpecific.qll
```diff
@@ -223,3 +221,19 @@ predicate interpretInputSpecific(string c, InterpretNode mid, InterpretNode n) {
     a.getUnboundDeclaration() = mid.asElement()
   )
 }
+
+/** Gets the argument position obtained by parsing `X` in `Parameter[X]`. */
+bindingset[s]
+ArgumentPosition parseParamBody(string s) {
+  result.getPosition() = s.regexpCapture("([-0-9]+)", 1).toInt()
```
aschackmull, line 228: Seems cleaner to have a separate private predicate that does the regex parsing from string to int, and then have `parseParamBody` and `parseArgBody` both use that.

docs/ql-libraries/dataflow/dataflow.md
```diff
@@ -148,23 +148,31 @@ methods, constructors, lambdas, etc.). It can also be useful to represent
...
-Flow into callables from arguments to parameters are matched up using an
-integer position, so these two predicates must be defined:
+Flow into callables from arguments to parameters are matched up using
+language-defined classes `ParameterPosition` and `ArgumentPosition`,
+so these three predicates must be defined:
 ```ql
-ArgumentNode::argumentOf(DataFlowCall call, int pos)
-predicate isParameterNode(ParameterNode p, DataFlowCallable c, int pos)
+/** Holds if `p` is a `ParameterNode` of `c` with position `pos`. */
+predicate isParameterNode(ParameterNodeImpl p, DataFlowCallable c, ParameterPosition pos)
```
aschackmull, line 168:
```suggestion
predicate isParameterNode(ParameterNode p, DataFlowCallable c, ParameterPosition pos)
```

aschackmull:
A few more inline comments, otherwise LGTM.

cpp/ql/lib/semmle/code/cpp/dataflow/internal/DataFlowImplCommon.qll
hvitved: Hmm, wonder how I missed that. Thanks.

csharp/ql/lib/semmle/code/csharp/dataflow/internal/FlowSummaryImpl.qll
hvitved: Good idea.

aschackmull (approved):

PR #7268, opened 2021-11-30
Java: Produce diffs for model generator changes

bmuskalla:
Introduces a new GitHub action that is setup for pull requests that contain changes to the model generator.

# PR usecase
It uses the `main` branch and the PR branch to produce two models for a variety of common libraries and shows the diff. The helps engineers working on the model generator to double-check the impact of their changes to the model generator.
By default, it is configured to generate models and their respective diffs for several apache commons projects as well as dubbo. The jobs are setup to not fail on changes as those are expected on PRs for the model generator. The results of the jobs are meant to be manually reviewed.

Screenshot 2021-11-30 at 12 28 10

The job is implemented as matrix job to parallelize the work of model generation. The current setup of models finishes within a few minutes.

The generated models and the diff reports are uploaded as Artifacts of the job.

Screenshot 2021-11-30 at 12 27 55

The `models.zip` archive contains all models (baseline and PR models), the `diff.zip` archive contains html reports of the diff between the two models.

Screenshot 2021-11-30 at 12 34 50

## Ad-hoc model usecase
The job can also be manually triggered for a specific project. This can be useful to let GitHub actions download a library database (anything that's available on lgtm.com) and generates a model for it. This is helpful to check the state of art on a specific library or to generate a model for ad-hoc security research.

Screenshot 2021-11-30 at 12 32 02


bmuskalla:
Here is a [sample job](https://github.com/bmuskalla/codeql/actions/runs/1520573734)

.github/workflows/java-model-diff.yml
```diff
@@ -0,0 +1,98 @@
...
+          SLUG: ${{ matrix.slugs }}
+        run: |
+          set -x
+          mkdir lib-dbs
+          SHORTNAME=${SLUG//[^a-zA-Z0-9_]/}
+          projectId=`curl -s https://lgtm.com/api/v1.0/projects/g/$SLUG | jq .id`
+          curl -L "https://lgtm.com/api/v1.0/snapshots/$projectId/java" -o $SHORTNAME.zip
+          unzip -q -d $SHORTNAME-db $SHORTNAME.zip
+          mkdir lib-dbs/$SHORTNAME/
+          mv $SHORTNAME-db/`ls -1 $SHORTNAME-db`/* lib-dbs/$SHORTNAME/
```
adityasharad, line 52: General recommendation: double-quote everywhere you're using variables or names controlled by the external file. Your replacement will ensure the short name won't have spaces, but it's good to be defensive.

I prefer to use `"$(...)"` over `` `...` `` to handle spaces in the output, but that is partly personal style.

.github/workflows/java-model-diff.yml
```diff
@@ -0,0 +1,98 @@
...
+jobs:
+  model-diff:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        # large
+        # ["google/guava", "spring-projects/spring-framework", "apache/poi"]
+        # others
+        # ["FasterXML/jackson-core", "FasterXML/jackson-databind", "google/gson", "JodaOrg/joda-time"
+        slugs: ${{fromJson(github.event.inputs.projects || '["apache/commons-codec", "apache/commons-io", "apache/commons-beanutils", "apache/commons-logging", "apache/commons-fileupload", "apache/commons-lang", "apache/commons-validator", "apache/commons-csv", "apache/dubbo"]' )}}
```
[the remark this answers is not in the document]
adityasharad: Minor: I would name this `slug` since you only get one for each matrix job, and can refer to it as `matrix.slug`.

.github/workflows/java-model-diff.yml
adityasharad: Also why do you have to unzip and move rather than unzipping straight into `lib-dbs/$SHORTNAME`?

.github/workflows/java-model-diff.yml
```diff
@@ -0,0 +1,98 @@
...
+          MODELS=`pwd`/tmp-models
+          MODE="baseline"
+          DATABASES=`pwd`/lib-dbs
+
+          analyzeDatabaseWithCheckout() {
+            QL_VARIANT=$1
+            DATABASE=$2
+            cd codeql-$QL_VARIANT
+            SHORTNAME=`basename $DATABASE`
+            python java/ql/src/utils/model-generator/GenerateFlowModel.py $DATABASE $MODELS/${SHORTNAME}.qll
```
adityasharad, line 66: Does this script do anything security-sensitive, such as uploading data?

.github/workflows/java-model-diff.yml
```diff
@@ -0,0 +1,98 @@
...
+          for d in $DATABASES/*/ ; do
+            ls -1 "$d"
+
+            analyzeDatabaseWithCheckout "baseline" $d
+            analyzeDatabaseWithCheckout "head" $d
+          done
+      - name: Generate Model Diff
+        run: |
+          set -x
+          npm install -g diff2html-cli
```
adityasharad, line 80: Minor: Pull this into its own step, so it's easier to spot in case it fails.

.github/workflows/java-model-diff.yml
```diff
@@ -0,0 +1,98 @@
...
+        # ["google/guava", "spring-projects/spring-framework", "apache/poi"]
+        # others
+        # ["FasterXML/jackson-core", "FasterXML/jackson-databind", "google/gson", "JodaOrg/joda-time"
+        slugs: ${{fromJson(github.event.inputs.projects || '["apache/commons-codec", "apache/commons-io", "apache/commons-beanutils", "apache/commons-logging", "apache/commons-fileupload", "apache/commons-lang", "apache/commons-validator", "apache/commons-csv", "apache/dubbo"]' )}}
+    steps:
+      - name: Clone self (github/codeql) for baseline
+        uses: actions/checkout@v2
+        with:
+          path: codeql-baseline
+          ref: ${{ github.base_ref }}
```
adityasharad, line 34: `github.base_ref` only exists for `pull_request` events. How do you want this to behave for manual runs or `push` events?

One option is that manual runs always compute the diff against `main`, and `push` runs don't attempt to compute a diff at all.

.github/workflows/java-model-diff.yml
```diff
@@ -0,0 +1,98 @@
...
+        required: true
+        default: '["netty/netty"]'
+  pull_request:
+    branches:
+      - main
+    paths:
+      - "java/ql/src/utils/model-generator/**/*.*"
+
+jobs:
+  model-diff:
```
adityasharad, line 20: Give this job a human-readable `name` too. It can be the same as the workflow name.

.github/workflows/java-model-diff.yml
```diff
@@ -0,0 +1,98 @@
...
+  push:
+    branches:
+      - main
+  workflow_dispatch:
+    inputs:
+      projects:
+        description: "The projects to generate models for"
+        required: true
+        default: '["netty/netty"]'
+  pull_request:
```
adityasharad, line 13: How strict do you want to keep this? Actions runs from external contributors will not start unless we manually approve them, but it's good to be careful. Should this run only for maintainers, or only from PRs based within this repository (no forks)?

.github/workflows/java-model-diff.yml
```diff
@@ -0,0 +1,98 @@
...
+          npm install -g diff2html-cli
+          MODELS=`pwd`/tmp-models
+          ls -1 tmp-models/
+          for m in $MODELS/*_baseline.qll ; do
+            t="${m/baseline/"head"}"
+            basename=`basename $m`
+            name="diff_${basename/_baseline.qll/""}"
+            (diff -w -u $m $t | diff2html -i stdin -F $MODELS/$name.html) || true
+          done
+      - uses: actions/upload-artifact@v2
```
adityasharad, line 89: PRs from forks get read-only tokens, so this might not work in that case.

.github/workflows/java-model-diff.yml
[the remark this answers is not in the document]
bmuskalla: The difference in default is intentional. We have essentially two usecases:
* PR feedback
* use the action to generate an adhoc model (that's why the dispatch default is just a simple example to showcase the format)
Removed the `push` trigger as that was only used for testing

.github/workflows/java-model-diff.yml
bmuskalla: Implemented skipping the diff for that case.

.github/workflows/java-model-diff.yml
bmuskalla: I can't seem to find any information about this. Let's try this out once we have the action setup. Worst case is that we need go back to `diff` in the build log :/

.github/workflows/java-model-diff.yml
bmuskalla: It only runs `codeql` and writes files

.github/workflows/java-model-diff.yml
```diff
@@ -0,0 +1,104 @@
...
+              analyzeDatabaseWithCheckout "head" $d
+            fi
+          done
+      - name: Install diff2html
+        run: |
+          npm install -g diff2html-cli
+      - name: Generate Model Diff
+        run: |
+          set -x
+          if [[ "$s1" == "$s2" ]]
```
adityasharad, line 82: Where are these set? I would find it easier to have `if: github.event.pull_request` in this Actions step, and drop the `if-then-else` from the Bash. Then the intent is obvious, and the UI will also show you that the step was skipped.

.github/workflows/java-model-diff.yml
```diff
@@ -0,0 +1,104 @@
...
+            python java/ql/src/utils/model-generator/GenerateFlowModel.py $DATABASE $MODELS/${SHORTNAME}.qll
+            mv $MODELS/${SHORTNAME}.qll $MODELS/${SHORTNAME}_${QL_VARIANT}.qll
+            cd ..
+          }
+
+          for d in $DATABASES/*/ ; do
+            ls -1 "$d"
+
+            analyzeDatabaseWithCheckout "baseline" $d
+            if [[ "$s1" != "$s2" ]]
```
adityasharad, line 71: Alternative: `"$GITHUB_EVENT_NAME" == "pull_request"`. Slightly less obscure IMO.

.github/workflows/java-model-diff.yml
adityasharad: Also skip the baseline checkout step if we're not in a PR run. `if: github.event.pull_request` should do it.

.github/workflows/java-model-diff.yml
```diff
@@ -0,0 +1,104 @@
...
+        uses: actions/checkout@v2
+        with:
+          path: codeql-head
+          ref: ${{ github.ref }}
+      - uses: ./codeql-baseline/.github/actions/fetch-codeql
+      - name: Download database
+        run: |
+          set -x
+          mkdir lib-dbs
+          SHORTNAME=${SLUG//[^a-zA-Z0-9_]/}
```
adityasharad, line 43: You haven't set `SLUG`: I think you need
```
env:
  SLUG: ${{ matrix.slug }}
```

.github/workflows/java-model-diff.yml
```diff
@@ -0,0 +1,104 @@
...
+          projectId=`curl -s https://lgtm.com/api/v1.0/projects/g/${matrix.slugs}} | jq .id`
```
adityasharad, line 44: After the above change.
```suggestion
          projectId=`curl -s "https://lgtm.com/api/v1.0/projects/g/${SLUG}" | jq .id`
```

.github/workflows/java-model-diff.yml
```diff
@@ -0,0 +1,104 @@
...
+          curl -L "https://lgtm.com/api/v1.0/snapshots/$projectId/java" -o $SHORTNAME.zip
+          unzip -q -d $SHORTNAME-db $SHORTNAME.zip
```
adityasharad, line 46: I think my suggestions about quoting have been lost.

adityasharad (commented):
General structure of the diffing mechanism looks sensible, thanks. A few more suggestions on the basis of that.

.github/workflows/java-model-diff.yml
```diff
@@ -0,0 +1,98 @@
...
+        default: '["netty/netty"]'
+  pull_request:
+    branches:
+      - main
+    paths:
+      - "java/ql/src/utils/model-generator/**/*.*"
+
+jobs:
+  model-diff:
+    runs-on: ubuntu-latest
```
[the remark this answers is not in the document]
bmuskalla: Thanks!

.github/workflows/java-model-diff.yml
bmuskalla: We restrict it to only this repo now. Generally, I think external contribtors might find this useful but it's too early to invest into that. Is there anything else we need to only run this for maintainers for now?

bmuskalla:
Friendly ping @adityasharad - let me know if you don't have capacity for this one and I'll let someone else take it from here.

.github/workflows/java-model-diff.yml
adityasharad: `if: github.repository == 'github/codeql'` is true even for PRs opened against this repo from forks. Let's chat separately so I can understand exactly which PRs you want this to run on.

adityasharad (commented):
Thanks for addressing the earlier comments. A few more observations, and one question I'll figure out with you live.

.github/workflows/java-model-diff.yml
bmuskalla: This is to handle the fact that we don't know the root folder of the zip archive (due to the naming pattern). With this indirection, we can move it to a known place (`lib-dbs/$SHORTNAME`)

bmuskalla:
As discussed offline, we'll move the job elsewhere.

.github/workflows/java-model-diff.yml
aibaars: @adityasharad Could you clarify you security concerns? The `pull_request` trigger is not particularly dangerous because pull requests from forks only run with (short-lived) read-only tokens.

bmuskalla:
Thanks @adityasharad !

PR #7269, opened 2021-11-30
C#: Update the make_stubs_nuget script

michaelnebel:
- Hardcode that we target the .NET 5.0 framework independently of what is the default on the machine, where the script is executed.
- Create a global.json file which hardcoded the use SDK to 5.0.402.
- Other minor changes.

### DIFF
--- a/csharp/ql/src/Stubs/make_stubs_nuget.py
+++ b/csharp/ql/src/Stubs/make_stubs_nuget.py
@@ -109,7 +109,7 @@ with open(jsonFile) as json_data:
 print("\n --> Generated stub files: " + rawSrcOutputDir)
 
 print("\n* Formatting files")
-run_cmd(['dotnet', 'format', rawSrcOutputDir], workDir)
+run_cmd(['dotnet', 'format', rawSrcOutputDir])
 
 print("\n --> Generated (formatted) stub files: " + rawSrcOutputDir)
 
--- a/csharp/ql/src/Stubs/make_stubs_nuget.py
+++ b/csharp/ql/src/Stubs/make_stubs_nuget.py
@@ -34,6 +34,9 @@ workDir = os.path.abspath(helpers.get_argv(3, "tempDir"))
 projectNameIn = "input"
 projectDirIn = os.path.join(workDir, projectNameIn)
 
+def run_cmd(cmd, msg="Failed to run command"):
+    helpers.run_cmd_cwd(cmd, workDir, msg)
+
 # /output contains the output of the stub generation
 outputDirName = "output"
 outputDir = os.path.join(workDir, outputDirName)
@@ -57,7 +60,7 @@ jsonFile = os.path.join(rawOutputDir, outputName + '.json')
 version = helpers.get_argv(2, "latest")
 
 print("\n* Creating new input project")
-helpers.run_cmd(['dotnet', 'new', 'classlib', "-f", "net5.0", "--language", "C#", '--name',
+run_cmd(['dotnet', 'new', 'classlib', "-f", "net5.0", "--language", "C#", '--name',
                  projectNameIn, '--output', projectDirIn])
 helpers.remove_files(projectDirIn, '.cs')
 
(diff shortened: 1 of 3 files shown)


csharp/ql/src/Stubs/make_stubs_nuget.py
```diff
@@ -68,9 +68,13 @@ def write_csproj_prefix(ioWrapper):
     cmd.append(version)
 helpers.run_cmd(cmd)
 
+sdk_version = '5.0.402'
+print("\n* Creating new global.json file and setting SDK to " + sdk_version)
+helpers.run_cmd(['dotnet', 'new', 'globaljson', '--force', '--sdk-version', sdk_version])
```
hvitved, line 73: Shouldn't this command somehow mention `projectDirIn`?

csharp/ql/test/resources/stubs/_frameworks/Microsoft.NETCore.App/System.Runtime.cs
```diff
@@ -5994,7 +5994,7 @@ public class Stopwatch
 
         namespace CodeAnalysis
         {
-            // Generated from `System.Diagnostics.CodeAnalysis.AllowNullAttribute` in `System.Diagnostics.DiagnosticSource, Version=5.0.0.0, Culture=neutral, PublicKeyToken=cc7b13ffcd2ddd51; System.Formats.Asn1, Version=5.0.0.0, Culture=neutral, PublicKeyToken=cc7b13ffcd2ddd51; System.Runtime, Version=5.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a; System.Threading.Tasks.Dataflow, Version=5.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a`
+            // Generated from `System.Diagnostics.CodeAnalysis.AllowNullAttribute` in `Newtonsoft.Json, Version=13.0.0.0, Culture=neutral, PublicKeyToken=30ad4fe6b2a6aeed; System.Diagnostics.DiagnosticSource, Version=5.0.0.0, Culture=neutral, PublicKeyToken=cc7b13ffcd2ddd51; System.Formats.Asn1, Version=5.0.0.0, Culture=neutral, PublicKeyToken=cc7b13ffcd2ddd51; System.Runtime, Version=5.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a; System.Threading.Tasks.Dataflow, Version=5.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a`
```
hvitved, line 5997: This looks a bit weird, what do you think @tamasvajk ?

csharp/ql/src/Stubs/make_stubs_nuget.py
michaelnebel: dotnet looks traverses the directory structure upwards towards the root from the current work dir to find a global.json file.
We need to change the working directory to the tempDir and store the global.json here as we execute dotnet commands against multiple subfolders in the the tempDir.

michaelnebel: @hvitved : But you are right - without any changes we will start poluting the folder from which the nuget script is executed.

csharp/ql/test/resources/stubs/_frameworks/Microsoft.NETCore.App/System.Runtime.cs
tamasvajk: Yes, I think we would manually need to touch up the files in `_frameworks` folder, and remove any mention of `Newtonsoft.Json`.

What happens here is that `System.Diagnostics.CodeAnalysis.AllowNullAttribute` is defined in multiple assemblies, and one of them happens to be `Newtonsoft.Json`, the others are all in the runtime.

csharp/ql/src/Stubs/make_stubs_nuget.py
```diff
@@ -106,7 +109,7 @@ def write_csproj_prefix(ioWrapper):
 print("\n --> Generated stub files: " + rawSrcOutputDir)
 
 print("\n* Formatting files")
-helpers.run_cmd(['dotnet', 'format', rawSrcOutputDir])
+run_cmd(['dotnet', 'format', rawSrcOutputDir], workDir)
```
hvitved, line 112: spurious `workDir` arg?

csharp/ql/test/resources/stubs/_frameworks/Microsoft.NETCore.App/System.Runtime.cs
michaelnebel: Alright. I will just roll-back the changes made in this file as the only difference between the previous version and this one is the mentioning of Newtonsoft.Json.

csharp/ql/src/Stubs/make_stubs_nuget.py
michaelnebel: Very nice catch @hvitved! Now the directory would be printed as error message... Will fix!

hvitved (approved):

PR #7270, opened 2021-11-30
Dataflow: Stage 2 refactor

aschackmull:
Minor refactor that makes stage 2 binary equal to stages 3 and 4. This makes future maintenance easier and is a prerequisite for a future parameterized module refactor of stages 2-4.

Quick local perf check and RA comparison showed only a very minor cost. (And some further incentive for better compiler elimination of constant columns).


hvitved (approved):
Great to have these be identical.

PR #7272, opened 2021-11-30
C++: refactor buffer overwrite queries with estimate reasons

redsun82:
This is a first step towards having different buffer overwrite queries with different precisions. The idea is to add a reason for why an estimate was given, allowing to detect when exact range analysis kicks in, in order to split out the current `OverrunWrite.ql` into two separate high precision and medium precision queries.

Next steps will be to apply the same idea to the float special case (possibly applying range analysis there as well) and actually splitting the query in two.

### DIFF
--- a/cpp/ql/lib/semmle/code/cpp/commons/Printf.qll
+++ b/cpp/ql/lib/semmle/code/cpp/commons/Printf.qll
@@ -9,6 +9,30 @@ import semmle.code.cpp.models.interfaces.FormattingFunction
 private import semmle.code.cpp.rangeanalysis.SimpleRangeAnalysis
 private import semmle.code.cpp.rangeanalysis.RangeAnalysisUtils
 
+private newtype TBufferWriteEstimationReason =
+  TTypeBoundsAnalysis() or
+  TValueFlowAnalysis()
+
+class BufferWriteEstimationReason extends TBufferWriteEstimationReason {
+  BufferWriteEstimationReason() {
+    this = TTypeBoundsAnalysis() or
+    this = TValueFlowAnalysis()
+  }
+
+  string toString() {
+    this = TTypeBoundsAnalysis() and result = "based on type bounds" or
+    this = TValueFlowAnalysis() and result = "based on flow analysis of value bounds"
+  }
+
+  BufferWriteEstimationReason combineWith(BufferWriteEstimationReason other) {
+    (this = TTypeBoundsAnalysis() or other = TTypeBoundsAnalysis()) and result = TTypeBoundsAnalysis() or
+    (this = TValueFlowAnalysis() and other = TValueFlowAnalysis()) and result = TValueFlowAnalysis()
+  }
+}
+
+BufferWriteEstimationReason typeBoundsAnalysis() { result = TTypeBoundsAnalysis() }
+BufferWriteEstimationReason valueFlowAnalysis() { result = TValueFlowAnalysis() }
+
 class PrintfFormatAttribute extends FormatAttribute {
   PrintfFormatAttribute() { this.getArchetype() = ["printf", "__printf__"] }
 }
@@ -990,7 +1014,7 @@ class FormatLiteral extends Literal {
    * conversion specifier of this format string; has no result if this cannot
    * be determined.
    */
-  int getMaxConvertedLength(int n) {
+  int getMaxConvertedLength(int n, BufferWriteEstimationReason reason) {
     exists(int len |
       (
         (
@@ -1002,10 +1026,10 @@ class FormatLiteral extends Literal {
       ) and
       (
         this.getConversionChar(n) = "%" and
-        len = 1
+        len = 1 and reason = TValueFlowAnalysis()
         or
         this.getConversionChar(n).toLowerCase() = "c" and
-        len = 1 // e.g. 'a'
+        len = 1 and reason = TValueFlowAnalysis() // e.g. 'a'
         or
         this.getConversionChar(n).toLowerCase() = "f" and
         exists(int dot, int afterdot |
@@ -1019,7 +1043,7 @@ class FormatLiteral extends Literal {
             afterdot = 6
           ) and
           len = 1 + 309 + dot + afterdot
-        ) // e.g. -1e308="-100000"...
+        ) and reason = TTypeBoundsAnalysis() // e.g. -1e308="-100000"...
         or
         this.getConversionChar(n).toLowerCase() = "e" and
         exists(int dot, int afterdot |
@@ -1033,7 +1057,7 @@ class FormatLiteral extends Literal {
             afterdot = 6
           ) and
           len = 1 + 1 + dot + afterdot + 1 + 1 + 3
-        ) // -1e308="-1.000000e+308"
+        ) and reason = TTypeBoundsAnalysis() // -1e308="-1.000000e+308"
         or
         this.getConversionChar(n).toLowerCase() = "g" and
         exists(int dot, int afterdot |
@@ -1056,67 +1080,78 @@ class FormatLiteral extends Literal {
           //       (e.g. 123456, 0.000123456 are just OK)
           //       so case %f can be at most P characters + 4 zeroes, sign, dot = P + 6
           len = (afterdot.maximum(1) + 6).maximum(1 + 1 + dot + afterdot + 1 + 1 + 3)
-        ) // (e.g. "-1.59203e-319")
+        ) and reason = TTypeBoundsAnalysis() // (e.g. "-1.59203e-319")
         or
         this.getConversionChar(n).toLowerCase() = ["d", "i"] and
         // e.g. -2^31 = "-2147483648"
-        len =
-          min(float cand |
+        exists(float typeBasedBound, float valueBasedBound |
             // The first case handles length sub-specifiers
             // Subtract one in the exponent because one bit is for the sign.
             // Add 1 to account for the possible sign in the output.
-            cand = 1 + lengthInBase10(2.pow(this.getIntegralDisplayType(n).getSize() * 8 - 1))
-            or
+            typeBasedBound = 1 + lengthInBase10(2.pow(this.getIntegralDisplayType(n).getSize() * 8 - 1)) and
             // The second case uses range analysis to deduce a length that's shorter than the length
             // of the number -2^31.
-            exists(Expr arg, float lower, float upper |
+            exists(Expr arg, float lower, float upper, float typeLower, float typeUpper |
               arg = this.getUse().getConversionArgument(n) and
               lower = lowerBound(arg.getFullyConverted()) and
-              upper = upperBound(arg.getFullyConverted())
+              upper = upperBound(arg.getFullyConverted()) and
+              typeLower = exprMinVal(arg.getFullyConverted()) and
+              typeUpper = exprMaxVal(arg.getFullyConverted()) 
             |
-              cand =
-                max(int cand0 |
+              valueBasedBound =
+                max(int cand |
                   // Include the sign bit in the length if it can be negative
                   (
                     if lower < 0
-                    then cand0 = 1 + lengthInBase10(lower.abs())
-                    else cand0 = lengthInBase10(lower)
+                    then cand = 1 + lengthInBase10(lower.abs())
+                    else cand = lengthInBase10(lower)
                   )
                   or
                   (
                     if upper < 0
-                    then cand0 = 1 + lengthInBase10(upper.abs())
-                    else cand0 = lengthInBase10(upper)
+                    then cand = 1 + lengthInBase10(upper.abs())
+                    else cand = lengthInBase10(upper)
                   )
-                )
-            )
-          )
+                ) and
+              (
+                if lower > typeLower or upper < typeUpper
+                then reason = TValueFlowAnalysis()
+                else reason = TTypeBoundsAnalysis()
+              )
+            ) and
+            len = valueBasedBound.minimum(typeBasedBound)
+        )
         or
         this.getConversionChar(n).toLowerCase() = "u" and
         // e.g. 2^32 - 1 = "4294967295"
-        len =
-          min(float cand |
+        exists(float typeBasedBound, float valueBasedBound |
             // The first case handles length sub-specifiers
-            cand = 2.pow(this.getIntegralDisplayType(n).getSize() * 8)
-            or
+            typeBasedBound = lengthInBase10(2.pow(this.getIntegralDisplayType(n).getSize() * 8) - 1) and
             // The second case uses range analysis to deduce a length that's shorter than
             // the length of the number 2^31 - 1.
-            exists(Expr arg, float lower |
+            exists(Expr arg, float lower, float upper, float typeLower, float typeUpper |
               arg = this.getUse().getConversionArgument(n) and
-              lower = lowerBound(arg.getFullyConverted())
+              lower = lowerBound(arg.getFullyConverted()) and
+              upper = upperBound(arg.getFullyConverted()) and
+              typeLower = exprMinVal(arg.getFullyConverted()) and
+              typeUpper = exprMaxVal(arg.getFullyConverted()) 
             |
-              cand =
-                max(float cand0 |
+              valueBasedBound =
+                lengthInBase10(max(float cand |
                   // If lower can be negative we use `(unsigned)-1` as the candidate value.
                   lower < 0 and
-                  cand0 = 2.pow(any(IntType t | t.isUnsigned()).getSize() * 8)
+                  cand = 2.pow(any(IntType t | t.isUnsigned()).getSize() * 8)
                   or
-                  cand0 = upperBound(arg.getFullyConverted())
-                )
-            )
-          |
-            lengthInBase10(cand)
-          )
+                  cand = upper
+                )) and
+              (
+                if lower > typeLower or upper < typeUpper
+                then reason = TValueFlowAnalysis()
+                else reason = TTypeBoundsAnalysis()
+              )
+            ) and
+            len = valueBasedBound.minimum(typeBasedBound)
+        )
         or
         this.getConversionChar(n).toLowerCase() = "x" and
         // e.g. "12345678"
@@ -1135,7 +1170,7 @@ class FormatLiteral extends Literal {
           (
             if this.hasAlternateFlag(n) then len = 2 + baseLen else len = baseLen // "0x"
           )
-        )
+        ) and reason = TTypeBoundsAnalysis()
         or
         this.getConversionChar(n).toLowerCase() = "p" and
         exists(PointerType ptrType, int baseLen |
@@ -1144,7 +1179,7 @@ class FormatLiteral extends Literal {
           (
             if this.hasAlternateFlag(n) then len = 2 + baseLen else len = baseLen // "0x"
           )
-        )
+        ) and reason = TValueFlowAnalysis()
         or
         this.getConversionChar(n).toLowerCase() = "o" and
         // e.g. 2^32 - 1 = "37777777777"
@@ -1163,14 +1198,15 @@ class FormatLiteral extends Literal {
           (
             if this.hasAlternateFlag(n) then len = 1 + baseLen else len = baseLen // "0"
           )
-        )
+        ) and reason = TTypeBoundsAnalysis()
         or
         this.getConversionChar(n).toLowerCase() = "s" and
         len =
           min(int v |
             v = this.getPrecision(n) or
             v = this.getUse().getFormatArgument(n).(AnalysedString).getMaxLength() - 1 // (don't count null terminator)
-          )
+          ) and
+        reason = TValueFlowAnalysis()
       )
     )
   }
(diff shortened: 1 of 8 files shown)


redsun82:
Juist a few words of context: I wanted to have a kind of enum splitting the buffer size estimation results. Something like
* best effort: normal size calculations on non-sprintf like functions, no meaningful range analysis possible on sprintf, `%f` capped to 8
* extreme floats: only giving a size on sprintf without exact range analysis, taking into account the true maximum width of double
* range analysis: only giving a size on sprintf with exact range analysis
In particular I'd like to avoid the kind of code that in the `cpp/overrunning-write-with-float` avoids the query overlap, as with three different queries (maybe more in the future) it'd start to be a bit ugly I think. So I want one method accepting this strategy enum and giving disjoint values (so, no value if the strategy does not apply to a specific case), so that the three corresponding queries would be written by just comparing this value.

For the moment I just want to apply this reasoning to the existing two cases, but this particular PR is not getting me there yet though.

cpp/ql/lib/semmle/code/cpp/commons/Printf.qll
```diff
@@ -990,7 +994,8 @@ class FormatLiteral extends Literal {
    * conversion specifier of this format string; has no result if this cannot
    * be determined.
    */
-  int getMaxConvertedLength(int n) {
+  int getMaxConvertedLength(int n, TFormattedSizeEstimateKind strategy, boolean strategyRelevant) {
+    (strategyRelevant = true or strategyRelevant = false) and // avoid "is not bound"...
```
MathiasVP, line 998: The "is not bound" warning is really there for a good reason: You're introducing a cartesian product: previously, the `getMaxConvertedLength` table would have 2 (ignoring the `this` argument) columns: `(result, n)`, with the interpretation that the `n`'th conversation can product a string of length at most `result`.

Now the table will look like `(result, n, strategy, strategyRelevant)`, where many results will have:
- A `(result, n, strategy, true)` row, and
- A `(result, n, strategy, false)` row

The `strategy` argument is actually also introducing a cartesian product, but unlike booleans, the compiler doesn't emit an error for this. If we're not in the `this.getConversionChar(n).toLowerCase() = "f"` case, then every previous row of the form `(result, n)` will now have both:
- A `(result, n, TFormattedSizeWithExtremeFloat(), strategyRelevant)` row, and
- A `(result, n, TFormattedSizeLimited(), strategyRelevant)` row.

We should either figure out how to make sure that we don't duplicate rows (i.e., by making sure that every argument is bound in every disjunction), or avoid the need for this change.

MathiasVP (commented):
Let me know if something isn't clear in what I wrote. This is the kind of change that would make total sense in most other languages, but in QL it will be difficult to implement.

cpp/ql/lib/semmle/code/cpp/commons/Printf.qll
```diff
@@ -1225,35 +1225,30 @@ class FormatLiteral extends Literal {
...
   }
 
-  private int getMaxConvertedLengthAfterLimited(int n) {
-    if n = this.getNumConvSpec()
-    then result = this.getConstantSuffix().length() + 1
-    else
-      result =
-        this.getConstantPart(n).length() + this.getMaxConvertedLengthLimited(n) +
-          this.getMaxConvertedLengthAfterLimited(n + 1)
+  int getMaxConvertedLength(TFormattedSizeEstimateKind strategy) {
```
geoffw0, line 1247: This is part of the public interface, so it should have a QLDoc comment.  Also, we should preserve the original `getMaxConvertedLength()` and `getMaxConvertedLengthLimited()` predicates for backwards compatibility, even if all they do is call into the new one (and you can deprecate them if you like).

cpp/ql/src/Security/CWE/CWE-120/OverrunWriteFloat.ql
```diff
@@ -26,9 +26,9 @@ where
   not bw.hasExplicitLimit() and
   // has no explicit size limit
   destSize = getBufferSize(bw.getDest(), _) and
-  bw.getMaxData() > destSize and
   // and we can deduce that too much data may be copied
-  bw.getMaxDataLimited() <= destSize // but it would fit without long '%f' conversions
+  bw.getMaxDataEstimate(TFormattedSizeWithExtremeFloat()) > destSize
```
geoffw0, line 30: You seem to have lost the line:
```
bw.getMaxDataLimited() <= destSize // but it would fit without long '%f' conversions
```

cpp/ql/lib/semmle/code/cpp/security/BufferWrite.qll
```diff
@@ -65,19 +65,7 @@ abstract class BufferWrite extends Expr {
...
-  int getMaxData() { none() }
-
-  /**
-   * Gets an upper bound to the amount of data that's being written (if one
-   * can be found), except that float to string conversions are assumed to be
-   * much smaller (8 bytes) than their true maximum length.  This can be
-   * helpful in determining the cause of a buffer overflow issue.
-   */
-  int getMaxDataLimited() { result = this.getMaxData() }
+  int getMaxDataEstimate(TFormattedSizeEstimateKind strategy) { none() }
```
geoffw0, line 68: `getMaxDataLimited()` had a default behaviour `result = this.getMaxData()`, which is used in some of the subclasses.

geoffw0 (commented):
This PR is great cleanup on its own, and should make it easier to add the third strategy.  :+1:

cpp/ql/lib/semmle/code/cpp/commons/Printf.qll
redsun82: I see, I will go back on this tomorrow morning, but this makes it more clear. I think I will try to make something similar and have no duplicate rows.

MathiasVP: Sounds good!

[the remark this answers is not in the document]
redsun82: my idea was kind of make the relevance of the strategy bubble up from the `getMaxConvertedLength` predicate. However I now see that the extreme float and the exact range analysis cases actually are different, in that the former is relevant if any of the conversions is a float, while the latter would be relevant if all conversions where sufficiently high precision.

I'll have to think about it

cpp/ql/src/Security/CWE/CWE-120/OverrunWriteFloat.ql
redsun82: my initial idea was not needing that line at all, but I understood in the meantime this was not possible, unless the buffer size itself was somehow injected down to the `FormatLiteral` methods doing the estimation. I'll get back to it

geoffw0: That line is there to avoid duplicating results that the `OverrunWrite.ql` query would report anyway.  It's necessary for a clean result set.

redsun82: yes, I'm aware, I hoped to not have to write that line by playing with non-existing results, but that led me nowhere :smile:

redsun82:
ok, here is a further take. Putting the special float case in was proving challenging, so I'm leaving it aside for the moment. From the tests it seems like the split is kind of working ok, but I'm eager to get some early comments on the logic behind it.

cpp/ql/lib/semmle/code/cpp/security/BufferWrite.qll
```diff
@@ -69,15 +69,30 @@ abstract class BufferWrite extends Expr {
    * Gets an upper bound to the amount of data that's being written (if one
    * can be found).
    */
-  int getMaxData() { none() }
+  int getMaxData() { result = max(getMaxData(_)) }
```
MathiasVP, line 72: Is this change really backward compatible? Previously, if a customer has written their own class that extends `BufferWrite` and implemented `getMaxData()` like so:
```ql
class MyWrite extends BufferWrite {
  MyWrite() { ... }

  override int getMaxData() { ... }
}
```
then our existing queries correctly pick up these classes and use their `getMaxData()` implementations.

But after this PR, they now override a `getMaxData()` that already has an implementation, but they now have a class where `getMaxData(BufferWriteEstimationReason reason)` implemented as `none()`. And so our queries (that now use the `getMaxData(BufferWriteEstimationReason reason)` version) won't pick up their class.

Does that make sense? Any idea how we can avoid this issue?

cpp/ql/src/Security/CWE/CWE-120/OverrunWrite.ql
```diff
@@ -21,14 +21,14 @@ import semmle.code.cpp.commons.Alloc
...
   dest = bw.getDest() and
   destSize = getBufferSize(dest, _) and
   // we can deduce that too much data may be copied (even without
   // long '%f' conversions)
-  bw.getMaxDataLimited() > destSize
+  bw.getMaxDataLimited(reason) > destSize
 select bw,
   "This '" + bw.getBWDesc() + "' operation requires " + bw.getMaxData() +
-    " bytes but the destination is only " + destSize + " bytes."
+    " bytes but the destination is only " + destSize + " bytes (" + reason.toString() + ")."
```
MathiasVP, line 34: Do we really want to write out this `reason` in the description? I'm guessing users don't really care if we obtained a bound using type-based bounds or range analysis-bounds?

MathiasVP (commented):
I like this approach much better than the first one :)

cpp/ql/lib/semmle/code/cpp/security/BufferWrite.qll
redsun82: would it be too rude to mark `getMaxData` as `final`, at leat to not silently not working?

cpp/ql/src/Security/CWE/CWE-120/OverrunWrite.ql
redsun82: well, the range analysis case will be moved to a separate query, but the plan is to have the float case here as well, then it will probably be good to point that out. But we could be less wordy when we get to that

cpp/ql/lib/semmle/code/cpp/commons/Printf.qll
```diff
@@ -1056,67 +1119,80 @@ class FormatLiteral extends Literal {
...
-            // the length of the number 2^31 - 1.
-            exists(Expr arg, float lower |
-              arg = this.getUse().getConversionArgument(n) and
-              lower = lowerBound(arg.getFullyConverted())
-            |
-              cand =
-                max(float cand0 |
+        exists(float typeBasedBound, float valueBasedBound |
+          // The first case handles length sub-specifiers
+          typeBasedBound = lengthInBase10(2.pow(this.getIntegralDisplayType(n).getSize() * 8) - 1) and
```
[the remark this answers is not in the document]
redsun82: I'm subtracting 1 from the power result, not the power exponent, so it's reflecting the actual maximum value for the unsigned int, which is `2^width - 1`. Does not make a difference in decimal, but I see it could create a problem if we copy and pasted the logic for the hex or octal cases.

cpp/ql/lib/semmle/code/cpp/security/BufferWrite.qll
MathiasVP: I hope we can avoid breaking customer's code like that, but that's at least better than silently removing alerts🤞. In any case, we should probably mark `getMaxData` as `deprecated`.

How about we implement a default version of `getMaxData(BufferWriteEstimationReason reason)` in terms of `getMaxData()` and either:
- Default the reason as `TypeBoundsAnalysis`, or
- Add a dummy reason IPA branch for this case?

MathiasVP: Hm, or maybe this will actually create non-monotonic recursion? 🤔. Let's see what happens!

cpp/ql/src/Security/CWE/CWE-120/OverrunWrite.ql
MathiasVP: Why would it be important for the float case?

cpp/ql/lib/semmle/code/cpp/commons/Printf.qll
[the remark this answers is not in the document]
redsun82: well, it did not have really an effect, as no power of 10 can also be an exact power of 2, so the number of digits of `2^x - 1` is always equal to the number of digits of `2^x`. It would be visible if we used the same kind of code for hex or octal, which we don't yet (and probably we don't need to, as there is a more direct and easily computable relation between number of bits and digits there)

But still, it was raising an eyebrow when reading the code (at least raising mine :raised_eyebrow: :smile: )

cpp/ql/lib/semmle/code/cpp/security/BufferWrite.qll
redsun82: Ok, I added a `NoSpecifiedReason` case used only for this. The current state of the PR will generate some warnings during library compilation, but no warnings in existing queries as far as I can tell from the tests.

cpp/ql/src/Security/CWE/CWE-120/OverrunWrite.ql
redsun82: I think for the same reason that we created the special float case query in the first place. Many might not know that a `%f` might have such extreme conversion lengths, so when they see that a format string containing "%f" is reported to potentially be hundreds of bytes long they might get surprised. I know I did not know that before (lucky me, I actually never needed to program with floating point numbers, ever :sweat_smile:).

Having something in the query result explicitly pointing to that fact might clear it up instantly.

cpp/ql/lib/semmle/code/cpp/security/BufferWrite.qll
MathiasVP: Sounds good. Is the warning because we now override a `deprecated` predicate?

cpp/ql/lib/semmle/code/cpp/commons/Printf.qll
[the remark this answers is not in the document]
MathiasVP: Ah, right. I see. Good catch!

cpp/ql/src/Security/CWE/CWE-120/OverrunWrite.ql
MathiasVP: I might be misunderstanding, but isn't the text in parentheses either:
- "based on type bounds", or
- "based on flow analysis of value bounds"
So is the idea that when programmers see the string "(based on type bounds)", they will know that we deduced a potential overflow due to floating-point specifiers?

cpp/ql/lib/semmle/code/cpp/security/BufferWrite.qll
redsun82: yes, in order to make a client class that override `getMaxData\0` work with our queries that use `getMaxData\1` now, the default implementation of `getMaxData\1` must call deprecated `getMaxData\0`. So when compiling the library a warning is produced. However the warnings are not produced when compiling/running the queries themselves.

cpp/ql/src/Security/CWE/CWE-120/OverrunWrite.ql
redsun82: sorry, I think I confused you :smile: Currently no, floating point estimates are not included in this query. I was speaking about plans for the near future, when the `based on flow analysis` reason will not be visible in this query as it will be moved to another high precision query, but we will then move here the float case. We can maybe then tweak the output then, when we cross that bridge.

So, I can remove this altogether now. It was quite useful to check the tests now, to look whether the reason was the correct one without changing the query results just yet. And also, it helped out spot one case in an earlier stage of development that was giving multiple reasons (without explicitly using the reason in the `select` part this was hidden in the output), but it's not really needed before the next steps.

redsun82: (which by the way will make this PR independent of the internal one)

MathiasVP: I see. Thanks for clearing up my confusion. Yeah, I think it's best to avoid this line currently. We might manage to do a dist upgrade on LGTM before the high-precision query work is done, and in that case, I don't think the message will be useful to programmers.

cpp/ql/lib/semmle/code/cpp/security/BufferWrite.qll
```diff
@@ -68,16 +68,37 @@ abstract class BufferWrite extends Expr {
   /**
    * Gets an upper bound to the amount of data that's being written (if one
    * can be found).
+   * DEPRECATED: getMaxData\1 should be used and overridden instead
```
MathiasVP, line 71: I think the syntax is:
```suggestion
   * DEPRECATED: getMaxData/1 should be used and overridden instead
```
And similarly for the other deprecation messages.

cpp/ql/lib/semmle/code/cpp/security/BufferWrite.qll
MathiasVP: Yeah, that makes sense. I guess that's okay.

cpp/ql/lib/semmle/code/cpp/security/BufferWrite.qll
redsun82: ok!

cpp/ql/lib/semmle/code/cpp/security/BufferWrite.qll
redsun82: hmm, actually it turns out it's not ok for CI to have warnings in the libraries :frowning_face: Any idea what to do @MathiasVP ? It seems there is no way we can deprecate a predicate that could be overridden by clients and both make those overrides still work as expected and not use deprecated predicates in our own library code without a warning... Are there some secret pragmas to suppress warnings maybe?

MathiasVP: As far as I know, there's no such secret pragma (but maybe there should be). 🤔

MathiasVP: One easy solution is to _not_ use the deprecated keyword for the old `getMaxData` predicate.

cpp/ql/lib/semmle/code/cpp/commons/Printf.qll
```diff
@@ -9,6 +9,83 @@ import semmle.code.cpp.models.interfaces.FormattingFunction
...
+  /**
+   * Combine estimate reasons. Used to give a reason for the size of a format string
+   * conversion given reasons coming from its individual specifiers.
+   */
+  abstract BufferWriteEstimationReason combineWith(BufferWriteEstimationReason other);
+}
+
+/**
+ * No particular reason given. This is currently used for backward compatibility so that
+ * classes derived from BufferWrite and overriding getMaxData\0 still work with the
```
MathiasVP, line 40:
```suggestion
 * classes derived from BufferWrite and overriding `getMaxData/0` still work with the
```

geoffw0:
The DCA run:
 - no significant performance changes.
 - no query result changes :tada: (I checked that `cpp/overrunning-write-with-float` was run, amongst others).
 - `microsoft__calculator` failed to build; but I would ignore this, its been flaky lately.

MathiasVP:
> * `microsoft__calculator` failed to build; but I would ignore this, its been flaky lately.

FWIW, it's failing because it's using C++/CX. It's not really flaky as much as it's using a dialect of C++ that we don't fully support.

redsun82:
> The DCA run:
> 
> * no significant performance changes.
> * no query result changes tada (I checked that `cpp/overrunning-write-with-float` was run, amongst others).
> * `microsoft__calculator` failed to build; but I would ignore this, its been flaky lately.

thanks for that @geoffw0 . So, shall we merge this?

geoffw0:
> shall we merge this?

I haven't kept up with the changes since my initial round of comments (I believe all of those have been addressed apart from adding explanation around the `newtype`, which is optional).  But this is a good sign.

MathiasVP:
> thanks for that @geoffw0 . So, shall we merge this?

I've added a review comment that hasn't been fixed yet :) It's an easy one!

redsun82:
> > thanks for that @geoffw0 . So, shall we merge this?
> 
> I've added a review comment that hasn't been fixed yet :) It's an easy one!

Uh, I had totally missed that, fixed now :smile: I wish there was a view showing only unresolved comments... or is there?

> As far as I can tell, there's also been a non-trivial amount of changes since the last DCA: https://github.com/github/codeql/compare/5ed705670752bda84aa6d53c562a60b04807d1c5..d88e4d60b57cc1d49f8702fee4162ade0ecee795#diff-303d2f7d1a901ce74c795fa2b729115169b2cf4c21678522f7cd9be9c3c0d8f5
> 
> I hope it's okay that I took the liberty to start another DCA run. Once that finishes (and my final lonely comment has been addressed) I'll be happy to merge it. Sorry I didn't raise this doubt on Slack earlier today. I didn't check which SHA the DCA experiment had actually run on.

Thanks for that!

redsun82:
It seems the DCA experiment is ok, right?

MathiasVP:
> It seems the DCA experiment is ok, right?

Indeed, all looks good :)

MathiasVP (approved):
LGTM!

PR #7273, opened 2021-11-30
Ruby: add CryptoAlgorithms library

alexrford:
This PR copies `CryptoAlgorithms.qll` to Ruby and extends it with some additional encryption algorithms (`ARIA`, `IDEA`, `SM4`, `SEED`, `GOST`, `CAST`, `CHACHA`, `CAMELLIA` - all classified as strong) and block modes (`ECB` - classified as weak, and `CBC`,  `CCM`, `CFB`, `CTR`, `GCM`, `OFB` - classified as strong).

It then adds a new library, `OpenSSL.qll`, for Ruby that models parts of the [`OpenSSL::Cipher`](https://ruby-doc.org/stdlib-3.0.1/libdoc/openssl/rdoc/OpenSSL/Cipher.html) library component. The aim here is mostly to be able to take a `string cipherName` that can be used to construct an `OpenSSL::Cipher` instance, s.t. we can pass judgement on whether that cipher is "strong" or "weak". This is used in https://github.com/github/codeql/pull/7313 to decide if this is a safe way to encrypt cookies or not. Some of the implementation here is taken from the [Java `Encryption.qll` library](https://github.com/github/codeql/blob/95d175e9e05d2b6a668d0c56afae0c3e7c8a1437/java/ql/lib/semmle/code/java/security/Encryption.qll#L104), in particular the regex matching code.

The `OpenSSL` Ruby library itself wraps a C `OpenSSL`/`LibreSSL` library. The known ciphers here were generated using `OpenSSL 1.1.1l` and `LibreSSL 3.4.1`, which each support some algorithms that the other does not.

<details>
<summary>Original description</summary>
<p dir="auto">Adds a <code>CryptoAlgorithms.qll</code> library for Ruby. This is similar to the libraries of the same name for JS and Python, but currently not identical to them. It would be good to unify as much of these as possible, but there may be some parts of this Ruby implementation that are language specific and need to move out into a separate file.</p>
<p dir="auto">The main change here, relative to the JS/Python version, is the addition of an <code>OpenSSL</code> module. This is specifically to support an upcoming PR which will use this to determine strong/weak ciphers supported by OpenSSL. These ciphers can include both the underlying block encryption algorithm and also the block mode (e.g. <code>aes-256-ecb</code>) - which in particular means that we say that a cipher is weak if either of these components are considered weak. These ciphers are encoded in the <code>OpenSSLCipher</code> class, which is added here as a type of <code>CryptographicAlgorithm</code>.</p>
<p dir="auto">I've uploaded the diff of this change relative to the existing versions at <a href="https://gist.github.com/alexrford/92426d26cb6b0def5d8967d49a0be0a4">https://gist.github.com/alexrford/92426d26cb6b0def5d8967d49a0be0a4</a>. In addition to the OpenSSL related changes, I've pulled in some of the regex matching code from the <a href="https://github.com/github/codeql/blob/main/cpp/ql/lib/semmle/code/cpp/security/Encryption.qll#L38">C++ Encryption.qll library</a>, and added a few more encryption algorithms (Cast, GOST, ChaCha).</p>
<p dir="auto"><a class="user-mention" data-hovercard-type="user" data-hovercard-url="/users/RasmusWL/hovercard" data-octo-click="hovercard-link-click" data-octo-dimensions="link_type:self" href="https://github.com/RasmusWL">@RasmusWL</a> I believe you worked on unifying the JS and Python versions of this library? I'd be interested to get your thoughts on what to do in this case, in particular if:</p>
<ol dir="auto">
<li>the OpenSSL related code here could make sense in a shared version of this library, or</li>
<li>if it should be factored out into a separate file and Ruby should use the existing <code>CryptoAlgorithms.qll</code> as it is (perhaps with Cast/GOST/ChaCha added)?</li>
</ol>
</details>


alexrford:
I've revisited this since the first version and changed it s.t. `CryptoAlgorithms.qll` is shared with the Python/JS versions and the OpenSSL related code is moved out to a separate file. This necessitated making the `AlgorithmNames` module public, which I'm not entirely satisfied with as it seems mostly like an implementation detail. I'm not sure of a good alternative, though.

ruby/ql/lib/codeql/ruby/frameworks/ActionController.qll
```diff
@@ -118,6 +118,25 @@ class ParamsSource extends RemoteFlowSource::Range {
...
+/**
+ * A `RemoteFlowSource::Range` to represent accessing the
+ * ActionController cookies available via the `cookies` method.
+ */
+class CookiesSource extends RemoteFlowSource::Range {
+  CookiesCall call;
+
+  CookiesSource() { this.asExpr().getExpr() = call }
+
+  override string getSourceType() { result = "ActionController::Metal#params" }
```
nickrolfe, line 137: s/params/cookies

But you can get rid of this change completely, if we merge https://github.com/github/codeql/pull/7341

alexrford: Thanks, this was a WIP change that I included by accident. I'll take a look at #7341 today.

alexrford:
Updated again, with a few significant changes:
 - Ciphers in the `OpenSSL` module are based on those available using gems built for both OpenSSL 1.1.1 and for LibreSSL 3.4.1.
 - This in turn adds a few more encryption algorithms: `ARIA`, `IDEA`, `SEED`, and `SM4`, and block modes: `CBC`, `CFB`, `CTR`, and `OFB`. These are all classified as strong as I haven't been able to find any attacks that seem to weaken these more than any other algorithms that we've already marked as strong, but I'm of course open to be corrected on this.
 - QLDoc fixes.

ruby/ql/lib/codeql/ruby/security/OpenSSL.qll
```diff
@@ -0,0 +1,310 @@
...
+  /**
+   * Gets the canonical version of `name`, as reported by `OpenSSL::Cipher#name`.
+   * No result if `name` is not a known OpenSSL cipher name.
+   */
+  string getCanonicalCipherName(string name) {
+    isOpenSSLCipher(name) and
+    (
+      if exists(string special | special = getSpecialCanonicalCipherName(name))
+      then result = getSpecialCanonicalCipherName(name)
+      else result = name.toUpperCase()
```
nickrolfe, line 250: Isn't this equivalent to the following?

```suggestion
      result = getSpecialCanonicalCipherName(name)
      or
      not exists(getSpecialCanonicalCipherName(name)) and
      result = name.toUpperCase()
```

nickrolfe (commented):
Otherwise, LGTM. It'd be good to get sign-off from JS/Python as well.

ruby/ql/lib/codeql/ruby/security/OpenSSL.qll
alexrford: That seems a bit neater 👍

erik-krogh:
You add the `isStrongBlockMode()` predicate, but you never use it?

alexrford:
> You add the `isStrongBlockMode()` predicate, but you never use it?

True, I added this for symmetry with `isWeakBlockMode()` but on reflection it's not necessary. We use the block mode as a reason to disqualify a cipher from being marked as strong. However, the block mode isn't sufficient to mark a cipher as strong and isn't always present in the cipher name (particularly for stream ciphers like `ChaCha` where a block mode wouldn't make sense).

I've removed this predicate.

ruby/ql/lib/codeql/ruby/security/CryptoAlgorithms.qll
```diff
@@ -0,0 +1,98 @@
+/**
+ * Provides classes modeling cryptographic algorithms, separated into strong and weak variants.
+ *
+ * The classification into strong and weak are based on Wikipedia, OWASP and google (2017).
```
aibaars, line 4: The year `2017` seems a bit long ago.

alexrford: Bumped to 2021

aibaars (approved):

aibaars (approved):

PR #7283, opened 2021-12-01
Ruby: pattern matching: CFG

aibaars:
This pull request adds the CFG implementation for pattern matching.

To help review here is a [rendered CFG](https://gist.githubusercontent.com/aibaars/3d658fc4c6bdd34dc20a62e4b7065d49/raw/6f2c7ecdca9a1e24a75a2fef98b2b0f0dc1dbb24/case.svg) for the CFG's of [case.rb](https://github.com/github/codeql/pull/7283/files#diff-e0ce958a2b71e898f8051e21222ee5f4e5ffa59b8374552cabb97e45be68cafd)


ruby/ql/lib/codeql/ruby/controlflow/internal/Completion.qll
```diff
@@ -227,6 +227,8 @@ private predicate inMatchingContext(AstNode n) {
   or
   n instanceof CasePattern
   or
+  n = any(CasePattern c).getDesugared()
```
aibaars, line 230: @hvitved Is this something that should be addressed more generally in the CFG library? I guess we'd want to preserve the completion context when desugaring a node. Should `isValidFor(node) implies isValidFor(node.getDesugared())` hold in general?

hvitved: Yes, that sounds right.

aibaars: Does this look right to you? https://github.com/github/codeql/pull/7283/commits/06b3d583e512f11f37f9320517cef9d7f3dbc13f

ruby/ql/lib/codeql/ruby/controlflow/internal/Completion.qll
```diff
@@ -310,7 +310,7 @@ class MatchingCompletion extends ConditionalCompletion {
     this = TNestedCompletion(_, TMatchingCompletion(value), _)
   }
 
-  override SuccessorType getAMatchingSuccessorType() {
+  override ConditionalSuccessor getAMatchingSuccessorType() {
```
hvitved, line 313: Could you change it to `MatchingSuccessor` instead and remove the infix cast below?

aibaars: That does not work for `class NestedMatchingCompletion extends NestedCompletion, MatchingCompletion`, does it?

hvitved: `Completion`s are different from `SuccessorType`s (the latter are never nested).

aibaars: What I meant is that `override BooleanSuccessor getAMatchingSuccessorType()` defined in `NestedMatchingCompletion` would not match the signature if we restricted it to `MatchingSuccessor` in the super class (`MatchingCompletion`)

hvitved: Ah, right, sorry.

hvitved (approved):

PR #7286, opened 2021-12-02
Java: CWE-552 Query to detect unsafe request dispatcher usage

luchua-bc:
Directly incorporating user input into HTTP requests dispatched from the Java EE RequestDispatcher without proper validation of the input can allow any web application resource such as configuration files and source code to be disclosed.

As stated in the [Java API doc](https://jakarta.ee/specifications/webprofile/9/apidocs/jakarta/servlet/servletrequest#getRequestDispatcher-java.lang.String-), when using a Java EE RequestDispatcher, requests may be dispatched to any part of the web application bypassing both implicit (no direct access to WEB-INF or META-INF) and explicit (defined by the web application) security constraints. Unsanitized user provided data must not be used to construct the path passed to the RequestDispatcher as it is very likely to create a security vulnerability in the application.

This query detects unsafe invocations of RequestDispatcher with user controlled input. Important features include:
+ RequestDispatcher constructed from both HTTP request and servlet context
+ Path traversal check
+ Path encoding check
+ Check of path normalization using the `java.nio.file.Path` package

The query was previously submitted then closed to wait for the closing of another relevant PR. The current query provided further enhancement in addition to new scenarios and was validated against published CVEs. 

Please consider to merge the PR. Thanks.


java/ql/src/experimental/Security/CWE/CWE-552/UnsafeUrlForward.ql
```diff
@@ -14,20 +14,154 @@ import java
...
+predicate isVirtualMethod(MethodAccess ma, Expr expr) {
+  ma.getMethod().getDeclaringType() instanceof TypeObject and
+  ma.getMethod().hasName("toString") and
+  (expr = ma or expr = ma.getQualifier())
+  or
+  (
+    ma.getMethod().getDeclaringType().getASupertype*().hasQualifiedName("java.util", "Map") and
+    ma.getMethod().hasName(["get", "getOrDefault"])
+  ) and
+  (expr = ma or expr = ma.getAnArgument())
```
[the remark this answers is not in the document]
atorralba: Also, I understand this ad-hoc sanitizer was added because there are many FPs in certain projects that have classes implementing `Object.toString` and `Map.get[OrDefault]`. What I'm worried about is that we could lose potentially interesting flows that go through a `Map` because of this.

After reviewing the FPs, for instance in `apache/struts`, I'm seeing that the spurious flow is actually caused by the lack of call context in sources, e.g.

```java
public String toString() {
  return getTitle() + ":" + getSomethingElse(); // getTitle() is a source, and taints any call to `toString` from anywhere.
}
```

So I think this could be fixed using flow features instead. Try adding this predicate to your `Configuration` and remove `VirtualMethodSanitizer`:

```codeql
override DataFlow::FlowFeature getAFeature() {
  result instanceof DataFlow::FeatureHasSourceCallContext
}
```

That should get rid of FPs caused by sources lacking call context.

java/ql/src/experimental/Security/CWE/CWE-552/UnsafeUrlForward.qll
```diff
@@ -70,3 +70,28 @@ private class SpringUrlForwardSink extends UnsafeUrlForwardSink {
...
+
+/** Taint model related to `java.nio.file.Path`. */
+private class FilePathFlowStep extends SummaryModelCsv {
+  override predicate row(string row) {
+    row =
+      [
+        "java.nio.file;Paths;true;get;;;Argument[-1];ReturnValue;taint",
+        "java.nio.file;Path;true;resolve;;;Argument[0];ReturnValue;taint",
+        "java.nio.file;Path;true;normalize;;;Argument[-1];ReturnValue;taint",
+        "java.nio.file;Path;true;startsWith;;;Argument[-1];ReturnValue;taint",
```
atorralba, line 93: I don't think `startsWith` should propagate taint to the return value, since it returns a `boolean`.

java/ql/src/experimental/Security/CWE/CWE-552/UnsafeUrlForward.qll
```diff
@@ -70,3 +70,28 @@ private class SpringUrlForwardSink extends UnsafeUrlForwardSink {
...
+      ]
+  }
+}
+
+/** Taint model related to `java.nio.file.Path`. */
+private class FilePathFlowStep extends SummaryModelCsv {
+  override predicate row(string row) {
+    row =
+      [
+        "java.nio.file;Paths;true;get;;;Argument[-1];ReturnValue;taint",
```
atorralba, line 90: `Paths.get` is static, so a tainted qualifier doesn't make much sense. Wouldn't you want to add a summary from the arguments to the return value instead?

java/ql/src/experimental/Security/CWE/CWE-552/UnsafeUrlForward.qll
```diff
@@ -70,3 +70,28 @@ private class SpringUrlForwardSink extends UnsafeUrlForwardSink {
...
+        "java.nio.file;Path;true;resolve;;;Argument[0];ReturnValue;taint",
```
atorralba, line 91: I think we should add a summary for `Argument[-1]` -> `ReturnValue` too. The qualifier gets directly returned if the argument is empty.

java/ql/src/experimental/Security/CWE/CWE-552/UnsafeUrlForward.qhelp
```diff
@@ -5,27 +5,45 @@
...
+untrusted URL forwarding, it is recommended to avoid concatenating user input directly into the forwarding URL.
+</p>
 
 </recommendation>
 <example>
 
 <p>The following examples show the bad case and the good case respectively.
 The <code>bad</code> methods show an HTTP request parameter being used directly in a URL forward
-without validating the input, which may cause file leakage. In <code>good1</code> method, 
+without validating the input, which may cause file leakage. In <code>good1</code> method,
```
atorralba, line 23:
```suggestion
without validating the input, which may cause file leakage. In the <code>good1</code> method,
```

java/ql/src/experimental/Security/CWE/CWE-552/UnsafeUrlForward.ql
```diff
@@ -14,20 +14,154 @@ import java
...
 import DataFlow::PathGraph
 
-private class StartsWithSanitizer extends DataFlow::BarrierGuard {
-  StartsWithSanitizer() {
-    this.(MethodAccess).getMethod().hasName("startsWith") and
-    this.(MethodAccess).getMethod().getDeclaringType() instanceof TypeString and
-    this.(MethodAccess).getMethod().getNumberOfParameters() = 1
-  }
+/**
+ * Holds if `ma` is a method call of matching with a path string, probably a whitelisted one.
```
atorralba, line 21:
```suggestion
 * Holds if `ma` is a call to a method that checks a path string, probably a whitelisted one.
```

java/ql/src/experimental/Security/CWE/CWE-552/UnsafeUrlForward.ql
```diff
@@ -14,20 +14,154 @@ import java
...
+ */
+predicate isStringPathMatch(MethodAccess ma) {
+  ma.getMethod().getDeclaringType() instanceof TypeString and
+  ma.getMethod().getName() = ["startsWith", "matches", "regionMatches"]
+}
 
-  override predicate checks(Expr e, boolean branch) {
-    e = this.(MethodAccess).getQualifier() and branch = true
+/**
+ * Holds if `ma` is a method call of `java.nio.file.Path` which matches with another
```
atorralba, line 29:
```suggestion
 * Holds if `ma` is a call to a method of `java.nio.file.Path` that checks a
```

java/ql/src/experimental/Security/CWE/CWE-552/UnsafeUrlForward.ql
```diff
@@ -14,20 +14,154 @@ import java
...
+ * Holds if `ma` is a method call to check special characters `..` used in path traversal.
+ */
+predicate isPathTraversalCheck(MethodAccess ma) {
+  ma.getMethod().getDeclaringType() instanceof TypeString and
+  ma.getMethod().hasName(["contains", "indexOf"]) and
+  ma.getAnArgument().(CompileTimeConstantExpr).getStringValue() = ".."
+}
+
+/**
+ * Holds if `ma` is a method call to decode a url string or check url encoding.
```
atorralba, line 58:
```suggestion
 * Holds if `ma` is a method call to decode a URL string or check URL encoding.
```

atorralba (commented):
Thanks for your submission @luchua-bc. I added some inline comments.

java/ql/src/experimental/Security/CWE/CWE-552/UnsafeUrlForward.ql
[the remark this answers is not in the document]
luchua-bc: Thanks @atorralba for reviewing this PR. The new approach works like a charm and helps to get rid of FPs caused by sources lacking call context.

java/ql/src/experimental/Security/CWE/CWE-552/UnsafeUrlForward.qhelp
luchua-bc: Changed

java/ql/src/experimental/Security/CWE/CWE-552/UnsafeUrlForward.ql
luchua-bc: Changed

java/ql/src/experimental/Security/CWE/CWE-552/UnsafeUrlForward.ql
luchua-bc: Changed

java/ql/src/experimental/Security/CWE/CWE-552/UnsafeUrlForward.ql
luchua-bc: Changed

java/ql/src/experimental/Security/CWE/CWE-552/UnsafeUrlForward.ql
```diff
@@ -14,20 +14,154 @@ import java
...
+}
+
+private class PathMatchSanitizer extends DataFlow::Node {
+  PathMatchSanitizer() {
+    exists(MethodAccess ma |
+      (
+        isStringPathMatch(ma) and
+        exists(MethodAccess ma2 |
+          isPathTraversalCheck(ma2) and
+          ma.getQualifier().(VarAccess).getVariable().getAnAccess() = ma2.getQualifier()
```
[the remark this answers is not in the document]
luchua-bc: Changed

java/ql/src/experimental/Security/CWE/CWE-552/UnsafeUrlForward.ql
```diff
@@ -14,20 +14,154 @@ import java
...
+        isFilePathMatch(ma)
+      ) and
+      (
+        not checkNoPathMatch(ma)
+        or
+        // non-match check needs decoding e.g. !path.startsWith("/WEB-INF/") won't detect /%57EB-INF/web.xml, which will be decoded and served by RequestDispatcher
+        checkNoPathMatch(ma) and
+        exists(MethodAccess ma2 |
+          isPathDecoding(ma2) and
+          ma.getQualifier().(VarAccess).getVariable().getAnAccess() = ma2.getQualifier()
```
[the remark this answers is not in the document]
luchua-bc: Changed

java/ql/src/experimental/Security/CWE/CWE-552/UnsafeUrlForward.ql
```diff
@@ -14,20 +14,154 @@ import java
...
+  ma.getMethod().getDeclaringType() instanceof TypeString and
+  ma.getMethod().hasName(["contains", "indexOf"]) and
+  ma.getAnArgument().(CompileTimeConstantExpr).getStringValue() = "%"
+  or
+  // Call to `URLDecoder` assuming the implementation handles double encoding correctly
+  ma.getMethod().getDeclaringType().hasQualifiedName("java.net", "URLDecoder") and
+  ma.getMethod().hasName("decode")
+}
+
+private class PathMatchSanitizer extends DataFlow::Node {
```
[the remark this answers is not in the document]
luchua-bc: I've made the change, which simplified it to certain degree. Please help to validate whether it's exactly what you want. Thanks.

java/ql/src/experimental/Security/CWE/CWE-552/UnsafeUrlForward.ql
```diff
@@ -14,20 +14,154 @@ import java
...
+      this.asExpr() = ma.getQualifier()
+    )
   }
 }
 
+/**
+ * Holds if `ma` is a method call to check string content, which means an input string is not
+ * blindly trusted and helps to reduce FPs.
+ */
+predicate checkStringContent(MethodAccess ma, Expr expr) {
```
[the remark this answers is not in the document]
luchua-bc: Changed

java/ql/src/experimental/Security/CWE/CWE-552/UnsafeUrlForward.ql
```diff
@@ -14,20 +14,154 @@ import java
...
+        ma.getMethod().getDeclaringType() instanceof TypeString and
+        ma.getMethod().hasName("equals") and
+        ma.getArgument(0).(CompileTimeConstantExpr).getStringValue() = ""
+      )
+    ) and
+    expr.getParent+() = rt
+  )
+}
+
+private class NullOrEmptyCheckSanitizer extends DataFlow::Node {
```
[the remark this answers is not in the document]
luchua-bc: This sanitizer is for a special edge case found in a few projects, which is:
```
    public static String decodePath(String path)
    {
        if (path == null || path.length() == 0)
            return path;

        path = URLDecoder.decode(path, "UTF-8");
        return path;
    }
```

It's to detect the null check which returns null or a blank string directly although a valid non-empty input will be decoded. I tried <code>BarrierGuard</code> but it doesn't seem to help. I'm not sure whether my implementation is not right or whether <code>BarrierGuard</code> helps with this scenario. I can also remove this sanitizer since it's only for an edge case. Please advise. Thanks.

java/ql/src/experimental/Security/CWE/CWE-552/UnsafeUrlForward.qll
luchua-bc: Removed

java/ql/src/experimental/Security/CWE/CWE-552/UnsafeUrlForward.qll
luchua-bc: Changed

java/ql/src/experimental/Security/CWE/CWE-552/UnsafeUrlForward.qll
luchua-bc: Added

atorralba:
Hey @luchua-bc. I reviewed your changes and took the liberty of adjusting some things.

* In cd9a485 I changed `NullOrEmptyCheckSanitizer` to be a `BarrierGuard`. Since you mentioned you couldn't make it work, I thought providing this example would help you. In any case, if this is only for a rare few edge cases, and considering that nullness checks are always a little problematic, I think the query would be simpler if we just removed this.
* In 81feaae I removed `NullOrEmptyCheckGuard`, for the reasons mentioned above, and also made a significant refactor to `PatchMatchSanitizer`. Since this is a big change, I want to explain why I did it. Feel free to disregard (revert) this commit if you don't agree with it:
  * You were checking for uses of the `!` operator, assuming that the use of the negation meant that the developer wanted to exclude some words (blocklist), and if it wasn't used then it was checking for safe paths (allowlist). This isn't accurate — a developer could just use the negation with a safe path to implement a guard clause, or use an unsafe path (like WEB-INF) without the negation for the same reason.
  * Instead of that, what I thought could work better was just checking the strings being compared: if they're known "bad" strings like "WEB-INF" or "..", we consider that the check is a blocklist (and thus we set `branch = false` in the BarrierGuard). Otherwise, we consider it an allowlist and set `branch = true`. That way you don't need to worry about whether the check is being negated or not, or any other potential complexity around boolean checks. Of course this has caveats, but as an heuristic I think it improves the previous approach a little.
  * Also, I got rid of `StringOperationSanitizer`, since in my opinion it was too broad and wasn't modeling any security protection, just an access to the string contents. Even if the string is read, its contents need to be checked in some way to be considered safe, and I think the current guards already do that job.
  * Note that I changed the test case `doHead5` from GOOD to BAD. If I'm not mistaken, that case would still be exploitable via double URL encoding, since that would bypass the "WEB-INF" check and would get decoded by `getRequestDispatcher`. That also fits with the simplified approach of requiring decoding protections for blocklist checks and path traversal protections for allowlist checks.
* Finally, in b6886b8 I just moved everything except the configuration to the QLL file, for convenience.

luchua-bc:
Thanks @atorralba a lot for the changes and I learned a new way to make the query more concise:-)

java/ql/lib/semmle/code/java/frameworks/Servlets.qll
```diff
@@ -347,3 +347,27 @@ predicate isRequestGetParamMethod(MethodAccess ma) {
   ma.getMethod() instanceof ServletRequestGetParameterMapMethod or
   ma.getMethod() instanceof HttpServletRequestGetQueryStringMethod
 }
+
+/** The Java EE RequestDispatcher. */
+library class RequestDispatcher extends RefType {
```
atorralba, line 352: Oh, don't use `library`, it's deprecated.

```suggestion
class RequestDispatcher extends RefType {
```

java/ql/lib/semmle/code/java/frameworks/Servlets.qll
```diff
@@ -347,3 +347,27 @@ predicate isRequestGetParamMethod(MethodAccess ma) {
...
+  RequestDispatcher() {
+    this.hasQualifiedName(["javax.servlet", "jakarta.servlet"], "RequestDispatcher") or
+    this.hasQualifiedName("javax.portlet", "PortletRequestDispatcher")
+  }
+}
+
+/** The `getRequestDispatcher` method. */
+library class GetRequestDispatcherMethod extends Method {
```
atorralba, line 360:
```suggestion
class GetRequestDispatcherMethod extends Method {
```

java/ql/lib/semmle/code/java/frameworks/Servlets.qll
```diff
@@ -347,3 +347,27 @@ predicate isRequestGetParamMethod(MethodAccess ma) {
...
+  GetRequestDispatcherMethod() {
+    this.getReturnType() instanceof RequestDispatcher and
+    this.getName() = "getRequestDispatcher"
+  }
+}
+
+/** The request dispatch method. */
+library class RequestDispatchMethod extends Method {
```
atorralba, line 368:
```suggestion
class RequestDispatchMethod extends Method {
```

java/ql/src/experimental/Security/CWE/CWE-552/UnsafeServletRequestDispatch.java
```diff
@@ -0,0 +1,82 @@
+public class UnsafeServletRequestDispatch extends HttpServlet {
+	private static final String BASE_PATH = "/pages";
+
+	@Override
+	protected void doGet(HttpServletRequest request, HttpServletResponse response)
+			throws ServletException, IOException {
+		{
+			// GOOD: whitelisted URI
```
smowton, line 8:
```suggestion
			// GOOD: check for an explicitly permitted URI
```

java/ql/src/experimental/Security/CWE/CWE-552/UnsafeUrlForward.qll
```diff
@@ -30,13 +31,132 @@ private class FollowsSanitizingPrefix extends UnsafeUrlForwardSanitizer {
...
+ */
+private predicate isExactStringPathMatch(MethodAccess ma) {
+  ma.getMethod().getDeclaringType() instanceof TypeString and
+  ma.getMethod().getName() = ["equals", "equalsIgnoreCase"]
+}
+
+/**
+ * Holds if `ma` is a call to a method that checks a path string.
+ */
+private predicate isStringPathMatch(MethodAccess ma) {
```
smowton, line 48:
```suggestion
 * Holds if `ma` is a call to a method that checks a partial string match.
 */
private predicate isStringPartialMatch(MethodAccess ma) {
```

java/ql/src/experimental/Security/CWE/CWE-552/UnsafeUrlForward.qll
```diff
@@ -30,13 +31,132 @@ private class FollowsSanitizingPrefix extends UnsafeUrlForwardSanitizer {
...
+  ma.getMethod().getDeclaringType() instanceof TypeString and
+  ma.getMethod().getName() =
+    ["contains", "startsWith", "matches", "regionMatches", "indexOf", "lastIndexOf"]
+}
+
+/**
+ * Holds if `ma` is a call to a method of `java.nio.file.Path` that checks a path.
+ */
+private predicate isFilePathMatch(MethodAccess ma) {
```
smowton, line 57:
```suggestion
 * Holds if `ma` is a call to a method of `java.nio.file.Path` that checks a partial path match.
 */
private predicate isPathPartialMatch(MethodAccess ma) {
```

In general, name predicates for what they do, not how you intend to use them.

java/ql/src/experimental/Security/CWE/CWE-552/UnsafeUrlForward.qll
```diff
@@ -30,13 +31,132 @@ private class FollowsSanitizingPrefix extends UnsafeUrlForwardSanitizer {
...
+ * * performing URL decoding
+ */
+private predicate isURLEncodingCheck(MethodAccess ma, Expr checked) {
+  // Search the special character `%` used in url encoding
+  ma.getMethod().getDeclaringType() instanceof TypeString and
+  ma.getMethod().hasName(["contains", "indexOf"]) and
+  ma.getAnArgument().(CompileTimeConstantExpr).getStringValue() = "%" and
+  ma.(Guard).controls(checked.getBasicBlock(), false)
+  or
+  // Call to `URLDecoder` assuming the implementation handles double encoding correctly
```
smowton, line 89: It does not.

```
jshell> URLDecoder.decode("%25C3%2596", "UTF-8");
$1 ==> "%C3%96"
```

(from https://stackoverflow.com/questions/20679604/does-class-urldecoder-decode-handle-double-encoding)

You don't necessarily have to handle this in this query -- if it's difficult to determine whether they're accounting for double encoding then it might be an acceptable false negative -- but you shouldn't claim you're doing so in the QL comments.

java/ql/src/experimental/Security/CWE/CWE-552/UnsafeUrlForward.qll
```diff
@@ -30,13 +31,132 @@ private class FollowsSanitizingPrefix extends UnsafeUrlForwardSanitizer {
...
+
+/**
+ * A guard that checks a path with the following methods:
+ *  1. Exact string match
+ *  2. Path matches allowed values (needs to protect against path traversal)
+ *  3. Path matches disallowed values (needs to protect against URL encoding)
+ */
+private class PathMatchGuard extends UnsafeUrlForwardBarrierGuard {
+  PathMatchGuard() {
+    isExactStringPathMatch(this) or isStringPathMatch(this) or isFilePathMatch(this)
```
smowton, line 125:
```suggestion
    isExactStringPathMatch(this) or isAllowListCheck(this) or isDisallowListCheck(this)
```

`checks` can't match without at least one of these holding.

java/ql/src/experimental/Security/CWE/CWE-552/UnsafeUrlForward.qll
```diff
@@ -30,13 +31,132 @@ private class FollowsSanitizingPrefix extends UnsafeUrlForwardSanitizer {
...
+  isDisallowedWord(ma.getAnArgument())
+}
+
+/**
+ * A guard that checks a path with the following methods:
+ *  1. Exact string match
+ *  2. Path matches allowed values (needs to protect against path traversal)
+ *  3. Path matches disallowed values (needs to protect against URL encoding)
+ */
+private class PathMatchGuard extends UnsafeUrlForwardBarrierGuard {
```
smowton, line 123:
```suggestion
private class PathMatchGuard extends UnsafeUrlForwardBarrierGuard, MethodAccess {
```
And remove all the `.(MethodAccess)` casts

java/ql/src/experimental/Security/CWE/CWE-552/UnsafeUrlForward.qll
```diff
@@ -30,13 +31,132 @@ private class FollowsSanitizingPrefix extends UnsafeUrlForwardSanitizer {
...
+    e = this.(MethodAccess).getQualifier() and
+    (
+      isExactStringPathMatch(this) and
+      branch = true
+      or
+      isAllowListCheck(this) and
+      exists(MethodAccess ma, Expr checked | isPathTraversalCheck(ma, checked) |
+        DataFlow::localExprFlow(checked, e)
+        or
+        ma.getParent*().(BinaryExpr) = this.(MethodAccess).getParent*()
```
smowton, line 138: This is hard to visualise -- give a comment indicating what nesting of the allow-list check and the path-traversal check you intend?

java/ql/src/experimental/Security/CWE/CWE-552/UnsafeUrlForward.qll
```diff
@@ -70,3 +190,27 @@ private class SpringUrlForwardSink extends UnsafeUrlForwardSink {
...
+  override predicate row(string row) {
+    row =
+      [
+        "javax.servlet.http;HttpServletRequest;true;getServletPath;;;ReturnValue;remote",
+        "jakarta.servlet.http;HttpServletRequest;true;getServletPath;;;ReturnValue;remote"
+      ]
+  }
+}
+
+/** Taint model related to `java.nio.file.Path`. */
```
smowton, line 205: These are worth having in general; would appreciate a followup PR upstreaming these

java/ql/src/experimental/Security/CWE/CWE-552/UnsafeServletRequestDispatch.java
```diff
@@ -0,0 +1,82 @@
+public class UnsafeServletRequestDispatch extends HttpServlet {
+	private static final String BASE_PATH = "/pages";
+
+	@Override
+	protected void doGet(HttpServletRequest request, HttpServletResponse response)
```
smowton, line 5: This example is too long for a qhelp file, mainly by repeating lots of boilerplate. For this purpose it would be enough to use a style like


```
// To check there won't be unexpected path traversal, we can use Path.normalize:
(code here)
// ... or alternatively ensure URL encoding is removed and then check for any `..` sequences ...
(code here)
// and so on
```

Basically bear in mind this is supposed to be easy to read, not a full test case.

java/ql/src/experimental/Security/CWE/CWE-552/UnsafeUrlForward.qhelp
```diff
@@ -5,27 +5,45 @@
...
-<p>Constructing a server-side redirect path with user input could allow an attacker to download application binaries 
+<p>Constructing a server-side redirect path with user input could allow an attacker to download application binaries
 (including application classes or jar files) or view arbitrary files within protected directories.</p>
 
 </overview>
 <recommendation>
 
-<p>In order to prevent untrusted URL forwarding, it is recommended to avoid concatenating user input directly into the forwarding URL.</p>
+<p>Unsanitized user provided data must not be used to construct the path for URL forwarding. In order to prevent
+untrusted URL forwarding, it is recommended to avoid concatenating user input directly into the forwarding URL.
```
smowton, line 15: Mention what should be done. For example,

"Instead, user input should be checked for path traversal using `..` sequences, taking care to remove any URL encoding first." ?

java/ql/src/experimental/Security/CWE/CWE-552/UnsafeUrlForward.qll
```diff
@@ -30,13 +31,132 @@ private class FollowsSanitizingPrefix extends UnsafeUrlForwardSanitizer {
...
+    isExactStringPathMatch(this) or isStringPathMatch(this) or isFilePathMatch(this)
+  }
+
+  override predicate checks(Expr e, boolean branch) {
+    e = this.(MethodAccess).getQualifier() and
+    (
+      isExactStringPathMatch(this) and
+      branch = true
+      or
+      isAllowListCheck(this) and
```
smowton, line 134: Add a comment to the `.ql` file indicating what the user ought to do, and how this query is approximating it.

For example, I expect what you ideally would check for is the user (a) removing all URL encoding, and then (b) checking for path traversal, but in fact the query sanitizes more than that? Looks like it would mark as probably safe anything that passes a check for `..` and then flows into a check for an allowed path, or anything that get url-decoded and then passes a check for `..`, `WEB-INF` or `META-INF`?

This is quite a complicated set of rules, so it would be good to have your intended meaning written down somewhere.

java/ql/lib/semmle/code/java/frameworks/Servlets.qll
luchua-bc: Removed

java/ql/lib/semmle/code/java/frameworks/Servlets.qll
luchua-bc: Removed

java/ql/src/experimental/Security/CWE/CWE-552/UnsafeUrlForward.qll
luchua-bc: Thanks @smowton for reviewing this PR. Shall I just remove the comment since we cannot assume double encoding is handled correctly in the QL comments? Or shall we remove this check completely? I can also add a check to find out whether URLDecoder is used in a loop just like the example https://github.com/github/codeql/blob/877c52981ff3b333d13cdb802b7f3592e5d753bc/java/ql/test/experimental/query-tests/security/CWE-552/UnsafeServletRequestDispatch.java#L119-L122. Will this suffice? Please advise. Thanks.

java/ql/src/experimental/Security/CWE/CWE-552/UnsafeUrlForward.qll
atorralba: I think that throws "ambiguous inherited methods" errors because some common predicates between `BarrierGuard` and `MethodAccess` have different implementations. Changing it to `extends UnsafeUrlForwardBarrierGuard instanceof MethodAccess` also doesn't seem to help.

java/ql/src/experimental/Security/CWE/CWE-552/UnsafeUrlForward.qll
smowton: Let's keep it simple for now; just note that double-encoding is a risk this query doesn't handle and maybe we'll look into that as future work.

java/ql/src/experimental/Security/CWE/CWE-552/UnsafeUrlForward.qll
atorralba: See 5a1dbd4.

java/ql/src/experimental/Security/CWE/CWE-552/UnsafeUrlForward.qll
atorralba: See 5a1dbd4.

atorralba:
Thanks for your review @smowton. 

@luchua-bc, I applied @smowton's comments regarding the rewrite I pushed. All the other suggestions are yours to handle :)

java/ql/src/experimental/Security/CWE/CWE-552/UnsafeUrlForward.qll
luchua-bc: Updated

java/ql/src/experimental/Security/CWE/CWE-552/UnsafeUrlForward.qll
luchua-bc: Updated

java/ql/src/experimental/Security/CWE/CWE-552/UnsafeUrlForward.qll
luchua-bc: Thanks @smowton for the advice. We'll keep it simple for now.

java/ql/src/experimental/Security/CWE/CWE-552/UnsafeUrlForward.qll
luchua-bc: Sounds like a good idea.

java/ql/src/experimental/Security/CWE/CWE-552/UnsafeServletRequestDispatch.java
luchua-bc: Changed

java/ql/src/experimental/Security/CWE/CWE-552/UnsafeServletRequestDispatch.java
luchua-bc: I've updated the test cases to be much more concise and easier to read. Please help to check whether further improvement is required. Thanks.

java/ql/src/experimental/Security/CWE/CWE-552/UnsafeUrlForward.qhelp
luchua-bc: I've added what should be done. Please validate. Thanks.

luchua-bc:
Thanks @atorralba for those new commits. I've made changes for all the other suggestions. 

@smowton - please let me know if more changes are needed. Thanks.

java/ql/src/experimental/Security/CWE/CWE-552/UnsafeUrlForward.qhelp
```diff
@@ -5,27 +5,48 @@
...
+Instead, user input should be checked for path traversal using <code>..</code> sequences or the user input should
+be normalized using <code>Path.normalize</code>, any URL encoding should be removed, and user input should be
+checked against a permitted URI.
```
smowton, line 18:
```suggestion
Instead, user input should be checked against allowed (e.g., must come within <code>user_content/</code>) or disallowed (e.g. must not come within <code>/internal</code>) paths, ensuring that neither path traversal using <code>../</code> or URL encoding are used to evade these checks.
```

java/ql/src/experimental/Security/CWE/CWE-552/UnsafeServletRequestDispatch.java
```diff
@@ -0,0 +1,55 @@
...
+				path = URLDecoder.decode(path, "UTF-8");
+				hasEncoding = path.contains("%");
+			}
+
+			if (!path.startsWith("/WEB-INF/") && !path.contains("..")) {
+				request.getServletContext().getRequestDispatcher(path).include(request, response);
+			}
+		}
+	}
+}
```
smowton, line 55:
```suggestion
// BAD: no URI validation
String returnURL = request.getParameter("returnURL");
RequestDispatcher rd = sc.getRequestDispatcher(returnURL);
rd.forward(request, response);

// GOOD: check for a trusted prefix, ensuring path traversal is not used to erase that prefix:
// (alternatively use `Path.normalize` instead of checking for `..`)
if(!returnURL.contains("..") && returnURL.hasPrefix("/pages")) { ... }

// Also GOOD: check for a forbidden prefix, ensuring URL-encoding is not used to evade the check:
// (alternatively use `URLDecoder.decode` before `hasPrefix`)
if(returnURL.hasPrefix("/internal") && !returnURL.contains("%")) { ... }
```

smowton: Suggest also dropping UnsafeRequestPath.java: it's exactly the same concept but with `chain.doFilter` instead of `forward`.

smowton (dismissed):

smowton (approved):

luchua-bc:
Thanks a lot for all the help from @atorralba and @smowton.

PR #7231, opened 2021-11-24
C#: Enable data-flow consistency queries

hvitved:
This PR enables data-flow consistency queries for C#, and also fixes a whole bunch of inconsistencies (others were fixed in preliminary PRs).

PR #7288, opened 2021-12-02
Shared CFG: Include CFG scope in `TElementNode`

hvitved:
Enables the fix of an inconsistency for C# identified via https://github.com/github/codeql/pull/7231.

### DIFF
--- a/ruby/ql/lib/codeql/ruby/controlflow/internal/ControlFlowGraphImplShared.qll
+++ b/ruby/ql/lib/codeql/ruby/controlflow/internal/ControlFlowGraphImplShared.qll
@@ -342,7 +342,7 @@ private predicate succExitSplits(
   ControlFlowElement pred, Splits predSplits, CfgScope succ, SuccessorType t
 ) {
   exists(Reachability::SameSplitsBlock b, Completion c | pred = b.getAnElement() |
-    b.isReachable(predSplits) and
+    b.isReachable(succ, predSplits) and
     t = getAMatchingSuccessorType(c) and
     scopeLast(succ, pred, c) and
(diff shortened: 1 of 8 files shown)


ruby/ql/lib/codeql/ruby/controlflow/internal/ControlFlowGraphImplShared.qll
```diff
@@ -853,6 +855,23 @@ private module Cached {
    */
   cached
   ControlFlowElement getAControlFlowExitNode(ControlFlowElement cfe) { last(cfe, result, _) }
+
+  /**
+   * Gets the CFG scope of node `n`. Unlike `getCfgScope`, this predicate
+   * is calculated based on reachability from an entry node, and it may
+   * yield different results for AST elements that are split into multiple
+   * scopes.
+   */
```
hmac, line 864: Would it be possible to give a minimal example of code that would cause `getCfgScope` and `getNodeCfgScope` to give different results? I ran a brief Ruby query and couldn't find any cases, but maybe there's a simple one in C#?

hvitved: There are no examples for Ruby, but for C# we inline field initializations into the constructors, so that means that each initializer assignment may occur in multiple scopes. Example:

```csharp
class C
{
    public int Foo = 0;
    public C()
    {
    }

    public C(int i)
    {
        this.Foo += i;
    }
}
```

gets turned into

```csharp
class C
{
    public int Foo;
    public C()
    {
        Foo = 0;
    }

    public C(int i)
    {
        Foo = 0;
        this.Foo += i;
    }
}
```

aibaars (approved):
Looks good to me.

PR #7289, opened 2021-12-02
C#: Convert some of the existing flow summaries to CSV

michaelnebel:
In this PR we cover the conversion of flow summaries for

- System.Boolean.
- System.Uri.


csharp/ql/lib/semmle/code/csharp/frameworks/System.qll
```diff
@@ -92,6 +92,20 @@ class SystemBooleanStruct extends BoolType {
   }
 }
 
+/** Data flow for `System.Boolean`. */
+private class SystemBoolean32FlowModelCsv extends SummaryModelCsv {
```
hvitved, line 96: Remove `32`.

hvitved (changes requested):
Otherwise LGTM.

hvitved (approved):

PR #7293, opened 2021-12-02
Python: Arbitrary file read and delete

haby0:
From: https://github.com/github/codeql/pull/6977

A taint analysis was done on arbitrary file reading. Related frameworks were modeled, such as FastAPI, Flask view, etc.
In the Python path injection, some of the sinks where arbitrary files and folders are deleted are missing. I modeled this in this PR.

### DIFF
--- a/python/ql/lib/semmle/python/frameworks/Django.qll
+++ b/python/ql/lib/semmle/python/frameworks/Django.qll
@@ -1791,6 +1791,49 @@ module PrivateDjango {
        * See https://docs.djangoproject.com/en/3.1/topics/http/shortcuts/#redirect
        */
       API::Node redirect() { result = shortcuts().getMember("redirect") }
+
+      /**
+       * A call to the `django.shortcuts.render` function.
+       *
+       * See https://docs.djangoproject.com/en/3.1/topics/http/shortcuts/#render
+       */
+      class ShortcutsRenderCall extends HTTP::Server::HttpResponse::Range, DataFlow::CallCfgNode {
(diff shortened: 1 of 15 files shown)


RasmusWL (changes requested):
Thanks for this PR.

I think there is too much overlap with our path injection query, so as such, I do not want to accept this new query.

I do understand that knowing the user-controlled file is returned or deleted, has a higher signal-to-noise ratio than simply knowing there is path-injection; it's an interesting idea, but for now doesn't fit easily within how we do things (and we will need to think a bit more on how to properly incorporate this).

It seems there are a few additions to path injection sinks (like that `flask.send_file` is just an alias of `flask.helpers.send_file`), and a few improvements to our modeling of HTTP libraries. If you split these improvements into separate PRs, I'd be happy to look at including this :+1:

haby0:
> Thanks for this PR.
> 
> I think there is too much overlap with our path injection query, so as such, I do not want to accept this new query.
> 
> I do understand that knowing the user-controlled file is returned or deleted, has a higher signal-to-noise ratio than simply knowing there is path-injection; it's an interesting idea, but for now doesn't fit easily within how we do things (and we will need to think a bit more on how to properly incorporate this).
> 
> It seems there are a few additions to path injection sinks (like that `flask.send_file` is just an alias of `flask.helpers.send_file`), and a few improvements to our modeling of HTTP libraries. If you split these improvements into separate PRs, I'd be happy to look at including this 👍

Thanks for the reply. I am closed now.

PR #7295, opened 2021-12-02
C++: Add CWE tags to some queries.

geoffw0:
Add [CWE-326: Inadequate Encryption Strength](https://cwe.mitre.org/data/definitions/326.html) tag to `cpp/boost/tls-settings-misconfiguration`.  This query previously had the `security` tag but no specific CWE, and this is the closest one I can find.  The query is about using `boost::asio` for TLS or SSLv23 without disabling deprecated protocols such as TLS 1.1.

Add [CWE-327: Use of a Broken or Risky Cryptographic Algorithm](https://cwe.mitre.org/data/definitions/327.html) tag to `cpp/boost/use-of-deprecated-hardcoded-security-protocol`.  This query previously had the `security` tag but no specific CWE, and this is the closest one I can find.  The query is about explicitly specifying a deprecated protocol such as SSLv2 with the `boost::asio` library.

Add [CWE-260: Password in Configuration File](https://cwe.mitre.org/data/definitions/260.html) tag to `cpp/cleartext-storage-file`.  This query is already tagged with [CWE-313: Cleartext Storage in a File or on Disk](https://cwe.mitre.org/data/definitions/313.html), but we believe `CWE-260` has the benefit of association with [OWASP A05:2021 – Security Misconfiguration](https://owasp.org/Top10/A05_2021-Security_Misconfiguration/).  The query is about storing data we believe to be an unencrypted password or similarly sensitive information in a file.

### DIFF
--- a/cpp/ql/src/Likely Bugs/Protocols/TlsSettingsMisconfiguration.ql
+++ b/cpp/ql/src/Likely Bugs/Protocols/TlsSettingsMisconfiguration.ql
@@ -5,6 +5,7 @@
  * @problem.severity error
  * @id cpp/boost/tls-settings-misconfiguration
  * @tags security
+ *       external/cwe/cwe-326
  */
 
 import cpp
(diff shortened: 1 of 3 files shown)


MathiasVP (commented):
The changes LGTM! Two questions:
- Do we get any security severity for these queries now?
- Should this have a change-note?

geoffw0:
> Should this have a change-note?

I don't think so just for tags, but if you disagree I will write one.

MathiasVP:
> I don't think so just for tags, but if you disagree I will write one.

Hm, nah. I think it's fine without it 👍.

geoffw0:
> Do we get any security severity for these queries now?

Good point.  Done.

MathiasVP (approved):
Great! LGTM!

PR #7297, opened 2021-12-02
C#: Make it possible to define flow for property backing methods.

michaelnebel:
In this PR we make it possible to define flow for the backing methods for property methods for getting and setting values.
These were discarded prior to this to change as they were not "Members".


hvitved (approved):
Perfect!

PR #7304, opened 2021-12-03
C#: Convert flow summaries to CSV format.

michaelnebel:
The following is covered by this PR

- System.String
- System.Text.StringBuilder
- System.Lazy
- System.Nullable
- System.Convert


csharp/ql/lib/semmle/code/csharp/frameworks/system/Text.qll
```diff
@@ -24,6 +25,97 @@ class SystemTextStringBuilderClass extends SystemTextClass {
   Method getAppendFormatMethod() { result = this.getAMethod("AppendFormat") }
 }
 
+/** Data flow for `System.Text.StringBuilder`. */
+private class SystemTextStringBuilderFlowModelCsv extends SummaryModelCsv {
```
hvitved, line 29: Add a `private import semmle.code.csharp.frameworks.System` to the module `Frameworks` in `ExternalFlow.qll` to ensure bidirectional importing.

hvitved (changes requested):
LGTM, just one comment.

csharp/ql/lib/semmle/code/csharp/frameworks/system/Text.qll
hvitved: (Comment above updated)

michaelnebel: Will fix, but why is it needed?

hvitved (approved):
LGTM. But let's not merge before we see the result of the DCA run.

csharp/ql/lib/semmle/code/csharp/frameworks/system/Text.qll
hvitved: The flow summaries are compiled down to data flow nodes and atomic data-flow steps, which is a cached computation, so we want to make sure that this cached computation knows all flow summaries. If not, we may end up having to recompute this expensive computation.

PR #7305, opened 2021-12-03
Ruby: query to find user-controlled bypass of sensitive actions

nickrolfe:
Closes https://github.com/github/codeql-team/issues/419

I've put this query in the `experimental` folder, for now. I don't want to promote it as a default query because I haven't seen any real-world vulnerabilities of this type (CWE-807), so I'm not very confident about the heuristic for identifying a sensitive action, or that the kind of code pattern the query finds is how a user-controlled bypass bug would manifest in Ruby.


ruby/ql/lib/codeql/ruby/security/SensitiveActions.qll
```diff
@@ -0,0 +1,68 @@
+/**
+ * Provides classes and predicates for identifying sensitive data and methods for security.
+ *
+ * 'Sensitive' data in general is anything that should not be sent around in unencrypted form. This
+ * library tries to guess where sensitive data may either be stored in a variable or produced by a
+ * method.
```
alexrford, line 6: I guess that this file is based off the JS version of this library? Looks like it will be a good place to include heuristics for the existing `rb/hardcoded-credentials` query and for future queries about leaking sensitive data.

ruby/ql/lib/codeql/ruby/security/SensitiveActions.qll
```diff
@@ -0,0 +1,68 @@
...
+
+/** A call that may perform authorization. */
+class AuthorizationCall extends SensitiveAction, DataFlow::CallNode {
+  AuthorizationCall() {
+    exists(MethodCall c, string s |
+      c = this.asExpr().getExpr() and
+      s = c.getMethodName() // name contains `login` or `auth`, but not as part of `loginfo` or `unauth`;
+    |
+      // also exclude `author`
+      s.regexpMatch("(?i).*(log_?in(?!fo)|(?<!un)auth(?!or\\b)|verify).*") and
```
alexrford, line 53: Perhaps we can include the shared [`SensitiveDataHeuristics.qll`](https://github.com/github/codeql/blob/main/javascript/ql/lib/semmle/javascript/security/internal/SensitiveDataHeuristics.qll) and move this regex into `HeuristicNames`, e.g. as `string maybeAuthorizationMethodName()`? The JS and Ruby versions of this regex are slightly different at the moment as this one matches on `log_in` in addition to `login`.

ruby/ql/lib/codeql/ruby/security/ConditionalBypassCustomizations.qll
```diff
@@ -0,0 +1,58 @@
...
+ * adding your own.
+ */
+
+private import codeql.ruby.CFG
+private import codeql.ruby.DataFlow
+private import codeql.ruby.controlflow.BasicBlocks
+private import codeql.ruby.dataflow.RemoteFlowSources
+private import codeql.ruby.security.SensitiveActions
+
+module ConditionalBypass {
```
alexrford, line 13: Missing a QLDoc.

alexrford (commented):
Looks good, very minor suggestions.

Having this as an experimental query seems sensible for the moment. It doesn't seem to be generate many FPs based on the DCA run, though I also don't know of any real world positive examples to test this against.

ruby/ql/lib/codeql/ruby/security/SensitiveActions.qll
nickrolfe: Yes, it's based on the JS version.

ruby/ql/lib/codeql/ruby/security/ConditionalBypassCustomizations.qll
nickrolfe: We're a little bit inconsistent about this, but I notice that most of the other `*Customizations.qll` files don't have a QLDoc on the module they define.

ruby/ql/lib/codeql/ruby/security/SensitiveActions.qll
nickrolfe: Good idea. Mind if I do that as a followup PR?

ruby/ql/lib/codeql/ruby/security/ConditionalBypassCustomizations.qll
alexrford: In this case we're failing a PR check because of the missing QLDoc, it looks like that check only applies to changed files.

nickrolfe: Whoops, I didn't spot that. I'll fix it.

alexrford (approved):
LGTM

PR #7307, opened 2021-12-03
JS/ATM: Various compilation fixes and performance improvements

adityasharad:
This PR attempts to improve the evaluation performance of the ATM queries and libraries, focused on slow bottlenecks that take 10s or more on a moderately large database.

Semantics should be largely unchanged, with the exception of a few places where I've used the strict version of aggregates, which will now have no results (rather than the unit of the aggregation operation) if there are no tuples satisfying the aggregation expressions.

Commit-wise review highly recommended. See commit descriptions and inline comments for further explanation of each change. Careful testing with the actual models against some candidate databases highly recommended, to make sure I haven't changed how features are represented.

### DIFF
--- a/javascript/ql/experimental/adaptivethreatmodeling/lib/experimental/adaptivethreatmodeling/EndpointFeatures.qll
+++ b/javascript/ql/experimental/adaptivethreatmodeling/lib/experimental/adaptivethreatmodeling/EndpointFeatures.qll
@@ -277,10 +277,11 @@ private string getASupportedFeatureName() {
  * `featureValue` for the endpoint `endpoint`.
  */
 predicate tokenFeatures(DataFlow::Node endpoint, string featureName, string featureValue) {
-  featureName = getASupportedFeatureName() and
+  ModelScoring::endpoints(endpoint) and
   (
-    featureValue = unique(string x | x = getTokenFeature(endpoint, featureName))
-    or
-    not exists(unique(string x | x = getTokenFeature(endpoint, featureName))) and featureValue = ""
+    if strictcount(getTokenFeature(endpoint, featureName)) = 1
+    then featureValue = getTokenFeature(endpoint, featureName)
+    // Performance note: this is a Cartesian product between all endpoints and feature names.
+    else (featureValue = "" and featureName = getASupportedFeatureName())
   )
 }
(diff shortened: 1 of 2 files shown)


erik-krogh:
LGTM :+1: 

[An evaluation](https://github.com/github/codeql-dca-main/tree/data/erik-krogh/pr-7307-nightly-lgtm-full/reports) looks great :+1:  

But you need to run the autoformatter on `EndpointFeatures.qll`.

esbena:
> An evaluation looks great +1

While impressive, the evaluation is not representative since it does not make use of the ml-model. A small DCA change is required to download the model on the fly. 
I'm about to look into fixing ATM FPs so I might just add DCA ml-model support immediately.

henrymercer:
https://github.com/github/codeql/pull/7307/commits/0e31439b7e4075d2c2dedd0d45ab3d7c811ec866 appears to some small side effects where we duplicate some tokens in the function body feature.  I've pushed a couple of commits: the first removes that side effect, and the second autoformats `EndpointFeatures.qll` so that the language tests should start passing.

henrymercer (commented):
Will do some final testing of this internally, but LGTM otherwise.

henrymercer:
End-to-end evaluations: https://github.com/github/ml-ql-adaptive-threat-modeling-backend/actions/workflows/evaluate-boosted-query.yml?query=branch%3Ahenrymercer%2Fupdate-for-ql-optimizations

henrymercer (approved):
LGTM for the ATM changes.  There look to be a couple of failing tests in the standard JS library.

adityasharad:
Looks like the regex-combining wasn't actually behaviour-preserving. Will see if I can fix that so it has no diffs, otherwise back out that change.

henrymercer (approved):
LGTM.  Could I get a signoff from @github/codeql-javascript on the standard library changes?

erik-krogh (approved):
> LGTM. Could I get a signoff from @github/codeql-javascript on the standard library changes? 

Yes you can :+1: 
I got [an evaluation running of the stdlib changes](https://github.com/github/codeql-dca-main/tree/data/erik-krogh/pr-7307-nightly-lgtm-full-2/reports), and that looks good.  

I have tried to fix the performance of `FrameworkLibraries.qll` a few times myself, but without success. 
So it's great to see that someone was finally able to do it :tada:

PR #7314, opened 2021-12-06
C#: Update tests dataflow/global tests to use stubs.

michaelnebel:
It seems like a good idea to keep the code / dll's used for these tests in sync with the code / dll's used in the flow summaries as the output of the dataflow/global test cases depend on the defined summaries (which will differ when the code is changed from the ql dataflow models to hardcoded CSV rows).

In this PR we update the dataflow/global tests for C# to use stubs instead of dll's.


csharp/ql/test/library-tests/dataflow/global/DataFlow.expected
```diff
@@ -1,14 +1,11 @@
 | Capture.cs:12:19:12:24 | access to local variable sink27 |
 | Capture.cs:21:23:21:28 | access to local variable sink28 |
-| Capture.cs:30:19:30:24 | access to local variable sink29 |
```
hvitved, line 3: Hmm, something is wrong: we should see any changes to this file. Perhaps some stubs are missing?

michaelnebel: The stubs are there, but for some reason the model for "Select" seems be ignored by the test. If I manually implemeneted the body of Select the "missing" rows in the DataFlow.expected related to Select will re-appear.

hvitved: As discussed offline, this is because stubs are no longer treated as `fromLibrary`. https://github.com/github/codeql/pull/7314/commits/e9583fdf217a032ead1989748a249f63602a7d51 fixes this.

hvitved (approved):

hvitved (approved):

PR #7317, opened 2021-12-06
Ruby: Restructure `ParameterNode(Impl)`

hvitved:
Use `abstract` predicates internally, and make `ParameterNode::getParameter` `final`.


aibaars (approved):
LGTM

PR #7320, opened 2021-12-06
C#: Populate `UnknownType`

hvitved:
It turns out that this entity was never populated. It is required to avoid e.g. data-flow nodes without types.

Identified via https://github.com/github/codeql/pull/7231.

### DIFF
--- a/csharp/ql/lib/semmle/code/csharp/TypeRef.qll
+++ b/csharp/ql/lib/semmle/code/csharp/TypeRef.qll
@@ -16,8 +16,7 @@ private class TypeRef extends @typeref {
     typeref_type(this, result)
     or
     not typeref_type(this, _) and
-    result instanceof UnknownType and
-    types(result, _, "<unknown type>")
+    result instanceof UnknownType
   }
 }
 
--- a/csharp/ql/lib/semmle/code/csharp/TypeRef.qll
+++ b/csharp/ql/lib/semmle/code/csharp/TypeRef.qll
@@ -12,7 +12,13 @@ private class TypeRef extends @typeref {
 
   string toString() { result = this.getName() }
 
-  Type getReferencedType() { typeref_type(this, result) }
+  Type getReferencedType() {
+    typeref_type(this, result)
+    or
+    not typeref_type(this, _) and
+    result instanceof UnknownType and
+    types(result, _, "<unknown type>")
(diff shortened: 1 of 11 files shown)


csharp/ql/lib/semmle/code/csharp/TypeRef.qll
```diff
@@ -12,7 +12,13 @@ private class TypeRef extends @typeref {
 
   string toString() { result = this.getName() }
 
-  Type getReferencedType() { typeref_type(this, result) }
+  Type getReferencedType() {
+    typeref_type(this, result)
+    or
+    not typeref_type(this, _) and
+    result instanceof UnknownType and
+    types(result, _, "<unknown type>")
```
tamasvajk, line 20: Isn't `not typeref_type(this, _) and result instanceof UnknownType` enough?

csharp/ql/test/library-tests/expressions/PrintAst.expected
```diff
@@ -825,6 +825,8 @@ expressions.cs:
 #  122|                   9: [TypeofExpr] typeof(...)
 #  122|                     0: [TypeAccess] access to type Y<,>
 #  122|                       0: [TypeMention] Y<T, U>
+#  122|                         1: [TypeMention] <unknown type>
+#  122|                         2: [TypeMention] <unknown type>
```
tamasvajk, line 829: Are these expected? This is extracted from `typeof(Y<,>)`

csharp/ql/lib/semmle/code/csharp/TypeRef.qll
hvitved: Yes, we should only ever have one `UnknownType` entity.

csharp/ql/test/library-tests/expressions/PrintAst.expected
hvitved: Not sure if it is expected, but this is because `Y<T>` in `class Y<T> { ... }` and `Y<>` are considered the same (the unbound declaration). So in `typeof(Y<>)` there is an implicit type argument, which is the type parameter itself, and apparently Roslyn struggles with that and gives it kind `TypeKind.Error`.

tamasvajk (approved):

PR #7321, opened 2021-12-06
C#: Avoid CIL instructions with multiple types

hvitved:
Identified via https://github.com/github/codeql/pull/7231.


csharp/ql/lib/semmle/code/cil/ControlFlow.qll
```diff
@@ -56,6 +56,32 @@ class ControlFlowNode extends @cil_controlflow_node {
...
+      op1 = this.getOperand(i) and
+      op2 = this.getOperand(i) and
+      op1 != op2 and
+      result = op1.getType() and
+      t2 = op2.getType()
+    |
+      result = t2
+      or
+      result.(PrimitiveType).getUnderlyingType().getConversionIndex() >
+        t2.(PrimitiveType).getUnderlyingType().getConversionIndex()
```
tamasvajk, line 79: Does this basically check if there's an implicit numeric conversion from `t2` to `t1`? If so, don't we need to also cover other implicit conversions?

csharp/ql/lib/semmle/code/cil/InstructionGroups.qll
```diff
@@ -73,8 +73,8 @@ class ComparisonOperation extends BinaryExpr, @cil_comparison_operation {
 class BinaryArithmeticExpr extends BinaryExpr, @cil_binary_arithmetic_operation {
   override Type getType() {
     exists(Type t0, Type t1 |
-      t0 = this.getOperand(0).getType().getUnderlyingType() and
```
tamasvajk, line 76: Maybe adding a comment onto `Expr::getType()` mentioning that there's an alternative `getOperandType` which should be used instead of `getOperand(X).getType()` would be beneficial. 

Should all instances be changed?

I found that `ex.getExpr().getType()` also calls `getOperand(X).getType` in https://github.com/github/codeql/blob/c9f1c9839065d5b63bd6084d64f209ed417cf6eb/csharp/ql/lib/semmle/code/cil/CallableReturns.qll#L28

csharp/ql/lib/semmle/code/cil/ControlFlow.qll
hvitved: Yes, this only covers implicit conversions. There is also a case for when one of the operands is `null`. I haven't run into other cases.

tamasvajk (approved):

PR #7322, opened 2021-12-06
C++: Fix performance of 'cpp/unused-static-function'.

MathiasVP:
This PR fixes some scary tuple counts on the `cpp/unused-static-function` query with some manual magic:

```ql
Tuple counts for #UnusedStaticFunctions::Thing::callsOrAccesses_dispredPlus#fb#flipped/2@i19#da588w9l after 35.7s:
    46418979  ~0%      {2} r1 = SCAN #UnusedStaticFunctions::Thing::callsOrAccesses_dispredPlus#fb#flipped#prev_delta OUTPUT In.1, In.0 'result'
    121519019 ~42%     {2} r2 = JOIN r1 WITH UnusedStaticFunctions::Thing::callsOrAccesses_dispred#ff_10#join_rhs ON FIRST 1 OUTPUT Rhs.1 'this', Lhs.1 'result'
    48732165  ~7%      {2} r3 = r2 AND NOT #UnusedStaticFunctions::Thing::callsOrAccesses_dispredPlus#fb#flipped#prev(Lhs.1 'result', Lhs.0 'this')
    48732165  ~4%      {2} r4 = SCAN r3 OUTPUT In.1 'result', In.0 'this'
                                         return r4
```

### DIFF
--- a/cpp/ql/src/Best Practices/Unused Entities/UnusedStaticFunctions.ql
+++ b/cpp/ql/src/Best Practices/Unused Entities/UnusedStaticFunctions.ql
@@ -50,6 +50,16 @@ predicate reachableThing(Thing t) {
   exists(Thing mid | reachableThing(mid) and mid.callsOrAccesses() = t)
 }
 
+pragma[nomagic]
+predicate callsOrAccessesPlus(Thing thing1, FunctionToRemove thing2) {
+  thing1.callsOrAccesses() = thing2
+  or
+  exists(Thing mid |
+    thing1.callsOrAccesses() = mid and
+    callsOrAccessesPlus(mid, thing2)
+  )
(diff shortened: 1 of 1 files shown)


geoffw0 (approved):
I don't see a speed-up locally, but I don't see a slow down either.  Most likely my local setup doesn't have the change to magic that's suspected to cause the problem in the first place.

Anyway I'm happy with the QL.

MathiasVP:
> I don't see a speed-up locally, but I don't see a slow down either. Most likely my local setup doesn't have the change to magic that's suspected to cause the problem in the first place.

FWIW, you should be able to verify it by pulling the latest changes in the internal repo and rebuilding the CLI.

PR #7323, opened 2021-12-06
JS: Performance improvements to libraries using regex matching

adityasharad:
A series of attempted improvements to library code that performs regex matching against potentially large sets of candidate strings on large databases. Most useful for ATM but also affects some standard queries that look for sensitive operations and framework libraries.

Commit-wise review recommended. In order:
- In `SensitiveActions.qll`: factor out a helper predicate that restricts the set of candidate `name`s before computing the Cartesian product with the set of regexes. This product is necessary before we can do a `regexpMatch`.
- In `GeneratedCode.qll`: multiline regex match against `Comment.getText()` rather than `Comment.getLine(_)`. This results in fewer and smaller scans of comment text before regex matching.
- In `FrameworkLibraries.qll`: Combine two regexes into a single one. This results in only 1 scan of the comments table and  regex match operation.
  - This is the only intended behaviour change. The combined regex is slightly more permissive than the original two, since it allows a combination of the two matched formats. A string that matches one of the original regexes will, however, match the combined regex.


adityasharad:
Fixed combined regex (and hence failing test). Fresh evaluation in progress.

Z80coder:
Hi @adityasharad can you assign someone to review your PR and get this merged as soon as possible? Thanks.

erik-krogh (approved):
Changes looks good to me, tests are passing, and the evaluation suggests that there might be a slight performance improvement.  

:+1:

erik-krogh (approved):
Changes looks good to me, tests are passing, and the evaluation suggests that there might be a slight performance improvement.  

:+1:

PR #7325, opened 2021-12-07
Ruby: Don't count private methods as Rails actions

hmac:
Private instance methods on ActionController classes aren't valid
request handlers. Routing to them will raise an exception.


hvitved (approved):

hvitved (dismissed):

hmac:
@hvitved sorry about the false starts on this, I missed updating a few fixtures. Should be good to go now.

hvitved (approved):

PR #7326, opened 2021-12-07
fix dependency cycle by removing superfluous classes

ginsbach:
These classes were introducing dependency cycles during type compilation. This only even compiled due to a bug in the compiler: https://github.com/github/semmle-code/pull/41019

The problem (by example) is this: `ReflectedXSS::Sink` is an abstract class with only one extension (`ReflectedXSS::AnySink`). Therefore, computing this type requires calculating the type `ReflectedXSS::AnySink` first. At the same time, `ReflectedXSS::AnySink` is `instanceof Shared::Sink`, so that type needs to be calculated first. But then `Shared::Sink` is also abstract and requires `ReflectedXSS::Sink`, introducing a cycle.

These classes aren'y currently used, so I just removed them. Feel free to suggest a better approach.

This seems to be the only case of that pattern occurring in existing QL code, so it is blocking a PR that will raise an appropriate error message for these cases in the future (https://github.com/github/semmle-code/pull/41091).

### DIFF
--- a/ruby/ql/lib/codeql/ruby/security/XSS.qll
+++ b/ruby/ql/lib/codeql/ruby/security/XSS.qll
@@ -266,13 +266,13 @@ module ReflectedXSS {
   abstract class Source extends Shared::Source { }
 
   /** A data flow sink for stored XSS vulnerabilities. */
-  abstract class Sink extends Shared::Sink { }
+  class Sink = Shared::Sink;
 
   /** A sanitizer for stored XSS vulnerabilities. */
-  abstract class Sanitizer extends Shared::Sanitizer { }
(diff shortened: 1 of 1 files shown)


ginsbach:
There are test failures, so this probably isn't the correct approach to fixing this problem. @alexrford do you have a suggestion about how to do this the right way?

alexrford:
> There are test failures, so this probably isn't the correct approach to fixing this problem. @alexrford do you have a suggestion about how to do this the right way?

Thanks for the fix. I was trying out some other approaches locally, but I think your alias approach works. I've triggered a DCA run of this that we should let complete before merging this.

alexrford (approved):

PR #7328, opened 2021-12-07
Ruby: add `Customizations.qll` file

alexrford:
Allows for global customizations to things like remote flow sources - we have this mechanism in other languages already (JS, Python, C#, Java).


ruby/ql/lib/ruby.qll
```diff
@@ -1 +1,6 @@
+/**
+ * Provides classes for working with Ruby programs.
+ */
+
+import Customizations
```
hvitved, line 5: private import?

hvitved (approved):

PR #7333, opened 2021-12-07
Ruby: rename codeql-extractor.yml template file

hmac:
This file is copied into the extractor pack when it is built, but having
it in the source tree confuses the codeql CLI into thinking that
`codeql/ruby` is itself an extractor pack. This can cause errors when
running `codeql database create`.

This change renames the file to `codeql-extractor.template.yml`, so the
CLI ignores it.


hmac:
@aibaars does this seem sensible to you or am I just using the CLI incorrectly?

aibaars:
I'd rather not rename this file like that. None of the other languages do this, so there should be no need. Are you working in an up-to-date `github/codeql` checkout? There was a broken version that removed the `.codeqlmanifest.json` from `codeql/ruby` but forgot to remove the `codeql/ruby` folder from the top-level `.codeqlmanifest.json` file.

hmac:
I can reliably hit a CLI error with this workflow:

1. run `scripts/create-extractor-pack.sh`
2. `cd` to the root of some ruby project
3. run `codeql database create some-db-name -l ruby -s $(pwd) --search-path /path/to/codeql/ruby`

I can avoid it with `--search-path /path/to/codeql/ruby/extractor-pack`, but that directory onyl exists _if_ I've run `create-extractor-pack.sh`, and would only do that if we've made extractor changes since the last public release.

aibaars:
Indeed that won't work. You should use psth/to/codeql instead as the search path.

hmac:
Ah, yes, that works fine. So I was using the CLI incorrectly then! Thanks.

I don't really understand why `/codeql` works but `/codeql/ruby` doesn't but 🤷

PR #7334, opened 2021-12-07
Ruby: Resolve simple string interpolations

hmac:
When calculating `StringlikeLiteral.getValueText`, include results from
interpolations where we can determine their string value. For example:

```ruby
b = "b" # local variable
D = "d" # constant

"a#{b}c"     # getValueText() = "abc"
"a#{b}c{D}"  # getValueText() = "abcd"
/#a#{b}c{D}/ # getValueText() = "abcd"
```

Because this also applies to regexes, it increases the sensitivity of our ReDoS queries. With this change we find a true positive in https://github.com/advisories/GHSA-jxhc-q857-3j6g.

There's one shortcoming to this change: ReDoS alerts may not include the correct location of the problematic (sub)string if it is included via an interpolation. The alert itself, and the suggested problematic input, will still be correct.


ruby/ql/lib/codeql/ruby/ast/Literal.qll
```diff
@@ -321,9 +322,16 @@ class StringInterpolationComponent extends StringComponent, StmtSequence,
...
   final override Stmt getStmt(int n) { toGenerated(result) = g.getChild(n) }
 
-  final override string getValueText() { none() }
-
   final override string getAPrimaryQlClass() { result = "StringInterpolationComponent" }
+
+  // If last statement in the interpolation is a constant or local variable read,
+  // we attempt to look up its string value.
+  // If there's a result, we return that as the string value of the interpolation.
+  final override string getValueText() {
```
hvitved, line 330: This (and the `getValueText` further down) should be implemented in `CfgNodes.qll` instead (i.e., on CFG nodes instead of AST nodes). In general, if `getValueText` is defined recursively, it should be done in `CfgNodes` for potential higher precision (in context of CFG splitting). Example
```rb
if (b)
  x = "true"
else
  x = "false"

if (b)
  puts x
```

If (when) Ruby has Boolean CFG splitting, the access to `x` in the `b = true` case will have constant value `"true"`, while it will have constant value `"false"` in the `b = false` case. However, the AST node `x` in `puts x` will not have a constant value.

hvitved (changes requested):

ruby/ql/lib/codeql/ruby/ast/Literal.qll
hmac: OK - I think I understand what you're saying here. I've updated this PR to push more calculation into the CFG. We still need a few changes at the AST level, I think, because regular string constants aren't CFG nodes. Let me know if I've got the right idea or if I'm still off.

hvitved:
I have added some suggested changes to this PR at https://github.com/github/codeql/pull/7440.

hmac:
Performance is improved from the last DCA run. Still slower than `main`, but not catastrophically.

`head~1`:

| source                                   | `a` targets | `b` targets | a\_m | a\_std | b\_m | b\_std | diff | relative |
| ---------------------------------------- | ----------- | ----------- | ---- | ------ | ---- | ------ | ---- | -------- |
| ruby\_\_ruby                             | 1           | 1           | 619  | 0      | 847  | 0      | 228  | 0.368    |
| opal\_\_opal                             | 1           | 1           | 987  | 0      | 1778 | 0      | 791  | 0.801    |


`head`:

source | a targets | b targets | a_m | a_std | b_m | b_std | diff | relative
-- | -- | -- | -- | -- | -- | -- | -- | --
opal__opal | 3 | 3 | 862.3 | 13.58 | 1060 | 27.87 | 197.7 | 0.229
ruby__ruby | 3 | 3 | 590.7 | 95.34 | 809.7 | 136.5 | 219 | 0.371

hvitved (approved):
Will merge if the latest DCA run looks good.

PR #7339, opened 2021-12-08
Python: Cache more predicates to improve performance. 

erik-krogh:
For this PR I have chosen not to rebase and squash, the commit history are the iterations I went through in making this PR.  
(And the commit-messages are accordingly). 

I recommend not reviewing each commit as much gets reverted along the way.  
But the commits give a good idea about the process that I followed.  

[nightly/lgtm-full evaluation](https://github.com/github/codeql-dca-main/tree/data/erik-krogh/pyPerf__nightly__lgtm-full__15/reports) (~11% average speedup).  
[nightly/lgtm-full evaluation (single-threaded)](https://github.com/github/codeql-dca-main/tree/data/erik-krogh/pyPerf__nightly__lgtm-full__16/reports) (~15% average speedup).  

The speedup gets larger the more queries are added in an evaluation, so the above represents a best-case performance gain.  
Single-query performance will be slower if you start with an empty cache.  

[Here is the DB size increase](https://github.com/github/codeql-dca-main/blob/data/erik-krogh/pyPerf__nightly__lgtm-full__15/reports/db-sizes/entire-db.afterNormalCleanup.txt). 

---- 

Here is the process I followed: 

1. Copy-paste the `CachedStages.qll` file from JavaScript into Python. 
2. Do a single-threaded DCA run on `main`, to figure out which queries are slow.  
3. Run a slow and advanced query. Check the output log and look sections starting with `Results in:`.  
    These sections tell you which predicates are the cached predicates in a given stage. 
    Use `CachedStages.qll` to group some of these stages in a way that makes sense. 
4. Look at the evaluation, find some good pairs of slow query and slow source. 
5. Get databases of the slow sources locally, run some slow queries with tuple counting, and fix the performance by doing these: 
    5.1. Look at the second to last `Clause timing report` (the report for the last stage) of a slow query, add `cache` to expensive non-query specific predicates (and add them to `CachedStages.qll`). 
    5.2. Fix any bad join orders that are introduced. 
    5.3. Look at `Results in:` again, and check if there are any `cached` predicates that should be grouped. 
    5.4. If you've accidentally grouped too many stages into one big group, then revert whatever you did. 
    5.5. Revert some caching if the DB size increases too much. 
    5.6. Revert some imports if I caused `ql/abstract-class-import` alerts. 
6. Do a single-threaded DCA run with your new code.  
   ( `Start an experiment` -> `Use saved experiment` -> `From previous experiment` is helpful). 
7. If you are happy with the new evaluation go to 8, else go to 4. 
8. Do a validation evaluation using the default `codeql-action` options. 
9. Open a PR.

### DIFF
--- a/python/ql/lib/semmle/python/Function.qll
+++ b/python/ql/lib/semmle/python/Function.qll
@@ -167,17 +167,19 @@ class Function extends Function_, Scope, AstNode {
 
 /** A def statement. Note that FunctionDef extends Assign as a function definition binds the newly created function */
 class FunctionDef extends Assign {
+  FunctionExpr f;
+
   /* syntax: def name(...): ... */
   FunctionDef() {
     /* This is an artificial assignment the rhs of which is a (possibly decorated) FunctionExpr */
-    exists(FunctionExpr f | this.getValue() = f or this.getValue() = f.getADecoratorCall())
+    this.getValue() = f or this.getValue() = f.getADecoratorCall()
   }
 
   override string toString() { result = "FunctionDef" }
 
   /** Gets the function for this statement */
   Function getDefinedFunction() {
-    exists(FunctionExpr func | this.containsInScope(func) and result = func.getInnerScope())
+    result = f.getInnerScope() // XXX: This behaves very differently. But from inspecting the results of the previous version, that had every function in the same scope as the result.
   }
 
   override Stmt getLastStatement() { result = this.getDefinedFunction().getLastStatement() }
--- a/python/ql/lib/semmle/python/Function.qll
+++ b/python/ql/lib/semmle/python/Function.qll
@@ -178,9 +178,7 @@ class FunctionDef extends Assign {
   override string toString() { result = "FunctionDef" }
 
   /** Gets the function for this statement */
-  Function getDefinedFunction() {
-    result = f.getInnerScope() // XXX: This behaves very differently. But from inspecting the results of the previous version, that had every function in the same scope as the result.
-  }
+  Function getDefinedFunction() { result = f.getInnerScope() }
 
   override Stmt getLastStatement() { result = this.getDefinedFunction().getLastStatement() }
 }
--- a/python/ql/lib/semmle/python/filters/GeneratedCode.qll
+++ b/python/ql/lib/semmle/python/filters/GeneratedCode.qll
@@ -51,7 +51,7 @@ private string comment_or_docstring(File f, boolean before_code) {
     c.getLocation().getFile() = f and
     result = c.getText()
   |
-    if isBeforeCode(c, f) then before_code = true else before_code = false
+    if isBeforeCode(c, f) then before_code = false else before_code = true
   )
   or
   exists(Module m | m.getFile() = f |
--- a/python/ql/lib/semmle/python/filters/GeneratedCode.qll
+++ b/python/ql/lib/semmle/python/filters/GeneratedCode.qll
@@ -41,7 +41,7 @@ private int minStmtLine(File file) {
 }
 
 pragma[nomagic]
-private predicate isBeforeCode(Comment c, File f) {
+private predicate isCommentAfterCode(Comment c, File f) {
   f = c.getLocation().getFile() and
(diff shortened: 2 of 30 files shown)


python/ql/lib/semmle/python/Function.qll
```diff
@@ -167,17 +167,19 @@ class Function extends Function_, Scope, AstNode {
...
-    exists(FunctionExpr f | this.getValue() = f or this.getValue() = f.getADecoratorCall())
+    this.getValue() = f or this.getValue() = f.getADecoratorCall()
   }
 
   override string toString() { result = "FunctionDef" }
 
   /** Gets the function for this statement */
   Function getDefinedFunction() {
-    exists(FunctionExpr func | this.containsInScope(func) and result = func.getInnerScope())
+    result = f.getInnerScope() // XXX: This behaves very differently. But from inspecting the results of the previous version, that had every function in the same scope as the result.
```
tausbn: Oh no. :astonished: 
This looks more correct to me.

python/ql/lib/semmle/python/pointsto/PointsTo.qll
```diff
@@ -1312,7 +1312,7 @@ module InterProceduralPointsTo {
    * `var = ...; foo(); use(var)`
    * Where var may be redefined in call to `foo` if `var` escapes (is global or non-local).
    */
-  pragma[noinline]
+  pragma[inline]
```
tausbn: This one was surprising to me, but I assume the performance tests show that it's a net improvement.

python/ql/lib/semmle/python/filters/GeneratedCode.qll
```diff
@@ -32,18 +32,26 @@ class GenericGeneratedFile extends GeneratedFile {
...
+    min(int line |
+      line = any(Stmt s | s.getLocation().getFile() = file).getLocation().getStartLine()
+    )
+}
+
+pragma[nomagic]
+private predicate isBeforeCode(Comment c, File f) {
+  f = c.getLocation().getFile() and
+  minStmtLine(f) < c.getLocation().getStartLine()
+}
```
tausbn, line 47: I know that you're just following the naming scheme set out in the file already, but isn't this predicate actually expressing that the comment `c` appears _after_ code?

tausbn (approved):
Thanks for doing this!
All of the changes look sensible to me (modulo one comment). :+1:

Now I guess I need to rejigger my intuitions for what happens in which stages...

python/ql/lib/semmle/python/pointsto/PointsTo.qll
erik-krogh: It's been a while, so I don't really remember.  
But I suppose it was an improvement.

python/ql/lib/semmle/python/Function.qll
erik-krogh: :+1:  
I'll delete the comment.

erik-krogh:
Code-scanning [was complaining about an `ql/name-casing` issue](https://github.com/github/codeql/security/code-scanning/32420) that was not introduced in this PR.  

(I did introduce it, but that was in the def-node PR).

python/ql/lib/semmle/python/filters/GeneratedCode.qll
tausbn: I'm very confused. I'm sure I saw a commit that fixed this, but it appears to be unfixed. :thinking:

erik-krogh: You definitely saw that, I'm not sure why that disappeared in [the rebase](https://github.com/github/codeql/compare/1ee4ba21f80d65a4a4192e12968df32570c7e699..1847a5713b0690fa8d39b171ed0971c44e2e6705).   

I reintroduced the fix, and also updated the expected output for the location test (that has an extra result due to the Django models getting imported).

tausbn (approved):
Nice! Let's get it in before it starts conflicting with other stuff. :shipit:

PR #7340, opened 2021-12-09
Ruby: handle private module methods

hmac:
Extend our modelling of private methods in two ways:

1. Extend `private` modelling to uses in modules, as well as classes.
2. Model `private_class_method`, which makes class methods (`def self.foo`) private.


alexrford (dismissed):
Looks like a nice improvement 👍

ruby/ql/lib/codeql/ruby/ast/Method.qll
```diff
@@ -48,30 +48,37 @@ private class MethodModifier extends MethodCall {
...
-  MethodBase getMethod() {
-    result = this.getMethodArgument()
-    or
-    exists(Namespace n |
-      n.getAStmt() = this and
-      n.getAStmt() = result and
-      result.getName() = this.getMethodArgument().(StringlikeLiteral).getValueText()
-    )
+  /** Holds if this call modifies a method with name `name` in namespace `n`. */
+  predicate modifiesMethod(Namespace n, string name) {
```
hvitved, line 52: Better safe than sorry: add `pragma [noinline]`.

ruby/ql/lib/codeql/ruby/ast/Method.qll
```diff
@@ -132,13 +139,7 @@ class Method extends MethodBase, TMethod {
...
    */
   override predicate isPrivate() {
-    this = any(Private p).getMethod()
-    or
-    exists(Namespace c, Private p, int i, int j |
-      p.hasNoArg(c, i) and
-      this = c.getStmt(j) and
-      j > i
-    )
+    any(Private p).modifiesMethod(this.getEnclosingModule(), this.getName())
```
hvitved, line 142: This can still join on just `this.getName()`, so it is better to add a top-level predicate
```ql
pragma[noinline]
private predicate isDeclaredIn(MethodBase m, Namespace n, string name) {
  n = m.getEnclosingModule() and
  name = m.getName()
}
```

and then change the above to

```ql
exists(Namespace n, string name |
  any(Private p).modifiesMethod(n, name) and
  isDeclaredIn(this, n, name)
)
```

ruby/ql/lib/codeql/ruby/ast/Method.qll
```diff
@@ -203,7 +204,9 @@ class SingletonMethod extends MethodBase, TSingletonMethod {
    * end
    * ```
    */
-  override predicate isPrivate() { this = any(PrivateClassMethod p).getMethod() }
+  override predicate isPrivate() {
+    any(PrivateClassMethod p).modifiesMethod(this.getEnclosingModule(), this.getName())
```
hvitved, line 208: Same as above.

hvitved (changes requested):

ruby/ql/lib/codeql/ruby/ast/Method.qll
hmac: Interesting! That seems to work, and from some rough testing it also looks faster as well. But I don't really understand why it is faster. Why is it that the previous version can join on just `this.getName()`, rather than `this.getName()` _and_ `this.getEnclosingModule()`? Are there some docs that I could read to better understand this?

hvitved: The code above involves three relations (`modifiesMethod`, `getEnclosingModule`, and `getName`), so the join-orderer has to pick one of the 6 possible join-orders. The join-order `modifiesMethod <-> getName <-> getEnclosingModule` is the one we want to avoid. By folding `getEnclosingModule` and `getName` into a separate predicate, we are effectively making sure that `modifiesMethod` joins on the two relations simultaneously, and there is only one possible (good) join-order.

You can try to look at the Datalog Intermediate Language (DIL) code before and after (right-click a query run in VS Code and select "View DIL").

hvitved (approved):

ruby/ql/lib/codeql/ruby/ast/Method.qll
hmac: Ah that makes sense! Had I thought about it a bit harder I probably should have guessed something like that 🤔 . Thanks for the explanation.

PR #7341, opened 2021-12-09
Ruby: treat ActionController#cookies as a remote flow source

nickrolfe:


alexrford (approved):
LGTM with one comment.

On a side note (not as a suggested change for this PR), `rb/reflected-xss` currently considers any `RemoteFlowSource` as a potential flow source. The implicit assumption is that the value from the source is controlled by the attacker, but that the input itself is inadvertently submitted by the victim.

In some cases, such as the attacker convincing the victim to click on a link with malicious query params, this is a realistic assumption. However, it seems less feasible if the attacker has to manipulate the victim into a SelfXSS attack or similar in order to set a cookie. I think that this means we should limit potential `rb/reflected-xss` sources to ones where this assumption may hold.

alexrford (approved):

PR #7343, opened 2021-12-09
Shared CFG: Add another consistency test

hvitved:
Finds nodes with multiple normal successors, where one is the special simple successor. For example, this would flag a node that has both a "simple" and a "true" successor.

### DIFF
--- a/csharp/ql/lib/semmle/code/csharp/controlflow/internal/ControlFlowGraphImplShared.qll
+++ b/csharp/ql/lib/semmle/code/csharp/controlflow/internal/ControlFlowGraphImplShared.qll
@@ -957,10 +957,31 @@ module Consistency {
     not split.hasEntry(pred, succ, c)
   }
 
+  private class SimpleSuccessorType extends SuccessorType {
+    SimpleSuccessorType() {
+      this = getAMatchingSuccessorType(any(Completion c | completionIsSimple(c)))
+    }
+  }
+
+  private class NormalSuccessorType extends SuccessorType {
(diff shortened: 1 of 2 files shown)


csharp/ql/lib/semmle/code/csharp/controlflow/internal/ControlFlowGraphImplShared.qll
```diff
@@ -963,6 +963,20 @@ module Consistency {
     successor = getASuccessor(node, t)
   }
 
+  query predicate simpleAndNormalSuccessors(Node node, SuccessorType t, Node successor) {
+    not node instanceof TEntryNode and
```
aibaars, line 967: why exclude `TEntryNode` here?

hvitved: We probably don't need that after all. We need it for `multipleSuccessors` though, since C# allows for methods with multiple bodies (when a method with the same fully qualified name, but different implementations, gets extracted multiple times).

aibaars: ok, makes sense. Although the edges from the entry node should all be labelled "simple", right?

hvitved: Yes, I can strengthen the check to that

aibaars (approved):
LGTM

PR #68, opened 2018-08-17
JS: Improved inter-procedural type inference

esben-semmle:
As discussed previously, this PR extends the interprocedural type inference for function declarations that provably have a single, simple, call site. These function declarations are now analyzed in the same way as IIFEs, i.e. their argument- and return-flow are analyzed.

   The [performance overhead](https://git.semmle.com/gist/esben/1a29161ec4a42fe848c77757405d502c) is negligible, and the new alerts are impressive at times.

   This PR is responsible for true positives of for several queries:
   - many `js/superfluous-trailing-arguments`, `js/comparison-between-incompatible-types`, `js/trivial-conditional`
   - some `js/property-access-on-non-object`, `js/call-to-non-callable`, `js/implicit-operand-conversion`

   This PR also flags a `js/path-injection`, although that seems like a FP in practice.

   Some alert filters have been added to prevent us from annoying programmers with some kinds of true positives that this PR enable. These true positives mainly appear when a polymorphic function only is used in a monomorphic way.

PR #7344, opened 2021-12-09
JS: Improve inter-procedural type inference for FunctionExpr

SZFsir:
To improve inter-procedural type inference for FunctionExpr.

codeql conduct an [Inter-procedural type inference](https://github.com/github/codeql/blob/753d886b0dd8217cfa3fafcc36afd9de9a2aed26/javascript/ql/lib/semmle/javascript/dataflow/internal/InterProceduralTypeInference.qll#L194) for `LocalFunction` .  But `LocalFunction` represents `FunctionDeclStmt` which is only called once locally, so we can easily handle the case below. 

```javascript
function f(source){return source;}
function c(source){return source;}
var sink = c(f)(easySource);
```

But codeql can not handle the case below. Function c is declare by a `FunctionExpr`. 
```javascript
function f(source){return source;}
var c = function c(source){return source;}
var sink = c(f)(easySource);
```

This usage is common so I improve it by extending the LocalFunction definition. BTW, this issue is mentioned in the previous [pull request](https://github.com/github/codeql/pull/68/files#r211149568) but still not improved.

I test all unit tests [before](https://blog.szfszf.top/static/codeqltest/before.json) and [after](https://blog.szfszf.top/static/codeqltest/after.json) modification, I found there are no noticeable performance changes.

### DIFF
--- a/javascript/ql/lib/semmle/javascript/dataflow/internal/InterProceduralTypeInference.qll
+++ b/javascript/ql/lib/semmle/javascript/dataflow/internal/InterProceduralTypeInference.qll
@@ -203,19 +203,14 @@ class LocalFunction extends Function {
   DataFlow::Impl::ExplicitInvokeNode invk;
 
   LocalFunction() {
-    (
-      exists(LocalVariable v |
-        getOnlyAccess(this, v) = invk.getCalleeNode().asExpr() and
-        not exists(v.getAnAssignedExpr()) and
-        not exists(ExportDeclaration export | export.exportsAs(v, _))
-      )
-    or
-      exists(LocalVariable v |
-        getOnlyAccessToFunctionExpr(this, v) = invk.getCalleeNode().asExpr() and
-        not exists(ExportDeclaration export | export.exportsAs(v, _))
-      )
-    )
-    and
+    exists(LocalVariable v |
+      getOnlyAccess(this, v) = invk.getCalleeNode().asExpr() and
+      not exists(v.getAnAssignedExpr()) and
+      not exists(ExportDeclaration export | export.exportsAs(v, _))
+      or
+      getOnlyAccessToFunctionExpr(this, v) = invk.getCalleeNode().asExpr() and
+      not exists(ExportDeclaration export | export.exportsAs(v, _))
+    ) and
     // if the function is non-strict and its `arguments` object is accessed, we
     // also assume that there may be other calls (through `arguments.callee`)
     (isStrict() or not usesArgumentsObject())
--- a/javascript/ql/lib/semmle/javascript/dataflow/internal/InterProceduralTypeInference.qll
+++ b/javascript/ql/lib/semmle/javascript/dataflow/internal/InterProceduralTypeInference.qll
@@ -190,16 +190,32 @@ private VarAccess getOnlyAccess(FunctionDeclStmt fn, LocalVariable v) {
   result = unique(VarAccess acc | acc = v.getAnAccess())
 }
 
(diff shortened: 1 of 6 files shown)


javascript/ql/lib/semmle/javascript/dataflow/internal/InterProceduralTypeInference.qll
```diff
@@ -190,16 +190,32 @@ private VarAccess getOnlyAccess(FunctionDeclStmt fn, LocalVariable v) {
   result = unique(VarAccess acc | acc = v.getAnAccess())
 }
 
+private VarAccess getOnlyAccessToFunctionExpr(FunctionExpr fn, LocalVariable v) {
+  exists( DeclStmt st |
+    fn = st.(DeclStmt).getADecl().getInit() and 
+    v = st.(DeclStmt).getADecl().getBindingPattern().getVariable() and
+    result = unique(VarAccess acc | acc = v.getAnAccess())
+  )
+}
```
erik-krogh, line 199:
```suggestion
private VarAccess getOnlyAccessToFunctionExpr(FunctionExpr fn, LocalVariable v) {
  exists(VariableDeclarator decl |
    fn = decl.getInit() and
    v = decl.getBindingPattern().getVariable() and
    result = unique(VarAccess acc | acc = v.getAnAccess())
  )
}
```

Two small issues here: 
1: A `DeclStmt` might declare multiple variables (e.g. `var i = 1, j = 2`), and with your code it is possible to consider the function from one declaration and the variable from another. You need to ensure that the two `getADecl()` are the same declaration (or use `VariableDeclarator` directly).    
2: You need to run the auto-formatter (in VSCode: Ctrl + Shift + p -> `Format Document`).   

My suggestion does both, it's up to you whether you want to use it.

javascript/ql/lib/semmle/javascript/dataflow/internal/InterProceduralTypeInference.qll
```diff
@@ -190,16 +190,32 @@ private VarAccess getOnlyAccess(FunctionDeclStmt fn, LocalVariable v) {
...
+        not exists(v.getAnAssignedExpr()) and
+        not exists(ExportDeclaration export | export.exportsAs(v, _))
+      )
+    or
+      exists(LocalVariable v |
+        getOnlyAccessToFunctionExpr(this, v) = invk.getCalleeNode().asExpr() and
+        not exists(ExportDeclaration export | export.exportsAs(v, _))
+      )
+    )
+    and
```
erik-krogh, line 218:
```suggestion
    exists(LocalVariable v |
      getOnlyAccess(this, v) = invk.getCalleeNode().asExpr() and
      not exists(v.getAnAssignedExpr()) and
      not exists(ExportDeclaration export | export.exportsAs(v, _))
      or
      getOnlyAccessToFunctionExpr(this, v) = invk.getCalleeNode().asExpr() and
      not exists(ExportDeclaration export | export.exportsAs(v, _))
    ) and
```

My suggestion here is purely stylistic and I haven't changed the behavior.  

It's up to you whether you want to use my suggestion, but if you don't, then you have to run the autoformatter.

erik-krogh (commented):
Nice :+1: 

You code is almost good to go (see comments below). 

Really nice that you found a new TP among our own test-cases :tada:

erik-krogh (approved):
Looks good :+1: nice work.  

I did my own performance evaluation, and that turned out just fine. 

So we are good to merge :tada:

PR #7349, opened 2021-12-10
Dataflow: Add support for flow state

aschackmull:
Flow state aka flow labels.

The type `DataFlow::FlowState` is added as an abstract class extending string, and 4 new predicates are added to `Configuration`s:
```
  predicate isSource(Node source, FlowState state)
  predicate isSink(Node source, FlowState state)
  predicate isBarrier(Node node, FlowState state)
  predicate isAdditionalFlowStep(Node node1, FlowState state1, Node node2, FlowState state2)
```
The existing `isSource` and `isSink` use the default state `""`, and the existing steps become state-preserving and applies to all states.

Furthermore, the state can be inspected after calculating data flow using `PathNode.getState()`.

To consider:
* Should we change the existing `isSource` and `isSink` from `abstract` to `{ none() }`?
* Should the state be part of `PathNode.toString` such that it shows up in path explanations? If so, what format do we prefer?

### DIFF
--- a/cpp/ql/lib/semmle/code/cpp/dataflow/internal/DataFlowImpl.qll
+++ b/cpp/ql/lib/semmle/code/cpp/dataflow/internal/DataFlowImpl.qll
@@ -57,15 +57,32 @@ abstract class Configuration extends string {
   abstract predicate isSource(Node source);
 
   /**
+   * Holds if `source` is a relevant data flow source with the given initial
+   * `state`.
+   */
+  predicate isSource(Node source, FlowState state) { none() }
+
+  /**
    * Holds if `sink` is a relevant data flow sink.
    */
   abstract predicate isSink(Node sink);
+
+  /**
+   * Holds if `sink` is a relevant data flow sink accepting `state`.
+   */
+  predicate isSink(Node source, FlowState state) { none() }
 
   /**
    * Holds if data flow through `node` is prohibited. This completely removes
    * `node` from the data flow graph.
    */
   predicate isBarrier(Node node) { none() }
+
+  /**
+   * Holds if data flow through `node` is prohibited when the flow state is
+   * `state`.
+   */
+  predicate isBarrier(Node node, FlowState state) { none() }
 
   /** Holds if data flow into `node` is prohibited. */
   predicate isBarrierIn(Node node) { none() }
@@ -81,6 +98,15 @@ abstract class Configuration extends string {
    * into account in the analysis.
    */
   predicate isAdditionalFlowStep(Node node1, Node node2) { none() }
+
+  /**
+   * Holds if the additional flow step from `node1` to `node2` must be taken
+   * into account in the analysis. This step is only applicable in `state1` and
+   * updates the flow state to `state2`.
+   */
+  predicate isAdditionalFlowStep(Node node1, FlowState state1, Node node2, FlowState state2) {
+    none()
+  }
 
   /**
    * Holds if an arbitrary number of implicit read steps of content `c` may be
@@ -197,9 +223,15 @@ abstract private class ConfigurationRecursionPrevention extends Configuration {
   override predicate hasFlow(Node source, Node sink) {
     strictcount(Node n | this.isSource(n)) < 0
     or
+    strictcount(Node n | this.isSource(n, _)) < 0
+    or
     strictcount(Node n | this.isSink(n)) < 0
     or
+    strictcount(Node n | this.isSink(n, _)) < 0
+    or
     strictcount(Node n1, Node n2 | this.isAdditionalFlowStep(n1, n2)) < 0
+    or
+    strictcount(Node n1, Node n2 | this.isAdditionalFlowStep(n1, _, n2, _)) < 0
     or
     super.hasFlow(source, sink)
   }
@@ -277,7 +309,7 @@ private predicate inBarrier(NodeEx node, Configuration config) {
   exists(Node n |
     node.asNode() = n and
     config.isBarrierIn(n) and
-    config.isSource(n)
+    (config.isSource(n) or config.isSource(n, _))
   )
 }
 
@@ -285,7 +317,7 @@ private predicate outBarrier(NodeEx node, Configuration config) {
   exists(Node n |
     node.asNode() = n and
     config.isBarrierOut(n) and
-    config.isSink(n)
+    (config.isSink(n) or config.isSink(n, _))
   )
 }
 
@@ -295,10 +327,12 @@ private predicate fullBarrier(NodeEx node, Configuration config) {
     config.isBarrier(n)
     or
     config.isBarrierIn(n) and
-    not config.isSource(n)
+    not config.isSource(n) and
+    not config.isSource(n, _)
     or
     config.isBarrierOut(n) and
-    not config.isSink(n)
+    not config.isSink(n) and
+    not config.isSink(n, _)
     or
     exists(BarrierGuard g |
       config.isBarrierGuard(g) and
@@ -308,13 +342,34 @@ private predicate fullBarrier(NodeEx node, Configuration config) {
 }
 
 pragma[nomagic]
-private predicate sourceNode(NodeEx node, Configuration config) {
-  config.isSource(node.asNode()) and
-  not fullBarrier(node, config)
+private predicate stateBarrier(NodeEx node, FlowState state, Configuration config) {
+  exists(Node n |
+    node.asNode() = n and
+    config.isBarrier(n, state)
+  )
 }
 
 pragma[nomagic]
-private predicate sinkNode(NodeEx node, Configuration config) { config.isSink(node.asNode()) }
+private predicate sourceNode(NodeEx node, FlowState state, Configuration config) {
+  (
+    config.isSource(node.asNode()) and state instanceof FlowStateEmpty
+    or
+    config.isSource(node.asNode(), state)
+  ) and
+  not fullBarrier(node, config) and
+  not stateBarrier(node, state, config)
+}
+
+pragma[nomagic]
+private predicate sinkNode(NodeEx node, FlowState state, Configuration config) {
+  (
+    config.isSink(node.asNode()) and state instanceof FlowStateEmpty
+    or
+    config.isSink(node.asNode(), state)
+  ) and
+  not fullBarrier(node, config) and
+  not stateBarrier(node, state, config)
+}
 
 /** Provides the relevant barriers for a step from `node1` to `node2`. */
 pragma[inline]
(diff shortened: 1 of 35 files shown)


cpp/ql/lib/semmle/code/cpp/dataflow/internal/DataFlowImpl.qll
```diff
@@ -295,10 +327,12 @@ private predicate fullBarrier(NodeEx node, Configuration config) {
     config.isBarrier(n)
     or
     config.isBarrierIn(n) and
-    not config.isSource(n)
+    not config.isSource(n) and
+    not config.isSource(n, _)
```
atorralba, line 331: Isn't this redundant? If `config.isSource(n)` is equivalent to `config.isSource(n, "")`, it should mean that if `not config.isSource(n, _)` holds, `not config.isSource(n, "")` must also hold.

java/ql/lib/semmle/code/java/dataflow/internal/DataFlowImplCommon.qll
```diff
@@ -3,6 +3,25 @@ private import DataFlowImplSpecific::Public
 import Cached
 
 module DataFlowImplCommonPublic {
+  /** A state value to track during data flow. */
+  abstract class FlowState extends string {
```
hvitved, line 7: This is a potential foot gun: For data flow configurations in libraries like https://github.com/github/codeql/blob/c5728b2951560649b158b330e5051d8fe1215695/cpp/ql/lib/semmle/code/cpp/dataflow/internal/DataFlowUtil.qll#L737
if someone were to add new states to a query (which I assume would be a valid thing to do), then such library analyses would have to be recomputed.

Can't `FlowState` just be an alias for `string`?

java/ql/lib/semmle/code/java/dataflow/internal/DataFlowImpl.qll
```diff
@@ -82,6 +99,15 @@ abstract class Configuration extends string {
    */
   predicate isAdditionalFlowStep(Node node1, Node node2) { none() }
 
+  /**
+   * Holds if the additional flow step from `node1` to `node2` must be taken
+   * into account in the analysis. This step is only applicable in `state1` and
+   * updates the flow state to `state2`.
+   */
+  predicate isAdditionalFlowStep(Node node1, FlowState state1, Node node2, FlowState state2) {
```
hvitved, line 107: Won't we need something like this for cases where we previously chained together configurations?

```ql
/**
 * Holds if reaching `node1` updates the state from `state1` to `state2`.
 */
predicate updatesState(Node node, FlowState state1, FlowState state2)
```

hvitved (commented):
Initial (design) comments (I haven't looked at the code yet).

java/ql/lib/semmle/code/java/dataflow/internal/DataFlowImplCommon.qll
aschackmull: Good point. There might be one place in DataFlowImpl where this change will require a refactor.

java/ql/lib/semmle/code/java/dataflow/internal/DataFlowImpl.qll
aschackmull: You can simulate that fairly well by using `isAdditionalStep(node, state1, node, state2)`. So I guess we could add it as a convenience, but it shouldn't be strictly necessary.

hvitved: Except that will only work for taint steps and not value preserving steps.

java/ql/lib/semmle/code/java/dataflow/internal/DataFlowImplCommon.qll
intrigus-lgtm: Maybe this has an obvious answer, but isn't codeql-javascript having the same potential foot gun?
Or is the plan to (later) integrate this PR with the javascript data-flow library?
https://github.com/github/codeql/blob/15caaa7ad6d91c533596204b6a863a1c56240628/javascript/ql/lib/semmle/javascript/dataflow/Configuration.qll#L284-L293

hvitved: Better to let @github/codeql-javascript answer that.

aschackmull: I don't think the javascript data-flow library supports usage in library code - it only exists in one copy, so that should be reserved for queries. It's a good observation and a good question, though, and certainly something that needs to be taken into account if the javascript data-flow library is to allow in-library use.

java/ql/lib/semmle/code/java/dataflow/internal/DataFlowImpl.qll
aschackmull: I wasn't planning on allowing state-dependent or state-changing value steps. Do you think we'll have strong use-cases for that?

hvitved: Not sure, but I do know that we have previously chained together configurations to model the case where flow must go through a certain node. Whether the step into that node is alway a taint step I don't know, but it seems a bit restrictive to assume so.

java/ql/lib/semmle/code/java/dataflow/internal/DataFlowImplCommon.qll
erik-krogh: > I don't think the javascript data-flow library supports usage in library code - it only exists in one copy, so that should be reserved for queries. 

Yep. We only create new `FlowLabel`s in query specific code, and never in library code.   
We also have [a test that fails](https://github.com/github/codeql/blob/main/javascript/ql/test/library-tests/FlowLabels/DefaultFlowLabels.ql) if we every accidentally add more default flow labels.

hvitved: The problem is not whether you create flow labels in library code, but whether you use global data flow in library code.

erik-krogh: Ahh. We don't use global dataflow in library code, we use type-tracking instead.

asgerf: I'm supportive of the idea of using an alias `class FlowState = string` instead of the abstract class.

In practice you want to define a predicate for accessing the flow state constant. Once you have that predicate, you don't gain much by forcing users to declare a subclass as well. In JS this has led to various foot-gun situations with [awkward solutions](https://github.com/github/codeql/blob/main/javascript/ql/lib/semmle/javascript/security/dataflow/TaintedPathQuery.qll#L14).

ruby/ql/lib/codeql/ruby/dataflow/internal/DataFlowPrivate.qll
```diff
@@ -710,14 +710,16 @@ predicate clearsContent(Node n, Content c) {
   FlowSummaryImpl::Private::Steps::summaryClearsContent(n, c)
 }
 
-private newtype TDataFlowType = TTodoDataFlowType()
+private newtype TDataFlowType =
+  TTodoDataFlowType() or
+  TTodoDataFlowType2()
```
hvitved, line 715: This deserves a comment.  I would expect that this is also a problem for Python, where `DataFlowType` is also a singleton.

cpp/ql/lib/semmle/code/cpp/dataflow/internal/DataFlowImpl.qll
```diff
@@ -994,15 +1111,23 @@ private module Stage2 {
...
+      localFlowStepNodeCand1(node1, node2, config) and
+      state1 = state2
+      or
+      preservesValue = false and
+      additionalLocalFlowStepNodeCand1(node1, node2, config) and
+      state1 = state2
       or
       preservesValue = false and
-      additionalLocalFlowStepNodeCand1(node1, node2, config)
+      additionalLocalStateStep(node1, state1, node2, state2, config)
```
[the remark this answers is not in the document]
aschackmull: Where the `localFlowStep` and `additionalLocalFlowStep` relations tend to be quite large, and therefore benefits from this restriction, I don't think it matters for `additionalLocalStateStep`.

cpp/ql/lib/semmle/code/cpp/dataflow/internal/DataFlowImpl.qll
```diff
@@ -3639,11 +3987,12 @@ private predicate pathOutOfCallable(PathNodeMid mid, NodeEx out, CallContext cc)
...
 pragma[noinline]
 private predicate pathIntoArg(
-  PathNodeMid mid, ParameterPosition ppos, CallContext cc, DataFlowCall call, AccessPath ap,
-  AccessPathApprox apa, Configuration config
+  PathNodeMid mid, ParameterPosition ppos, FlowState state, CallContext cc, DataFlowCall call,
+  AccessPath ap, AccessPathApprox apa, Configuration config
 ) {
   exists(ArgNode arg, ArgumentPosition apos |
     arg = mid.getNodeEx().asNode() and
+    state = mid.getState() and
```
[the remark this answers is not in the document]
aschackmull: It's a bit shorter and ought to compile to the same thing, so sure.

ruby/ql/lib/codeql/ruby/dataflow/internal/DataFlowPrivate.qll
hvitved: Are you sure we shouldn't add a second constructor for Python as well?

cpp/ql/lib/semmle/code/cpp/dataflow/internal/DataFlowImpl.qll
```diff
@@ -994,15 +1111,23 @@ private module Stage2 {
   bindingset[node, cc, config]
   private LocalCc getLocalCc(NodeEx node, Cc cc, Configuration config) { any() }
 
+  bindingset[node1, state1, config]
+  bindingset[node2, state2, config]
   private predicate localStep(
```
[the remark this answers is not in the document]
aschackmull: I know. I was indeed thinking about readability in terms of those predicates - maybe I should have said maintainability instead. Reading a predicate in which another complex predicate is being inlined and reasoning about its expected behaviour is somewhat tricky, I think. Note also that even `localStep` is being duplicated in `revFlow` currently, so the disjuncts aren't all collapsed in that case either.
It is the fact that these 3 disjuncts have different binding patterns that makes them difficult to collapse in a simple way.

java/ql/lib/semmle/code/java/dataflow/internal/DataFlowImpl.qll
rdmarsh2: I think I've found a use case for state-changing value steps when there's a multi-part barrier (in this case, two different methods that need to be called to secure an XML parser against XXE vulnerabilities).

PR #7350, opened 2021-12-10
Dataflow: order step side-conditions ahead of mapping Node <-> NodeEx and cartesian product with Configuration

smowton:


cpp/ql/lib/semmle/code/cpp/dataflow/internal/DataFlowImpl.qll
```diff
@@ -351,7 +351,8 @@ private predicate additionalLocalFlowStep(NodeEx node1, NodeEx node2, Configurat
   exists(Node n1, Node n2 |
     node1.asNode() = n1 and
     node2.asNode() = n2 and
-    config.isAdditionalFlowStep(n1, n2) and
+    pragma[only_bind_into](config)
```
aschackmull, line 354: This pragma is superfluous, since nothing else binds `config` here.

cpp/ql/lib/semmle/code/cpp/dataflow/internal/DataFlowImpl.qll
```diff
@@ -384,16 +385,21 @@ private predicate additionalJumpStep(NodeEx node1, NodeEx node2, Configuration c
...
```
aschackmull, line 388: same

aschackmull:
Looks like a speed-up for java as well.

smowton:
Looks probably beneficial but lots of unexplained failures including one that ran for v1 then failed for v2? Asking on Slack why I can't retry those failures.

smowton:
@aschackmull given the results look positive, have marked ready for review and removed the superfluous bindings you mentioned.

MathiasVP:
Could you run DCA for the other languages as well, please?

smowton:
@MathiasVP done. C/C++ had a project and query suite named "DCA-PR" which seemed like the obvious thing to use; for other languages I've started Differences / lgtm.qls.

MathiasVP:
> @MathiasVP done. C/C++ had a project and query suite named "DCA-PR" which seemed like the obvious thing to use; for other languages I've started Differences / lgtm.qls.

That's not actually the one to use :) Please use `Differences.yml` for C++ as well. The DCA-PR is one that's used internally for DCA changes (we really should change that name).

smowton:
Doh ok, fixed, new run started

aschackmull:
The two C/C++ slowdowns look like something that needs to be looked into.

tausbn:
This PR seems to have gone a bit stale. Is it something that should be revived, or can it just be closed?

smowton:
I'll re-check whether the join order is still a problem.

calumgrant:
Do you have any updates on this?

aschackmull:
This has been superseded by https://github.com/github/codeql/pull/9686. Closing.

PR #7351, opened 2021-12-10
JS: Improve handling of heuristic sinks in endpoint filters

henrymercer:
Previously heuristic sinks were always included, to avoid us filtering them out due to not being an argument to an external library call. In this PR, we move the "argument to an external library call" filtering to the query-specific endpoint filters. This lets us filter out heuristic sinks if they match one of the other endpoint filters, reducing FPs.


Z80coder (approved):
I understand the intent of this refactoring, which is to apply per-query endpoint filters to heuristic sinks, and therefore reduce FPs. On the end-to-end metrics the improvement is 1 less FP. In the wild, this may be more. I can't comment on the precise details of the QL refactoring but otherwise LGTM.

PR #7352, opened 2021-12-10
ATM Endpoint filtering improvements

esbena:
Based on sheet `atm-results (2021-12-08)`: 

- 66 triaged results
- 45 addressed FP results: should disappear

- Primary approach: extend the endpoint filters directly
- Secondary approach: model more database accesses

The secondary approach requires some manual result inspection, especially for the missing rate-limiting query.

Core library changes:

- More database accesses recognized (redis, ioredis, marsdb)
- The `express-fileupload` `mv` function has a path-injection sink

### DIFF
--- a/javascript/ql/lib/semmle/javascript/frameworks/NoSQL.qll
+++ b/javascript/ql/lib/semmle/javascript/frameworks/NoSQL.qll
@@ -733,4 +733,37 @@ private module Redis {
       )
     }
   }
+
+  /**
+   * An access to a database through redis
+   */
+  class RedisDatabaseAccess extends DatabaseAccess {
+    RedisDatabaseAccess() { this = redis().getMember(_).getACall() }
+
+    override DataFlow::Node getAQueryArgument() { none() }
+  }
+}
+
+/**
+ * Provides classes modeling the `ioredis` library.
+ *
+ * ```
+ * import Redis from 'ioredis'
+ * let client = new Redis(...)
+ * ```
+ */
+private module IoRedis {
+  /**
+   * Gets an `ioredis` client.
+   */
+  API::Node ioredis() { result = API::moduleImport("ioredis").getInstance() }
+
+  /**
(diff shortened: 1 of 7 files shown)


javascript/ql/lib/semmle/javascript/frameworks/NoSQL.qll
```diff
@@ -618,14 +618,30 @@ private module Minimongo {
...
 private module MarsDB {
+  private class MarsDBAccess extends DatabaseAccess {
+    string method;
+
+    MarsDBAccess() {
+      this =
+        API::moduleImport("marsdb")
+            .getMember("Collection")
+            .getInstance()
+            .getMember(method)
```
asgerf, line 629: On these new `DatabaseAccess` subclasses (this one and the others), there is no restriction on the method name. Have you checked the API that this is reasonable? Maybe for the ATM use-case it doesn't matter, but there may be other uses where you don't want calls that aren't actually accessing the database.

For example, I'm not sure registering an event handler ought to be considered a database access.

asgerf: Btw this isn't a blocker if merging this PR would unblock others.

esbena: I did think along those lines, but I think all siblings to methods that could run queries are fair game in practice. For your example and the missing rate-limit query, I suppose it could be real problem if you add lots of event handlers to an active database connection.

The evaluation did not show any new alerts from this change.

esbena: So I'll merge as is, and volunteer to fix the FPs this heuristic causes.

esbena:
Had to format, I need a new approval.

Z80coder (approved):
Thanks @esbena and @asgerf!

PR #7054, opened 2021-11-04
Java: Promote Log Injection from experimental

atorralba:
PR to promote the Log Injection query created in #5099.

## Changes

* Existing files were moved out of experimental.
* Refactored into different libraries. CSV models were added to `Logging.qll`.
* Added tests, both for the query and for the new CSV summaries. 

## To Consider

* No sanitizers were added, for more information see the original discussion in https://github.com/github/codeql/pull/5099#issuecomment-795508576.
* `security-severity` was calculated automatically, but it seems too high.
* Given the high amount of results found by the query (and the difficulty of validating all of them), maybe it makes sense to lower the `precision` to `medium`.

PR #7354, opened 2021-12-10
Java: Experimental query for Log4j JNDI Injection

atorralba:
This is an experimental query for detecting CVE-2021-44228.

### DIFF
--- a/java/ql/src/experimental/Security/CWE/CWE-020/Log4jJndiInjection.qhelp
+++ b/java/ql/src/experimental/Security/CWE/CWE-020/Log4jJndiInjection.qhelp
@@ -1,0 +1,53 @@
+<!DOCTYPE qhelp PUBLIC
+"-//Semmle//qhelp//EN"
+"qhelp.dtd">
+<qhelp>
+
+<overview>
+<p>
+Log4j versions prior to 2.15.0 are subject to a remote code execution vulnerability via the ldap JNDI parser.
+</p>
+<p>
+As per Apache's Log4j security guide: Apache Log4j2 &lt;=2.14.1 JNDI features used in configuration, log messages, and parameters 
+do not protect against attacker controlled LDAP and other JNDI related endpoints. An attacker who can control log messages or 
+log message parameters can execute arbitrary code loaded from LDAP servers when message lookup substitution is enabled. 
+From log4j 2.15.0, this behavior has been disabled by default.
+</p>
+</overview>
+
+<recommendation>
+<p>
+This issue was remediated in Log4J v2.15.0. The Apache Logging Services team provides the following mitigation advice:
+</p>
+<p>
+In previous releases (>=2.10) this behavior can be mitigated by setting system property "log4j2.formatMsgNoLookups" to “true” 
+or by removing the JndiLookup class from the classpath (example: zip -q -d log4j-core-*.jar org/apache/logging/log4j/core/lookup/JndiLookup.class). 
+Java 8u121 protects against RCE by defaulting "com.sun.jndi.rmi.object.trustURLCodebase" and "com.sun.jndi.cosnaming.object.trustURLCodebase" to "false".
+</p>
+<p>
+You can manually check for use of affected versions of Log4J by searching your project repository for Log4J use, which is often in a pom.xml file.
+</p>
+<p>
+Where possible, upgrade to Log4J version 2.15.0. If you are using Log4J v1 there is a migration guide available.
+</p>
+<p>
+Please note that Log4J v1 is End Of Life (EOL) and will not receive patches for this issue. Log4J v1 is also vulnerable to other RCE vectors and we 
+recommend you migrate to Log4J 2.15.0 where possible.
+</p>
+<p>
+If upgrading is not possible, then ensure the -Dlog4j2.formatMsgNoLookups=true system property is set on both client- and server-side components.
+</p>
+</recommendation>
+
+<example>
+<p>In this example, a username, provided by the user, is logged using <code>logger.warn</code> (from  <code>org.apache.logging.log4j.Logger</code>). 
+  If a malicious user provides <code>${jndi:ldap://127.0.0.1:1389/a}</code> as a username parameter, 
+  Log4j will make a JNDI lookup on the specified LDAP server and potentially load arbitrary code. 
+</p>
+<sample src="Log4jJndiInjection.java" />
+</example>
+
+<references>
+<li>GitHub Advisory Database: <a href="https://github.com/advisories/GHSA-jfh8-c2jp-5v3q">Remote code injection in Log4j</a>.</li>
+</references>
+</qhelp>
--- a/java/ql/src/experimental/Security/CWE/CWE-020/Log4jJndiInjection.qhelp
+++ b/java/ql/src/experimental/Security/CWE/CWE-020/Log4jJndiInjection.qhelp
@@ -5,19 +5,19 @@
 
 <overview>
 <p>
-Log4j versions prior to 2.15.0 are subject to a remote code execution vulnerability via the ldap JNDI parser.
+This query flags up situations in which untrusted user data is included in Log4j messages. If an application uses a Log4j version prior to 2.15.0, using untrusted user data in log messages will make an application vulnerable to remote code execution through Log4j's LDAP JNDI parser (CVE-2021-44228).
 </p>
 <p>
 As per Apache's Log4j security guide: Apache Log4j2 &lt;=2.14.1 JNDI features used in configuration, log messages, and parameters 
 do not protect against attacker controlled LDAP and other JNDI related endpoints. An attacker who can control log messages or 
 log message parameters can execute arbitrary code loaded from LDAP servers when message lookup substitution is enabled. 
-From log4j 2.15.0, this behavior has been disabled by default.
+From Log4j 2.15.0, this behavior has been disabled by default. Note that this query will not try to determine which version of Log4j is used.
 </p>
 </overview>
 
 <recommendation>
 <p>
-This issue was remediated in Log4J v2.15.0. The Apache Logging Services team provides the following mitigation advice:
+This issue was remediated in Log4j v2.15.0. The Apache Logging Services team provides the following mitigation advice:
 </p>
 <p>
 In previous releases (>=2.10) this behavior can be mitigated by setting system property "log4j2.formatMsgNoLookups" to “true” 
@@ -25,14 +25,14 @@ or by removing the JndiLookup class from the classpath (example: zip -q -d log4j
 Java 8u121 protects against RCE by defaulting "com.sun.jndi.rmi.object.trustURLCodebase" and "com.sun.jndi.cosnaming.object.trustURLCodebase" to "false".
 </p>
 <p>
-You can manually check for use of affected versions of Log4J by searching your project repository for Log4J use, which is often in a pom.xml file.
+You can manually check for use of affected versions of Log4j by searching your project repository for Log4j use, which is often in a pom.xml file.
 </p>
 <p>
 Where possible, upgrade to Log4J version 2.15.0. If you are using Log4J v1 there is a migration guide available.
 </p>
 <p>
-Please note that Log4J v1 is End Of Life (EOL) and will not receive patches for this issue. Log4J v1 is also vulnerable to other RCE vectors and we 
-recommend you migrate to Log4J 2.15.0 where possible.
+Please note that Log4j v1 is End Of Life (EOL) and will not receive patches for this issue. Log4j v1 is also vulnerable to other RCE vectors and we
+recommend you migrate to Log4j 2.15.0 where possible.
 </p>
 <p>
 If upgrading is not possible, then ensure the -Dlog4j2.formatMsgNoLookups=true system property is set on both client- and server-side components.
--- a/java/ql/src/experimental/Security/CWE/CWE-020/Log4jJndiInjection.qhelp
+++ b/java/ql/src/experimental/Security/CWE/CWE-020/Log4jJndiInjection.qhelp
@@ -20,15 +20,14 @@ From Log4j 2.15.0, this behavior has been disabled by default. Note that this qu
 This issue was remediated in Log4j v2.15.0. The Apache Logging Services team provides the following mitigation advice:
 </p>
 <p>
-In previous releases (>=2.10) this behavior can be mitigated by setting system property "log4j2.formatMsgNoLookups" to “true” 
-or by removing the JndiLookup class from the classpath (example: zip -q -d log4j-core-*.jar org/apache/logging/log4j/core/lookup/JndiLookup.class). 
-Java 8u121 protects against RCE by defaulting "com.sun.jndi.rmi.object.trustURLCodebase" and "com.sun.jndi.cosnaming.object.trustURLCodebase" to "false".
+In previous releases (>=2.10) this behavior can be mitigated by setting system property <code>log4j2.formatMsgNoLookups</code> to <code>true</code> 
+or by removing the <code>JndiLookup</code> class from the classpath (example: <code>zip -q -d log4j-core-*.jar org/apache/logging/log4j/core/lookup/JndiLookup.class</code>). 
 </p>
 <p>
 You can manually check for use of affected versions of Log4j by searching your project repository for Log4j use, which is often in a pom.xml file.
 </p>
 <p>
-Where possible, upgrade to Log4J version 2.15.0. If you are using Log4J v1 there is a migration guide available.
+Where possible, upgrade to Log4j version 2.15.0. If you are using Log4j v1 there is a migration guide available.
 </p>
 <p>
 Please note that Log4j v1 is End Of Life (EOL) and will not receive patches for this issue. Log4j v1 is also vulnerable to other RCE vectors and we
--- a/java/ql/src/experimental/Security/CWE/CWE-020/Log4jJndiInjection.ql
+++ b/java/ql/src/experimental/Security/CWE/CWE-020/Log4jJndiInjection.ql
@@ -1,11 +1,11 @@
 /**
- * @name Log4j JNDI Injection
+ * @name Log4j log injection and LDAP JNDI injection
  * @description Building Log4j log entries from user-controlled data may allow
  *              attackers to inject malicious code through JNDI lookups.
  * @kind path-problem
  * @problem.severity error
  * @precision high
- * @id java/log4j-jndi-injection
+ * @id java/log4j-injection
  * @tags security
  *       external/cwe/cwe-020
  *       external/cwe/cwe-074
@@ -168,7 +168,7 @@ private class LoggingSinkModels extends SinkModelCsv {
         "org.apache.logging.log4j;LogBuilder;true;log;(String,Object,Object,Object,Object,Object,Object,Object,Object);;Argument[0..8];logging",
         "org.apache.logging.log4j;LogBuilder;true;log;(String,Object,Object,Object,Object,Object,Object,Object,Object,Object);;Argument[0..9];logging",
         "org.apache.logging.log4j;LogBuilder;true;log;(String,Object,Object,Object,Object,Object,Object,Object,Object,Object,Object);;Argument[0..10];logging",
-        "org.apache.logging.log4j;LogBuilder;true;log;(String,Supplier);;Argument[0..1];logging",
+        "org.apache.logging.log4j;LogBuilder;true;log;(String,Supplier[]);;Argument[0..1];logging",
         "org.apache.logging.log4j;LogBuilder;true;log;(Supplier);;Argument[0];logging"
       ]
   }
--- a/java/ql/src/experimental/Security/CWE/CWE-020/Log4jJndiInjection.ql
+++ b/java/ql/src/experimental/Security/CWE/CWE-020/Log4jJndiInjection.ql
@@ -18,21 +18,6 @@ import semmle.code.java.dataflow.FlowSources
 import semmle.code.java.dataflow.ExternalFlow
 import DataFlow::PathGraph
 
-private class LoggingSummaryModels extends SummaryModelCsv {
-  override predicate row(string row) {
-    row =
-      [
-        "org.apache.logging.log4j;Logger;true;traceEntry;(Message);;Argument[0];ReturnValue;taint",
-        "org.apache.logging.log4j;Logger;true;traceEntry;(String,Object[]);;Argument[0..1];ReturnValue;taint",
-        "org.apache.logging.log4j;Logger;true;traceEntry;(String,Supplier[]);;Argument[0..1];ReturnValue;taint",
-        "org.apache.logging.log4j;Logger;true;traceEntry;(Supplier[]);;Argument[0];ReturnValue;taint",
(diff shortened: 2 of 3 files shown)


fproulx:
Good job @atorralba was looking for that throughout the day =)

java/ql/src/experimental/Security/CWE/CWE-020/Log4jJndiInjection.qhelp
```diff
@@ -0,0 +1,53 @@
...
+</overview>
+
+<recommendation>
+<p>
+This issue was remediated in Log4J v2.15.0. The Apache Logging Services team provides the following mitigation advice:
+</p>
+<p>
+In previous releases (>=2.10) this behavior can be mitigated by setting system property "log4j2.formatMsgNoLookups" to “true” 
+or by removing the JndiLookup class from the classpath (example: zip -q -d log4j-core-*.jar org/apache/logging/log4j/core/lookup/JndiLookup.class). 
+Java 8u121 protects against RCE by defaulting "com.sun.jndi.rmi.object.trustURLCodebase" and "com.sun.jndi.cosnaming.object.trustURLCodebase" to "false".
```
Marcono1234, line 25: That part is rather misleading / incorrect and should probably be removed (even though it was originally in the official security notes as well), see also https://github.com/apache/logging-log4j2/pull/608#issuecomment-991396409.

In the documentation here there are also some inconsistent capitalizations of "Log4j" (with uppercase 'J'), and it might be good to use `<code>...</code>` for the class name and `zip` command listed above. However, I did not want to add separate review comments to avoid hindering this important query.

java/ql/src/experimental/Security/CWE/CWE-020/Log4jJndiInjection.ql
```diff
@@ -0,0 +1,207 @@
...
+            ";(Level,String,Object,Object,Object,Object,Object,Object,Object,Object,Object,Object);;Argument[1..11];logging",
+            ";(Level,String,Supplier);;Argument[1..2];logging",
+            ";(Level,String,Throwable);;Argument[1];logging",
+            ";(Level,Supplier);;Argument[1];logging",
+            ";(Level,Supplier,Throwable);;Argument[1];logging"
+          ], "org.apache.logging.log4j;Logger;true;entry;(Object[]);;Argument[0];logging",
+        "org.apache.logging.log4j;Logger;true;logMessage;(Level,Marker,String,StackTraceElement,Message,Throwable);;Argument[4];logging",
+        "org.apache.logging.log4j;Logger;true;printf;(Level,Marker,String,Object[]);;Argument[2..3];logging",
+        "org.apache.logging.log4j;Logger;true;printf;(Level,String,Object[]);;Argument[1..2];logging",
+        "org.apache.logging.log4j;Logger;true;traceEntry;(Message);;Argument[0];logging",
```
Marcono1234, line 146: Should this consider the two [`throwing`](https://logging.apache.org/log4j/2.x/log4j-api/apidocs/org/apache/logging/log4j/Logger.html#throwing-org.apache.logging.log4j.Level-T-) methods as well? Though exception messages might not actually be affected by this vulnerability. And it looks like for the other methods you explicitly did not consider the exceptions either.

java/ql/src/experimental/Security/CWE/CWE-020/Log4jJndiInjection.ql
```diff
@@ -0,0 +1,207 @@
...
+        "org.apache.logging.log4j;LogBuilder;true;log;(String,Object,Object,Object);;Argument[0..3];logging",
+        "org.apache.logging.log4j;LogBuilder;true;log;(String,Object,Object,Object,Object);;Argument[0..4];logging",
+        "org.apache.logging.log4j;LogBuilder;true;log;(String,Object,Object,Object,Object,Object);;Argument[0..5];logging",
+        "org.apache.logging.log4j;LogBuilder;true;log;(String,Object,Object,Object,Object,Object,Object);;Argument[0..6];logging",
+        "org.apache.logging.log4j;LogBuilder;true;log;(String,Object,Object,Object,Object,Object,Object,Object);;Argument[0..7];logging",
+        "org.apache.logging.log4j;LogBuilder;true;log;(String,Object,Object,Object,Object,Object,Object,Object,Object);;Argument[0..8];logging",
+        "org.apache.logging.log4j;LogBuilder;true;log;(String,Object,Object,Object,Object,Object,Object,Object,Object,Object);;Argument[0..9];logging",
+        "org.apache.logging.log4j;LogBuilder;true;log;(String,Object,Object,Object,Object,Object,Object,Object,Object,Object,Object);;Argument[0..10];logging",
+        "org.apache.logging.log4j;LogBuilder;true;log;(String,Supplier);;Argument[0..1];logging",
```
Marcono1234, line 171:
```suggestion
        "org.apache.logging.log4j;LogBuilder;true;log;(String,Supplier[]);;Argument[0..1];logging",
```
The `Supplier` parameter is varargs, see [doc](https://logging.apache.org/log4j/2.x/log4j-api/apidocs/org/apache/logging/log4j/LogBuilder.html#log-java.lang.String-org.apache.logging.log4j.util.Supplier...-)

Marcono1234:
Is this based on / related to #7054? Would it make sense to use a kind other than `logging` here to avoid clashes with that other pull request, and because these findings here have a higher impact?

java/ql/src/experimental/Security/CWE/CWE-020/Log4jJndiInjection.qhelp
```diff
@@ -0,0 +1,53 @@
...
+
+<overview>
+<p>
+Log4j versions prior to 2.15.0 are subject to a remote code execution vulnerability via the ldap JNDI parser.
+</p>
+<p>
+As per Apache's Log4j security guide: Apache Log4j2 &lt;=2.14.1 JNDI features used in configuration, log messages, and parameters 
+do not protect against attacker controlled LDAP and other JNDI related endpoints. An attacker who can control log messages or 
+log message parameters can execute arbitrary code loaded from LDAP servers when message lookup substitution is enabled. 
+From log4j 2.15.0, this behavior has been disabled by default.
```
sj, line 14:
```suggestion
From Log4j 2.15.0, this behavior has been disabled by default. Note that this query will not try to determine which version of Log4j is used.
```

securingdev:
Have we considered looking at [this open source query](https://github.com/cldrn/codeql-queries/blob/master/log4j-injection.ql) as an alternative way of finding this vulnerability in code?

java/ql/src/experimental/Security/CWE/CWE-020/Log4jJndiInjection.qhelp
```diff
@@ -0,0 +1,53 @@
...
+</p>
+</overview>
+
+<recommendation>
+<p>
+This issue was remediated in Log4J v2.15.0. The Apache Logging Services team provides the following mitigation advice:
```
sj, line 20:
```suggestion
This issue was remediated in Log4j v2.15.0. The Apache Logging Services team provides the following mitigation advice:
```

sj: Let's follow Apache's [own guidelines](https://logging.apache.org/log4j/2.x/) on formatting "Log4j": uppercase "L" and lowercase "j".

java/ql/src/experimental/Security/CWE/CWE-020/Log4jJndiInjection.qhelp
```diff
@@ -0,0 +1,53 @@
...
+</p>
+<p>
+In previous releases (>=2.10) this behavior can be mitigated by setting system property "log4j2.formatMsgNoLookups" to “true” 
+or by removing the JndiLookup class from the classpath (example: zip -q -d log4j-core-*.jar org/apache/logging/log4j/core/lookup/JndiLookup.class). 
+Java 8u121 protects against RCE by defaulting "com.sun.jndi.rmi.object.trustURLCodebase" and "com.sun.jndi.cosnaming.object.trustURLCodebase" to "false".
+</p>
+<p>
+You can manually check for use of affected versions of Log4J by searching your project repository for Log4J use, which is often in a pom.xml file.
```
sj, line 28:
```suggestion
You can manually check for use of affected versions of Log4j by searching your project repository for Log4j use, which is often in a pom.xml file.
```

java/ql/src/experimental/Security/CWE/CWE-020/Log4jJndiInjection.qhelp
```diff
@@ -0,0 +1,53 @@
...
+</p>
+<p>
+Where possible, upgrade to Log4J version 2.15.0. If you are using Log4J v1 there is a migration guide available.
```
sj, line 31:
```suggestion
Where possible, upgrade to Log4j version 2.15.0. If you are using Log4j v1 there is a migration guide available.
```

java/ql/src/experimental/Security/CWE/CWE-020/Log4jJndiInjection.qhelp
```diff
@@ -0,0 +1,53 @@
...
+</p>
+<p>
+Please note that Log4J v1 is End Of Life (EOL) and will not receive patches for this issue. Log4J v1 is also vulnerable to other RCE vectors and we
```
sj, line 34:
```suggestion
Please note that Log4j v1 is End Of Life (EOL) and will not receive patches for this issue. Log4j v1 is also vulnerable to other RCE vectors and we
```

java/ql/src/experimental/Security/CWE/CWE-020/Log4jJndiInjection.qhelp
```diff
@@ -0,0 +1,53 @@
...
+</p>
+<p>
+You can manually check for use of affected versions of Log4J by searching your project repository for Log4J use, which is often in a pom.xml file.
+</p>
+<p>
+Where possible, upgrade to Log4J version 2.15.0. If you are using Log4J v1 there is a migration guide available.
+</p>
+<p>
+Please note that Log4J v1 is End Of Life (EOL) and will not receive patches for this issue. Log4J v1 is also vulnerable to other RCE vectors and we 
+recommend you migrate to Log4J 2.15.0 where possible.
```
sj, line 35:
```suggestion
recommend you migrate to Log4j 2.15.0 where possible.
```

java/ql/src/experimental/Security/CWE/CWE-020/Log4jJndiInjection.ql
```diff
@@ -0,0 +1,207 @@
+/**
+ * @name Log4j JNDI Injection
```
sj, line 2:
```suggestion
 * @name Log4j log injection and LDAP JNDI injection
```

sj (commented):
Few suggestions for improvements to the documentation

smowton (approved):

java/ql/src/experimental/Security/CWE/CWE-020/Log4jJndiInjection.ql
```diff
@@ -0,0 +1,207 @@
...
+ * @description Building Log4j log entries from user-controlled data may allow
+ *              attackers to inject malicious code through JNDI lookups.
+ * @kind path-problem
+ * @problem.severity error
+ * @precision high
+ * @id java/log4j-jndi-injection
```
sj, line 8:
```suggestion
 * @id java/log4j-injection
```

java/ql/src/experimental/Security/CWE/CWE-020/Log4jJndiInjection.ql
atorralba: I tested this, and `throwing` and similar methods don't seem to make the JNDI connection in the vulnerable version, so no need to model those. Thanks for pointing it out nevertheless!

java/ql/src/experimental/Security/CWE/CWE-020/Log4jJndiInjection.qhelp
atorralba: Good catch, thanks! That's been removed from the GHSA too, so it's a good idea to not have it here either.
```suggestion
```

java/ql/src/experimental/Security/CWE/CWE-020/Log4jJndiInjection.qhelp
```diff
@@ -0,0 +1,53 @@
...
+</p>
+</overview>
+
+<recommendation>
+<p>
+This issue was remediated in Log4j v2.15.0. The Apache Logging Services team provides the following mitigation advice:
+</p>
+<p>
+In previous releases (>=2.10) this behavior can be mitigated by setting system property "log4j2.formatMsgNoLookups" to “true” 
+or by removing the JndiLookup class from the classpath (example: zip -q -d log4j-core-*.jar org/apache/logging/log4j/core/lookup/JndiLookup.class).
```
atorralba, line 24: As suggested by @Marcono1234:

```suggestion
In previous releases (>=2.10) this behavior can be mitigated by setting system property <code>log4j2.formatMsgNoLookups</code> to <code>true</code> 
or by removing the <code>JndiLookup</code> class from the classpath (example: <code>zip -q -d log4j-core-*.jar org/apache/logging/log4j/core/lookup/JndiLookup.class</code>). 
```

atorralba:
Thank you so much for your review @Marcono1234 and @sj!

> Is this based on / related to #7054? Would it make sense to use a kind other than `logging` here to avoid clashes with that other pull request, and because these findings here have a higher impact?

This query is intended to be short-lived and remain in experimental. The intention is to provide a tool for users to check for potential exploitation paths in their code if they need it, but not to run it continuously as part of code scanning. As such, we don't intend to merge the Log Injection query for the moment, which should avoid potential clashes. Of course, if this was going to be merged into the main query set, your suggestion would make total sense, so thank you!

> Though I assume this query here is mostly only useful for security researchers to find attack vectors (which can still be valuable); it should not be used to reason about whether an application is affected by the Log4j 2 vulnerability because:
> 
> * Dependencies might use Log4j 2; this will not be detected by this query
> * This query likely has false negatives
> * There are bridging bindings (e.g. "Log4j 2 SLF4J Binding") which redirect log messages of other logging frameworks to Log4j 2

As stated above, this query is intended to be a tool to help users evaluating their code, but of course it's not a silver bullet. For instance, Dependabot is actually more useful for detecting the issue, because it'll report if the vulnerable version is used. But for those that aren't able to activate Dependabot (or additionally to it), this query should provide a good alternative to obtain some information.

Regarding bindings, while you're right that they _could_ be also sinks, we've discussed it internally and, for the moment, we are going to try to maintain a high TP ratio in this query. If we added this without knowing whether the binding is actually bridging to Log4j, we'd be raising FP alerts.

atorralba:
> Have we considered looking at [this open source query](https://github.com/cldrn/codeql-queries/blob/master/log4j-injection.ql) as an alternative way of finding this vulnerability in code?

Yes, thanks @securingdev! There are several community-provided queries that do similar things out there. In this case, we decide to keep the explicit CSV format for defining sources, because we think it's more comprehensive. Is there anything that you think our query is lacking compared to those?

securingdev:
I would recommend we look at merging in the following section of code (or something similar) to catch any edge cases we might have missed:

``` 
exists(RefType t, Method m |
      t.hasQualifiedName("org.apache.log4j", ["Category", "Logger", "LogBuilder"]) // Log4j v1
      or
      t.hasQualifiedName("org.apache.logging.log4j", ["Logger", "LogBuilder", "LoggerManager"]) // Log4j v2 or
      or
      t.hasQualifiedName("org.apache.logging.log4j.core", ["Logger", "LogBuilder", "LoggerManager"]) // Log4j v2
      or
      t.hasQualifiedName("org.apache.logging.log4j.status", "StatusLogger") // Log4j Status logger
      or
      t.hasQualifiedName("org.slf4j", ["Logger", "LoggingEventBuilder"]) and // SLF4J Logger is used when Log4j core is on classpath
      log4JJarCoreJarFilePresent()
    |
      (
        m.getDeclaringType().getASourceSupertype*() = t or
        m.getDeclaringType().extendsOrImplements*(t)
      ) and
      m.getReturnType() instanceof VoidType and
      this = m.getAReference()
    )
  }
  ```

The following idea was also worth considering in some capacity:

```
predicate log4JJCoreJarFile(JarFile file) { file.getBaseName().matches("%log4j-core%") }

predicate log4JJarCoreJarFilePresent() { log4JJCoreJarFile(_) }
```

**Comment by atorralba** [comment]
The problem with that approach is that it would probably introduce some FPs like, as stated above, [`catching`](https://logging.apache.org/log4j/2.x/log4j-api/apidocs/org/apache/logging/log4j/Logger.html#catching-org.apache.logging.log4j.Level-java.lang.Throwable-) and [`throwing`](https://logging.apache.org/log4j/2.x/log4j-api/apidocs/org/apache/logging/log4j/Logger.html#throwing-org.apache.logging.log4j.Level-T-). Given how sensitive this query is security-wise, I would prefer not to raise false issues and keep a high TP ratio.

Happy to change it, though, if several of us think the risk of raising some FP is acceptable in exchange of having less FNs.

File: java/ql/src/experimental/Security/CWE/CWE-020/Log4jJndiInjection.ql [hunk 13]
```diff
@@ -0,0 +1,207 @@
...
+        "org.apache.logging.log4j;Logger;true;printf;(Level,String,Object[]);;Argument[1..2];logging",
+        "org.apache.logging.log4j;Logger;true;traceEntry;(Message);;Argument[0];logging",
+        "org.apache.logging.log4j;Logger;true;traceEntry;(String,Object[]);;Argument[0..1];logging",
+        "org.apache.logging.log4j;Logger;true;traceEntry;(String,Supplier[]);;Argument[0..1];logging",
+        "org.apache.logging.log4j;Logger;true;traceEntry;(Supplier[]);;Argument[0];logging",
+        "org.apache.logging.log4j;Logger;true;traceExit;(EntryMessage);;Argument[0];logging",
+        "org.apache.logging.log4j;Logger;true;traceExit;(EntryMessage,Object);;Argument[0..1];logging",
+        "org.apache.logging.log4j;Logger;true;traceExit;(Message,Object);;Argument[0..1];logging",
+        "org.apache.logging.log4j;Logger;true;traceExit;(Object);;Argument[0];logging",
+        "org.apache.logging.log4j;Logger;true;traceExit;(String,Object);;Argument[0..1];logging",
```
atorralba, line 154:
```suggestion
```

java/ql/src/experimental/Security/CWE/CWE-020/Log4jJndiInjection.ql
```diff
@@ -0,0 +1,207 @@
...
+        "org.apache.logging.log4j;Logger;true;traceEntry;(String,Object[]);;Argument[0..1];ReturnValue;taint",
+        "org.apache.logging.log4j;Logger;true;traceEntry;(String,Supplier[]);;Argument[0..1];ReturnValue;taint",
+        "org.apache.logging.log4j;Logger;true;traceEntry;(Supplier[]);;Argument[0];ReturnValue;taint",
+        "org.apache.logging.log4j;Logger;true;traceExit;(EntryMessage,Object);;Argument[1];ReturnValue;value",
+        "org.apache.logging.log4j;Logger;true;traceExit;(Message,Object);;Argument[1];ReturnValue;value",
+        "org.apache.logging.log4j;Logger;true;traceExit;(Object);;Argument[0];ReturnValue;value",
+        "org.apache.logging.log4j;Logger;true;traceExit;(String,Object);;Argument[1];ReturnValue;value",
+      ]
+  }
+}
```
atorralba, line 35:
```suggestion
```

smowton (approved):

ambiso:
Do I understand correctly - these queries are to detect whether a project is vulnerable to specifically CVE-2021-44228?

Are there queries that should/would have detected the vulnerability inside of log4j2?

sj:
> Do I understand correctly - these queries are to detect whether a project is vulnerable to specifically CVE-2021-44228?

That's correct. And it will detect other cases of Log4j log injection (which is something you [might generally want to avoid](https://owasp.org/www-community/attacks/Log_Injection)).

> Are there queries that should/would have detected the vulnerability inside of log4j2?

That's under active investigation. If you have any suggestions on how to approach this with CodeQL we'd love to hear from you!

securingdev:
Does this also catch cases of [CVE-2021-45046](https://nvd.nist.gov/vuln/detail/CVE-2021-45046) related to **Log4J**?

atorralba:
> Will this query also catch cases of [CVE-2021-45046](https://nvd.nist.gov/vuln/detail/CVE-2021-45046) related to **Log4J**?

Yes, the vulnerability pattern remains the same (untrusted data written to logs), it's just that the patch was incomplete for certain non-default configurations.

Marcono1234:
> > Do I understand correctly - these queries are to detect whether a project is vulnerable to specifically CVE-2021-44228?
> 
> That's correct.

@ambiso, might be good to clarify here, as [mentioned by atorralba](https://github.com/github/codeql/pull/7354#issuecomment-992771662) that this is no "silver bullet". If this query does not have any findings, you should not interpret that as not being vulnerable. If you have `log4j-core` as (transitive) dependency, you should definitely use one of the mitigation methods listed on the [Log4j 2 security notes](https://logging.apache.org/log4j/2.x/security.html). They have also released a new version targeting Java 7, in case that was an issue for you.

---

> Yes, the vulnerability pattern remains the same (untrusted data written to logs), it's just that the patch was incomplete for certain non-default configurations.

CVE-2021-45046 seems to be caused by legitimate usage of lookups in the log config (e.g. `log4j2.xml`) whose substituted value is attacker controlled (and resolved recursively). This blog post explains it quite well: https://www.lunasec.io/docs/blog/log4j-zero-day-update-on-cve-2021-45046/

That vulnerability is normally not triggered directly by log message injection (unless someone uses `$${event:Message}` in their log config, which is rather unlikely), but instead by other values being stored in some way, e.g. in the Log4j 2 `ThreadContext` or possibly implicitly in the `ServletContext`, see the lookup types here: https://logging.apache.org/log4j/2.x/manual/lookups.html
Therefore to me it looks like this query does not cover CVE-2021-45046.

atorralba:
> (unless someone uses `$${event:Message}` in their log config, which is rather unlikely)

That's what I was referring to with non-default configurations, in the sense that the patch mentioned an "incomplete fix" of the original vulnerability.

> instead by other values being stored in some way, e.g. in the Log4j 2 `ThreadContext` or possibly implicitly in the `ServletContext`

But you're right, there are more sinks that could be modeled to cover CVE-2021-45046 (e.g. `ThreadContext.put`). The urgency of doing that, however, is up to debate because of the somewhat limited impact of this in `>2.15.0` (DDoS), although it seems to be another RCE vector in `<=2.14.1`.

Thanks for the clarification @Marcono1234.

atorralba:
Also, there seems to be a [third vulnerability](https://www.praetorian.com/blog/log4j-2-15-0-stills-allows-for-exfiltration-of-sensitive-data/) affecting `2.15.0` being disclosed soon. Depending on how that works, maybe it will make sense to extend this query to both completely cover CVE-2021-45046 and this new variant.

sirdarckcat:
is anyone working on that atm? @atorralba

atorralba:
#7423 aims to cover CVE-2021-45046 completely.

PR #7166, opened 2021-11-18
Move upgrades into standard library packs

dbartol:
This PR moves the upgrade scripts for C#, C++, Java, JavaScript, and Python into the corresponding standard library packs (the same pack that contains the dbscheme).

Having upgrades in a separate pack gives us little or no advantage, and increases the overhead of package management. We wind up with the standard library (say, `codeql/cpp-all`) depending on another library pack (`codeql/cpp-upgrades`), but the version number of that dependency always needs to be kept in sync with the version number of the upgrade pack. In addition, no other pack for that language would have any reason to declare a dependency on the upgrade pack, so no query pack would ever have a reason to depend on a version of the upgrade pack other than the one that was published at the same time as that version of the library pack. By simply merging the packs we ensure that the right upgrade scripts are always available without having to keep two package versions in sync.

The actual changes were as follows:
- Moved contents of `<lang>/upgrades` to `<lang>/ql/lib/upgrades` verbatim.
- Added `upgrades: upgrades` property to `qlpack.yml` for each standard library pack.
- Deleted `qlpack*.yml` for upgrade packs.
- Removed from each standard library pack the declared dependency on the now-deleted upgrade pack.

The corresponding internal PR fixes up the build system to handle the change.

Note that Ruby already put its upgrades in its standard library.

I have not yet applied the same change to Go. I'll do that in a follow-up PR to avoid having three PRs that need to be merged simultaneously.

PR #7355, opened 2021-12-10
Move upgrades into standard library packs

aeisenberg:
Continuation of https://github.com/github/codeql/pull/7166. This time it is not in a fork.

This PR moves the upgrade scripts for C#, C++, Java, JavaScript, and Python into the corresponding standard library packs (the same pack that contains the dbscheme).

Having upgrades in a separate pack gives us little or no advantage, and increases the overhead of package management. We wind up with the standard library (say, codeql/cpp-all) depending on another library pack (codeql/cpp-upgrades), but the version number of that dependency always needs to be kept in sync with the version number of the upgrade pack. In addition, no other pack for that language would have any reason to declare a dependency on the upgrade pack, so no query pack would ever have a reason to depend on a version of the upgrade pack other than the one that was published at the same time as that version of the library pack. By simply merging the packs we ensure that the right upgrade scripts are always available without having to keep two package versions in sync.

The actual changes were as follows:

- Moved contents of `<lang>/upgrades` to `<lang>/ql/lib/upgrades` verbatim.
- Added `upgrades: upgrades` property to qlpack.yml for each standard library pack.
- Deleted `qlpack*.yml` for upgrade packs.
- Removed from each standard library pack the declared dependency on the now-deleted upgrade pack.

The corresponding internal PR fixes up the build system to handle the change.

Note that Ruby already put its upgrades in its standard library.


aeisenberg:
Failing C# tests. Should be working when run against the [PR](https://github.com/github/codeql/pull/7166). Here is an attempt at running it:

https://github.com/github/semmle-code/actions/runs/1565148791

dbartol (approved):
LGTM

PR #7356, opened 2021-12-10
Python: Basic support for match statement

yoff:
QL part of https://github.com/github/semmle-code/pull/41110.

- AST nodes for patterns
  - sub expressions, sub statements, and sub patterns
  - add sub patterns to the new `case` statement
  - scopes for patterns computed as the scope of the enclosing case (could be added to `py_scopes` instead)
- SSA for variables created inside match statements
  - `PatternCaptureDefinition` and `PatternAliasDefinition`
- data flow through patterns
  - subject of match flows to each top-level pattern (could be more like use-use-flow and only flow directly to the first one)
  - shape based flow inside patterns
    - [x] MatchAsPattern
    - [x] MatchOrPattern
    - [x] MatchLiteralPattern (no flow)
    - [x] MatchCapturePattern
    - [x] MatchWildcardPattern (no flow)
    - [x] MatchValuePattern (no flow)
    - [x] MatchSequencePattern
       - [x] when tuple
       - [x] when list
    - [x] MatchStarPattern
    - [x] MatchMappingPattern
    - [x] MatchDoubleStarPattern
    - [x] MatchKeyValuePattern
    - [x] MatchClassPattern
       - [x] for keyword arguments
       - [ ] for positional arguments
    - [x] MatchKeywordPattern
- [x] create upgrade script
- [x] create downgrade script


yoff:
Superseeded by https://github.com/github/codeql/pull/7635

PR #7368, opened 2021-12-13
Update creating-codeql-databases.rst

hvitved:
Always use `/p:UseSharedCompilation=false` for `msbuild` / `dotnet build`.

### DIFF
--- a/docs/codeql/codeql-cli/creating-codeql-databases.rst
+++ b/docs/codeql/codeql-cli/creating-codeql-databases.rst
@@ -210,6 +210,10 @@ commands that you can specify for compiled languages.
 
 - C# project built using ``dotnet build``::
 
+     For C# projects using either `dotnet build` or `msbuild`, you should specify `/p:UseSharedCompilation=false`
+     in the build command. It is also a good idea to add `/t:rebuild` to ensure that all code will be built (code
+     that is not built will not be included in the CodeQL database):
+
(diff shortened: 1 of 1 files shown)


docs/codeql/codeql-cli/creating-codeql-databases.rst
```diff
@@ -208,14 +208,9 @@ commands that you can specify for compiled languages.
 
      codeql database create cpp-database --language=cpp --command=make
 
-- C# project built using ``dotnet build`` (.NET Core 3.0 or later)::
+- C# project built using ``dotnet build``::
```
jf205, line 211: If the extra options `p:UseSharedCompilation=false /t:rebuild` are always required for projects built using dotnet, then I guess you might want to call that out in the text as well as the example command.

Also which version of the CLI was this changed introduced? If earlier versions of the CLI require different behaviour then that should be mentioned here too.

hvitved: Added. This was not a change that got introduced, but a bug that has been there all along. We are working on a long term solution that will inject the `/p:UseSharedCompilation=false` flag, but until we have that it has to be added manually.

jf205 (approved):

docs/codeql/codeql-cli/creating-codeql-databases.rst
jf205: I see. Thanks for the explanation.

PR #7371, opened 2021-12-13
Ruby: use Ruby object instantiation syntax in a comment

alexrford:
This was using ES6 style class instantiation syntax.


nickrolfe (approved):

aibaars (approved):
Good catch.

PR #7377, opened 2021-12-13
C#: Introduce class `Overridable`

hvitved:
The class `Overridable` generalizes the existing class `Virtualizable` by also including accessors. This allows for quite a bit of code to be simplified.


michaelnebel (approved):
Very nice!!

PR #7378, opened 2021-12-13
Ruby: Prevent infinite recursion in module resolution library

hvitved:
The following Ruby code made the module resolution algorithm diverge:
```rb
class B::A
end

class A::B
end

class A < B
  include B
end

include A
```


aibaars (approved):
Looks good to me.

aibaars (approved):

PR #6760, opened 2021-09-27
Increase precision to high for cpp/static-buffer-overflow

andersfugmann:
**As this changes the precision on static-buffer-overflow query, please do not merge before GHES 3.3 has been branched.**

This PR includes:
* Further relax predicate `memberMaybeVarsized` to include all members of size 0 or 1
* Remove warnings for static buffer overflows in when fields are indexed as an argument to `offsetof`
* Do not warn on static buffer overflows if upper bound analysis used widening. 
* Weaken wording on overflow static alert text (`is accessed` -> `may be accessed`) 

Statistics from running OverflowStatic on all LGTM projects:
|  | |
| ----------- | ----------- |
| Total Alerts | 206 | 
| Projects with alerts | 91 | 
| False positives | 16 | 
| Unknown* | 6 | 

[*] Unknown alerts denotes alerts, where I cannot judge if the alerts is correct. These usually includes nested macro expansions. 

The false positives can be grouped into:
* Copying multiple fields (7 FP's)
* Limitations in the range analysis (1 FP)
  * 1 reasonably trivial 
  * 3 are non trivial, and would require a deeper invariant analysis. 
* Dead code (2 FP's)
* Wrong assumption of `page_size` (evaluated to 0) (1 FP)
* Variable sized member of size 4 (`char data[sizeof(int)]`) (2 FP)

Detailed results and the actual analysis can be found here: 
https://gist.github.com/andersfugmann/5091e7c63cd4f6ae94d3b86e426efd77

PR #7386, opened 2021-12-14
C++: split `cpp/overrunning-write` into two

redsun82:
This splits the `cpp/overruning-write` into two separate queries based
off on the reason for the estimation. If the overrun is detected based
on non-trivial range analysis, the results are now marked by the new
`cpp/very-likely-overruning-write` high precision query. If it is based
on less precise, usually type based bounds, then it will still be marked
by `cpp/overruning-write` which remains at medium precision.

Closes: https://github.com/github/codeql-c-team/issues/751


redsun82:
I'll now create the internal request and start a DCA experiment on this, and then maybe play around with some LGTM databases.

cpp/ql/src/Security/CWE/CWE-120/OverrunWrite.ql
```diff
@@ -21,12 +21,12 @@ import semmle.code.cpp.commons.Alloc
  * See CWE-120/UnboundedWrite.ql for a summary of CWE-120 alert cases.
  */
 
-from BufferWrite bw, Expr dest, int destSize, int estimated
+from BufferWrite bw, Expr dest, int destSize, int estimated, TypeBoundsAnalysis reason
```
MathiasVP, line 24: We need to make sure that we don't report any alerts twice with these two queries (i.e., there's no result for which `bw.getMaxDataLimited(any(TypeBoundsAnalysis reason))` and `bw.getMaxDataLimited(any(ValueFlowAnalysis reason))` holds). That sounds like something that's true. Is that correct?

MathiasVP (commented):
The code LGTM, I think (see my one comment). A few things need to happen before we can merge this:
1. We need to run a DCA on the code-scanning suite to make sure the performance on that suite is good with this new query.
2. We need to add a .qhelp file, which is probably just a copy/paste of the one we have for `cpp/overrunning-write`.
3. We need to run the new query on the ~140 projects we usually test on (or even better: on all of LGTM) and make sure that the new query doesn't have too many FPs.
4. Once that's done, we need to add the `ready-for-docs-review` label on this PR so that the docs team can review the .qhelp file, the metadata, and stuff like that.

MathiasVP:
> I'll now create the internal request and start a DCA experiment on this, and then maybe play around with some LGTM databases.

Great!

cpp/ql/src/Security/CWE/CWE-120/OverrunWrite.ql
redsun82: yes. Apart from making sure the code had this property, in a previous version of my other PR where the reasons where printed I double checked there were no duplicate reasons being reported for the same cases. I think the DCA experiment that @geoffw0 run had that reason printing in, so I'm pretty confident there is no overlap. I'll still check that further though.

cpp/ql/src/Security/CWE/CWE-120/OverrunWrite.qhelp
```diff
@@ -13,13 +13,13 @@
...
+<p>In this example, the call to <code>sprintf</code> writes a message of 14 characters (including the terminating null) plus the length of the string conversion of `userId` into a buffer with space for just 18 characters.  As such, if `userId` is greater or equal to `10000`, the last characters overflow the buffer resulting in undefined behavior.</p>
 
-<p>To fix this issue three changes should be made:</p>
+<p>To fix this issue one of three changes should be made:</p>
 <ul>
-  <li>Control the size of the buffer using a preprocessor define.</li>
-  <li>Replace the call to <code>strcpy</code> with <code>strncpy</code>, specifying the define as the maximum length to copy.  This will prevent the buffer overflow.</li>
-  <li>Consider increasing the buffer size, say to 20 characters, so that the message is displayed correctly.</li>
+  <li>Preferably, replace the call to <code>sprintf</code> with <code>snprintf</code>, specifying a define or `sizeof(buffer)` as maximum length to copy. This will prevent the buffer overflow.</li>
```
MathiasVP, line 20: I understand what "specifying a define" means, but it sounds very imprecise. The previous text had a bullet before this one saying "Control the size of the buffer using a preprocessor define.". Could we add that here as well to explain what "specifying a define" mean? Maybe even change it to:

- Control the size of the buffer by declaring it with a constant size.
- [...], specifying the constant size, or `sizeof(buffer)` as maximum length to copy. This will prevent the buffer overflow.

And the same applies to the `VeryLikelyOverrunWrite.qhelp` file.

cpp/ql/src/Security/CWE/CWE-120/VeryLikelyOverrunWrite.ql
```diff
@@ -0,0 +1,34 @@
+/**
+ * @name Likely overrunning write based on non-trivial analysis of value ranges
```
MathiasVP, line 2:
```suggestion
 * @name Likely overrunning write
```
Again, I think it's best to avoid implementation details in the name.

cpp/change-notes/2021-12-14-overruning-write-split.md
```diff
@@ -0,0 +1,2 @@
+lgtm,codescanning
+* A new `cpp/very-likely-overruning-write` query has been added for C/C++, which reports results formerly flagged by `cpp/overruning-write` where a non-trivial range analysis shows a higher likelyhood of an actual problem. As a consequence, the new query is tagged with high precision.
```
MathiasVP, line 2: I think this is maybe too technical for a change-note. How about this instead? (I've also suggested a note about the query being enabled by default.)

```suggestion
* A new `cpp/very-likely-overruning-write` query has been added to the default query suite for C/C++. The query reports some results that were formerly flagged by `cpp/overruning-write`.
```

MathiasVP (commented):
Oops. Forgot to add the rest of the comments. Sorry!

cpp/ql/src/Security/CWE/CWE-120/OverrunWrite.c
```diff
@@ -1,9 +1,9 @@
-void sayHello()
+void sayHello(uint32_t userId)
 {
-	char buffer[10];
+	char buffer[18];
 
-	// BAD: this message overflows the buffer
-	strcpy(buffer, "Hello, world!");
+	// BAD: this message overflows the buffer if userId >= 10000
+	sprintf(buffer, "Hello, user %d!", userId);
```
geoffw0, line 6: I see, the original example would be a result for `VeryLikelyOverrunWrite.c` not `OverrunWrite.c` now. :+1:

redsun82:
On `vim/vim`, I'm getting a couple of false positives on the new query that boils down to to the trailing array idiom:
```cpp
v = &fc->fixvar[fixvar_idx++].var;
name = v->di_key;
STRCPY(name, "self");
```
where `v` is of type
```cpp
struct dictitem_S
{
    typval_T	di_tv;		// type and value of the variable
    char_u	di_flags;	// DI_FLAGS_ flags (only used for variable)
    char_u	di_key[1];	// key (actually longer!)
};
typedef struct dictitem_S dictitem_T;
```
and the items are `malloc`ed with a bigger size than the size of the item.

Any thoughts? Is this ok, or should we try to detect such a case somehow?

cpp/ql/src/Security/CWE/CWE-120/VeryLikelyOverrunWrite.qhelp
```diff
@@ -0,0 +1,35 @@
...
+<overview>
+<p>The program performs a buffer copy or write operation with no upper limit on the size of the copy, and by analysing the bounds of the expressions involved it appears that certain inputs will cause a buffer overflow to occur in this case.  In addition to causing program instability, techniques exist which may allow an attacker to use this vulnerability to execute arbitrary code.</p>
+
+</overview>
+<recommendation>
+<p>Always control the length of buffer copy and buffer write operations.  <code>strncpy</code> should be used over <code>strcpy</code>, <code>snprintf</code> over <code>sprintf</code>, and in other cases 'n-variant' functions should be preferred.</p>
+
+</recommendation>
+<example>
+<sample src="OverrunWrite.c" />
```
geoffw0, line 14:
```suggestion
<sample src="VeryLikelyOverrunWrite.c" />
```

cpp/ql/src/Security/CWE/CWE-120/VeryLikelyOverrunWrite.ql
```diff
@@ -0,0 +1,34 @@
...
+ *       external/cwe/cwe-120
+ *       external/cwe/cwe-787
+ *       external/cwe/cwe-805
+ */
+
+import semmle.code.cpp.security.BufferWrite
+import semmle.code.cpp.commons.Alloc
+
+/*
+ * See CWE-120/UnboundedWrite.ql for a summary of CWE-120 alert cases.
```
geoffw0, line 21: The table in `CWE-120/UnboundedWrite.ql` should have a reference to the new query actually, alongside `OverrunWrite.ql` and `OverrunWriteFloat.ql`.

cpp/config/suites/security/cwe-120
```diff
@@ -5,9 +5,11 @@
   @name Badly bounded write (CWE-120)
 + semmlecode-cpp-queries/Security/CWE/CWE-120/OverrunWrite.ql: /CWE/CWE-120
   @name Potentially overrunning write (CWE-120)
++ semmlecode-cpp-queries/Security/CWE/CWE-120/VeryLikelyOverrunWrite.ql: /CWE/CWE-120
+  @name Likely overrunning write based on non-trivial analysis of value ranges
```
geoffw0:
```suggestion
  @name Likely overrunning write
```
to match the change to the query `@name`.

geoffw0 (commented):
A few comments, otherwise LGTM.

MathiasVP:
> Any thoughts? Is this ok, or should we try to detect such a case somehow?

It would be great if we can filter those out. We should have code for this already. Take a look at https://github.com/github/codeql/blob/main/cpp/ql/lib/semmle/code/cpp/commons/Buffer.qll#L15. That predicate is actually already used to filter out some results in `getBufferSize` (which this new query is actually using). So I'm surprised we're seeing these FPs.

redsun82:
on `kraflab/dsda-doom` I found a bunch of quite interesting FPs.  [Here's an example][1]: `j` is bound between characters `'0'` and `'9'`, so that `j - 48` is actually bounded between `0` and `9`, so that the formatting `"STTNUM%.1d",j-48` actually fits in 8 bytes, but the query estimates the maximum length to 10. There are two other similar FP around that, but interestingly a bunch of other `sprintf`'s around there on the same buffer are correctly not flagged.

[1]: https://lgtm.com/projects/g/kraflab/dsda-doom/snapshot/9780cd0c0c64d5d7a71f7bd3834d35b1a6177d7b/files/prboom2/src/hu_stuff.c?sort=name&dir=ASC&mode=heatmap#xe36c196b770e580f:1

redsun82:
> > Any thoughts? Is this ok, or should we try to detect such a case somehow?
> 
> It would be great if we can filter those out. We should have code for this already. Take a look at https://github.com/github/codeql/blob/main/cpp/ql/lib/semmle/code/cpp/commons/Buffer.qll#L15. That predicate is actually already used to filter out some results in `getBufferSize` (which this new query is actually using). So I'm surprised we're seeing these FPs.

Hmm, actually the full code snippet is
```cpp
// Set l:self to "selfdict".  Use "name" to avoid a warning from
// some compiler that checks the destination size.
v = &fc->fixvar[fixvar_idx++].var;
name = v->di_key;
STRCPY(name, "self");
```
Might it be that the aliasing they use to avoid compiler warnings is actually stumping our own code detecting this case?

redsun82:
> on `kraflab/dsda-doom` I found a bunch of quite interesting FPs. [Here's an example](https://lgtm.com/projects/g/kraflab/dsda-doom/snapshot/9780cd0c0c64d5d7a71f7bd3834d35b1a6177d7b/files/prboom2/src/hu_stuff.c?sort=name&dir=ASC&mode=heatmap#xe36c196b770e580f:1): `j` is bound between characters `'0'` and `'9'`, so that `j - 48` is actually bounded between `0` and `9`, so that the formatting `"STTNUM%.1d",j-48` actually fits in 8 bytes, but the query estimates the maximum length to 10. There are two other similar FP around that, but interestingly a bunch of other `sprintf`'s around there on the same buffer are correctly not flagged.

This indeed seems like a bug with range analysis, I've managed to reduce the false positive case to https://github.com/github/codeql/pull/7396 . Can you have a look @geoffw0 ?

redsun82:
> > Any thoughts? Is this ok, or should we try to detect such a case somehow?
> 
> It would be great if we can filter those out. We should have code for this already. Take a look at https://github.com/github/codeql/blob/main/cpp/ql/lib/semmle/code/cpp/commons/Buffer.qll#L15. That predicate is actually already used to filter out some results in `getBufferSize` (which this new query is actually using). So I'm surprised we're seeing these FPs.

I think

redsun82:
> > > Any thoughts? Is this ok, or should we try to detect such a case somehow?
> > 
> > 
> > It would be great if we can filter those out. We should have code for this already. Take a look at https://github.com/github/codeql/blob/main/cpp/ql/lib/semmle/code/cpp/commons/Buffer.qll#L15. That predicate is actually already used to filter out some results in `getBufferSize` (which this new query is actually using). So I'm surprised we're seeing these FPs.
> 
> Hmm, actually the full code snippet is
> 
> ```c++
> // Set l:self to "selfdict".  Use "name" to avoid a warning from
> // some compiler that checks the destination size.
> v = &fc->fixvar[fixvar_idx++].var;
> name = v->di_key;
> STRCPY(name, "self");
> ```
> 
> Might it be that the aliasing they use to avoid compiler warnings is actually stumping our own code detecting this case?
> 
> Or maybe it is because that field is not `malloc`ed in the same scope?

yes, I think that's it. As the copy is not done directly on `v->di_key` but on the alias `name`, then `memberMayBeVarSize(_, name)` does not hold.

Do you think we should enhance `memberMayBeVarSize` to follow aliases, or should we just accept this kind of FP?

MathiasVP:
> Do you think we should enhance `memberMayBeVarSize to` follow aliases, or should we just accept this kind of FP?

I _think_ it's okay to accept this kind of FP for now (since it didn't stop increasing the precision of another query in https://github.com/github/codeql/pull/6760).

cpp/ql/lib/semmle/code/cpp/commons/Printf.qll
```diff
@@ -359,6 +376,17 @@ private int lengthInBase10(float f) {
   result = f.log10().floor() + 1
 }
 
+private BufferWriteEstimationReason getEstimationReasonForIntegralExpression(Expr expr) {
```
MathiasVP, line 379: I'm not sure `expr` is sufficiently bound in this predicate, but let's test this with DCA before we make any changes. I'd be much less nervous about this if we added a `bindingset[expr]` to it.

cpp/ql/lib/semmle/code/cpp/commons/Printf.qll
```diff
@@ -60,12 +67,26 @@ class TypeBoundsAnalysis extends BufferWriteEstimationReason, TTypeBoundsAnalysi
...
+ * but a widening aproximation might have been used for variables in loops.
+ * For example
+ * ```
+ * for (int i = 0; i < 10; ++i) {
+ *    int j = i + i;
+ *    //...  <- estimation done here based on j
+ * }
+ * ```
+ */
+class WidenedValueFlowAnalysis extends BufferWriteEstimationReason, TWidenedValueFlowAnalysis {
```
MathiasVP, line 83: Great to see the strategy concept already being used an extended :)

cpp/ql/lib/semmle/code/cpp/commons/Printf.qll
redsun82: If this is a problem we might see it on performance? Should I add the tuple counting option?

MathiasVP: Yeah, you can try adding tuple counting. Be sure to update DCA to get the latest updates to that feature.

intrigus-lgtm:
I assume that it is not possible to do some sharing of the query help using [query help inclusion](https://github.com/github/codeql/blob/15caaa7ad6d91c533596204b6a863a1c56240628/docs/codeql/writing-codeql-queries/query-help-files.rst#block-level-include-elements)?

redsun82:
> I assume that it is not possible to do some sharing of the query help using [query help inclusion](https://github.com/github/codeql/blob/15caaa7ad6d91c533596204b6a863a1c56240628/docs/codeql/writing-codeql-queries/query-help-files.rst#block-level-include-elements)?

I did not know about this mechanism, I'll have a look, thanks!

redsun82:
I've run the query on the whole LGTM, results will be available here:

https://lgtm.com/query/2517655735938554920

MathiasVP:
The DCA experiment failed due to a memory issue on the join-order badness analysis (that's my fault, sorry!)

It's still not fixed, but you can still generate the data _locally_ by doing:
```
./dca reports --name redsun82/pr-7386-cc8a646__Differences__code-scanning --redo-report readme
```

when I do that, I get the following table (I've removed a few of the columns for simplicity):
```
| source                  | predicate                                                                           | b\_result\_size | b\_max\_tuple\_count | b\_metric
| nlohmann\_\_json        | Printf::FormatLiteral::getMaxConvertedLengthAfterLimited\#ffff \(recursive delta\)  | 158             | 225                  | 1.424
| microsoft\_\_calculator | Printf::FormatLiteral::getMaxConvertedLengthAfterLimited\#ffff \(recursive delta\)  | 4               | 6                    | 1.5  
| vim\_\_vim              | m\#Declaration::Declaration::hasGlobalOrStdName\_dispred\#bf                        | 8901            | 14635                | 1.644
| vim\_\_vim              | Printf::FormatLiteral::getMaxConvertedLengthAfterLimited\#ffff \(recursive delta\)  | 1978            | 3711                 | 1.876
| microsoft\_\_ChakraCore | Printf::FormatLiteral::getMaxConvertedLengthAfterLimited\#ffff \(recursive delta\)  | 3323            | 6561                 | 1.974
| dsp-testing\_\_putty    | Printf::FormatLiteral::getMaxConvertedLengthAfterLimited\#ffff \(recursive delta\)  | 1958            | 4341                 | 2.217
| git\_\_git              | Printf::FormatLiteral::getMaxConvertedLengthAfterLimited\#ffff \(recursive delta\)  | 5984            | 14886                | 2.488
| microsoft\_\_calculator | Printf::FormatLiteral::getIntegralDisplayType\_dispred\#bff                         | 6               | 54                   | 3    
| git\_\_git              | Scanf::ScanfFormatLiteral::parseConvSpec\_dispred\#ffffff                           | 2               | 8                    | 4    
| vim\_\_vim              | Scanf::ScanfFormatLiteral::parseConvSpec\_dispred\#ffffff                           | 16              | 64                   | 4    
| dsp-testing\_\_putty    | Scanf::ScanfFormatLiteral::parseConvSpec\_dispred\#ffffff                           | 22              | 88                   | 4    
| vim\_\_vim              | StringAnalysis::AnalysedString::getMaxLength\#ff \(recursive delta\)                | 16058           | 74460                | 4.637
| git\_\_git              | StringAnalysis::AnalysedString::getMaxLength\#ff \(recursive delta\)                | 27983           | 134538               | 4.808
| facebook\_\_yoga        | StringAnalysis::AnalysedString::getMaxLength\#ff \(recursive delta\)                | 241             | 1302                 | 5.402
| abseil\_\_abseil-cpp    | StringAnalysis::AnalysedString::getMaxLength\#ff \(recursive delta\)                | 1855            | 10062                | 5.424
| dsp-testing\_\_putty    | StringAnalysis::AnalysedString::getMaxLength\#ff \(recursive delta\)                | 9202            | 50688                | 5.508
| microsoft\_\_ChakraCore | StringAnalysis::AnalysedString::getMaxLength\#ff \(recursive delta\)                | 23079           | 134778               | 5.84 
| microsoft\_\_calculator | StringAnalysis::AnalysedString::getMaxLength\#ff \(recursive delta\)                | 1849            | 10944                | 5.919
| nlohmann\_\_json        | StringAnalysis::AnalysedString::getMaxLength\#ff \(recursive delta\)                | 31419           | 186444               | 5.934
| an-tao\_\_drogon        | StringAnalysis::AnalysedString::getMaxLength\#ff \(recursive delta\)                | 8410            | 50064                | 5.953
| nlohmann\_\_json        | Printf::FormatLiteral::getConvSpec\_dispred\#fff\#shared\#1                         | 83              | 581                  | 7    
| dsp-testing\_\_putty    | Printf::FormatLiteral::getConvSpec\_dispred\#fff\#shared\#1                         | 1455            | 10185                | 7    
| microsoft\_\_ChakraCore | Printf::FormatLiteral::getConvSpec\_dispred\#fff\#shared\#1                         | 2335            | 16345                | 7    
| git\_\_git              | Printf::FormatLiteral::getConvSpec\_dispred\#fff\#shared\#1                         | 5035            | 35245                | 7    
```

I think this shows that there's a potential performance problem that we need to look into before we merge this.

skedwards88:
Docs first responder here :wave: I've added this to our review board for a writer to review.

cpp/ql/src/Security/CWE/CWE-120/OverrunWrite.qhelp
```diff
@@ -6,30 +6,20 @@
...
-<p>In this example, the call to <code>strcpy</code> copies a message of 14 characters (including the terminating null) into a buffer with space for just 10 characters.  As such, the last four characters overflow the buffer resulting in undefined behavior.</p>
+<p>In this example, the call to <code>sprintf</code> writes a message of 14 characters (including the terminating null) plus the length of the string conversion of `userId` into a buffer with space for just 18 characters.  As such, if `userId` is greater or equal to `10000`, the last characters overflow the buffer resulting in undefined behavior.</p>
 
-<p>To fix this issue three changes should be made:</p>
+<p>To fix this issue these changes should be made:</p>
 <ul>
-  <li>Control the size of the buffer using a preprocessor define.</li>
-  <li>Replace the call to <code>strcpy</code> with <code>strncpy</code>, specifying the define as the maximum length to copy.  This will prevent the buffer overflow.</li>
-  <li>Consider increasing the buffer size, say to 20 characters, so that the message is displayed correctly.</li>
+  <li>Control the size of the buffer by declaring it with a compile time constant</li>
```
skedwards88, line 17: nit to be consistent with the other list items
```suggestion
  <li>Control the size of the buffer by declaring it with a compile time constant.</li>
```

cpp/ql/src/Security/CWE/CWE-120/VeryLikelyOverrunWrite.qhelp
```diff
@@ -0,0 +1,24 @@
+<!DOCTYPE qhelp PUBLIC
+  "-//Semmle//qhelp//EN"
+  "qhelp.dtd">
+<qhelp>
+<overview>
+<p>The program performs a buffer copy or write operation with no upper limit on the size of the copy, and by analysing the bounds of the expressions involved it appears that certain inputs will cause a buffer overflow to occur in this case.  In addition to causing program instability, techniques exist which may allow an attacker to use this vulnerability to execute arbitrary code.</p>
```
skedwards88, line 6: nit
```suggestion
<p>The program performs a buffer copy or write operation with no upper limit on the size of the copy. By analyzing the bounds of the expressions involved, it appears that certain inputs will cause a buffer overflow to occur in this case.  In addition to causing program instability, techniques exist which may allow an attacker to use this vulnerability to execute arbitrary code.</p>
```

cpp/ql/src/Security/CWE/CWE-120/VeryLikelyOverrunWrite.qhelp
```diff
@@ -0,0 +1,24 @@
...
+</overview>
+<include src="OverrunWriteRecommendation.inc.qhelp" />
+<example>
+<sample src="VeryLikelyOverrunWrite.c" />
+
+<p>In this example, the call to <code>sprintf</code> writes a message of 14 characters (including the terminating null) plus the length of the string conversion of `userId` into a buffer with space for just 17 characters. While `userId` is checked to occupy no more than 4 characters when converted, there is no space in the buffer for the terminating null character if `userId >= 1000`. In this case, the null character overflows the buffer resulting in undefined behavior.</p>
+
+<p>To fix this issue these changes should be made:</p>
+<ul>
+  <li>Control the size of the buffer by declaring it with a compile time constant</li>
```
skedwards88, line 17:
```suggestion
  <li>Control the size of the buffer by declaring it with a compile time constant.</li>
```

skedwards88 (approved):
Looks good from a writer perspective!

MathiasVP (approved):
LGTM!

PR #7390, opened 2021-12-14
Ruby: Deprecate `Pattern` classes

hvitved:
The various `Pattern` classes got introduced before we supported pattern matching, as a common way of describing destructured assignments and destructured parameters. Now, however, they give more rise to confusion (they have nothing to do with pattern matching) than benefit, so I think we should simply deprecate them.

I also fixed a bug in the computation of callables belonging to nested destructured parameters.


ruby/ql/lib/codeql/ruby/ast/Expr.qll
```diff
@@ -66,6 +66,83 @@ class ArgumentList extends Expr, TArgumentList {
...
+
+  override Variable getAVariable() {
+    result = this.getElement(_).(VariableWriteAccess).getVariable()
+    or
+    result = this.getElement(_).(DestructuredLhsExpr).getAVariable()
+  }
+
+  override string toString() { result = "(..., ...)" }
+
+  override AstNode getAChild(string pred) { pred = "getElement" and result = this.getElement(_) }
```
nickrolfe, line 143: Should this also have `or result = super.getChild(pred)`?

ruby/ql/lib/codeql/ruby/ast/internal/AST.qll
```diff
@@ -342,7 +342,7 @@ private module Cached {
         TStringConcatenation or TStringEscapeSequenceComponent or TStringInterpolationComponent or
         TStringTextComponent or TSubExprReal or TSubshellLiteral or TSymbolArrayLiteral or
         TTernaryIfExpr or TThen or TTokenConstantAccess or TTokenMethodName or TTokenSuperCall or
-        TToplevel or TTrueLiteral or TTuplePatternParameter or TUnaryMinusExpr or TUnaryPlusExpr or
+        TToplevel or TTrueLiteral or TDestructuredParameter or TUnaryMinusExpr or TUnaryPlusExpr or
```
nickrolfe, line 345: Perhaps move this to keep the alphabetical ordering?

nickrolfe:
A couple of trivial questions, but otherwise this looks very sensible.

hvitved:
> A couple of trivial questions, but otherwise this looks very sensible.

I don't see any questions?

nickrolfe:
> > A couple of trivial questions, but otherwise this looks very sensible.
> 
> I don't see any questions?

Sorry, you should see them now. I was reviewing from VS Code, but the extension seems very flaky when it comes to submitting reviews.

nickrolfe (approved):
LGTM

PR #7393, opened 2021-12-14
Ruby: SimpleParameter should not be an Expr

aibaars:
PR #7394 makes all expressions "post-order" nodes in the control flow graph except for [`SimpleParameter`](https://github.com/github/codeql/pull/7394/files#diff-32ce71534428bfb48247fdc1f6ca40704b369d31bb38b73e2972cadd97367b1aR1) . I think parameters should not be instance of `Expr` , therefore, this pull requests removes the `Expr` class from the super class list of `SimpleParameter`.

### DIFF
--- a/ruby/ql/lib/codeql/ruby/ast/Erb.qll
+++ b/ruby/ql/lib/codeql/ruby/ast/Erb.qll
@@ -118,7 +118,7 @@ private class ErbDirectiveFile extends File {
 
   /** Gets a statement in this file. */
   pragma[nomagic]
-  Stmt getAStmt(int startLine, int startColumn) {
+  AstNode getAnAstNode(int startLine, int startColumn) {
     exists(Location loc |
       result.getLocation() = loc and
       loc.getFile() = this and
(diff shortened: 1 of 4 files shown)


ruby/ql/lib/codeql/ruby/ast/Erb.qll
```diff
@@ -159,7 +159,7 @@ class ErbDirective extends TDirectiveNode, ErbAstNode {
    */
   Stmt getAChildStmt() {
     this.containsStmtStart(result) and
-    not this.containsStmtStart(result.getParent())
+    not this.containsStmtStart(parent*(result).getParent())
```
hvitved, line 162: Why this unrelated change?

hvitved (commented):
LGTM, but I don't understand the ERB change.

ruby/ql/lib/codeql/ruby/ast/Erb.qll
aibaars: It's not unrelated. One of the synthesized variable accesses has a SimpleParameter as parent and because SimpleParameter is not longer a Stmt( Expr extends Stmt) this negation suddenly held. The `parent*` change is to find the nearest parent that is instanceof `Stmt`.

alexrford (dismissed):

ruby/ql/lib/codeql/ruby/ast/Erb.qll
hvitved: What if we generalize `containsStmtStart(Stmt s)` to `containsNodeStart(AstNode n)` instead?

aibaars: That should work too.

hvitved (approved):

PR #7394, opened 2021-12-14
Ruby: CFG: make all expressions "post-order" nodes

aibaars:


ruby/ql/lib/codeql/ruby/controlflow/internal/ControlFlowGraphImpl.qll
```diff
@@ -416,17 +399,31 @@ module Trees {
         first(this.getBranch(i + 1), succ) and
         c.(ConditionalCompletion).getValue() = false
       )
+      or
+      succ = this and
```
hvitved, line 401: Don't we need a `c instanceof NormalCompletion` check here?

ruby/ql/lib/codeql/ruby/controlflow/internal/ControlFlowGraphImpl.qll
```diff
@@ -452,6 +445,15 @@ module Trees {
...
         c.(MatchingCompletion).getValue() = false
       )
+      or
+      succ = this and
+      (
+        last(this.getABranch(), pred, c) and
+        not c.(MatchingCompletion).getValue() = false
+        or
+        last(this.getElseBranch(), pred, c) and
+        c instanceof NormalCompletion
```
hvitved, line 450: Move up after `succ = this and` to also cover the first disjunct.

ruby/ql/lib/codeql/ruby/controlflow/internal/ControlFlowGraphImpl.qll
```diff
@@ -748,51 +751,44 @@ module Trees {
...
-      c instanceof SimpleCompletion
-      or
-      exists(Expr next |
-        last(this.getPattern(), pred, c) and
-        c.(MatchingCompletion).getValue() = true and
-        first(next, succ)
-      |
-        next = this.getCondition()
+      last(this.getPattern(), pred, c) and
+      not c.(MatchingCompletion).getValue() = false and
```
hvitved, line 772: I think this should be `c.(MatchingCompletion).getValue() = true`.

ruby/ql/lib/codeql/ruby/controlflow/internal/ControlFlowGraphImpl.qll
```diff
@@ -1146,24 +1146,23 @@ module Trees {
...
-      first(this.getHandler(), succ)
+      (
+        c instanceof RaiseCompletion and
+        first(this.getHandler(), succ)
+        or
+        not c instanceof RaiseCompletion and
+        succ = this
+      )
+      or
+      last(this.getHandler(), pred, c) and
```
hvitved, line 1164: add `c instanceof NormalCompletion`

nickrolfe (commented):
I haven't reviewed the test output, but the changes in `ControlFlowGraphImpl.qll` all look sensible.

hvitved (changes requested):
Looks great, it even simplified some existing logic. I suggest we rebase this once https://github.com/github/codeql/pull/7393 has been merged, and then get rid of the last commit.

ruby/ql/lib/codeql/ruby/controlflow/internal/ControlFlowGraphImpl.qll
aibaars: A pattern may also end with a simple completion, for example a `*var` never fails.

ruby/ql/lib/codeql/ruby/controlflow/internal/ControlFlowGraphImpl.qll
aibaars: don't we get that "for free" because `value` and `branch(_)` are mentioned in `propagatesAbnormal`?

ruby/ql/lib/codeql/ruby/controlflow/internal/ControlFlowGraphImpl.qll
aibaars: Could write the following instead `c instanceof MatchingCompletion implies c.(MatchingCompletion).getValue() = true`

ruby/ql/consistency-queries/CfgConsistency.ql
```diff
@@ -1 +1,24 @@
 import codeql.ruby.controlflow.internal.ControlFlowGraphImplShared::Consistency
+import codeql.ruby.AST
+import codeql.ruby.controlflow.internal.ControlFlowGraphImpl
+
+/**
+ * All `Expr` nodes are `PostOrderTree`s
+ */
+query predicate nonPostOrderExprTypes(string cls) {
```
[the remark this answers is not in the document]
aibaars: That gives quite a few results : https://github.com/github/codeql/runs/4549392461?check_suite_focus=true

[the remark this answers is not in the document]
hvitved: That made me realize that the modelling of `InClause` is actually not like (I think) it should be. There is a subtle difference: before, the node for the `InClause` itself was visited regardless of a match, but now it is only visited if there is a match.

There are two options: Revert 1343ed58a21eec2954876d8d42e877a382ba89c8 or use CFG splitting to make sure that `InClauses` are truly post-order. I don't think splitting is worth the effort, so I suggest a revert.

With a revert, we can update the check above to

```ql
query predicate nonPostOrderExpr(Expr e, string cls) {
  cls = e.getPrimaryQlClasses() and
  not exists(e.getDesugared()) and
  not e instanceof InClause and
  not e instanceof BeginExpr and
  not e instanceof Namespace and
  not e instanceof Toplevel and
  exists(AstNode last, Completion c |
    last(e, last, c) and
    last != e and
    c instanceof NormalCompletion
  )
}
```

ruby/ql/lib/codeql/ruby/controlflow/internal/ControlFlowGraphImpl.qll
hvitved: `propagatesAbnormal` only gives you the escaping `raise` edge (typically to the exit point), it doesn't prevent an internal edge.

ruby/ql/consistency-queries/CfgConsistency.ql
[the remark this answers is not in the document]
aibaars: > That made me realize that the modelling of `InClause` is actually not like (I think) it should be. There is a subtle difference: before, the node for the `InClause` itself was visited regardless of a match, but now it is only visited if there is a match.

I think this is the correct behaviour; if there is no match then the `InClause` has no value (not even `nil`).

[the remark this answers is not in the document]
aibaars: As discussed, perhaps `InClause` shouldn't be an `Expr` in the first place.

hvitved (approved):
LGTM. I have started a final DCA run; let's wait for that before we merge.

hvitved (dismissed):

hvitved (approved):

Issue #7385, opened 2021-12-14
LGTM.com - false positive `cpp/comparison-of-identical-expressions`


bblanchon:
**Description of the false positive**

I'm using a meta-programming trick that reduces the number of overloads by sorting the arguments to a function like so:

```c++
template <typename TString1, typename TString2>
typename std::enable_if<TString1::typeSortKey <= TString2::typeSortKey, int>::type
stringCompare(TString1 s1, TString2 s2) {
  // IMPLEMENTATION
}

template <typename TString1, typename TString2>
typename std::enable_if<(TString1::typeSortKey > TString2::typeSortKey), int>::type
stringCompare(TString1 s1, TString2 s2) {
  return -stringCompare(s2, s1);
}
```

As you can see, this code uses the same implementation even if the arguments  are reversed.

The alert "Self Comparison" triggers on the expression `TString1::typeSortKey <= TString2::typeSortKey`.
Indeed, it can be a self-comparison if `TString1` and `TString2` are the same type, but in the general case, it's not.

**URL to the alert on the project page on LGTM.com**

https://lgtm.com/projects/g/bblanchon/ArduinoJson/snapshot/6d857fd73270f4ee1e03f5e9aa5c1e7d1ee0f179/files/src/ArduinoJson/Strings/StringAdapters.hpp?sort=name&dir=ASC&mode=heatmap#x262128796e29999c:1

MathiasVP:
Hi @bblanchon,

Thanks for the report. Indeed, this looks like a false positive. I've opened a PR with a fix here: https://github.com/github/codeql/pull/7395.

MathiasVP:
The PR has been merged 🎉. The fix will reach LGTM within the next couple of weeks :)

bblanchon:
Awesome. Thank Mathias!

MathiasVP:
This is unfortunately expected, yes. We update LGTM roughly 14 days, but the last update before Christmas had to be canceled due to technical issues. I expect the warning to disappear within the next couple of days, though.

MathiasVP:
We've just updated LGTM and the false positive is now gone.

bblanchon:
Awesome! ArduinoJson is now A+ 😍

PR #7395, opened 2021-12-14
C++: Fix FP in `cpp/comparison-of-identical-expressions`

MathiasVP:
Fixes https://github.com/github/codeql/issues/7385.

Initially, I wanted to add this case to `Expr.isUnevaluated` since that would make this fix apply more broadly. But I guess, `Expr.isUnevaluated` shouldn't hold for things that are evaluated _at compile-time only_. So for now I've just special-cased this in the query containing the actual FP report.

### DIFF
--- a/cpp/ql/test/query-tests/Likely Bugs/Arithmetic/BadAdditionOverflowCheck/templates.cpp
+++ b/cpp/ql/test/query-tests/Likely Bugs/Arithmetic/BadAdditionOverflowCheck/templates.cpp
@@ -37,5 +37,5 @@ struct Value0 {
 };
 
 void instantiation_with_pointless_comparison() {
-  constant_comparison<Value0, Value0>();
+  constant_comparison<Value0, Value0>(); // GOOD
 }
--- a/cpp/ql/test/query-tests/Likely Bugs/Arithmetic/BadAdditionOverflowCheck/templates.cpp
+++ b/cpp/ql/test/query-tests/Likely Bugs/Arithmetic/BadAdditionOverflowCheck/templates.cpp
@@ -20,3 +20,22 @@ bool compareValues() {
 bool callCompareValues() {
   return compareValues<C1, C2> || compareValues<C1, C1>();
 }
+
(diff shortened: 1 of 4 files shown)


cpp/ql/test/query-tests/Likely Bugs/Arithmetic/BadAdditionOverflowCheck/templates.cpp
```diff
@@ -20,3 +20,22 @@ bool compareValues() {
...
+typename enable_if<T1::value <= T2::value, bool>::type constant_comparison() {
+  return true;
+}
+
+struct Value0 {
+  const static int value = 0;
+};
+
+void instantiation_with_pointless_comparison() {
+  constant_comparison<Value0, Value0>();
```
geoffw0, line 40:
```suggestion
  constant_comparison<Value0, Value0>(); // GOOD
```

geoffw0 (commented):
LGTM.

I'd like to see a real-world or DCA test of some sort to confirm this doesn't affect results more widely than expected.

MathiasVP:
> LGTM.
> 
> I'd like to see a real-world or DCA test of some sort to confirm this doesn't affect results more widely than expected.

Good point. I'll do an LGTM-diff query on our usual projects since I don't think the query is actually run as part of security-extended.

MathiasVP:
No result changes on ~130 projects: https://lgtm.com/query/2133265528072233558/

cpp/ql/test/query-tests/Likely Bugs/Arithmetic/BadAdditionOverflowCheck/templates.cpp
MathiasVP: Doh. Thanks!

geoffw0:
Looks like we lose a (false positive) result for `cpp/comparison-of-identical-expressions`: https://lgtm.com/query/5961269154262927431/ :+1:

geoffw0 (approved):
All good.

PR #7398, opened 2021-12-14
Fix change notes

dbartol:
This PR has two separate commits.

- The first commit moves all change notes added since the last dist upgrade into the affected pack's change-notes directory, where the automatic changelog generation will pick them up. 
- The second commit renames the `change-notes` directory at the _language_ level (e.g., `cpp/change-notes`) to `old-change-notes` to reduce the temptation to add new change notes there. I'm working separately on a document to describe how to add change notes for a pack.


aeisenberg (approved):
This all looks good to me.

aeisenberg:
Are there any prepare release workflow changes that need to happen as well?

dbartol:
> Are there any prepare release workflow changes that need to happen as well?

Nope. This just moves everything so that the existing release prep workflow will work, and so that people won't keep adding new change notes in the wrong place. I did the same thing for the previous release, except for the renaming the old directory part.

dbartol:
See https://github.com/github/codeql/pull/7400 for the new instructions on creating change notes.

aschackmull:
Looks like `java/change-notes/2021-11-15-overrides.md` was accidentally deleted?

PR #7399, opened 2021-12-14
Ruby: Model what is written to the log from stdlib `Logger` methods

alexrford:
The `Logging` concept is borrowed from Python's `Concepts.qll`.

This encodes what may be present in log output from stdlib [`Logger`](https://ruby-doc.org/stdlib-3.0.2/libdoc/logger/rdoc/Logger.html) instances, covering the following methods:
- `add`/`log`
- `debug`, `error`, `fatal`, `info`, `unknown`, and `warn`
- `<<`
- `progname=`

`add`/`log` only execute their block and include its return value in log output when the `message` argument is `nil` (this is also its default value). See the test cases or [`Logger#add`](https://github.com/ruby/logger/blob/0fc272a3d472b3db635ae8e846c0d925dfb09ff4/lib/logger.rb#L459) for details.

### DIFF
--- a/ruby/ql/lib/codeql/ruby/Concepts.qll
+++ b/ruby/ql/lib/codeql/ruby/Concepts.qll
@@ -644,3 +644,32 @@ module Path {
     abstract class Range extends DataFlow::Node { }
   }
 }
+
+/**
+ * A data-flow node that logs data.
+ *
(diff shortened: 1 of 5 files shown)


nickrolfe (approved):
LGTM

Issue #7336, opened 2021-12-08
[Feature Request] Support Trusted Types in JavaScript queries


tosmolka:
Hello, I did not find much info about this elsewhere so asking here. 

It would be amazing if CodeQL had support for [Trusted Types](https://w3c.github.io/webappsec-trusted-types/dist/spec/) in JavaScript language and we could use CodeQL to query DOM injections sinks and catch violations early. Similar to what [tsec](https://github.com/google/tsec) is doing in their [rules](https://github.com/google/tsec/tree/main/src/rules/dom_security). 

I know many sinks (ScriptContentSink, ScriptUrlSink, EvalJavaScriptSink, DomSink, HtmlParserSink, TemplateTagInScriptSink, ...) are already covered to some extend. But it seems we might be missing some sinks (I did not check extensively, maybe SharedWorker constructor, React<iframe srcdoc>, HTMLBaseElement.href, ...). 

And I could not find anything related to trusted types (TrustedHTML, TrustedScript, TrustedScriptURL) or other relevant methods (TrustedTypePolicyFactory.createPolicy, TrustedTypePolicy.createHTML, TrustedTypePolicy.createScript, TrustedTypePolicy.createScriptURL). 

Now, without this support, it is challenging to continue scanning large and complex JavaScript applications that adopted Trusted Types. CodeQL does not understand "guard" methods and detected issues are not as precise as they could be. 

Thanks.

RasmusWL:
Hi @tosmolka, thanks for your suggestion. I've forwarded this to the JS team :+1:

asgerf:
Hi @tosmolka,

I've opened https://github.com/github/codeql/pull/7408 which adds sinks at the points where a value is boxed in a trusted type wrapper (i.e. the return value from the policy callback)

> CodeQL does not understand "guard" methods and detected issues are not as precise as they could be.

Without more information I can't really help with this. If you're seeing false positives due to a missed sanitizer guard, feel free to open an issue.

asgerf:
Closing as https://github.com/github/codeql/pull/7408 has been merged

PR #7408, opened 2021-12-15
JS: Use return value of trusted type policy callback as a sink

asgerf:
Treats a value being boxed in a trusted type as a sink. Concretely, we use the return value from the policy callback as a sink:
```js
trustedTypes.createPolicy('foo', { createHTML: x => /* sink */ })
```

Adding this since it was requested in https://github.com/github/codeql/issues/7336 - previously I have been hesitant to do this due lack of real-world examples of trusted types.

The only uses of trusted types I could find had flow around the policy invocation to the true sink,
```js
const policy = trustedTypes?.createPolicy('something', { createHTML: x => x });
...
function foo(data) {
  if (policy != null) { data = policy.createHTML(data); } // new sink here
  ...
  elm.innerHTML = data; // but also flow to the usual sink
}
```
For cases like this, we just get an alert for both sinks. But in principle, the policy invocation could occur far away from the traditional sink, in which case we get better coverage from the added sinks.

[Evaluation](https://github.com/github/codeql-dca-main/tree/data/asgerf/trusted-types-s__default__dist-compare/reports) looks fine

### DIFF
--- a/javascript/ql/lib/semmle/javascript/frameworks/TrustedTypes.qll
+++ b/javascript/ql/lib/semmle/javascript/frameworks/TrustedTypes.qll
@@ -25,6 +25,7 @@ module TrustedTypes {
   class PolicyCreation extends API::CallNode {
     PolicyCreation() { this = trustedTypesObj().getMember("createPolicy").getACall() }
 
+    /** Gets the function passed as the given option. */
     DataFlow::FunctionNode getPolicyCallback(string method) {
       // Require local callback to avoid potential call/return mismatch in the uses below
       result = getOptionArgument(1, method).getALocalSource()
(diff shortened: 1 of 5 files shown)


javascript/ql/lib/semmle/javascript/frameworks/TrustedTypes.qll
```diff
@@ -0,0 +1,68 @@
+/** Module for working with uses of the Trusted Types API. */
```
esbena, line 1:
```suggestion
/** Module for working with uses of the [Trusted Types API](https://w3c.github.io/webappsec-trusted-types/dist/spec/). */
```

esbena (dismissed):
LGTM. 
I'll let you decide if we need the spec URL.

javascript/ql/lib/semmle/javascript/frameworks/TrustedTypes.qll
asgerf: Added a link to the [MDN docs](https://developer.mozilla.org/en-US/docs/Web/API/Trusted_Types_API) instead as spec isn't that useful for our purposes.

esbena (dismissed):

esbena (approved):

PR #7410, opened 2021-12-15
Add QL for QL

erik-krogh:
This PR merges the QL for QL project into `github/codeql`.  
Included is a new workflow that runs QL for QL on all PRs.   

I moved the implementation into a `ql/` folder, and I updated the workflow files such that they should work well in this repository.  

[You can see the implementation here](https://github.com/github/codeql/tree/24b94e1c0fe709d5f04532fcf873ecfe21fc8b92).  
That is the commit right before the merge (the merge was clean, there were no conflicting files). 

You can inspect the results found by QL for QL by clicking on [the failing code-scanning check in the bottom of this PR](https://github.com/github/codeql/pull/7410/checks?check_run_id=4533567461). 

# How to review 

There are really only two things that need review:
 - [The readme](https://github.com/github/codeql/blob/24b94e1c0fe709d5f04532fcf873ecfe21fc8b92/ql/README.md). 
 - [The workflows](https://github.com/github/codeql/tree/24b94e1c0fe709d5f04532fcf873ecfe21fc8b92/.github/workflows). (@esbena :eyes: )
 
The rest is the unmodified QL for QL implementation. 

We don't have a QL for QL team, so I think everyone that is part of a CodeQL team is qualified to review (and approve) this PR.


.github/workflows/ql-for-ql-build.yml
```diff
@@ -0,0 +1,160 @@
...
+  queries:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - name: Cache queries
+        id: cache-queries
+        uses: actions/cache@v2
+        with:
+          path: ${{ runner.temp }}/query-pack.zip
+          key: ${{ runner.os }}-queries-${{ hashFiles('**/*.ql*') }}
```
esbena, line 22: `runner.os` should not have a say in how queries are compiled. Can we remove that part of the key?

.github/workflows/ql-for-ql-build.yml
```diff
@@ -0,0 +1,160 @@
...
+      - name: Find codeql
+        if: steps.cache-queries.outputs.cache-hit != 'true'
+        id: find-codeql
+        uses: github/codeql-action/init@esbena/ql
```
esbena, line 26: https://github.com/github/codeql-action/pull/779 should ideally get merged, but this works.

.github/workflows/ql-for-ql-build.yml
```diff
@@ -0,0 +1,160 @@
...
+        with:
+          languages: javascript # does not matter
```
esbena, line 28:
```suggestion
          languages: ql # does not matter
```
sligthly nicer

.github/workflows/ql-for-ql-build.yml
```diff
@@ -0,0 +1,160 @@
...
+          languages: ql
+          db-location: ${{ runner.temp }}/db
+      - name: Print debug info
+        run: |
+          echo "Pack"
+          ls ${{ runner.temp }}/pack
+          echo "Tools"
+          ls ${{ runner.temp }}/pack/tools
+          echo "index-files"
+          cat ${{ runner.temp }}/pack/tools/index-files.sh
```
esbena, line 150:
```suggestion
```
This was for debugging

ql/Cargo.toml
```diff
@@ -0,0 +1,6 @@
+[workspace]
+members = [
+    "extractor",
+    "generator",
+    "node-types",
```
aibaars, line 5: Have you tried using the components from `ruby` instead. The following might just work and allow us to delete the copied code.
```suggestion
    "../ruby/generator",
    "../ruby/node-types",
```

.github/workflows/ql-for-ql-build.yml
```diff
@@ -0,0 +1,160 @@
...
+
+      - name: Perform CodeQL Analysis
+        uses: github/codeql-action/analyze@esbena/ql
+
+      - name: Upload db
+        uses: actions/upload-artifact@v2
+        with:
+          name: db
+          path: ${{ runner.temp }}/db
+          retention-days: 1
```
esbena, line 160:
```suggestion
```
This was for debugging. 
The codeql-action/init has the `debug: true` flag if we need it. 
(I need to bump the ql branch to get the feature though)

.github/workflows/ql-for-ql-build.yml
```diff
@@ -0,0 +1,160 @@
...
+        run: |
+          echo "Pack"
+          ls ${{ runner.temp }}/pack
+          echo "Tools"
+          ls ${{ runner.temp }}/pack/tools
+          echo "index-files"
+          cat ${{ runner.temp }}/pack/tools/index-files.sh
+
+      - name: Perform CodeQL Analysis
+        uses: github/codeql-action/analyze@esbena/ql
```
RasmusWL, line 153: is using `@esbena/ql` a good long term solution?

.github/workflows/ql-for-ql-dataset_measure.yml
```diff
@@ -0,0 +1,80 @@
...
+  workflow_dispatch:
+
+jobs:
+  measure:
+    env:
+      CODEQL_THREADS: 4 # TODO: remove this once it's set by the CLI
+    strategy:
+      fail-fast: false
+      matrix:
+        repo: [github/codeql]
```
esbena, line 21:
```suggestion
        repo: 
          - github/codeql
          - github/codeql-go
```

.github/workflows/ql-for-ql-dataset_measure.yml
```diff
@@ -0,0 +1,80 @@
...
+    paths:
+      - ql/ql/src/ql.dbscheme
+  workflow_dispatch:
+
+jobs:
+  measure:
+    env:
+      CODEQL_THREADS: 4 # TODO: remove this once it's set by the CLI
+    strategy:
+      fail-fast: false
```
esbena, line 19:
```suggestion
```
I think it is fine to fail fast here.

RasmusWL (commented):
Would it be possible to configure [categories](https://docs.github.com/en/code-security/code-scanning/automatically-scanning-your-code-for-vulnerabilities-and-errors/configuring-code-scanning#configuring-a-category-for-the-analysis) for each of our CodeQL languages? (python, javascript, java, ...) that should allows us to filter down on just a single language when viewing Code Scanning alerts.

.github/workflows/ql-for-ql-dataset_measure.yml
```diff
@@ -0,0 +1,80 @@
...
+      matrix:
+        repo: [github/codeql]
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+
+      - name: Fetch CodeQL
```
esbena, line 26: See https://github.com/github/codeql/pull/7410/files#diff-c3c811f2f3a821590acf45fc281ba009b4886ec06d156f573019acf671e2d817R23 for how to obtain the codeql-cli that the codeql-action would be using. It is likely present on disk already, and it will by construction be the latest stable release.

.github/workflows/ql-for-ql-tests.yml
```diff
@@ -0,0 +1,42 @@
...
+
+env:
+  CARGO_TERM_COLOR: always
+
+jobs:
+  qltest:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - name: Fetch CodeQL
```
esbena, line 21: As above:

> See https://github.com/github/codeql/pull/7410/files#diff-c3c811f2f3a821590acf45fc281ba009b4886ec06d156f573019acf671e2d817R23 for how to obtain the codeql-cli that the codeql-action would be using. It is likely present on disk already, and it will by construction be the latest stable release.

.github/workflows/ql-for-ql-dataset_measure.yml
```diff
@@ -0,0 +1,80 @@
...
+    needs: measure
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/download-artifact@v2
+        with:
+          name: measurements
+          path: stats
+      - run: |
+          python -m pip install --user lxml
+          find stats -name 'stats.xml' | sort | xargs python ql/scripts/merge_stats.py --output ql/ql/src/ql.dbscheme.stats --normalise ql_tokeninfo
```
esbena, line 76: `-print0 | sort -z | xargs -0` is more standard

esbena (commented):
Some initial comments on the workflows. It may be best to try out the suggested changed outside of this PR to avoid the commit notification noise repeated failures could result in.

.github/workflows/ql-for-ql-build.yml
esbena: No. See https://github.com/github/codeql-action/pull/779. It was left as is due to priorities.

erik-krogh: For now it's the only way to get the `ql` language to work with `codeql-action`.  

https://github.com/github/codeql-action/pull/779 is the long-term solution (see Esbens comment further down).

erik-krogh: > Would it be possible to configure categories for each of our CodeQL languages? (python, javascript, java, ...) that should allows us to filter down on just a single language when viewing Code Scanning alerts.

I think that is possible, and it seems like a good idea, but I think it requires a significant change to the analyze workflow.  
I would like to wait with that until after this PR has been merged (or at the least after fixing Esbens comments).

ql/ql/src/codeql_ql/ast/internal/Builtins.qll
```diff
@@ -0,0 +1,96 @@
...
+  exists(string sig, string re | re = "(\\w+) (\\w+)\\.(\\w+)\\(([\\w, ]*)\\)" |
+    isBuiltinMember(sig) and
+    ret = sig.regexpCapture(re, 1) and
+    qual = sig.regexpCapture(re, 2) and
+    name = sig.regexpCapture(re, 3) and
+    args = sig.regexpCapture(re, 4)
+  )
+}
+
+module BuildinsConsistency {
```
nickrolfe, line 63: Typo? s/Buildins/Builtins

.github/workflows/ql-for-ql-build.yml
erik-krogh: Nope, doesn't work, I'm not sure why.

esbena: Can you link the failure here?

ql/Cargo.toml
erik-krogh: I have not.   
And I would like to not do that for now, and wait until we try to share code between the two extractors.

.github/workflows/ql-for-ql-build.yml
erik-krogh: I think it's this one: https://github.com/github/codeql-ql/runs/4534309639?check_suite_focus=true

MathiasVP: ql-code-scanning.qls 😂 if only

MathiasVP: Oh, wait. That file is actually there 🤔

esbena: The reason it does not work here, but later, is that the default query suite name does not exist yet. It exists once we do the proper setup in the context of our codeql-cli-with-ql.

esbena: So it is just a bootstrapping problem. Lets just stick to `javascript` then.

aibaars: We don't really need the `codeql-action` at all and simply use the `codeql` CLI directly. We shouldn't even need the CodeQL bundle, a plain CLI should be enough.

esbena: I'm abusing the codeql-action here to tell me where to find the codeql cli on disk.

erik-krogh:
I've fixed all the comments (some things are left for followup PRs).

.github/workflows/ql-for-ql-build.yml
aibaars: > I'm abusing the codeql-action here to tell me where to find the codeql cli on disk.

Yes, I figured that. I think we could rewrite the workflows to use `gh codeql` instead and reduce the amount of (ab)use of the codeql-action ;-) Best left for a follow-up pull request though.

erik-krogh: > Yes, I figured that. I think we could rewrite the workflows to use gh codeql instead and reduce the amount of (ab)use of the codeql-action ;-) Best left for a follow-up pull request though. 

`codeql-action` uses a CodeQL installation that is cached on all the workers, and therefore we don't have to download a fresh CodeQL installation each time.  
So yes, we are (ab)using `codeql-action`, but I don't think we are going to change that.

aibaars: @erik-krogh Is that to "speed things up". The CodeQL Action Init step takes about 9 seconds and fetching a fresh CodeQL CLI takes 13 seconds. Using CodeQL from the tool-cache is indeed faster, but not by much .

ql/ql/src/queries/style/docs/NonUSSpelling.ql
```diff
@@ -0,0 +1,38 @@
...
+ * @precision very-high
+ */
+
+import ql
+
+predicate non_us_word(string wrong, string right) {
+  exists(string s |
+    wrong = s.splitAt("/", 0) and
+    right = s.splitAt("/", 1) and
+    s = ["colour/color", "authorise/authorize", "analyse/analyze"]
```
intrigus-lgtm, line 17: Maybe add "modelling/modeling"

erik-krogh: Nice, that catches a whole lot of mistakes (a lot of which are mine :see_no_evil:).   
Could you open a PR with that addition after I've merged this one?  
We can use that new PR to fix the existing mistakes.

intrigus-lgtm: Will do.

esbena (commented):
The workflow parts look good enough to merge now. 

I do agree with Arthur that the codeql-cli download trick through codeql-action appears to just complicate matters with little tangible benefit. Perhaps the next iteration of that logic should use https://github.com/github/gh-codeql?


Some integration concerns:

- Could we exclude the known test directories from extraction to avoid lots of noisy alerts?
- Will we be overriding the results from the ordinary codeql-action run if that runs happen before this customized codeql action run (or vice versa)? I'm too fuzzy on how code-scanning behaves.

erik-krogh:
> Could we exclude the known test directories from extraction to avoid lots of noisy alerts? 

Maybe... I think I'll look into that in a followup PR.  
(I'll look into it when I address [this comment](https://github.com/github/codeql/pull/7410#pullrequestreview-832741984) by Rasmus). 

> Will we be overriding the results from the ordinary codeql-action run if that runs happen before this customized codeql action run (or vice versa)? I'm too fuzzy on how code-scanning behaves. 

That is not an issue.  
code-scanning is fine with multiple workflows uploading results.  
Code-scanning uses [categories](https://docs.github.com/en/code-security/code-scanning/automatically-scanning-your-code-for-vulnerabilities-and-errors/configuring-code-scanning#configuring-a-category-for-the-analysis) to split out results of multiple workflows

.github/workflows/ql-for-ql-build.yml
```diff
@@ -0,0 +1,146 @@
...
+  queries:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - name: Cache queries
+        id: cache-queries
+        uses: actions/cache@v2
+        with:
+          path: ${{ runner.temp }}/query-pack.zip
+          key: queries-${{ hashFiles('**/*.ql*') }}
```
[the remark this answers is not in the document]
erik-krogh: I've added the version (from `codeql-version`) and the hash of the dbscheme files to the key.

esbena (approved):

ql/Cargo.toml
erik-krogh: I tried to share them naively, but I run into this issue: https://github.com/rust-lang/cargo/issues/5042  
(I got the `is a member of the wrong workspace` error).   

So it takes a bit more effort than just changing the Cargo.toml.  
I've leave it for now.

PR #7416, opened 2021-12-15
Remove experimental tag from non-ATM queries

sampart:
When https://github.com/github/code-scanning/issues/4291 ships, we will be using the `experimental` tag on CodeQL queries as a signifier for Adaptive Threat Modeling boosted queries, and showing extra UI around them like this:

image

We therefore don't want to use this tag for queries that aren't to do with ATM, hence this PR.

@henrymercer @AlonaHlobina


tausbn (approved):
Looks good to me. I'll try to remember that external contributions (that go into the `experimental` folder) should not have this tag.

sampart:
Sounds like you need a QL-for-QL query so you don't have to remember 😜

PR #7417, opened 2021-12-15
Java: Start running telemetry queries on Code Scanning

henrymercer:
This PR incorporates the telemetry queries for Java into the summary metrics framework.  Specifically we use the support for summary metrics with messages that was introduced in CodeQL CLI v2.8.0.

This has a few effects:

1. We'll start running these queries on Code Scanning.
2. We will start displaying a summary output of these queries in the metrics summary table produced on Code Scanning and in the CLI (see example below).
3. We can ingest the results of these queries as part of generic internal pipelines that operate on summary metrics.

<details>
<summary>Example metrics summary table additions</summary>

```
Analysis produced the following metric data:

|                          Metric                          |      Value     |
+----------------------------------------------------------+----------------+
| Supported sinks in external libraries                    |  (574 results) |
| Usage of unsupported APIs coming from external libraries | (3949 results) |
| Supported sources in external libraries                  |   (15 results) |
| External libraries                                       |   (75 results) |
| Supported sinks in external libraries                    |   (22 results) |
```

</details>

/cc @bmuskalla @turbo @yo-h

### DIFF
--- a/java/ql/src/Telemetry/SupportedExternalTaint.ql
+++ b/java/ql/src/Telemetry/SupportedExternalTaint.ql
@@ -1,9 +1,9 @@
 /**
  * @name Supported sinks in external libraries
  * @description A list of 3rd party APIs detected as sinks. Excludes test and generated code.
- * @id java/telemetry/supported-external-api-taint
  * @kind metric
- * @metricType callable
+ * @tags summary
+ * @id java/telemetry/supported-external-api-taint
  */
 
 import java
--- a/java/ql/src/Telemetry/SupportedExternalTaint.ql
+++ b/java/ql/src/Telemetry/SupportedExternalTaint.ql
@@ -1,6 +1,6 @@
 /**
- * @name Supported sinks in external libraries
- * @description A list of 3rd party APIs detected as sinks. Excludes test and generated code.
+ * @name Supported flow steps in external libraries
+ * @description A list of 3rd party APIs detected as flow steps. Excludes test and generated code.
  * @kind metric
  * @tags summary
  * @id java/telemetry/supported-external-api-taint
--- a/java/ql/src/Telemetry/ExternalLibraryUsage.ql
+++ b/java/ql/src/Telemetry/ExternalLibraryUsage.ql
@@ -2,7 +2,7 @@
  * @name External libraries
  * @description A list of external libraries used in the code
  * @kind metric
- * @metricType callable
+ * @tags summary
  * @id java/telemetry/external-libs
  */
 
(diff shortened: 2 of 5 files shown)


java/ql/src/Telemetry/SupportedExternalTaint.ql
```diff
@@ -1,9 +1,9 @@
 /**
  * @name Supported sinks in external libraries
```
yo-h, line 2:
```suggestion
 * @name Supported flow steps in external libraries
```

java/ql/src/Telemetry/SupportedExternalTaint.ql
```diff
@@ -1,9 +1,9 @@
...
  * @description A list of 3rd party APIs detected as sinks. Excludes test and generated code.
```
yo-h, line 3:
```suggestion
 * @description A list of 3rd party APIs detected as flow steps. Excludes test and generated code.
```

yo-h (commented):
Not caused by this PR, but it seems worth fixing what looks like a copy/paste error to avoid the following confusion (taken from the example in the PR description).

> | Supported sinks in external libraries                    |  (574 results) |
> | Supported sinks in external libraries                    |   (22 results) |

bmuskalla:
Thanks for fixing up the tags (I had a draft PR but never merged it, sorry) 

@henrymercer - a more general question: Do we run all `summary` queries alongside every other invocation of codeql for regular users?
Because that makes sense for queries that are looking at the database in question (e.g. `UnsupportedExternalAPIs.ql`) but is not necessary for database-independant queries (e.g. the new `FrameworkCoverageMetrics.ql`). WDYT?

henrymercer:
> @henrymercer - a more general question: Do we run all `summary` queries alongside every other invocation of codeql for regular users?
> Because that makes sense for queries that are looking at the database in question (e.g. `UnsupportedExternalAPIs.ql`) but is not necessary for database-independant queries (e.g. the new `FrameworkCoverageMetrics.ql`). WDYT?

The queries that we'll run on Code Scanning are defined within https://github.com/github/codeql/tree/main/java/ql/src/codeql-suites.  I can't find `FrameworkCoverageMetrics.ql`, but I agree we likely want to avoid running queries like these that are targetted at query writers in Code Scanning.  We can do this for instance by adding another exclusion rule to https://github.com/github/codeql/blob/main/misc/suite-helpers/code-scanning-selectors.yml like the one we have for `experimental`.

henrymercer:
@github/codeql-java Would this PR require a changenote?

yo-h (approved):
> Would this PR require a changenote?

We are still evaluating the output of these queries, so I would merge without a change note for now. Unless @turbo knows of a reason to advertise more widely at this time?

PR #7420, opened 2021-12-16
Python: Add SSRF queries

RasmusWL:
I've added 2 queries:

- one that detects full SSRF, where an attacker can control the full URL, which is always bad
- and one for partial SSRF, where an attacker can only control parts of an URL (such as the path, query parameters, or fragment), which is not a big problem in many cases (but could still be exploitable)

full SSRF should run by default, and partial SSRF should not (but having the query included makes it easy to run). I got inspired by this setup from Java where they have a [precise](https://github.com/github/codeql/blob/b53e3499cbbc517c639d876ebbbd1b4cadc3d52b/java/ql/src/Security/CWE/CWE-327/BrokenCryptoAlgorithm.ql#L7) and [imprecise](https://github.com/github/codeql/blob/b53e3499cbbc517c639d876ebbbd1b4cadc3d52b/java/ql/src/Security/CWE/CWE-327/MaybeBrokenCryptoAlgorithm.ql#L7) version of the same query.

### Current status

Most of the query work/library modeling is done (although we could always add support for more libraries). Still need to do:

- [x] write qhelp 
- [x] write change-note
- [x] Polish sanitizer for full SSRF query so we're able to detect `"https://" + user_input` is in fact controlling the full URL.
- [x] verify FP rates from run across many repos
- [x] verify performance looks ok
- [ ] make some updates to the Ruby code so we're better aligned (but that can wait until after this PR is merged I think)

### Commits

Some of the commits changes the concepts that was added in the very first commit. I've kept things this way so it could help to illustrate _why_ I wanted to diverge from the Ruby code.

### What is SSRF even?

See https://portswigger.net/web-security/ssrf if you need a refresher on SSRF :blush:


python/ql/test/library-tests/frameworks/requests/taint_test.py
```diff
@@ -0,0 +1,54 @@
...
+    # response from a request to a user-controlled URL should be considered
+    # user-controlled as well.
+    resp = requests.get(url) # $ clientRequestUrlPart=url
+
+    requests.Response
+    requests.models.Response
+
+    ensure_tainted(
+        # see https://docs.python-requests.org/en/latest/api/#requests.Response
+        resp, # $ tainted
```
yoff, line 21:
```suggestion
    ensure_tainted(
        url, # $ tainted
        # see https://docs.python-requests.org/en/latest/api/#requests.Response
        resp, # $ tainted
```
I would add the url in here as a sanity check

python/ql/lib/semmle/python/frameworks/Requests.qll
```diff
@@ -0,0 +1,171 @@
...
+    string methodName;
+
+    OutgoingRequestCall() {
+      methodName in [HTTP::httpVerbLower(), "request"] and
+      (
+        this = API::moduleImport("requests").getMember(methodName).getACall()
+        or
+        exists(API::Node moduleExporting, API::Node sessionInstance |
+          moduleExporting in [
+              API::moduleImport("requests"), //
```
intrigus-lgtm, line 37: Intentionally empty comment (for formatting?)?

RasmusWL: indeed :+1:

python/ql/test/query-tests/Security/CWE-918-ServerSideRequestForgery/full_partial_test.py
```diff
@@ -22,6 +22,13 @@ def full_ssrf():
     url = "https://" + user_input + "/foo?key=" + query_val
     requests.get(url) # NOT OK -- user has full control
 
+# currently it's not possible to sanitize a step, so due to use-use flow, if we sanitize
+# a value that is used later on, it will NOT be tainted later on... so we need to make
+# separate tests for each of the vairant cases, so show that we handle all of them.
```
yoff, line 27:
```suggestion
# currently it's not possible to sanitize a step, so due to use-use flow, if we sanitize
# a value that is used later on, it will NOT be tainted later on... so we need to make
# separate tests for each of the variant cases, to show that we handle all of them.
```
I am not sure what you mean by it not being possible to sanitize a step. It is true that `isBarrier` removes a node entirely from the data-flow graph, but `isBarrierIn` and `isBarrierOut` both remove edges.

On use-use flow: It is supposed to respect the control flow graph and thus dominance. Do you have a case of sanitising a value and still wanting it to be tainted in a place (only) reachable by use-use flow from the sanitized value?

RasmusWL: You can add a taint-flow step `fromNode -> toNode`, but when adding a sanitizer, you can only do so on `fromNode` or `toNode` (either all edges in and out, or just the edges in or out). So if you have multiple taint edges going out from `fromNode`, you can choose to only sanitize some of them.

Highly contrived example:

```py
url = "https://example.com/" + user_input # <-- sanitizer is applied to `user_input` for full-ssrf
requests.get(url)

requests.get(user_input) # <-- since user_input has been sanitized, and we have use-use flow, there will be no alert here
```

This caused some problems since I added support for fstrings before % formatting -- but I realize that I failed to capture this important bit in the commit history.

RasmusWL:
```suggestion
# taint-steps are added as `fromNode -> toNode`, but when adding a sanitizer it's
# currently only possible to so on either `fromNode` or `toNode` (either all edges in
# and out, or just the edges in or out). The sanitizers for full URL control is applied
# on the `fromNode`, since for `"https://{}/{}".format(user_input1, user_input2)` there
# is still a valid taint-step for `user_input1`. When coupled with use-use flow, this
# means that later uses of a sanitized value will no longer be tainted, so
# `requests.get(user_input2)` would no longer give an alert. To overcome this problem,
# we split these tests into multiple functions, so we do not get this use-use flow, and
# therefore know we are able to see where the sanitizers are applied.
```

RasmusWL: > Although I do not understand why, in your contrived example, you sanitize `user_input` (which is not changed by the code) rather than the `+`-expression.

To be able to handle `requests.get(user_input + "/foo")` which would still be considered full SSRF. (if all `+`-expression are sanitizers, then we would not be able to produce a full SSRF results here).

This is what I aimed at illustrating with the `"https://{}/{}".format(user_input1, user_input2)` example in the suggested comment. I've tried to expand that part a bit.


```suggestion
# taint-steps are added as `fromNode -> toNode`, but when adding a sanitizer it's
# currently only possible to so on either `fromNode` or `toNode` (either all edges in
# and out, or just the edges in or out). The sanitizers for full URL control is applied
# on the `fromNode`, since for `"https://{}/{}".format(user_input1, user_input2)` there
# is still a valid taint-step for `user_input1` -- if we made `toNode` a sanitizer that
# would also remove this flow that we actually want. When coupled with use-use flow,
# this means that later uses of a sanitized value will no longer be tainted, so
# `requests.get(user_input2)` would no longer give an alert. To overcome this problem,
# we split these tests into multiple functions, so we do not get this use-use flow, and
# therefore know we are able to see where the sanitizers are applied.
```

yoff (approved):
LGTM - thanks for the offline explanations

yoff (approved):
Lgtm

RasmusWL:
:shipit:

PR #7423, opened 2021-12-16
Java: Cover CVE-2021-45046 in the Log4jJndiInjection query

atorralba:
This PR adds sinks and taint steps for covering the new attack vectors discovered in CVE-2021-45046.

Tests were added too.


smowton:
What's going on with the removed model lines? Most seem to end with a `Throwable` parameter, though some Throwables remain in the revised models.

I'd strongly recommend using an inline-expectation test rather than committing a 5,000 line expectation.

atorralba:
> What's going on with the removed model lines? Most seem to end with a `Throwable` parameter, though some Throwables remain in the revised models.

No lines were removed, it's just that I renamed the sinks from `logging` to `log4j` and that changes the formatting. Should I make a first commit only with that change, so that it's easier to see the new additions? (the new things are at the bottom of the models in any case)

> I'd strongly recommend using an inline-expectation test rather than committing a 5,000 line expectation.

Yes, that was my intention, but the problem is that we can't reuse anything from the query in the test because it's in experimental, and I didn't want to duplicate everything in the test 😞

smowton:
> No lines were removed, it's just that I renamed the sinks from `logging` to `log4j` and that changes the formatting. Should I make a first commit only with that change, so that it's easier to see the new additions? (the new things are at the bottom of the models in any case)

Are you sure? The first block for example adds 5 less lines than it removes?

atorralba:
> Are you sure? The first block for example adds 5 less lines than it removes?

Note that, e.g. line 57 contains two rows in one line because of the change in formatting: https://github.com/github/codeql/pull/7423/files#diff-c87c9c5a74ba94cb105e161c4d11190f6ea16f301da6e98c2779209abb3196ccR57

smowton:
Doh right, so it does! I guess we can take care of fixing the test on promotion if that ever comes...

java/ql/src/experimental/Security/CWE/CWE-020/Log4jJndiInjection.ql
```diff
@@ -19,140 +19,146 @@ import semmle.code.java.dataflow.FlowSources
...
+        "org.apache.logging.log4j;LogBuilder;true;log;(String,Object,Object,Object,Object);;Argument[0..4];log4j",
+        "org.apache.logging.log4j;LogBuilder;true;log;(String,Object,Object,Object,Object,Object);;Argument[0..5];log4j",
+        "org.apache.logging.log4j;LogBuilder;true;log;(String,Object,Object,Object,Object,Object,Object);;Argument[0..6];log4j",
+        "org.apache.logging.log4j;LogBuilder;true;log;(String,Object,Object,Object,Object,Object,Object,Object);;Argument[0..7];log4j",
+        "org.apache.logging.log4j;LogBuilder;true;log;(String,Object,Object,Object,Object,Object,Object,Object,Object);;Argument[0..8];log4j",
+        "org.apache.logging.log4j;LogBuilder;true;log;(String,Object,Object,Object,Object,Object,Object,Object,Object,Object);;Argument[0..9];log4j",
+        "org.apache.logging.log4j;LogBuilder;true;log;(String,Object,Object,Object,Object,Object,Object,Object,Object,Object,Object);;Argument[0..10];log4j",
+        "org.apache.logging.log4j;LogBuilder;true;log;(String,Supplier[]);;Argument[0..1];log4j",
+        "org.apache.logging.log4j;LogBuilder;true;log;(Supplier);;Argument[0];log4j",
+        // org.apache.logging.log4j.ThreadContet
```
Marcono1234, line 139: There are also [`CloseableThreadContext`](https://logging.apache.org/log4j/2.x/log4j-api/apidocs/org/apache/logging/log4j/CloseableThreadContext.html) and [`CloseableThreadContext.Instance`](https://logging.apache.org/log4j/2.x/log4j-api/apidocs/org/apache/logging/log4j/CloseableThreadContext.Instance.html), in case you want to cover them as well.

atorralba: Good catch @Marcono1234, thanks. See #7435.

PR #7424, opened 2021-12-16
C#: Flow summaries in CSV format.

michaelnebel:
In this PR we convert most of the remaining flow summaries to MaD.
The targeted summaries are for the following classes, interfaces - including their subtypes and relevant extensions methods.
* System.Collections.IEnumerable
* System.Collections[.Generic].IDictionary
* System.Collections[.Generic].IList
* System.Collections[.Generic].ICollection

Besides the conversion, some minor improvements has also been introduced as there didn't exist a flow summary for the generic IEnumerable<>.GetEnumerator method.

As the existing flow summary implementation overlapped quite heavily, the strategy for doing the conversion was as follows: Remove a QL class and/or predicate and introduce "at least" the corresponding flow summaries in CSV format.

Some suggestions for follow up improvements and re-factors
* Instead of defining flow summaries for a single C# class/interface in a QL class, we can consider to only define a single QL class pr. namespace. This will reduce the amount of boilerplate code.
* Move the AccesPath implementation from LibraryTypeDataFlow.qll into a separate file, as this will allow us to define synthetic fields without referencing LibraryTypeDataFlow (and move the corresponding synthethic field definitions into the appropriate files)
* Rewrite ForeachCapture test to get rid of the callableFlow summary for Add.

Furthermore, we need to consider what (if anything) should be done about the "clearsContent" type of flow summaries that already exist in LibraryTypeDataFlow.qll. Do we need to introduce syntax for this in MaD framework?


csharp/ql/lib/semmle/code/csharp/frameworks/system/xml/Serialization.qll
```diff
@@ -0,0 +1,62 @@
+/** Provides definitions related to the `System.Xml.Schema` namespace. */
```
hvitved, line 1: typo

hvitved (changes requested):
Thanks for the heroic work on converting all these summaries.

csharp/ql/lib/semmle/code/csharp/frameworks/system/xml/Serialization.qll
michaelnebel: Nice catch! Will fix.

hvitved (approved):

PR #7440, opened 2021-12-17
Ruby: Include StringComponents in the CFG

hvitved:
Code review suggestions for https://github.com/github/codeql/pull/7334.


ruby/ql/lib/codeql/ruby/controlflow/internal/ControlFlowGraphImpl.qll
```diff
@@ -1365,6 +1365,10 @@ module Trees {
     final override ControlFlowTree getChildElement(int i) { result = this.getComponent(i) }
   }
 
+  private class StringComponentComponentTree extends LeafTree, StringComponent {
+    StringComponentComponentTree() { not this instanceof StringInterpolationComponent }
```
nickrolfe, line 1369: Are interpolations included in the CFG elsewhere?

hmac:
Interesting! I went down this route at first, but decided against it because it seemed wasteful to bloat the CFG with nodes that play no part in control flow.

hvitved:
> Interesting! I went down this route at first, but decided against it because it seemed wasteful to bloat the CFG with nodes that play no part in control flow.

We already include stuff in the CFG that does not really get evaluated, e.g. the LHS of a local variable assignment, so I don' think this is too bad.

ruby/ql/lib/codeql/ruby/controlflow/internal/ControlFlowGraphImpl.qll
hvitved: I don't think so, but perhaps they should. But I would rather that we do that in a separate PR, since adding them naively makes the CFG inconsistent (some nodes will have multiple successors).

nickrolfe (approved):
LGTM

ruby/ql/lib/codeql/ruby/controlflow/internal/ControlFlowGraphImpl.qll
hmac: Why is this called `StringComponentComponentTree` rather than `StringComponentTree`?

ruby/ql/lib/codeql/ruby/controlflow/CfgNodes.qll
```diff
@@ -147,10 +147,10 @@ abstract private class ExprChildMapping extends Expr {
    * Holds if `child` is a (possibly nested) child of this expression
    * for which we would like to find a matching CFG child.
    */
-  abstract predicate relevantChild(Expr child);
+  abstract predicate relevantChild(AstNode child);
```
hmac, line 150: Should this class now  be called `AstChildMapping` instead of `ExprChildMapping`?

ruby/ql/test/library-tests/ast/ValueText.expected
```diff
@@ -602,7 +602,6 @@
 | modules/modules.rb:55:8:55:30 | "module Foo::Bar again" | module Foo::Bar again |
 | modules/modules.rb:56:17:56:17 | 4 | 4 |
 | modules/toplevel.rb:1:6:1:12 | "world" | world |
-| modules/toplevel.rb:3:12:3:16 | "!!!" | !!! |
```
hmac, line 604: I guess this has disappeared because the `puts` call in `END { puts "!!!" }` is unreachable?

hmac:
This is a nice simplification, thanks for doing it!

ruby/ql/lib/codeql/ruby/controlflow/internal/ControlFlowGraphImpl.qll
hvitved: That's a typo.

ruby/ql/lib/codeql/ruby/controlflow/CfgNodes.qll
hvitved: I think it is OK to leave as is, since the class extends `Expr`.

ruby/ql/test/library-tests/ast/ValueText.expected
hvitved: yes.

PR #7441, opened 2021-12-17
QL-for-QL: Followup changes

erik-krogh:
Changes: 
- Updated the .stats file from [this workflow](https://github.com/github/codeql/actions/runs/1589511679).  
- Fixed some bad join orders that appeared from the above. 
- Make the change-note checker ignore QL-for-QL
- Fix the QL tests, such that they now also run when content inside a nested folder has changed. 
- Cache the extractor in the analysis workflow. 
- Updated to the latest `codeql-action` (with a change that allows the `ql` language to be used). 
- Split out analysis of each CodeQL language. 
  This means that each folder is now it's own code-scanning category, which will hopefully make navigating results easier. 
  (and turn-around time is now ~3 minutes). 
  That's also why code-scanning is confused about this PR: it says there are lots of new results (there aren't). 
- The above required fixing the autobuilder. That fix is temporary until [we figure out what should happen](https://github.com/github/codeql/pull/7444).  
- All the errors and warnings that QL-for-QL found in QL-for-QL have been fixed. 

The turn-around time for the PR check is now ~3 minutes.


.github/workflows/ql-for-ql-build.yml
```diff
@@ -146,12 +166,26 @@ jobs:
 
       - name: Checkout repository
         uses: actions/checkout@v2
+      - name: Create CodeQL config file
+        run: |
+          echo "paths:" > ${CONF}
+          echo "  - ${{ matrix.folder }}" >> ${CONF}
```
esbena, line 172:
```suggestion
          echo "  - ${FOLDER}" >> ${CONF}
```
+ env change below

(this replaces dynamically generated bash code with simple variable substitution which is much safer in general)

.github/workflows/ql-for-ql-build.yml
```diff
@@ -146,12 +166,26 @@ jobs:
...
+          echo "paths-ignore:" >> ${CONF}
+          echo "  - ql/ql/test" >> ${CONF}
+          echo "Config file: "
+          cat ${CONF}
+        env: 
+          CONF: ./ql-for-ql-config.yml
```
esbena, line 178:
```suggestion
          CONF: ./ql-for-ql-config.yml
          FOLDER: ${{ matrix.folder }}
```

esbena (commented):
LGTM with a (for now) innocent security nit. But I can't say anything about the autobuilder/rust changes.

erik-krogh:
> LGTM with a (for now) innocent security nit. But I can't say anything about the autobuilder/rust changes. 

The autobuilder change is temporary, but it works in the one case where it needs to work (the QL-for-QL PR check). 
I think `codeql database index-files` and/or the Ruby autobuilder need to be fixed.  
When that's done, then I'll revisit the QL-for-QL autobuilder.

esbena (approved):

PR #7443, opened 2021-12-17
QL4QL: catch behaviour/behavior in ql/non-us-spelling

nickrolfe:


owen-mc (approved):
I don't understand the Code scanning warning, but this PR is obviously good.

erik-krogh:
> I don't understand the Code scanning warning, but this PR is obviously good.

If you're talking about the `Consistency predicate that should be empty`-warning, then it's caused by the QL-for-QL parser being confused about trailing commas in set literals.  
We know about it, and it will get fixed (after Taus has reworked the Tree-sitter grammars).

erik-krogh (approved):
JS and QL-for-QL :+1:

hvitved (approved):
C# 👍

MathiasVP (approved):
C++ 👍

nickrolfe:
I haven't had a review from the Python team, but this seems like such an uncontroversial change that I'm sure they won't mind.

PR #7444, opened 2021-12-17
QL for QL: add autobuilder that respects LGTM_INDEX_FILTERS

nickrolfe:
A fairly trivial modification of the Ruby autobuilder.

Closes https://github.com/github/codeql-ql/issues/141


erik-krogh (approved):
LGTM :tada:

erik-krogh:
Something seems to be wrong.  
When I try to use `LGTM_INDEX_FILTERS` the extractor tries to extract every file in the folder, not just the files with the right extension: https://github.com/github/codeql/runs/4562952843?check_suite_focus=true  
@nickrolfe ?

nickrolfe:
I see you have this config:

```
paths:
  - ruby
paths-ignore:
  - ql/ql/test
```

And I guess that gets translated somehow to something like `LGTM_INDEX_FILTERS=include:ruby/**\nexclude:ql/ql/test/**`.

The autobuilder will then run

```
codeql database index-files --include-extension=.ql --include-extension=.qll --include-extension=.dbscheme --include=**/qlpack.yml --size-limit=5m --language=ql --working-dir=. $CODEQL_EXTRACTOR_QL_WIP_DATABASE --include=ruby/** --exclude=ql/ql/test/**
```

And I suppose the `--include=ruby/**` argument tells `index-files` to include everything in the `ruby` directory, ignoring the extension filter.

@aibaars do we need to make the Ruby/QL autobuilder smarter at expanding the `include:` options, or is there a better way of handling this?

erik-krogh:
Could it be `codeql database index-files` bug.  
It seems like `--include-extension` is pretty much ignored if `--include` is used?

erik-krogh:
I did a little experiment.  
And apparently `codeql database index-files --include-extension=.ql --include=somefolder` means to include every file `.ql` reachable from the current working folder, and every file inside `somefolder`.   

So the autobuilder is broken both for QL-for-QL and Ruby right now.

PR #7446, opened 2021-12-18
[Ruby] Bugfix: ConstantWriteAccess::getQualifiedName() returns wrong value in some cases

jeffgran:
## Description of the bug

given this code:

```ruby
module Foo
  class Bar
  end
end
```
the `ConstantWriteAccess` node for the `Bar` class declaration *correctly* returns `Foo::Bar` from `getQualifiedName()`

However, given this code:

```ruby
class Foo::Bar
end
```
the `ConstantWriteAccess` node for the `Bar` class declaration *incorrectly* returns `Bar` from `getQualifiedName()`

---

I noticed that the test cases for this predicate exist and it covers this case, but the expected values were incorrect. So first I fixed the expectation, in https://github.com/github/codeql/commit/7315c542d7587a1487f49e8cd80564b46065ab5c . Then, I fixed the bug in https://github.com/github/codeql/commit/5bb1795559bf671a78b625eba67291a85dc6504d by adding a recursive `getFullName` predicate, similar to and modeled after the existing `getQualifiedName()` predicate.

This may not be the best solution. I suspect that this should apply to more than just `ConstantWriteAccess`. It probably needs to apply to `ConstantReadAccess` too. I think maybe the whole `getQualifiedName()` should be moved to `ConstantAccess`.  It will be very important in Ruby to be able to get the fully qualified constant name for any constant regardless of whether it's a read or write access.

For example, in this code:

```ruby
module ModuleA
  @@myvar = ModuleB::MAX_SIZE
end
```

`ModuleB::MAX_SIZE` is a `ConstantReadAccess`, but there is currently no way to get the fully qualified name of that constant. We would want to have a predicate that will return `ModuleA::ModuleB::MAX_SIZE`, like `getQualifiedName()` does for `ConstantWriteAccess`


alexrford:
Hi @jeffgran, thanks for identifying this problem and providing a bugfix. Looking into the results of this predicate on some other test cases, I've observed that it has incorrect results in other cases involving nested modules (apologies for the wall of text below). For example, in the following case:

```rb
module Mod1 end

module Mod2
   module Mod1::Mod3 end
end

module Mod4
  include Mod1
  module Mod3::Mod5 end
end
```

we incorrectly assume that `Mod1::Mod3` is nested inside `Mod2` because it's declaration is syntactically within `module Mod2 ... end`, but as `Mod1` is declared outside of `Mod2` (and there is no `Mod1` declared as a nested module of `Mod2`), this is not the case and `Mod3` should have qualified name `Mod1::Mod3`.

Separately, `Mod5` can be referred to by either `Mod1::Mod3::Mod5` or `Mod4::Mod3::Mod5`, and neither is necessarily "more qualified" than the other.

Another interesting case is a pair of files like:

```rb
# a.rb
module Foo
  module Bar end
end
```

```rb
# b.rb
module Bar end
module Foo
  module Bar::Baz end
end
```

and a file that `require`s them:
```rb
# main.rb
require './a'
require './b'

p !(defined? Foo::Bar::Baz).nil? # Baz has qualified name Foo::Bar::Baz
p !(defined? Bar::Baz).nil?      # Baz has qualified name Bar::Baz
```

will output:
```
true
false
```
as `a.rb` declares `Bar` as a module nested in `Foo` before the declaration of `module Bar::Baz` syntactically within `module Foo`. However, reversing the requires (`require './b'; require './a'`) will instead cause `Baz` to be declared in the unnested `Bar` module, as this declaration occurs before the nested `Foo::Bar` module is declared. So in this second case the output is flipped:
```
false
true
```

I think this suggests that picking a single qualified name isn't possible in general, either due to there being multiple valid qualified names, or because we don't know the qualified name until runtime (as in the case of the last example).

---

We have an internal predicate, [`string resolveConstantWriteAccess(ConstantWriteAccess c)`](https://github.com/github/codeql/blob/aa92fe8c90d0209600e597d99d84a959898a309e/ruby/ql/lib/codeql/ruby/ast/internal/Module.qll#L367) that I believe handles these cases and can return multiple results - we probably should have defined `getQualifiedName()` effectively as an alias for this internal predicate when we first added it.

My suggestion for this PR would be to add a member predicate:
```
  string getAQualifiedName() {
    result = resolveConstantWriteAccess(this)
  }
```
to `ConstantWriteAccess` to make use of this implementation, and to redefine `getQualifiedName()` as a deprecated functional alias for it:
```
  deprecated string getQualifiedName() { result = this.getAQualifiedName() }
```
(the deprecation and renaming is just to comply with naming conventions for predicates that are expected to return multiple results).

Thanks again for picking up on this - hope the suggestion above makes sense.

---

Regarding generalizing this to `ConstantReadAccess`es as well - I agree that this seems useful and we should support it. As with `ConstantWriteAccess`, there are some internal predicates in `internal/Module.qll` that deal with similar complexities. I think this change would be better in a separate follow-up PR if that's okay, as there are a few different predicates in `internal/Module.qll` that deal with `ConstantReadAccess`es and I'm not sure yet which would make the most sense for the corresponding public predicate.

jeffgran:
@alexrford thank you for the detailed response! This all makes perfect sense to me. I made the changes you suggested, added some more documentation about this, and added a few more test cases based on your explanations.

> Separately, `Mod5` can be referred to by either `Mod1::Mod3::Mod5` or `Mod4::Mod3::Mod5`, and neither is necessarily "more qualified" than the other.

Wow, in all my years of ruby I have never come across this -- I didn't know that including a module in another module means it can be used as the "namespace" for the constants defined in the parent module.

^ I added that test case and noticed that even using `resolveConstantWriteAccess` doesn't seem to pick that up. I suppose maybe it doesn't take this `include X` into account?

ruby/ql/lib/codeql/ruby/ast/Constant.qll
```diff
@@ -184,16 +184,34 @@ class ConstantWriteAccess extends ConstantAccess {
...
-    or
-    /* base case - there's no parent module */
-    not exists(ConstantWriteAccess parent | parent = this.getEnclosingModule()) and
-    result = this.getName()
+  string getAQualifiedName() {
+    result = resolveConstantWriteAccess(this)
+  }
+
+  /**
+   * gets a qualified name for this constant. Deprecated in favor of
```
aibaars, line 210:
```suggestion
   * Gets a qualified name for this constant. Deprecated in favor of
```

ruby/ql/lib/codeql/ruby/ast/Constant.qll
```diff
@@ -184,16 +184,34 @@ class ConstantWriteAccess extends ConstantAccess {
    *
    * the constant `Baz` has the fully qualified name `Foo::Bar::Baz`, and
    * `CONST_A` has the fully qualified name `Foo::CONST_A`.
+   *
```
aibaars, line 187: In the above `Gets the fully qualified name` -> `Gets a fully qualified name`

ruby/ql/test/library-tests/ast/constants/constants.rb
```diff
@@ -40,3 +40,22 @@ module ModuleA::ModuleC
 
 puts GREETING
 puts ::GREETING
+
+module ModuleA::ModuleB
+  class ClassB < Base
+  end
```
aibaars, line 46: Let's add a few constant definitions to the classes and modules defined in the examples.

```
module ModuleA::ModuleB
  class ClassB < Base
    X = 42
  end
end
``` 

Similar for the others.

aibaars (commented):
Thank for fixing this bug.

aibaars:
Please also fix the following warning:
```
WARNING: Predicate getQualifiedName has been deprecated and may be removed in future
(/home/runner/work/codeql/codeql/ruby/ql/lib/codeql/ruby/frameworks/ActiveRecord.qll:281,33-49)
```
If I'm not mistaken uses of the deprecated `getQualifiedName` should be replaced in
```
ruby/ql/lib/codeql/ruby/frameworks/ActiveRecord.qll:    resolveConstant(recv) = cls.getQualifiedName() and
ruby/ql/src/queries/analysis/Definitions.ql:      w.getQualifiedName() = fqn and l = w.getLocation()
```

aibaars:
The test data is a bit out of date, see the failed `qltest` job. To generate new test data you can run `codeql test run --learn  ql/test` in the `codeql/ruby` folder.

ruby/ql/test/library-tests/ast/constants/constants.expected
```diff
@@ -65,6 +79,17 @@ constantWriteAccessQualifiedName
...
+| constants.rb:45:3:46:5 | ClassB | ModuleA::ModuleB::ClassB |
+| constants.rb:49:1:52:3 | ModuleA | ModuleA |
+| constants.rb:50:3:51:5 | ClassB | ModuleA::ModuleB::ClassB |
+| constants.rb:50:3:51:5 | ClassB | ModuleB::ClassB |
+| constants.rb:54:1:56:3 | Mod1 | Mod1 |
+| constants.rb:55:3:55:17 | Mod3 | Mod1::Mod3 |
+| constants.rb:58:1:61:3 | Mod4 | Mod4 |
+| constants.rb:60:3:60:23 | Mod5 | Mod1::Mod3::Mod5 |
+| constants.rb:60:3:60:23 | Mod5 | Mod4::Mod3::Mod5 |
+| constants.rb:60:3:60:23 | Mod5 | Mod3::Mod5 |
```
aibaars, line 95: I think this should actually:
```suggestion
| constants.rb:60:3:60:23 | Mod5 | Mod1::Mod3::Mod5 |
| constants.rb:60:3:60:23 | Mod5 | Mod4::Mod3::Mod5 |
```
I wonder why the `resolveConstantWriteAccess` drops `Mod4::Mod3::Mod5` instead of  `Mod3::Mod5`

aibaars: No need to look at this problem in this pull request. I think this because the implementation of `resolveConstantWriteAccess` is an over-approximation. If it were correct it should drop both `Mod3::Mod5` and `Mod4::Mod3::Mod5` .

jeffgran: I think if this were correct, the possible fully qualified names for `Mod5` would be:
```ruby
Mod1::Mod3::Mod5
Mod4::Mod3::Mod5
Mod3::Mod5
```
But yeah this PR is definitely an improvement even if it's still not a perfect solution. I'm fine with merging this as-is if you are.

aibaars:
@jeffgran I think the version of the Ruby extractor in your copy of CodeQL is too old. Here is a PR with test data generated with a more recent version: https://github.com/jeffgran/codeql/pull/5

Feel free to merge/cherry-pick/etc the commit above. Alternatively you can build a new version of the Ruby extractor by running `ruby/scripts/create-extractor-pack.sh` . To use it with `codeql test run` you need to supply `--search-path /path/to/github/codeql/ruby/extractor-pack/` flag.

jeffgran:
> @jeffgran I think the version of the Ruby extractor in your copy of CodeQL is too old.

Thanks @aibaars , I was able to build a new extractor and get the same results. I did wonder why I would have all those change to unrelated files when I re-learned the results the first time!

Also added a `change-notes` entry because the CI told me to :)

ruby/ql/lib/codeql/ruby/ast/Constant.qll
```diff
@@ -184,16 +184,34 @@ class ConstantWriteAccess extends ConstantAccess {
...
-    exists(ConstantWriteAccess parent | parent = this.getEnclosingModule() |
-      result = parent.getQualifiedName() + "::" + this.getName()
-    )
-    or
-    /* base case - there's no parent module */
-    not exists(ConstantWriteAccess parent | parent = this.getEnclosingModule()) and
-    result = this.getName()
+  string getAQualifiedName() {
+    result = resolveConstantWriteAccess(this)
+  }
```
aibaars, line 207:
```suggestion
  string getAQualifiedName() { result = resolveConstantWriteAccess(this) }
```

ruby/ql/lib/codeql/ruby/ast/Constant.qll
```diff
@@ -184,16 +184,34 @@ class ConstantWriteAccess extends ConstantAccess {
...
+
+  /**
+   * Gets a qualified name for this constant. Deprecated in favor of
+   * `getAQualifiedName` because this can return more than one value
+   */
+  deprecated string getQualifiedName() {
+    result = this.getAQualifiedName()
   }
```
aibaars, line 215:
```suggestion
  deprecated string getQualifiedName() { result = this.getAQualifiedName() }
```

aibaars (commented):
The QL formatter check fails. I think the only problems are the ones I posted as suggested changes. You can auto-format the code yourself using VSCode (format document option, or setting format-on-save). Alternatively you can run `codeql query format -i` :
```
find ql "(" -name "*.ql" -or -name "*.qll" ")" -print0 | xargs -0 codeql query format -i
```

jeffgran:
> You can auto-format the code yourself using VSCode (format document option, or setting format-on-save). Alternatively you can run codeql query format -i 

Oops, sorry about that. I'm a stubborn emacs user so I haven't tried the VSCode tooling yet. Thanks for this snipped to format everything at once. I rebased on `main` again, ran the formatter and re-ran the tests again to make sure. Hopefully all good now.

aibaars (approved):

PR #7450, opened 2021-12-18
QL: Add query detecting suspiciously missing parameters from the QLDoc of a predicate

erik-krogh:
I added a QL-for-QL query that flags suspicious QLDoc for a predicate when: 
- At least one parameter is documented in the QLDoc. 
- At least one paramter is **not** documented in the QLDoc. 
- The QLDoc has a codeblock (enclosed in backticks) mentioning something that looks like an identifier and is not a parameter. 

This turns out to be good at identifying lots of QLDocs where the parameters mentioned in the qldoc do not match up with the actual parameter names.  

I fixed most of the issues identified in the `javascript/` folder. The remaining issues in `javascript/` seem benign.  
There are still lots of issues for the other languages. I have not tried to fix those.


esbena:
Could this be tuned to handle the case where all parameters have been renamed? That case can not be caught currently due to the first of your requirements. (I'm particularly interested in the /1 case)

erik-krogh:
> Could this be tuned to handle the case where all parameters have been renamed? That case can not be caught currently due to the first of your requirements. (I'm particularly interested in the /1 case)

I thought it would be bad, but it worked great :+1:  (See the extra JS fixes). 

I also found a few more tweaks to reduce the FP rate. 
There are also more benign cases, but I think the precision is OK.

javascript/ql/examples/queries/dataflow/TemplateInjection/TemplateInjection.ql
```diff
@@ -14,7 +14,7 @@ import DataFlow::PathGraph
 /**
  * Gets the name of an unescaped placeholder in a lodash template.
  *
- * For example, the string `<h1><%= title %></h1>` contains the placeholder `title`.
+ * For example, the string "<h1><%= title %></h1>" contains the placeholder "title".
```
tausbn, line 17: I assume the `<h1>...` bit was originally in backticks so as to preserve the formatting, and now it's not. Perhaps it should be changed to
```
`"<h1><%= title %></h1>"`
```
so that both string-ness and pre-formatting are evident.

ql/ql/src/queries/style/MissingParameterInQlDoc.ql
```diff
@@ -0,0 +1,85 @@
...
+ * Get something that looks like a parameter name from the QLDoc,
+ * but which is not a parameter of `p`.
+ */
+private string getAMentionedNonParameter(Predicate p) {
+  exists(string fragment | fragment = getACodeFragment(p) |
+    result = fragment.substring(1, fragment.length() - 1)
+  ) and
+  result.regexpMatch("^[a-z]\\w+$") and
+  not result.toLowerCase() = getAParameterName(p).toLowerCase() and
+  not result = ["true", "false", "NaN", "this"] and // keywords
```
tausbn, line 45: I would add `forall` and `exists`, and possibly the names of various aggregates to this mix.

tausbn (changes requested):
Overall this looks okay to me, but it's still fairly noisy. One common source of noise is when the QLDoc mentions language-specific syntax (like `null`, `break`, or `return`). 

I don't know if there's a way to annotate inline code snippets with the desired highlighting language. Some MarkDown flavours support this, but it's not clear to me that the one in VSCode does.

Either way, I think we need to cut down on the number of false positives before merging this.

rRGN3087741:
###

ql/ql/src/queries/style/MissingParameterInQlDoc.ql
```diff
@@ -0,0 +1,91 @@
...
+  result = strictconcat(string param | param = getAnUndocumentedParameter(p) | param, ", or ")
+}
+
+/** Gets the parameter-like names mentioned in the QLDoc of `p` that are not parameters. */
+private string getMentionedNonParameters(Predicate p) {
+  result = strictconcat(string param | param = getAMentionedNonParameter(p) | param, ", and ")
+}
+
+from Predicate p
+where not p.getLocation().getFile().getBaseName() = "Aliases.qll" // these are OK
```
tausbn, line 88:
```suggestion
where not p.getLocation().getFile().getBaseName() in ["Aliases.qll", "TreeSitter.qll"] // these are OK
```

ql/ql/src/queries/style/MissingParameterInQlDoc.ql
```diff
@@ -0,0 +1,91 @@
...
+pragma[noinline]
+private string getMentionedThings(File file) {
+  // predicates get mentioned all the time, it's fine.
+  result = any(Predicate pred | pred.getLocation().getFile() = file).getName() or
+  result = any(Call c | c.getLocation().getFile() = file).getTarget().getName() or
+  result = any(String s | s.getLocation().getFile() = file).getValue()
+}
+
+/** Gets a parameter name from `p` that is not mentioned in the qldoc. */
+private string getAnUndocumentedParameter(Predicate p) {
```
tausbn, line 66: Maybe also exclude things that belong to type (back-)trackers?
Documenting the meaning of the `t` parameter everywhere is unlikely to be terribly useful.

ql/ql/src/queries/style/MissingParameterInQlDoc.ql
```diff
@@ -0,0 +1,91 @@
...
+ */
+private string getAMentionedNonParameter(Predicate p) {
+  exists(string fragment | fragment = getACodeFragment(p) |
+    result = fragment.substring(1, fragment.length() - 1)
+  ) and
+  result.regexpMatch("^[a-z]\\w+$") and
+  not result.toLowerCase() = getAParameterName(p).toLowerCase() and
+  // keywords
+  not result =
+    ["true", "false", "NaN", "this", "forall", "exists", "null", "break", "return", "not"] and
```
tausbn, line 47:
```suggestion
    ["true", "false", "NaN", "this", "forall", "exists", "null", "break", "return", "not", "if", "then", "else", "import"] and
```
... and we're probably missing a few still.

nickrolfe (approved):
Ruby 👍

tausbn (approved):
Most of the results look legit to me now. Ship it. :ship:

PR #7455, opened 2021-12-20
Python: Add shutil module sinks for path injection query

haby0:
If the path is controlled by some function calls of the `shutil` module, it may cause malicious attacks. For example, the `rmtree` function will delete the specified directory. The `move` function will move the specified file or move the specified file to the specified location, causing the file content to leak.


python/ql/lib/semmle/python/frameworks/Stdlib.qll
```diff
@@ -2942,6 +2942,97 @@ private module StdlibPrivate {
...
+  private class ShutilCopyfileobjCall extends FileSystemAccess::Range, DataFlow::CallCfgNode {
+    ShutilCopyfileobjCall() { this = shutil().getMember("copyfileobj").getACall() }
+
+    override DataFlow::Node getAPathArgument() {
+      result in [
+          this.getArg(0), this.getArgByName("fsrc"), this.getArg(1), this.getArgByName("fdst")
+        ]
+    }
+  }
+
```
RasmusWL, line 3010: These arguments are not paths, but file-likes (so already opened files).

I think there is good value in keeping a note around, so we can easily remember that there is potential in modeling this function :blush: 

```suggestion
  // TODO: once we have flow summaries, model `shutil.copyfileobj` which copies the content between its' file-like arguments.
  // See https://docs.python.org/3/library/shutil.html#shutil.copyfileobj 
```

RasmusWL (changes requested):
Thanks :muscle: 

Can you add the tests to the bottom of https://github.com/github/codeql/blob/cbd7434a7e5d48a16f0ae5840711ef9cabae3988/python/ql/test/library-tests/frameworks/stdlib/FileSystemAccess.py instead? Would be great to see tests using both positional and keyword arguments, like the rest of the tests :blush: 

Besides that, just a minor thing about one of the functions you modeled :blush:

python/ql/lib/semmle/python/frameworks/Stdlib.qll
haby0: Does `ShutilCopyfileobjCall` need to be deleted?

haby0:
> Besides that, just a minor thing about one of the functions you modeled 😊

I don't quite understand the meaning of this sentence.

RasmusWL:
> > Besides that, just a minor thing about one of the functions you modeled blush
> 
> I don't quite understand the meaning of this sentence.

Oh, that was just referring to the inline comment I made.

python/ql/lib/semmle/python/frameworks/Stdlib.qll
RasmusWL: Either that (as in my suggestion, which you can commit directly by accepting the suggestion), or by making the body of `getAPathArgument()` to be `none()`

haby0: Updated.

haby0: Do you think I can apply for a bounty next?

python/ql/test/library-tests/frameworks/stdlib/FileSystemAccess.py
```diff
@@ -238,3 +239,30 @@ def test_fspath():
 
 tempfile.TemporaryDirectory("suffix", "prefix", "dir") # $ getAPathArgument="suffix" getAPathArgument="prefix" getAPathArgument="dir"
 tempfile.TemporaryDirectory(suffix="suffix", prefix="prefix", dir="dir") # $ getAPathArgument="suffix" getAPathArgument="prefix" getAPathArgument="dir"
+
```
RasmusWL, line 242:
```suggestion

# ------------------------------------------------------------------------------
# shutil
# ------------------------------------------------------------------------------

```

RasmusWL (commented):
Can you remove the entire file `python/ql/test/query-tests/Security/CWE-022-PathInjection/shutil_path_injection.py`? Since we now highlight that we model these in `python/ql/test/library-tests/frameworks/stdlib/FileSystemAccess.py`, that is good enough :+1:

haby0:
> Can you remove the entire file `python/ql/test/query-tests/Security/CWE-022-PathInjection/shutil_path_injection.py`? Since we now highlight that we model these in `python/ql/test/library-tests/frameworks/stdlib/FileSystemAccess.py`, that is good enough 👍

No problem!

python/ql/test/library-tests/frameworks/stdlib/FileSystemAccess.py
haby0: Updated!

RasmusWL (approved):
LGTM :+1:

Will need to run tests before merging though

python/ql/test/library-tests/frameworks/stdlib/FileSystemAccess.py
RasmusWL: Thanks :+1:

haby0:
> LGTM 👍
> 
> Will need to run tests before merging though


Ok. I will try to apply for the bounty next. Do you think it's okay?

RasmusWL:
> Ok. I will try to apply for the bounty next. Do you think it's okay?

I'm not officially approving/rejecting bounties. If this additional modeling finds real vulnerabilities (that you can show), I don't see why you shouldn't be able to apply for a bounty.

haby0:
> > Ok. I will try to apply for the bounty next. Do you think it's okay?
> 
> I'm not officially approving/rejecting bounties. If this additional modeling finds real vulnerabilities (that you can show), I don't see why you shouldn't be able to apply for a bounty.

The application has been submitted.

RasmusWL:
We had some problems running tests on this one, but they all pass, so we can merge this now :+1:

haby0:
> We had some problems running tests on this one, but they all pass, so we can merge this now 👍

Thanks.

PR #7458, opened 2021-12-20
QL: add "modelling/modeling" to `ql/non-us-spelling`

erik-krogh:
/cc @intrigus-lgtm 

I also made a patch thing that automatically fix the issues found by the new query.

### DIFF
--- a/ql/ql/src/codeql_ql/style/docs/NonUSSpellingQuery.qll
+++ b/ql/ql/src/codeql_ql/style/docs/NonUSSpellingQuery.qll
@@ -1,0 +1,24 @@
+predicate non_us_word(string wrong, string right) {
+  exists(string s |
+    wrong = s.splitAt("/", 0) and
+    right = s.splitAt("/", 1) and
+    s =
+      [
+        "colour/color", "authorise/authorize", "analyse/analyze", "behaviour/behavior",
+        "modelling/modeling"
+      ]
+  )
+}
+
+bindingset[s]
+predicate contains_non_us_spelling(string s, string wrong, string right) {
+  non_us_word(wrong, right) and
+  (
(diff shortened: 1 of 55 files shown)


hvitved (dismissed):
C# 👍

intrigus-lgtm:
You could also add modelled/modeled.

owen-mc (approved):

ql/ql/src/codeql_ql/style/docs/NonUSSpellingQuery.qll
```diff
@@ -0,0 +1,24 @@
...
+bindingset[s]
+predicate contains_non_us_spelling(string s, string wrong, string right) {
+  non_us_word(wrong, right) and
+  (
+    s.matches("%" + wrong + "%") and
+    wrong != "analyse"
+    or
+    // analyses (as a noun) is fine
+    s.regexpMatch(".*analyse[^s].*") and
+    wrong = "analyse"
```
tausbn, line 22: Perhaps it would make more sense to use an explicit word-boundary match, so something like `s.regexpMatch(".*\\b" + wrong + "\\b.*")`.

tausbn (approved):
Python looks good. One minor comment about the QL.

MathiasVP (approved):
C++ 👍

ql/ql/src/codeql_ql/style/docs/NonUSSpellingQuery.qll
erik-krogh: Good point.  
But I think I'll wait with that until we encounter FPs.  

With the current approach we also catch when the bad spelling is used as a prefix/suffix.  
(But luckily English is not like Danish, so that will be rare).

PR #7400, opened 2021-12-14
Add instructions for creating change notes.

dbartol:

PR #7459, opened 2021-12-20
C++: Increase precision of `cpp/arithmetic-uncontrolled` to `high`

MathiasVP:
Fixes https://github.com/github/codeql-c-team/issues/553.

This PR does two things:

- The first commit removes a bunch of false positives by restricting the sinks to be **signed** integers in the case of potentially overflowing expressions. This removes a ton of results, and I think this is overall a good thing since (unlike signed-integer overflow) unsigned integer overflow is well-defined.

  [Here's](https://lgtm.com/query/2680252163988322272/) a list of 70 random projects that produce alerts on this query, and [here](https://lgtm.com/query/3189303705633211258/) is a difference query that shows how many results we remove with this PR.

  A note on Samate:
  On `main` this query raises `22798` TPs on Samate and after this PR it raises `18012` TPs.
  So we lose `4786` results. Out of those results, 1920 are related to CWE-680 (which we have a [seperate query](https://github.com/github/codeql/blob/main/cpp/ql/src/Security/CWE/CWE-190/TaintedAllocationSize.ql) for), and the remaining 2866 results are CWE-190 related that look like:
    ```cpp
    /* POTENTIAL FLAW: Use a random value */
    data = (unsigned int)RAND32();
    /* POTENTIAL FLAW: Adding 1 to data could cause an overflow */
    unsigned int result = data + 1;
    printUnsignedLine(result);
    ```
  which I'm not convinced is an issue that we want to highlight in real code.

- The second commit raises the query's precision to high, meaning it'll be part of the code-scanning suite and be displayed by default on LGTM.

Here's the results on our usual list of test projects: https://lgtm.com/query/7783254823209558200/.

I'm leaving this PR as a draft while I  run the query on a larger set of LGTM  projects.


geoffw0:
Code changes LGTM.  I still need to look at results.

> The first commit removes a bunch of false positives by restricting the sinks to be signed integers in the case of potentially overflowing expressions. This removes a ton of results, and I think this is overall a good thing since (unlike signed-integer overflow) unsigned integer overflow is well-defined.

I agree with your thinking, I wonder if we should have a separate `medium` precision query `cpp/arithmetic-uncontrolled-unsigned` though (or not).  I suppose the answer depends on the general quality of those lost results.

geoffw0:
LGTM diff - I agree that we're better off without the `REMOVED` results.  The results in `transmission/transmission`, `ntop/nDPI` and `ciso/openh264` *might* be correct but none of the other results on the first page were convincing, which is a poor ratio.  Surprisingly many results are in hashing algorithms where unsigned overflow can be an integral part of the algorithm.

New LGTM results - the first four results in `systemd/systemd` appear to be FPs because there's a range check a few lines above the sink, likewise the fifth result in `systemd/systemd` because `m <= n` and the two results in `libretro-uae` for similar reasons.  We could potentially improve precision further by fixing this (in future).  All other results appear correct / reasonable, and I particularly enjoyed the `haiku/haiku` one in an expression parser (where the user could presumably write an expression with this bug in it).

geoffw0:
> A note on Samate:

We're mostly interested in SAMATE Juliet suite results for queries that are enabled on code scanning, so a narrower query at `high` precision is an improvement on a wider one at `medium` anyway (the latter is largely ignored).

geoffw0 (dismissed):
I'm happy with this change.

rdmarsh2:
I agree with @geoffw0 on the removed results - all of the ones on the first two pages are either tests or crypto code where the overflow looks intended.

cpp/ql/src/change-notes/2022-01-05-promote-uncontrolled-arithmetic.md
```diff
@@ -0,0 +1,4 @@
+---
+category: newQuery
+---
```
geoffw0, line 3: I take it the change note process has changed again - is this documented somewhere?

MathiasVP: Indeed, it has. There's an open PR that documents the process, but it hasn't been merged yet. See here: https://github.com/github/codeql/pull/7400.

geoffw0 (approved):
Approved - my question about the change note is really for my own benefit, not because I think there's anything wrong here.

cpp/ql/src/change-notes/2022-01-05-promote-uncontrolled-arithmetic.md
geoffw0: Ah, thanks!

PR #7469, opened 2021-12-22
C#: Promote existing ad-hoc consistency checks to consistency queries

hvitved:
Identified two issues that are fixed on this PR.


csharp/extractor/Semmle.Extraction.CSharp/Entities/Constructor.cs
```diff
@@ -113,11 +113,7 @@ protected override void ExtractInitializers(TextWriter trapFile)
 
             trapFile.expr_call(init, target);
 
-            var child = 0;
-            foreach (var arg in initializer.ArgumentList.Arguments)
-            {
-                Expression.Create(Context, arg.Expression, init, child++);
-            }
```
tamasvajk, line 120: I think the same changes should be applied to the similar cases in:
- [Tuple.cs](https://github.com/github/codeql/blob/main/csharp/extractor/Semmle.Extraction.CSharp/Entities/Expressions/Tuple.cs#L19), and
- [Initializer.cs](https://github.com/github/codeql/blob/main/csharp/extractor/Semmle.Extraction.CSharp/Entities/Expressions/Initializer.cs#L110).

Maybe it's not possible in all cases to use each of `ref`, `out` or `in`. But `in` works in the below, and I don't think we extract it:
```
using System;
public class C {
    public void M() {
        var i = 1;
        new C { [in i] = 42 };
    }
    
    public int this[in int i] {
        get { return 42; }
        set { Console.WriteLine(value); }
    }
}
```

hvitved: Thanks, I have added those as well.

tamasvajk (approved):

PR #7472, opened 2021-12-22
QL-for-QL: Add a could-be-cast query

erik-krogh:
Inspired by https://github.com/github/codeql/discussions/5160 

I haven't fixed the issues identified by this query (I got a patch ready, but I wanted to reduce the scope of the PR). 

----- 

This query identifies `exists(..)`/`any(..)` that can be simplified to a type-cast:    
`exists(Type t | t = foo())` can be simplified to ```foo() instanceof Type```.  
And `any(Type t | t = foo())` can be simplified to ```foo().(Type)```.


ql/ql/src/queries/style/RedundantAggregate.ql
```diff
@@ -0,0 +1,18 @@
...
+ * @description An exists/any aggregate where a single variable is assigned in the range,
+ *              and used once in the formula, can be replaced with something simpler.
+ * @kind problem
+ * @problem.severity warning
+ * @id ql/redundant-aggregate
+ * @precision high
+ */
+
+import ql
+import codeql_ql.style.RedundatAggregateQuery
```
calumgrant, line 12: Typo - missing `n`

ql/ql/src/codeql_ql/style/RedundatAggregateQuery.qll
```diff
@@ -0,0 +1,50 @@
...
+    count(anyy.getArgument(_)) = 1 and
+    anyy.getArgument(0) = var
+  )
+}
+
+/**
+ * Holds if `aggr` is a redundant aggregate.
+ * The aggregate declares a single variable `var`, the value of which is `operand`, and it is only used in `formula`.
+ */
+predicate redundatAggregate(AstNode aggr, AstNode formula, VarDecl var, AstNode operand) {
```
calumgrant, line 28: Random question, would this also consider rewriting aggregates as inline casts, e.g.
```
exists(Type t | t = foo())
```
into
```
foo().(Type)
```

erik-krogh: It doesn't currently, but it could.  

Your example would become `foo() instanceof Type` as `exists(..)` is a formula and not an expression. 

(`any(Type t | t = foo())` would become `foo().(Type)`).

erik-krogh: And I've done it. 
The patch is grouped together one. 
But if you check out the `run the redundant-inline-cast patch` commit, you'll see plenty of `exists(Type t | t = foo())` getting simplified.  

If the added type-cast itself would be redundant then I don't emit a type-cast.  
(For the above that would be the case if the type of `foo()` is `Type`, in that case it simplifies to `exists(foo())`).

aschackmull:
I disagree with simplifying the `exists` case in general. There are many cases where it's nice to name an intermediate result to make code more readable. Changing a single-use exists-variable to an instanceof is probably a good thing, though.

aschackmull (changes requested):
The `exists` case catches too much.

aschackmull:
> Note: an exists(..) is actually a qualifier and not an aggregate, but I'm unsure what else to call something that is either an exists(..) or an any(..).

I think it's fine to just call it an "exists" in the query name, and then use either "exists" or "any" in the message. In any case, calling it an aggregate when it definitely isn't an aggregate is somewhat problematic, I think.

aschackmull:
There's a FP here: https://github.com/github/codeql/security/code-scanning/5199?query=ref%3Arefs%2Fpull%2F7472%2Fmerge+ref%3Arefs%2Fpull%2F7472%2Fhead

aschackmull:
There are a couple of those FPs - when the single use is in just one branch of an `or`.

aschackmull:
Looking through some more results, I find that I generally agree with the "can replaced with an instanceof expression." suggestions, but I mostly wouldn't change any of the `exists` - the additional variable name typically helps by breaking up an otherwise large expression and documents an intermediate value through a variable name.

aschackmull:
So I think we should limit the query to the cases where the entire `exists` can be replaced by `instanceof`.

erik-krogh:
> So I think we should limit the query to the cases where the entire exists can be replaced by instanceof.

I'll do that, and rewrite the query into a `could be instanceof` query.

aschackmull (approved):
Much better! Looking through a bunch of the alerts, they now all look like something I'd want to fix.

MathiasVP (approved):

PR #7473, opened 2021-12-22
Update ql-style-guide.md

bobychaudhary:
# Minor typo mistake
```diff
-from
+form


intrigus-lgtm:
I believe the current version is correct as is.

Codeql queries contain a `from`, `where`, and `select` clause.
The `from` clause can introduce variables.
Also see:
https://codeql.github.com/docs/ql-language-reference/queries/

bobychaudhary:
yaa you are right, sorry again😅

Issue #7488, opened 2021-12-29
JavaScript type confusion documentation example mistake 


Naman-ntc:
In the [Type confusion through parameter tampering](https://codeql.github.com/codeql-query-help/javascript/js-type-confusion-through-parameter-tampering/) documentation, the fixed example seems to be wrong. 

It does a check on `typeof path !== string` which should be `typeof file !== string`. The comments in the example also do not make sense. I suppose updating [TypeConfusionThroughParameterTampering_fixed.js](https://github.com/github/codeql/blob/main/javascript/ql/src/Security/CWE-843/examples/TypeConfusionThroughParameterTampering_fixed.js) javascript file would be sufficient (the query-tests folder doesn't have this example)?

edoardopirovano:
Greetings, many thanks for spotting this issue! I've opened a PR that fixes that example here: https://github.com/github/codeql/pull/7489

PR #7489, opened 2021-12-29
Fix example in JavaScript query

edoardopirovano:
Closes https://github.com/github/codeql/issues/7488

Fixes the issues with the example pointed out in the linked issues.

### DIFF
--- a/javascript/ql/src/Security/CWE-843/examples/TypeConfusionThroughParameterTampering.js
+++ b/javascript/ql/src/Security/CWE-843/examples/TypeConfusionThroughParameterTampering.js
@@ -4,8 +4,8 @@ var app = require("express")(),
 app.get("/user-files", function(req, res) {
   var file = req.param("file");
   if (file.indexOf("..") !== -1) {
-    // BAD
-    // forbid paths outside the /public directory
+    // BAD: we forbid relative paths that contain ..
+    // as these could leave the public directory
     res.status(400).send("Bad request");
   } else {
     var absolute = path.resolve("/public/" + file);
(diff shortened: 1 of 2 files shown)


Naman-ntc (commented):
Hi, 
I think the comments are still not completely relevant to the type-confusion vulnerability. 
For example, in  `TypeConfusionThroughParameterTampering.js` file, I would write something along the lines of `BAD : user-controlled file variable can be of type array causing exception`.

javascript/ql/src/Security/CWE-843/examples/TypeConfusionThroughParameterTampering.js
```diff
@@ -4,8 +4,8 @@ var app = require("express")(),
 app.get("/user-files", function(req, res) {
   var file = req.param("file");
   if (file.indexOf("..") !== -1) {
-    // BAD
-    // forbid paths outside the /public directory
+    // BAD: we forbid relative paths that contain ..
```
asgerf, line 7: The `BAD` refers to the code example being a "good" or "bad"; it's not about the fact that the incoming request is classified as a "bad request" or not. It looks like it was accidentally copy/pasted into the fixed example without changing it to `GOOD`.

The rest of the comment is really just part of code snippet so should be on a separate line
```suggestion
    // BAD
    // we forbid relative paths that contain ..
```

javascript/ql/src/Security/CWE-843/examples/TypeConfusionThroughParameterTampering_fixed.js
```diff
@@ -3,9 +3,9 @@ var app = require("express")(),
 
 app.get("/user-files", function(req, res) {
   var file = req.param("file");
-  if (typeof path !== 'string' || file.indexOf("..") !== -1) {
-    // BAD
-    // forbid paths outside the /public directory
+  if (typeof file !== 'string' || file.indexOf("..") !== -1) {
+    // BAD: we forbid relative paths that contain ..
```
asgerf, line 7:
```suggestion
    // GOOD
    // we forbid relative paths that contain ..
```

asgerf (approved):

PR #7500, opened 2022-01-03
Propagate taint through AbstractStringBuilder.reverse()

zbazztian:
... and its overrides. Currently, we do not propagate taint through `AbstractStringBuilder.reverse()` calls, which leads us to miss out on results e.g. within [this webgoat example](https://github.com/zbazztian/webgoat-delombok/blob/main/webgoat-lessons/http-basics/src/main/java/org/owasp/webgoat/http_basics/HttpBasicsLesson.java#L43).

Is there a motivation for this or was this just an oversight?

### DIFF
--- a/java/ql/lib/semmle/code/java/frameworks/Strings.qll
+++ b/java/ql/lib/semmle/code/java/frameworks/Strings.qll
@@ -50,6 +50,8 @@ private class StringSummaryCsv extends SummaryModelCsv {
         "java.lang;AbstractStringBuilder;true;insert;;;Argument[-1];ReturnValue;value",
         "java.lang;AbstractStringBuilder;true;replace;;;Argument[-1];ReturnValue;value",
         "java.lang;AbstractStringBuilder;true;replace;;;Argument[2];Argument[-1];taint",
+        "java.lang;AbstractStringBuilder;true;reverse;;;Argument[-1];ReturnValue;value",
+        "java.lang;AbstractStringBuilder;true;reverse;;;Argument[-1];ReturnValue;taint",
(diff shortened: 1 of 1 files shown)


zbazztian:
Note, we might also want to do it for `AbstractStringBuilder.substring()` and `AbstractStringBuilder.subSequence()` as those appear to have been modeled for `java.lang.String` and `java.lang.CharSequence` as well.

java/ql/lib/semmle/code/java/frameworks/Strings.qll
```diff
@@ -50,6 +50,8 @@ private class StringSummaryCsv extends SummaryModelCsv {
         "java.lang;AbstractStringBuilder;true;insert;;;Argument[-1];ReturnValue;value",
         "java.lang;AbstractStringBuilder;true;replace;;;Argument[-1];ReturnValue;value",
         "java.lang;AbstractStringBuilder;true;replace;;;Argument[2];Argument[-1];taint",
+        "java.lang;AbstractStringBuilder;true;reverse;;;Argument[-1];ReturnValue;value",
+        "java.lang;AbstractStringBuilder;true;reverse;;;Argument[-1];ReturnValue;taint",
```
aschackmull, line 54:
```suggestion
```
Adding the value step is all that's needed.

aschackmull (approved):

zbazztian:
@aschackmull Given that we already modeled `subSequence` and `substring` to propagate taint in other classes, would you mind if I also add those for `AbstractStringBuilder`?

aschackmull:
> Given that we already modeled `subSequence` and `substring` to propagate taint in other classes, would you mind if I also add those for `AbstractStringBuilder`?

No, feel free.

PR #7502, opened 2022-01-03
QL: Support trailing comma in set literals

tausbn:
See

https://github.com/tausbn/tree-sitter-ql/commit/725395405e65814f10095a451404b0ced5dc6289

for the grammar changes and corresponding test.

### DIFF
--- a/ql/Cargo.lock
+++ b/ql/Cargo.lock
@@ -514,7 +514,7 @@ dependencies = [
 [[package]]
 name = "tree-sitter-ql"
 version = "0.19.0"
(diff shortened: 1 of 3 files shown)


erik-krogh:
The workflows didn't recompile the extractor, it just used a cached version.  

The cause seems to be that the `Cargo.lock` file hasn't changed.  
I think you need to run a `cargo build` or something.

tausbn:
Aha! Thanks. I figured I had forgotten something.

erik-krogh (approved):
:+1: 

Lots of fixed consistency issues according to code-scanning :tada:

PR #7507, opened 2022-01-04
C#: Refactor and cleanup LibraryTypeDataFlow

michaelnebel:
In this PR, we move and/or migrate the remaining code in the LibrabryTypeDataflow and then delete the file.

- The deletion of LibraryTypeDataFlow is done incrementally in different commits to make it transparent that code is dead before it is deleted.
- Some code has been moved to other files as this is where it logically should be placed.
- clearContent has been migrated from the legacy dataflow to the FlowSummary framework.


csharp/ql/lib/semmle/code/csharp/dataflow/FlowSummary.qll
```diff
@@ -129,3 +152,17 @@ private class SummarizedCallableDefaultClearsContent extends Impl::Public::Summa
...
 class RequiredSummaryComponentStack = Impl::Public::RequiredSummaryComponentStack;
+
+private class RecordConstructorFlowRequiredSummaryComponentStack extends RequiredSummaryComponentStack {
+  private SummaryComponent head;
+
+  RecordConstructorFlowRequiredSummaryComponentStack() {
+    exists(Property p |
+      recordConstructorFlow(_, _, p) and
+      head = SummaryComponent::property(p) and
+      this = SummaryComponentStack::singleton(SummaryComponent::return())
```
hvitved, line 162: I think this can be simplified to `this = SummaryComponentStack::return()`

csharp/ql/lib/semmle/code/csharp/frameworks/system/Text.qll
```diff
@@ -25,6 +26,19 @@ class SystemTextStringBuilderClass extends SystemTextClass {
   Method getAppendFormatMethod() { result = this.getAMethod("AppendFormat") }
 }
 
+/** Clear content for `System.Text.StringBuilder.Clear`. */
+private class SystemTextStringBuilderClearFlow extends SummarizedCallable {
+  SystemTextStringBuilderClearFlow() {
+    this.getDeclaringType() instanceof SystemTextStringBuilderClass and
+    this.hasName("Clear")
```
hvitved, line 33: Alternatively
```ql
this = any(SystemTextStringBuilderClass c).getAMethod("Clear")
```

hvitved (commented):
Great work, so happy to finally get rid of this legacy library.

csharp/ql/lib/semmle/code/csharp/dataflow/FlowSummary.qll
michaelnebel: Yes, that is a good point! To be honest I just moved this code from the LibraryTypeDataFlow without looking that much into it 😸

csharp/ql/lib/semmle/code/csharp/frameworks/system/Text.qll
michaelnebel: Yes - great - that looks much nicer! Thank you!

hvitved (approved):

michaelnebel:
@hvitved: I force pushed the branch again as the "Checks" GitHub Actions just stalls.
This is still the case and happens sometime :-(
Have you experienced something similar?

hvitved (approved):

PR #7171, opened 2021-11-18
JS: Initial models-as-data implementation

asgerf:
Initial implementation of models-as-data. Please use the backlinked issue for high-level discussion of the format itself.

The implementation is split in three files:
- `Shared.qll` is intended to be shared between languages (some day; currently only works for JS)
- `Impl.qll` is intended to contain the langauge-specific details.
- `ModelsAsData.qll` is the public interface from the rest of the library

Some predicates from `Shared.qll` must be usable from `Impl.qll`, but should not be public in the library, so `ModelsAsData.qll` acts as a gatekeeper toward the public API.

The API exposes two modules:
- `ModelInput` is used for contributing models
  - For example, subclass `ModelInput::SinkModelCsv` to add new sinks
- `ModelOutput` is used for accessing interpreted models in terms of API nodes
  - For example, use `ModelOutput::getASinkNode("sql-injection")` to get SQL injection sinks

In this PR I have ported Sequelize and Spanner, as initial validation.

[Evaluation](https://github.com/github/codeql-dca-main/tree/data/asgerf/mad__sequelize__dist-compare/reports) from a slightly earlier version.
- The new result is due to MaD always using API graphs whereas the old model used local flow for credentials. 
- (The new taint sources were due to the inclusion of an rough Express MaD model in the earlier version. They only show up in the diff because `getSourceType()` returned a different string for those source, not because they were truly new)

PR #7511, opened 2022-01-04
Python: remove duplicated spaces in qldoc

erik-krogh:
I was inspired by a duplicated space in some QLDoc in the [MaD implementation](https://github.com/github/codeql/pull/7171/). 

So I tried to see if that would be viable as a QL-for-QL query.  
But it only found a few results in Python, so I think I'll just skip the QL-for-QL query. 

(PR was   written automatically by a tool).


tausbn (approved):
Looks    good to me.

PR #7512, opened 2022-01-04
Experiment with a fixed threshold

tiferet:
After all the changes made on the modeling side, there is now better separation in scores between TPs and FPs. 

This PR experiments with setting a hard-coded score threshold > 0.5 to determine which alerts to surface, rather than always selecting the max-likelihood class, thereby reducing FPs.

----

Replaced by https://github.com/github/codeql/pull/7519 because I need to update the worse rather than current libraries.


tiferet:
The above experiments use a cutoff of 0.65 on the model that's on `main` ☝️

tiferet:
Replaced by https://github.com/github/codeql/pull/7519 because I need to update the worse rather than current libraries.

PR #7515, opened 2022-01-05
C#: Introduce extractor mode to identify DBs created with `codeql test run`

hvitved:
Passing `--qltest` to the extractor when run though `codeql test run` is not done yet, that will be done on an internal follow-up PR.

### DIFF
--- a/csharp/extractor/Semmle.Extraction.CIL.Driver/ExtractorOptions.cs
+++ b/csharp/extractor/Semmle.Extraction.CIL.Driver/ExtractorOptions.cs
@@ -6,6 +6,7 @@ using System.Reflection.Metadata;
 using System.Reflection;
 using System.Runtime.InteropServices;
 using System.Globalization;
+using Semmle.Util;
 using Semmle.Util.Logging;
 
 namespace Semmle.Extraction.CIL.Driver
@@ -169,18 +170,14 @@ namespace Semmle.Extraction.CIL.Driver
     /// <summary>
     /// Parses the command line and collates a list of DLLs/EXEs to extract.
     /// </summary>
-    internal class ExtractorOptions
+    internal class ExtractorOptions : CommonOptions
     {
         private readonly AssemblyList assemblyList = new AssemblyList();
 
         public ExtractorOptions(string[] args)
         {
-            Verbosity = Verbosity.Info;
-            Threads = System.Environment.ProcessorCount;
             PDB = true;
-            TrapCompression = TrapWriter.CompressionMode.Gzip;
-
-            ParseArgs(args);
+            this.ParseArguments(args);
 
             AddFrameworkDirectories(false);
 
@@ -202,12 +199,6 @@ namespace Semmle.Extraction.CIL.Driver
         {
             AddDirectory(RuntimeEnvironment.GetRuntimeDirectory(), extractAll);
         }
-
-        public Verbosity Verbosity { get; private set; }
-        public bool NoCache { get; private set; }
-        public int Threads { get; private set; }
-        public bool PDB { get; private set; }
-        public TrapWriter.CompressionMode TrapCompression { get; private set; }
 
         private void AddFileOrDirectory(string path)
         {
@@ -237,43 +228,25 @@ namespace Semmle.Extraction.CIL.Driver
         /// </summary>
         public IEnumerable<AssemblyName> MissingReferences => assemblyList.MissingReferences;
 
-        private void ParseArgs(string[] args)
+        public override bool HandleFlag(string flag, bool value)
         {
-            foreach (var arg in args)
+            switch (flag)
             {
-                if (arg == "--verbose")
-                {
-                    Verbosity = Verbosity.All;
-                }
-                else if (arg == "--silent")
-                {
-                    Verbosity = Verbosity.Off;
-                }
-                else if (arg.StartsWith("--verbosity:"))
-                {
-                    Verbosity = (Verbosity)int.Parse(arg.Substring(12));
-                }
-                else if (arg == "--dotnet")
-                {
-                    AddFrameworkDirectories(true);
-                }
-                else if (arg == "--nocache")
-                {
-                    NoCache = true;
-                }
-                else if (arg.StartsWith("--threads:"))
-                {
-                    Threads = int.Parse(arg.Substring(10));
-                }
-                else if (arg == "--no-pdb")
-                {
-                    PDB = false;
-                }
-                else
-                {
-                    AddFileOrDirectory(arg);
-                }
+                case "dotnet":
+                    if (value)
+                        AddFrameworkDirectories(true);
+                    return true;
+                default:
+                    return base.HandleFlag(flag, value);
             }
         }
+
+        public override bool HandleArgument(string argument)
+        {
+            AddFileOrDirectory(argument);
+            return true;
+        }
+
+        public override void InvalidArgument(string argument) { }
     }
 }
--- a/csharp/extractor/Semmle.Extraction.CIL.Driver/ExtractorOptions.cs
+++ b/csharp/extractor/Semmle.Extraction.CIL.Driver/ExtractorOptions.cs
@@ -176,8 +176,7 @@ namespace Semmle.Extraction.CIL.Driver
 
         public ExtractorOptions(string[] args)
         {
-            PDB = true;
-            this.ParseArguments(args);
+            this.ParseArguments(args.Append("--pdb").ToArray());
 
             AddFrameworkDirectories(false);
 
--- a/csharp/extractor/Semmle.Extraction.CIL/Entities/PdbSourceFile.cs
+++ b/csharp/extractor/Semmle.Extraction.CIL/Entities/PdbSourceFile.cs
@@ -25,7 +25,7 @@ namespace Semmle.Extraction.CIL.Entities
                 else
                     Context.TrapWriter.Archive(TransformedPath, text);
 
-                yield return Tuples.file_extraction_mode(this, 2);
+                yield return Tuples.file_extraction_mode(this, Context.Extractor.Mode | ExtractorMode.Pdb);
             }
         }
     }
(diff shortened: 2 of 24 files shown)


csharp/extractor/Semmle.Extraction.CIL.Driver/ExtractorOptions.cs
```diff
@@ -237,43 +228,25 @@ private void AddFileOrDirectory(string path)
         /// </summary>
         public IEnumerable<AssemblyName> MissingReferences => assemblyList.MissingReferences;
 
-        private void ParseArgs(string[] args)
+        public override bool HandleFlag(string flag, bool value)
         {
-            foreach (var arg in args)
+            switch (flag)
             {
-                if (arg == "--verbose")
```
michaelnebel, line 244: The semantics of the --verbose flag has now been changed by using the base.HandleFlag method. Is this intentional?

csharp/extractor/Semmle.Extraction.CIL.Driver/ExtractorOptions.cs
```diff
@@ -169,18 +170,14 @@ public void ResolveReferences()
...
-    internal class ExtractorOptions
+    internal class ExtractorOptions : CommonOptions
     {
         private readonly AssemblyList assemblyList = new AssemblyList();
 
         public ExtractorOptions(string[] args)
         {
-            Verbosity = Verbosity.Info;
-            Threads = System.Environment.ProcessorCount;
             PDB = true;
```
michaelnebel, line 179: In the CommonOptions the logic for the PDB and CIL flags are connected. I suppose it doesn't matter as we didn't use the CIL flag before in the ExtractorOptions.

csharp/extractor/Semmle.Extraction.CIL/Entities/PdbSourceFile.cs
```diff
@@ -25,7 +25,7 @@ public override IEnumerable<IExtractionProduct> Contents
                 else
                     Context.TrapWriter.Archive(TransformedPath, text);
 
-                yield return Tuples.file_extraction_mode(this, 2);
+                yield return Tuples.file_extraction_mode(this, Context.Extractor.Mode | ExtractorMode.Pdb);
```
michaelnebel, line 28: Fantastic that we get rid of the magic number and rely on the flag, but shouldn't this only be ExtractorMode.Pdb?

csharp/extractor/Semmle.Extraction.CSharp/Entities/File.cs
```diff
@@ -61,7 +61,7 @@ public override void Populate(TextWriter trapFile)
                 }
             }
 
-            trapFile.file_extraction_mode(this, Context.Extractor.Standalone ? 1 : 0);
+            trapFile.file_extraction_mode(this, Context.Extractor.Mode);
```
michaelnebel, line 64: Now printing the "entire" mode and not only, whether the it is standalone extraction to the trap file. This is this intentional?

csharp/extractor/Semmle.Extraction/Extractor/StandaloneExtractor.cs
```diff
@@ -4,15 +4,19 @@ namespace Semmle.Extraction
 {
     public class StandaloneExtractor : Extractor
     {
-        public override bool Standalone => true;
+        private readonly ExtractorMode mode;
+        public override ExtractorMode Mode => mode;
```
michaelnebel, line 8:
```suggestion
        public override ExtractorMode Mode { get; };
```

csharp/extractor/Semmle.Extraction/Extractor/StandaloneExtractor.cs
```diff
@@ -4,15 +4,19 @@ namespace Semmle.Extraction
...
         /// Creates a new extractor instance for one compilation unit.
         /// </summary>
         /// <param name="logger">The object used for logging.</param>
         /// <param name="pathTransformer">The object used for path transformations.</param>
-        public StandaloneExtractor(ILogger logger, PathTransformer pathTransformer) : base(logger, pathTransformer)
+        public StandaloneExtractor(ILogger logger, PathTransformer pathTransformer, CommonOptions options) : base(logger, pathTransformer)
         {
+            mode = ExtractorMode.Standalone;
+            if (options.QlTest)
+                mode |= ExtractorMode.QlTest;
```
michaelnebel, line 19:
```suggestion
            Mode = ExtractorMode.Standalone;
            if (options.QlTest)
                Mode |= ExtractorMode.QlTest;
```

michaelnebel, line 19: Are we not strict on always using { } for single line if-statements (sometimes this can be overlooked in code-reviews if more statements are added to the if-block)?

csharp/extractor/Semmle.Extraction/Extractor/TracingExtractor.cs
```diff
@@ -4,8 +4,8 @@ namespace Semmle.Extraction
 {
     public class TracingExtractor : Extractor
     {
-        public override bool Standalone => false;
-
+        private readonly ExtractorMode mode;
+        public override ExtractorMode Mode => mode;
```
michaelnebel, line 8: Same comments as for the StandaloneExtractor.

csharp/extractor/Semmle.Extraction/Options.cs
```diff
@@ -37,13 +37,18 @@ public abstract class CommonOptions : ICommandLineOptions
         /// <summary>
         /// Whether to extract PDB information.
         /// </summary>
-        public bool PDB { get; private set; } = false;
+        public bool PDB { get; set; } = false;
```
michaelnebel, line 40: Maybe consider to use the HandleFlag method to set PDB instead of making PDB mutable (however, this will impact the CIL flag as well - but it seems to be un-unused).

csharp/extractor/Semmle.Extraction/Options.cs
```diff
@@ -37,13 +37,18 @@ public abstract class CommonOptions : ICommandLineOptions
...
 
         /// <summary>
         /// Whether "fast extraction mode" has been enabled.
         /// </summary>
         public bool Fast { get; private set; } = false;
 
+        /// <summary>
+        /// Whether extraction is done using `codeql test run`.
+        /// </summary>
+        public bool QlTest { get; private set; } = false;
```
michaelnebel, line 50: Excellent!

csharp/ql/lib/semmle/code/csharp/File.qll
```diff
@@ -209,7 +209,12 @@ class File extends Container, @file {
    * A source file can come from a PDB and from regular extraction
    * in the same snapshot.
    */
-  predicate isPdbSourceFile() { file_extraction_mode(this, 2) }
+  predicate isPdbSourceFile() {
+    exists(int i |
+      file_extraction_mode(this, i) and
+      i.bitAnd(2) = 2
```
michaelnebel, line 215: So this is needed because the entire "mode" is now used and you need to ensure that the specific flag (in this "2" which corresponds to PDB) is present?

michaelnebel (commented):
Looks nice! I have added some questions/comments and suggestions.

csharp/extractor/Semmle.Extraction.CIL.Driver/ExtractorOptions.cs
hvitved: Good spot; I think it makes sense that `--verbose` means the same thing for all `.exe`s.

csharp/extractor/Semmle.Extraction.CIL/Entities/PdbSourceFile.cs
hvitved: No; extractor mode is now a bit field, which allows for more information. For example, a PDB file could be extracted in either a QL test or not, and now we record all combinations.

csharp/extractor/Semmle.Extraction.CSharp/Entities/File.cs
hvitved: Yes, we want to persist all bit fields.

csharp/ql/lib/semmle/code/csharp/File.qll
hvitved: Yes. And this should also be backwards compatible for old DBs.

csharp/extractor/Semmle.Extraction/Extractor/StandaloneExtractor.cs
```diff
@@ -14,9 +13,11 @@ public class StandaloneExtractor : Extractor
         /// <param name="pathTransformer">The object used for path transformations.</param>
         public StandaloneExtractor(ILogger logger, PathTransformer pathTransformer, CommonOptions options) : base(logger, pathTransformer)
         {
-            mode = ExtractorMode.Standalone;
+            Mode = ExtractorMode.Standalone;
             if (options.QlTest)
-                mode |= ExtractorMode.QlTest;
+            {
```
michaelnebel, line 18: The comment related to the use of { } was merely a question. I do see the value in not using { }, when the formatting we have chosen puts all { } on new lines (not sure if this is purely bike-shedding). I just wanted to know, if we have a strict C# coding standard for one way or the other.

michaelnebel (approved):
Looks really nice!

csharp/extractor/Semmle.Extraction/Extractor/StandaloneExtractor.cs
hvitved: @tamasvajk has set up some [automated coding standard rules](https://github.com/github/codeql/blob/main/csharp/.editorconfig), which I think should take effect on format (I use format on save by default). So perhaps the rules do not require the braces.

tamasvajk: I prefer braces everywhere, but I didn't want to enforce my preferences onto the team. Instead I took the `.editorconfig` from the .NET docs site and the roslyn repo.

hvitved:
Rebased to resolve merge conflict.

PR #7516, opened 2022-01-05
C#: Introduce Csv validation on kind.

michaelnebel:


csharp/ql/lib/semmle/code/csharp/dataflow/ExternalFlow.qll
```diff
@@ -348,6 +348,24 @@ module CsvValidation {
...
+    exists(string row, string kind | sinkModel(row) |
+      kind = row.splitAt(";", 7) and
+      not kind = ["code", "sql", "xss", "remote", "html"] and
+      msg = "Invalid kind \"" + kind + "\" in sink model."
+    )
+    or
+    exists(string row, string kind | sourceModel(row) |
+      kind = row.splitAt(";", 7) and
+      not kind = "local" and
+      msg = "Invalid kind \"" + kind + "\" in sink model."
```
intrigus-lgtm, line 367:
```suggestion
      msg = "Invalid kind \"" + kind + "\" in source model."
```
Looks like a copy and paste error.

The index in the `kind = row.split` stuff is the same for source and sink models?

hvitved (approved):

csharp/ql/lib/semmle/code/csharp/dataflow/ExternalFlow.qll
michaelnebel: @intrigus-lgtm : Thank you very much for catching this!

PR #7518, opened 2022-01-05
Python: Extend unreachable statement test

tausbn:
Adds a test demostrating the false positive observed by andersfugmann.

Note that this does not change the `.expected` file, and so the tests
will fail. This is expected.


tausbn:
The internal change has been merged, so now the tests pass again. :tada:

RasmusWL (approved):
Nice :muscle:

PR #7521, opened 2022-01-05
C++: Use Guards library in Overflow.qll

rdmarsh2:
This replaces a partial reimplementation of the Guards library in Overflow.qll with uses of `GuardCondition::ensuresLt`, which handles cases where the controlled expression is not syntactically included in the controlling statement.

### DIFF
--- a/cpp/ql/lib/semmle/code/cpp/security/Overflow.qll
+++ b/cpp/ql/lib/semmle/code/cpp/security/Overflow.qll
@@ -5,10 +5,10 @@
 
 import cpp
 import semmle.code.cpp.controlflow.Dominance
-// `GlobalValueNumbering` is only imported to prevent IR re-evaluation.
 private import semmle.code.cpp.valuenumbering.GlobalValueNumbering
 import semmle.code.cpp.rangeanalysis.SimpleRangeAnalysis
 import semmle.code.cpp.rangeanalysis.RangeAnalysisUtils
+import semmle.code.cpp.controlflow.Guards
 
 /**
  * Holds if the value of `use` is guarded using `abs`.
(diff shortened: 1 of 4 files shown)


cpp/ql/lib/semmle/code/cpp/security/Overflow.qll
```diff
@@ -5,64 +5,27 @@
 
 import cpp
 import semmle.code.cpp.controlflow.Dominance
-// `GlobalValueNumbering` is only imported to prevent IR re-evaluation.
```
geoffw0, line 8: Is this no longer true?

geoffw0 (commented):
Code changes LGTM, perhaps we should do a DCA run to check for wider results and performance changes?

cpp/ql/lib/semmle/code/cpp/security/Overflow.qll
rdmarsh2: Oh, I experimented with using GVN but it didn't work out well. I'll re-add that line

rdmarsh2:
DCA run: https://github.com/github/codeql-dca-main/issues/2269

geoffw0:
The DCA run looks fine.

geoffw0 (dismissed):
Quick diff query here: https://lgtm.com/query/1144214176301359103/
 - the four lost results are exactly what we want to see
 - the new result is a surprise, but I'm not sure its genuinely new - I think I've seen it before.  In any case it looks correct up to the lack of understanding the return value of `read`.

rdmarsh2:
The old version used `Loop`, so the do-while bounds check excluded that new result. With the new version, the first iteration isn't guarded by the bounds check.

cpp/ql/test/query-tests/Security/CWE/CWE-190/semmle/tainted/ArithmeticTainted.expected
```diff
@@ -63,6 +87,14 @@ nodes
 | test5.cpp:17:6:17:18 | call to getTaintedInt | test5.cpp:9:7:9:9 | buf | test5.cpp:17:6:17:18 | call to getTaintedInt | $@ flows to here and is used in arithmetic, potentially causing an overflow. | test5.cpp:9:7:9:9 | buf | User-provided value |
 | test5.cpp:19:6:19:6 | y | test5.cpp:9:7:9:9 | buf | test5.cpp:19:6:19:6 | y | $@ flows to here and is used in arithmetic, potentially causing an overflow. | test5.cpp:9:7:9:9 | buf | User-provided value |
 | test5.cpp:19:6:19:6 | y | test5.cpp:9:7:9:9 | buf | test5.cpp:19:6:19:6 | y | $@ flows to here and is used in arithmetic, potentially causing an underflow. | test5.cpp:9:7:9:9 | buf | User-provided value |
+| test5.cpp:30:17:30:23 | tainted | test5.cpp:9:7:9:9 | buf | test5.cpp:30:17:30:23 | tainted | $@ flows to here and is used in arithmetic, potentially causing an overflow. | test5.cpp:9:7:9:9 | buf | User-provided value |
```
MathiasVP, line 90: It's a shame we don't catch this one anymore. I remember those `imaxabs` having a way-too-good influence on the Samate score (even though it's probably quite useful on real-world code).

rdmarsh2: That's caused by a typo in `guardedAbs`. Unfortunately we also lose the good result on line 40 by fixing it.

MathiasVP: Oh, great! (I mean, it's not great that we lose the good result on line 40, but it's great that we don't gain any additional FPs 🎉).

MathiasVP (approved):

PR #7525, opened 2022-01-06
C++: Remove `rank` aggregate in `SsaInternals`

MathiasVP:
This is hopefully a behavior-preserving optimization 🤞.

Thanks to @hvitved for suggesting this.

### DIFF
--- a/cpp/ql/lib/semmle/code/cpp/ir/dataflow/internal/SsaInternals.qll
+++ b/cpp/ql/lib/semmle/code/cpp/ir/dataflow/internal/SsaInternals.qll
@@ -304,8 +304,8 @@ cached
 private module Cached {
   private predicate defUseFlow(Node nodeFrom, Node nodeTo) {
     exists(IRBlock bb1, int i1, IRBlock bb2, int i2, DefOrUse defOrUse, Use use |
-      defOrUse.hasRankInBlock(bb1, i1) and
-      use.hasRankInBlock(bb2, i2) and
+      defOrUse.hasIndexInBlock(bb1, i1) and
+      use.hasIndexInBlock(bb2, i2) and
       adjacentDefRead(_, bb1, i1, bb2, i2) and
       nodeFrom.asInstruction() = toInstruction(defOrUse) and
       flowOutOfAddressStep(use.getOperand(), nodeTo)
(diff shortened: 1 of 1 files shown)


hvitved (approved):

cpp/ql/lib/semmle/code/cpp/ir/dataflow/internal/SsaInternals.qll
```diff
@@ -74,9 +64,10 @@ private class DefOrUse extends TDefOrUse {
   /** Gets the block of this definition or use. */
   abstract IRBlock getBlock();
 
-  /** Holds if this definition or use has rank `rank` in block `block`. */
-  cached
-  final predicate hasRankInBlock(IRBlock block, int rnk) { rnk = getRank(this, block) }
+  /** Holds if this definition or use has index `index` in block `block`. */
+  final predicate hasIndexInBlock(IRBlock block, int index) {
```
intrigus-lgtm, line 68: OOI: Why has the cached annotation been removed?

intrigus-lgtm, line 68: Previously, `hasRankInBlock` could only hold for `rnk \in [1, ...]` because it delegates to `getRank` which uses a `rank` aggregate (and rank aggregates start at 1).
But `hasIndexInBlock` will now also hold for `index in [0, ...]` because it delegates to `block.getInstruction` which is zero-based I assume.

Isn't this a behavior change or is `rnk`/`index` only used transparently and so this does not matter?

rdmarsh2: The annotation was there to force the work done by `getRank` to be cached, but the new predicate uses the `IRBlock::getInstruction` member predicate, which itself uses a cached [nonmember `getInstruction` predicate](https://github.com/github/codeql/blob/b3cb250ecede42a7294f866a84bd3cabeb39fb0e/cpp/ql/lib/semmle/code/cpp/ir/implementation/aliased_ssa/IRBlock.qll#L254). Those predicates will all have the same contents and column order, so in practice any use of `hasIndexInBlock` will use the cached `getInstruction` directly.

MathiasVP: Good point. This is _completely_ behavior preserving, but all current uses of this predicate shouldn't depend on this change for the exact reason that you mentioned :)

rdmarsh2 (approved):

PR #7526, opened 2022-01-06
Don't include arg -> param edges in PathGraph::edges where arg is not reachable

smowton:
This avoids lots of missing-node warnings from `codeql bqrs interpret` as it discards the nodes that occur in the `edges` relation but not `nodes`. The problem arises because subpaths introduced two variants of `reach`, one of which is more restrictive than simply `reach(succ) and succ = pred.getASuccessor()`, so it no longer suffices to just check that the successor is reachable.

Suggestions re: the best way to test this welcome.


cpp/ql/lib/semmle/code/cpp/dataflow/internal/DataFlowImpl.qll
```diff
@@ -3416,7 +3416,7 @@ private predicate pathSuccPlus(PathNode n1, PathNode n2) = fastTC(pathSucc/2)(n1
  */
 module PathGraph {
   /** Holds if `(a,b)` is an edge in the graph of data flow path explanations. */
-  query predicate edges(PathNode a, PathNode b) { a.getASuccessor() = b and reach(b) }
+  query predicate edges(PathNode a, PathNode b) { a.getASuccessor() = b and reach(a) and reach(b) }
```
hvitved, line 3419: I wonder if we shouldn't restrict sub path nodes to those that can reach a sink as well. I.e., redefine `retReach` as
```ql
  predicate retReach(PathNode n) {
    exists(PathNodeMid out |
      subpaths(_, _, n, out) and
      reach(out)
    )
    or
    exists(PathNode mid |
      retReach(mid) and
      n.getASuccessor() = mid and
      not subpaths(_, mid, _, _)
    )
  }
```

@aschackmull : What do you think?

aschackmull: That makes sense, although I think I'd prefer
```ql
  predicate retReach(PathNode n) {
    exists(PathNodeMid out |
      subpaths(_, _, n, out)
    |
      directReach(out) or retReach(out)
    )
    or
    exists(PathNode mid |
      retReach(mid) and
      n.getASuccessor() = mid and
      not subpaths(_, mid, _, _)
    )
  }
```
to avoid making it mutually recursive with `reach`.

aschackmull: We might also want to restrict the final `subpaths` relation if we want to get rid of the `Skipping subpaths tuple with missing node` warnings.

hvitved: Yes, we should certainly also restrict `subpaths`.

MathiasVP:
The changes to the `.expected` files LGTM.

aschackmull:
Looks like the qltest failures are highlighting a pre-existing bug that I was actually just investigating in a separate context. I don't have a complete overview yet, but I expect to fix it in a separate PR.

aschackmull:
> Looks like the qltest failures are highlighting a pre-existing bug that I was actually just investigating in a separate context. I don't have a complete overview yet, but I expect to fix it in a separate PR.

Actually, the fix is simple and would likely conflict, so I'll just push it here.

aschackmull (approved):

PR #7063, opened 2021-11-04
C++: Use `SimpleRangeAnalysis` in `Printf.qll`

MathiasVP:
Fixes some of the cases of https://github.com/github/codeql-c-team/issues/199.

This PR uses `SimpleRangeAnalysis` to infer smaller bounds for arguments to `sprintf` and related functions. Previously, `cpp/overrunning-write` would raise an alert in this case:
```cpp
char buffer[2];

if (unsigned_value < 10) {
	sprintf(buffer, "%u", unsigned_value);
}
```
and now we infer that the value of `unsigned_value` at the point where we call `sprintf` will be at most 9, and so the written value fits into `buffer`.

This has quite a positive impact on LGTM. Here's a differences query: https://lgtm.com/query/6430583417087825070/. I haven't looked at all of them yet, but so far I've only seen FPs being removed.

PR #7543, opened 2022-01-07
C++: Use range analysis for maximum lengths of `%x` formats

rdmarsh2:


MathiasVP (commented):
LGTM! I'd appreciate a DCA run on this before we merge it, though. This change will probably also conflict with https://github.com/github/codeql/pull/7386, but it shouldn't be difficult to fixup, I guess.

redsun82:
I don't mind this getting merged before https://github.com/github/codeql/pull/7386, I can easily update that :+1:

Should we maybe open another issue to do the same for octal as well?

MathiasVP:
We might as well. Then we can fix it once we find someone actually using octal formats 😄.

MathiasVP:
> LGTM. I think this deserves a change note.

We didn't add one for the [base-10 case](https://github.com/github/codeql/pull/7063), though. But I guess we can add a change note saying "we did this for both base-10 and base-16".

redsun82 (approved):
:ship:

PR #7550, opened 2022-01-10
C#: Support for identifying whether a using directive is "global".

michaelnebel:
In this PR we

- Make support for global using statements in the C# extractor (C# 10 feature).
- Make database upgrade/downgrade scripts.
- Update the stats file using a DCA run, where some projects uses C# 10 features are included.


michaelnebel:
Setting this for review now in case any changes are required prior to running the DCA job.

michaelnebel:
@hvitved : It looks like there are a couple of projects that fails to build during the DCA run. One of this is the "Orchard" projekt.
A couple of observations.

- The Orchard project was updated to target .NET6 on November 8th 2021.
- Our CI agents doesn't have .NET6 installed (explains why the Orchard project doesn't build).
- C# 10 is only available when using .NET6.

Does this mean that we can't make any meaningful stats for tables containing information related to the C# 10 language features as we can't provide a SHA in the DCA run that requires .NET6 in the target projects (as this will not built on the agents)?

hvitved (commented):
LGTM. Let's see what the performance comparison looks like after adding new stats.

michaelnebel:
Sounds good to me!

hvitved (approved):

PR #7551, opened 2022-01-10
C++: Fix join orders in `cpp/unsigned-difference-expression-compared-zero`

MathiasVP:


dbartol (approved):
LGTM pending performance numbers, although happy for others to weigh in.

cpp/ql/src/Security/CWE/CWE-191/UnsignedDifferenceExpressionComparedZero.ql
```diff
@@ -36,47 +38,56 @@ predicate isGuarded(SubExpr sub, Expr left, Expr right) {
...
 predicate exprIsSubLeftOrLess(SubExpr sub, DataFlow::Node n) {
-  n.asExpr() = sub.getLeftOperand()
-  or
-  exists(DataFlow::Node other |
-    // dataflow
-    exprIsSubLeftOrLess(sub, other) and
-    (
-      DataFlow::localFlowStep(n, other) or
-      DataFlow::localFlowStep(other, n)
+  interestingSubExpr(sub, _) and // Manual magic
```
geoffw0, line 41: Because `exprIsSubLeftOrLess` is recursive, this is not simply a copy of the call to `interestingSubExpr` in the query `where` block - it also enforces constraints on each recursive step that weren't there before.  I'd like to see either reasoning or a diff query demonstrating that this hasn't changed behaviour in a way we care about.

geoffw0 (commented):
Query changes otherwise LGTM.

cpp/ql/src/Security/CWE/CWE-191/UnsignedDifferenceExpressionComparedZero.ql
MathiasVP: The `sub` column never changes in any of the recursive calls, so the constraint placed on `sub` here shouldn't make any difference since the same `SubExpr` will be restricted in the `where` block anyway (unless I'm missing something in the query?)

geoffw0: Ah I see, the `_` is important here.  :+1:

geoffw0 (approved):
I'm happy with this PR, subject to the performance stats we're waiting on.  I have confirmed this change resolves the timeout in `gcwnow/linux/ci`, so 'no worse' performance on DCA would be sufficient for a merge IMO.

MathiasVP:
DCA looks good. Merging! 🎉

PR #7552, opened 2022-01-10
Note that parameterizations of local classes are themselves local

smowton:
Previously `LocalClass` itself would match `.isLocal()` whereas `LocalClass<Param>` would not. Rather than require each individual user to check for `.getSourceDeclaration().isLocal()`, let's note that the specializations themselves are local.


atorralba (approved):
LGTM

PR #7567, opened 2022-01-11
ATM: Optimize body tokens by pushing in size restriction

henrymercer:
Pushing the restriction to 256 tokens into the `bodyTokens` predicate means we avoid this predicate blowing up due to very large functions.

This results in a runtime improvement from 1800s+ to 294s as measured on a problematic repo on my machine (I didn't wait for the query to finish running).


Z80coder (approved):
I think this is gonna have a significant impact.

PR #7577, opened 2022-01-12
C#: Make support for Line span pragma

michaelnebel:
C# 10 introduced a new #line pragma, where it is possible to specify, how start and end (line and column) including an offset should be printed, when a compiler warning is issued.
In the PR we

- Introduce support for the new construct
- Include some testcases
- Make some minor refactoring to avoid magic numbers (introducing an enum for #line kind)
- Update stats to include the new tables (unfortunately they are empty).


michaelnebel:
DCA run for new stats: https://github.com/github/codeql-dca-main/issues/2425

michaelnebel:
Need to also look into the SourceLocation stuff (forgot that). Will write something here, when investigation is finished.

michaelnebel:
> Need to also look into the SourceLocation stuff (forgot that). Will write something here, when investigation is finished.

Tried to do some dev testing of this and looks like it is working out of the box.
```codeql
query predicate elements(Assignment statement, File f) { statement.getLocation().getFile() = f }
```
Which among other things produced the output
```codeql
| LinePragmasRef1.cs:1:9:1:13 | Int32 i = ... | LinePragmasRef1.cs:0:0:0:0 | LinePragmasRef1.cs |
| LinePragmasRef2.cs:2:13:2:17 | Int32 j = ... | LinePragmasRef2.cs:0:0:0:0 | LinePragmasRef2.cs |
```
and it looks like this respects the line span pragma info.

michaelnebel:
Ready for review.

csharp/ql/lib/semmle/code/csharp/Preprocessor.qll
```diff
@@ -208,21 +208,40 @@ class HiddenLineDirective extends LineDirective {
...
+class SpanLineDirective extends NumericOrSpanLineDirective {
+  SpanLineDirective() { directive_lines(this, 3) }
+
+  /** Gets the offset of this directive. */
+  int getOffset() { directive_line_offset(this, result) }
+
+  /** Gets the span of this directive. */
+  predicate span(int startLine, int startColumn, int endLine, int endColumn) {
+    directive_line_span(this, startLine, startColumn, endLine, endColumn)
+  }
```
tamasvajk, line 240: The comment doesn't match the predicate. `/** Holds if the specified start and end positions match this SpanLineDirective. */`

michaelnebel: Ah yes - you are right! I will fix that!

csharp/extractor/Semmle.Extraction.CSharp/Entities/PreprocessorDirectives/LineOrSpanDirective.cs
```diff
@@ -0,0 +1,32 @@
...
+    }
+
+    internal abstract class LineOrSpanDirective<T> : PreprocessorDirective<T> where T : LineOrSpanDirectiveTriviaSyntax
+    {
+        protected LineOrSpanDirective(Context cx, T trivia)
+            : base(cx, trivia)
+        {
+        }
+
+        protected override void PopulatePreprocessor(TextWriter trapFile)
```
hvitved, line 23: I think I would add a `LindeDirectiveKind kind` field to this class, which should be passed in to the constructor, and then do `trapFile.directive_lines(this, kind);` here instead of in the base classes.

csharp/extractor/Semmle.Extraction.CSharp/Entities/PreprocessorDirectives/LineSpanDirective.cs
```diff
@@ -0,0 +1,40 @@
...
+using Microsoft.CodeAnalysis.CSharp.Syntax;
+using System.IO;
+
+namespace Semmle.Extraction.CSharp.Entities
+{
+    internal class LineSpanDirective : LineOrSpanDirective<LineSpanDirectiveTriviaSyntax>
+    {
+        private LineSpanDirective(Context cx, LineSpanDirectiveTriviaSyntax trivia)
+            : base(cx, trivia) { }
+        public static LineSpanDirective Create(Context cx, LineSpanDirectiveTriviaSyntax line) =>
```
hvitved, line 12: newline

csharp/extractor/Semmle.Extraction.CSharp/Entities/PreprocessorDirectives/LineSpanDirective.cs
```diff
@@ -0,0 +1,40 @@
...
+            var startColumn = (int)Symbol.Start.Character.Value!;
+            var endLine = (int)Symbol.End.Line.Value!;
+            var endColumn = (int)Symbol.End.Character.Value!;
+            trapFile.directive_line_span(this, startLine, startColumn, endLine, endColumn);
+
+            var offset = Symbol.CharacterOffset.Value;
+            if (offset is not null)
+            {
+                trapFile.directive_line_offset(this, (int)offset);
+            }
```
hvitved, line 29:
```suggestion
            if (Symbol.CharacterOffset.Value is int offset)
            {
                trapFile.directive_line_offset(this, offset);
            }
```

csharp/ql/lib/semmle/code/csharp/Preprocessor.qll
```diff
@@ -208,21 +208,40 @@ class HiddenLineDirective extends LineDirective {
   override string getAPrimaryQlClass() { result = "HiddenLineDirective" }
 }
 
+abstract private class NumericOrSpanLineDirective extends LineDirective {
```
hvitved, line 211: No `abstract` classes. So you will need to add a charpred
```
NumericOrSpanLineDirective() { directive_lines(this, [2, 3]) }
```

hvitved (changes requested):
Approach LGTM.

csharp/extractor/Semmle.Extraction.CSharp/Entities/PreprocessorDirectives/LineOrSpanDirective.cs
michaelnebel: Yes, that is a good idea! I will do that.

csharp/extractor/Semmle.Extraction.CSharp/Entities/PreprocessorDirectives/LineSpanDirective.cs
michaelnebel: Yep

csharp/extractor/Semmle.Extraction.CSharp/Entities/PreprocessorDirectives/LineSpanDirective.cs
michaelnebel: Ah yes!

csharp/ql/lib/semmle/code/csharp/Preprocessor.qll
hvitved: No worries. I also now realized that the class is `private`, so `abstract` would probably have been OK. But let's keep it as is.

hvitved (approved):
LGTM. Then we just need a final DCA run, since the stats have been updated.

michaelnebel:
> LGTM. Then we just need a final DCA run, since the stats have been updated.

Great! I will start one.

michaelnebel:
Execution times are looking good. Merging now.

PR #7578, opened 2022-01-12
C++: Store destinations should not be uses for dataflow SSA

MathiasVP:
This PR is basically a glorified cherry-pick of https://github.com/github/codeql/commit/092beb8b73cc23c20f568246fb8e6568d51ed608. The code was written as a response to one of the reviews on https://github.com/github/codeql/pull/6825, but the commit had to be reverted because it caused a loss of dataflow on some projects.

I've now looked at those results, and it turns out they were all FPs caused by weird interaction with phi nodes. e8afec413aabf7407e5502bf1245d50aacb22af9 demonstrates the kind of false-positive that underlies all of the lost results, and 25253c7b8df9a5bf800841881575fc9803f8d7c0 fixes the false-positive by doing exactly what 092beb8b73cc23c20f568246fb8e6568d51ed608 did.

Performance on DCA looks fine.

### DIFF
--- a/cpp/ql/lib/semmle/code/cpp/ir/dataflow/internal/SsaInternals.qll
+++ b/cpp/ql/lib/semmle/code/cpp/ir/dataflow/internal/SsaInternals.qll
@@ -178,7 +178,7 @@ private predicate isExplicitUse(Operand op) {
       load.getAUse().getUse() instanceof InitializeIndirectionInstruction
     ) and
     // Don't include this operand as a use if the only use of the address is for a write
-    // that definately overrides a variable.
+    // that definitely overrides a variable.
     not (explicitWrite(true, _, vai) and exists(unique( | | vai.getAUse())))
   )
 }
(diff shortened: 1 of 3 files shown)


cpp/ql/lib/semmle/code/cpp/ir/dataflow/internal/SsaInternals.qll
```diff
@@ -170,10 +170,16 @@ private class ReturnParameterIndirection extends Use, TReturnParamIndirection {
...
-    load.getAUse().getUse() instanceof InitializeIndirectionInstruction
+  exists(VariableAddressInstruction vai | vai = op.getDef() |
+    // Don't include this operand as a use if it only exists to initialize the
+    // indirection of a parameter.
+    not exists(LoadInstruction load |
+      load.getSourceAddressOperand() = op and
+      load.getAUse().getUse() instanceof InitializeIndirectionInstruction
+    ) and
+    // Don't include this operand as a use if the only use of the address is for a write
+    // that definately overrides a variable.
```
redsun82, line 181: definately -> definitely

MathiasVP: Thanks! Fixed in e1598aba5eedbf9652280f8323b8025063b6c43a.

rdmarsh2 (approved):

PR #7588, opened 2022-01-13
Dataflow: Add language-specific NeedsReference predicates

owen-mc:
This is needed for codeql-go, which wants to specify that "ReturnValue[n]" needs a reference


hvitved (approved):
LGTM

PR #7598, opened 2022-01-14
QL: field only used in charPred

erik-krogh:
Flags some bad style I recently encountered during a code-review. 

I also went ahead and fixed all the alerts found in JS. (Except for [this one](https://github.com/github/codeql/security/code-scanning/6736?query=ref%3Arefs%2Fpull%2F7598%2Fmerge+ref%3Arefs%2Fpull%2F7598%2Fhead), and the alerts in shared files).  

Most of the time the fix is to somehow inline the field into the charpred.   
But the right way to inline the field can vary a lot, so I'm not writing an automated patch for this. 

I think that for some of the issues (e.g. [this one](https://github.com/github/codeql/security/code-scanning/6745?query=ref%3Arefs%2Fpull%2F7598%2Fmerge+ref%3Arefs%2Fpull%2F7598%2Fhead)) the problem is that the field should have been used somewhere.

### DIFF
--- a/javascript/ql/test/library-tests/TypeScript/LocalTypeResolution/tests.ql
+++ b/javascript/ql/test/library-tests/TypeScript/LocalTypeResolution/tests.ql
@@ -11,17 +11,17 @@ abstract class Violation extends ASTNode {
  * The assertion holds if `name1 = name2`, indicating that `X` resolved to the right interface.
  */
 class TypeResolutionAssertion extends TupleTypeExpr, Violation {
-  InterfaceDeclaration interface;
-  LocalTypeAccess typeAccess;
   string expected;
   string actual;
 
   TypeResolutionAssertion() {
-    typeAccess = getElementType(0) and
-    expected = getElementType(1).(StringLiteralTypeExpr).getValue() and
-    typeAccess.getLocalTypeName() = interface.getIdentifier().(TypeDecl).getLocalTypeName() and
-    actual = interface.getField("where").getTypeAnnotation().(StringLiteralTypeExpr).getValue() and
-    actual != expected
+    exists(InterfaceDeclaration interface, LocalTypeAccess typeAccess |
+      typeAccess = getElementType(0) and
+      expected = getElementType(1).(StringLiteralTypeExpr).getValue() and
+      typeAccess.getLocalTypeName() = interface.getIdentifier().(TypeDecl).getLocalTypeName() and
+      actual = interface.getField("where").getTypeAnnotation().(StringLiteralTypeExpr).getValue() and
+      actual != expected
+    )
   }
 
   override string reason() {
--- a/ql/ql/src/codeql/GlobalValueNumbering.qll
+++ b/ql/ql/src/codeql/GlobalValueNumbering.qll
@@ -5,7 +5,7 @@ private import codeql_ql.ast.internal.Builtins
 
 private newtype TValueNumber =
   TVariableValueNumber(VarDecl var) { variableAccessValueNumber(_, var) } or
-  TFieldValueNumber(VarDecl var) { fieldAccessValueNumber(_, var) } or
+  TFieldValueNumber(FieldDecl var) { fieldAccessValueNumber(_, var) } or
   TThisValueNumber(Predicate pred) { thisAccessValueNumber(_, pred) } or
   TPredicateValueNumber(PredicateOrBuiltin pred, ValueNumberArgumentList args) {
     predicateCallValueNumber(_, pred, args)
@@ -103,7 +103,7 @@ private predicate variableAccessValueNumber(VarAccess access, VarDef var) {
   access.getDeclaration() = var
 }
 
-private predicate fieldAccessValueNumber(FieldAccess access, VarDef var) {
+private predicate fieldAccessValueNumber(FieldAccess access, FieldDecl var) {
   access.getDeclaration() = var
 }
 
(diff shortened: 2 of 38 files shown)


aschackmull:
> I think that for some of the issues (e.g. [this one](https://github.com/github/codeql/security/code-scanning/6745?query=ref%3Arefs%2Fpull%2F7598%2Fmerge+ref%3Arefs%2Fpull%2F7598%2Fhead)) the problem is that the field should have been used somewhere.

That code is fine. The field can just be deleted there, if we want.

aschackmull:
It's an instance of a harmless pattern that appears to also occur many other places: When a class wraps an IPA-injector, all the columns from the IPA case are added as fields, and it may well be that not all of those needs to be referenced in member predicates.

aschackmull:
We may want to exclude that pattern from this query.

ql/ql/src/queries/style/FieldOnlyUsedInCharPred.ql
```diff
@@ -0,0 +1,25 @@
...
+import ql
+
+from Class c, FieldDecl f
+where
+  c.getAField() = f and
+  forex(FieldAccess access | access.getDeclaration() = f |
+    access.getEnclosingPredicate() = c.getCharPred()
+  ) and
+  // excluding fields that are uniquely used in call to an IPA constructor
+  not unique(FieldAccess access | access.getDeclaration() = f | access) =
```
kaeluka, line 20: Could you explain this (line 20-23) to me, perhaps in the huddle for a faster turnaround time? It reads a bit like you might exclude cases where the field is used multiple times in the charpred. But maybe that's not an issue here, because we're looking at IPA constructors only? Anyway, it's a bit opaque to me.

ql/ql/src/queries/style/FieldOnlyUsedInCharPred.ql
```diff
@@ -0,0 +1,25 @@
+/**
```
kaeluka, line 1: As you're adding a new query here, does this need a change-note?

kaeluka (commented):
It looks good to me — but I've left two comments where I'm a bit uncertain.

ql/ql/src/queries/style/FieldOnlyUsedInCharPred.ql
erik-krogh: QL-for-QL is different, it's not a product.  
No change-note or tests are needed.  
(The `Code scanning results / CodeQL` check is effectively a test).

kaeluka: 👍

ql/ql/src/queries/style/FieldOnlyUsedInCharPred.ql
erik-krogh: Yes, I'm excluding cases where the field is used multiple times in the charpred.  

It's to only exclude the pattern where the field is only used as an argument to a call to an IPA constructor. 
That pattern is quite common, and perfectly readable, so we decided to exclude it.  

You can see the pattern a lot in [`Ast.qll` in QL-for-QL](https://github.com/github/codeql/blob/6aad8d6897e8b24790a8be1338656c81e80f10f5/ql/ql/src/codeql_ql/ast/Ast.qll#L511).

kaeluka: OK, we also had a chat about this. Makes sense to me. 👍

javascript/ql/test/library-tests/TypeScript/LocalTypeResolution/tests.ql
```diff
@@ -11,17 +11,17 @@ abstract class Violation extends ASTNode {
  * The assertion holds if `name1 = name2`, indicating that `X` resolved to the right interface.
  */
 class TypeResolutionAssertion extends TupleTypeExpr, Violation {
-  InterfaceDeclaration interface;
-  LocalTypeAccess typeAccess;
```
kaeluka, line 15: oh my, TWICE! :)

ql/ql/src/codeql/GlobalValueNumbering.qll
```diff
@@ -5,7 +5,7 @@ private import codeql_ql.ast.internal.Builtins
 
 private newtype TValueNumber =
   TVariableValueNumber(VarDecl var) { variableAccessValueNumber(_, var) } or
-  TFieldValueNumber(VarDecl var) { fieldAccessValueNumber(_, var) } or
+  TFieldValueNumber(FieldDecl var) { fieldAccessValueNumber(_, var) } or
```
kaeluka, line 8: is this an unrelated drive-by bugfix?

kaeluka:
The changes to the javascript analyses look good to me! 👍

ql/ql/src/codeql/GlobalValueNumbering.qll
MathiasVP: I think `var intanceof FieldDecl` was implied by `fieldAccessValueNumber` already. So this is just making the branch more readable, I think.

erik-krogh: Yes, I was trying to make it clear when something is a `FieldDecl` and when something is a `VarDecl`.  

There is a lot of drive-by "Make `FieldDecl` and it's uses better"-fixes in this PR.

kaeluka (approved):
The changes to the javascript analyses look good to me! 👍

PR #7599, opened 2022-01-14
Java: Model Appenable and Writer

bmuskalla:
This helps with downstream models to properly track taint through methods like `write(char[])`.


java/ql/lib/semmle/code/java/frameworks/Strings.qll
```diff
@@ -40,9 +40,6 @@ private class StringSummaryCsv extends SummaryModelCsv {
         "java.lang;String;false;valueOf;(char);;Argument[0];ReturnValue;taint",
         "java.lang;String;false;valueOf;(char[],int,int);;Argument[0];ReturnValue;taint",
         "java.lang;String;false;valueOf;(char[]);;Argument[0];ReturnValue;taint",
-        "java.io;StringWriter;true;append;;;Argument[0];Argument[-1];taint",
-        "java.io;StringWriter;true;append;;;Argument[-1];ReturnValue;value",
```
atorralba, line 44: Was this row accidentally removed? Do we need a test for this (since all tests passed anyway)?

bmuskalla: No, it was replaced by the `Appendable` rows. We already have tests for Appendable-like types (e.g. `AbstractStringBuilder`). But can add explicit tests for `Appendable` if you want

atorralba: Then `"java.lang;Appendable;true;append;;;Argument[-1];ReturnValue;taint"` should be of `value` type instead of `taint`, shouldn't it? At least I think we need a test for this fluent logic, since this change went unnoticed by the tests.

bmuskalla: Done, added test as well

bmuskalla:
@atorralba Fixed your concerns and added a few more small models that we need to proceed with commons IO. Didn't feel worth another PR

atorralba (approved):

PR #7603, opened 2022-01-14
Java: Replace Commons IO model

bmuskalla:
Replace the manual model with a generated model.
We keep some manual models that cannot be auto-generated today but will likely be replaced by future iterations on the model generator.


atorralba:
Should this follow the same naming convention as https://github.com/github/codeql/pull/7634? i.e. `IO.qll` -> `IOGenerated.qll` and `IOCustom.qll` -> `IO.qll`.

atorralba (approved):

PR #7605, opened 2022-01-14
C#: Support for record structs

michaelnebel:
In this PR we make support for record structs and change the original implementation for `record`.
Characteristics of the old implementation

- When records were introduced in  C# they were implicitly record types.
- The QL code used the presence of a <Clone> method to decide, whether a type should be considered a record type. The <Clone> method doesn't exist for record struct's, which means we can't reuse this idea to identify, whether a type is a `record`.

It appears that there is nothing in IL level that identifies, whether something is a record type (it seems native to C# only).
In C# 10, it is now also possible to declare a `record struct` and `record class`. `record class` is clarification of `record` (that is `record` is just a shorthand for `record class`). That is `record` implicitly means `record class` and as such, we can consider the `record` keyword to be a modifier that applies to `class` and `struct`.


michaelnebel:
@tamasvajk @hvitved This is the initial suggestion for record struct support. I am not sure,

- if it is acceptable that we remove some of the functionality from cil
- if we should delete the RecordClone and friends classes.

csharp/ql/lib/semmle/code/cil/Types.qll
```diff
@@ -60,11 +60,6 @@ class Class extends ValueOrRefType {
   Class() { this.isClass() }
 }
 
-/** A `record`. */
-class Record extends Class {
-  Record() { this.isRecord() }
-}
-
```
tamasvajk, line 67: Should we delete this? It is probably not used by too many users, but instead of deleting it, we could deprecate it to not introduce a breaking change. Also, we could introduce `RecordClass` instead of `Record`. (We wouldn't need `RecordStruct` because that's indistinguishable from `Struct` in IL.)

tamasvajk:
> * if we should delete the RecordClone and friends classes.

I don't think we should delete `RecordClone` altogether. That's used as the target of a `... with { ... }` expression and might be useful for query writers. BTW, we might need to change the `WithExpr` in QL to support the C# 10 changes. I added a separate point to the internal C# 10 issue.

csharp/ql/lib/semmle/code/cil/Types.qll
michaelnebel: Alright - thank you! Let me try and re-cap
- We deprecate the <code>Record</code> class in cil/Types.qll and introduce a <code>RecordClass</code> in cil/types.qll (ie. we do not deprecate the support for records in cil, we de-facto rename it instead).
- We deprecate <code>Record</code> in csharp/type.qll and make <code>RecordClass</code> instead (ie. just rename).
Question: Why does it make sense to talk about records (classes) in IL, when they can only be observed by their side effect of having a generated <Clone> method (so basically, I am asking why we shouldn't deprecate the support for "Record" entirely in cil/Types.qll)? It seems that the record concept is native only to C#.

michaelnebel:
> > * if we should delete the RecordClone and friends classes.
> 
> I don't think we should delete `RecordClone` altogether. That's used as the target of a `... with { ... }` expression and might be useful for query writers. BTW, we might need to change the `WithExpr` in QL to support the C# 10 changes. I added a separate point to the internal C# 10 issue.

Great - thank you! 😄

csharp/ql/lib/semmle/code/cil/Types.qll
tamasvajk: > so basically, I am asking why we shouldn't deprecate the support for "Record" entirely in cil/Types.qll?

This is also an option. And probably the more pragmatic option, so let's go this way.

csharp/ql/lib/semmle/code/csharp/Type.qll
```diff
@@ -449,6 +451,14 @@ class SimpleType extends ValueType, @simple_type {
   override SystemNamespace getDeclaringNamespace() { any() }
 }
 
+/**
+ * A `record` like type.
+ * This can be either a `class` or a `struct`.
+ */
+abstract class RecordType extends ValueOrRefType {
```
hvitved, line 458: Remove `abstract`.

csharp/ql/lib/semmle/code/csharp/Type.qll
```diff
@@ -711,6 +721,20 @@ class Struct extends ValueType, @struct_type {
...
+/**
+ * A `record struct`, for example
+ * ```csharp
+ * record struct RS {
+ *   ...
+ * }
+ * ```
+ */
+class RecordStruct extends RecordType, Struct {
+  RecordStruct() { this.isRecord() }
```
hvitved, line 733: Redundant.

csharp/ql/lib/semmle/code/csharp/Type.qll
```diff
@@ -765,6 +789,18 @@ class Class extends RefType, @class_type {
...
+/**
+ * DEPRECATED: Use `RecordClass` instead.
+ */
+deprecated class Record extends Class {
+  Record() { this.isRecord() }
+
+  /** Gets the clone method of this record. */
+  RecordCloneMethod getCloneMethod() { result = this.getAMember() }
+
+  override string getAPrimaryQlClass() { result = "Record" }
```
hvitved, line 801: Remove; we do don't want to encourage the use of deprecated classes.

csharp/ql/test/library-tests/dataflow/fields/J.cs
```diff
@@ -27,6 +29,14 @@ private void M1()
         Sink(r4.Prop2); // no flow
     }
 
+    private void M2()
+    {
+        var o = Source<object>(1);
```
hvitved, line 34: Change `1` to `2` (and in the inline expectation comment below).

hvitved (changes requested):
Approach looks good to me.

csharp/ql/lib/semmle/code/csharp/Type.qll
michaelnebel: May I ask why?
When declared abstract we are forcing RecordType = RecordClass \union RecordStruct. ValueOrRefType is pretty general and maybe "one day" other types will get the "record" modifier. Do we then automatically wan't them to be a part of RecordType (this might then break FlowSummary.qll implementation)? Right now there is a built in assumption that the RecordType is the disjoint union of two other sets.

csharp/ql/lib/semmle/code/csharp/Type.qll
michaelnebel: Oops! Will fix!

csharp/ql/lib/semmle/code/csharp/Type.qll
michaelnebel: Alright - I will kill the override.

csharp/ql/test/library-tests/dataflow/fields/J.cs
michaelnebel: Good idea! I will do that!

csharp/ql/lib/semmle/code/csharp/Type.qll
hvitved: Given the name `RecordType`, I would expect it to include all `record` types (suppose `record interface` became a thing). Also, we never use `abstract` classes in user-facing libraries for closed unions, as a user writing `extends` would than mistakenly be adding to the type, rather than refining it.

csharp/extractor/Semmle.Extraction.CSharp/Entities/Modifier.cs
```diff
@@ -111,6 +111,9 @@ public static void ExtractModifiers(Context cx, TextWriter trapFile, IEntity key
                 if (nt is null)
                     throw new InternalError(symbol, "Symbol kind is inconsistent with its type");
 
+                if (nt.IsRecord)
```
hvitved, line 114: This requires, in principle, a DB upgrade script, since DBs created before this extractor change should have `record` modifiers populated whenever a clone method exists. Upgrades without DB scheme changes can be done by making a dummy change to the DB scheme, namely updating the date in the header of the file.

csharp/ql/lib/semmle/code/csharp/Type.qll
michaelnebel: Ah yes, you are absolutely right - exposing abstract towards clients is not a good idea! I didn't think about that! Thank you for the clarification!

csharp/extractor/Semmle.Extraction.CSharp/Entities/Modifier.cs
michaelnebel: Aha, I see. So providing a "dummy" upgrade script (and now also downgrade script) will re-create the database to ensure the consistency with the change of semantics.

michaelnebel: Nice catch btw!

hvitved: I just realized that this can't be done, since we can't create a new `@modifier` entity for `record` in an upgrade script. So let's just not touch the DB scheme on this PR.

hvitved (approved):

PR #7611, opened 2022-01-16
Ruby: flag up `protect_from_forgery` calls without an exception strategy

alexrford:
Calling `protect_from_forgery` without specifying a protection strategy defaults to nulling the `session`. However, if parts of the session are memoized (e.g. in previous `before_action`s), then this may not be sufficient to avoid CSRF attacks, as this memoization can take place before the session is nulled.

See https://www.veracode.com/blog/managing-appsec/when-rails-protectfromforgery-fails for details

### DIFF
--- a/ruby/ql/test/query-tests/security/cwe-352/railsapp/app/controllers/application_controller.rb
+++ b/ruby/ql/test/query-tests/security/cwe-352/railsapp/app/controllers/application_controller.rb
@@ -1,2 +1,20 @@
 class ApplicationController < ActionController::Base
+
+  # BAD: `protect_from_forgery` without `with: :exception` can expose an
+  # application to CSRF attacks in some circumstances
+  protect_from_forgery
+
+  before_action authz_guard
+
+  def current_user
+    @current_user ||= User.find_by_id(session[:user_id])
+  end
+
+  def logged_in?
+    !current_user.nil?
+  end
+
+  def authz_guard
+    render(plain: "not logged in") unless logged_in?
+  end
 end
--- a/ruby/ql/test/query-tests/security/cwe-352/railsapp/app/controllers/articles_controller.rb
+++ b/ruby/ql/test/query-tests/security/cwe-352/railsapp/app/controllers/articles_controller.rb
@@ -1,0 +1,15 @@
+class ArticlesController < ApplicationController
+  prepend_before_action :user_authored_article?, only: [:delete_authored_article]
+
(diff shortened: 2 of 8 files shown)


ruby/ql/test/query-tests/security/cwe-352/railsapp/app/controllers/articles_controller.rb
```diff
@@ -0,0 +1,15 @@
...
+  end
+
+  def article
+    @article ||= Article.find(params[:article_id])
+  end
+
+  def user_authored_article?
+    @article.author_id = current_user.id
+  end
+end
```
hmac, line 15: Is this file here to demonstrate the danger of having memoised values (`current_user`) and weakened CSRF settings? As far as I can tell the tests don't actually depend on this file, right?

ruby/ql/test/query-tests/security/cwe-352/railsapp/app/controllers/application_controller.rb
```diff
@@ -1,2 +1,20 @@
 class ApplicationController < ActionController::Base
+
+  # BAD: `protect_from_forgery` without `with: :exception` can expose an
+  # application to CSRF attacks in some circumstances
+  protect_from_forgery
```
hmac, line 5: It would be good to have an example controller with `protect_from_forgery with: :exception` to confirm that the query doesn't flag that.

hmac (commented):
This looks good to me. Just had a couple of comments. Since we're expanding the range of this query, I guess we might as well add a change note too?

Ping me when you're ready for a re-review

ruby/ql/test/query-tests/security/cwe-352/railsapp/app/controllers/articles_controller.rb
alexrford: Yeah, the file is technically not necessary. It's just for some context into how the non-`with: :exception` calls can be vulnerable.

ruby/ql/test/query-tests/security/cwe-352/railsapp/app/controllers/application_controller.rb
alexrford: Added a case of this (with a new action) to the `ArticlesController`.

alexrford:
@hmac thanks for the review - updated.

hmac (approved):
Nice!

Issue #7601, opened 2022-01-14
General issue - RandonUsedOnce.ql treated as a critical security issue


garethhurley:
**Description of the issue**
The following query is reported in Github alerts as a Critical severity security issue (with a severity of 9.8).
However there is no 'security' tag.

The impact is that other tools (e.g. LGTM) that parse the query differently do not treat this as a security issue.  Can this be clarified?

https://github.com/github/codeql/blob/main/java/ql/src/Likely%20Bugs/Arithmetic/RandomUsedOnce.ql

smowton:
Seems like this should be fixed, raised https://github.com/github/codeql/pull/7613

PR #7613, opened 2022-01-17
Remove security-severity tag to java/random-used-once

smowton:
Raised in https://github.com/github/codeql/issues/7601, this is one of the only .ql files that has a security-severity score but not the tag "security", including many other queries that live outside the `Security/` subdirectory.

Besides this the only other files with this security-severity-but-no-security-tag combination are:

```
java/ql/src/Frameworks/JavaEE/EJB/EjbContainerInterference.ql
java/ql/src/Frameworks/JavaEE/EJB/EjbFileIO.ql
java/ql/src/Frameworks/JavaEE/EJB/EjbNative.ql
java/ql/src/Frameworks/JavaEE/EJB/EjbReflection.ql
java/ql/src/Frameworks/JavaEE/EJB/EjbSecurityConfiguration.ql
java/ql/src/Frameworks/JavaEE/EJB/EjbSerialization.ql
java/ql/src/Frameworks/JavaEE/EJB/EjbSetSocketOrUrlFactory.ql
```

Given their location I'm assuming these queries are disabled by default and likely shouldn't changed?

### DIFF
--- a/java/ql/src/Likely Bugs/Arithmetic/RandomUsedOnce.ql
+++ b/java/ql/src/Likely Bugs/Arithmetic/RandomUsedOnce.ql
@@ -4,12 +4,10 @@
  *              guarantee an evenly distributed sequence of random numbers.
  * @kind problem
  * @problem.severity warning
- * @security-severity 9.8
  * @precision medium
  * @id java/random-used-once
  * @tags reliability
  *       maintainability
- *       security
  *       external/cwe/cwe-335
  */
 
--- a/java/ql/src/Likely Bugs/Arithmetic/RandomUsedOnce.ql
+++ b/java/ql/src/Likely Bugs/Arithmetic/RandomUsedOnce.ql
@@ -9,6 +9,7 @@
  * @id java/random-used-once
  * @tags reliability
  *       maintainability
+ *       security
  *       external/cwe/cwe-335
  */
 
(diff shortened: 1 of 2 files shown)


atorralba:
I think that the actual problem was that the query had been (incorrectly IMHO) assigned a CWE, and because of that a security-severity score was automatically assigned to it. I don't think this query actually models `CWE-335: Incorrect Usage of Seeds in Pseudo-Random Number Generator (PRNG)` (or a security problem at all), so in my opinion we should instead remove the CWE and security-severity tags.

smowton:
If the complaint in the query description (a less-than-even distribution) is true, then that would make random numbers easier to predict than they should be. That seems like a (perhaps mild) security issue to me.

atorralba (dismissed):
Ok, I saw the `readability` and `maintainability` tags and assumed that this was a code quality matter, but after reading more about it I agree it _may_ be a low-severity security issue (because I think this would be very difficult to exploit practically). I also think it shouldn't have a severity of 9.8, but I'm hesitant to change that because it'll probably be changed back the next time the automation runs.

smowton:
Yeah that does seem dubious. @aschackmull do you have an opinion?

aschackmull:
I think the query likely does a decent job as a code quality query, but a poor job as a security query, so for that reason I think the simplest solution is to just consider it to be the former and not the latter.

smowton:
Changed this to remove the security-severity score, leaving the tagging in the same state as the C# query of the same name.

aschackmull (approved):

PR #7614, opened 2022-01-17
Ruby: add more Array/Enumerable flow summaries

nickrolfe:
WIP


nickrolfe:
I've added a commit to enable taint-flow checks in the array-flow test, but it fails with this unexpected result, that I am unable to explain:

```
| array_flow.rb:4:10:4:13 | ...[...] | Unexpected result: hasTaintFlow=0.1 |
```

ruby/ql/lib/codeql/ruby/frameworks/Array.qll
```diff
@@ -0,0 +1,2407 @@
...
+    }
+
+    predicate debugDeleteMe(MethodCall c, string input, string output, int s, int e, int ln) {
+      c = mc and
+      s = start and
+      e = end and
+      propagatesFlowExt(input, output, _) and
+      ln = mc.getLocation().getStartLine()
+    }
+  }
```
hmac, line 1399: I guess this wants to go before merging

ruby/ql/lib/codeql/ruby/frameworks/Array.qll
```diff
@@ -0,0 +1,2407 @@
...
+
+    override predicate propagatesFlowExt(string input, string output, boolean preservesValue) {
+      preservesValue = true and
+      (
+        input = "ArrayElement[?] of Receiver" and
+        output = "ArrayElement[?] of ReturnValue"
+        or
+        exists(ArrayIndex elementIndex, int argIndex |
+          argIndex in [0 .. mc.getNumberOfArguments() - 1] and
+          elementIndex = mc.getArgument(argIndex).getConstantValue().getInt()
```
hmac, line 1460: Out of interest, how does this line differ from this alternative?
```ql
elementIndex = getKnownArrayElementContent(mc.getArgument(argIndex))
```

ruby/ql/lib/codeql/ruby/frameworks/Array.qll
```diff
@@ -0,0 +1,2407 @@
...
+
+private class ArrayIndex extends int {
+  ArrayIndex() { this = any(DataFlow::Content::KnownArrayElementContent c).getIndex() }
+}
+
+/**
+ * Provides flow summaries for the `Array` class.
+ *
+ * The summaries are ordered (and implemented) based on
+ * https://ruby-doc.org/core-3.1.0/Array.html, however for methods that have the
```
hmac, line 17: I think we should probably stop using `ruby-doc.org` as a source of truth. We've encountered several spurious entries already, and I don't think it's an official project.

https://docs.ruby-lang.org/en/3.1/Array.html is probably a better source.

hmac (commented):
This is a mammoth piece of work @nickrolfe! I'm impressed and also a bit terrified 😅

I've reviewed most of the `Array` summaries. I'll take a look at `Enumerable` tomorrow.

I spent a bit of time trying to find the source of that spurious taint result, but got nowhere I'm afraid.

ruby/ql/lib/codeql/ruby/frameworks/Array.qll
```diff
@@ -0,0 +1,2407 @@
...
+
+    override predicate propagatesFlowExt(string input, string output, boolean preservesValue) {
+      input = "ArrayElement of Receiver" and
+      output =
+        [
+          "ArrayElement[?] of ReturnValue", "ArrayElement[?] of Receiver",
+          "Parameter[0] of BlockArgument"
+        ] and
+      preservesValue = true
+    }
```
hmac, line 2098: Confusingly, this is actually a method on `Array`, not `Enumerable`. I think maybe because it modifies the receiver? 🤷

ruby/ql/lib/codeql/ruby/frameworks/Array.qll
```diff
@@ -0,0 +1,2407 @@
...
+  }
+
+  private class SliceBeforeAfterSummary extends SimpleSummarizedCallable {
+    SliceBeforeAfterSummary() { this = ["slice_before", "slice_after"] }
+
+    override predicate propagatesFlowExt(string input, string output, boolean preservesValue) {
+      input = "ArrayElement of Receiver" and
+      output = "Parameter[0] of BlockArgument" and
+      preservesValue = true
+    }
```
hmac, line 2124:
```rb
[1,2,3].slice_before { false }.to_a
=> [[1, 2, 3]]
[1,2,3].slice_before { true }.to_a
=> [[1], [2], [3]]
```
`ArrayElement of Receiver` -> `ArrayElement[?] of ArrayElement[?] of ReturnValue`

ruby/ql/lib/codeql/ruby/frameworks/Array.qll
```diff
@@ -0,0 +1,2407 @@
...
+  }
+
+  private class SliceWhenSummary extends SimpleSummarizedCallable {
+    SliceWhenSummary() { this = "slice_when" }
+
+    override predicate propagatesFlowExt(string input, string output, boolean preservesValue) {
+      input = "ArrayElement of Receiver" and
+      output = ["Parameter[0] of BlockArgument", "Parameter[1] of BlockArgument"] and
+      preservesValue = true
```
hmac, line 2133: Same here:
`ArrayElement of Receiver` -> `ArrayElement[?] of ArrayElement[?] of ReturnValue`

ruby/ql/lib/codeql/ruby/frameworks/Array.qll
```diff
@@ -0,0 +1,2407 @@
...
+        [
+          "Parameter[0] of BlockArgument", "Parameter[1] of BlockArgument",
+          "ArrayElement[?] of ReturnValue"
+        ] and
+      preservesValue = true
+    }
+  }
+
+  private class SortBangSummary extends SimpleSummarizedCallable {
+    SortBangSummary() { this = "sort!" }
```
hmac, line 2152: Like `reject!`, this is a method on `Array`.

ruby/ql/lib/codeql/ruby/frameworks/Array.qll
```diff
@@ -0,0 +1,2407 @@
...
+
+    override predicate propagatesFlowExt(string input, string output, boolean preservesValue) {
+      input = "ArrayElement of Receiver" and
+      output = ["Parameter[0] of BlockArgument", "ArrayElement[?] of ReturnValue"] and
+      preservesValue = true
+    }
+  }
+
+  private class SortByBangSummary extends SimpleSummarizedCallable {
+    SortByBangSummary() { this = "sort_by!" }
```
hmac, line 2181: I'll stop adding these comments - suffice to say that I think all the bang method variants are on `Array`, not `Enumerable`.

ruby/ql/lib/codeql/ruby/frameworks/Array.qll
```diff
@@ -0,0 +1,2407 @@
...
+
+    override predicate propagatesFlowExt(string input, string output, boolean preservesValue) {
+      input = "Receiver" and
+      output = "ReturnValue" and
+      preservesValue = true
+    }
+  }
+
+  private class TransposeSummary extends SimpleSummarizedCallable {
+    TransposeSummary() { this = "transpose" }
```
hmac, line 2286: This is an Array method too.

ruby/ql/lib/codeql/ruby/frameworks/Array.qll
```diff
@@ -0,0 +1,2407 @@
...
+
+    override predicate propagatesFlowExt(string input, string output, boolean preservesValue) {
+      preservesValue = true and
+      (
+        input = "ArrayElement[?] of ArrayElement[?] of Receiver" and
+        output = "ArrayElement[?] of ArrayElement[?] of ReturnValue"
+        or
+        exists(ArrayIndex i, ArrayIndex j |
+          input = "ArrayElement[" + i + "] of ArrayElement[" + j + "] of Receiver" and
+          output = "ArrayElement[" + j + "] of ArrayElement[" + i + "] of ReturnValue"
```
hmac, line 2296: It's cool we can express something as complex as this 👍

ruby/ql/lib/codeql/ruby/frameworks/Array.qll
```diff
@@ -0,0 +1,2407 @@
...
+        )
+      )
+    }
+  }
+
+  private class UnionSummary extends SummarizedCallable {
+    MethodCall mc;
+
+    UnionSummary() { this = "union" and mc.getMethodName() = this }
```
hmac, line 2305: Array method.

ruby/ql/lib/codeql/ruby/frameworks/Array.qll
```diff
@@ -0,0 +1,2407 @@
...
+
+    override predicate propagatesFlowExt(string input, string output, boolean preservesValue) {
+      (
+        // receiver[i] -> return_value[i][0]
+        exists(ArrayIndex i |
+          input = "ArrayElement[" + i + "] of Receiver" and
+          output = "ArrayElement[0] of ArrayElement[" + i + "] of ReturnValue"
+        )
+        or
+        // receiver[?] -> return_value[0][?]
```
hmac, line 2388: `return_value[?][0]` I think

hmac (commented):
Pretty much reviewed everything now. I think there's a few missing flow steps and some other super minor things but otherwise it looks 👌

hvitved:
> I've added a commit to enable taint-flow checks in the array-flow test, but it fails with this unexpected result, that I am unable to explain:

This is expected, and it happens because `*` has a direct flow step from its argument to the return value (`x` may already be an array in `*x`). And all array read steps are also lifted to taint-steps (to account for the case where a taint flow source is an array), so we should simply tag this result as expected.

ruby/ql/lib/codeql/ruby/frameworks/Array.qll
```diff
@@ -0,0 +1,2407 @@
...
+        or
+        exists(ArrayIndex i, ArrayIndex j |
+          input = "ArrayElement[" + i + "] of ArrayElement[" + j + "] of Receiver" and
+          output = "ArrayElement[" + j + "] of ArrayElement[" + i + "] of ReturnValue"
+        )
+      )
+    }
+  }
+
+  private class UnionSummary extends SummarizedCallable {
```
hvitved, line 2302: Can be made a `SimpleSummarizedCallable`

ruby/ql/lib/codeql/ruby/frameworks/StandardLibrary.qll
```diff
@@ -1513,53 +1489,484 @@ module Enumerable {
...
+
+    bindingset[this]
+    MinOrMaxSummary() { mc.getMethodName() = ["min", "max"] }
+
+    override MethodCall getACall() { result = mc }
+  }
+
+  private class MinOrMaxNoArgNoBlockSummary extends MinOrMaxSummary {
+    MinOrMaxNoArgNoBlockSummary() {
+      this = "min_or_max_no_arg_no_block" and
```
hvitved: I think I would prefer `this = mc.getMethodName() + "_no_arg_no_block`. This way we will synthesize two different methods for calls to `min` and `max`.

hvitved: This comment may apply to other classes as well.

ruby/ql/lib/codeql/ruby/frameworks/StandardLibrary.qll
```diff
@@ -712,7 +712,7 @@ module Array {
     ElementReferenceRangeReadKnownSummary() {
       mc.getNumberOfArguments() = 2 and
       start = getKnownArrayElementContent(mc.getArgument(0)).getIndex() and
-      exists(int length | length = mc.getArgument(1).getValueText().toInt() |
+      exists(int length | length = mc.getArgument(1).getConstantValue().getInt() |
```
hvitved: Could also use `mc.getArgument(1).getConstantValue().isInt(length)` (which I think I slightly prefer).

ruby/ql/lib/codeql/ruby/frameworks/StandardLibrary.qll
```diff
@@ -762,16 +762,16 @@ module Array {
       (
         mc.getNumberOfArguments() = 2 and
         (
-          not exists(mc.getArgument(0).getValueText().toInt()) or
-          not exists(mc.getArgument(1).getValueText().toInt())
+          not exists(mc.getArgument(0).getConstantValue().getInt()) or
```
hvitved: `not mc.getArgument(0).getConstantValue().isInt(_)` is a bit shorter

hvitved (commented):
Very impressive and thorough work Nick. I have only spot-checked a few summaries, and they looked fine.

ruby/ql/lib/codeql/ruby/frameworks/Array.qll
```diff
@@ -0,0 +1,2407 @@
...
+        input = "ArrayElement of Receiver" and
+        output = "ArrayElement[?] of Parameter[0] of BlockArgument"
+        or
+        input = "ArrayElement[?] of Receiver" and
+        output = "ArrayElement[?] of ReturnValue"
+        or
+        exists(ArrayIndex i |
+          input = "ArrayElement[" + i + "] of Receiver" and
+          output = "ArrayElement[" + i + "] of ReturnValue"
+        )
```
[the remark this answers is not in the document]
nickrolfe: @hvitved – you wrote this one, originally. Any reason not to make the simplification @hmac is suggesting?

ruby/ql/lib/codeql/ruby/frameworks/Array.qll
nickrolfe: Good catch. I've moved it.

nickrolfe (commented):
I've addressed a lot of the feedback. Thanks @hmac and @hvitved!

To summarise the remaining issues:
1. So far, the summaries just kind of ignore the methods (or variants) that return enumerators. @hmac is effectively proposing that we treat them as if they were arrays. I'm not convinced.
    - Side note: a lot of the methods return `self` if given a block, and an enumerator otherwise. We're currently ignoring the latter variants, but I think the summaries we do have for the block variants should be restricted with `exists(mc.getBlock())`.
2. I couldn't make @hvitved's suggested changes to use `SimpleSummarizedCallable` without introducing non-monotonic recursion errors. Is there a better way to avoid those?

nickrolfe:
Oh, I see I missed some of Tom's comments, because they were made on an earlier commit, before I moved everything to a new file.

ruby/ql/lib/codeql/ruby/frameworks/StandardLibrary.qll
nickrolfe: Sure, I've done that for this and other classes that covered multiple methods.

ruby/ql/lib/codeql/ruby/frameworks/Array.qll
```diff
@@ -0,0 +1,2407 @@
...
+      )
+    }
+
+    override predicate clearsContent(ParameterPosition pos, DataFlow::Content content) {
+      pos.isSelf() and
+      content instanceof DataFlow::Content::KnownArrayElementContent
+    }
+  }
+
+  private class ProductSummary extends SummarizedCallable {
```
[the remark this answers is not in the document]
hvitved: I'll push a fix.

ruby/ql/lib/codeql/ruby/frameworks/Array.qll
```diff
@@ -0,0 +1,2407 @@
...
+    }
+  }
+
+  private class IndexSummary extends SimpleSummarizedCallable {
+    IndexSummary() { this = ["index", "rindex"] }
+
+    override predicate propagatesFlowExt(string input, string output, boolean preservesValue) {
+      input = "ArrayElement of Receiver" and
+      output = "Parameter[0] of BlockArgument" and
+      preservesValue = true
```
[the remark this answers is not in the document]
hmac: 🤦 you're totally right, dunno what I was thinking here...

ruby/ql/test/library-tests/dataflow/array-flow/array_flow.rb
```diff
@@ -145,379 +163,1411 @@ def m17
...
     end
+    sink(b[0]) # $ hasValueFlow=28.1 $ hasValueFlow=28.2
 end
 
 def m29
     a = [0, 1, source(29)]
-    a.cycle(2) do |x|
-        sink x # $ hasValueFlow=29
+    a.combination(1) do |x|
+        sink(x[0]) # $ hasValueFlow=29
```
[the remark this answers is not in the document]
hmac: Sorry, I was sloppy with my description here. Without a block it returns an enumerator, but the enumerator's elements are combinations, i.e.
```ruby
pry(main)> [1,2,3].combination(2).to_a
=> [[1, 2], [1, 3], [2, 3]]
```

So it's the same deal as with the others, where we might want to pretend the enumerator is an array. A [quick search](https://cs.github.com/?scopeName=All+repos&scope=&q=.combination%28) indicates it's common to do things like `foo.combination(n).to_a` or `foo.combination(n).each`, so we'd need `ArrayElement of Receiver` -> `ArrayElement[?] of ArrayElement[?] of ReturnValue` to track flow through these expressions.

ruby/ql/test/library-tests/dataflow/array-flow/array_flow.rb
```diff
@@ -145,379 +163,1411 @@ def m17
...
+    a.count do |x|
+        sink x # $ hasValueFlow=33
+    end
+end
+
+def m34
+    a = [0, 1, source(34)]
+    a.cycle(2) do |x|
+        sink x # $ hasValueFlow=34
+    end
```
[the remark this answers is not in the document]
hmac: :+1: sorry for the confusing comment, I think I meant to tack a `.to_a` on to the end of the first line. This sounds like the same are-enumerators-arrays issue we've discussed in a few comments, so whatever we decide re that would apply here too.

hmac:
I think the only remaining question mark is what we do about the methods that return enumerators. My argument is we should track flow through them as if they were arrays, so that expressions like `arr.each_cons(2).map { |x, y| sink y }` will work. [Here](https://cs.github.com/?q=%2F%5C.%28chain%7Cchunk%7Ccollect%7Ccollect_concat%7Cdrop_while%7Cmap%7Creject%7Ccombination%7Cchunk_while%7Ccycle%7Cdetect%7Cdrop_while%7Ceach_cons%7Ceach_entry%7Ceach_slice%7Ceach_with_index%7Ceach_with_object%7Cfilter_map%7Cfind_all%7Cfind_index%7Cflat_map%7Cgroup_by%7Clazy%7Cmax_by%7Cmin_by%7Cminmax_by%7Cpartition%7Creject%7Creverse_each%7Cslice_after%7Cslice_before%7Cslice_when%7Csort_by%7Ctake_while%29%5C%28%5Cw%2B%5C%29%5C.%2F+language%3ARuby&scopeName=All+repos) are some examples of this kind of enumerator chaining in real code.

asgerf:
+1 for treating enumerables as arrays, or more specifically: treating `ArrayElement` to mean the content of any kind of enumerable (maybe it could then be renamed to something more appropriate).

We could also use an access path token `Element` to mean any kind of enumerable content, but then initially implement that as an alias for `ArrayElement`. That way, the summary rows are accurate, and there's a single place to update if it is later deemed important to distinguish arrays from other enumerables.

asgerf:
Does the enumerable stuff need to be in this PR though? I'd would simplify the access path rewrite if this PR gets merged ASAP.

hvitved:
> maybe it could then be renamed to something more appropriate

It will be renamed as part of my work in implementing flow through hashes.

hvitved:
> However, there were two questions for @hvitved that I would not like to be missed:

D'oh, my comments were still pending.

> 1. Is there any meaningful difference berween elementIndex = mc.getArgument(argIndex).getConstantValue().getInt() and elementIndex = getKnownArrayElementContent(mc.getArgument(argIndex))?

No.

> Is there any reason not to make @hmac's suggested simplification of EachSliceSummary?

That would allow flow in this case, which we don't want:
```rb
x = "tainted"
y = x.each_slice
sink(x)
```

hvitved (approved):

hmac (approved):
Awesome work 👍

PR #7615, opened 2022-01-17
QL: support this.method() calls in the charpred that references non-extending supertypes

erik-krogh:
See the below example.  
I've added support for the `this might be a compile error?` line. 

If it turns out that it's actually supposed to be a compile-error, then we'll revert this PR. 

```CodeQL
class Expr extends int {
  Expr() { this = 1 or this = 2 }
}

class Function extends int {
  Function() { this = 1 }

  int getParameter(int i) { result = i * this }
}

class Foo extends Expr instanceof Function {
  Foo() {
    exists(super.getParameter(0))
    or
    exists(this.getParameter(0)) // <- this might be a compile error? But currently the compiler doesn't think so. 
  }

  predicate bar() {
    exists(super.getParameter(0))
    or
    exists(this.getParameter(0)) // <- this is a compile error, everyone agrees on that. 
  }
}
```

### DIFF
--- a/ql/ql/src/codeql_ql/ast/internal/Predicate.qll
+++ b/ql/ql/src/codeql_ql/ast/internal/Predicate.qll
@@ -77,8 +77,12 @@ private module Cached {
       p = t.getClassPredicate(mc.getMemberName(), mc.getNumberOfArguments())
     )
     or
-    // super calls
-    exists(Super sup, ClassType type, Type supertype |
+    // super calls - and `this.method()` calls in charpreds (for confusing reasons)
+    exists(AstNode sup, ClassType type, Type supertype |
+      sup instanceof Super
+      or
+      sup.(ThisAccess).getEnclosingPredicate() instanceof CharPred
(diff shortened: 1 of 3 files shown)


ql/ql/src/codeql_ql/ast/internal/Predicate.qll
```diff
@@ -77,8 +77,12 @@ private module Cached {
       p = t.getClassPredicate(mc.getMemberName(), mc.getNumberOfArguments())
     )
     or
-    // super calls
-    exists(Super sup, ClassType type, Type supertype |
+    // super calls - and `this.method()` calls in charpreds (for confusing reasons)
```
MathiasVP, line 80: I'm not really a big fan of "for confusing reasons". Nor do I think this is the right place to tell the whole story. Is this PR unblocking anything? If not, shouldn't we just wait and see the what Core team decides on this?

MathiasVP (approved):

PR #7616, opened 2022-01-17
JS: Add query help for ML-powered queries

henrymercer:
Query help is identical to the original query, except for a new paragraph prepended to the overview explaining that the queries are experimental.

We add Markdown query help since only Markdown query help is embedded in SARIF via `--sarif-add-query-help`.

### DIFF
--- a/javascript/ql/experimental/adaptivethreatmodeling/src/NosqlInjectionATM.md
+++ b/javascript/ql/experimental/adaptivethreatmodeling/src/NosqlInjectionATM.md
@@ -1,0 +1,50 @@
+# NoSQL database query built from user-controlled sources (experimental)
+This is an experimental query. Experimental queries generate alerts using machine learning. They might include more false positives but they will improve over time.
+
+If a database query (such as a SQL or NoSQL query) is built from user-provided data without sufficient sanitization, a malicious user may be able to run malicious database queries.
+
+
+## Recommendation
+Most database connector libraries offer a way of safely embedding untrusted data into a query by means of query parameters or prepared statements.
+
+For NoSQL queries, make use of an operator like MongoDB's `$eq` to ensure that untrusted data is interpreted as a literal value and not as a query object.
+
+
+## Example
+In the following example, assume the function `handler` is an HTTP request handler in a web application, whose parameter `req` contains the request object.
+
+The handler constructs two copies of the same SQL query involving user input taken from the request object, once unsafely using string concatenation, and once safely using query parameters.
+
+In the first case, the query string `query1` is built by directly concatenating a user-supplied request parameter with some string literals. The parameter may include quote characters, so this code is vulnerable to a SQL injection attack.
+
+In the second case, the parameter is embedded into the query string `query2` using query parameters. In this example, we use the API offered by the `pg` Postgres database connector library, but other libraries offer similar features. This version is immune to injection attacks.
+
+
+```javascript
+const app = require("express")(),
+      pg = require("pg"),
+      pool = new pg.Pool(config);
+
+app.get("search", function handler(req, res) {
+  // BAD: the category might have SQL special characters in it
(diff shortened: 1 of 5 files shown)


esbena:
I'm aware that this is in the experimental folder, but this way of creating near-duplicate documentation does not seem maintainable to me.

1. Is this really a required caveat to have in the qhelp?
2. If yes to 1: would it be possible to create a script that generates these markdown files with the added line at the top?


> We add Markdown query help since only Markdown query help is embedded in SARIF via --sarif-add-query-help

Is this a special thing due to the `experimental` folder, or why does documentation work differently from the ordinary queries?

henrymercer:
> I'm aware that this is in the experimental folder, but this way of creating near-duplicate documentation does not seem maintainable to me.

I agree that this approach is non-ideal, but I would remark that the relevant query helps have only changed 1, 2, 2 times in the past 3 years for SQL/NoSQL, path injection, and XSS respectively.  As such I expect the maintenance burden of updating them will be less than the initial engineering effort and maintenance burden of a script to generate them.



> Is this a special thing due to the `experimental` folder, or why does documentation work differently from the ordinary queries?

This is because we don't deploy the experimental ML-powered packs to Code Scanning in the same way we do for the standard packs.  The deployment process for the standard packs includes a stage where the qhelp for all of the standard queries is compiled into a SARIF file which is loaded into Turboscan.  However since the ML-powered packs are instead downloaded when required, we need to populate the SARIF with the query help via the CodeQL Action.  Currently the CodeQL CLI only supports populating the SARIF with query help that's already been compiled to Markdown.

javascript/ql/experimental/adaptivethreatmodeling/src/NosqlInjectionATM.md
```diff
@@ -0,0 +1,50 @@
+# NoSQL database query built from user-controlled sources (experimental)
+This is an experimental query. Experimental queries generate alerts using machine learning. They might include more false positives but they will improve over time.
+
+If a database query (such as a SQL or NoSQL query) is built from user-provided data without sufficient sanitization, a malicious user may be able to run malicious database queries.
```
Z80coder, line 4: Perhaps this should be "If a NoSQL query is built ..."

Z80coder: Ah I see the problem ... this is 1 standard query that we split into 2 boosted queries. So we don't have 2 original docs to copy and paste. Therefore please ignore the following comment.

javascript/ql/experimental/adaptivethreatmodeling/src/codeql-suites/javascript-atm-code-scanning.qls
```diff
@@ -1,8 +1,2 @@
 - description: ATM boosted Code Scanning queries for JavaScript
 - queries: .
-- include:
-    id: 
-    - adaptive-threat-modeling/js/nosql-injection
-    - adaptive-threat-modeling/js/sql-injection
-    - adaptive-threat-modeling/js/path-injection
-    - adaptive-threat-modeling/js/xss
```
Z80coder, line 8: Why has this been removed in this PR?

Z80coder (approved):
Just one comment to double-check, otherwise LGTM.

javascript/ql/experimental/adaptivethreatmodeling/src/codeql-suites/javascript-atm-code-scanning.qls
henrymercer: Yeah arguably that commit should have been in a separate PR.  The justification is:

> It's simpler to just run all the queries in the pack instead of specifying the IDs.

esbena (approved):

PR #7617, opened 2022-01-17
JS: Update alert messages for ML-powered queries

henrymercer:
We follow a template of `(Experimental) <alert message for original query> Identified using machine learning.`, but also modify the original message to indicate uncertainty, for example "This is a path that depends on a user-provided value." -> "This may be a path that depends on a user-provided value."

### DIFF
--- a/javascript/ql/experimental/adaptivethreatmodeling/src/NosqlInjectionATM.ql
+++ b/javascript/ql/experimental/adaptivethreatmodeling/src/NosqlInjectionATM.ql
@@ -16,15 +16,11 @@ import ATM::ResultsInfo
 import DataFlow::PathGraph
 import experimental.adaptivethreatmodeling.NosqlInjectionATM
 
-from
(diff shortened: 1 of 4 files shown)


Z80coder (commented):
Will an example message be:

"(Experimental) This may be a database query that depends on a user-provided value. Identified using machine learning. 0.567"

?

In other words, the `score` will also be present?

henrymercer:
> Will an example message be:
> 
> "(Experimental) This may be a database query that depends on a user-provided value. Identified using machine learning. 0.567"
> 
> ?
> 
> In other words, the `score` will also be present?

The message will be "(Experimental) This may be a database query that depends on a user-provided value. Identified using machine learning.", i.e. not including the score.  The score will be included in the SARIF metadata about the alert, but it isn't part of the alert message, and won't (yet) be displayed in the UI.

Z80coder (approved):
LGTM

PR #7626, opened 2022-01-18
JS: add query for detecting insecure temporary files

erik-krogh:
To safely create a temporary file in the OS temp dir in a multi-user environment you must: 
- Ensure that the file is created with `0o600` permissions or similar. 
- Ensure that you're not reusing an existing file (Use [the `O_EXCL` and `O_CREAT` flags](https://nodejs.org/api/fs.html#fs_fs_constants))   
  (An exploit requires extremely precise timing by inserting a file between an `fs.exists(..)` and a `fs.open(..)` call.) 

This query checks for the first of these requirements.   

The only way to ensure the second requirement is to use `fs.open(.., O_EXCL | O_CREAT)`, which is what e.g. [`tmp`](https://www.npmjs.com/package/tmp) does.  
Checking for the second requirement gets complicated. One of the complications is that NodeJS programs very rarely use file handles, and instead just calls the `fs` module multiple times with the same string `path`.   
So I went the easy route and just look for bad permissions, which is the most important part.  
(But the recommendation in the `.qhelp` basically says: use [`tmp`](https://www.npmjs.com/package/tmp). Because that package does everything right). 

I haven't found any FPs.  
But I suspect that FPs might happen when a program safely creates a temp file in a utility method, but the query still sees that utility method as a source. 

[An evaluation](https://github.com/github/codeql-dca-main/tree/data/erik-krogh/CWE-377__default__CustomSuite__1/reports) shows no new results, and a constant 30s slowdown (that's the extra time it takes to compile the query, something went wrong in a preparation step).   
So the evaluation show no noticeable slowdown. 

[A test run](https://lgtm.com/query/2488387209030888032/) finds some results.  
All the results are either in test code or benign, which is why I've set the precision to `medium`.

### DIFF
--- a/javascript/ql/lib/semmle/javascript/security/dataflow/InsecureTemporaryFileCustomizations.qll
+++ b/javascript/ql/lib/semmle/javascript/security/dataflow/InsecureTemporaryFileCustomizations.qll
@@ -1,0 +1,96 @@
+/**
+ * Provides default sources, sinks and sanitizers for reasoning about
+ * insecure temporary file creation, as well as
+ * extension points for adding your own.
+ */
+
+import javascript
+
+/**
+ * Classes and predicates for reasoning about insecure temporary file creation.
+ */
+module InsecureTemporaryFile {
+  /**
+   * A data flow source for insecure temporary file creation.
+   */
+  abstract class Source extends DataFlow::Node { }
+
+  /**
+   * A data flow sink for insecure temporary file creation.
+   */
+  abstract class Sink extends DataFlow::Node { }
+
+  /**
+   * A sanitizer for random insecure temporary file creation.
+   */
+  abstract class Sanitizer extends DataFlow::Node { }
+
+  /** A call that opens a file with a given path. */
+  class OpenFileCall extends DataFlow::CallNode {
+    string methodName;
+
+    OpenFileCall() {
+      methodName = ["open", "openSync", "writeFile", "writeFileSync"] and
+      this = NodeJSLib::FS::moduleMember(methodName).getACall()
+    }
+
+    DataFlow::Node getPath() { result = this.getArgument(0) }
+
+    DataFlow::Node getMode() {
+      methodName = ["open", "openSync"] and
+      result = this.getArgument(2)
+      or
+      methodName = ["writeFile", "writeFileSync"] and
+      result = this.getOptionArgument(2, "mode")
+    }
+  }
+
+  /** Holds if the `mode` ensure no access to other users. */
+  bindingset[mode]
+  private predicate isSecureMode(int mode) {
+    // the lowest 6 bits should be 0.
+    // E.g. `0o600` is secure (each digit in a octal number is 3 bits)
+    mode.bitAnd(1) = 0 and
+    mode.bitAnd(2) = 0 and
+    mode.bitAnd(4) = 0 and
+    mode.bitAnd(8) = 0 and
+    mode.bitAnd(16) = 0 and
+    mode.bitAnd(32) = 0
+  }
+
+  /** The path in a call that opens a file without specifying a secure `mode`. Seen as a sink for insecure temporary file creation. */
+  class InsecureFileOpen extends Sink {
+    InsecureFileOpen() {
+      exists(OpenFileCall call |
+        not exists(call.getMode())
+        or
+        exists(int mode | mode = call.getMode().getIntValue() | not isSecureMode(mode))
+      |
+        this = call.getPath()
+      )
+    }
+  }
+
+  /** A a string that references the global tmp dir. Seen as a source for insecure temporary file creation. */
+  class OSTempDir extends Source {
+    OSTempDir() {
+      this = DataFlow::moduleImport("os").getAMemberCall("tmpdir")
+      or
+      this.getStringValue().matches("/tmp/%")
+    }
+  }
+
+  /** A non-first leaf in a string-concatenation. Seen as a sanitizer for insecure temporary file creation. */
+  class NonFirstStringConcatLeaf extends Sanitizer {
+    NonFirstStringConcatLeaf() {
+      exists(StringOps::ConcatenationRoot root |
+        this = root.getALeaf() and
+        not this = root.getFirstLeaf()
+      )
+      or
+      exists(DataFlow::CallNode join | join = DataFlow::moduleMember("path", "join").getACall() |
+        this = join.getArgument([1 .. join.getNumArgument() - 1])
+      )
+    }
+  }
+}
--- a/javascript/ql/lib/semmle/javascript/security/dataflow/InsecureTemporaryFileCustomizations.qll
+++ b/javascript/ql/lib/semmle/javascript/security/dataflow/InsecureTemporaryFileCustomizations.qll
@@ -30,7 +30,12 @@ module InsecureTemporaryFile {
     string methodName;
 
     OpenFileCall() {
-      methodName = ["open", "openSync", "writeFile", "writeFileSync"] and
+      methodName =
+        [
+          "open", "openSync", "writeFile", "writeFileSync", "writeJson", "writeJSON",
+          "writeJsonSync", "writeJSONSync", "outputJson", "outputJSON", "outputJsonSync",
+          "outputJSONSync", "outputFile", "outputFileSync"
+        ] and
       this = NodeJSLib::FS::moduleMember(methodName).getACall()
     }
 
@@ -40,7 +45,7 @@ module InsecureTemporaryFile {
       methodName = ["open", "openSync"] and
       result = this.getArgument(2)
       or
-      methodName = ["writeFile", "writeFileSync"] and
+      not methodName = ["open", "openSync"] and
       result = this.getOptionArgument(2, "mode")
     }
   }
@@ -88,7 +93,8 @@ module InsecureTemporaryFile {
         not this = root.getFirstLeaf()
       )
       or
-      exists(DataFlow::CallNode join | join = DataFlow::moduleMember("path", "join").getACall() |
+      exists(DataFlow::CallNode join |
+        join = DataFlow::moduleMember("path", "join").getACall() and
         this = join.getArgument([1 .. join.getNumArgument() - 1])
       )
     }
--- a/javascript/ql/lib/semmle/javascript/security/dataflow/InsecureTemporaryFileCustomizations.qll
+++ b/javascript/ql/lib/semmle/javascript/security/dataflow/InsecureTemporaryFileCustomizations.qll
@@ -71,7 +71,7 @@ module InsecureTemporaryFile {
     }
   }
 
-  /** A a string that references the global tmp dir. Seen as a source for insecure temporary file creation. */
+  /** A string that references the global tmp dir. Seen as a source for insecure temporary file creation. */
   class OSTempDir extends Source {
     OSTempDir() {
       this = DataFlow::moduleImport("os").getAMemberCall("tmpdir")
--- a/javascript/ql/src/Security/CWE-377/InsecureTemporaryFile.qhelp
+++ b/javascript/ql/src/Security/CWE-377/InsecureTemporaryFile.qhelp
@@ -1,0 +1,43 @@
+<!DOCTYPE qhelp PUBLIC "-//Semmle//qhelp//EN" "qhelp.dtd">
+<qhelp>
+
+<overview>
+<p>
+Temporary files created in the operating system tmp directory are by default accessible 
+to other users. This can in some cases lead to information exposure, or in the worst 
+case to remote code execution.
+</p>
+</overview>
+
+<recommendation>
+<p>
+Use a well tested library like <a href="https://www.npmjs.com/package/tmp">tmp</a> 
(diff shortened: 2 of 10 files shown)


javascript/ql/lib/semmle/javascript/security/dataflow/InsecureTemporaryFileCustomizations.qll
```diff
@@ -0,0 +1,96 @@
...
+    mode.bitAnd(32) = 0
+  }
+
+  /** The path in a call that opens a file without specifying a secure `mode`. Seen as a sink for insecure temporary file creation. */
+  class InsecureFileOpen extends Sink {
+    InsecureFileOpen() {
+      exists(OpenFileCall call |
+        not exists(call.getMode())
+        or
+        exists(int mode | mode = call.getMode().getIntValue() | not isSecureMode(mode))
```
kaeluka, line 67: I think this could be simpler if getMode would return an `int` (applying the default value where none is specified explicitly).

javascript/ql/lib/semmle/javascript/security/dataflow/InsecureTemporaryFileCustomizations.qll
```diff
@@ -0,0 +1,96 @@
...
+      |
+        this = call.getPath()
+      )
+    }
+  }
+
+  /** A a string that references the global tmp dir. Seen as a source for insecure temporary file creation. */
```
kaeluka, line 74:
```suggestion
  /** A string that references the global tmp dir. Seen as a source for insecure temporary file creation. */
```

javascript/ql/lib/semmle/javascript/security/dataflow/InsecureTemporaryFileCustomizations.qll
```diff
@@ -0,0 +1,96 @@
...
+  class OSTempDir extends Source {
+    OSTempDir() {
+      this = DataFlow::moduleImport("os").getAMemberCall("tmpdir")
+      or
+      this.getStringValue().matches("/tmp/%")
```
kaeluka, line 79: Does it make sense to include windows paths here as well?

javascript/ql/test/query-tests/Security/CWE-377/InsecureTemporaryFile.expected
```diff
@@ -0,0 +1,53 @@
+nodes
```
kaeluka, line 1: Is the fact that you're using a `qlref` file the a reason you're not using consistency checking (`// OK` / `// NOT OK`) based tests?

javascript/ql/lib/semmle/javascript/security/dataflow/InsecureTemporaryFileCustomizations.qll
```diff
@@ -0,0 +1,96 @@
...
+    }
+  }
+
+  /** A non-first leaf in a string-concatenation. Seen as a sanitizer for insecure temporary file creation. */
```
kaeluka, line 83: I don't understand why this is a sanitizer. Can you elaborate?

javascript/ql/lib/semmle/javascript/security/dataflow/InsecureTemporaryFileCustomizations.qll
```diff
@@ -0,0 +1,96 @@
...
+  class NonFirstStringConcatLeaf extends Sanitizer {
+    NonFirstStringConcatLeaf() {
+      exists(StringOps::ConcatenationRoot root |
+        this = root.getALeaf() and
+        not this = root.getFirstLeaf()
+      )
+      or
+      exists(DataFlow::CallNode join | join = DataFlow::moduleMember("path", "join").getACall() |
```
kaeluka, line 91: should `resolve` also be included here?

javascript/ql/lib/semmle/javascript/security/dataflow/InsecureTemporaryFileCustomizations.qll
```diff
@@ -0,0 +1,96 @@
...
+   * A sanitizer for random insecure temporary file creation.
+   */
+  abstract class Sanitizer extends DataFlow::Node { }
+
+  /** A call that opens a file with a given path. */
+  class OpenFileCall extends DataFlow::CallNode {
+    string methodName;
+
+    OpenFileCall() {
+      methodName = ["open", "openSync", "writeFile", "writeFileSync"] and
```
kaeluka, line 33: there's also functions in `fs-extra` that pbly should be included here (eg, `writeJson`, and I think there's also aliases like `writeJSON`, etc)

kaeluka: writeJson doc: https://github.com/jprichardson/node-fs-extra/blob/master/docs/writeJson.md

kaeluka (changes requested):
I found a few typos and minor issues, but nothing all that serious. Rest LGTM

kaeluka:
@erik-krogh and me just had a small follow-up call; I asked whether `fs.constantc` (https://nodejs.org/api/fs.html#fsconstants) is/should be supported; his reply was that it isn't since that using it implies that users are likely doing it right. That argument makes sense to me.

javascript/ql/lib/semmle/javascript/security/dataflow/InsecureTemporaryFileCustomizations.qll
erik-krogh: People using windows are (hopefully) using the `require("os").tmpdir()` method, so I think that's covered.

javascript/ql/lib/semmle/javascript/security/dataflow/InsecureTemporaryFileCustomizations.qll
erik-krogh: Then I would have to duplicate the `getIntValue()` across the two cases in `getMode()`.  
I think I still prefer this way, where `getMode()` returns the node that has the mode.  

Also, in `InsecureFileOpen()` I check whether a `mode` argument exists. 
I couldn't do that if the predicate returns an `int`.

javascript/ql/lib/semmle/javascript/security/dataflow/InsecureTemporaryFileCustomizations.qll
erik-krogh: I'm only interested in paths that point to somewhere inside the global temp directory.  

If `var source = "/tmp/foo"` is my source, then a path like `var foo = "~/myDir/" + source` is fine, because it no longer points to the global temp directory.  
In general, taint is only preserved in string-concatenations when taint starts from the left-most leaf, because it's a completely different path in the filesystem.

javascript/ql/lib/semmle/javascript/security/dataflow/InsecureTemporaryFileCustomizations.qll
erik-krogh: I'm not sure it should. 
If your path is absolute (and paths to the global temp-dir are likely absolute), then `path.resolve()` will return the absolute path, even though it's not the first argument.

javascript/ql/lib/semmle/javascript/security/dataflow/InsecureTemporaryFileCustomizations.qll
kaeluka: ok, makes sense :)

javascript/ql/lib/semmle/javascript/security/dataflow/InsecureTemporaryFileCustomizations.qll
kaeluka: 👍

javascript/ql/lib/semmle/javascript/security/dataflow/InsecureTemporaryFileCustomizations.qll
kaeluka: What I didn't understand was that you're limiting the analysis to the global temp dir. Just to confirm my understanding: using a tempfile that is contained in `~/myDir/tmp/foo` is not technically safer, right? An attacker still might race on that file. But it's much harder for the attacker to guess the file path to race on.

javascript/ql/lib/semmle/javascript/security/dataflow/InsecureTemporaryFileCustomizations.qll
kaeluka: Related to the thread above — this makes sense under the assumption that we stick to analysing the global temp dir :)

javascript/ql/lib/semmle/javascript/security/dataflow/InsecureTemporaryFileCustomizations.qll
erik-krogh: >  ~/myDir/tmp/foo is not technically safer 

It is safer, because another user doesn't have access to that path.  
The global temp dir is shared between all users, and that's why bad things can happen.

javascript/ql/test/query-tests/Security/CWE-377/InsecureTemporaryFile.expected
erik-krogh: It's just simpler to add a `.qlref` file, that's why.  
No deep reasoning.

kaeluka:
Ah, silly me! Of course :) glad I asked, though

Erik Krogh Kristensen ***@***.***> schrieb am Mi., 2. Feb.
2022, 16:44:

kaeluka (approved):

kaeluka:
I have now approved this change because it's good in my opinion. Take my opinion with a grain of salt. If you want an extra review by someone who's been here longer, that might make sense.

erik-krogh:
> I'll leave that decision to you. 

I think I'll merge it.  
After it has gone through a doc review.

javascript/ql/src/Security/CWE-377/InsecureTemporaryFile.qhelp
```diff
@@ -0,0 +1,43 @@
+<!DOCTYPE qhelp PUBLIC "-//Semmle//qhelp//EN" "qhelp.dtd">
+<qhelp>
+
+<overview>
+<p>
+Temporary files created in the operating system tmp directory are by default accessible
```
mattpollard, line 6:
```suggestion
Temporary files created in the operating system's temporary directory are by default accessible 
```

javascript/ql/src/Security/CWE-377/InsecureTemporaryFile.qhelp
```diff
@@ -0,0 +1,43 @@
+<!DOCTYPE qhelp PUBLIC "-//Semmle//qhelp//EN" "qhelp.dtd">
+<qhelp>
+
+<overview>
+<p>
+Temporary files created in the operating system tmp directory are by default accessible 
+to other users. This can in some cases lead to information exposure, or in the worst
```
mattpollard, line 7:
```suggestion
to other users. In some cases, this can lead to information exposure, or in the worst 
```

javascript/ql/src/Security/CWE-377/InsecureTemporaryFile.qhelp
```diff
@@ -0,0 +1,43 @@
+<!DOCTYPE qhelp PUBLIC "-//Semmle//qhelp//EN" "qhelp.dtd">
+<qhelp>
+
+<overview>
+<p>
+Temporary files created in the operating system tmp directory are by default accessible 
+to other users. This can in some cases lead to information exposure, or in the worst 
+case to remote code execution.
```
mattpollard, line 8:
```suggestion
case, to remote code execution.
```

javascript/ql/src/Security/CWE-377/InsecureTemporaryFile.qhelp
```diff
@@ -0,0 +1,43 @@
...
+</p>
+</overview>
+
+<recommendation>
+<p>
+Use a well tested library like <a href="https://www.npmjs.com/package/tmp">tmp</a>
```
mattpollard, line 14:
```suggestion
Use a well-tested library like <a href="https://www.npmjs.com/package/tmp">tmp</a> 
```

javascript/ql/src/Security/CWE-377/InsecureTemporaryFile.qhelp
```diff
@@ -0,0 +1,43 @@
...
+<p>
+Use a well tested library like <a href="https://www.npmjs.com/package/tmp">tmp</a> 
+for creating temporary files. These libraries ensure both that the file is inaccessible
+to other users and that the file does not already exist.
+</p>
+</recommendation>
+
+<example>
+<p>
+The following example creates a temporary file in the operating system tmp directory.
```
mattpollard, line 22:
```suggestion
The following example creates a temporary file in the operating system's temporary directory.
```

mattpollard (approved):
Hello from Docs 👋🏻 I've left a few minor suggestions for you to consider, but otherwise, the changes looks good to me from a documentation perspective!

kaeluka (approved):
Still looks good!

Issue #7609, opened 2022-01-16
LGTM.com - false positive C# record classes (records) - Local scope variable shadows member


ldeluigi:
**Description of the false positive**
When using recors, a false positive about local variable shadowing pops up.
![image](https://user-images.githubusercontent.com/44567586/149657088-69b3a800-0e1b-4b6f-821d-ef35a8e5ea62.png)

**URL to the alert on the project page on LGTM.com**

https://github.com/e-scooter-2077/rent.manage-scooter-availability/security/code-scanning/16?query=ref%3Arefs%2Fheads%2Fmaster

tausbn:
Thank you for your report!

I agree that this looks like a false positive. I will forward it to the CodeQL C# team.

smowton:
Fixed in https://github.com/github/codeql/pull/7628 -- the fix will show up as of CodeQL 2.12.1, which is estimated to begin rolling out on 24th January.

PR #7628, opened 2022-01-18
C#: Fix false positive alert for shadowing on record types.

michaelnebel:
When defining a record type, the compiler generates a Deconstruct method on the class or struct representing the record type.
This method has out parameters with the same name as property accessors.
In this PR we add this exception to the rule that identifies bad practices related to shadowing.

### DIFF
--- a/csharp/ql/lib/change-notes/2022-01-18-local-shadows-member.md
+++ b/csharp/ql/lib/change-notes/2022-01-18-local-shadows-member.md
@@ -1,0 +1,4 @@
+---
+category: minorAnalysis
+---
+* The query `cs/local-shadows-member` no longer highlights parameters of `record` types.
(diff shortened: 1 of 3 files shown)


michaelnebel:
Fixes https://github.com/github/codeql/issues/7609

hvitved (approved):
Excellent!

csharp/ql/lib/change-notes/2022-01-18-local-shadows-member.md
```diff
@@ -0,0 +1,4 @@
+---
+category: minorAnalysis
+---
+* Added exception for record types auto generated Deconstruct method.
```
hvitved, line 4: The change note should mention `cs/local-shadows-member`. So something like
```
The query `cs/local-shadows-member` no longer highlights parameters of `record` types.
```

hvitved (dismissed):

hvitved:
> Will need to rebase after #7605 is merged.

Why is that?

hvitved (approved):

PR #7631, opened 2022-01-18
Python: Support SQLAlchemy `scoped_session`

RasmusWL:
I don't think this is big enough for a change-note, but let me know if you disagree.

The .py file can be run, and still doesn't raise any exceptions on those `assert`s, just FYI.


tausbn (approved):
Looks good. I don't think a change note is required.

RasmusWL:
:shipit:

PR #7635, opened 2022-01-18
Python/support match

yoff:
Cleaned up version of https://github.com/github/codeql/pull/7356

QL part of https://github.com/github/semmle-code/pull/41454

- AST nodes for patterns
  - sub expressions, sub statements, and sub patterns
  - add sub patterns to the new `case` statement
  - scopes for patterns computed as the scope of the enclosing case (could be added to `py_scopes` instead)
- SSA for variables created inside match statements
  - `PatternCaptureDefinition` and `PatternAliasDefinition`
- data flow through patterns
  - subject of match flows to each top-level pattern (could be more like use-use-flow and only flow directly to the first one)
  - shape based flow inside patterns
    - [x] MatchAsPattern
    - [x] MatchOrPattern
    - [x] MatchLiteralPattern (no flow)
    - [x] MatchCapturePattern
    - [x] MatchWildcardPattern (no flow)
    - [x] MatchValuePattern (no flow)
    - [x] MatchSequencePattern
       - [x] when tuple
       - [x] when list
    - [x] MatchStarPattern
    - [x] MatchMappingPattern
    - [x] MatchDoubleStarPattern
    - [x] MatchKeyValuePattern
    - [x] MatchClassPattern
       - [x] for keyword arguments
       - [ ] for positional arguments
    - [x] MatchKeywordPattern
- [x] create upgrade script
- [x] create downgrade script


yoff:
Superseeded by https://github.com/github/codeql/pull/7635

yoff:
The language tests pass in the Semmle-code PR.
Here, I see errors of the form
```
[VALUE_NOT_IN_TYPE] predicate py_bools(@py_bool_parent parent, int idx): Value 4333 of field parent is not in type @py_bool_parent. The value is however in the following types: @py_Try. Appears in tuple (4333,4)
```
Does that indicate a problem with the upgrade script?

yoff:
> The language tests pass in the Semmle-code PR. Here, I see errors of the form
> 
> ```
> [VALUE_NOT_IN_TYPE] predicate py_bools(@py_bool_parent parent, int idx): Value 4333 of field parent is not in type @py_bool_parent. The value is however in the following types: @py_Try. Appears in tuple (4333,4)
> ```
> 
> Does that indicate a problem with the upgrade script?

Ah, I guess that is expected since this repository contains the new dbscheme, but not the updated extractor.

python/ql/lib/semmle/python/AstExtended.qll
```diff
@@ -82,6 +82,12 @@ library class StrListParent extends StrListParent_ { }
 /** Internal implementation class */
 library class ExprParent extends ExprParent_ { }
 
+/** Internal implementation class */
+library class PatternListParent extends PatternListParent_ { }
```
tausbn, line 86: You can leave off the `library` bit. We don't need that any more.

python/ql/lib/semmle/python/Patterns.qll
```diff
@@ -0,0 +1,123 @@
...
+import python
+
+/** A pattern in a match statement */
+class Pattern extends Pattern_, AstNode {
+  /** Gets the scope of this pattern */
+  override Scope getScope() {
+    // TODO: Should it be defined as
+    //   py_scopes(this, result)
+    // instead?
+    result = this.getCase().getScope()
```
tausbn, line 14: I think reducing it to the scope of the surrounding `case` makes sense. Does using `py_scopes` actually make any observable difference?

python/ql/lib/semmle/python/Patterns.qll
```diff
@@ -0,0 +1,123 @@
...
+class MatchKeyValuePattern extends MatchKeyValuePattern_ {
+  override Pattern getASubPattern() { result = this.getKey() or result = this.getValue() }
+}
+
+/** A class pattern in a match statement: `Circle(radius = 3)` */
+class MatchClassPattern extends MatchClassPattern_ {
+  override Expr getASubExpression() { result = this.getClassName() }
+
+  override Pattern getASubPattern() {
+    result = this.getAPositional() or result = this.getAKeyword()
```
tausbn, line 114: The name `getAPositional` feels a bit awkward to me, since `Positional` reads as an adjective (whereas `Keyword` can be a noun). I wonder if there's a better name we could use... :thinking:

python/ql/lib/semmle/python/dataflow/new/internal/DataFlowPrivate.qll
```diff
@@ -1553,6 +1559,290 @@ module IterableUnpacking {
...
+      element = subject.getPattern(index)
+    |
+      nodeFrom.asCfgNode().getNode() = subject and
+      nodeTo.asCfgNode().getNode() = element and
+      (
+        // tuple content
+        c.(TupleElementContent).getIndex() = index
+        or
+        // list content
+        c instanceof ListElementContent
```
tausbn, line 1687: Do we need this always, or just when the pattern is sufficiently complicated that accurate indices are not available? 

It seems to me that for `case [a, b]:` we should know exactly which indices go where, but maybe I'm overlooking something.

python/ql/lib/semmle/python/dataflow/new/internal/DataFlowPrivate.qll
```diff
@@ -1553,6 +1559,290 @@ module IterableUnpacking {
...
+      )
+    )
+  }
+
+  /**
+   * star pattern: subject flows to the variable, possibly via a conversion
+   * syntax (toplevel): `case *var:`
+   *
+   * We decompose this flow into a read step and a store step. The read step
+   * reads both tupe and list content, the store step only stores list content.
```
tausbn, line 1727:
```suggestion
   * reads both tuple and list content, the store step only stores list content.
```

python/ql/lib/semmle/python/dataflow/new/internal/DataFlowPrivate.qll
```diff
@@ -1553,6 +1559,290 @@ module IterableUnpacking {
...
```
tausbn, line 1699:
```suggestion
   * reads both tuple and list content, the store step only stores list content.
```

python/ql/lib/semmle/python/dataflow/new/internal/DataFlowPublic.qll
```diff
@@ -476,6 +484,21 @@ class IterableElementNode extends Node, TIterableElementNode {
   override Location getLocation() { result = consumer.getLocation() }
 }
 
+/**
+ * A synthetic node representing elemnt content of a star pattern.
```
tausbn, line 488:
```suggestion
 * A synthetic node representing element content of a star pattern.
```

python/ql/test/experimental/dataflow/match/test.py
```diff
@@ -0,0 +1,152 @@
...
+
+# No flow for wildcard pattern
+
+class Unsafe:
+    VALUE = SOURCE
+
+def test_value_pattern():
+    match SOURCE:
+        case Unsafe.VALUE as x:
+            SINK(x) #$ flow="SOURCE, l:-2 -> x" MISSING: flow="SOURCE, l:-5 -> x"
```
tausbn, line 63: This missing flow is curious. Do you know what causes it?

python/ql/test/experimental/dataflow/match/test.py
```diff
@@ -0,0 +1,152 @@
...
+    match (NONSOURCE, SOURCE):
+        case (x, y):
+            SINK_F(x)
+            SINK(y) #$ flow="SOURCE, l:-3 -> y"
+
+@expects(2)
+def test_sequence_pattern_list():
+    match [NONSOURCE, SOURCE]:
+        case [x, y]:
+            SINK_F(x) #$ SPURIOUS: flow="SOURCE, l:-2 -> x"
```
tausbn, line 76: I note the spurious flow here, which I assume is caused by the list content mentioned elsewhere. I feel like we should be able to tighten this up a bit.

tausbn (changes requested):
A bunch of comments, but overall this is a wonderful piece of work!
I especially like the thorough comments and the clear division regarding setting/removing `Content` in flow steps. Made it very easy to review. :slightly_smiling_face:

python/ql/lib/semmle/python/dataflow/new/internal/DataFlowPrivate.qll
```diff
@@ -1553,6 +1559,290 @@ module IterableUnpacking {
...
+ * stemming from the bindings and the matching of shape. Below, 'subject' is not necessarily the
+ * top-level subject of the match, but rather the part recursively matched by the current pattern.
+ * For instance, in the example:
+ * ```python
+ * match command:
+ *  case ('quit' as c) | ('go', ('up'|'down') as c):
+ *   ...body
+ * ```
+ * `command` is the subject of the as-pattern, while the second component of `command` is the subject
+ * of the first capture pattern. As such, 'subject' refers to the pattern under evaluation.
```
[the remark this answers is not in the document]
yoff: Yes, that description is wrong, I think it has not been updated. Thanks for catching that.

python/ql/lib/semmle/python/Patterns.qll
yoff: I think that using `py_scopes` requires some work which is why I did it like this. The `TODO` was to remind me to ask the reviewers if that would be fine :-)

python/ql/test/experimental/dataflow/match/test.py
yoff: Yes. I thought we would need the guard aspect to be implemented, but now that you mention it, there is already flow from the value pattern `Unsafe.VALUE` to the alias `x`. So all we should need is flow from the actual value to the value pattern.

python/ql/test/experimental/dataflow/match/test.py
yoff: In this case, the incoming subject `[NONSOURCE, SOURCE]` has list content, so we really do not know where in the list the taint resides. In more realistic code, it would be written as
```
match l:
  case [x, y]:
    ... use x ...
```
If `l` has list content, we have to read it into `x`.

python/ql/lib/semmle/python/dataflow/new/internal/DataFlowPrivate.qll
yoff: `case [a, b]` is equivalent to `case (a, b)`. Just like for utterable unpacking, they are the same internally. But `match [a, b]` is different from `match (a, b)` since they receive two different values. We also model these two values differently, as in the list case we forget about indices, just assigning the whole list `ListElementContent`. So when a case sees a sequence, it could have either list content or tuple content depending on the subject coming from above. So we have to read both.

python/ql/test/experimental/dataflow/match/test.py
yoff: Actually, I think we are also missing flow from the SSA variable on line 58 to the attribute at line 62.

tausbn (approved):
Looks good! Ship it! :shipit:

tausbn (approved):

PR #7643, opened 2022-01-19
C#: Struct (and to a minor extent anonymous types) improvements 

michaelnebel:
In this PR we make

- Support for user declared default constructors for structs (test-only).
- Support for with expressions for structs, anonymous types and record structs (test only).
- Flow tests for the type of with expressions.
- Support for flow via anonymous type object initializer.


hvitved (approved):
Looks great! I have started a DCA run.

PR #7650, opened 2022-01-19
C++: Improve cpp/cleartext-transmission

geoffw0:
This PR is a collection of simple improvements to `cpp/cleartext-transmission`, each ruling out a small source of false positive results.  There may be more work to be done before upping the query to `@precision high`, but we're getting closer.


MathiasVP (dismissed):
LGTM! (except for the missing autoformat.)

geoffw0:
Autoformatted.

PR #7651, opened 2022-01-19
JS: add CWE-471 to the prototype-pollution queries

erik-krogh:
All the [CWE-471 CVEs in the advisory db](https://github.com/advisories?query=cwe%3A471) are prototype-pollution CVEs.  

So I've added CWE-471 to the prototype-pollution queries.


esbena (commented):
What about https://github.com/github/codeql/blob/cb098df4ea5afb5b1156d1b0a304d1258391433f/javascript/ql/src/Security/CWE-915/PrototypePollutingMergeCall.ql?

esbena (approved):

asgerf (approved):

PR #7652, opened 2022-01-19
Python: Remove usernames as sensitive source for cleartext queries

RasmusWL:
Draft while double checking that we wouldn't loose any important current results from this.

### DIFF
--- a/python/ql/lib/semmle/python/security/dataflow/CleartextLoggingCustomizations.qll
+++ b/python/ql/lib/semmle/python/security/dataflow/CleartextLoggingCustomizations.qll
@@ -40,6 +40,10 @@ module CleartextLogging {
    * A source of sensitive data, considered as a flow source.
    */
   class SensitiveDataSourceAsSource extends Source, SensitiveDataSource {
+    SensitiveDataSourceAsSource() {
+      not SensitiveDataSource.super.getClassification() = SensitiveDataClassification::id()
+    }
+
     override SensitiveDataClassification getClassification() {
       result = SensitiveDataSource.super.getClassification()
     }
--- a/python/ql/src/change-notes/2021-01-19-remove-cleartext-fps.md
+++ b/python/ql/src/change-notes/2021-01-19-remove-cleartext-fps.md
@@ -1,0 +1,4 @@
+---
+category: majorAnalysis
+---
+* No longer consider usernames or other account information as sensitive data for the queries `py/clear-text-logging-sensitive-data` and `py/clear-text-storage-sensitive-data`, since this has lead to many false positives.
--- a/python/ql/src/change-notes/2021-01-19-remove-cleartext-fps.md
+++ b/python/ql/src/change-notes/2021-01-19-remove-cleartext-fps.md
@@ -1,4 +1,4 @@
 ---
 category: majorAnalysis
 ---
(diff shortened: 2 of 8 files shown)


RasmusWL:
I looked through 100 projects that had results with id/username, and nothing was interesting. So I think we should do this :+1:

python/ql/lib/semmle/python/security/dataflow/CleartextLoggingCustomizations.qll
```diff
@@ -40,6 +40,10 @@ module CleartextLogging {
    * A source of sensitive data, considered as a flow source.
    */
   class SensitiveDataSourceAsSource extends Source, SensitiveDataSource {
+    SensitiveDataSourceAsSource() {
+      not SensitiveDataSource.super.getClassification() = SensitiveDataClassification::id()
```
tausbn, line 44: Just to double-check, do we want `not ... = ...` here, and not `... != ...`? If I'm not mistaken, the latter would allow something that's both `id` and some other classification, whereas the former would disallow anything that can be seen as `id` (regardless of any other classifications).

python/ql/src/change-notes/2021-01-19-remove-cleartext-fps.md
```diff
@@ -0,0 +1,4 @@
+---
+category: majorAnalysis
+---
+* No longer consider usernames or other account information as sensitive data for the queries `py/clear-text-logging-sensitive-data` and `py/clear-text-storage-sensitive-data`, since this has lead to many false positives.
```
tausbn, line 4:
```suggestion
* User names and other account information is no longer considered to be sensitive data for the queries `py/clear-text-logging-sensitive-data` and `py/clear-text-storage-sensitive-data`, since this lead to many false positives.
```
An attempt at a slight reformulation.

python/ql/test/query-tests/Security/CWE-312-CleartextLogging/test.py
```diff
@@ -39,6 +39,36 @@ def print_password():
...
+    # https://github.com/github/codeql/issues/6726
+    # https://github.com/github/codeql/issues/7497
+    x = generate_uuid4()
+    print(x) # OK
+
+    # username not considered sensitive
+    # https://github.com/github/codeql/issues/7116
+    logging.error("Misc Exception. User %s: %s", request.user.username)
+
+    # dictionary taint-flow corss-talk
```
tausbn, line 62:
```suggestion
    # dictionary taint-flow cross-talk
```

tausbn (changes requested):
A few questions and suggestions, but otherwise I think it looks good! :+1:

python/ql/lib/semmle/python/security/dataflow/CleartextLoggingCustomizations.qll
RasmusWL: Not something I had thought about. I agree with your analysis on what the code would do.

Given the problem about cross-talk in dictionaries, I guess we're better off with disallowing anything that can be seen as an `id` :thinking:

tausbn: Makes sense. :+1:

tausbn (approved):
:shipit: :rocket:

yoff (approved):
LGTM
I would hope that we can also solve the cross-talk properly in the future.

PR #7654, opened 2022-01-19
Python: Cleanup: Remove old points-to versions of queries

RasmusWL:
Since we've internally agreed that we've reached the same or better set of results.


yoff (commented):
I wondered if there was some flow-label-modelling that we wanted to port, now that we have flow state in the new data flow library, but that is easily recovered from history.

What about the libraries, say `semmle.python.security.Paths`, are they still needed?

RasmusWL:
I wanted to take things one step at a time, so this is only about removing the old points-to versions of the queries. For removing the old modeling, we need to deprecate it first, which I'm going to do in an other PR.

> I wondered if there was some flow-label-modelling that we wanted to port, now that we have flow state in the new data flow library, but that is easily recovered from history.

I looked through these queries, and they don't contain any interesting flow-state modeling. and as you point out, it's possible to check out the old versions if we do need to look at it in the future. But I would think that much of the old flow-state modeling would be handled by flow-summaries in the future instead.

yoff (approved):
Out with the old, in with the new...

PR #7656, opened 2022-01-19
Java: Exclude irrelevant rows from models

bmuskalla:
* Exclude `main` methods
* Exclude any `logging` sinks as are not actionable for the users of specific libraries


atorralba (approved):
LGTM

PR #7658, opened 2022-01-19
C#: Get rid of negative parameter/argument data-flow positions

hvitved:
This PR makes use of https://github.com/github/codeql/pull/7260 to get rid of negative parameter/argument positions. Negative indices were used in two places:

- `-1`: representing qualifiers/`this` parameters.
- `-2 ..`: used to model flow through captured variables.

The former are used in flow summaries, which means that `Argument[Qualifier]` replaces `Argument[-1]`. The latter are only used internally, but getting rid of those as well means that we can remove some hacky `equivalenceRelation`-based ranking.


michaelnebel (approved):
Looks good!

csharp/ql/lib/semmle/code/csharp/dataflow/internal/DataFlowDispatch.qll
```diff
@@ -116,19 +116,17 @@ private module Cached {
...
+    TPositionalParameterPosition(int i) { i = any(Parameter p).getPosition() } or
+    TThisParameterPosition() or
+    TImplicitCapturedParameterPosition(SsaCapturedEntryDefinition def)
 
   cached
-  newtype TArgumentPosition = MkArgumentPosition(int i) { i = parameterPosition() }
+  newtype TArgumentPosition =
+    TPositionalArgumentPosition(int i) { i = any(Parameter p).getPosition() } or
+    TQualifierArgumentPosition() or
+    TImplicitCapturedArgumentPosition(SsaCapturedEntryDefinition def)
```
aschackmull, line 129: This looks risky! This might introduce virtual-dispatch-sized blowups in any part of the dataflow lib that work with an argument without having joined that to a call target.

aschackmull: Or is this a case where the call-callee relation is 1-1?

hvitved: Performance-wise this should be no different from before, where we translated each `SsaCapturedEntryDefinition` into a magic number (using `equivalenceRelation`), and used that magic number as both parameter and argument positions.

hvitved: But now that I think about it, this can probably be made a bit nicer (and smaller) by simply using the variable that is captured as the positions.

aschackmull: I guess the bad case requires some sort of convoluted setup where a single call might dispatch to many different lambdas (or other kinds of closures) that each capture some variable, and that this single call somehow provides a value in the corresponding argument-position(s). I don't know if such a construction is even possible - it depends on how these arguments are recognised.

michaelnebel (approved):
LGTM!

PR #7659, opened 2022-01-19
Python: Move regex injection configuration files

RasmusWL:
I did not notice that these went to the wrong location in #6693. They should be in the `dataflow` folder with the rest of the data-flow configurations files, the `injection` folder is for old points-to based modeling.


yoff (approved):
Thanks for cleaning this up

PR #7660, opened 2022-01-19
Python: Deprecate old points-to based modeling

RasmusWL:
I did this with search/replace regex. Draft PR to see what sort of failures come up.


RasmusWL:
Very excited to announce that all checks now pass, so this can be reviewed :tada:

tausbn (approved):
Looks good to me! Thanks for doing this. 

Regarding the tests, instead of adding the deprecation warnings to the expected output, I wonder if we should just delete them outright.

RasmusWL:
> Regarding the tests, instead of adding the deprecation warnings to the expected output, I wonder if we should just delete them outright.

I think some of the tests could still have value if we want to port the (old) experimental queries away from points-to.

PR #7661, opened 2022-01-19
Docs: Mention packaging commands in CodeQL extension

shati-patel:
We've introduced two new commands in https://github.com/github/vscode-codeql/pull/1076 which will make it easier to work with CodeQL packs from within VS Code. You can download packs and install pack dependencies. I think this is a suitable place to document the new commands, but other suggestions are welcome! 📝 

The linked PR also has some demo GIFs for reference 😊 

No real rush to get this merged, since the commands should be fairly self-explanatory. Just whenever the next CLI release happens!

### DIFF
--- a/docs/codeql/codeql-for-visual-studio-code/working-with-codeql-packs-in-visual-studio-code.rst
+++ b/docs/codeql/codeql-for-visual-studio-code/working-with-codeql-packs-in-visual-studio-code.rst
@@ -23,8 +23,14 @@ You can then use the CodeQL CLI to publish your pack to share with others. For m
 
 Viewing CodeQL packs and their dependencies in Visual Studio Code
 -----------------------------------------------------------------
-Whether you have used the CodeQL CLI to download a CodeQL pack that someone else has created, or created your own, you can open the ``qlpack.yml`` file in the root of a CodeQL pack directory in Visual Studio Code and view the dependencies section to see what libraries the pack depends on.
+To download a CodeQL pack that someone else has created, run the **CodeQL: Download Packs** command from the Command Palette.
+
+Whether you have downloaded a CodeQL pack or created your own, you can open the ``qlpack.yml`` file in the root of a CodeQL pack directory in Visual Studio Code and view the dependencies section to see what libraries the pack depends on.
 
 If you want to understand a query in a CodeQL pack better, you can open the query file and view the code, using the IntelliSense code editing features of Visual Studio Code. For example, if you hover over an element from a library depended on by the pack, Visual Studio Code will resolve it so you can see documentation about the element. 
 
 To view the full definition of an element of a query, you can right-click and choose **Go to Definition**. If the library pack is present within the same Visual Studio Code workspace, this will take you to the definition within the workspace. Otherwise it will take you to the definition within your package cache, the shared location where downloaded dependencies are stored, which is in your home directory by default.
+
+Installing dependencies for CodeQL packs in Visual Studio Code
+--------------------------------------------------------------
+To  install dependencies for a CodeQL pack in your Visual Studio Code workspace, run the **CodeQL: Install Pack Dependencies** command from the Command Palette and select the packs you want to install dependencies for.


docs/codeql/codeql-for-visual-studio-code/working-with-codeql-packs-in-visual-studio-code.rst
```diff
@@ -23,8 +23,14 @@ You can then use the CodeQL CLI to publish your pack to share with others. For m
...
 
-To view the full definition of an element of a query, you can right-click and choose **Go to Definition**. If the library pack is present within the same Visual Studio Code workspace, this will take you to the definition within the workspace. Otherwise it will take you to the definition within your package cache, the shared location where downloaded dependencies are stored, which is in your home directory by default.
\ No newline at end of file
+To view the full definition of an element of a query, you can right-click and choose **Go to Definition**. If the library pack is present within the same Visual Studio Code workspace, this will take you to the definition within the workspace. Otherwise it will take you to the definition within your package cache, the shared location where downloaded dependencies are stored, which is in your home directory by default.
+
+Installing dependencies for CodeQL packs in Visual Studio Code
```
adityasharad, line 34: I suggest putting this first; it's more important at the moment.

docs/codeql/codeql-for-visual-studio-code/working-with-codeql-packs-in-visual-studio-code.rst
```diff
@@ -23,8 +23,14 @@ You can then use the CodeQL CLI to publish your pack to share with others. For m
...
+--------------------------------------------------------------
+To  install dependencies for a CodeQL pack in your Visual Studio Code workspace, run the **CodeQL: Install Pack Dependencies** command from the Command Palette and select the packs you want to install dependencies for.
```
adityasharad, line 36: In particular, this allows the user to write and run query packs that depend on the CodeQL standard libraries, without needing the standard libraries checked out as part of the workspace. I think we should briefly mention that example.

adityasharad (dismissed):
Good stuff. Couple of minor suggestions.

docs/codeql/codeql-for-visual-studio-code/working-with-codeql-packs-in-visual-studio-code.rst
shati-patel: Thanks, that's a useful example! Updated ⚡

docs/codeql/codeql-for-visual-studio-code/working-with-codeql-packs-in-visual-studio-code.rst
```diff
@@ -23,8 +29,10 @@ You can then use the CodeQL CLI to publish your pack to share with others. For m
 
 Viewing CodeQL packs and their dependencies in Visual Studio Code
 -----------------------------------------------------------------
-Whether you have used the CodeQL CLI to download a CodeQL pack that someone else has created, or created your own, you can open the ``qlpack.yml`` file in the root of a CodeQL pack directory in Visual Studio Code and view the dependencies section to see what libraries the pack depends on.
+To download a CodeQL pack that someone else has created, run the **CodeQL: Download Packs** command from the Command Palette.
```
felicitymay, line 32: Having watched the gif (thanks), I wonder if we need to mention that you'll need to enter the full pack name and give an example. This may be obvious to the user but I wanted to check.

docs/codeql/codeql-for-visual-studio-code/working-with-codeql-packs-in-visual-studio-code.rst
```diff
@@ -13,6 +13,12 @@ About CodeQL packs
 ------------------
 CodeQL packs are used to create, share, depend on, and run CodeQL queries and libraries. You can publish your own CodeQL packs and download packs created by others. For more information, see ":ref:`About CodeQL packs <about-codeql-packs>`."
 
+Installing dependencies for CodeQL packs in Visual Studio Code
+--------------------------------------------------------------
+To install dependencies for a CodeQL pack in your Visual Studio Code workspace, run the **CodeQL: Install Pack Dependencies** command from the Command Palette and select the packs you want to install dependencies for.
```
felicitymay, line 18: In the demo gif, this is called **CodeQL: Install Packs** - I assume that the name changed since you recorded that?

docs/codeql/codeql-for-visual-studio-code/working-with-codeql-packs-in-visual-studio-code.rst
```diff
@@ -13,6 +13,12 @@ About CodeQL packs
 ------------------
 CodeQL packs are used to create, share, depend on, and run CodeQL queries and libraries. You can publish your own CodeQL packs and download packs created by others. For more information, see ":ref:`About CodeQL packs <about-codeql-packs>`."
 
+Installing dependencies for CodeQL packs in Visual Studio Code
```
felicitymay, line 16: I wonder whether this should be more task focused, maybe:

```suggestion
Using standard CodeQL packs in Visual Studio Code
```

docs/codeql/codeql-for-visual-studio-code/working-with-codeql-packs-in-visual-studio-code.rst
```diff
@@ -13,6 +13,12 @@ About CodeQL packs
...
+--------------------------------------------------------------
+To install dependencies for a CodeQL pack in your Visual Studio Code workspace, run the **CodeQL: Install Pack Dependencies** command from the Command Palette and select the packs you want to install dependencies for.
+
+In particular, installing pack dependencies lets you write and run query packs that depend on the CodeQL standard libraries, without needing to check out the standard libraries in your workspace.
```
felicitymay, line 20: I'd be inclined to put the explanation ahead of the procedure. If you agree, then the text would need a tweak - maybe:

```suggestion
You can write and run query packs that depend on the CodeQL standard libraries, without needing to check out the standard libraries in your workspace. Instead, you can install the only the dependencies required by the query packs you want to use.
```

docs/codeql/codeql-for-visual-studio-code/working-with-codeql-packs-in-visual-studio-code.rst
felicitymay: Also - potentially, this deserves its own heading, even though it will be a very short section, to make it easier for users to find when they scan this article.

felicitymay (approved):
@shati-patel - thank you so much for updating the documentation for these new commands. 💖 

I've made various suggestions for possible improvements, but these are based just on viewing your gifs as I haven't been able to use the commands. Feel free to ignore any that don't make sense.

I think this change should be merged into the CLI 2.7.6 branch, assuming that we'll want to update the docs to include this change when we release 2.7.6? Otherwise it won't be published until a new CLI branch is taken from `main` and that CLI version released.

docs/codeql/codeql-for-visual-studio-code/working-with-codeql-packs-in-visual-studio-code.rst
shati-patel: Yes. sorry about that - the GIF is slightly outdated! We decided to change the command name to be more explicit 😺

docs/codeql/codeql-for-visual-studio-code/working-with-codeql-packs-in-visual-studio-code.rst
shati-patel: Thanks! I've added a bit of clarification and an example in https://github.com/github/codeql/pull/7661/commits/1462565810a1d63d84b2d6fec1aea443fab138a8 ⚡ 

Haven't added a new heading just yet, since we don't expect too many users to use this command at the moment. Once we've added a way to directly run queries from a downloaded pack, we can make this more discoverable! (Currently, they'd have to find the queries in their package cache, which isn't ideal...)

shati-patel:
> @shati-patel - thank you so much for updating the documentation for these new commands. 💖
> 
> I've made various suggestions for possible improvements, but these are based just on viewing your gifs as I haven't been able to use the commands. Feel free to ignore any that don't make sense.
> 
> I think this change should be merged into the CLI 2.7.6 branch, assuming that we'll want to update the docs to include this change when we release 2.7.6? Otherwise it won't be published until a new CLI branch is taken from `main` and that CLI version released.

Thanks for the review! ✨ 🎉 

This feature hasn't been released yet in the extension so merging into `main` is okay, and the docs changes can be published during the next CLI release cycle 😊

felicitymay (approved):
Thanks for the clarifications and assurance about the target branch. This LGTM 🚀

PR #5417, opened 2021-03-16
Inline testing: Allow multiple actual values in InlineExpectationsTest

MathiasVP:
In the C++ dataflow tests we annotate sinks with their source's location (as a value for a test tag). To save a few keystrokes we only require the source location when there's more than 1 source.

However, it recently came up in a [PR discussion](https://github.com/github/codeql/pull/3364#discussion_r594941441) that it's sometimes nice to optionally be able to specify the source location even though there's just one source.

This PR changes the InlineExpectationsTest framework so that a `TActualResult` can have a set of values for a tag, and a test annotation is satisfied if the value of that tag is in the set of actual values.

PR #7663, opened 2022-01-19
Ruby: Add basic subclassing support to API Graphs

hmac:
Given the code

```ruby
class A; end
class B < A; end
class C < A; end
```

You can find uses of B and C with the expression

```ql
API::getTopLevelMember("A").getASubclass()
```

We do this by adding edges from the use of `A` in `class B < A` to all uses of `B`.

API Graph paths that use `getASubclass()` will, I think, always be longer than an equivalent "canonical" path which doesn't use `getASubclass()`. For example, a use of the class `B` above is accessible via `API::getTopLevelMember("B")` and `API::getTopLevelMember("A").getASubclass()`. Therefore when it comes to testing `getASubclass()`, we're interested in non-canonical paths, for example:

```
class A; end
class B < A; end

B # API::getTopLevelMember("B") [canonical], API::getTopLevelMember("A").getASubclass()
```

To test subclass support using inline tests, I've extended the `InlineTest` framework to support _optional_ results, which are matched against annotations but do not trigger a failure if there is no matching annotation. This allows us to add annotations for non-canonical paths where we want to test subclassing, but leave existing tests alone.


ruby/ql/lib/codeql/ruby/frameworks/ActionController.qll
```diff
@@ -35,16 +22,13 @@ private class ApplicationControllerAccess extends ConstantReadAccess {
...
-    exists(ActionControllerControllerClass other |
-      other.getModule() = resolveConstantReadAccess(this.getSuperclassExpr())
-    )
+    this.getSuperclassExpr() =
+      [
+        API::getTopLevelMember("ActionController").getMember("Base"),
+        // In Rails applications `ApplicationController` typically extends `ActionController::Base`, but we
+        // treat it separately in case the `ApplicationController` definition is not in the database.
+        API::getTopLevelMember("ApplicationController")
+      ].getASubclass*().getAUse().asExpr().getExpr()
```
aibaars, line 31: I think it would be nice to have a method named `getADirectSubclass` and define `getASubclass` to be `getADirectSubclass*`. I expect users in almost all cases need the transitive closure and they are likely to forget adding `*`. Having the nicely named method "just do the right thing in most cases" would be helpful.

ruby/ql/lib/codeql/ruby/ApiGraphs.qll
```diff
@@ -174,8 +174,7 @@ module API {
           // avoid producing strings longer than 1MB
           result.length() < 1000 * 1000
         )
-      ) and
-      length in [1 .. Impl::distanceFromRoot(this)]
```
hvitved, line 178: I don't think we should change this just so we can use it in a test. I think it would be OK to replicate the above in the test itself instead.

ruby/ql/lib/codeql/ruby/ApiGraphs.qll
```diff
@@ -389,6 +387,17 @@ module API {
...
+      or
+      // `pred` is a use of class A
+      // `succ` is a use of class B
+      // there exists a class declaration B < A
+      exists(ClassDeclaration c, DataFlow::Node a, DataFlow::Node b |
+        use(pred, a) and
+        use(succ, b) and
+        resolveConstant(b.asExpr().getExpr()) = resolveConstantWriteAccess(c) and
+        c.getSuperclassExpr() = a.asExpr().getExpr() and
+        lbl = Label::subclass()
```
hvitved, line 399: I wonder if this should include the name of the sub class.

hvitved (commented):
I don't have much experience with sub classes in API graphs, but @max-schaefer mentioned that it might be a good principle for API graphs to adhere to the substitution principle. The way I understand this, it would mean that when asking, say, for `getMember(C)`, we should already then be getting `C` and all sub classes thereof. However, whether we would also like to be able to get just `C`, I don't know.

tausbn:
> I remember the @github/codeql-python folks remarking how their API-graph code ended up using `getASubclass*()` all over the place, which could perhaps be avoided by this scheme.

Indeed. When I originally implemented support for subclassing, you mentioned this issue and I discussed it with the other members of the Python team (as I was eager to avoid peppering our code with `getASubclass*()`).

If memory serves, one of the concerns raised was that if `some_class_api_node` implicitly represents not just itself but all of its subclasses, then we lose the ability to distinguish between these cases (which might be relevant at some point). So instead we opted for the more explicit (and in turn perhaps more Pythonic) `getASubclass`, to make it explicitly clear that we're now including all subclasses. In practice, this isn't terribly much of a nuisance. (Also, checking our libraries just now, _most_ instances use `getASubclass*`, but there is one instance where we do `getASubclass+` instead.)

max-schaefer:
> If memory serves, one of the concerns raised was that if `some_class_api_node` implicitly represents not just itself but all of its subclasses, then we lose the ability to distinguish between these cases (which might be relevant at some point)

Absolutely! That's why my proposal was to only bring in the substitution principle when referring to instances or members (not when referring to the class itself), but there are still things you can't express that way, and it's entirely possible that they are practically relevant.

max-schaefer:
I like that suggestion! It's similar to how we have `getAUse` and `getAnImmediateUse` that do/don't follow interprocedural flow: you usually want the former but the latter is available for the occasional case where you don't.

ruby/ql/lib/codeql/ruby/ApiGraphs.qll
hmac: I removed this because I noticed that we were only calling this predicate in
```ql
string getPath() {
  result = min(string p | p = this.getAPath(Impl::distanceFromRoot(this)) | p)
}
```
which already passes in a max length of `Impl::distanceFromRoot(this)`, so I thought it was redundant. Am I mistaken?

tausbn: >  [...] which already passes in a max length of `Impl::distanceFromRoot(this)`, so I thought it was redundant. Am I mistaken?

Careful! This is "top-down" reasoning ("passes in"), but the evaluation is actually bottom-up. 

Consider how `getAPath` is evaluated in isolation. The lines you deleted would prevent this predicate from containing a `result` and `length` where `length` is greater than the shortest path to the given node. Without those lines, there's no such restriction (except the fact that `result` can't be more than 1000000 characters, which is likely something that happens much further out). In particular, if you have a loop in the API graph (which is almost certain), this predicate will keep going round and round in that loop, producing path strings of greater and greater length, until finally hitting the 1M limit.

In general, it's useful to consider the question "which tuples will be in this relation" _regardless_ of any context that may limit that set. If we're lucky, that context will be automatically magicked in, but it might not (and so it's better to just limit the number of tuples in advance).

hmac: Thanks for explaining! I'm definitely still getting used to thinking in the bottom-up evaluation mindset :)

hmac:
Thanks everyone for your comments and the insightful discussion! I've pushed a change to the exposed API such that:
- `getAnImmediateSubclass()` returns direct subclasses of the receiver
- `getASubclass()` returns the transitive closure of the above

As a result, to get a class `A` and all its subclasses you still have to explicitly call `getASubclass()`.

If you want to match calls to `Foo::Bar.baz` whilst also including any subclasses of `Foo` _and_ subclasses of `Foo::Bar`, you will have to use
```ql
API::getTopLevelMember("Foo").getASubclass().getMember("Bar").getASubclass().getMethodCall("baz")
```
However I think this is a rare case. It is more common for `Foo` to be a module, in which case you can't create subclasses of it.

I think this a reasonable middle ground, and I think it would be nice to merge this PR and try using it in our queries etc, and then make further changes/improvements as we encounter the need. How does that sound?

ruby/ql/lib/codeql/ruby/ApiGraphs.qll
hmac: That sounds reasonable. I can't immediately see what benefit it would bring though. Does it make anything particular easier/harder?

ruby/ql/lib/codeql/ruby/ApiGraphs.qll
hmac: I've reverted this change in a followup commit - the test now uses a copy of `getAPath` without the length restriction.

MathiasVP:
> To test subclass support using inline tests, I've extended the `InlineTest` framework to support _optional_ results, which are matched against annotations but do not trigger a failure if there is no matching annotation. This allows us to add annotations for non-canonical paths where we want to test subclassing, but leave existing tests alone.

I've wanted this feature myself for a while ❤️. In fact, I did it the wrong way in https://github.com/github/codeql/pull/5417 and closed it again following the discussion with `@aschackmull`. But I like the API you went with here a lot more 👍.

MathiasVP (dismissed):

hmac:
(Sorry about the mass ping, language teams! Changes to the shared inline test framework caused GitHub to request reviews from everyone.)

erik-krogh:
I'm currently implementing def nodes for the API graph implementation in Python, and I also encountered similar testing issues.   
I ended up porting the [API-graph testing framework from JS](https://github.com/github/codeql/blob/main/javascript/ql/test/ApiGraphs/VerifyAssertions.qll), but using the same syntax as the inline expectation tests. 

You can see [my implementation here](https://github.com/github/codeql/pull/7806/commits/e3a3c66c931d101ceb7d610eacd63818eb51e1ab).  
A clear advantage is that the error messages you get are usually very good, and there is no potential performance issue from computing a basically unbounded number of paths.  

E.g. I think your test will blow up if you try a test [like this one](https://github.com/github/codeql/pull/7806/commits/ca370c2f37dce8f719dc8618eb88ab0dfcaf98b7).  
(Some quick math suggests that there is about `2^(1000000/17)≈10^17707` paths of length of length less than 1 million to the API-nodes in that function).

ruby/ql/lib/codeql/ruby/ApiGraphs.qll
```diff
@@ -101,9 +101,26 @@ module API {
...
+     * Gets a node representing a (direct or indirect) subclass of the class represented by this node.
+     * ```rb
+     * class A; end
+     * class B < A; end
+     * class C < B; end
+     * ```
+     * In the example above, `getMember("A").getASubclass()` will return uses of `A`, `B` and `C`.
      */
-    Node getASubclass() { result = this.getASuccessor(Label::subclass()) }
+    Node getASubclass() { result = this.getAnImmediateSubclass*() }
```
hvitved, line 112: Should we update
```ql
Node getInstance() { result = this.getASuccessor(Label::instance()) }
```

to

```ql
Node getInstance() { result = this.getASubclass().getASuccessor(Label::instance()) }
```

@asgerf , @aibaars I think this is related to our discussion yesterday

asgerf: I agree that's probably better (though I still don't have a good intuition about what it's like to model Ruby frameworks).

hmac: I think that makes sense. If we do so, we should probably also do the same for `getReturn`. Together that should cover both instance and class methods.

hmac: Added this in https://github.com/github/codeql/pull/7663/commits/704b58519fb1346049ab015c15f6b0618a9e29db

ruby/ql/lib/codeql/ruby/ApiGraphs.qll
hvitved: Let's just skip it for now.

hvitved (approved):

hmac:
> I'm currently implementing def nodes for the API graph implementation in Python, and I also encountered similar testing issues. I ended up porting the [API-graph testing framework from JS](https://github.com/github/codeql/blob/main/javascript/ql/test/ApiGraphs/VerifyAssertions.qll), but using the same syntax as the inline expectation tests.
> 
> You can see [my implementation here](https://github.com/github/codeql/pull/7806/commits/e3a3c66c931d101ceb7d610eacd63818eb51e1ab). A clear advantage is that the error messages you get are usually very good, and there is no potential performance issue from computing a basically unbounded number of paths.
> 
> E.g. I think your test will blow up if you try a test [like this one](https://github.com/github/codeql/pull/7806/commits/ca370c2f37dce8f719dc8618eb88ab0dfcaf98b7). (Some quick math suggests that there are at least `2^(1000000/17)≈10^17707` paths of length less than 1 million to the API-nodes in that function).

Thanks @erik-krogh! I think for now I'm going to stick with the simple approach in this PR but if we encounter performance problems in future then it's good to know there's a better implementation we can switch to 👍

PR #7668, opened 2022-01-20
simplify expressions that could be type-casts

erik-krogh:
Fixes the issues identified by [this QL-for-QL query](https://github.com/github/codeql/pull/7472). 

--- 

Consider the below: 
```CodeQL
predicate even(int i) { exists(Even even | even = i) }
```
That is basically just a type-cast, so we might as well simplify it as such:  
```CodeQL
predicate even(int i) { i instanceof Even }
```

--- 

There are also some cases where the resulting type-cast would be redundant. 
Here's an example: 
```CodeQL
predicate hasAttribute(string name) { exists(XMLAttribute a | a = this.getAttribute(name)) }
```
The `getAttribute` predicate has `XMLAttribute` as its result type, so adding a type-cast would be redundant. 
It is therefore simplified as follows:  
```CodeQL
predicate hasAttribute(string name) { exists(this.getAttribute(name)) }
```

--- 

The commit was generated automatically by a tool. 
(GitHub people: you can see how in the backref below).

### DIFF
--- a/ruby/ql/lib/codeql/ruby/frameworks/http_clients/Httparty.qll
+++ b/ruby/ql/lib/codeql/ruby/frameworks/http_clients/Httparty.qll
@@ -39,7 +39,7 @@ class HttpartyRequest extends HTTP::Client::Request::Range {
     exists(DataFlow::Node r | r = requestNode.getAMethodCall("body") | result = r)
     or
     // Otherwise, treat the response as the response body.
-    not exists(DataFlow::Node r | r = requestNode.getAMethodCall("body")) and
+    not exists(requestNode.getAMethodCall("body")) and
     result = requestUse
   }
 
--- a/cpp/ql/lib/semmle/code/cpp/Class.qll
+++ b/cpp/ql/lib/semmle/code/cpp/Class.qll
@@ -206,9 +206,7 @@ class Class extends UserType {
    * it is callable by a particular caller. For C++11, there's also a question
    * of whether to include members that are defaulted or deleted.
    */
-  deprecated predicate hasCopyConstructor() {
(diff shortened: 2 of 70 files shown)


nickrolfe (approved):
👍 for Ruby

smowton (approved):
Java :+1:

tausbn (approved):
Python :+1:

aschackmull (approved):

esbena (approved):

ruby/ql/lib/codeql/ruby/frameworks/http_clients/Httparty.qll
```diff
@@ -39,7 +39,7 @@ class HttpartyRequest extends HTTP::Client::Request::Range {
     exists(DataFlow::Node r | r = requestNode.getAMethodCall("body") | result = r)
     or
     // Otherwise, treat the response as the response body.
-    not exists(DataFlow::Node r | r = requestNode.getAMethodCall("body")) and
```
yoff, line 42: I wonder if the line
```
exists(DataFlow::Node r | r = requestNode.getAMethodCall("body") | result = r)
```
could also be simplified to
```
result = requestNode.getAMethodCall("body")
```
if the query was extended slightly? (We know that `result` has type `DataFlow::Node` in this case.)

erik-krogh: It could, and originally my query did just that (see [the discussion in the PR](https://github.com/github/codeql/pull/7472)).  
But that had too many FPs, so I dropped it.

MathiasVP (approved):

hvitved (approved):
C# 👍

PR #7674, opened 2022-01-20
QL: Use of db-type outside language core.

erik-krogh:
I fixed the 1 alert for JS by moving the file into the `lib/` folder.


aschackmull:
Sounds reasonable.

PR #7677, opened 2022-01-20
Ruby: Replace `getValueText` with `getConstantValue`

hvitved:
This PR replaces `string getValueText()` with a typed `ConstantValue getConstantValue()` predicate. `ConstantValue` is a union type consisting of integers, floats, rationals, complex numbers, strings, Booleans, and `nil`.

Existing `getValueText` overloads have been replaced with appropriate `getConstantValue` overloads (e.g., `IntegerLiteral` now returns a `ConstantValue::ConstantIntegerValue`, as opposed to merely a string). This means that we can implement constant folding for binary operations appropriately, and I have also implemented constant folding for unary operations (`+x` and `-x`). As discussed on previous occasions, I have also removed the logic for truncating string concatenations `x + y`, and now such concatenations will simply have no value if the combined length exceeds 10,000.

### DIFF
--- a/ruby/ql/lib/codeql/ruby/ast/Literal.qll
+++ b/ruby/ql/lib/codeql/ruby/ast/Literal.qll
@@ -761,8 +761,8 @@ class SymbolLiteral extends StringlikeLiteral, TSymbolLiteral {
 // Tree-sitter gives us value text including the colon, which we skip.
 private string getSimpleSymbolValue(Ruby::SimpleSymbol ss) { result = ss.getValue().suffix(1) }
 
-private class RequiredStringConstantValue7 extends RequiredConstantValue {
-  override predicate requiredString(string s) { s = getSimpleSymbolValue(_) }
+private class RequiredSymbolConstantValue extends RequiredConstantValue {
+  override predicate requiredSymbol(string s) { s = getSimpleSymbolValue(_) }
 }
 
 private class SimpleSymbolLiteral extends SymbolLiteral, TSimpleSymbolLiteral {
@@ -770,8 +770,8 @@ private class SimpleSymbolLiteral extends SymbolLiteral, TSimpleSymbolLiteral {
 
   SimpleSymbolLiteral() { this = TSimpleSymbolLiteral(g) }
 
-  final override ConstantValue::ConstantStringValue getConstantValue() {
-    result.isString(getSimpleSymbolValue(g))
+  final override ConstantValue::ConstantSymbolValue getConstantValue() {
+    result.isSymbol(getSimpleSymbolValue(g))
   }
 
   final override string toString() { result = g.getValue() }
@@ -795,8 +795,8 @@ private class BareSymbolLiteral extends ComplexSymbolLiteral, TBareSymbolLiteral
   final override StringComponent getComponent(int i) { toGenerated(result) = g.getChild(i) }
 }
 
-private class RequiredStringConstantValue8 extends RequiredConstantValue {
-  override predicate requiredString(string s) { s = any(Ruby::HashKeySymbol h).getValue() }
+private class RequiredSymbolConstantValue2 extends RequiredConstantValue {
+  override predicate requiredSymbol(string s) { s = any(Ruby::HashKeySymbol h).getValue() }
 }
 
 private class HashKeySymbolLiteral extends SymbolLiteral, THashKeySymbolLiteral {
@@ -804,8 +804,8 @@ private class HashKeySymbolLiteral extends SymbolLiteral, THashKeySymbolLiteral 
 
   HashKeySymbolLiteral() { this = THashKeySymbolLiteral(g) }
 
-  final override ConstantValue::ConstantStringValue getConstantValue() {
-    result.isString(g.getValue())
+  final override ConstantValue::ConstantSymbolValue getConstantValue() {
+    result.isSymbol(g.getValue())
   }
 
   final override string toString() { result = ":" + g.getValue() }
@@ -829,7 +829,7 @@ class SubshellLiteral extends StringlikeLiteral, TSubshellLiteral {
   final override StringComponent getComponent(int i) { toGenerated(result) = g.getChild(i) }
 }
 
-private class RequiredStringConstantValue9 extends RequiredConstantValue {
+private class RequiredStringConstantValue7 extends RequiredConstantValue {
   override predicate requiredString(string s) { s = any(Ruby::Character c).getValue() }
 }
 
@@ -1131,7 +1131,7 @@ private string getMethodName(MethodName::Token t) {
   result = t.(Ruby::Setter).getName().getValue() + "="
 }
 
-private class RequiredStringConstantValue10 extends RequiredConstantValue {
+private class RequiredStringConstantValue8 extends RequiredConstantValue {
   override predicate requiredString(string s) { s = getMethodName(_) }
 }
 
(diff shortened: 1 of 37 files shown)


ruby/ql/lib/codeql/ruby/ast/Literal.qll
```diff
@@ -129,13 +167,32 @@ class RationalLiteral extends NumericLiteral, TRationalLiteral {
...
+    result = s.prefix(s.length() - 1).toFloat()
+  )
+}
+
+private class RequiredComplexConstantValue extends RequiredConstantValue {
+  override predicate requiredComplex(float real, float imaginary) {
+    real = 0 and
+    imaginary = getComplexValue(_)
+  }
+}
```
hmac, line 194: This doesn't appear to support complex numbers like `1+2i` or `3-4i`. Is it worth adding that?

ruby/ql/lib/codeql/ruby/ast/Literal.qll
```diff
@@ -211,11 +272,15 @@ private class FalseLiteral extends BooleanLiteral, TFalseLiteral {
 
   FalseLiteral() { this = TFalseLiteral(g) }
 
-  final override string getValueText() { result = g.getValue() }
+  final override string toString() { result = g.getValue() }
 
   final override predicate isFalse() { any() }
 }
 
+private class RequiredStringConstantValue extends RequiredConstantValue {
```
hmac, line 280: For these and the `RequiredStringConstantValueN` variations below, it would be nice to give them a name that makes it clear what they apply to. E.g. in this case I think this should be something like `RequiredEncodingConstantValue`.

ruby/ql/lib/codeql/ruby/ast/internal/Constant.qll
```diff
@@ -0,0 +1,32 @@
...
+    any(RequiredConstantValue x).requiredComplex(real, imaginary)
+  } or
+  TString(string s) { any(RequiredConstantValue x).requiredString(s) } or
+  TSymbol(string s) { any(RequiredConstantValue x).requiredSymbol(s) } or
+  TBoolean(boolean b) { b in [false, true] } or
+  TNil()
+
+private newtype TRequiredConstantValue = MkRequiredConstantValue()
+
+class RequiredConstantValue extends MkRequiredConstantValue {
```
hmac, line 18: I don't really understand the meaning of "required" in this name. This seems to be a class that holds a set of predicates defining all constant values, but I'm not sure what is "required" about them. Maybe doc comment here would help clarify?

ruby/ql/test/library-tests/ast/ValueText.expected
```diff
@@ -355,29 +357,27 @@
...
-| literals/literals.rb:7:1:7:4 | TRUE | TRUE |
+| literals/literals.rb:7:1:7:4 | TRUE | true |
 | literals/literals.rb:10:1:10:4 | 1234 | 1234 |
-| literals/literals.rb:11:1:11:5 | 5_678 | 5_678 |
+| literals/literals.rb:11:1:11:5 | 5_678 | 5678 |
 | literals/literals.rb:12:1:12:1 | 0 | 0 |
-| literals/literals.rb:13:1:13:5 | 0d900 | 0d900 |
-| literals/literals.rb:16:1:16:6 | 0x1234 | 0x1234 |
-| literals/literals.rb:17:1:17:10 | 0xdeadbeef | 0xdeadbeef |
-| literals/literals.rb:18:1:18:11 | 0xF00D_face | 0xF00D_face |
```
hmac, line 369: We appear to have lost these two hexadecimal literals.

ruby/ql/test/library-tests/ast/ValueText.expected
```diff
@@ -355,29 +357,27 @@
...
+| literals/literals.rb:26:1:26:10 | 0b10010100 | 148 |
+| literals/literals.rb:27:1:27:11 | 0B011_01101 | 109 |
 | literals/literals.rb:30:1:30:5 | 12.34 | 12.34 |
-| literals/literals.rb:31:1:31:7 | 1234e-2 | 1234e-2 |
-| literals/literals.rb:32:1:32:7 | 1.234E1 | 1.234E1 |
-| literals/literals.rb:35:1:35:3 | 23r | 23r |
-| literals/literals.rb:36:1:36:5 | 9.85r | 9.85r |
-| literals/literals.rb:39:1:39:2 | 2i | 2i |
+| literals/literals.rb:31:1:31:7 | 1234e-2 | 12.34 |
+| literals/literals.rb:32:1:32:7 | 1.234E1 | 12.34 |
```
hmac, line 377: I guess this is because we're relying on `string::toFloat` to parse floats at the moment?

hmac (commented):
This looks great - I just had a few questions on some edge cases.

ruby/ql/lib/codeql/ruby/ast/Literal.qll
hvitved: That would have to be done in `BinaryOperationCfgNode`. We could add constant folding for complex and rational numbers there, but I think that should be done as follow-up work.

ruby/ql/lib/codeql/ruby/ast/internal/Constant.qll
hvitved: I'll add ql doc.

ruby/ql/test/library-tests/ast/ValueText.expected
hvitved: Good spot, I'll fix.

ruby/ql/test/library-tests/ast/ValueText.expected
hvitved: yes

ruby/ql/lib/codeql/ruby/ast/Literal.qll
hmac: Unless I'm missing something, I don't think so? `1+2i` is parsed as a single `ComplexLiteral` value, not an application of `+` to two separate literals.

hmac: (Totally happy if we do this as follow-up work, but I think this is the right part of the code to be changing)

hvitved: 👍 Let's do it as follow-up.

ruby/ql/test/library-tests/ast/ValueText.expected
```diff
@@ -367,6 +367,8 @@
 | literals/literals.rb:12:1:12:1 | 0 | 0 |
 | literals/literals.rb:13:1:13:5 | 0d900 | 0 |
 | literals/literals.rb:16:1:16:6 | 0x1234 | 4660 |
+| literals/literals.rb:17:1:17:10 | 0xdeadbeef | -559038737 |
+| literals/literals.rb:18:1:18:11 | 0xF00D_face | -267519282 |
```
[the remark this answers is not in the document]
hmac: Just a guess, is CodeQL's `int` a signed 32-bit and these values are overflowing it?

hmac (approved):
This is a real nice improvement!

ruby/ql/test/library-tests/ast/ValueText.expected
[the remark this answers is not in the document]
hvitved: OK, presumably because Ruby uses 64 bit integers, and QL uses 32 bits.

PR #7680, opened 2022-01-20
VSCode: Autosave when running query

JLLeitschuh:
Configure VS Code to autosave when running a query from within the IDE. This prevents an annoying pop-up every time you attempt to execute a CodeQL query from a file that hasn't been saved yet. It also guarntees that what you're seeing in the IDE will be what gets exceuted by the CodeQL plugin.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,3 +1,4 @@
 {
-    "omnisharp.autoStart": false
+    "omnisharp.autoStart": false,
+    "codeQL.runningQueries.autoSave": true
 }


JLLeitschuh:
👋  anyone able to take a look at this?

owen-mc:
I already have this setting. I have no memory of ever setting it. I can imagine it would be annoying to get a pop-up that you describe.

aeisenberg:
We're chatting about this internally. I personally don't have any strong feeling about having it for myself. I do have a concern that adding this setting to the workspace will override any user setting and will prevent a user who does not want autosave for their queries.

Is there a reason why setting it globally isn't working for you?

JLLeitschuh:
There didn't seem to be a pop-up option allowing this to be set globally when I experienced it

PR #6779, opened 2021-09-30
Java: CWE-927 - Query to detect the use of implicit PendingIntents

atorralba:
This PR introduces a query that detects the issue described below.

## Description

A [`PendingIntent`](https://developer.android.com/reference/android/app/PendingIntent) describes an action in the form of an Intent that is intended to be given and executed at a later time by another application. The Intent wrapped by a `PendingIntent` is executed on behalf of the application that created it, and with its same privileges.

If a `PendingIntent` is configured to be mutable, the fields of its internal Intent can be changed by the receiving application if they were not previously set. This means that a mutable `PendingIntent` that has not defined a destination component (that is, an implicit `PendingIntent`) can be altered to execute an arbitrary action with the privileges of the application that created it.

If an implicit `PendingIntent` is obtainable by a malicious application by any of the following means:

* It is wrapped and sent as an extra of another implicit Intent
* It is sent as the action of a `Slide`
* It is sent as the action of a `Notification`

the attacker could modify the underlying Intent and execute an arbitrary action with elevated privileges. This could give the malicious application access to private components of the victim application, or the ability to perform actions without having the necessary permissions.

## References

* [All About PendingIntents](https://medium.com/androiddevelopers/all-about-pendingintents-748c8eb8619)
* [Remediation for Implicit PendingIntent Vulnerability](https://support.google.com/faqs/answer/10437428?hl=en)
* [PIAnalyzer: A precise approach for PendingIntent vulnerability analysis](https://www.cs.uni-potsdam.de/se/papers/esorics18.pdf)

PR #7681, opened 2022-01-20
Java: Improvements to the Android query Use of implicit PendingIntents

atorralba:
This PR introduces fixes and improvements on top of github/codeql#6779.

### DIFF
--- a/java/ql/src/Security/CWE/CWE-927/ImplicitPendingIntents.qhelp
+++ b/java/ql/src/Security/CWE/CWE-927/ImplicitPendingIntents.qhelp
@@ -27,7 +27,7 @@
 <recommendation>
 <p>Avoid creating implicit <code>PendingIntent</code>s. This means that the underlying <code>Intent</code> should always have an 
   explicit destination component.</p>
-<p>Also, when adding the <code>PendingIntent</code> as an extra of another <code>Intent</code>, make sure that said <code>Intent</code> also has
+<p>When you add the <code>PendingIntent</code> as an extra of another <code>Intent</code>, make sure that this second <code>Intent</code> also has
 an explicit destination component, so that it is not delivered to untrusted applications.</p>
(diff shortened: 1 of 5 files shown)


java/ql/src/Security/CWE/CWE-927/ImplicitPendingIntents.qhelp
```diff
@@ -5,42 +5,42 @@
...
+  application, or the ability to perform actions without having the necessary permissions.</p>
 </overview>
 
 <recommendation>
-<p>Avoid creating implicit <code>PendingIntent</code>s. This means that the underlying Intent should always have an 
+<p>Avoid creating implicit <code>PendingIntent</code>s. This means that the underlying <code>Intent</code> should always have an 
   explicit destination component.</p>
-<p>Also, when adding the <code>PendingIntent</code> as an extra of another Intent, make sure that said Intent also has
+<p>Also, when adding the <code>PendingIntent</code> as an extra of another <code>Intent</code>, make sure that said <code>Intent</code> also has
 an explicit destination component, so that it is not delivered to untrusted applications.</p>
```
felicitymay, line 31: Thanks for the explanation in: https://github.com/github/codeql/pull/6779#issuecomment-1018375595.

One thing that remains unclear is which `Intent` is meant by "said <code>Intent</code>". I think what we mean is that all `Intent`s should have an explicit destination component. So maybe:

```suggestion
<p>When you add the <code>PendingIntent</code> as an extra of another <code>Intent</code>, make sure that this second <code>Intent</code> also has
an explicit destination component, so that it is not delivered to untrusted applications.</p>
```

felicitymay (dismissed):
Thanks for opening this PR with changes to the text ✨

java/ql/src/Security/CWE/CWE-927/ImplicitPendingIntents.qhelp
atorralba: Exactly. Good suggestion, thanks!

aschackmull (approved):
LGTM

PR #4644, opened 2020-11-10
C++: Add 'unsafe use of this' query

MathiasVP:
Fixes https://github.com/github/codeql-c-analysis-team/issues/163.

A couple of points are worth making:
- I've restricted the scope of the query to just be uses of `this` in C++ constructors (as suggested in the discussion).
- I've restricted the flow to be 'must' dataflow.

Since this is 'must' flow, I've designed a tiny ad-hoc 'must' dataflow library in the query. I suspect we could move this ad-hoc library out into a library file and generalize it if it turns out to be useful somewhere else.

- I've specialized the query to only flag calls will cause undefined behavior, and not flag somewhat suspicious uses of calls to pure virtual in constructors that will actually resolve correctly to a function.

For instance, this will _not_ be flagged:

```cpp
struct Base { virtual void f() = 0; };

struct Derived : public Base {
  Derived() { f(); }

  void f() override {}
};
```
since the call to `f` in `Derived`'s constructor will call `f`. However, the query does flag this:

```cpp
struct Base { virtual void f() = 0; };

struct Derived : public Base {
  Derived() { call_f(this); }

  void f() override {}
};

void call_f(Base* b) { b->f(); }
```
since this is undefined behavior (and in practice will crash on the compilers I've tried).

I'm open to discussion on whether or not we want to flag the first example. It should only make the implementation simpler.
We do find some results on LGTM even though we throw away the somewhat benign cases: https://lgtm.com/query/3109442847251804192/.

TODO:

- [x] Write qhelp.
- [x] Check whether we need to add any CVE tags.

PR #7682, opened 2022-01-20
C++: Use the IR for `cpp/return-stack-allocated-memory`.

MathiasVP:
This query is a perfect query for the IR. Without really thinking too hard about it I was able to remove all the false positives from our tests for this query 🎉.

I changed the behavior of the query to not report 'may' flow, but only 'must' flow. This removes the possibility of FPs from complex path conditions which we cannot reason about.

This has no influence on Samate since Samate has a very low number of tests for this CWE. It does find a bunch of good results on LGTM, though (I've provided a link in the internal issue).


geoffw0 (approved):
I'm a bit sad that so much query code is needed that feels like it should be done by a library.  But the results on tests are very impressive, the upgrade to a `path-problem` is progress, and real world results I've looked at LGTM.  :+1:

MathiasVP:
> I'm a bit sad that so much query code is needed that feels like it should be done by a library. 

Yeah, most of the code is really because I re-invented the 'must flow' library like I did in https://github.com/github/codeql/pull/4644. I think we can pull some shared parts out into a "intraprocedural must flow" library in the future, which should hopefully cut most of this code away.

geoffw0:
> I think we can pull some shared parts out into a "intraprocedural must flow" library in the future, which should hopefully cut most of this code away.

Sounds good, and I agree this should be done outside of this PR.

PR #7683, opened 2022-01-20
Ruby: Use multiple threads in QL test CI job

hvitved:


.github/workflows/ruby-qltest.yml
```diff
@@ -32,14 +32,14 @@ jobs:
       - uses: ./ruby/actions/create-extractor-pack
       - name: Run QL tests
         run: |
-          codeql test run --search-path "${{ github.workspace }}/ruby/extractor-pack" --check-databases --check-unused-labels --check-repeated-labels --check-redefined-labels --check-use-before-definition --consistency-queries ql/consistency-queries ql/test
+          codeql test run --threads=0 --ram 5000 --search-path "${{ github.workspace }}/ruby/extractor-pack" --check-databases --check-unused-labels --check-repeated-labels --check-redefined-labels --check-use-before-definition --consistency-queries ql/consistency-queries ql/test
```
hmac, line 35: The [Actions docs](https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners) state that the VM has 7GB of RAM, so maybe we can bump this even further?

aibaars (approved):

PR #7688, opened 2022-01-21
Data flow: Restructure `RequiredSummaryComponentStack`

hvitved:
This avoids non-linear recursion through a self-join of `interpretSpec`:

```
[2022-01-21 09:04:55] (24s) Tuple counts for dom#FlowSummaryImpl::Private::TConsSummaryComponentStack#ff/2@i28#07224gwr after 14.1s:
                      0         ~0%         {2} r1 = SCAN FlowSummaryImpl::Private::derivedFluentFlowPush#ffffff#reorder_3_4_5_0_1_2#prev_delta OUTPUT In.1 'tail', In.0 'head'
                      
                      110216000 ~6471%      {2} r2 = JOIN FlowSummaryImpl::Private::External::interpretSpec#ffff#reorder_3_2_0_1#prev_delta WITH FlowSummaryImpl::Private::External::interpretSpec#ffff#reorder_3_2_0_1#prev ON FIRST 1 OUTPUT Lhs.0 'tail', Rhs.1 'head'
                      
                      110215000 ~6471%      {2} r3 = JOIN FlowSummaryImpl::Private::External::interpretSpec#ffff#reorder_3_2_0_1#prev WITH FlowSummaryImpl::Private::External::interpretSpec#ffff#reorder_3_2_0_1#prev_delta ON FIRST 1 OUTPUT Lhs.0, Rhs.1
                      
                      220431000 ~13042%     {2} r4 = r2 UNION r3
                      220431000 ~13042%     {2} r5 = r1 UNION r4
                      6027000   ~259%       {2} r6 = r5 AND NOT dom#FlowSummaryImpl::Private::TConsSummaryComponentStack#ff#prev(Lhs.1 'head', Lhs.0 'tail')
                      6027000   ~269%       {2} r7 = SCAN r6 OUTPUT In.1 'head', In.0 'tail'
                                            return r7
```

### DIFF
--- a/csharp/ql/lib/semmle/code/csharp/dataflow/internal/FlowSummaryImpl.qll
+++ b/csharp/ql/lib/semmle/code/csharp/dataflow/internal/FlowSummaryImpl.qll
@@ -171,15 +171,20 @@ module Public {
     )
   }
 
+  private newtype TRequiredSummaryComponentStack = MkRequiredSummaryComponentStack()
+
   /**
    * A class that exists for QL technical reasons only (the IPA type used
    * to represent component stacks needs to be bounded).
    */
-  abstract class RequiredSummaryComponentStack extends SummaryComponentStack {
+  class RequiredSummaryComponentStack extends TRequiredSummaryComponentStack {
+    /** Gets a textual representation of this element. */
+    string toString() { none() }
+
     /**
      * Holds if the stack obtained by pushing `head` onto `tail` is required.
      */
-    abstract predicate required(SummaryComponent c);
+    predicate required(SummaryComponent head, SummaryComponentStack tail) { none() }
   }
 
   /** A callable with a flow summary. */
@@ -240,9 +245,9 @@ module Private {
   newtype TSummaryComponentStack =
     TSingletonSummaryComponentStack(SummaryComponent c) or
     TConsSummaryComponentStack(SummaryComponent head, SummaryComponentStack tail) {
-      tail.(RequiredSummaryComponentStack).required(head)
+      any(RequiredSummaryComponentStack x).required(head, tail)
       or
-      tail.(RequiredSummaryComponentStack).required(TParameterSummaryComponent(_)) and
+      any(RequiredSummaryComponentStack x).required(TParameterSummaryComponent(_), tail) and
       head = thisParam()
       or
       derivedFluentFlowPush(_, _, _, head, tail, _)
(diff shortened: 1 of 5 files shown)


michaelnebel (dismissed):
Looks good from C# perspective!

csharp/ql/lib/semmle/code/csharp/dataflow/internal/FlowSummaryImpl.qll
```diff
@@ -171,15 +171,20 @@ module Public {
...
   }
 
+  private newtype TRequiredSummaryComponentStack = MkRequiredSummaryComponentStack()
+
   /**
    * A class that exists for QL technical reasons only (the IPA type used
    * to represent component stacks needs to be bounded).
    */
-  abstract class RequiredSummaryComponentStack extends SummaryComponentStack {
+  class RequiredSummaryComponentStack extends TRequiredSummaryComponentStack {
```
aschackmull, line 180: We should have access to the `Unit` type here (provided by `DataFlowImplSpecific::Private`), so might as well use that.

csharp/ql/lib/semmle/code/csharp/dataflow/internal/FlowSummaryImpl.qll
```diff
@@ -171,15 +171,20 @@ module Public {
...
+    /** Gets a textual representation of this element. */
+    string toString() { none() }
+
     /**
      * Holds if the stack obtained by pushing `head` onto `tail` is required.
      */
-    abstract predicate required(SummaryComponent c);
+    predicate required(SummaryComponent head, SummaryComponentStack tail) { none() }
```
aschackmull, line 187: Might as well make this `abstract`.

csharp/ql/lib/semmle/code/csharp/dataflow/internal/FlowSummaryImpl.qll
```diff
@@ -171,15 +171,20 @@ module Public {
...
+  private newtype TRequiredSummaryComponentStack = MkRequiredSummaryComponentStack()
+
   /**
    * A class that exists for QL technical reasons only (the IPA type used
    * to represent component stacks needs to be bounded).
    */
-  abstract class RequiredSummaryComponentStack extends SummaryComponentStack {
+  class RequiredSummaryComponentStack extends TRequiredSummaryComponentStack {
+    /** Gets a textual representation of this element. */
+    string toString() { none() }
```
aschackmull, line 182: Using `Unit` means that this `toString` is unnecessary. (And having a singleton value for `toString` like `Unit` provides is probably slightly preferable to `none()`)

aschackmull (commented):
Some minor comments, but otherwise LGTM.

csharp/ql/lib/semmle/code/csharp/dataflow/internal/FlowSummaryImpl.qll
hvitved: I am not a fan of `abstract` predicates in non-`abstract` classes, but ok.

aschackmull: Specifically on unit-classes I think it makes sense.

aschackmull (approved):

michaelnebel (approved):
Great!

PR #7700, opened 2022-01-21
JS: Fix copy/paste error in XSS ML-powered queries results patterns

henrymercer:
We didn’t catch this because our unit tests test only library code due to the previous difficulty of running queries with an ML model (the ML models in packs work should fix that), and because the end-to-end evaluation runs separate queries that have different result patterns.

Going forward we should create unit tests for the queries themselves, which will require using the ML model in tests. We should also be able to catch this type of error using DCA.


Z80coder (approved):

esbena:
FWIW https://github.com/github/codeql/commit/63de4ba93956c608dfa5ab83be5185f3c07bc361 contains StoredXssATM.ql as an entirely generated file, so this kind of bug will become less likely.

PR #7701, opened 2022-01-21
C++: Remove FPs from `cpp/return-stack-allocated-memory`

MathiasVP:
Quick follow-up to https://github.com/github/codeql/pull/7682 that removes results when the enclosing function's name suggests that its job is to return the value of the stack pointer.

See internal issue for differences.


geoffw0 (approved):
There are surprisingly many of these results.

:+1:

PR #7702, opened 2022-01-21
Java: Remove some JNDI Injection sinks

### LINKED ISSUE #7699: JNDI Injection - false positive
**Description of the false positive**

As reported by @JLLeitschuh on Slack the JNDI injection reported on [OpenClinica](https://lgtm.com/projects/g/OpenClinica/OpenClinica/snapshot/5c237e1d5e9d95572822e3c42c6dcdc9d03e740d/files/core/src/main/java/org/akaza/openclinica/service/user/LdapUserService.java?sort=name&dir=ASC&mode=heatmap#L145) is a false positive.

The `lookup` overload invoked is [`lookup(final String dn, final AttributesMapper<T> mapper)`](https://github.com/spring-projects/spring-ldap/blob/main/core/src/main/java/org/springframework/ldap/core/LdapTemplate.java#L868-L876). This overload does not call `DirContext.lookup` but just `DirContext.getAttributes()`:

```java
	public <T> T lookup(final String dn, final AttributesMapper<T> mapper) {

		return executeReadOnly(new ContextExecutor<T>() {
			public T executeWithContext(DirContext ctx) throws javax.naming.NamingException {
				Attributes attributes = ctx.getAttributes(dn);
				return mapper.mapFromAttributes(attributes);
			}
		});
	}
```

This sink was originally introduced [here](https://github.com/github/codeql/pull/3288/files#diff-b886eb20c5a31e4bf0a6cc3a366fcfd9ea37a874314c3a17d82110a31e7461aaR125)

All the `LdapTemplate.lookup` sinks taking an `AttributeMapper` object on any of their parameters should be excluded from the sink:

- `lookup(Name dn, AttributesMapper<T> mapper)`
- `lookup(Name dn, String[] attributes, AttributesMapper<T> mapper)`
- `lookup(String dn, AttributesMapper<T> mapper)`
- `lookup(String dn, String[] attributes, AttributesMapper<T> mapper)`

atorralba:
Removes JNDI Injection sinks that were causing false positives: `LdapOperations.lookup` methods receiving an `AttributesMapper` parameter don't actually call `DirContext.lookup`, but only `DirContext.getAttributes`. Also, adds tests and stubs for these cases.

Fixes #7699.

### DIFF
--- a/java/ql/lib/semmle/code/java/security/JndiInjection.qll
+++ b/java/ql/lib/semmle/code/java/security/JndiInjection.qll
@@ -88,7 +88,12 @@ private class DefaultJndiInjectionSinkModel extends SinkModelCsv {
         // Spring
         "org.springframework.jndi;JndiTemplate;false;lookup;;;Argument[0];jndi-injection",
         // spring-ldap 1.2.x and newer
-        "org.springframework.ldap.core;LdapOperations;true;lookup;;;Argument[0];jndi-injection",
+        "org.springframework.ldap.core;LdapOperations;true;lookup;(Name);;Argument[0];jndi-injection",
+        "org.springframework.ldap.core;LdapOperations;true;lookup;(Name,ContextMapper);;Argument[0];jndi-injection",
+        "org.springframework.ldap.core;LdapOperations;true;lookup;(Name,String[],ContextMapper);;Argument[0];jndi-injection",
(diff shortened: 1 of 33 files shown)


aschackmull (commented):
LGTM at a glance, but the test fails.

atorralba:
@aschackmull stubs needed fixing. All checks are now green.

aschackmull (approved):

PR #7703, opened 2022-01-21
C++: Split 'gets' model.

geoffw0:
As I understand it `gets` should be a `LocalFlowSourceFunction` not a `RemoteFlowSourceFunction` (unlike `fgets` which _may_ at least be a remote source).  Fixing this required splitting the model in two, though a few parts did simplify afterwards.

This change is mostly for correctness but will fix the odd false positive, such as the test case added for `cpp/cleartext-transmission` (which confusingly is and remains a true positive for  `cpp/cleartext-storage-buffer`, tested alongside, because that query is defined to not care about it being a remote source).


cpp/ql/lib/semmle/code/cpp/models/implementations/Gets.qll
```diff
@@ -54,17 +53,60 @@ private class GetsFunction extends DataFlowFunction, TaintFunction, ArrayFunctio
...
 
   override predicate hasArrayWithVariableSize(int bufParam, int countParam) {
-    not this.hasName("gets") and
     bufParam = 0 and
     countParam = 1
   }
 
+  override predicate hasArrayOutput(int bufParam) { bufParam = 0 }
+
+  override predicate hasSocketInput(FunctionInput input) { input.isParameter(2) }
```
MathiasVP, line 62: This isn't really something we need to fix in this PR, but should this really be `isParameter(2)` and not `isParameterDeref(2)`? Presumably, `isParameter(2)` should only hold if `fgets` had a prototype like
```cpp
char *fgets(char *str, int n, FILE stream)
```
(i.e., taking the `FILE` argument by value).

I see that all of our implementations of `hasSocketInput` uses `isParameter`, so it's probably not something we should change in this PR.

cpp/ql/lib/semmle/code/cpp/models/implementations/Gets.qll
```diff
@@ -54,17 +53,60 @@ private class GetsFunction extends DataFlowFunction, TaintFunction, ArrayFunctio
...
+private class GetsFunction extends DataFlowFunction, TaintFunction, ArrayFunction, AliasFunction,
+  SideEffectFunction, LocalFlowSourceFunction {
+  GetsFunction() {
+    // gets(str)
+    this.hasGlobalOrStdOrBslName("gets")
+  }
+
+  override predicate hasDataFlow(FunctionInput input, FunctionOutput output) {
+    input.isParameter(0) and
+    output.isReturnValue()
```
MathiasVP, line 78: Hm, shouldn't there also be a case like:
```ql
input.isParameterDeref(0) and
output.isReturnValueDeref(0)
```
to model the fact that we also propagate the pointed-to value to the returned pointed-to value?

(I see that we don't do that for the original model either, so that's probably _also_ a follow-up to this PR).

MathiasVP (approved):
LGTM! Both of my comments are really something that we should fix as follow-ups, so feel free to merge this PR without changing anything.

cpp/ql/lib/semmle/code/cpp/models/implementations/Gets.qll
geoffw0: It depends whether we're thinking about the `char *` pointer as a pointer to a buffer or a pointer to a string.  If the latter (and there isn't an error), any existing string is overwritten with new data from `stdin` so there's no data flow.

MathiasVP: Oh, yes. You're right 👍.

cpp/ql/lib/semmle/code/cpp/models/implementations/Gets.qll
geoffw0: I don't think it matters as long as we're consistent.  Note that `hasSocketInput` is not describing the data coming from the socket, just the parameter the socket is specified by (though perhaps the QLDoc could be clearer on this).

MathiasVP: True. We lose the option to distinguish between functions that take a socket inputs by value and by functions that take socket inputs by pointer or reference, though. So if we need that distinction at some point we need to go back and update the implementations to correctly reflect this.

geoffw0: Good point.

geoffw0: Fixed.  In fact most cases are correctly `isParameter`, because sockets (as opposed to streams) are expressed as a simple `int`, e.g.
```
ssize_t recv(int sockfd, void *buf, size_t len, int flags);
```

geoffw0:
Fixed formatting.

cpp/ql/lib/semmle/code/cpp/models/implementations/Gets.qll
MathiasVP: Good point. Thanks for looking into it!

cpp/ql/lib/semmle/code/cpp/models/implementations/Gets.qll
```diff
@@ -51,20 +50,65 @@ private class GetsFunction extends DataFlowFunction, TaintFunction, ArrayFunctio
...
   }
 
-  override predicate hasArrayOutput(int bufParam) { bufParam = 0 }
+  override predicate hasTaintFlow(FunctionInput input, FunctionOutput output) {
```
MathiasVP, line 83: Since this is now `none()` I guess we should just not extend `TaintFunction`?

geoffw0: Yep, good point.

Done.

MathiasVP (approved):
LGTM!

PR #7704, opened 2022-01-21
C++: Another improvement to cpp/cleartext-transmission

geoffw0:
This fixes the false positive we can see in https://lgtm.com/query/301759231708277869/ (note the type `cencrypted_v2_password_header`).  Its fairly narrow, but we need to do all we can to get this query up to `@precision high`.

No change note because its close enough to https://github.com/github/codeql/pull/7650 and covered by the change note there.


MathiasVP (approved):
LGTM!

PR #7713, opened 2022-01-23
Ruby: Add `rb/clear-text-logging-sensitive-data` query

alexrford:
Tracks cases where potentially sensitive data (e.g. passwords and other credentials) are logged as cleartext.

This is loosely based on the JS version of the same query, using the same `SensitiveDataHeuristics.qll` and a similar structure for the `CleartextLoggingCustomizations.qll`.


ruby/ql/lib/codeql/ruby/security/CleartextLoggingCustomizations.qll
```diff
@@ -0,0 +1,311 @@
...
+  }
+
+  /**
+   * A hash literal with an entry that may contain a password
+   */
+  private class HashLiteralPasswordSource extends Source {
+    private string name;
+
+    HashLiteralPasswordSource() {
+      exists(DataFlow::Node val, HashLiteral lit |
```
hvitved, line 213: Use `HashLiteralCfgNode` and `PairCfgNode` node below, instead.

hvitved (changes requested):

ruby/ql/lib/codeql/ruby/security/CleartextLoggingCustomizations.qll
```diff
@@ -0,0 +1,311 @@
...
+
+  // `writeNode` assigns pair with key `name` to `val`
+  private predicate hashKeyWrite(DataFlow::CallNode writeNode, string name, DataFlow::Node val) {
+    exists(SetterMethodCall setter |
+      setter = writeNode.asExpr().getExpr() and
+      // hash[name]
+      setter.getArgument(0).getConstantValue().getStringOrSymbol() = name and
+      // val
+      setter.getArgument(1).(Assignment).getRightOperand() = val.asExpr().getExpr()
+    )
```
hmac, line 168: I think this will also include things like `foo.name = val`, but maybe that's fine?

ruby/ql/lib/codeql/ruby/security/CleartextLoggingCustomizations.qll
```diff
@@ -0,0 +1,311 @@
...
+        not val instanceof NonCleartextPassword and
+        (
+          // hash[name] = val
+          hashKeyWrite(this, name, val) and
+          recv = this.(DataFlow::CallNode).getReceiver()
+        )
+      )
+    }
+
+    override string describe() { result = "an write to " + name }
```
hmac, line 192:
```suggestion
    override string describe() { result = "a write to " + name }
```

ruby/ql/lib/codeql/ruby/security/internal/SensitiveDataHeuristics.qll
```diff
@@ -0,0 +1,129 @@
...
+    result = maybeCertificate() and
+    classification = SensitiveDataClassification::certificate()
+  }
+
+  /**
+   * Gets a regular expression that identifies strings that may indicate the presence of data
+   * that is hashed or encrypted, and hence rendered non-sensitive, or contains special characters
+   * suggesting nouns within the string do not represent the meaning of the whole string (e.g. a URL or a SQL query).
+   */
+  string notSensitiveRegexp() {
```
hmac, line 100: I realise this is shared code, but it's a shame that we have inconsistent naming here. `maybePassword` etc are also regexes but don't have the `Regexp` suffix in their names. Not necessarily something we should change in this PR but the code would read a lot easier if we had a consistent naming scheme.

ruby/ql/lib/codeql/ruby/security/CleartextLoggingQuery.qll
```diff
@@ -0,0 +1,38 @@
...
+ */
+class Configuration extends TaintTracking::Configuration {
+  Configuration() { this = "CleartextLogging" }
+
+  override predicate isSource(DataFlow::Node source) { source instanceof CleartextLogging::Source }
+
+  override predicate isSink(DataFlow::Node sink) { sink instanceof CleartextLogging::Sink }
+
+  override predicate isSanitizer(DataFlow::Node node) {
+    super.isSanitizer(node)
```
hmac, line 26: `TaintTracking::Configuration.isSanitizer(_)` is `none()`, so isn't this line redundant?

ruby/ql/lib/codeql/ruby/security/CleartextLoggingCustomizations.qll
```diff
@@ -0,0 +1,311 @@
...
+  }
+
+  private string commonLogMethodName() {
+    result = ["info", "debug", "warn", "warning", "error", "log"]
+  }
+
+  /** Holds if `nodeFrom` taints `nodeTo`. */
+  predicate isAdditionalTaintStep(DataFlow::Node nodeFrom, DataFlow::Node nodeTo) {
+    exists(string name, ElementReference ref, LocalVariable hashVar |
+      // from `hsh[password] = "changeme"` to a `hsh[password]` read
```
hmac, line 282: Can this additional step be dropped once we support flow through hashes?

ruby/ql/test/query-tests/security/cwe-312/logging.rb
```diff
@@ -0,0 +1,84 @@
...
+stdout_logger.progname = password
+
+hsh1 = { password: "aec5058e61f7f122998b1a30ee2c66b6" }
+hsh2 = {}
+# GOOD: no backwards flow
+stdout_logger.info hsh2[:password]
+hsh2[:password] = "beeda625d7306b45784d91ea0336e201"
+
+# BAD: password logged as plaintext
+stdout_logger.info hsh1[:password]
```
hmac, line 37: Unless I've missed something, I don't think there's a result in the `.expected` file for this case.

hmac (commented):
I've left a few minor comments and questions. In general this looks super thorough, but I have trouble following some of the logic in `CleartextLoggingCustomizations.qll`, I think because we deal with heuristics both for "sensitive" data and "not sensitive" data. Hopefully a second read through will make it a bit clearer for me.

ruby/ql/lib/codeql/ruby/security/CleartextLoggingCustomizations.qll
```diff
@@ -0,0 +1,311 @@
...
+   * password.sub!(/.+/, "")
+   * Logger.new(STDOUT).info password
+   * ```
+   */
+  private class MaskingReplacerSanitizedNode extends SanitizerIn {
+    MaskingReplacerSanitizedNode() {
+      exists(MaskingReplacerSanitizer maskCall, Variable v |
+        maskCall.getMethodName() = ["sub!", "gsub!"] and
+        v = maskCall.getReceiver().asExpr().getExpr().(VariableReadAccess).getVariable() and
+        v = this.asExpr().getExpr().(VariableReadAccess).getVariable() and
```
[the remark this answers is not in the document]
hvitved: Yes, using SSA would be better, as it would also ensure that this variable access and the receiver of the `maskCall` actually refer to the same value. We will also be able to get rid of the manual successor gymnastics:

```ql
    MaskingReplacerSanitizedNode() {
      exists(Ssa::Definition def |
        exists(MaskingReplacerSanitizer maskCall |
          maskCall.getMethodName() = ["sub!", "gsub!"] and
          def.hasAdjacentReads(maskCall.getReceiver().asExpr(), this.asExpr())
        )
        or
        def.hasAdjacentReads(any(MaskingReplacerSanitizedNode read).asExpr(), this.asExpr())
      )
    }
```

ruby/ql/test/query-tests/security/cwe-312/logging.rb
alexrford: I must have missed this - Tom's suggested CFG node matching appears to fix this.

ruby/ql/lib/codeql/ruby/security/CleartextLoggingCustomizations.qll
```diff
@@ -0,0 +1,311 @@
...
+        maskCall.asExpr().getASuccessor*() = this.asExpr()
```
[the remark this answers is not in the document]
alexrford: I think that limiting restricting this to `SanitizerIn` rather than just `Sanitizer` was unnecessary, I'll change this to just being a `Sanitizer` and remove the `SanitizerIn` related code in this query.

As a side note (and a reminder to myself), this is a separate case to `MaskingReplacerSanitizer` because `sub!`/`gsub!` mutate the receiver object. In the case of:
```
foo = "password"
foo = foo.sub(/.*/, "[redacted]")
logger.info(foo)
```
there is an edge from the second definition of `foo = foo.sub(/.*/, "[redacted]")` to the read of `foo` in the `logger.info(foo)` call and no edge directly from the first definition `foo = "password"` to the last read of `foo` because `foo` is redefined. On the other hand, in:
```
foo = "password"
foo.sub!(/.*/, "[redacted]")
logger.info(foo)
```
the edge is from the definition of `foo = "password"` straight to the read of `foo` in the `logger.info(foo)` call, without any edges involving the `sub!` call.

ruby/ql/lib/codeql/ruby/security/CleartextLoggingCustomizations.qll
```diff
@@ -0,0 +1,311 @@
...
+  }
+
+  /**
+   * A data flow node that receives flow that is not a clear-text password.
+   */
+  private class NonCleartextPasswordFlow extends NonCleartextPassword {
+    NonCleartextPasswordFlow() {
+      any(NonCleartextPassword other).(DataFlow::LocalSourceNode).flowsTo(this)
+    }
+  }
```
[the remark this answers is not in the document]
alexrford: Yes, that's the kind of pattern that this is intended to filter out. I'll add a test case for it.

ruby/ql/lib/codeql/ruby/security/CleartextLoggingCustomizations.qll
alexrford: In practice this doesn't match against those assignments, though this is more down to the details of how we desugar these assignments than an intentional choice. Something like:
```
      // foo.name = val
      setter.getMethodName() = name + "=" and
      setter.getArgument(0).(Assignment).getRightOperand() = val.asExpr().getExpr()
```
would additionally match such methods, though we wouldn't automatically recognize subsequent calls to `foo.name` as "accesses" to that value, as far as I know.

ruby/ql/lib/codeql/ruby/security/internal/SensitiveDataHeuristics.qll
alexrford: Yes, agreed. It'll be worth fixing this as a followup - and this seems to have already started (e.g. with `deprecated predicate maybeSensitive = maybeSensitiveRegexp/1;` in this file).

ruby/ql/lib/codeql/ruby/security/CleartextLoggingQuery.qll
alexrford: I added it in case we extend the default `isSanitizer` predicate in the future, though I haven't been consistent in doing this with all dataflow/taint tracking queries.

ruby/ql/lib/codeql/ruby/security/CleartextLoggingCustomizations.qll
alexrford: Yes, I expect so.

ruby/ql/lib/codeql/ruby/security/CleartextLoggingCustomizations.qll
```diff
@@ -0,0 +1,312 @@
...
+      exists(Ssa::Definition def |
+        exists(MaskingReplacerSanitizer maskCall |
+          maskCall.getMethodName() = ["sub!", "gsub!"] and
+          def.hasAdjacentReads(maskCall.getReceiver().asExpr(), this.asExpr())
+        )
+        or
+        def.hasAdjacentReads(any(MaskingReplacerSanitizedNode read).asExpr(), this.asExpr())
+      )
+    }
+  }
```
hmac, line 92: I think this can be simplified by replacing it with something like this:
```ql
  /**
   * Like `MaskingReplacerSanitizer` but updates the receiver.
   * Taint is thereby cleared for any subsequent read.
   */
  private class InPlaceMaskingReplacerSanitizer extends Sanitizer {
    InPlaceMaskingReplacerSanitizer() {
      exists(MaskingReplacerSanitizer m | m.getMethodName() = ["gsub!", "sub!"] |
        m.getReceiver() = this
      )
    }
  }
```

The idea being that we make the _receiver_ of `sub!` be a sanitizer, which (I believe) removes it and any subsequent reads of the same variable from the dataflow graph.

hmac (approved):
This looks great. I understand it better on the second read through!

I've left one comment about a possible simiplifcation, but otherwise :+1:. I've also kicked off a DCA run that should be done by the time you read this.

ruby/ql/lib/codeql/ruby/security/CleartextLoggingCustomizations.qll
alexrford: Ah, that's really nice - much neater and truer to the intent of this sanitizer.

hmac (approved):

PR #7715, opened 2022-01-24
JS: add a predicate to recognize path arguments in calls to the fs-extra lib

kaeluka:
I'm sending this PR because the lack of complete support for `fs-extra` (ca 60.000.000 downloads weekly) has led to at least one false negative (see here: https://lgtm.com/projects/g/strongloop/strong-arc/snapshot/ac18d2c147b10c94552f215b1016216bb580d3ff/files/client/test/test-server.js?sort=name&dir=ASC&mode=heatmap#L89, it's a FN since fs.remove is not currently recognized as a sink. This PR fixes the problem.)

### DIFF
--- a/javascript/ql/lib/semmle/javascript/frameworks/NodeJSLib.qll
+++ b/javascript/ql/lib/semmle/javascript/frameworks/NodeJSLib.qll
@@ -408,7 +408,7 @@ module NodeJSLib {
 
   /**
    * Holds if the `i`th parameter of method `methodName` of the Node.js
-   * `fs` module or the `fs-extra` might represent a file path.
+   * `fs` module or the `fs-extra` module might represent a file path.
    *
    * For `fs`, we determine this by looking for an externs declaration for
    * `fs.methodName` where the `i`th parameter's name is `filename` or
@@ -435,13 +435,13 @@ module NodeJSLib {
    * method might represent a file path.
    */
   private predicate fsExtraExtensionFileParam(string methodName, int i) {
-    methodName = ["copy", "copySync", "copyFile"] and i = [0 .. 1]
+    methodName = ["copy", "copySync", "copyFile"] and i = [0, 1]
     or
-    methodName = ["move", "moveSync"] and i = [0 .. 1]
+    methodName = ["move", "moveSync"] and i = [0, 1]
     or
     methodName = ["createFile", "createFileSync"] and i = 0
     or
-    methodName = ["createSymLink", "createSymlinkSync"] and i = [0 .. 1]
+    methodName = ["createSymLink", "createSymlinkSync"] and i = [0, 1]
     or
     methodName = ["ensureDir", "ensureDirSync"] and i = 0
     or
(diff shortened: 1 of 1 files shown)


kaeluka:
Adding @esbena as reviewer, as he is already in the loop.

javascript/ql/lib/semmle/javascript/frameworks/NodeJSLib.qll
```diff
@@ -408,11 +408,13 @@ module NodeJSLib {
...
    * Holds if the `i`th parameter of method `methodName` of the Node.js
-   * `fs` module might represent a file path.
+   * `fs` module or the `fs-extra` might represent a file path.
    *
-   * We determine this by looking for an externs declaration for
+   * For `fs`, we determine this by looking for an externs declaration for
    * `fs.methodName` where the `i`th parameter's name is `filename` or
    * `path` or a variation thereof.
+   *
+   * For `fs-extra`, we use the `fsExtraExtensionFileParam` predicate.
```
esbena, line 417: QLDoc should summarize the semantics, not the implementation. We only have a few technical corners of our libraries that need to refer explicitly to other predicates in their QLDoc.

```suggestion
   * For `fs-extra`, we use a manually maintained list.
```

javascript/ql/lib/semmle/javascript/frameworks/NodeJSLib.qll
```diff
@@ -423,6 +425,47 @@ module NodeJSLib {
...
+    or
+    fsExtraExtensionFileParam(methodName, i)
+  }
+
+  /**
+   * Holds if `methodName` is a function defined in the `fs-extra` library
+   * as an extension to node.js' `fs` module and parameter `i` of of the
+   * method might represent a file path.
+   */
+  private predicate fsExtraExtensionFileParam(string methodName, int i) {
```
esbena, line 437: @asgerf Could I get your opinion here? Should we skip directly to models-as-data for this predicate?

More generally: should we have (rough) guidelines for when we prefer models-as-data over QL implementations?

esbena (changes requested):
Thanks for doing this! I'm excited about the evaluation, could you do one with code-scanning.qls and default.yml?

The semantics of this looks fine, but there are two tangential issues. See inline comments.

javascript/ql/lib/semmle/javascript/frameworks/NodeJSLib.qll
kaeluka: 👍  good question!

asgerf: There's generally no pressure to use MaD at the moment.

esbena (approved):

esbena:
Great. Merge at will.
The results are a bit underwhelming, so I have started an evaluation that just lists the new sinks (see backlink below)

javascript/ql/lib/semmle/javascript/frameworks/NodeJSLib.qll
kaeluka: Alright, that settles it then, @esbena?

PR #7718, opened 2022-01-24
C++: Add `security` tag to `cpp/return-stack-allocated-memory`

MathiasVP:
The query was rewritten to be more robust in https://github.com/github/codeql/pull/7682 and https://github.com/github/codeql/pull/7701 removed all of the FPs that I've seen on LGTM, so I think this query is ready to go into the Code Scanning suite.

This will make the results of `cpp/return-stack-allocated-memory` show up by default.

Before we merge this I'll run a DCA to make sure we don't get any unintended IR re-evaluations.


rdmarsh2 (approved):
DCA looks good

PR #7720, opened 2022-01-24
C#: Desugar property patterns that uses member access syntax.

michaelnebel:
In this PR we make support for [Extended property patterns](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/proposals/csharp-10.0/extended-property-patterns).
The idea is that the new form is *de-sugared* into a pattern expression in the old form. The only difference will be some slightly different locations.

### DIFF
--- a/csharp/extractor/Semmle.Extraction.CSharp/Entities/Expressions/Patterns/PropertyPattern.cs
+++ b/csharp/extractor/Semmle.Extraction.CSharp/Entities/Expressions/Patterns/PropertyPattern.cs
@@ -23,17 +23,7 @@ namespace Semmle.Extraction.CSharp.Entities.Expressions
             }
         }
 
-        private class AccessStep
-        {
-            public readonly string Identifier;
-            public readonly Microsoft.CodeAnalysis.Location Location;
-
-            public AccessStep(string identifier, Microsoft.CodeAnalysis.Location location)
-            {
-                Identifier = identifier;
-                Location = location;
-            }
-        }
+        private record AccessStep(string Identifier, Microsoft.CodeAnalysis.Location Location);
 
         private class AccessStepPack
         {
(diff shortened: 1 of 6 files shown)


csharp/extractor/Semmle.Extraction.CSharp/Entities/Expressions/Patterns/PropertyPattern.cs
```diff
@@ -10,17 +12,91 @@ internal class PropertyPattern : Expression
...
-                    Context.ModelError(sub, "Expected to find 'Name:' in pattern.");
+                    Context.ModelError(sub, "Expected to find 'Expression:' in pattern.");
                     continue;
                 }
-                trapFile.exprorstmt_name(p, sub.NameColon.Name.ToString());
+                MakeExpressions(cx, this, sub, child++);
+            }
+        }
+
+        private class AccessStep
```
hvitved, line 26: Use `private record AccessStep(string Identifier, Microsoft.CodeAnalysis.Location Location)` instead?

csharp/extractor/Semmle.Extraction.CSharp/Entities/Expressions/Patterns/PropertyPattern.cs
```diff
@@ -10,17 +12,91 @@ internal class PropertyPattern : Expression
...
+
+            public AccessStepPack(string identifier, Microsoft.CodeAnalysis.Location location)
+            {
+                Last = new AccessStep(identifier, location);
+            }
+        }
+
+        private static AccessStepPack GetAccessStepPack(ExpressionSyntax syntax)
+        {
+            switch (syntax)
```
hvitved, line 58: Can use a switch expression instead.

csharp/extractor/Semmle.Extraction.CSharp/Entities/Expressions/Patterns/PropertyPattern.cs
```diff
@@ -10,17 +12,91 @@ internal class PropertyPattern : Expression
...
+                case IdentifierNameSyntax identifier:
+                    return new AccessStepPack(identifier.Identifier.Text, identifier.GetLocation());
+                default:
+                    throw new InternalError(syntax, "Unexpected expression syntax in property patterns.");
             }
         }
+
+        private static AccessStepPack GetAccessStepPack(BaseExpressionColonSyntax syntax)
+        {
+            switch (syntax)
```
hvitved, line 71: Can use a switch expression instead.

hvitved (commented):
LGTM. It would be nice to have data flow tests for this as well, e.g. in `ql/test/library-tests/dataflow/fields/`.

michaelnebel:
> LGTM. It would be nice to have data flow tests for this as well, e.g. in `ql/test/library-tests/dataflow/fields/`.

Yes, that is a good idea, but I don't think that pattern matching is really supported in the dataflow library.
As an example, the following code results in a failure of flow not being detected by dataflow library.
```csharp
public struct Struct
{
    public object Field;
    public object Prop { get; init; }
    public Struct(object field, object prop) => (Field, Prop) = (field, prop);
}

private void M1()
{
    var o = Source<object>(1);
    var s = new Struct(o, o);
    if (s is Struct { Field: object p })
    {
        Sink(p); // $ hasValueFlow=1
    }
}
```

Maybe we should create a seperate issue, where we investigate this?

michaelnebel:
https://github.com/github/codeql-csharp-team/issues/220

hvitved (approved):

PR #7721, opened 2022-01-24
JS: add a js/samesite-none-cookie cookie

erik-krogh:
[MDN Web Docs: SameSite](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Set-Cookie/SameSite). 

I've set the `@precision` to `medium`, because `SameSite=None` is only an issue if there isn't any other CSRF protection.  
And the query makes no attempt to detect the existence of CSRF protection. 

The default value for `SameSite` is `Lax`. 
The `Lax` setting is not great, but also not terrible, so I don't want to flag that. 
(Also: The `SameSite` attribute is somewhat recent. E.g. there is no IE version that supports it). 

[Evaluation shows no new results, and a constant slow-down caused by having to compile the new query](https://github.com/github/codeql-dca-main/tree/data/erik-krogh/CWE-1275__default__CustomSuite/reports).  
/cc @esbena: This is at least the second time I've seen an evaluation where a new query is compiled by `codeql-action/analyze` because it didn't hit anything in the cache.

### DIFF
--- a/javascript/ql/lib/semmle/javascript/frameworks/CookieLibraries.qll
+++ b/javascript/ql/lib/semmle/javascript/frameworks/CookieLibraries.qll
@@ -110,6 +110,23 @@ private string getCookieValue(string s, string attribute) {
 }
 
 /**
+ * Gets the "SameSite" value for a given `node`.
+ * Converts boolean values to the corresponding string value.
+ *
+ * Not all libraries support boolean values for the `SameSite` attribute,
+ * but here we assume that they do.
+ */
+private string getSameSiteValue(DataFlow::Node node) {
+  node.mayHaveStringValue(result)
+  or
+  node.mayHaveBooleanValue(true) and
+  result = "Strict"
+  or
+  node.mayHaveBooleanValue(false) and
+  result = "Lax"
+}
+
+/**
  * A model of the `js-cookie` library (https://github.com/js-cookie/js-cookie).
  */
 private module JsCookie {
@@ -150,7 +167,7 @@ private module JsCookie {
     override predicate isSensitive() { canHaveSensitiveCookie(this.getArgument(0)) }
 
     override string getSameSite() {
-      this.getOptionArgument(2, "sameSite").mayHaveStringValue(result)
+      result = getSameSiteValue(this.getOptionArgument(2, "sameSite"))
     }
   }
 }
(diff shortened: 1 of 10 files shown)


javascript/ql/lib/semmle/javascript/frameworks/CookieLibraries.qll
```diff
@@ -132,6 +148,10 @@ private module JsCookie {
     }
 
     override predicate isSensitive() { canHaveSensitiveCookie(this.getArgument(0)) }
+
+    override string getSameSite() {
+      this.getOptionArgument(2, "sameSite").mayHaveStringValue(result)
```
esbena, line 153: Should we introduce a shared (inline) predicate for this string/bool -> string conversion? I list the suggestion here because I do not see the harm in adding it for this smaller predicate as well.

esbena (commented):
LGTM, with one suggestino for some code sharing.

javascript/ql/src/change-notes/2022-01-24-samesite-cookie.md
```diff
@@ -0,0 +1,4 @@
+---
+category: newQuery
+---
+* A new query `js/samesite-none-cookie` has been added. The query detects when the SameSite attribute on a sensitive cookie is set to None.
```
mchammer01, line 4: Perhaps reads slightly better?

```suggestion
* A new query `js/samesite-none-cookie` has been added. The query detects when the SameSite attribute is set to None on a sensitive cookie.
```

javascript/ql/src/Security/CWE-1275/SameSiteNoneCookie.qhelp
```diff
@@ -0,0 +1,43 @@
+<!DOCTYPE qhelp PUBLIC
+  "-//Semmle//qhelp//EN"
+  "qhelp.dtd">
+<qhelp>
+
+<overview>
+<p>
+Authentication cookies where the SameSite attribute is set to "None" can 
+potentially be used to perform cross-site request forgery (CSRF) attacks
```
mchammer01, line 9: Same comment as for the ql file: 
```suggestion
potentially be used to perform Cross-Site Request Forgery (CSRF) attacks 
```

javascript/ql/src/Security/CWE-1275/SameSiteNoneCookie.qhelp
```diff
@@ -0,0 +1,43 @@
...
+<qhelp>
+
+<overview>
+<p>
+Authentication cookies where the SameSite attribute is set to "None" can 
+potentially be used to perform cross-site request forgery (CSRF) attacks 
+if no other CSRF protections are in place.
+</p>
+<p>
+With SameSite set to "None" a third party website may create an authorized cross-site request
```
mchammer01, line 13:
```suggestion
With SameSite set to "None", a third party website may create an authorized cross-site request
```

mchammer01 (dismissed):
@erik-krogh - LGTM from an editorial point of view  ✨ 
Made a few minor suggestions.

mchammer01 (approved):

esbena (approved):

PR #7723, opened 2022-01-24
Java: Add ReDoS queries

joefarebrother:
Ports the ReDoS and PolynomialReDoS queries from dynamic languages - primarily based off of the implementation for Python.

Todo:
- ~~Tests~~
- ~~Documentation~~
- ~~Ensure displayed locations of regex terms are accurate~~
- ~~Look into the possibility of dertermining whether a regex is matched against a full sring or a substring; for increased accuracy~~


erik-krogh:
I'm loving it :heart: 

I can see you are using slightly outdated versions of the ReDoS libraries. 
You should set up [identical files](https://github.com/github/codeql/blob/main/config/identical-files.json#L470-L483).  

Something that that can help debug the regexp parser is to implement PrintAST for regexp literals.  
Python has PrintAST support for regexp literals, so that shouldn't be too hard to copy.  

And you should definitely have a lot of tests to ensure the syntax is correctly parsed (and check that the regexp is actually vulnerable).   

I see you have a `Implement remaining syntax differences` commit, how did you determine that you got all the differences?  
(I'm just asking because there exists surprisingly subtle differences between regexp dialects. E.g. I recently learned that some regexp dialects have atomic groups). 

I don't have much experience with creating a regexp parser (that's `@yoff` and `@nickrolfe`). 
So if you want comments on the parser, you should request a review from one of them.

java/ql/lib/semmle/code/java/regex/RegexTreeView.qll
```diff
@@ -0,0 +1,1031 @@
...
+ * ```
+ */
+class RegExpCharacterClass extends RegExpTerm, TRegExpCharacterClass {
+  RegExpCharacterClass() { this = TRegExpCharacterClass(re, start, end) }
+
+  /** Holds if this character class is inverted, matching the opposite of its content. */
+  predicate isInverted() { re.getChar(start + 1) = "^" }
+
+  /** Gets the `i`th char inside this charater class. */
+  string getCharThing(int i) { result = re.getChar(i + start) }
```
smowton, line 582: Thing?

java/ql/lib/semmle/code/java/regex/RegexTreeView.qll
```diff
@@ -0,0 +1,1031 @@
...
+      cce1 != cce2 and cce1.toLowerCase() = cce2.toLowerCase()
+    )
+  }
+
+  override RegExpTerm getChild(int i) {
+    i = 0 and
+    result.getRegex() = re and
+    exists(int itemStart, int itemEnd |
+      result.getStart() = itemStart and
+      re.char_set_start(start, itemStart) and
```
smowton, line 604: snake case -> camel case everywhere

java/ql/lib/semmle/code/java/regex/RegexTreeView.qll
```diff
@@ -0,0 +1,1031 @@
...
+ * ```
+ */
+class RegExpNormalChar extends RegExpTerm, TRegExpNormalChar {
+  RegExpNormalChar() { this = TRegExpNormalChar(re, start, end) }
+
+  /**
+   * Holds if this constant represents a valid Unicode character (as opposed
+   * to a surrogate code point that does not correspond to a character by itself.)
+   */
+  predicate isCharacter() { any() }
```
smowton, line 676: Overridden anywhere, or...?

java/ql/lib/semmle/code/java/regex/RegexTreeView.qll
```diff
@@ -0,0 +1,1031 @@
...
+    result.getRegex() = re and
+    result.getStart() = upper_start and
+    result.getEnd() = end
+  }
+
+  override string getPrimaryQLClass() { result = "RegExpCharacterRange" }
+}
+
+/**
+ * A normal character in a regular expression, that is, a character
```
smowton, line 660: This doesn't seem to be true, since predefined character classes are being parsed as normal chars. Split them out into their own syntactic class?

java/ql/lib/semmle/code/java/regex/RegexTreeView.qll
```diff
@@ -0,0 +1,1031 @@
...
+ *
+ * ```
+ * a
+ * ```
+ */
+class RegExpConstant extends RegExpTerm {
+  string value;
+
+  RegExpConstant() {
+    (this = TRegExpNormalChar(re, start, end) or this = TRegExpQuote(re, start, end)) and
```
smowton, line 725: Redundant with the `getValue` constraints below?

java/ql/lib/semmle/code/java/regex/RegexTreeView.qll
```diff
@@ -0,0 +1,1031 @@
...
+    not this instanceof RegExpCharacterClassEscape and
+    // exclude chars in qualifiers
```
smowton, line 727: Clarify why? Even in the context `a*`, the subexpression `a` seems like a constant? We also don't check for other syntactic parents that would make this non-constant -- `a|b`, `a(nonconst)` etc

java/ql/lib/semmle/code/java/regex/RegexTreeView.qll
```diff
@@ -0,0 +1,1031 @@
...
+  RegExpSpecialChar() {
+    this = TRegExpSpecialChar(re, start, end) and
+    re.specialCharacter(start, end, char)
+  }
+
+  /**
+   * Holds if this constant represents a valid Unicode character (as opposed
+   * to a surrogate code point that does not correspond to a character by itself.)
+   */
+  predicate isCharacter() { any() }
```
smowton, line 811: Again, overridden anywhere?

java/ql/lib/semmle/code/java/regex/regex.qll
```diff
@@ -0,0 +1,943 @@
...
+private import RegexFlowConfigs
+
+/**
+ * A string literal that is used as a regular exprssion.
+ */
+abstract class RegexString extends Expr {
+  RegexString() { this instanceof StringLiteral }
+
+  /** Holds if a character set starts between `start` and `end`. */
+  private predicate char_set_start0(int start, int end) {
```
smowton, line 11: snake -> camel case everywhere

java/ql/lib/semmle/code/java/regex/regex.qll
```diff
@@ -0,0 +1,943 @@
...
+    this.nonEscapedCharAt(start) = "[" and
+    (
+      this.getChar(start + 1) = "^" and end = start + 2
```
smowton, line 14: use `if` expression

java/ql/lib/semmle/code/java/regex/regex.qll
```diff
@@ -0,0 +1,943 @@
...
+
+/**
+ * A string literal that is used as a regular exprssion.
+ */
+abstract class RegexString extends Expr {
+  RegexString() { this instanceof StringLiteral }
+
+  /** Holds if a character set starts between `start` and `end`. */
+  private predicate char_set_start0(int start, int end) {
+    this.nonEscapedCharAt(start) = "[" and
```
smowton, line 12: Can `[` itself occur in a character class? `[[]` for example?

java/ql/lib/semmle/code/java/regex/regex.qll
```diff
@@ -0,0 +1,943 @@
...
+      or
+      not this.getChar(start + 1) = "^" and end = start + 1
+    )
+  }
+
+  /** Holds if the character at `pos` marks the end of a character class. */
+  private predicate char_set_end0(int pos) {
+    this.nonEscapedCharAt(pos) = "]" and
+    /* special case: `[]]` and `[^]]` are valid char classes.  */
+    not this.char_set_start0(_, pos - 1)
```
smowton, line 24: With specifically `pos - 1` does this account for `[^]]`?

java/ql/lib/semmle/code/java/regex/regex.qll
```diff
@@ -0,0 +1,943 @@
...
+  }
+
+  /**
+   * Gets the nesting depth of charcter classes at position `pos`
```
smowton, line 28:
```suggestion
   * Gets the nesting depth of character classes at position `pos`
```

java/ql/lib/semmle/code/java/regex/regex.qll
```diff
@@ -0,0 +1,943 @@
...
+   */
+  private int char_set_depth(int pos) {
+    exists(this.getChar(pos)) and
+    result =
+      max(int j |
+        j = 0 or
+        j =
+          count(int i | i < pos and this.char_set_start0(i, _)) -
```
smowton, line 36: This is a question for people with better QL wizardry than me, but I think this will build intermediate expressions quadratic in the number of `[` or `]` characters in a string, which could be a problem if we run against a database with a very large string in it.

If that is a problem, a recursive version could work nicely, counting and maintaining state as it traverses a string from left to right.

java/ql/lib/semmle/code/java/regex/regex.qll
```diff
@@ -0,0 +1,943 @@
...
+  }
+
+  /** Holds if a top-level character set ends at `pos`. */
+  predicate char_set_end(int pos) {
+    this.char_set_end0(pos) and
+    this.char_set_depth(pos) = 1
+  }
+
+  /**
+   * Whether there is a top-level character class, between start (inclusive) and end (exclusive)
```
smowton, line 54:
```suggestion
   * Holds if there is a top-level character class beginning at `start` (inclusive) and ending at `end` (exclusive)
```

java/ql/lib/semmle/code/java/regex/regex.qll
```diff
@@ -0,0 +1,943 @@
...
+   *
+   * For now, nested character classes are approximated by only considering the top-level class for parsing.
+   * This leads to very similar results for ReDoS queries.
+   */
+  predicate charSet(int start, int end) {
```
smowton, line 59:
```suggestion
  predicate isTopLevelCharacterClass(int start, int end) {
```

java/ql/lib/semmle/code/java/regex/regex.qll
```diff
@@ -0,0 +1,943 @@
+import java
+private import RegexFlowConfigs
+
+/**
+ * A string literal that is used as a regular exprssion.
+ */
+abstract class RegexString extends Expr {
+  RegexString() { this instanceof StringLiteral }
```
smowton, line 8: Derived from `StringLiteral` not `Expr`?

java/ql/lib/semmle/code/java/regex/regex.qll
```diff
@@ -0,0 +1,943 @@
...
+
+  /** Holds if a character set starts between `start` and `end`. */
```
smowton, line 10:
```suggestion
  /** Holds if a character set starts between `start` and `end`, including any negation character. */
```

java/ql/lib/semmle/code/java/regex/regex.qll
```diff
@@ -0,0 +1,943 @@
...
+   */
+  predicate charSet(int start, int end) {
+    exists(int inner_start, int inner_end |
+      this.char_set_start(start, inner_start) and
+      not this.char_set_start(_, start)
+    |
+      end = inner_end + 1 and
+      inner_end > inner_start and
+      this.char_set_end(inner_end) and
+      not exists(int mid | this.char_set_end(mid) | mid > inner_start and mid < inner_end)
```
smowton, line 67: Would this be better expressed as a regex match of `[^]]+` or similar?

smowton (commented):
Hopefully some helpful style comments, but on the substance of `regex.qll` I'll have to request a review from someone familiar with writing parsers in QL

java/ql/lib/semmle/code/java/regex/RegexTreeView.qll
```diff
@@ -0,0 +1,1031 @@
...
+    this.getUnescaped() = "n" and result = "\n"
+    or
+    this.getUnescaped() = "r" and result = "\r"
+    or
+    this.getUnescaped() = "t" and result = "\t"
+    or
+    // TODO: Find a way to include a formfeed character
+    // also the alert/bell character for \a and escape character for \e.
+    // this.getUnescaped() = "f" and result = "
"
```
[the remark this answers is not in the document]
erik-krogh: Note: 
In QL you can use `11.toUnicode()` to get \v and `12.toUnicode()` to get \f.

smowton:
@yoff @nickrolfe I tried reviewing this just now, but particularly concerning `regex.qll` which parses nested character classes and such, I don't know enough about writing text parsers in QL to comment meaningfully. If either of you are able to comment I'd be very grateful.

java/ql/lib/semmle/code/java/regex/RegexTreeView.qll
```diff
@@ -0,0 +1,1031 @@
...
+  TRegExpSequence(Regex re, int start, int end) {
+    re.sequence(start, end) and
+    exists(seqChild(re, start, end, 1)) // if a sequence does not have more than one element, it should be treated as that element instead.
+  } or
+  /** An alternation term */
+  TRegExpAlt(Regex re, int start, int end) {
+    re.alternation(start, end) and
+    exists(int part_end |
+      re.alternationOption(start, end, start, part_end) and
+      part_end < end
```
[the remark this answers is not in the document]
joefarebrother: This tests that there *exists* an alternation part that ends strictly before the end of the alternation itself; i.e. it has more than one element.

java/ql/lib/semmle/code/java/regex/RegexTreeView.qll
```diff
@@ -0,0 +1,1031 @@
...
+    this.isRootTerm() and result = this
+    or
+    result = this.getParent().(RegExpTerm).getRootTerm()
+  }
+
+  /**
+   * Holds if this term is part of a string literal
+   * that is interpreted as a regular expression.
+   */
+  predicate isUsedAsRegExp() { any() }
```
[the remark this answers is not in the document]
joefarebrother: It's used by the shared redos library (which is presumably the case for most of these trivial predicates)

java/ql/lib/semmle/code/java/regex/RegexTreeView.qll
```diff
@@ -0,0 +1,1031 @@
...
+  /** Gets the string literal in which this term is found. */
+  RegExpLiteral getLiteral() { result = TRegExpLiteral(re) }
+
+  /** Gets the regular expression term that is matched (textually) before this one, if any. */
+  RegExpTerm getPredecessor() {
+    exists(RegExpTerm parent | parent = this.getParent() |
+      result = parent.(RegExpSequence).previousElement(this)
+      or
+      not exists(parent.(RegExpSequence).previousElement(this)) and
+      not parent instanceof RegExpSubPattern and
```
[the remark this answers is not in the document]
joefarebrother: A RegExpSubPattern is a lookahead/lookbehind assertion; which doesn't really make sense to have a predecessor or successor.

java/ql/lib/semmle/code/java/regex/RegexTreeView.qll
```diff
@@ -0,0 +1,1031 @@
...
+ * A star-quantified term.
+ *
+ * Example:
+ *
+ * ```
+ * \w*
+ * ```
+ */
+class RegExpStar extends InfiniteRepetitionQuantifier {
+  RegExpStar() { this.getQualifier().charAt(0) = "*" }
```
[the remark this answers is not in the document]
joefarebrother: Indeed; should inclde `*`, `*?`, and `*+`.

java/ql/lib/semmle/code/java/regex/RegexTreeView.qll
joefarebrother: It looks like it actually exludes the `*` in `a*` from being considered a constant

java/ql/lib/semmle/code/java/regex/regex.qll
joefarebrother: No; in `[[ ]` the `[ ]` would be considred as a nested class, making the outer class unfinished and thus a parse error.

java/ql/lib/semmle/code/java/regex/regex.qll
joefarebrother: Should do; since `char_set_start0` accounts for `[^`

java/ql/lib/semmle/code/java/regex/regex.qll
joefarebrother: I think that would require taking a substring, which is something that it looks like this library is trying to avoid except at the top-level by mostly handling indicies. Doing so could cause quadratically many substrings to be constructed.

java/ql/lib/semmle/code/java/regex/RegexFlowConfigs.qll
```diff
@@ -0,0 +1,192 @@
...
+  override predicate isSink(DataFlow::Node node) { sinkNode(node, "regex-compile") }
+}
+
+/**
+ * Holds if `s` is used as a regex, with the mode `mode` (if known).
+ * If regex mode is not known, `mode` will be `"None"`.
+ */
+predicate used_as_regex(Expr s, string mode) {
+  any(RegexCompileFlowConf c).hasFlow(DataFlow2::exprNode(s), _) and
+  mode = "None" // TODO: proper mode detection
```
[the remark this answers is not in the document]
joefarebrother: The optional flags that can be passed to methods like `Pattern.compile`.

java/ql/lib/semmle/code/java/regex/regex.qll
joefarebrother: (additionally, that regex seems less clear to me what it's doing than this ql does. It also qould fail on escaped `]`s in the class, or nested classes)

java/ql/lib/semmle/code/java/regex/RegexTreeView.qll
```diff
@@ -0,0 +1,1031 @@
...
+private import semmle.code.java.regex.regex
+
+/**
+ * An element containing a regular expression term, that is, either
+ * a string literal (parsed as a regular expression)
+ * or another regular expression term.
+ *
+ * For sequences and alternations, we require at least one child.
+ * Otherwise, we wish to represent the term differently.
+ * This avoids multiple representations of the same term.
```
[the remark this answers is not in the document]
joefarebrother: I believe it should indeed say two (and so should the python version of this file)

java/ql/lib/semmle/code/java/regex/regex.qll
yo-h: The [QL style guide](https://github.com/github/codeql/blob/main/docs/ql-style-guide.md) specifies camel case for predicate names.

java/ql/lib/semmle/code/java/regex/regex.qll
yo-h: @aschackmull?

java/ql/lib/semmle/code/java/regex/RegexFlowConfigs.qll
```diff
@@ -0,0 +1,192 @@
...
+  override predicate isSource(DataFlow::Node node) { node.asExpr() instanceof StringLiteral }
+
+  override predicate isSink(DataFlow::Node node) { sinkNode(node, "regex-compile") }
+}
+
+/**
+ * Holds if `s` is used as a regex, with the mode `mode` (if known).
+ * If regex mode is not known, `mode` will be `"None"`.
+ */
+predicate used_as_regex(Expr s, string mode) {
```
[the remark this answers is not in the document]
aschackmull: `s` is guaranteed to be a `StringLiteral`, so might as well use the more precise type, right?

java/ql/src/Security/CWE/CWE-730/PolynomialReDoS.ql
```diff
@@ -0,0 +1,48 @@
...
+
+class PolynomialRedosSink extends DataFlow::Node {
+  RegExpLiteral reg;
+
+  PolynomialRedosSink() { regex_match(reg.getRegex(), this.asExpr()) }
+
+  RegExpTerm getRegExp() { result = reg }
+}
+
+class PolynomialRedosConfig extends DataFlow::Configuration {
```
aschackmull, line 30: Should this be `TaintTracking`? Most `RemoteFlowSource`s aren't necessarily strings, so without some additional taint steps then this might not catch much.

java/ql/lib/semmle/code/java/security/performance/RegExpTreeView.qll
```diff
@@ -0,0 +1,57 @@
+/**
+ * This module should provide a class hierarchy corresponding to a parse tree of regular expressions.
```
aschackmull, line 2: "should"? I guess hopefully it does. This is the interface to the shared redos libs, right?, so it should probably say so.

java/ql/lib/semmle/code/java/security/performance/RegExpTreeView.qll
```diff
@@ -0,0 +1,57 @@
...
+ */
+
+import java
+import semmle.code.java.regex.RegexTreeView
```
aschackmull, line 6: It is quite confusing to have both a `RegexTreeView.qll` and a `RegExpTreeView.qll`. Could we improve on that? It took me a few look-throughs to figure out that they were two different files.

java/ql/lib/semmle/code/java/security/performance/RegExpTreeView.qll
```diff
@@ -0,0 +1,57 @@
+/**
+ * This module should provide a class hierarchy corresponding to a parse tree of regular expressions.
+ */
+
+import java
```
aschackmull, line 5: Make the imports private? Presumably since this implements the shared lib interface we shouldn't expose anything language-specific.

java/ql/lib/semmle/code/java/security/performance/RegExpTreeView.qll
```diff
@@ -0,0 +1,57 @@
...
+import semmle.code.java.regex.RegexTreeView
+
+/**
+ * Holds if `term` is an ecape class representing e.g. `\d`.
```
aschackmull, line 9:
```suggestion
 * Holds if `term` is an escape class representing e.g. `\d`.
```

java/ql/lib/semmle/code/java/security/performance/RegExpTreeView.qll
```diff
@@ -0,0 +1,57 @@
...
+ * `clazz` is which character class it represents, e.g. "d" for `\d`.
+ */
+predicate isEscapeClass(RegExpTerm term, string clazz) {
+  exists(RegExpCharacterClassEscape escape | term = escape | escape.getValue() = clazz)
```
aschackmull, line 13:
```suggestion
  term.(RegExpCharacterClassEscape).getValue() = clazz
```

java/ql/lib/semmle/code/java/security/performance/RegExpTreeView.qll
```diff
@@ -0,0 +1,57 @@
...
+}
+
+/**
+ * Holds if the regular expression should not be considered.
+ *
+ * We make the pragmatic performance optimization to ignore regular expressions in files
+ * that does not belong to the project code (such as installed dependencies).
```
aschackmull, line 20:
```suggestion
 * that do not belong to the project code (such as installed dependencies).
```

java/ql/lib/semmle/code/java/security/performance/RegExpTreeView.qll
```diff
@@ -0,0 +1,57 @@
...
+ */
+predicate isExcluded(RegExpParent parent) {
+  not exists(parent.getRegex().getLocation().getFile().getRelativePath())
+  or
+  // Regexes with many occurrences of ".*" may cause the polynomial ReDoS computation to explode, so
+  // we explicitly exclude these.
+  count(int i | exists(parent.getRegex().getText().regexpFind("\\.\\*", i, _)) | i) > 10
```
aschackmull, line 27: Equivalent and skips generating the empty case.
```suggestion
  strictcount(int i | exists(parent.getRegex().getText().regexpFind("\\.\\*", i, _)) | i) > 10
```

java/ql/lib/semmle/code/java/regex/regex.qll
```diff
@@ -0,0 +1,943 @@
+import java
+private import RegexFlowConfigs
+
+/**
+ * A string literal that is used as a regular exprssion.
```
aschackmull, line 5:
```suggestion
 * A string literal that is used as a regular expression.
```

java/ql/lib/semmle/code/java/regex/regex.qll
```diff
@@ -0,0 +1,943 @@
...
+    not exists(int x, int y | this.charSet(x, y) and x <= start and y >= end)
+    or
+    this.group(start, end)
+    or
+    this.charSet(start, end)
+    or
+    this.backreference(start, end)
+  }
+
+  private predicate qualifier(int start, int end, boolean maybe_empty, boolean may_repeat_forever) {
```
aschackmull, line 665: s/qualifier/quantifier/ everywhere?

java/ql/lib/semmle/code/java/regex/RegexTreeView.qll
```diff
@@ -0,0 +1,1031 @@
...
+  Location getLocation() { result = re.getLocation() }
+
+  /** Holds if this term is found at the specified location offsets. */
+  predicate hasLocationInfo(
+    string filepath, int startline, int startcolumn, int endline, int endcolumn
+  ) {
+    exists(int re_start, int re_end |
+      re.getLocation().hasLocationInfo(filepath, startline, re_start, endline, re_end) and
+      startcolumn = re_start + start + 1 and
+      endcolumn = re_start + end
```
aschackmull, line 193: This only appears to work for the most simple cases. But covering everything is likely going to be too much work for very little benefit. Are there some common cases that it might be worth covering as well? E.g. backslashes? At the very least I think the QL code here should have a comment stating that this is a simple best-effort approach and sketch some of the things that aren't covered.
To elaborate on why this fails in general: Our offsets are necessarily in terms of the string _value_, but the location offsets are in terms of the _literal_ in the code. The most basic difference is escape sequences, a value `\` usually has the literal code representation `\\`, there can also be unicode escape sequences, and the string can have compile-time concatenation (`"a" + "b"` is a string literal with the value `ab`). String literals with compile-time concatenation can also be spread across multiple lines, which further throws off the column offsets. And then to complicate things further a string literal can also be a Java 15 text block, which has a whole set of additional rules.

java/ql/lib/semmle/code/java/regex/RegexTreeView.qll
```diff
@@ -0,0 +1,1031 @@
...
+    )
+  }
+
+  /** Gets the file in which this term is found. */
+  File getFile() { result = this.getLocation().getFile() }
+
+  /** Gets the raw source text of this term. */
```
aschackmull, line 200: That qldoc is not entirely accurate. In fact, I don't think there's any way we can reliably produce "the raw source text of this term" (barring a herculean re-parsing of Java strings, which we don't want to do), nor do I think we want to. We should probably dig a bit into why and for what the shared lib uses this predicate.

java/ql/lib/semmle/code/java/regex/RegexTreeView.qll
```diff
@@ -0,0 +1,1031 @@
...
+  /**
+   * Holds if this term is part of a string literal
+   * that is interpreted as a regular expression.
+   */
+  predicate isUsedAsRegExp() { any() }
+
+  /**
+   * Holds if this is the root term of a regular expression.
+   */
+  predicate isRootTerm() { start = 0 and end = re.getText().length() }
```
aschackmull, line 141: This indicates to me that `start` is inclusive and `end` is exclusive`. Presumably that's the convention that's followed for all these start-end-ranges? Could we document this?

java/ql/lib/semmle/code/java/regex/RegexTreeView.qll
```diff
@@ -0,0 +1,1031 @@
...
+  File getFile() { result = this.getLocation().getFile() }
+
+  /** Gets the raw source text of this term. */
+  string getRawValue() { result = this.toString() }
+
+  /** Gets the string literal in which this term is found. */
+  RegExpLiteral getLiteral() { result = TRegExpLiteral(re) }
+
+  /** Gets the regular expression term that is matched (textually) before this one, if any. */
+  RegExpTerm getPredecessor() {
```
aschackmull, line 207: Why is this predicate not in the shared library? It seems language-independent to me.

java/ql/lib/semmle/code/java/regex/RegexTreeView.qll
joefarebrother: I'm aware of this issue; I'm working on handling the most common cases (i.e. backslashes).

java/ql/lib/semmle/code/java/regex/RegexTreeView.qll
```diff
@@ -0,0 +1,1031 @@
...
+      result = parent.(RegExpSequence).previousElement(this)
+      or
+      not exists(parent.(RegExpSequence).previousElement(this)) and
+      not parent instanceof RegExpSubPattern and
+      result = parent.getPredecessor()
+    )
+  }
+
+  /** Gets the regular expression term that is matched (textually) after this one, if any. */
+  RegExpTerm getSuccessor() {
```
aschackmull, line 218: Is `r1.getSuccessor() = r2` equivalent to `r1 = r2.getPredecessor()`? If so, then there's no need for a duplicate implementation. (And if it's somehow not then it needs some more qldoc!)

java/ql/lib/semmle/code/java/regex/RegexTreeView.qll
```diff
@@ -0,0 +1,1031 @@
...
+  string getFlags() { result = concat(string mode | mode = re.getAMode() | mode, " | ") }
+
+  override Regex getRegex() { result = re }
+
+  /** Gets the primary QL class for this regex. */
+  string getPrimaryQLClass() { result = "RegExpLiteral" }
+}
+
+/**
+ * A regular expression term, that is, a syntactic part of a regular expression.
```
aschackmull, line 94:
```suggestion
 * A regular expression term, that is, a syntactic part of a regular expression.
 * These are the tree nodes that form the parse tree of a regular expression literal.
```
Hopefully that's correctly understood?

java/ql/lib/semmle/code/java/regex/RegexTreeView.qll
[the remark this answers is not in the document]
aschackmull: I think `.matches` ought to perform better in principle than `.charAt(0) = "X"`. It should perform fewer string pool operations.
```suggestion
  RegExpStar() { this.getQualifier().matches("*%") }
```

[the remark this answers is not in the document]
aschackmull: I guess a rewrite like this might also be a job for the optimiser, but take a look at the RA and see for yourself.

java/ql/lib/semmle/code/java/regex/RegexTreeView.qll
```diff
@@ -0,0 +1,1031 @@
...
+ * \w{2}
+ * ```
+ */
+class RegExpRange extends RegExpQuantifier {
+  string upper;
+  string lower;
+
+  RegExpRange() { re.multiples(part_end, end, lower, upper) }
+
+  /** Gets the string defining the upper bound of this range, if any. */
```
aschackmull, line 336: There's always a result, so drop "if any"?
```suggestion
  /** Gets the string defining the upper bound of this range. */
```

java/ql/lib/semmle/code/java/regex/RegexTreeView.qll
joefarebrother: It looks like the shared library is using it to choose a canonical representation for all terms that share the same string representation.

java/ql/lib/semmle/code/java/regex/RegexTreeView.qll
```diff
@@ -0,0 +1,1031 @@
...
+    )
+    or
+    i > 0 and
+    result.getRegex() = re and
+    exists(int part_start |
+      part_start = this.getChild(i - 1).getEnd() + 1 // allow for the |
+    |
+      result.getStart() = part_start and
+      re.alternationOption(start, end, part_start, result.getEnd())
+    )
```
aschackmull, line 441: Possibly replace the recursion here with `rank`?

java/ql/lib/semmle/code/java/regex/RegexTreeView.qll
```diff
@@ -0,0 +1,1031 @@
...
+  }
+}
+
+/**
+ * Gets the hex number for the `hex` char.
+ */
+private int toHex(string hex) {
+  hex = [0 .. 9].toString() and
+  result = hex.toInt()
```
aschackmull, line 528:
```suggestion
  result = [0 .. 9] and hex = result.toString()
```

java/ql/lib/semmle/code/java/regex/RegexTreeView.qll
joefarebrother: Doesn't look like this is used anywhere anyway; so removed

java/ql/lib/semmle/code/java/security/performance/RegExpTreeView.qll
joefarebrother: "should" in the sense of "it needs to" rather than "I think it does"

java/ql/lib/semmle/code/java/regex/regex.qll
```diff
@@ -0,0 +1,933 @@
...
+        this.isOctal(end) and
+        (end = start + 4 implies this.getChar(start + 2) <= "3")
+      )
+      or
+      // 16-bit hex value \uhhhh
+      this.getChar(start + 1) = "u" and end = start + 6
+      or
+      this.escapedBraces(start, end)
+      or
+      // Boundry matchers \b, \b{g}
```
nickrolfe, line 304: I notice you're not handling the other forms of boundary matchers I see in [the docs](https://docs.oracle.com/javase/7/docs/api/java/util/regex/Pattern.html), such as `\A`, `\Z`, and `\G`.

We missed those originally in Ruby as well, and they caused weird messages from the ReDoS query alerts, where it thought for example that `\G` was just an escape of the character `G`, so it suggested the regex `\G(a*)*b` would backtrack exponentially for strings of the form `Gaaaaaa...`. I fixed that in https://github.com/github/codeql/commit/f63c768d9f48db8d57604f290dc442876b6812ff.

nickrolfe (commented):
I haven't reviewed the entire parser line-by-line, but I've commented on some things that stood out. The main thing I'd say is that the parser is pretty tedious to work on, to understand, or to review, so I think it's really worth getting good test coverage.

java/ql/test/library-tests/regex/Test.java
```diff
@@ -0,0 +1,23 @@
...
+        "\\Q hello world [ *** \\Q ) ( \\E",
+        "[\\Q hi ] \\E]",
+        "[]]",
+        "[^]]",
+        "[abc[defg]]",
+        "[abc&&[\\W\\p{Lower}\\P{Space}\\N{degree sign}]]\\b7\\b{g}8",
+        "\\cA",
+        "\\c(",
+        "\\c\\(ab)",
+        "(?>hi)(?<name>hell*?o*+)123\\k<name>"
```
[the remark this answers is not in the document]
erik-krogh: Tests for `X{n, m`, `X{n, `, `X{n`, `X{` could also be a good idea. 
(Those were parsed wrong in the JS parser a while ago).

java/ql/lib/semmle/code/java/regex/RegexTreeView.qll
```diff
@@ -0,0 +1,1031 @@
...
+  result = 13 and hex = ["d", "D"]
+  or
+  result = 14 and hex = ["e", "E"]
+  or
+  result = 15 and hex = ["f", "F"]
+}
+
+/**
+ * A character class escape in a regular expression.
+ * That is, an escaped charachter that denotes multiple characters.
```
aschackmull, line 548:
```suggestion
 * That is, an escaped character that denotes multiple characters.
```

java/ql/lib/semmle/code/java/regex/RegexTreeView.qll
```diff
@@ -0,0 +1,1031 @@
...
+      re.charSetChild(start, itemStart, itemEnd) and
+      result.getEnd() = itemEnd
+    )
+    or
+    i > 0 and
+    result.getRegex() = re and
+    exists(int itemStart | itemStart = this.getChild(i - 1).getEnd() |
+      result.getStart() = itemStart and
+      re.charSetChild(start, itemStart, result.getEnd())
+    )
```
aschackmull, line 614: Another `getChild` recursion that could profitably be rewritten to use `rank` instead of recursion.

java/ql/lib/semmle/code/java/regex/RegexTreeView.qll
aschackmull: Nevermind, found the comment at the top of `regex.qll`.

java/ql/lib/semmle/code/java/regex/regex.qll
```diff
@@ -0,0 +1,933 @@
+import java
```
aschackmull, line 1: The ordering of the member predicates of `RegexString` seems a bit arbitrary. Could we topologically sort them in dependency order? I think that would make the code easier to read and any potential recursion would be easier to recognise.

java/ql/lib/semmle/code/java/regex/regex.qll
```diff
@@ -0,0 +1,933 @@
...
+  }
+
+  /**
+   * Helper predicate for `quoteSequence`.
+   * Holds if the char at `pos` is the one-based `index`th occourence of a quote delimiter (`\Q` or `\E`)
+   * Result is `true` for `\Q` and `false` for `\E`.
+   */
+  private boolean quoteDelimiter(int index, int pos) {
+    result = this.quoteDelimiter(pos) and
+    pos = rank[index](int p | this.quoteDelimiter(p) = [true, false])
```
aschackmull, line 202:
```suggestion
    pos = rank[index](int p | exists(this.quoteDelimiter(p)))
```

java/ql/lib/semmle/code/java/regex/regex.qll
```diff
@@ -0,0 +1,933 @@
...
+  }
+
+  /** Holds if a quoted sequence is found between `start` and `end` */
+  predicate quote(int start, int end) { this.quote(start, end, _, _) }
+
+  /** Holds if a quoted sequence is fund between `start` and `end`, with ontent found between `inner_start` and `inner_end`. */
```
aschackmull, line 208:
```suggestion
  /** Holds if a quoted sequence is found between `start` and `end`, with content found between `inner_start` and `inner_end`. */
```

java/ql/lib/semmle/code/java/regex/regex.qll
```diff
@@ -0,0 +1,933 @@
...
+  /** Holds if a character set starts between `start` and `end`, including any negation character (`^`). */
+  private predicate charSetStart0(int start, int end) {
+    this.nonEscapedCharAt(start) = "[" and
+    (if this.getChar(start + 1) = "^" then end = start + 2 else end = start + 1)
+  }
+
+  /** Holds if the character at `pos` marks the end of a character class. */
+  private predicate charSetEnd0(int pos) {
+    this.nonEscapedCharAt(pos) = "]" and
+    /* special case: `[]]` and `[^]]` are valid char classes.  */
```
aschackmull, line 18: Are both `[^]` and `[^]]` valid char classes? Are we handling both correctly?

java/ql/lib/semmle/code/java/regex/regex.qll
```diff
@@ -0,0 +1,933 @@
...
+   */
+  private int charSetDepth(int pos) {
+    exists(this.getChar(pos)) and
+    result =
+      max(int j |
+        j = 0 or
+        j =
+          count(int i | i < pos and this.charSetStart0(i, _)) -
+            count(int i | i < pos and this.charSetEnd0(i))
+      )
```
aschackmull, line 33: This seems inefficient, and given how common character classes are, we should probably fix this. Replace by applying `rank` to charSet starts and ends, then do a linear recursion over those rank-indices to calculate the depth, since you only need the depth info at those positions anyway.

java/ql/lib/semmle/code/java/regex/regex.qll
```diff
@@ -0,0 +1,933 @@
+import java
+private import RegexFlowConfigs
+
+// In all ranges handled by this library, `start` is inclusive and `end` is exclusive.
+/**
+ * A string literal that is used as a regular expression.
+ */
+abstract class RegexString extends StringLiteral {
+  /** Holds if a character set starts between `start` and `end`, including any negation character (`^`). */
```
aschackmull, line 9: Terminology should be fixed throughout: "character class" vs. "character set". I believe the former is the usual name.

java/ql/lib/semmle/code/java/regex/regex.qll
```diff
@@ -0,0 +1,933 @@
...
+  /**
+   * Holds if there is a top-level character class beginning at `start` (inclusive) and ending at `end` (exclusive)
+   *
+   * For now, nested character classes are approximated by only considering the top-level class for parsing.
+   * This leads to very similar results for ReDoS queries.
+   */
+  predicate charSet(int start, int end) {
+    exists(int inner_start, int inner_end |
+      this.charSetStart(start, inner_start) and
+      not this.charSetStart(_, start)
```
aschackmull, line 57: This is superfluous, right? `charSetStart` already restricts depth to 0.

java/ql/lib/semmle/code/java/regex/regex.qll
```diff
@@ -0,0 +1,933 @@
...
+    |
+      end = inner_end + 1 and
+      inner_end > inner_start and
+      this.charSetEnd(inner_end) and
+      not exists(int mid | this.charSetEnd(mid) | mid > inner_start and mid < inner_end)
```
aschackmull, line 62: Simplify: replace with `inner_end = min(...)`.

java/ql/lib/semmle/code/java/regex/RegexTreeView.qll
joefarebrother: I'm not sure that would work; a `rank` looking for `alternationOption`s in between `start` and `end` would find those in groups as well - e.g. in `ab(c|d)e|f`, `c` would be considerd a child of the top level term.

java/ql/lib/semmle/code/java/regex/regex.qll
aschackmull: The `min` version will likely be more efficient, but still quadratic in the number of top-level char classes. We can make it linear by keeping track of the depth=0 opening bracket in the rank-index traversal we need to do to calculate depth, and then immediately match up the top-level opening and closing brackets.

java/ql/lib/semmle/code/java/regex/regex.qll
joefarebrother: `[^]` isn't a valid class.

java/ql/lib/semmle/code/java/regex/regex.qll
```diff
@@ -0,0 +1,933 @@
...
+      // hex value \xhh
+      this.getChar(start + 1) = "x" and
+      this.getChar(start + 2) != "{" and
+      end = start + 4
+      or
+      // octal value \0o, \0oo, or \0ooo. Max of 0377.
+      this.getChar(start + 1) = "0" and
+      end in [start + 3 .. start + 5] and
+      forall(int i | i in [start + 1 .. end - 1] | this.isOctal(i)) and
+      (end = start + 5 implies this.getChar(start + 2) <= "3") and
```
aschackmull, line 292:
```suggestion
      (end = start + 5 implies this.getChar(start + 2) = ["0", "1", "2", "3"]) and
```

java/ql/test/library-tests/regex/Test.java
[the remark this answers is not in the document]
joefarebrother: Those cases should be parse failures.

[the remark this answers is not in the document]
erik-krogh: Those examples don't have any parse errors in JS, but yeah, they might be parse-errors in other regex dialects.

java/ql/lib/semmle/code/java/regex/regex.qll
```diff
@@ -0,0 +1,933 @@
...
+      not (
+        end < start + 5 and
+        this.isOctal(end) and
+        (end = start + 4 implies this.getChar(start + 2) <= "3")
+      )
```
aschackmull, line 297: This seems a bit complicated, how about just using some if-then-else to enumerate the 3 cases.
```suggestion
      // octal value \0o, \0oo, or \0ooo. Max of 0377.
      this.getChar(start + 1) = "0" and
      this.isOctal(start + 2) and
      if this.isOctal(start + 3)
      then
        if this.isOctal(start + 4) and this.getChar(start + 2) in ["0", "1", "2", "3"]
        then
          end = start + 5
        else
          end = start + 4
      else
        end = start + 3
```

java/ql/lib/semmle/code/java/regex/regex.qll
```diff
@@ -0,0 +1,933 @@
...
+   * but excludes backreferences.
+   */
+  predicate escapedCharacter(int start, int end) {
+    this.escapingChar(start) and
+    not this.backreference(start, _) and
+    (
+      // hex value \xhh
+      this.getChar(start + 1) = "x" and
+      this.getChar(start + 2) != "{" and
+      end = start + 4
```
aschackmull, line 286: Should the individual hex chars be checked?

java/ql/lib/semmle/code/java/regex/regex.qll
```diff
@@ -0,0 +1,933 @@
...
+      forall(int i | i in [start + 1 .. end - 1] | this.isOctal(i)) and
+      (end = start + 5 implies this.getChar(start + 2) <= "3") and
+      not (
+        end < start + 5 and
+        this.isOctal(end) and
+        (end = start + 4 implies this.getChar(start + 2) <= "3")
+      )
+      or
+      // 16-bit hex value \uhhhh
+      this.getChar(start + 1) = "u" and end = start + 6
```
aschackmull, line 300: Should the individual hex chars be checked?

java/ql/lib/semmle/code/java/regex/regex.qll
aschackmull: `InputSymbol.toString` in ReDoSUtil.qll indicates that it is, but that might be language-dependent? I would hope that a thing like `[^]` had mostly engine-independent semantics, but that might be too much to hope for?

erik-krogh: It's a valid class in JS, in some other dialects it's not.  
(I'm not sure about Java)

joefarebrother: It is in some regex dialects, but not in Java.

java/ql/lib/semmle/code/java/regex/RegexTreeView.qll
joefarebrother: In fact, its purpose was to exclude characters in repitition quantifiers (`a{3}`) from being considered constants. It has now ben moved into the RegexString class.

java/ql/lib/semmle/code/java/regex/regex.qll
joefarebrother: I'm not sure that would be worth it - doing so would only help to reject invalid regexes (which we don't expect there to be in real databases); and wouldn't help accept more valid regexes.

java/ql/lib/semmle/code/java/regex/RegexTreeView.qll
joefarebrother: In fact it's not quite equivalent - In a pattern like `(a|b)(c|d)`, the succesor of the `(a|b)` term, as well as that of its children `a` and `b`, is the `(c|d)` term, but not also any of its children; whereas the predecessor of the `c` term is the `(a|b)` term.

joefarebrother:
All-projects runs:

 -   Exponential redos: https://lgtm.com/query/7282163139420838370/
 -   Polynomial redos: https://lgtm.com/query/5280828096307312393/

erik-krogh (dismissed):
JS changed :+1:  

Did you try out modeling the possessive quantifiers as a "zero-or-one" quantifier?

atorralba (commented):
The `LengthRestrictedMethod` heuristic looks plausible to me. Added a minor inline comment.

erik-krogh:
> 2 .... I might have missed something however and suggest running this past the original ReDoS authors (I can't recall who off the top of my head?)

That would be me. 
However, what you are looking at there is regexp parsing, and the original authors there is @yoff / @nickrolfe.  

I think what you are essentially asking is if there exists regular expressions with deeply nested sequences.   
(I haven't dealt much with the parsers, so I don't know much about those predicates, so I might be wrong).  

The answer is [yes in JS](https://lgtm.com/query/6978396536563460329/) and [also in Ruby](https://lgtm.com/query/7673761226801577715/).  
(Probably also in Python, but I only checked a single project).

joefarebrother:
It looks like it shoulsn't change behaviour

nickrolfe:
> @erik-krogh / @yoff / @nickrolfe could you confirm whether you think [smowton@9493d8a](https://github.com/smowton/codeql/commit/9493d8aa1071c36c74c5b964580a22d531a6b985) is a valid optimisation, or do I accidentally change behaviour there?

I don't know/remember that code well enough to be really confident, but:
- it looks plausible
- it doesn't break any of the Ruby regexp tests
- I ran a quick query on a few large Ruby projects to find `TRegExpSequence`s where the conjunct you deleted holds and the one you added doesn't, or vice-versa, and got no results.

smowton:
Thanks @nickrolfe that's an exceedingly useful test. @joefarebrother over to you to DCA etc, but I think these improvements ought to make a substantial dent in the performance penalty of this query

erik-krogh:
> `erik-krogh` / `yoff` / `nickrolfe` could you confirm whether you think [smowton@9493d8a](https://github.com/smowton/codeql/commit/9493d8aa1071c36c74c5b964580a22d531a6b985) is a valid optimisation, or do I accidentally change behaviour there?

I had misunderstood above, which is good.  
The change looks fine to me.

hvitved:
@joefarebrother : You should rebase your PR now that https://github.com/github/codeql/pull/8589 has been merged, to get the latest changes to the shared files.

erik-krogh (approved):
Shared ReDoS changes :+1: 

You should do a new evaluation now that the shared ReDoS libraries have been updated.

erik-krogh (dismissed):

java/ql/lib/semmle/code/java/regex/regex.qll
aschackmull: There's still `escapedCharacter`, which is used before being defined.

java/ql/lib/semmle/code/java/security/performance/ReDoSUtilSpecific.qll
```diff
@@ -0,0 +1,75 @@
+/**
+ * This module should provide a class hierarchy corresponding to a parse tree of regular expressions.
+ * This is the interface to the shared ReDoS library.
+ */
+
+import java
```
aschackmull, line 6: This qll defines the input to a cross-language shared module, so the `import java` ought to be private.
```suggestion
private import java
```

java/ql/lib/semmle/code/java/regex/RegexTreeView.qll
```diff
@@ -0,0 +1,1081 @@
+/** Provides a class hierarchy corresponding to a parse tree of regular expressions. */
+
+import java
```
aschackmull, line 3:
```suggestion
private import java
```

java/ql/lib/semmle/code/java/regex/regex.qll
```diff
@@ -0,0 +1,921 @@
...
+    this.getChar(pos + 1) = "Q"
+    or
+    result = false and
+    this.getChar(pos) = "\\" and
+    this.getChar(pos + 1) = "E"
+  }
+
+  /**
+   * Helper predicate for `quote`.
+   * Holds if the char at `pos` is the one-based `index`th occourence of a quote delimiter (`\Q` or `\E`)
```
aschackmull, line 65:
```suggestion
   * Holds if the char at `pos` is the one-based `index`th occurrence of a quote delimiter (`\Q` or `\E`)
```

java/ql/lib/semmle/code/java/regex/regex.qll
```diff
@@ -0,0 +1,921 @@
...
+   */
+  private boolean quoteDelimiter(int index, int pos) {
+    result = this.quoteDelimiter(pos) and
+    pos = rank[index](int p | exists(this.quoteDelimiter(p)))
+  }
+
+  /** Holds if a quoted sequence is found between `start` and `end` */
+  predicate quote(int start, int end) { this.quote(start, end, _, _) }
+
+  /** Holds if a quoted sequence is found between `start` and `end`, with ontent found between `inner_start` and `inner_end`. */
```
aschackmull, line 76:
```suggestion
  /** Holds if a quoted sequence is found between `start` and `end`, with content found between `inner_start` and `inner_end`. */
```

java/ql/lib/semmle/code/java/regex/regex.qll
```diff
@@ -0,0 +1,921 @@
...
+  /** Holds if the character at `pos` marks the end of a character class. */
+  private predicate charSetEnd0(int pos) {
+    this.nonEscapedCharAt(pos) = "]" and
+    /* special case: `[]]` and `[^]]` are valid char classes.  */
+    not this.charSetStart0(_, pos)
+  }
+
+  /**
+   * Holds if the character at `pos` starts a character set delimiter.
+   * Result is 1 for `[` and 0 for `]`.
```
aschackmull, line 131:
```suggestion
   * Result is 1 for `[` and -1 for `]`.
```

java/ql/lib/semmle/code/java/regex/regex.qll
```diff
@@ -0,0 +1,921 @@
...
+  private int max_zero(int x) { result = max([x, 0]) }
+
+  /**
+   * Gets the nesting depth of character classes after position `pos`,
+   * where `pos` is the position of a character set delimiter.
+   */
+  private int charSetDepth(int index, int pos) {
+    index = 1 and result = max_zero(charSetDelimiter(index, pos))
+    or
+    result = max_zero(charSetDelimiter(index, pos) + charSetDepth(index - 1, _))
```
aschackmull, line 158: Fix 'implicit this' warnings.

java/ql/lib/semmle/code/java/regex/regex.qll
```diff
@@ -0,0 +1,921 @@
...
+      not this.charSetEnd(start)
+      or
+      this.quote(start, end)
+    )
+  }
+
+  /** An indexed version of `charSetToken/3` */
+  private predicate charSetToken(int charset_start, int index, int token_start, int token_end) {
+    token_start =
+      rank[index](int start, int end | this.charSetToken(charset_start, start, end) | start) and
```
aschackmull, line 213: Simplify:
```suggestion
      rank[index](int start | this.charSetToken(charset_start, start, _)) and
```

java/ql/lib/semmle/code/java/regex/regex.qll
```diff
@@ -0,0 +1,921 @@
...
+    this.charSetToken(charset_start, token_start, token_end)
+  }
+
+  /**
+   * Holds if the character set starting at `charset_start` contains either
+   * a character or a range found between `start` and `end`.
+   */
+  predicate charSetChild(int charset_start, int start, int end) {
```
aschackmull, line 221: Sort predicates in dependency order. This should go below `charRange`.

java/ql/lib/semmle/code/java/regex/regex.qll
```diff
@@ -0,0 +1,917 @@
...
+  }
+
+  /**
+   * Holds if an escaped character is found between `start` and `end`.
+   * Escaped characters include hex values, octal values and named escapes,
+   * but excludes backreferences.
+   */
+  predicate escapedCharacter(int start, int end) {
+    this.escapingChar(start) and
+    not this.backreference(start, _) and
```
aschackmull, line 128: `backreference` is used before being defined. Please reorder.

java/ql/lib/semmle/code/java/regex/regex.qll
aschackmull: Agree. I just found that I couldn't review this code efficiently without it being topologically sorted. So we should sort the others similarly as a follow-up.

java/ql/lib/semmle/code/java/regex/regex.qll
```diff
@@ -0,0 +1,917 @@
...
+    this.getChar(start + 1) = "c" and
+    end = start + 3
+  }
+
+  private predicate namedBackreference(int start, int end, string name) {
+    this.escapingChar(start) and
+    this.getChar(start + 1) = "k" and
+    this.getChar(start + 2) = "<" and
+    end = min(int i | i > start + 2 and this.getChar(i) = ">") + 1 and
+    name = this.getText().substring(start + 3, end - 2)
```
aschackmull, line 115: Should be `end - 1`, right? `.substring` uses exclusive endpoint:
```suggestion
    name = this.getText().substring(start + 3, end - 1)
```

aschackmull (approved):
I think there's a lot of further opportunity for simplifications and improvements, but as-is I think this already provides plenty of value and that we should merge and tackle further improvements as follow-up work.

joefarebrother:
Great!
Does this also need a docs review before merging, or is the fact that the docs are essentially the same as for the other languages sufficient?

aschackmull:
> Does this also need a docs review before merging, or is the fact that the docs are essentially the same as for the other languages sufficient?

I think that's sufficient.

PR #7724, opened 2022-01-24
Add new groups for examples packs

aeisenberg:
Will make it easier to avoid publishing them.

### DIFF
--- a/cpp/ql/src/qlpack.yml
+++ b/cpp/ql/src/qlpack.yml
@@ -1,6 +1,8 @@
 name: codeql/cpp-queries
 version: 0.0.8-dev
-groups: cpp
+groups: 
+  - cpp
+  - queries
 dependencies:
     codeql/cpp-all: "*"
     codeql/suite-helpers: "*"
--- a/csharp/ql/examples/qlpack.yml
+++ b/csharp/ql/examples/qlpack.yml
@@ -1,4 +1,6 @@
 name: codeql/csharp-examples
-version: 0.0.2
+groups: 
+  - csharp
+  - examples
 dependencies:
   codeql/csharp-all: "*"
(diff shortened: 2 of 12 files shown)


michaelnebel (dismissed):
Looks good to me!

hvitved (approved):

cpp/ql/src/qlpack.yml
```diff
@@ -1,6 +1,6 @@
 name: codeql/cpp-queries
 version: 0.0.7-dev
-groups: cpp
+groups: cpp,queries
```
aibaars, line 3: Why is `groups` not a list? Using a comma-separated string instead of an array-type feels weird .

aibaars (changes requested):

aibaars:
@aeisenberg Don't forget to resolve the conflicting files and use arrays for the groups property in those too.

aibaars (approved):

csharp/ql/examples/qlpack.yml
```diff
@@ -1,4 +1,6 @@
 name: codeql/csharp-examples
-version: 0.0.2
+groups:
```
michaelnebel, line 2: Why has the version number been removed?

michaelnebel (approved):
Looks good from C# perspective.

csharp/ql/examples/qlpack.yml
dbartol: We're not making any particular versioning guarantees for the example packs. They're just sample code that sits in a Git repo for customers to look at, rather than an officially supported pack that gets published to the package registry and consumed by real Code Scanning runs.

dbartol (approved):

PR #7726, opened 2022-01-24
Ruby: Introduce `TAnyArrayElementContent`

hvitved:
This avoids potential combinatorial explosion in `SummaryComponentStack`s. Until now it has not been a problem, since array indices are limited to `0..5`, but with this change we should be able to track arbitrary known indices.

One example is the `dig` method, where `dig(x, y, z)` (with unknowns `x`, `y`, and `z`) yields 6^3 `SummaryComponentStack`s. With this change it yields just a single `SummaryComponentStack`, regardless of whether we increase the precision of which indices to track.

### DIFF
--- a/ruby/ql/lib/codeql/ruby/dataflow/FlowSummary.qll
+++ b/ruby/ql/lib/codeql/ruby/dataflow/FlowSummary.qll
@@ -47,9 +47,7 @@ module SummaryComponent {
    * index or known index. This predicate should never be used in the output specification
    * of a flow summary; use `arrayElementUnknown()` instead.
    */
-  SummaryComponent arrayElementAny() {
-    result in [arrayElementUnknown(), SC::content(TKnownArrayElementContent(_))]
-  }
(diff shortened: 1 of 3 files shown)


hmac (approved):
Am I right in thinking that we let array indices range from [0..10](https://github.com/github/codeql/blob/main/ruby/ql/lib/codeql/ruby/dataflow/internal/DataFlowPrivate.qll#L298)? Then it looks like the problem is even worse than you describe, as we'd get 12^3 `SummaryComponentStacks` for `dig(x,y,z)` (the 12th being `TUnknownArrayElementContent`)!

hvitved:
> Am I right in thinking that we let array indices range from [0..10](https://github.com/github/codeql/blob/main/ruby/ql/lib/codeql/ruby/dataflow/internal/DataFlowPrivate.qll#L298)? Then it looks like the problem is even worse than you describe, as we'd get 12^3 `SummaryComponentStacks` for `dig(x,y,z)` (the 12th being `TUnknownArrayElementContent`)!

Correct!

PR #7729, opened 2022-01-25
Ruby extractor: bump clap

hmac:
This supersedes #7499 and #7556 by bumping the `clap` dependency in both `generator` and `extractor`. It also fixes a warning arising from the version bump (`clap::App::args_from_usage` is deprecated).

The original PRs didn't update `Cargo.lock`, which looked a bit suspicious.

### DIFF
--- a/ruby/Cargo.lock
+++ b/ruby/Cargo.lock
@@ -63,17 +63,17 @@ checksum = "baf1de4339761588bc0619e3cbc0120ee582ebb74b53b4efbf79117bd2da40fd"
 
 [[package]]
 name = "clap"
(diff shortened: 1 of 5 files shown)


aibaars:
Looks good to me, but make sure to check that is also works for M1 Macs. We've had problems in the past with new versions of crates not being available for that architecture.

hmac:
@aibaars that's a good point - how do I test that? I don't have one myself.

hmac:
Confirmed this builds successfully on M1.

aibaars (approved):

PR #7733, opened 2022-01-25
Java: Add models for java.util.regex.Pattern and Matcher

pwntester:
This PR adds summaries for java.util.regex Pattern and Matcher so that taint can flow in cases such as:

```java
import java.util.regex.Matcher;
import java.util.regex.Pattern;

class Main {  

  public static void main(String args[]) { 
    String pattern = "\\$\\{(.*)\\}";
    Pattern r = Pattern.compile(pattern);
    String tainted = "${foo}";
    Matcher m = r.matcher(tainted);
    if (m.find( )) {
      sink(m.group(1));
    }
  } 
}
```


smowton:
The test doesn't compile:

```
Blah.java:39: error: reference to replaceAll is ambiguous
			out = in.replaceAll(null);
			        ^
  both method replaceAll(String) in Matcher and method replaceAll(Function<MatchResult,String>) in Matcher match
Blah.java:54: error: reference to replaceFirst is ambiguous
			out = in.replaceFirst(null);
			        ^
  both method replaceFirst(String) in Matcher and method replaceFirst(Function<MatchResult,String>) in Matcher match
2 errors
```

Otherwise this looks good to me. I imagine this will expose FPs because people are using regexes to sanitize things and we've previously by default assumed that any regex is sanitizing, but I'd be in favour of accepting these models and selectively introducing regex sanitizer steps as required.

pwntester:
Thanks @smowton I updated the test file and made it more meaningful

smowton:
Might want to clear up the tabs/spaces situation. Otherwise LGTM

pwntester:
ready for review @smowton

smowton (approved):

PR #7735, opened 2022-01-25
Python: promote log injection

yoff:
This PR promotes the log injection (CWE-117) query out of experimental.

Some concerns:
 - the models of loggers, particularly for Django is a bit curated. In general, should we do way more than the contribution when we promote?
 - the sanitization seems very specific. I added the string const compare sanitizer, but there are probably other ways to sanitize also. Should we have a concept for sanitizers similar to escaping or just reuse escaping?


python/ql/lib/semmle/python/frameworks/Django.qll
```diff
@@ -2295,4 +2295,17 @@ module PrivateDjango {
 
     override string getMimetypeDefault() { none() }
   }
+
+  // ---------------------------------------------------------------------------
+  // Logging
+  // ---------------------------------------------------------------------------
+  /**
+   * Django provides a standard Python logger.
+   */
```
RasmusWL, line 2304:
```suggestion
  /**
   * A standard Python logger instance from Django.
   * see https://github.com/django/django/blob/stable/4.0.x/django/utils/log.py#L11
   */
```

python/ql/lib/semmle/python/frameworks/Stdlib.qll
```diff
@@ -2683,14 +2698,12 @@ private module StdlibPrivate {
         method = "log" and
         msgIndex = 1
       |
-        this = Logger::instance().getMember(method).getACall()
-        or
-        this = API::moduleImport("logging").getMember(method).getACall()
```
RasmusWL, line 2688: I don't agree about this removal. I see that `API::moduleImport("logging")` is now modeled as a LoggerInstance... but that is not true, these are helper methods defined on the module leve.

python/ql/lib/semmle/python/security/dataflow/LogInjectionCustomizations.qll
```diff
@@ -0,0 +1,65 @@
+/**
+ * Provides default sources, sinks and sanitizers for detecting
+ * "command injection"
```
RasmusWL, line 3: This file says `"command injection"` all over the place. We need to replace that with `"log injection"` :+1:

python/ql/lib/semmle/python/security/dataflow/LogInjectionCustomizations.qll
```diff
@@ -0,0 +1,65 @@
...
+
+  /**
+   * A call to replace line breaks functions as a sanitizer.
+   */
+  class ReplaceLineBreaksSanitizer extends Sanitizer, DataFlow::CallCfgNode {
+    ReplaceLineBreaksSanitizer() {
+      this.getFunction().(DataFlow::AttrRead).getAttributeName() = "replace" and
+      this.getArg(0).asExpr().(StrConst).getText() in ["\r\n", "\n"]
+    }
+  }
```
RasmusWL, line 64: This sanitizer check seems a bit weak. What if it only did `text.replace("\r\n", "\n")`? I guess it's not trivial to write a sanitizer that knows we've eliminated both, so maybe this is ok? (what are your thoughts on this)

python/ql/lib/semmle/python/frameworks/Stdlib.qll
```diff
@@ -237,6 +237,42 @@ module Stdlib {
...
+      or
+      result = subclassRef().getReturn()
+      or
+      result = API::moduleImport("logging")
+      or
+      result = API::moduleImport("logging").getMember("root")
+      or
+      result = API::moduleImport("logging").getMember("getLogger").getReturn()
+    }
+  }
```
RasmusWL, line 275: I see that I created this non-standard setup with just using `API::Node`, but now that we're exposing it more publicly, we need to rewrite this to use the standard library modeling setup with `InstanceSource` and type-tracking (which has a snippet, so it's quite easy).

Looking over the code again, I don't quite know why I did this in the first place. `LoggerLogCall` can easily be written with the standard approach together with `DataFlow::MethodCallNode` :shrug:

RasmusWL (changes requested):
`python/ql/src/Security/CWE-117/LogInjection.ql` needs to have meta-data updated, to include security severity at least. I haven't been involved in looking at this query before, do you feel that `@precision high` is reasonable? For security-severity, we might just copy [the one from java](https://github.com/github/codeql/blob/1030ff70636ee5a3b2b9e375c67d2e32972097a7/java/ql/src/Security/CWE/CWE-117/LogInjection.ql#L7) (which is the same as the [one for JS](https://github.com/github/codeql/blob/f6da0305721c32b86c8acd84d218fee1edf9c4c7/javascript/ql/src/Security/CWE-117/LogInjection.ql#L7))

Looking up those `@security-severity` scores, I noticed that both queires had `@precision medium`. That got me thinking, in the example below, neither `username` nor `foo` can contain newlines, but would probably end up causing an alert :neutral_face:  so we need to be mindful about the precision we set here.

```py
@app.route("/users/<username>")
def show_user(username):
    foo = request.args.get("foo", 10)
    LOGGER.debug(f"showing user {username} with foo={foo}")
```


nitpick: QLDoc for classes doesn't follow our style-guide (must start with `A` or `An`)
nitpick: Would be very nice with documentations links.

(I've made a suggestion for Django to fix the two nitpicks above)

For the qhelp, I see that it is in large part just a copy of the [qhelp from the java query](https://github.com/github/codeql/blob/474bf576a7c73ddd2a976e3d54049aa423f15f5c/java/ql/src/Security/CWE/CWE-117/LogInjection.qhelp#L5). I personally have a mild preference for the [JS version](https://github.com/github/codeql/blob/c98dacf8421c2af9470e004ffae1760950bd4ff2/javascript/ql/src/Security/CWE-117/LogInjection.qhelp#L17) (since it better highlights the problems of HTML). Have you made a consideration between the two?

yoff:
> For the qhelp, I see that it is in large part just a copy of the [qhelp from the java query](https://github.com/github/codeql/blob/474bf576a7c73ddd2a976e3d54049aa423f15f5c/java/ql/src/Security/CWE/CWE-117/LogInjection.qhelp#L5). I personally have a mild preference for the [JS version](https://github.com/github/codeql/blob/c98dacf8421c2af9470e004ffae1760950bd4ff2/javascript/ql/src/Security/CWE-117/LogInjection.qhelp#L17) (since it better highlights the problems of HTML). Have you made a consideration between the two?

I did not even consider that we would have such options, I just checked that it looked reasonable :-) Thanks for the heads-up!

python/ql/lib/semmle/python/security/dataflow/LogInjectionCustomizations.qll
yoff: I have had the same concerns from the beginning.

python/ql/lib/semmle/python/frameworks/Stdlib.qll
yoff: Surely it is nicer, if we can get away with using an `API:Node`?

python/ql/lib/semmle/python/security/dataflow/LogInjectionCustomizations.qll
yoff: :doh:

python/ql/lib/semmle/python/frameworks/Stdlib.qll
yoff: Is there value in modelling instances specifically? Or could we just rename `instance` to something like `loggingMethodProvider`?

python/ql/src/Security/CWE-117/LogInjection.ql
```diff
@@ -4,7 +4,8 @@
  *              insertion of forged log entries by a malicious user.
  * @kind path-problem
  * @problem.severity error
- * @precision high
+ * @security-severity 7.8
+ * @precision medium
```
RasmusWL, line 8: with this change, we need to update the change-note, since it talks about this query being part of the default pack, which it no longer will be. (but security-extended)

RasmusWL (commented):
nice updates so far :+1:

python/ql/lib/semmle/python/security/dataflow/LogInjectionCustomizations.qll
yoff: Good point

python/ql/src/Security/CWE-117/LogInjection.ql
yoff: Good point!

python/ql/lib/semmle/python/frameworks/Stdlib.qll
yoff: Reverted to previous modelling approach as discussed offline.

python/ql/lib/semmle/python/frameworks/Stdlib.qll
yoff: Moved to standard setup as discussed offline.

python/ql/lib/semmle/python/frameworks/Stdlib.qll
```diff
@@ -237,6 +237,54 @@ module Stdlib {
       }
     }
   }
+
+  // ---------------------------------------------------------------------------
+  // logging
+  // ---------------------------------------------------------------------------
+  /**
+   * Provides models for the `logging.Logger` class
```
RasmusWL, line 245:
```suggestion
   * Provides models for the `logging.Logger` class and subclasses.
```

python/ql/lib/semmle/python/security/dataflow/LogInjectionCustomizations.qll
RasmusWL: I think it would be great to have a small comment in the code to explain this dilemma, and explain why the solution implemented was deemed ok.

RasmusWL (changes requested):
Two minor things, then we should be good to go :+1:

yoff:
> I think it would be great to have a small comment in the code to explain this dilemma, and explain why the solution implemented was deemed ok.

I actually got a bit annoyed with the explanation. Do let me know if we can live with postponing this, or if I should attempt a rewrite in this PR..

python/ql/lib/semmle/python/security/dataflow/LogInjectionCustomizations.qll
```diff
@@ -0,0 +1,70 @@
...
+    LoggingAsSink() { this = any(Logging write).getAnInput() }
+  }
+
+  /**
+   * A comparison with a constant string, considered as a sanitizer-guard.
+   */
+  class StringConstCompareAsSanitizerGuard extends SanitizerGuard, StringConstCompare { }
+
+  /**
+   * A call to replace line breaks functions as a sanitizer.
```
RasmusWL, line 57:
```suggestion
   * A call to replace line breaks, considered as a sanitizer.
```

python/ql/lib/semmle/python/security/dataflow/LogInjectionCustomizations.qll
```diff
@@ -0,0 +1,70 @@
...
+   */
+  class ReplaceLineBreaksSanitizer extends Sanitizer, DataFlow::CallCfgNode {
+    // This is actually not safe:
+    // - we do not check that all kinds of line breaks are replaced
+    // - we do not check that one kind of line breaks is not replaced by another
+    // however, we lack a simple way to do better, and the query would likely
+    // be too noisy without this. Consider rewriting using flow states.
```
RasmusWL, line 64:
```suggestion
    // Note: This sanitizer is not 100% accurate, since:
    // - we do not check that all kinds of line breaks are replaced
    // - we do not check that one kind of line breaks is not replaced by another
    //
    // However, we lack a simple way to do better, and the query would likely
    // be too noisy without this. 
    //
    // TODO: Consider rewriting using flow states.
```

I think it's nice to have the `TODO:` part in there, so it's possible to grep for.

RasmusWL (commented):
> > I think it would be great to have a small comment in the code to explain this dilemma, and explain why the solution implemented was deemed ok.
> 
> I actually got a bit annoyed with the explanation. Do let me know if we can live with postponing this, or if I should attempt a rewrite in this PR..

Thanks for writing this :+1: I think it's very nice that the next person thinking about this code can just read the comment explaining we thought about this :+1: I have a small suggestion for writing it, but otherwise LGTM :+1:

python/ql/lib/semmle/python/security/dataflow/LogInjectionCustomizations.qll
yoff: Good suggestion

python/ql/lib/semmle/python/security/dataflow/LogInjectionCustomizations.qll
yoff: Thanks, I considered it briefly, but seeing it written out, the TODO is much better.

RasmusWL (approved):

yoff:
Thanks for the auto format fix :-) (it confused me for a bit that I could not get it to change the file locally 😁)

RasmusWL:
Python 3 Language tests failed seems to have been flaky, so I've rerun them. :crossed_fingers:

RasmusWL:
Tests are all good, so :shipit:

RasmusWL:
Performance does looks fine, so in it goes :+1:

PR #7737, opened 2022-01-25
C++: Upgrade cpp/cleartext-storage-file

geoffw0:
Resolves some of the inconsistency between `cpp/cleartext-transmission`, `cpp/cleartext-storage-file` and `cpp/cleartext-storage-buffer` by upgrading the second one to non-local taint flow (with paths).

LGTM run: https://lgtm.com/query/834234913031404157/
LGTM diff query: https://lgtm.com/query/495444284760597175/

As you can see above, this makes the query a little noisier, but the quality of results seems to be similar to before.


MathiasVP (approved):
LGTM!

PR #7740, opened 2022-01-25
JS: promote the js/jwt-missing-verification query out of experimental

erik-krogh:
[Evaluation looks OK](https://github.com/github/codeql-dca-main/tree/data/erik-krogh/CWE-347__default__CustomSuite__1/reports). No change in alerts, and performance that seems to be within the margin of error. 

The query flags basically the same issue as [this java query](https://github.com/github/codeql/blob/main/java/ql/src/Security/CWE/CWE-347/MissingJWTSignatureCheck.ql).


kaeluka:
it looks good to me, but I'd like someone else to take a look as well :)

javascript/ql/src/Security/CWE-347/MissingJWTKeyVerification.ql
```diff
@@ -10,12 +10,11 @@
...
 import javascript
-import DataFlow
 import semmle.javascript.RestrictedLocations
 
-from CallNode call
+from DataFlow::CallNode call
 where
-  call = moduleMember("jsonwebtoken", "verify").getACall() and
+  call = DataFlow::moduleMember("jsonwebtoken", "verify").getACall() and
   unique(boolean b | b = call.getArgument(1).analyze().getABooleanValue()) = false
```
esbena, line 18: This should be used instead: 
https://github.com/github/codeql/blob/ee52774e9079562ed30ce9998707f120de6192cb/javascript/ql/lib/semmle/javascript/dataflow/TypeInference.qll#L99-L100

javascript/ql/src/Security/CWE-347/MissingJWTKeyVerification.ql
```diff
@@ -10,12 +10,11 @@
...
 select call.asExpr().(FirstLineOf),
```
esbena, line 19: Could we select the faulty argument instead with a message like:

`This argument disables the integrity enforcement of the token verification.`

esbena (dismissed):

esbena (approved):

javascript/ql/src/Security/CWE-347/MissingJWTKeyVerification.qhelp
```diff
@@ -0,0 +1,42 @@
...
+
+<recommendation>
+<p>
+Calls to <code>verify()</code> functions should use a cryptographic secret or key to decode JWT payloads.
+</p>
+</recommendation>
+
+<example>
+<p>
+In the example below <code>false</code> is used to disable the integrity enforcement of a JWT payload.
```
mchammer01, line 24:
```suggestion
In the example below, <code>false</code> is used to disable the integrity enforcement of a JWT payload. 
```

javascript/ql/src/Security/CWE-347/MissingJWTKeyVerification.qhelp
```diff
@@ -0,0 +1,42 @@
+<!DOCTYPE qhelp PUBLIC "-//Semmle//qhelp//EN" "qhelp.dtd"> <qhelp>
+
+<overview>
+<p>
+Applications decoding JSON Web Token (JWT) may be misconfigured due to the <code>None</code> algorithm.
```
mchammer01, line 5: Not sure about this suggestion, but token in its singular form reads oddly in this sentence.
```suggestion
Applications decoding JSON Web Tokens (JWT) may be misconfigured due to the <code>None</code> algorithm.
```

mchammer01 (approved):
@erik-krogh - LGTM ✨ 
A few minor suggestions and a typo.

javascript/ql/src/Security/CWE-347/MissingJWTKeyVerification.qhelp
erik-krogh: The plural version seems is used often enough, and yes it sounds better :+1:

mchammer01 (approved):

esbena (approved):

PR #7741, opened 2022-01-25
C#: Exclude extractor arguments from `compilation_args` relation

hvitved:
The relation is only supposed to contain the arguments of the C# compilation call.


michaelnebel (approved):
Looks good to me!

tamasvajk (approved):

PR #7742, opened 2022-01-25
C++: Upgrade cpp/cleartext-storage-buffer

geoffw0:
Resolves a bit more of the inconsistency between `cpp/cleartext-transmission`, `cpp/cleartext-storage-file` and `cpp/cleartext-storage-buffer` by upgrading the third one from security to AST taint flow.

We still use the old security library's `isUserInput` to identify inputs, that could be upgraded as well but may have a bigger impact on results.

LGTM diff query: https://lgtm.com/query/6710910426361157574/ (sparse results, no differences)

### DIFF
--- a/cpp/ql/src/Security/CWE/CWE-311/CleartextBufferWrite.ql
+++ b/cpp/ql/src/Security/CWE/CWE-311/CleartextBufferWrite.ql
@@ -13,21 +13,32 @@
 
 import cpp
 import semmle.code.cpp.security.BufferWrite
-import semmle.code.cpp.security.TaintTracking
 import semmle.code.cpp.security.SensitiveExprs
-import TaintedWithPath
+import semmle.code.cpp.security.Security
+import semmle.code.cpp.dataflow.TaintTracking
+import DataFlow::PathGraph
 
-class Configuration extends TaintTrackingConfiguration {
-  override predicate isSink(Element tainted) { exists(BufferWrite w | w.getASource() = tainted) }
+/**
+ * Taint flow from user input to a buffer write.
+ */
+class ToBufferConfiguration extends TaintTracking::Configuration {
+  ToBufferConfiguration() { this = "ToBufferConfiguration" }
+
+  override predicate isSource(DataFlow::Node source) { isUserInput(source.asExpr(), _) }
+
+  override predicate isSink(DataFlow::Node sink) {
+    exists(BufferWrite w | w.getASource() = sink.asExpr())
+  }
 }
 
 from
-  BufferWrite w, Expr taintedArg, Expr taintSource, PathNode sourceNode, PathNode sinkNode,
-  string taintCause, SensitiveExpr dest
+  ToBufferConfiguration config, BufferWrite w, Expr taintSource, DataFlow::PathNode sourceNode,
+  DataFlow::PathNode sinkNode, string taintCause, SensitiveExpr dest
 where
-  taintedWithPath(taintSource, taintedArg, sourceNode, sinkNode) and
+  config.hasFlowPath(sourceNode, sinkNode) and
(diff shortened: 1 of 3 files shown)


cpp/ql/src/Security/CWE/CWE-311/CleartextBufferWrite.ql
```diff
@@ -13,21 +13,32 @@
 
 import cpp
 import semmle.code.cpp.security.BufferWrite
-import semmle.code.cpp.security.TaintTracking
 import semmle.code.cpp.security.SensitiveExprs
-import TaintedWithPath
+import semmle.code.cpp.security.Security
+import semmle.code.cpp.dataflow.TaintTracking
```
MathiasVP, line 18: This is going from IR taint tracking to AST taint tracking, which is a bit sad. Could we replace this with `semmle.code.cpp.ir.dataflow.TaintTracking`?

geoffw0: Ah that's interesting, I thought the the security lib / default taint tracking used AST taint.

I'll update it to be IR-based...

MathiasVP: Fantastic! Here's a quick summary:
- `semmle.code.cpp.security.TaintTracking` is an alias for `semmle.code.cpp.ir.dataflow.DefaultTaintTracking` (which is an IR taint tracking library that uses an IR `DataFlow` configuration)
- `semmle.code.cpp.dataflow.TaintTracking` is an AST taint tracking library
- `semmle.code.cpp.ir.dataflow.TaintTracking` is an IR taint tracking library built on an IR `TaintTracking` configuration. By now I think this library should mostly be superior to `semmle.code.cpp.dataflow.TaintTracking`, but there are a few limitations that mostly shouldn't matter in practice.

geoffw0: I've just done two things:
 - put it back on IR taint tracking, but no longer via relying on DefaultTaintTracking.  This is probably slightly untidier than what we started with, but opens the possibility to replace security library `isUserInput` with modern flow sources (which I should look at later today...)
 - I've also added to QLDoc at the top of two of the library files you described above.  That should help us (and others) remember which is which in future.

geoffw0: I've updated it to use `FlowSource`, which is cleaner and appears to give better textual explanations of the source.

cpp/ql/src/Security/CWE/CWE-311/CleartextBufferWrite.ql
```diff
@@ -12,31 +12,39 @@
...
+}
+
 /**
  * Taint flow from user input to a buffer write.
  */
 class ToBufferConfiguration extends TaintTracking::Configuration {
   ToBufferConfiguration() { this = "ToBufferConfiguration" }
 
-  override predicate isSource(DataFlow::Node source) { isUserInput(source.asExpr(), _) }
+  override predicate isSource(DataFlow::Node source) { isUserInput(exprForNode(source), _) }
```
MathiasVP: <s>Do we get the same results if we use `source instanceof FlowSource`? Then we don't need this ugly `exprForNode` (as `FlowSource.getSourceType` can be used instead of `isUserInput`  to get the `taintCause`).</s> Oh, you did exactly that in the last commit. Excellent!

MathiasVP (approved):
LGTM! Are the results still the expected set of results?

geoffw0:
Yep, results on the test and a very ugly (LGTM diff](https://lgtm.com/query/4199143955747163646/) are the same.  I was hoping to get some new ones, but there you go.

PR #7744, opened 2022-01-25
JS: Move experimental notice to the bottom of the ML-powered query help

henrymercer:
The Code Scanning UI shows just the first paragraph of the query help as a summary, until a user chooses to expand the help. We decided it was more useful to display the standard query help in this summary compared to the experimental query notice, since there is already a notice about experimental queries on the alert show page.


AlonaHlobina (approved):
Looks great! Thank you!

PR #7746, opened 2022-01-26
C#: Remove some unused legacy relations from the DB scheme

hvitved:
- `externalDefects`
- `externalMetrics`
- `externalData`
- `snapshotDate`
- `duplicateCode`
- `similarCode`
- `tokens.rel`


michaelnebel:
Looks good to me! Did you manually change the *stats* file and should we consider run (1) DCA stat generation (2) DCA performance test

michaelnebel (approved):
Excellent to eliminate dead stuff!

hvitved:
> Did you manually change the _stats_ file and should we consider run (1) DCA stat generation (2) DCA performance test

I manually deleted the entries for the deleted relations. A DCA performance test should not be needed, as these stats only pertained to relations that are not used in any queries.

PR #7749, opened 2022-01-26
C# 10 - Lambda improvements.

michaelnebel:
In this PR, we make support for C# 10 [Improvements on lambda expressions](https://docs.microsoft.com/en-us/dotnet/csharp/whats-new/csharp-10#lambda-expression-improvements)
* Lambda expressions may have a natural type, where the compiler can infer a delegate type from the lambda expression or method group (test only).
* Lambda expressions may declare a return type when the compiler can't infer it.
* Attributes can be applied to lambda expressions.

What is still needed in this PR
- [x] Rebase on PR related to https://github.com/github/codeql-csharp-team/issues/219
- [x] Change note.
- [x] Database upgrade / downgrade scripts.
- [x] DCA run for new stats.
- [x] DCA run for performance test.


hvitved (commented):
Looks great! Go ahead with generating and adding new stats.

michaelnebel:
DCA run shows that performance is not affected.

hvitved (approved):

PR #7752, opened 2022-01-26
Fix issues with downgrade pack releases

edoardopirovano:
This PR fixes two issues with the release process for downgrade packs:

1. An existing downgrade pack has a version number lower than the others because it didn't make it into the release branch and therefore didn't get released when the others did. Bump it to bring it back in sync.
2. The automation for releasing packs isn't picking up downgrade ones because they aren't listed in our `.codeqlmanifest.json`, so even if it _had_ been on the release branch it would not have been released correctly as things stand.

### DIFF
--- a/.codeqlmanifest.json
+++ b/.codeqlmanifest.json
@@ -4,6 +4,7 @@
         "*/ql/lib/qlpack.yml",
         "*/ql/test/qlpack.yml",
         "*/ql/examples/qlpack.yml",
+        "*/downgrades/qlpack.yml",
         "cpp/ql/test/query-tests/Security/CWE/CWE-190/semmle/tainted/qlpack.yml",
         "javascript/ql/experimental/adaptivethreatmodeling/lib/qlpack.yml",
         "javascript/ql/experimental/adaptivethreatmodeling/src/qlpack.yml",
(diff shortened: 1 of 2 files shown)


aibaars:
Have you checked that C# still works well in VSCode? There might be a reason why downgrades are not in the normal `ql` folder.

edoardopirovano:
> Have you checked that C# still works well in VSCode? There might be a reason why downgrades are not in the normal `ql` folder.

Good point. I've run a couple of queries in VS code and clicked around the standard library including in the parts affected by the downgrade script (i.e. the DB scheme for the new relation, and some bits of the QL source that use it). All seems to work smoothly.

As for where these packs _should_ be. I'm not sure how conscious the decision to not put them in `ql` was (@michaelnebel can comment on that). Regardless of where the pack goes, we'll need a new line in the manifest since the existing patterns won't match `ql/upgrades` either.

aibaars (approved):

cklin:
Thank you @edoardopirovano!  Tagging https://github.com/github/codeql-core/issues/1717 so that we can find this PR later.

edoardopirovano:
> Thank you @edoardopirovano!  Tagging https://github.com/github/codeql-core/issues/1717 so that we can find this PR later.

👍

Note I've held off on merging this as @dbartol mentioned we may not want to release the downgrade packs after all.

edoardopirovano:
Closing as we have decided to not publish these packa to the registry after all.

PR #7758, opened 2022-01-26
C++: Do not report "Declaration hides variable" for unnamed variables

jketema:


geoffw0 (commented):
LGTM.

I appreciate the second test case showing one we could flag, but currently don't due to the extractor limitation.

MathiasVP (approved):

PR #7763, opened 2022-01-26
QL: add unused-field query

erik-krogh:
Inspired by a recent code-review. 

Detects when a field is not bound by the charpred.  
Also handles the cases where the field is defined in an abstract super-class. 

I fixed all five errors detected by the query. They were pretty benign except for the Python warning in `Flask.qll`.  
(I think the error in `Flask.qll` was from not copy-pasting a line from the above class, so I did that).

### DIFF
--- a/ql/ql/src/queries/performance/UnusedField.ql
+++ b/ql/ql/src/queries/performance/UnusedField.ql
@@ -31,13 +31,6 @@ where
       access.getEnclosingPredicate() = p
     )
   ) and
-  (if clz = implClz then extraMsg = "." else extraMsg = " of any class between it and $@.") and
-  // The `implClz` does not override `field` with a more specific type.
-  not exists(FieldDecl override |
-    override = implClz.getDeclaration().getAField() and
-    override.getName() = field.getName() and
-    override.hasAnnotation("override") and
-    override.getVarDecl().getType() != field.getVarDecl().getType()
-  )
+  (if clz = implClz then extraMsg = "." else extraMsg = " of any class between it and $@.")
 select clz, "The field $@ declared in $@ is not used in the characteristic predicate" + extraMsg,
   field, field.getName(), clz, clz.getName(), implClz, implClz.getName()
--- a/ql/ql/src/queries/performance/UnusedField.ql
+++ b/ql/ql/src/queries/performance/UnusedField.ql
@@ -16,8 +16,7 @@ where
   implClz.getASuperType*() = clz and
   // The field is not accessed in the charpred (of any of the classes)
   not exists(FieldAccess access |
-    access.getEnclosingPredicate() =
-      [clz.getDeclaration().getCharPred(), implClz.getDeclaration().getCharPred()]
+    access.getEnclosingPredicate() = [clz, implClz].getDeclaration().getCharPred()
   ) and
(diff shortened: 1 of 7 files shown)


RasmusWL (approved):
:+1: for Python 

:relaxed: that is was only for an experimental query

atorralba (dismissed):
Good catch! Java 👍

erik-krogh:
> relaxed that is was only for an experimental query

The expected output changed after I fixed it, so I pushed the updated output.

ql/ql/src/queries/performance/UnusedField.ql
```diff
@@ -0,0 +1,43 @@
...
+    )
+  ) and
+  (if clz = implClz then extraMsg = "." else extraMsg = " of any class between it and $@.") and
+  // The `implClz` does not override `field` with a more specific type.
+  not exists(FieldDecl override |
+    override = implClz.getDeclaration().getAField() and
+    override.getName() = field.getName() and
+    override.hasAnnotation("override") and
+    override.getVarDecl().getType() != field.getVarDecl().getType()
+  )
```
aschackmull, line 41: Is this relevant? The field should still be mentioned in a char-pred, right? Merely overriding the type doesn't prevent the cartesian product.

erik-krogh: You're right, that should not be relevant.  
Removing that restriction produces 1 extra alert here: https://github.com/github/codeql/blob/f1ec2e3260b622dd03e46d2fe0b50441835eaa12/cpp/ql/lib/semmle/code/cpp/PrintAST.qll#L412.  
But that result is an FP for other reasons (the field is bound in `DeclarationEntryNode`).   

I'll look into it.

erik-krogh: I fixed the FP by just looking at the name of the field, instead of checking that the `FieldDecl` was the same.  
(The `FieldDecl` was different because of an `override`).

ql/ql/src/queries/performance/UnusedField.ql
```diff
@@ -0,0 +1,36 @@
...
+import ql
+
+from ClassType clz, ClassType implClz, FieldDecl field, string extraMsg
+where
+  clz.getDeclaration().getAField() = field and
+  implClz.getASuperType*() = clz and
+  // The field is not accessed in the charpred (of any of the classes)
+  not exists(FieldAccess access |
+    access.getEnclosingPredicate() =
+      [clz.getDeclaration().getCharPred(), implClz.getDeclaration().getCharPred()]
```
kaeluka, line 20: isn't `clz` implied by `implClz`, as you're using the `*` operator in L16?

kaeluka: If not, this can be `[clz, implClz].getDeclaration().getCharPred()`, right?

erik-krogh: > isn't `clz` implied by `implClz`, as you're using the * operator in L16?

Not necessarily. There are other restrictions in the query that makes it so `clz` and `implClz` must sometimes be different classes. (I don't have any examples to link to though).  

> If not, this can be [clz, implClz].getDeclaration().getCharPred(), right? 

Yes it can. I'll do that.

PR #7766, opened 2022-01-27
C#: Make `TypeParameterConstraints` a `CachedEntity`

hvitved:
Extracted from https://github.com/github/codeql/pull/7456, in preparation for shared extraction.

### DIFF
--- a/csharp/extractor/Semmle.Extraction.CSharp/Entities/Types/TypeParameterConstraints.cs
+++ b/csharp/extractor/Semmle.Extraction.CSharp/Entities/Types/TypeParameterConstraints.cs
@@ -1,14 +1,65 @@
+using Microsoft.CodeAnalysis;
 using System.IO;
 
 namespace Semmle.Extraction.CSharp.Entities
 {
-    internal class TypeParameterConstraints : FreshEntity
+    internal class TypeParameterConstraints : CachedEntity<ITypeParameterSymbol>
     {
-        public TypeParameterConstraints(Context cx)
-            : base(cx) { }
+        private readonly TypeParameter parent;
 
-        protected override void Populate(TextWriter trapFile)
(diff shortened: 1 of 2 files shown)


csharp/extractor/Semmle.Extraction.CSharp/Entities/Types/TypeParameterConstraints.cs
```diff
@@ -1,14 +1,65 @@
...
+        public TypeParameterConstraints(Context cx, TypeParameter parent)
+            : base(cx, parent.Symbol)
         {
+            this.parent = parent;
+        }
+
+        public override void WriteId(EscapingTextWriter trapFile)
+        {
+            trapFile.WriteSubId(parent);
+            trapFile.Write(";typeparameterconstraints");
```
michaelnebel, line 19: Could you remind me, why this is needed (we are only doing this for entities and not fresh entities).

michaelnebel (approved):
Looks good to me!

csharp/extractor/Semmle.Extraction.CSharp/Entities/Types/TypeParameterConstraints.cs
hvitved: It is to give a unique ID to the entity. If we simply did `trapFile.WriteSubId(parent);` it would overlap with the ID of the parent.

michaelnebel: Ok, so the ID we write into the trap, is basically the same as the parent id except that we postfix it with ";typeparameterconstraints"?

hvitved: Yes, exactly.

PR #7767, opened 2022-01-27
Java: Regenerate framework models automatically

bmuskalla:
Takes the model generator and any framework model we have on main and regenerates the model using the latest and greatest model generator. This enables us to update and evolve the models without manual intervention.

### DIFF
--- a/.github/workflows/mad_regenerate-models.yml
+++ b/.github/workflows/mad_regenerate-models.yml
@@ -4,12 +4,18 @@ on:
   workflow_dispatch:
   schedule:
     - cron: "30 2 * * *"
+  pull_request:
+    branches:
+      - main
+    paths:
+      - ".github/workflows/mad_regenerate-models.yml"
 
 jobs:
   regenerate-models:
     runs-on: ubuntu-latest
     strategy:
       matrix:
+        # placeholder required for each axis, excluded below, replaced by the actual combinations (see include)
         slug: ["placeholder"]
         ref: ["placeholder"]
         include:
--- a/.github/workflows/mad_regenerate-models.yml
+++ b/.github/workflows/mad_regenerate-models.yml
@@ -1,0 +1,56 @@
+name: Regenerate framework models
+
+on:
+  workflow_dispatch:
+  schedule:
+    - cron: "30 2 * * *"
+
+jobs:
+  regenerate-models:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        slug: ["placeholder"]
+        ref: ["placeholder"]
+        include:
+          - slug: "apache/commons-io"
+            ref: "8985de8fe74f6622a419b37a6eed0dbc484dc128"
+        exclude:
+          - slug: "placeholder"
+            ref: "placeholder"
+    steps:
+      - name: Clone self (github/codeql)
+        uses: actions/checkout@v2
+      - name: Setup CodeQL binaries
+        uses: ./.github/actions/fetch-codeql
+      - name: Clone repositories
+        uses: actions/checkout@v2
+        with:
+          path: repos/${{ matrix.ref }}
+          ref: ${{ matrix.ref }}
+          repository: ${{ matrix.slug }}
+      - name: Build database
+        env:
+          SLUG: ${{ matrix.slug }}
+          REF: ${{ matrix.ref }}
+        run: |
+          mkdir dbs
+          cd repos/${REF}
+          SHORTNAME=${SLUG//[^a-zA-Z0-9_]/}
+          codeql database create --language=java ../../dbs/${SHORTNAME}
+      - name: Regenerate models in-place
+        env:
+          SLUG: ${{ matrix.slug }}
+        run: |
+          SHORTNAME=${SLUG//[^a-zA-Z0-9_]/}
+          java/ql/src/utils/model-generator/RegenerateModels.py "${SLUG}" dbs/${SHORTNAME}
+      - name: Stage changes
+        run: |
+          find java -name "*.qll" -print0 | xargs -0 git add
+          git status
+          git diff --cached > models.patch
+      - uses: actions/upload-artifact@v2
+        with:
+          name: patch
+          path: models.patch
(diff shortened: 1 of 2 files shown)


.github/workflows/mad_regenerate-models.yml
```diff
@@ -0,0 +1,56 @@
...
+          repository: ${{ matrix.slug }}
+      - name: Build database
+        env:
+          SLUG: ${{ matrix.slug }}
+          REF: ${{ matrix.ref }}
+        run: |
+          mkdir dbs
+          cd repos/${REF}
+          SHORTNAME=${SLUG//[^a-zA-Z0-9_]/}
+          codeql database create --language=java ../../dbs/${SHORTNAME}
```
bmuskalla, line 40: We could consider using `actions/cache` for this

bmuskalla:
@tamasvajk In case you have some time, I'd appreciate if someone could look at this action. Happy to pair if anything is unclear

java/ql/src/utils/model-generator/RegenerateModels.py
```diff
@@ -0,0 +1,47 @@
...
+import json
+import os
+import requests
+import shutil
+import subprocess
+import tempfile
+import sys
+
+
+lgtmSlugToModelFile = {
```
michaelnebel, line 15: Should we consider introducing an invariant instead of an explicit mapping?
The root path could be `java/ql/lib/semmle/code/java/frameworks/` and the slug can then be used to decide subfolder an file name.
eg. slug = "apache/commons-beanutils" -> apache/CommonsBeanutilsGenerated.qll.

This would solve the minor code cohesion issue that we otherwise need to remember to update both the workflow (slug, ref) pair and add an entry into lgtmSlugToModelFile.

java/ql/src/utils/model-generator/RegenerateModels.py
```diff
@@ -0,0 +1,47 @@
...
+    return subprocess.check_output(
+        ["git", "rev-parse", "--show-toplevel"]).decode("utf-8").strip()
+
+
+def regenerateModel(lgtmSlug, extractedDb):
+    tmpDir = tempfile.mkdtemp()
+    print("============================================================")
+    print("Generating models for " + lgtmSlug)
+    print("============================================================")
+    modelFile = lgtmSlugToModelFile[lgtmSlug]
```
michaelnebel, line 33: Maybe add some exception handling as we might have forgotten to update the lgtmSlugToModelFile, when adding entries to the strategy matrix in the workflow?

michaelnebel (approved):
This is really good work!

I have added a couple of suggestions, but these could also just be follow ups - so not anything blocking.

java/ql/src/utils/model-generator/RegenerateModels.py
bmuskalla: I used to do that but quickly realized that we have different locations for our models. And yes, we could try and move/rename them all into the same spot but hold back on that. I'm happy either way.

bmuskalla:
@michaelnebel Good points overall on the `lgtmSlugToModelFile` - one major consideration I had (and I'm still torn) is whether we even need the python script at all. We could simplify the whole PR by moving everything into the action. The only downside would be that we loose the ability to locally regenerate those (but that case is already a lot harder than initially as I had to move the database building into the action. Thinking about it right now, I feel like moving this into the action makes the most sense. Locally, you can always run `GenerateFlowModel.py` manually and point it to the right database. Thoughts?

java/ql/src/utils/model-generator/RegenerateModels.py
michaelnebel: Are you thinking about models that has been auto generated and then don't follow this pattern on location or hand written models?

michaelnebel:
> @michaelnebel Good points overall on the `lgtmSlugToModelFile` - one major consideration I had (and I'm still torn) is whether we even need the python script at all. We could simplify the whole PR by moving everything into the action. The only downside would be that we loose the ability to locally regenerate those (but that case is already a lot harder than initially as I had to move the database building into the action. Thinking about it right now, I feel like moving this into the action makes the most sense. Locally, you can always run `GenerateFlowModel.py` manually and point it to the right database. Thoughts?

Without any experience, it is hard for me to say, but my *immediate* response would be to keep it as a python script, just to ease potential debugging/extensions/rewriting of the script.

.github/workflows/mad_regenerate-models.yml
```diff
@@ -0,0 +1,56 @@
...
+    strategy:
+      matrix:
+        slug: ["placeholder"]
+        ref: ["placeholder"]
+        include:
+          - slug: "apache/commons-io"
+            ref: "8985de8fe74f6622a419b37a6eed0dbc484dc128"
+        exclude:
+          - slug: "placeholder"
+            ref: "placeholder"
```
tamasvajk, line 20: Why do we need the `placeholder`s here (, which are excluded)?

tamasvajk (commented):
Looks plausible to me.

BTW, where did you test this workflow? I used to push my workflow changes first to `dsp-testing` to see them in action (pun intended).

.github/workflows/mad_regenerate-models.yml
bmuskalla: In order to create a paramterized job without the combinatoric nature of a matrix build (and all axis are required to have at least one value)

bmuskalla:
> Without any experience, it is hard for me to say, but my immediate response would be to keep it as a python script, just to ease potential debugging/extensions/rewriting of the script

Fair enough, keeping it for now

> BTW, where did you test this workflow?
Used to test it on a branch. But will actually make it so it runs on a PR that changes the workflow itself

java/ql/src/utils/model-generator/RegenerateModels.py
michaelnebel: Alright. Thank you for the context. If this is undecided, then just leave it as it is :-)

michaelnebel (approved):
Looks good to me!

PR #7772, opened 2022-01-27
C#: Guard against `AssociatedSymbol` not being an `IEventSymbol`

hvitved:
Apply same logic as for property/indexer accessors to account for cases where the associated event cannot be determined. I have not been able to reproduce such cases locally, though we have seen reports of it happening.


csharp/extractor/Semmle.Extraction.CSharp/Entities/EventAccessor.cs
```diff
@@ -3,45 +3,46 @@
...
+        /// if there is no associated symbol.
         /// </summary>
-        private IEventSymbol? EventSymbol => Symbol.AssociatedSymbol as IEventSymbol;
+        public static IEventSymbol? GetEventSymbol(IMethodSymbol symbol) =>
+            symbol.AssociatedSymbol as IEventSymbol;
 
         public override void Populate(TextWriter trapFile)
         {
             PopulateMethod(trapFile);
             ContainingType!.PopulateGenerics();
```
tamasvajk, line 26: Shouldn't we have a `PopulateModifiers(trapFile);` call here?

csharp/extractor/Semmle.Extraction.CSharp/Entities/EventAccessor.cs
```diff
@@ -53,14 +54,14 @@ public override void Populate(TextWriter trapFile)
             Overrides(trapFile);
         }
```
tamasvajk, line 55: Properties have the following too. Shouldn't it be added?

```
            if (Symbol.FromSource() && Block is null)
            {
                trapFile.compiler_generated(this);
            }
```

tamasvajk (commented):
I've added some minor questions.

csharp/extractor/Semmle.Extraction.CSharp/Entities/EventAccessor.cs
hvitved: Apparently not:
```
error CS1609: Modifiers cannot be placed on event accessor declarations
```

csharp/extractor/Semmle.Extraction.CSharp/Entities/EventAccessor.cs
hvitved: Makes sense.

tamasvajk (approved):

tamasvajk (approved):

PR #7773, opened 2022-01-27
JS: add a js/file-system-race query

erik-krogh:
Inspired by [this C++ query](https://github.com/github/codeql/blob/main/cpp/ql/src/Security/CWE/CWE-367/TOCTOUFilesystemRace.ql).  
The precision is set to `medium`, because protecting against these kinds of races is uncommon in JS, which is because the attack is irrelevant in most JavaScript applications. 

[Here is a test run](https://lgtm.com/query/112645255730185844/). It's noisy. 

The documentation of the `fs` module has some description of the bad pattern: https://nodejs.org/api/fs.html  
(Search for: "recommended"). 

[An evaluation](https://github.com/github/codeql-dca-main/tree/data/erik-krogh/CWE-367__default__CustomSuite__1/reports) is about as noisy as expected.  
Performance is not great, not terrible. It might be a fluke, but I think it's OK for a non-default query.

### DIFF
--- a/javascript/ql/src/Security/CWE-367/FileSystemRace.qhelp
+++ b/javascript/ql/src/Security/CWE-367/FileSystemRace.qhelp
@@ -44,7 +44,7 @@ This can be avoided by using <code>fs.open</code> to get a file descriptor, and 
 
 <references>
 <li>
-Wikipedia: <a href="https://en.wikipedia.org/wiki/Time-of-check_to_time-of-use">Time-of-check to time-of-use.</a>
+Wikipedia: <a href="https://en.wikipedia.org/wiki/Time-of-check_to_time-of-use">Time-of-check to time-of-use</a>.
 </li>
 
 <li>
--- a/javascript/ql/src/Security/CWE-367/FileSystemRace.qhelp
+++ b/javascript/ql/src/Security/CWE-367/FileSystemRace.qhelp
@@ -1,0 +1,61 @@
+<!DOCTYPE qhelp PUBLIC
+  "-//Semmle//qhelp//EN"
+  "qhelp.dtd">
+<qhelp>
+
+<overview>
+<p>
+Often it is necessary to check the state of a file before using it. These checks usually take a file name
+to be checked, and if the check returns positively, then the file is opened or otherwise operated upon.
+</p>
+
+<p>
+However, in the time between the check and the operation, the underlying file referenced by the
+file name could be changed by an attacker, causing unexpected behavior.
+</p>
+
+</overview>
+
+<recommendation>
+<p>
+Use file descriptors instead of file names whenever possible.
+</p>
+</recommendation>
+
+<example>
+<p>
+The following example shows a case where it is checked whether a file inside the <code>/tmp/</code> folder exists, 
+and if it didn't exist the file is written to.
+</p>
+
+<sample src="examples/file-race-bad.js" />
+
+<p>
+However, in a multi-user environment the file might be created by another user in between the existence check and the write. 
+</p>
+
+<p>
+This can be avoided by using <code>fs.open</code> to get a file descriptor, and then use that file descriptor in the write operation.
+</p>
+
+<sample src="examples/file-race-good.js" />
+
+</example>
+
+<references>
+<li>
+Wikipedia: <a href="https://en.wikipedia.org/wiki/Time-of-check_to_time-of-use">Time-of-check to time-of-use</a>
+</li>
+
+<li>
+  The CERT Oracle Secure Coding Standard for C:
+  <a href="https://www.securecoding.cert.org/confluence/display/c/FIO01-C.+Be+careful+using+functions+that+use+file+names+for+identification">
+    FIO01-C. Be careful using functions that use file names for identification
+  </a>.
+</li>
+<li>
+NodeJS: 
+  <a href="https://nodejs.org/api/fs.html">The FS module</a>
+</li>
+</references>
+</qhelp>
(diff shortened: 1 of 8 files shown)


javascript/ql/src/Security/CWE-367/FileSystemRace.ql
```diff
@@ -0,0 +1,105 @@
...
+  )
+  or
+  AccessPath::getAnAliasedSourceNode(check.getPathArgument()).flowsTo(use.getPathArgument())
+}
+
+/**
+ * Holds if `check` happens before `use`.
+ */
+pragma[inline]
+predicate useAfterCheck(FileCheck check, FileUse use) {
```
esbena, line 84: Not a blocker: I suspect this to be the primary cause of slowness. Does it use the same form as the other "does A happen before B" predicates that we have?

erik-krogh: In `DeadStoreOfProperty.ql` we use `CFGNode::getASuccessor()`,   
but most other places seem to use `BasicBlock::getNode(i)`. 

It could be optimized in the same way as I did for `GlobalAccessPaths::DominatingPaths::hasDominatingWrite()`.  
(I'll do that if you think we could use a performance improvement here).

esbena: OK. Sounds like we have a solution that is not unreasonable then. I agree with "but I think it's OK for a non-default query.".

esbena (dismissed):

javascript/ql/src/Security/CWE-367/FileSystemRace.qhelp
```diff
@@ -0,0 +1,61 @@
...
+
+<li>
+  The CERT Oracle Secure Coding Standard for C:
+  <a href="https://www.securecoding.cert.org/confluence/display/c/FIO01-C.+Be+careful+using+functions+that+use+file+names+for+identification">
+    FIO01-C. Be careful using functions that use file names for identification
+  </a>.
+</li>
+<li>
+NodeJS: 
+  <a href="https://nodejs.org/api/fs.html">The FS module</a>
```
mchammer01, line 58:
```suggestion
  <a href="https://nodejs.org/api/fs.html">The FS module</a>.
```

javascript/ql/src/Security/CWE-367/FileSystemRace.qhelp
```diff
@@ -0,0 +1,61 @@
...
+This can be avoided by using <code>fs.open</code> to get a file descriptor, and then use that file descriptor in the write operation.
+</p>
+
+<sample src="examples/file-race-good.js" />
+
+</example>
+
+<references>
+<li>
+Wikipedia: <a href="https://en.wikipedia.org/wiki/Time-of-check_to_time-of-use">Time-of-check to time-of-use</a>
```
mchammer01, line 47:
```suggestion
Wikipedia: <a href="https://en.wikipedia.org/wiki/Time-of-check_to_time-of-use">Time-of-check to time-of-use.</a>
```

javascript/ql/src/Security/CWE-367/FileSystemRace.qhelp
```diff
@@ -0,0 +1,61 @@
...
+<recommendation>
+<p>
+Use file descriptors instead of file names whenever possible.
+</p>
+</recommendation>
+
+<example>
+<p>
+The following example shows a case where it is checked whether a file inside the <code>/tmp/</code> folder exists, 
+and if it didn't exist the file is written to.
```
mchammer01, line 28:
```suggestion
and if it doesn't, the file is written to that location.
```

javascript/ql/src/Security/CWE-367/FileSystemRace.qhelp
```diff
@@ -0,0 +1,61 @@
...
+
+<recommendation>
+<p>
+Use file descriptors instead of file names whenever possible.
+</p>
+</recommendation>
+
+<example>
+<p>
+The following example shows a case where it is checked whether a file inside the <code>/tmp/</code> folder exists,
```
mchammer01, line 27: To avoid passive voice
```suggestion
The following example shows a case where the code checks whether a file inside the <code>/tmp/</code> folder exists, 
```

javascript/ql/src/Security/CWE-367/FileSystemRace.qhelp
```diff
@@ -0,0 +1,61 @@
...
+<example>
+<p>
+The following example shows a case where it is checked whether a file inside the <code>/tmp/</code> folder exists, 
+and if it didn't exist the file is written to.
+</p>
+
+<sample src="examples/file-race-bad.js" />
+
+<p>
+However, in a multi-user environment the file might be created by another user in between the existence check and the write.
```
mchammer01, line 34:
```suggestion
However, in a multi-user environment the file might be created by another user between the existence check and the write. 
```

mchammer01 (approved):
@erik-krogh - LGTM ✨ 
Some minor comments for your consideration.

mchammer01 (approved):

javascript/ql/src/Security/CWE-367/FileSystemRace.qhelp
```diff
@@ -0,0 +1,61 @@
...
+This can be avoided by using <code>fs.open</code> to get a file descriptor, and then use that file descriptor in the write operation.
+</p>
+
+<sample src="examples/file-race-good.js" />
+
+</example>
+
+<references>
+<li>
+Wikipedia: <a href="https://en.wikipedia.org/wiki/Time-of-check_to_time-of-use">Time-of-check to time-of-use.</a>
```
esbena, line 47:
```suggestion
Wikipedia: <a href="https://en.wikipedia.org/wiki/Time-of-check_to_time-of-use">Time-of-check to time-of-use</a>.
```
(consistency with the other links below)

esbena (approved):

PR #7780, opened 2022-01-28
Ruby: Split up CI jobs

hmac:
Change the Ruby CI workflow to run the following jobs in parallel:
- Formatting
- Query compilation check
- Database upgrade check
- Tests (1/2)
- Test (2/2)

The overall effect is that Ruby CI takes ~10 minutes to complete, down from 15-20 minutes.

Other improvements that I haven't figured out yet:
- Cache and restore the CodeQL cache between runs
- Fix Cargo caching, which currently seems to be broken
- Cache the fetching of the CLI

### DIFF
--- a/.github/workflows/ruby-qltest.yml
+++ b/.github/workflows/ruby-qltest.yml
@@ -24,27 +24,45 @@ defaults:
     working-directory: ruby
 
 jobs:
-  qltest:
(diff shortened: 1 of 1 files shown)


aibaars (approved):
LGTM!

PR #7782, opened 2022-01-28
C++: Fix FPs for cpp/cleartext-storage-file

geoffw0:
Fix the FPs for `cpp/cleartext-storage-file` that we discovered on `kamailio/kamailio`, where the 'file' was in fact `stderr`.

The code I've used is simple compared to the exclusions in `cpp/cleartext-transmission`, but I found this simple solution worked well in practice.

Diff query: https://lgtm.com/query/1738869787752851711/

### DIFF
--- a/cpp/ql/src/Security/CWE/CWE-311/CleartextFileWrite.ql
+++ b/cpp/ql/src/Security/CWE/CWE-311/CleartextFileWrite.ql
@@ -65,6 +65,7 @@ where
   midNode.getNode().asExpr() = mid and
   mid = w.getASource() and
   dest = w.getDest() and
+  not dest.(VariableAccess).getTarget().getName() = ["stdin", "stdout", "stderr"] and // exclude calls with standard streams
   not isFileName(globalValueNumber(source)) and // file names are not passwords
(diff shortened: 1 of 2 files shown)


cpp/ql/src/Security/CWE/CWE-311/CleartextFileWrite.ql
```diff
@@ -65,6 +65,7 @@ where
   midNode.getNode().asExpr() = mid and
   mid = w.getASource() and
   dest = w.getDest() and
+  not dest.(VariableAccess).getTarget().getName() = ["stdin", "stdout", "stderr"] and // exclude calls with standard streams
```
MathiasVP, line 68: Could this exclusion be done in the `isSink` predicate instead? That will reduce the number of sinks, and thus result in better performance.

geoffw0: Not trivially / as cleanly, because the sink is the source of the `FileWrite` and this is the dest.  It could be done, if you have performance concerns.

MathiasVP: I don't really have performance concerns, no. The performance clearly won't be worse than what we have already. So let's just leave it as-is 👍.

MathiasVP (approved):

PR #7783, opened 2022-01-28
Python: promote LDAP injection query

yoff:
Some questions:
- Is the use of flow states acceptable or would you rather see two configurations?
- Should we rename the query to simply "LDAP injection"?


python/ql/lib/semmle/python/Concepts.qll
```diff
@@ -443,6 +443,41 @@ module RegexExecution {
   }
 }
 
+/**
```
RasmusWL, line 446: I'm thinking we should create an `Ldap` module that will contain all LDAP related concepts, like we've done for HTTP related concepts. I know that we have multiple experimental queries, so I think it could be worth the effort :+1:

python/ql/lib/semmle/python/Concepts.qll
```diff
@@ -526,6 +573,21 @@ class RegexEscaping extends Escaping {
...
+  LdapDnEscaping() { range.getKind() = Escaping::getLdapDnKind() }
+}
+
+/**
+ * An escape of a string so it can be safely used as a filter in an LDAP search.
+ */
+class LdapFilterEscaping extends Escaping {
+  LdapFilterEscaping() { range.getKind() = Escaping::getLdapFilterKind() }
+}
+
```
RasmusWL, line 590: If we do add such a `Ldap` module, I'm really at odds with whether we should move those ldap escaping classes in there (since they're clearly only ldap related), or whether we should keep them here close to the other escaping classes :thinking: I guess that since we have `HtmlEscaping` defined right here, we should not move them.

(keeping this note as a way to highlight this, but nothing to do for this code)

python/ql/lib/semmle/python/frameworks/Ldap.qll
```diff
@@ -0,0 +1,80 @@
...
+  }
+
+  /** The execution of an `ldap` query. */
+  private class LdapQueryExecution extends DataFlow::CallCfgNode, LdapExecution::Range {
+    LdapQueryExecution() {
+      this =
+        API::moduleImport("ldap")
+            .getMember("initialize")
+            .getReturn()
+            .getMember(ldapQueryMethodName())
```
RasmusWL, line 33:
```suggestion
            .getMember(["search", "search_s", "search_st", "search_ext", "search_ext_s"])
```

python/ql/lib/semmle/python/frameworks/Ldap.qll
```diff
@@ -0,0 +1,80 @@
...
+  /**
+   * The name of an `ldap` method used to execute a query.
+   *
+   * See https://www.python-ldap.org/en/python-ldap-3.3.0/reference/ldap.html#functions
+   */
+  private string ldapQueryMethodName() {
+    result in ["search", "search_s", "search_st", "search_ext", "search_ext_s"]
+  }
+
+  /** The execution of an `ldap` query. */
```
RasmusWL, line 26: No reason to keep this in a predicate, since it is only used once

```suggestion
  /**
   * The execution of an `ldap` query.
   *
   * See https://www.python-ldap.org/en/python-ldap-3.3.0/reference/ldap.html#functions
   */
```

python/ql/lib/semmle/python/frameworks/Ldap.qll
```diff
@@ -0,0 +1,80 @@
...
+
+    override DataFlow::Node getFilter() {
+      result in [this.getArg(2), this.getArgByName("filterstr")]
+    }
+
+    override DataFlow::Node getBaseDn() { result in [this.getArg(0), this.getArgByName("base")] }
+  }
+
+  /**
+   * A class to find calls to `ldap.dn.escape_dn_chars`.
```
RasmusWL, line 45:
```suggestion
   * A call to `ldap.dn.escape_dn_chars`.
```

python/ql/lib/semmle/python/frameworks/Ldap.qll
```diff
@@ -0,0 +1,80 @@
...
+
+    override DataFlow::Node getAnInput() { result in [this.getArg(0), this.getArgByName("s")] }
+
+    override DataFlow::Node getOutput() { result = this }
+
+    override string getKind() { result = Escaping::getLdapDnKind() }
+  }
+
+  /**
+   * A class to find calls to `ldap.filter.escape_filter_chars`.
```
RasmusWL, line 62:
```suggestion
   * A call to `ldap.filter.escape_filter_chars`.
```

python/ql/lib/semmle/python/frameworks/Ldap3.qll
```diff
@@ -0,0 +1,80 @@
...
+      result in [this.getArg(1), this.getArgByName("search_filter")]
+    }
+
+    override DataFlow::Node getBaseDn() {
+      result in [this.getArg(0), this.getArgByName("search_base")]
+    }
+  }
+
+  /**
+   * A class to find calls to `ldap3.utils.dn.escape_rdn`.
```
RasmusWL, line 38:
```suggestion
   * A call to `ldap3.utils.dn.escape_rdn`.
```

python/ql/lib/semmle/python/frameworks/Ldap3.qll
```diff
@@ -0,0 +1,80 @@
...
+
+    override DataFlow::Node getAnInput() { result in [this.getArg(0), this.getArgByName("rdn")] }
+
+    override DataFlow::Node getOutput() { result = this }
+
+    override string getKind() { result = Escaping::getLdapDnKind() }
+  }
+
+  /**
+   * A class to find calls to `ldap3.utils.conv.escape_filter_chars`.
```
RasmusWL, line 60:
```suggestion
   * A call to `ldap3.utils.conv.escape_filter_chars`.
```

python/ql/src/Security/CWE-090/LdapInjection.ql
```diff
@@ -5,13 +5,13 @@
  * @kind path-problem
  * @problem.severity error
```
RasmusWL, line 6: We need additional metadata. I think `@precision high` is fair, but we need to confirm this.

```suggestion
 * @problem.severity error
 * @security-severity 9.8
 * @precision high
```

`@security-severity 9.8` match [Java](https://github.com/github/codeql/blob/main/java/ql/src/Security/CWE/CWE-090/LdapInjection.ql) and [C#](https://github.com/github/codeql/blob/main/csharp/ql/src/Security%20Features/CWE-090/LDAPInjection.ql) version.

python/ql/lib/semmle/python/security/dataflow/LdapInjection.qll
```diff
@@ -0,0 +1,87 @@
...
+    // All additional flow steps signify a state change.
+    //    (n, `state`) --> (`node`, s)
+    // Thus, if a node in `state` transitions to `node` and a new state,
+    // then `state` should be blocked at `node`.
+    isAdditionalFlowStep(_, state, node, _)
+  }
+
+  override predicate isAdditionalFlowStep(
+    DataFlow::Node node1, DataFlow::FlowState state1, DataFlow::Node node2,
+    DataFlow::FlowState state2
```
RasmusWL, line 40: is there a reason you're not using the typical `nodeFrom` -> `nodeTo` naming?

RasmusWL (changes requested):
> The original configuration did not match sinks with sanitizers.
> Here it is resolved using flow state, it could also be done by using two configurations.

Good catch. I remember this bit, and thought it was called out in the original PR, but I couldn't see it anywhere. 

I think it would be a simpler setup with two configurations (if there were 3 different kinds, the state-transitions would become overly complex). 

## Other things

This is no `LdapInjectionCustomizations.qll` file, which should be added.

Please also update `docs/codeql/support/reusables/frameworks.rst`

> Should we rename the query to simply "LDAP injection"?

I don't understand what you're suggesting. That use that for the `@name` field? (I'm not against it, but it would also be nice to match what it's called in Java/C#)

python/ql/lib/semmle/python/Concepts.qll
yoff: I am happy to do that, it was like that in the external contribution, actually  :-)

python/ql/lib/semmle/python/frameworks/Ldap.qll
yoff: Hm, I thought I fixed that.

python/ql/lib/semmle/python/frameworks/Ldap.qll
yoff: Agreed.

python/ql/src/Security/CWE-090/LdapInjection.ql
yoff: Thanks for reminding me! It originally got the maximal false positive score, so `@precision high` should be correct :-)

python/ql/lib/semmle/python/security/dataflow/LdapInjection.qll
yoff: That comes from the data flow API. I would change it (but I will do a two-config version instead first).

python/ql/lib/semmle/python/Concepts.qll
yoff: Oh, I misunderstood. Yes, that is a good idea.

yoff: Done

python/ql/lib/semmle/python/Concepts.qll
```diff
@@ -443,38 +443,41 @@ module RegexExecution {
...
-  DataFlow::Node getFilter() { result = range.getFilter() }
-
-  /** Gets the argument containing the base DN. */
-  DataFlow::Node getBaseDn() { result = range.getBaseDn() }
-}
-
-/** Provides classes for modeling new LDAP query execution-related APIs. */
-module LdapExecution {
+/** Provides classes for modeling LDAP-related APIs. */
+module LDAP {
```
RasmusWL, line 447: nitpick: our style guide suggests that because this acronym is longer than 2 characters, it should be written as `Ldap`. I guess there is some value in having consistency between the existing `HTTP` module. I'm not really sure whether this is something we need to change...

python/ql/lib/semmle/python/security/dataflow/LdapInjection.qll
```diff
@@ -8,80 +8,42 @@ import semmle.python.dataflow.new.DataFlow
...
-class Safe extends DataFlow::FlowState {
-  Safe() { this = "Safe" }
+    override predicate isSanitizer(DataFlow::Node node) { node instanceof FilterSanitizer }
+
+    override predicate isSanitizerGuard(DataFlow::BarrierGuard guard) {
+      guard instanceof FilterSanitizerGuard
+    }
+  }
+
+  import DataFlow::PathGraph
```
RasmusWL, line 61: this part should be in the query itself, otherwise you will always expose the extra query predicates by importing this file :disappointed:

RasmusWL (changes requested):
Just a few minor things, then we should be good to go :+1:

python/ql/lib/semmle/python/Concepts.qll
yoff: Yeah, I modelled it after the HTTP module. I guess we can change both in a clean-up PR? Or are we creating deprecation work for ourselves that way?

python/ql/lib/semmle/python/security/dataflow/LdapInjection.qll
```diff
@@ -8,80 +8,42 @@ import semmle.python.dataflow.new.DataFlow
...
+
+  predicate ldapInjection(DataFlow::PathNode source, DataFlow::PathNode sink) {
+    any(DnConfiguration dnConfig).hasFlowPath(source, sink)
+    or
+    any(FilterConfiguration filterConfig).hasFlowPath(source, sink)
+  }
```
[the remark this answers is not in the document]
yoff: I think using the information is a good idea :-)

python/ql/lib/semmle/python/security/dataflow/LdapInjection.qll
yoff: I am not sure I understand this (that line was also in the query and superfluous here), but it has disappeared anyway with the rewrite proposed below.

python/ql/lib/semmle/python/Concepts.qll
RasmusWL: I think it's OK

docs/codeql/support/reusables/frameworks.rst
```diff
@@ -171,6 +171,8 @@ Python built-in support
    Twisted, Web framework
    Flask-Admin, Web framework
    starlette, Asynchronous Server Gateway Interface (ASGI)
+   ldap, Lightweight Directory Access Protocol (LDAP)
```
RasmusWL, line 174: Since we've listed `PyYAML` and not `yaml`, I guess we should continue listing the PyPI package name instead of the name used for importing in the Python code. We could probably do better to highlight what this is expected to be though :grimacing: 

```suggestion
   python-ldap, Lightweight Directory Access Protocol (LDAP)
```

RasmusWL (commented):
Found a really minor detail, otherwise LGTM :+1:

RasmusWL (approved):
Thanks :+1:

PR #7784, opened 2022-01-28
C#: Use .NET 6

hvitved:


csharp/extractor/Semmle.Extraction.CSharp/Entities/UserOperator.cs
```diff
@@ -178,6 +178,9 @@ public static bool OperatorSymbol(string methodName, out string operatorName)
         /// <returns>The converted name.</returns>
         private static string OperatorSymbol(Context cx, IMethodSymbol method)
         {
+            if (method.ExplicitInterfaceImplementations.Any())
+                return OperatorSymbol(cx, method.ExplicitInterfaceImplementations.First());
```
tamasvajk, line 182: Why is this change needed?

tamasvajk (approved):
looks plausible

csharp/extractor/Semmle.Extraction.CSharp/Entities/UserOperator.cs
hvitved: It has been an issue all along, but with the upgrade we ran into it in the tests, which we didn't before.

michaelnebel (approved):
Looks plausible to me!
I suppose the failing test are caused by a needed update in the semmle-code repo, which will be made in another PR?

hvitved:
> I suppose the failing test are caused by a needed update in the semmle-code repo, which will be made in another PR?

Yes.

PR #7785, opened 2022-01-28
Restrict AST nodes according to string length

Z80coder:
Fix for the catastrophic error identified in https://github.com/github/ml-ql-adaptive-threat-modeling/issues/1618

The catastrophic error is `CatastrophicError "String too long (5395415 characters)"`, which occurs when we tokenize the string associated with the `ASTNode`. I've added a restriction to throw away strings that are longer than 1,000,000 characters before we attempt to tokenize them.

### DIFF
--- a/javascript/ql/experimental/adaptivethreatmodeling/lib/experimental/adaptivethreatmodeling/FunctionBodyFeatures.qll
+++ b/javascript/ql/experimental/adaptivethreatmodeling/lib/experimental/adaptivethreatmodeling/FunctionBodyFeatures.qll
@@ -127,6 +127,7 @@ ASTNode getAnASTNodeWithAFeature(Function f) {
   result = getAnASTNodeToFeaturize(f)
 }
 
+/** Returns the number of source-code characters in a function.  */
 int getNumCharsInFunction(Function f) {
   result =
     strictsum(ASTNode node | node = getAnASTNodeWithAFeature(f) | getTokenizedAstNode(node).length())
@@ -134,10 +135,6 @@ int getNumCharsInFunction(Function f) {
 
 // Evaluator string limit is 5395415 characters. We choose a limit lower than this.
 private int getMaxChars() { result = 1000000 }
-
-Function getFeaturizableFunction(Function f) {
-  result = f and getNumCharsInFunction(f) <= getMaxChars()
-}
 
 /**
  * Returns a featurized representation of the function that can be used to populate the
@@ -147,13 +144,15 @@ string getBodyTokensFeature(Function function) {
   // Performance optimization: If a function has more than 256 body subtokens, then featurize it as
   // absent. This approximates the behavior of the classifer on non-generic body features where
   // large body features are replaced by the absent token.
+  //
+  // We count nodes instead of tokens because tokens are often not unique.
   strictcount(ASTNode node |
     node = getAnASTNodeToFeaturize(function) and
     exists(getTokenizedAstNode(node))
   ) <= 256 and
   // Performance optimization: If a function has more than getMaxChars() characters in its body subtokens,
   // then featurize it as absent.
-  function = getFeaturizableFunction(function) and
+  getNumCharsInFunction(function) <= getMaxChars() and
   result =
     strictconcat(Location l, string token |
       // The use of a nested exists here allows us to avoid duplicates due to two AST nodes in the
--- a/javascript/ql/experimental/adaptivethreatmodeling/lib/experimental/adaptivethreatmodeling/FunctionBodyFeatures.qll
+++ b/javascript/ql/experimental/adaptivethreatmodeling/lib/experimental/adaptivethreatmodeling/FunctionBodyFeatures.qll
@@ -133,7 +133,10 @@ int getNumCharsInFunction(Function f) {
     strictsum(ASTNode node | node = getAnASTNodeWithAFeature(f) | getTokenizedAstNode(node).length())
 }
 
-// Evaluator string limit is 5395415 characters. We choose a limit lower than this.
+/**
+ * The maximum number of characters a feature can be.
+ * The evaluator string limit is 5395415 characters. We choose a limit lower than this.
+ */
 private int getMaxChars() { result = 1000000 }
 
 /**
--- a/javascript/ql/experimental/adaptivethreatmodeling/lib/experimental/adaptivethreatmodeling/FunctionBodyFeatures.qll
+++ b/javascript/ql/experimental/adaptivethreatmodeling/lib/experimental/adaptivethreatmodeling/FunctionBodyFeatures.qll
@@ -128,17 +128,15 @@ ASTNode getAnASTNodeWithAFeature(Function f) {
 }
 
 int getNumCharsInFunction(Function f) {
-    result = strictsum(int i |
-      exists(ASTNode node | node = getAnASTNodeWithAFeature(f) and i = getTokenizedAstNode(node).length()) |
-      i
-    )
+  result =
+    strictsum(ASTNode node | node = getAnASTNodeWithAFeature(f) | getTokenizedAstNode(node).length())
 }
 
 // Evaluator string limit is 5395415 characters. We choose a limit lower than this.
 private int getMaxChars() { result = 1000000 }
 
 Function getFeaturizableFunction(Function f) {
-   result = f and getNumCharsInFunction(f) <= getMaxChars()
+  result = f and getNumCharsInFunction(f) <= getMaxChars()
 }
 
 /**
@@ -146,10 +144,11 @@ Function getFeaturizableFunction(Function f) {
  * `enclosingFunctionBody` feature for an endpoint.
  */
 string getBodyTokensFeature(Function function) {
-  // Performance optimization: If a function has more than getMaxChars() characters in its body subtokens, 
+  // Performance optimization: If a function has more than getMaxChars() characters in its body subtokens,
   // then featurize it as absent.
   function = getFeaturizableFunction(function) and
-  result = strictconcat(Location l, string token |
+  result =
+    strictconcat(Location l, string token |
       // The use of a nested exists here allows us to avoid duplicates due to two AST nodes in the
       // same location featurizing to the same token. By using a nested exists, we take only unique
       // (location, token) pairs.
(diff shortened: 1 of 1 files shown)


henrymercer (commented):
Have you verified that the extraction query now works on the relevant database?  When we concatenate the body tokens from each AST node together to form the full body token feature, we could still potentially reach the length limit.

From a brief dive into the code, the string limit may be 1,048,576 bytes.  Therefore a safer limit might be 1,048,576/(max num ast nodes in body feature) = 1,048,576/1024 = 1,024 bytes / characters.  We'd need to subtract one from that for the space separator, so perhaps a 1000 character limit is sensible?

Z80coder:
Sorry this is a draft, and I shouldn't have asked you to take a look yet.

No this does not solve the problem at all.

The difficulty, seems to be, that the size restriction in `getBodyTokenFeature`, which is currently `256`, isn't strict enough. Lowering that to, say, `32`, avoids this problem.

Perhaps I need to implement a separate predicate that somehow counts the total number of characters in all the `ASTNode`s of a `Function` and then chooses the node restriction appropriately?

Z80coder:
@henrymercer I've tested that this change now fixes the problem. However, it may have other knock-ons that I don't understand. Ready for review now.

javascript/ql/experimental/adaptivethreatmodeling/lib/experimental/adaptivethreatmodeling/FunctionBodyFeatures.qll
```diff
@@ -127,22 +127,29 @@ ASTNode getAnASTNodeWithAFeature(Function f) {
   result = getAnASTNodeToFeaturize(f)
 }
 
+int getNumCharsInFunction(Function f) {
+    result = strictsum(int i |
+      exists(ASTNode node | node = getAnASTNodeWithAFeature(f) and i = getTokenizedAstNode(node).length()) |
+      i
+    )
```
henrymercer, line 134: This sums all integers i such that there exists a tokenized AST node of that length.  This will give a result of 2 instead of 4 if we have for instance two AST nodes of length 2.  Instead we should sum over the AST nodes, i.e. `strictsum(ASTNode node | node = getAnASTNodeWithAFeature(f) | getTokenizedAstNode(node).length())`.

javascript/ql/experimental/adaptivethreatmodeling/lib/experimental/adaptivethreatmodeling/FunctionBodyFeatures.qll
```diff
@@ -127,22 +127,29 @@ ASTNode getAnASTNodeWithAFeature(Function f) {
   result = getAnASTNodeToFeaturize(f)
 }
 
+int getNumCharsInFunction(Function f) {
```
henrymercer, line 130: This needs autoformatting.

javascript/ql/experimental/adaptivethreatmodeling/lib/experimental/adaptivethreatmodeling/FunctionBodyFeatures.qll
```diff
@@ -127,22 +127,29 @@ ASTNode getAnASTNodeWithAFeature(Function f) {
...
-  strictcount(ASTNode node |
-    node = getAnASTNodeToFeaturize(function) and
-    exists(getTokenizedAstNode(node))
-  ) <= 256 and
-  result =
-    strictconcat(Location l, string token |
+  // Performance optimization: If a function has more than getMaxChars() characters in its body subtokens, 
+  // then featurize it as absent.
+  function = getFeaturizableFunction(function) and
+  result = strictconcat(Location l, string token |
```
henrymercer, line 152: This changes the behaviour of the function body features potentially quite considerably.  The AST node limit was a factor that noticeably affected end-to-end performance, and using a different restriction here could have important implications on the runtime of the ML-powered queries.

We need to validate the runtime implications of this change before we merge it.  Furthermore, we should either (a) stick with the removal of the AST node limit, train a new model on the new features and run an end-to-end evaluation, or (b) reintrodue the AST node limit, so the feature values should not change considerably.

javascript/ql/experimental/adaptivethreatmodeling/lib/experimental/adaptivethreatmodeling/FunctionBodyFeatures.qll
Z80coder: Oof .. thanks for the explanation. Will fix.

javascript/ql/experimental/adaptivethreatmodeling/lib/experimental/adaptivethreatmodeling/FunctionBodyFeatures.qll
Z80coder: - Will reinstate the AST node limit
- Will collect timing measures
- Will run end-to-end evaluation

Z80coder:
End-to-end evaluation results are [here](https://github.com/github/ml-ql-adaptive-threat-modeling-backend/pull/1185).

There is essentially no change _except_ for an increase in the number of results to recover for Path Injection. I suspect, but have not verified, that this might be due to the string length restriction resulting in fewer databases failing.

Z80coder:
End-to-end results are here:
https://github.com/github/ml-ql-adaptive-threat-modeling-backend/pull/1185
(this time against a properly worsened comparison branch). Summary:
- `nop` for `sqlinjection`, `nosqlinjection` and `xss`
- `pathinjection` has an increase in the number of results that can be recovered (compared to main), which has the knock-on that `recall` drops by about `8.58`% (and other metrics similarly affected).

This PR should have 2 effects:
- catastrophic failures on some databases should no longer occur
- some `enclosingFunctionBody` feature values will be the empty string (rather than a very long string).

More detective work is needed to understand the cause of the increase in recoverable alerts in the end-to-end metrics.

Z80coder:
Some timing tests. Conclusion: this change probably has no effect on query execution time.

N.B. Before each run I brutally cleared database caches and pre-compiled the queries.

<b style="font-weight:normal;" id="docs-internal-guid-b7b4b682-7fff-8f90-2bea-5585e8328ff1"><div dir="ltr" style="margin-left:0pt;" align="left">

repo | With change | Without change
-- | -- | --
willwillis_Crypto-Sentiment-Clusters_8830868/ | 49.026s | 49.610s
atm-query-suite-owasp-juice-shop | 1m13.680s | 1m12.604s
A2Z-F15 | 17m27.202s | 17m28.127s

</div></b>

Z80coder:
> More detective work is needed to understand the cause of the increase in recoverable alerts in the end-to-end metrics.

The difference is due to an improvement in modeling unrelated to this PR. See https://github.com/github/ml-ql-adaptive-threat-modeling-backend/pull/1185

Z80coder:
@henrymercer I've reinstated the AST node limit, collect timing measures (no effect) and performed end-to-end evaluation (no effect).

javascript/ql/experimental/adaptivethreatmodeling/lib/experimental/adaptivethreatmodeling/FunctionBodyFeatures.qll
```diff
@@ -127,6 +127,18 @@ ASTNode getAnASTNodeWithAFeature(Function f) {
   result = getAnASTNodeToFeaturize(f)
 }
 
+int getNumCharsInFunction(Function f) {
```
henrymercer, line 130: Add QLDoc

javascript/ql/experimental/adaptivethreatmodeling/lib/experimental/adaptivethreatmodeling/FunctionBodyFeatures.qll
```diff
@@ -127,6 +127,18 @@ ASTNode getAnASTNodeWithAFeature(Function f) {
...
+  result =
+    strictsum(ASTNode node | node = getAnASTNodeWithAFeature(f) | getTokenizedAstNode(node).length())
+}
+
+// Evaluator string limit is 5395415 characters. We choose a limit lower than this.
```
henrymercer, line 135:
```suggestion
/**
 * The maximum number of characters a feature can be.
 * The evaluator string limit is 5395415 characters. We choose a limit lower than this.
 */
```

javascript/ql/experimental/adaptivethreatmodeling/lib/experimental/adaptivethreatmodeling/FunctionBodyFeatures.qll
```diff
@@ -127,6 +127,18 @@ ASTNode getAnASTNodeWithAFeature(Function f) {
...
+private int getMaxChars() { result = 1000000 }
+
+Function getFeaturizableFunction(Function f) {
```
henrymercer, line 138: The structure of this predicate is a bit confusing.  I suggest converting to a `predicate isFunctionFeaturizable(Function f)` predicate instead, or (perhaps better) just inlining it.

javascript/ql/experimental/adaptivethreatmodeling/lib/experimental/adaptivethreatmodeling/FunctionBodyFeatures.qll
```diff
@@ -135,12 +147,13 @@ string getBodyTokensFeature(Function function) {
   // Performance optimization: If a function has more than 256 body subtokens, then featurize it as
   // absent. This approximates the behavior of the classifer on non-generic body features where
   // large body features are replaced by the absent token.
-  //
-  // We count nodes instead of tokens because tokens are often not unique.
```
henrymercer, line 139: Did you intend to remove this?

henrymercer (commented):
Thanks, this looks good overall.  Minor suggestions to improve the code, then let's get this in.

javascript/ql/experimental/adaptivethreatmodeling/lib/experimental/adaptivethreatmodeling/FunctionBodyFeatures.qll
Z80coder: No.

javascript/ql/experimental/adaptivethreatmodeling/lib/experimental/adaptivethreatmodeling/FunctionBodyFeatures.qll
Z80coder: Will inline.

javascript/ql/experimental/adaptivethreatmodeling/lib/experimental/adaptivethreatmodeling/FunctionBodyFeatures.qll
```diff
@@ -127,6 +127,18 @@ ASTNode getAnASTNodeWithAFeature(Function f) {
   result = getAnASTNodeToFeaturize(f)
 }
 
+/** Returns the number of source-code characters in a function.  */
```
henrymercer, line 130: Autoformat via code review:

```suggestion
/** Returns the number of source-code characters in a function. */
```

henrymercer (approved):

PR #7787, opened 2022-01-30
C/C++ : Useless test

Yonah125:
This query finds any useless test of kind a==8 && a!=7.

This query found an useless test on the Suricata project that can you see in this Pull Request : 

https://github.com/OISF/suricata/pull/6897

### DIFF
--- a/cpp/ql/src/experimental/Best Practices/UselessTest.qhelp
+++ b/cpp/ql/src/experimental/Best Practices/UselessTest.qhelp
@@ -1,0 +1,19 @@
+<!DOCTYPE qhelp PUBLIC
+  "-//Semmle//qhelp//EN"
+  "qhelp.dtd">
+<qhelp>
+
+<overview>
+<p>Comparisons operation like <code>a==8 && a!=7</code> contains an useless part : the non-equal part. This rule finds any test of this kind within an if or a while statement
+This rule will only find useless comparisons with a right literal operand. </p>
+</overview>
+
+<recommendation>
+<p>Remove the useless comparisons</p>
+</recommendation>
+
+<example>
+<sample src="UselessTest.cpp" />
+</example>
+
+</qhelp>
--- a/cpp/ql/src/experimental/Best Practices/UselessTest.qhelp
+++ b/cpp/ql/src/experimental/Best Practices/UselessTest.qhelp
@@ -4,8 +4,7 @@
 <qhelp>
 
 <overview>
-<p>Comparisons operation like <code>a==8 && a!=7</code> contains an useless part : the non-equal part. This rule finds any test of this kind within an if or a while statement
-This rule will only find useless comparisons with a right literal operand. </p>
+<p>Comparison operations like <code>a==8 &amp;&amp; a!=7</code> contain a useless part : the non-equal part. This rule finds tests of this kind within an <code>if</code> or a <code>while</code> statement</p>
 </overview>
 
 <recommendation>
--- a/cpp/ql/src/experimental/Best Practices/UselessTest.ql
+++ b/cpp/ql/src/experimental/Best Practices/UselessTest.ql
@@ -40,4 +40,4 @@ where
   ne.getRightOperand() instanceof Literal and
   nearestParent(eq) = nearestParent(ne) and
   sameExpr(eq.getLeftOperand(), ne.getLeftOperand())
-select  ne,"Useless Test"
+select ne, "Useless Test"
--- a/cpp/ql/src/experimental/Best Practices/UselessTest.ql
+++ b/cpp/ql/src/experimental/Best Practices/UselessTest.ql
@@ -1,0 +1,50 @@
+/**
+ * @name Useless Test
+ * @description Find any useless test of kind a==8 && a!=7
+ * @kind problem
+ * @problem.severity warning
+ * @id cpp/uselesstest
+ * @tags reliability
+ *       readability
+ */
+
+import cpp
+
+predicate sameStmt(Expr e1, Expr e2) {
+  e1.getNumChild() = e2.getNumChild() and
+  e1.toString() = e2.toString() and
+  (
+    e1.getNumChild() = 0
+    or
+    forall(int i | i in [0 .. e1.getNumChild() - 1] | sameStmt(e1.getChild(i), e2.getChild(i)))
+  )
+}
+
+Element nearestParent(Expr e) {
+  if
+    e.getParent().(Expr).getFullyConverted() instanceof ParenthesisExpr or
+    e.getParent() instanceof IfStmt or
+    e.getParent() instanceof WhileStmt
+  then result = e.getParent()
+  else result = nearestParent(e.getParent())
+}
+
+from LogicalAndExpr b, EQExpr eq, NEExpr ne
+where
+  (
+    b.getAChild*() = eq and
+    b.getAChild*() = ne and
+    eq.getParent() instanceof LogicalAndExpr and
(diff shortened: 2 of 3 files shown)


catenacyber:
cc @HauwaOtori : this is a first PR from my students project aka CodeQL versus Suricata
cc @xcorail

PS @Yonah125 do you want to co-author commits with your comrades ?

cpp/ql/src/experimental/Best Practices/UselessTest.ql
```diff
@@ -0,0 +1,50 @@
...
+ * @kind problem
+ * @problem.severity warning
+ * @id cpp/uselesstest
+ * @tags reliability
+ *       readability
+ */
+
+import cpp
+
+predicate sameStmt(Expr e1, Expr e2) {
```
MathiasVP, line 13: I'm not quite sure I understand the meaning of this predicate. The name makes it sound like you want to check if two statements are equivalent, but the two parameters are of type `Expr` (and not `Stmt`)?

If what you want is to compare two expressions, I'd consider trying out our global value numbering library (by importing `semmle.code.cpp.valuenumbering.GlobalValueNumbering`). Then you can check if two expressions are guarenteed to evaluate to the same value by check if `globalValueNumber(e1).getAnExpr() = e2` holds. We also have a hash cons library which is a more syntactic version of global value numbering. You can read up on both libraries here: https://codeql.github.com/docs/codeql-language-guides/hash-consing-and-value-numbering/

cpp/ql/src/experimental/Best Practices/UselessTest.ql
```diff
@@ -0,0 +1,50 @@
...
+  e1.getNumChild() = e2.getNumChild() and
+  e1.toString() = e2.toString() and
```
MathiasVP, line 15: You shouldn't be using the value produced by `toString` in your query logic as we don't guarantees any backwards compatibility on the strings produced by `toString` (i.e., we might silently change the format of the string in the future and thus break your query logic).

cpp/ql/src/experimental/Best Practices/UselessTest.ql
```diff
@@ -0,0 +1,50 @@
...
+from LogicalAndExpr b, EQExpr eq, NEExpr ne
+where
+  (
+    b.getAChild*() = eq and
+    b.getAChild*() = ne and
+    eq.getParent() instanceof LogicalAndExpr and
+    ne.getParent() instanceof LogicalAndExpr
+  ) and
+  (
+    eq.getChild(0) instanceof VariableAccess and ne.getChild(0) instanceof VariableAccess
```
MathiasVP, line 41: I'd avoid using `getChild` whenever there's a more convenient predicate available. For instance, `eq.getChild(0)` is equivalent to `eq.getLeftOperand()`m and `eq.getChild(1)` is equivalent to `eq.getRightOperand()`. The same holds for the uses of `ne.getChild`.

cpp/ql/src/experimental/Best Practices/UselessTest.qhelp
```diff
@@ -0,0 +1,19 @@
+<!DOCTYPE qhelp PUBLIC
+  "-//Semmle//qhelp//EN"
+  "qhelp.dtd">
+<qhelp>
+
+<overview>
+<p>Comparisons operation like <code>a==8 && a!=7</code> contains an useless part : the non-equal part. This rule finds any test of this kind within an if or a while statement
```
MathiasVP, line 7: A couple of grammar fixes :) Also, our PR checks are complaining because of the non-escaped `&&`, so I've replaced it with `&amp;&amp;`.
```suggestion
<p>Comparison operations like <code>a==8 &amp;&amp; a!=7</code> contain a useless part : the non-equal part. This rule finds tests of this kind within an <code>if</code> or a <code>while</code> statement
```

cpp/ql/src/experimental/Best Practices/UselessTest.ql
```diff
@@ -0,0 +1,50 @@
+/**
+ * @name Useless Test
+ * @description Find any useless test of kind a==8 && a!=7
```
MathiasVP, line 3: Code must be enclosed in single quotes. See https://github.com/github/codeql/blob/main/docs/query-metadata-style-guide.md#query-descriptions-description.

Also, ideally the query description should describe the problem that this query is trying to highlight, and not what the query is doing. So I'd really rephrase this as something like:
```
A boolean condition that is guaranteed to never be evaluated should be deleted.
```

MathiasVP (commented):
Thanks for your contribution @Yonah125,

Here are my initial comments. Note that CI is unhappy because the query isn't hasn't been auto-formatted (see the `Formatting` section in https://github.com/github/codeql/blob/main/CONTRIBUTING.md#submitting-a-new-experimental-query).

None of my comments are really blocking this PR: It's mostly to help you learn to be efficient writing CodeQL queries :) If you're happy with the results you're finding with the query I'll merge it as soon as the PR is green.

MathiasVP:
> About the auto-formatting, I used the 'Format Document' command in VSC, is the output bad ?

Excellent. That should all be fine now, then. Let's see what CI thinks about it now 🤞.

cpp/ql/src/experimental/Best Practices/UselessTest.ql
```diff
@@ -0,0 +1,43 @@
...
+    eq.getLeftOperand() instanceof VariableAccess and ne.getLeftOperand() instanceof VariableAccess
+    or
+    eq.getLeftOperand() instanceof PointerDereferenceExpr and
+    ne.getLeftOperand() instanceof PointerDereferenceExpr
+  ) and
+  eq.getRightOperand() instanceof Literal and
+  ne.getRightOperand() instanceof Literal and
+  nearestParent(eq) = nearestParent(ne) and
+  sameExpr(eq.getLeftOperand(), ne.getLeftOperand())
+select "Useless test", ne
```
MathiasVP, line 43: The PR check is failing because I missed an error in your query :)

Since this is a `@kind problem` query, your `select` clause must be of the form `select element, "string"` (instead of `select "string", element` like you have currently). Check out https://codeql.github.com/docs/writing-codeql-queries/about-codeql-queries/#select-clause for more info on this. This edit should make CI happy:

```suggestion
select ne, "Useless test"
```

cpp/ql/src/experimental/Best Practices/UselessTest.ql
```diff
@@ -0,0 +1,43 @@
...
+ */
+
+import cpp
+import semmle.code.cpp.valuenumbering.GlobalValueNumbering
+
+predicate sameExpr(Expr e1, Expr e2) { globalValueNumber(e1).getAnExpr() = e2 }
+
+Element nearestParent(Expr e) {
+  if
+    e.getParent().(Expr).getFullyConverted() instanceof ParenthesisExpr or
```
rdmarsh2, line 18: I think you want `getConverted*()` rather than `getFullyConverted` - if there's an explicit cast of a parenthesized expression, like `(short)(a + b)`, `getFullyConverted()` will be the cast rather than the parenthesis.

catenacyber:
CI looks green now 👏

MathiasVP:
> CI looks green now 👏

Indeed, it is 🎉!

@catenacyber has this been resolved:

> PS [@Yonah125](https://github.com/Yonah125) do you want to co-author commits with your comrades ?

Because I think we're ready to merge this PR once you're all happy with the commit history.

catenacyber:
Commit history is good enough for me :-)

MathiasVP (approved):
LGTM!

PR #7788, opened 2022-01-31
python: remove library annotation to clean up QL warnings.

yoff:
Should put these functions in a private module instead, as indicated by the documentation for `library`?

### DIFF
--- a/python/ql/lib/semmle/python/AstExtended.qll
+++ b/python/ql/lib/semmle/python/AstExtended.qll
@@ -62,33 +62,33 @@ abstract class AstNode extends AstNode_ {
 
 /* Parents */
 /** Internal implementation class */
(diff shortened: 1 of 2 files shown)


RasmusWL:
Maybe this should be obvious, but what warnings are we getting for these? (and why?)

tausbn:
> Maybe this should be obvious, but what warnings are we getting for these? (and why?)

There's a QL-for-QL query that checks for this presence of this annotation. The annotation itself is deprecated (though not to the degree that it actually emits a compiler warning).

yoff:
I made it also remove the auto-generated annotations. There is a PR to not auto-generate those.

tausbn (approved):
Looks good to me. We might want to also fix up the QLDoc to get rid of those warnings.

PR #7791, opened 2022-01-31
Data flow: Inline `local(Expr|Instruction)?(Flow|Taint)`

hvitved:
Computing a full transitive closure is often bad; by inlining all calls we are providing more context to the QL optimizer.


michaelnebel (approved):
Looks plausible to me!

aschackmull (approved):
LGTM

PR #7792, opened 2022-01-31
C#: Attribute kind and return value attributes.

michaelnebel:
In this PR we add support for return value attributes.

- Return value attributes are now extracted.
- Attributes will have different kind (so far Default, Return, Assembly and Module).
- [Microsoft documentation](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/language-specification/attributes#2124-attribute-parameter-types) of attributes.

In most cases, there is no ambiguity if the target specifier is omitted, when declaring an attribute, but there are some exceptions, which is why the notion of *kind* on an attribute is being introduced.
- For global attribute declarations as *target* (in this case directly translated to a kind) needs to be specified.
- For return value attributes, we need to be able to distinguish between the attribute *target*. Eg. for methods we need to distinguish whether an attribute has the method itself as target or the return value.

Closes https://github.com/github/codeql-csharp-team/issues/219

In this PR, we still need to
- [x] Run DCA to generate stats.
- [x] Run DCA validate that there are no performance regressions.


michaelnebel:
Closes https://github.com/github/codeql-csharp-team/issues/219

michaelnebel:
In this PR, we still need to
- [ ] Run DCA to generate stats
- [ ] Run DCA validate that there are no performance regressions.

tamasvajk (commented):
LGTM, let's see what DCA says.

michaelnebel:
> LGTM, let's see what DCA says.

Alright - good - thank you!

michaelnebel:
Below is the performance executions numbers on the newly generated stats file.
Not sure, whether we think this is within the statistical uncertainty as four projects are 3-5% slower. On the other hand, the overall only shows a degredation of 1%, which is within statistical uncertainty.
image

Should we run it again and see what happens or do we accept these numbers?

csharp/extractor/Semmle.Extraction.CSharp/Tuples.cs
```diff
@@ -32,8 +32,8 @@ internal static class Tuples
         internal static void array_element_type(this TextWriter trapFile, ArrayType array, int dimension, int rank, Type elementType) =>
             trapFile.WriteTuple("array_element_type", array, dimension, rank, elementType);
 
-        internal static void attributes(this TextWriter trapFile, Attribute attribute, Type attributeType, IEntity entity) =>
-            trapFile.WriteTuple("attributes", attribute, attributeType, entity);
+        internal static void attributes(this TextWriter trapFile, Attribute attribute, AttributeKind kind, Type attributeType, IEntity entity) =>
+            trapFile.WriteTuple("attributes", attribute, (int)kind, attributeType, entity);
```
hvitved, line 36: The `int` cast is not strictly necessary.

csharp/ql/lib/semmle/code/csharp/Attribute.qll
```diff
@@ -97,3 +97,45 @@ class Attribute extends TopLevelExprParent, @attribute {
...
+
+/**
+ * An attribute with return kind, for example `[...]` on line 1 in
+ * ```csharp
+ * [return: MyAttribute(0)]
+ * int SomeMethod() { return 1; }
+ * ```
+ */
+class ReturnAttribute extends Attribute, @attribute_return {
+  override string getAPrimaryQlClass() { result = "ReturnAttribute" }
```
hvitved, line 120: Lets add a top-level predicate
```ql
private string getAttributeName(Attribute a) {
  exists(string type | type = a.getType().getName() |
    if type.matches("%Attribute") then result = type.prefix(type.length() - 9) else result = type
  )
}
```

replace the body of `Attribute::toString` with

```ql
result = "[" + getAttributeName(this) + "(...)]"
```

and add an override here

```ql
override string toString() {
  result = "[return: " + getAttributeName(this) + "(...)]"
}
```

csharp/ql/lib/semmle/code/csharp/Attribute.qll
```diff
@@ -97,3 +97,45 @@ class Attribute extends TopLevelExprParent, @attribute {
...
+}
+
+/**
+ * An attribute with assembly kind, for example `[...]` on line 1 in
+ * ```csharp
+ * [assembly: MyAttribute(0)]
+ * ```
+ */
+class AssemblyAttribute extends Attribute, @attribute_assembly {
+  override string getAPrimaryQlClass() { result = "AssemblyAttribute" }
```
hvitved, line 130: Add

```ql
override string toString() {
  result = "[assembly: " + getAttributeName(this) + "(...)]"
}
```

csharp/ql/lib/semmle/code/csharp/Attribute.qll
```diff
@@ -97,3 +97,45 @@ class Attribute extends TopLevelExprParent, @attribute {
...
+}
+
+/**
+ * An attribute with module kind, for example `[...]` on line 1 in
+ * ```csharp
+ * [module: MyAttribute(0)]
+ * ```
+ */
+class ModuleAttribute extends Attribute, @attribute_module {
+  override string getAPrimaryQlClass() { result = "ModuleAttribute" }
```
hvitved, line 140: Add

```ql
override string toString() {
  result = "[module: " + getAttributeName(this) + "(...)]"
}
```

hvitved (changes requested):
Looks great to me. Some minor comments.

hvitved:
> Below is the performance executions numbers on the newly generated stats file.
> Not sure, whether we think this is within the statistical uncertainty as four projects are 3-5% slower. On the other hand, the overall only shows a degredation of 1%, which is within statistical uncertainty.

That is well within the statistical uncertainty, so it looks all good.

csharp/extractor/Semmle.Extraction.CSharp/Entities/Attribute.cs
```diff
@@ -125,26 +137,36 @@ private void ExtractArguments(TextWriter trapFile)
 
         public override bool NeedsPopulation => true;
 
+        private static void ExtractAttributes<T>(Context cx, T symbol, Func<T, IEnumerable<AttributeData>> getAttributes, IEntity entity, AttributeKind kind) where T : ISymbol
```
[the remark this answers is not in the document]
michaelnebel: Ah yes, the final solution ended up being so simple that this is not needed. Will change. Good catch!

csharp/extractor/Semmle.Extraction.CSharp/Populators/TypeContainerVisitor.cs
```diff
@@ -80,17 +80,26 @@ private void ExtractTypeDeclaration(BaseTypeDeclarationSyntax node)
             Entities.Type.Create(Cx, Cx.GetModel(node).GetDeclaredSymbol(node)).ExtractRecursive(TrapFile, Parent);
         }
 
+        private static Entities.AttributeKind ExtractGlobalTarget(AttributeListSyntax node) =>
```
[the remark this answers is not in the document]
michaelnebel: Alright. Will inline.

csharp/ql/lib/upgrades/ba2201248071b2bf0bb52909b35014091d2e18a6/attribute_kind.ql
```diff
@@ -0,0 +1,16 @@
...
+
+class Attributable extends @attributable {
+  string toString() { none() }
+}
+
+class TypeOrRef extends @type_or_ref {
+  string toString() { none() }
+}
+
+query predicate add_default_kind(Attribute id, int kind, TypeOrRef type_id, Attributable target) {
```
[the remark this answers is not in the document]
michaelnebel: hah, I was wondering the same thing, but thought it was safe, since the CI didn't complain.
Maybe the predicate has to be named *select* or something similar :-)
In any case, I will change it according to your recommendation!

csharp/extractor/Semmle.Extraction.CSharp/Tuples.cs
michaelnebel: Ah yes - it looks like the WriteColumn extensions method does the cast for us.
Will fix.

csharp/ql/lib/semmle/code/csharp/Attribute.qll
michaelnebel: Yes, that is a good idea!

hvitved (dismissed):

hvitved (approved):

PR #7794, opened 2022-01-31
Sign analysis for C++

dbartol:


java/ql/lib/semmle/code/java/dataflow/internal/rangeanalysis/SignAnalysisSpecific.qll
```diff
@@ -216,6 +212,10 @@ private module Impl {
   predicate numericExprWithUnknownSign(Expr e) {
     // The expression types handled in the predicate complements the expression
     // types handled in `specificSubExprSign`.
+    exists(IntegerLiteral lit | lit = e and not exists(lit.getValue().toInt()))
```
intrigus-lgtm, line 215: How is it possible for an integer literal to not have a value?

cpp/ql/lib/semmle/code/cpp/ir/rangeanalysis/internal/SsaReadPositionSpecific.qll
```diff
@@ -0,0 +1,32 @@
+private import semmle.code.cpp.ir.IR as IR
+
+class BasicBlock = IR::IRBlock;
+
+private newtype TSsaVariable = MkSsaVariable(IR::Instruction instr)
```
MathiasVP, line 5: This IPA doesn't seem to be used? Was `SsaVariable` supposed to extend it? (It looks like it's not actually necessary to have this type at all!)

MathiasVP (commented):
I've only given this a very superficial review so far, but it looks great!

PR #7796, opened 2022-01-31
C++: replace ReturnValue nodes in flow paths

rdmarsh2:
Improves data flow path readability for C++ by hiding the per-function `ReturnValue` nodes and forcing the nodes created for each `ReturnStmt` to be included in the displayed path. This also adds a new `nodeIsShown` predicate to the data flow library's per-language interface.

### DIFF
--- a/cpp/ql/lib/semmle/code/cpp/ir/dataflow/internal/DataFlowImpl.qll
+++ b/cpp/ql/lib/semmle/code/cpp/ir/dataflow/internal/DataFlowImpl.qll
@@ -1698,7 +1698,8 @@ private module LocalFlowBigStep {
   private class FlowCheckNode extends NodeEx {
     FlowCheckNode() {
       castNode(this.asNode()) or
-      clearsContentCached(this.asNode(), _)
+      clearsContentCached(this.asNode(), _) or
+      nodeIsShown(this.asNode())
     }
   }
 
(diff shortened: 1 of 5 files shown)


rdmarsh2:
Opened as draft so I can do performance evaluations and get input from the data flow library maintainers.

MathiasVP:
We _could_ hack this without changing the shared library by repurposing the `CastNode` class. However, I do like having a `Please show this node in the final path` feature that's more explicit than just marking everything as cast nodes when debugging paths 👍.

aschackmull:
Ah, right, yeah I see the problem you're trying to solve. I think we ought to solve this entirely within the shared library without the need for a new interface predicate.

aschackmull:
Do you envision other use-cases for a "Please show this node in the final path"-feature? Because I'm leaning towards considering the specific short-coming around synthesised return nodes as a bug in the shared library, which can be fixed without such a feature.

MathiasVP:
> Do you envision other use-cases for a "Please show this node in the final path"-feature? Because I'm leaning towards considering the specific short-coming around synthesised return nodes as a bug in the shared library, which can be fixed without such a feature.

The return node use-case is the only one I've encountered so far, at least. Can you think of other use-cases @rdmarsh2?

rdmarsh2:
I didn't envision any other use-cases. If you can make a shared library fix for it, that's better than adding a new predicate to the interface.

PR #7797, opened 2022-02-01
Ruby: Pin Rust to 1.54

hmac:
Add a `rust-toolchain` file to the Ruby directory, which instructs Rustup
to install a specific version of Rust (1.54). This will be used in CI,
so any use of language features or dependencies that don't support 1.54
will result in a CI failure.

This should ensure we have a documented minimum supported rust version
and an easy to way to update it in the future (update the rust-toolchain
file).

In addition to this, we may want to consider reverting the bump of `clap` and then dropping the version to 1.53.


ruby/rust-toolchain
```diff
@@ -0,0 +1,6 @@
+# This file specifies the Rust version used to develop and test the Ruby
```
aibaars, line 1: Shall we name the file `rust-toolchain.toml` . Both `rust-toolchain` and `rust-toolchain.toml` are valid, but I think with extension is preferable .

aibaars:
In the future we could also use https://doc.rust-lang.org/nightly/cargo/reference/manifest.html#the-rust-version-field

ruby/rust-toolchain
```diff
@@ -0,0 +1,6 @@
...
+# extractor. It is set to the lowest version of Rust we want to support.
+
+[toolchain]
```
aibaars, line 4: Let's also include this file in the cache-key in the following files https://github.com/github/codeql/pull/7799/commits/6451a71a7845b49b83802b8b2de370fbfed24f30

aibaars (approved):
These changes look good to me. On my laptop changing the `rust-toolchain` causes `cargo` to automatically download a new toolchain if the version is not present. Therefore, I don't think there is a need to revert back to the 1.53 version.

nickrolfe:
> I don't think there is a need to revert back to the 1.53 version.

You're probably right – I guess all the Ubuntu users who were on the 1.53 version from `apt` have now taken the hit of installing rustup. But let's keep an eye on when `apt` starts providing 1.54 or newer, so we can tell the Ubuntu users they can switch back, if they want.

ruby/rust-toolchain
hmac: Good point. I've pushed a commit to test that in CI, and if all is well then :+1:

aibaars (dismissed):

ruby/actions/create-extractor-pack/action.yml
```diff
@@ -9,7 +9,7 @@ runs:
           ~/.cargo/registry
           ~/.cargo/git
           ruby/target
-        key: ${{ runner.os }}-ruby-qltest-cargo-${{ hashFiles('ruby/**/Cargo.lock') }}
+        key: ${{ runner.os }}-ruby-qltest-${{hashFiles('ruby/rust-toolchain.toml')}}-cargo-${{ hashFiles('ruby/**/Cargo.lock') }}
```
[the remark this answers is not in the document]
hmac: Good point, I think that's slightly nicer.

aibaars (approved):

PR #7798, opened 2022-02-01
C++: Add query for missing mode argument in `open`/`openat` calls

jketema:
When calling `open`/`openat` with the `O_CREAT` flag, the mode argument must be supplied. The query introduced in this PR checks for this.

The meta-data on the query is wrong. Some guidelines on getting this correct are appreciated.

### DIFF
--- a/cpp/ql/lib/change-notes/2022-02-02-detect-creat-in-openat.md
+++ b/cpp/ql/lib/change-notes/2022-02-02-detect-creat-in-openat.md
@@ -1,4 +1,4 @@
 ---
 category: minorAnalysis
 ---
-* he `cpp/world-writable-file-creation` query now only detects `open` and `openat` calls with the `O_CREAT` or `O_TMPFILE` flag.
+* The `cpp/world-writable-file-creation` query now only detects `open` and `openat` calls with the `O_CREAT` or `O_TMPFILE` flag.
--- a/cpp/ql/lib/change-notes/2022-02-02-detect-creat-in-openat.md
+++ b/cpp/ql/lib/change-notes/2022-02-02-detect-creat-in-openat.md
@@ -1,4 +1,4 @@
 ---
 category: minorAnalysis
 ---
-* The `cpp/world-writable-file-creation` query now only detects `open` and `openat` calls with both the `O_CREAT` and `O_TMPFILE` flag, irrespective of whether the `mode` argument is present.
+* he `cpp/world-writable-file-creation` query now only detects `open` and `openat` calls with the `O_CREAT` or `O_TMPFILE` flag.
--- a/cpp/ql/lib/change-notes/2022-02-02-detect-creat-in-openat.md
+++ b/cpp/ql/lib/change-notes/2022-02-02-detect-creat-in-openat.md
@@ -1,4 +1,4 @@
 ---
 category: minorAnalysis
 ---
-* The `cpp/world-writable-file-creation` query now only detects `openat` calls with the `O_CREAT` flag, irrespective of whether the `mode` argument is present.
+* The `cpp/world-writable-file-creation` query now only detects `open` and `openat` calls with both the `O_CREAT` and `O_TMPFILE` flag, irrespective of whether the `mode` argument is present.
--- a/cpp/ql/lib/change-notes/2022-02-02-detect-creat-in-openat.md
+++ b/cpp/ql/lib/change-notes/2022-02-02-detect-creat-in-openat.md
@@ -1,0 +1,4 @@
+---
+category: minorAnalysis
+---
+* The `cpp/world-writable-file-creation` query now only detects `openat` calls with the `O_CREAT` flag, irrespective of whether the `mode` argument is present.
--- a/cpp/ql/lib/change-notes/2022-02-02-detect-missing-mode-argument-for-open.md
+++ b/cpp/ql/lib/change-notes/2022-02-02-detect-missing-mode-argument-for-open.md
@@ -1,0 +1,4 @@
+---
+category: newQuery
+---
+* Added a new query, `open-call-with-mode-argument`, to detect when `open` or `openat` is called with the `O_CREAT` flag but when the `mode` argument is omitted.
--- a/cpp/ql/lib/change-notes/2022-02-02-detect-missing-mode-argument-for-open.md
+++ b/cpp/ql/lib/change-notes/2022-02-02-detect-missing-mode-argument-for-open.md
@@ -1,4 +1,4 @@
 ---
 category: newQuery
 ---
-* Added a new query, `open-call-with-mode-argument`, to detect when `open` or `openat` is called with the `O_CREAT` flag but when the `mode` argument is omitted.
+* Added a new query, `cpp/open-call-with-mode-argument`, to detect when `open` or `openat` is called with the `O_CREAT` or `O_TMPFILE` flag but when the `mode` argument is omitted.
--- a/cpp/ql/lib/semmle/code/cpp/commons/unix/Constants.qll
+++ b/cpp/ql/lib/semmle/code/cpp/commons/unix/Constants.qll
@@ -11,10 +11,10 @@ import cpp
  */
 bindingset[input]
 int parseOctal(string input) {
-  input.charAt(0) = "0" and
+  input.regexpMatch("0[0-7]+") and
   result =
-    strictsum(int ix |
-      ix in [0 .. input.length()]
+    sum(int ix |
+      ix in [1 .. input.length()]
     |
       8.pow(input.length() - (ix + 1)) * input.charAt(ix).toInt()
     )
--- a/cpp/ql/lib/semmle/code/cpp/commons/unix/Constants.qll
+++ b/cpp/ql/lib/semmle/code/cpp/commons/unix/Constants.qll
@@ -13,7 +13,7 @@ bindingset[input]
 int parseOctal(string input) {
   input.regexpMatch("0[0-7]+") and
   result =
-    sum(int ix |
+    strictsum(int ix |
       ix in [1 .. input.length()]
     |
       8.pow(input.length() - (ix + 1)) * input.charAt(ix).toInt()
--- a/cpp/ql/src/Security/CWE/CWE-732/FilePermissions.qll
+++ b/cpp/ql/src/Security/CWE/CWE-732/FilePermissions.qll
@@ -6,7 +6,7 @@ int parseHex(string input) {
   exists(string lowerCaseInput | lowerCaseInput = input.toLowerCase() |
     lowerCaseInput.regexpMatch("0x[0-9a-f]+") and
     result =
-      sum(int ix |
+      strictsum(int ix |
         ix in [2 .. input.length()]
       |
         16.pow(input.length() - (ix + 1)) * "0123456789abcdef".indexOf(lowerCaseInput.charAt(ix))
@@ -14,6 +14,10 @@ int parseHex(string input) {
   )
 }
 
+/**
+ * Gets the value defined by the `O_CREAT` macro if the macro
+ * exists and if every definition defines the same value.
+ */
 int o_creat() {
   result =
     unique(int v |
@@ -23,6 +27,10 @@ int o_creat() {
     )
 }
 
+/**
+ * Gets the value defined by the `O_TMPFILE` macro if the macro
+ * exists and if every definition defines the same value.
+ */
 int o_tmpfile() {
   result =
     unique(int v |
--- a/cpp/ql/src/Security/CWE/CWE-732/FilePermissions.qll
+++ b/cpp/ql/src/Security/CWE/CWE-732/FilePermissions.qll
@@ -3,7 +3,7 @@ import semmle.code.cpp.commons.unix.Constants
 
 bindingset[input]
 int parseHex(string input) {
-  input.prefix(2) = "0x" and
+  input.matches("0x%") and
   result =
     strictsum(int ix |
       ix in [2 .. input.length()]
--- a/cpp/ql/src/Security/CWE/CWE-732/FilePermissions.qll
+++ b/cpp/ql/src/Security/CWE/CWE-732/FilePermissions.qll
@@ -134,7 +134,7 @@ abstract class FileCreationWithOptionalModeExpr extends FileCreationExpr {
 
 class OpenCreationExpr extends FileCreationWithOptionalModeExpr {
   OpenCreationExpr() {
-    this.getTarget().getName() = ["open", "_open", "_wopen"] and
+    this.getTarget().hasGlobalOrStdName(["open", "_open", "_wopen"]) and
     exists(int flag | flag = this.getArgument(1).getValue().toInt() |
       setsFlag(flag, o_creat()) or setsFlag(flag, o_tmpfile())
     )
@@ -163,7 +163,7 @@ class CreatCreationExpr extends FileCreationExpr {
 
 class OpenatCreationExpr extends FileCreationWithOptionalModeExpr {
   OpenatCreationExpr() {
-    this.getTarget().getName() = "openat" and
+    this.getTarget().hasGlobalOrStdName("openat") and
     exists(int flag | flag = this.getArgument(2).getValue().toInt() |
       setsFlag(flag, o_creat()) or setsFlag(flag, o_tmpfile())
     )
--- a/cpp/ql/src/Security/CWE/CWE-732/FilePermissions.qll
+++ b/cpp/ql/src/Security/CWE/CWE-732/FilePermissions.qll
@@ -98,7 +98,7 @@ class OpenCreationExpr extends FileCreationWithOptionalModeExpr {
   override predicate hasModeArgument() { exists(this.getArgument(2)) }
 
   override int getMode() {
-    if hasModeArgument()
+    if this.hasModeArgument()
     then result = this.getArgument(2).getValue().toInt()
     else
       // assume anything is permitted
@@ -125,7 +125,7 @@ class OpenatCreationExpr extends FileCreationWithOptionalModeExpr {
   override predicate hasModeArgument() { exists(this.getArgument(3)) }
 
   override int getMode() {
-    if hasModeArgument()
+    if this.hasModeArgument()
     then result = this.getArgument(3).getValue().toInt()
     else
       // assume anything is permitted
--- a/cpp/ql/src/Security/CWE/CWE-732/FilePermissions.qll
+++ b/cpp/ql/src/Security/CWE/CWE-732/FilePermissions.qll
@@ -1,6 +1,11 @@
 import cpp
 import semmle.code.cpp.commons.unix.Constants as UnixConstants
 
+/**
+ * Gets the number corresponding to the contents of `input` in base-16.
+ * Note: the first two characters of `input` must be `0x`. For example:
+ * `parseHex("0x123abc") = 1194684`.
+ */
 bindingset[input]
 int parseHex(string input) {
   exists(string lowerCaseInput | lowerCaseInput = input.toLowerCase() |
--- a/cpp/ql/src/Security/CWE/CWE-732/FilePermissions.qll
+++ b/cpp/ql/src/Security/CWE/CWE-732/FilePermissions.qll
@@ -111,12 +111,18 @@ class CreatCreationExpr extends FileCreationExpr {
 class OpenatCreationExpr extends FileCreationExpr {
   OpenatCreationExpr() {
     this.getTarget().getName() = "openat" and
-    this.getNumberOfArguments() = 4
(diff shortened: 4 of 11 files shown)


cpp/ql/src/Security/CWE/CWE-732/OpenCallMissingModeArgument.ql
```diff
@@ -0,0 +1,19 @@
+/**
+ * @name File opened with O_CREAT flag but without mode argument
+ * @description Opening a file with the O_CREAT flag but without mode argument reads arbitrary bytes from the stack.
+ * @kind problem
+ * @problem.severity warning
+ * @security-severity 7.8
+ * @precision medium
+ * @id cpp/open-call-with-mode-argument
+ * @tags security
+ *       external/cwe/cwe-732
```
geoffw0, line 10: `precision` and `problem.severity`: are very broadly described in https://github.com/github/codeql/blob/main/docs/query-metadata-style-guide.md

In my own words, precision describes the proportion of results a user should expect to be correct (not just technically correct according to the specification of the query, which may itself have shortcomings).  `medium` suggests a balance of good results and false positives, but not overwhelmingly false positives.  `high` suggests that most results will be correct.  `very-high` suggests you are unlikely to find any false positive results at all (we rarely use `very-high` in CPP because its a complex language with very varied practices).

The other dimension, severity, is about how bad a true positive result is for the project.  `error` means the result is likely to indicate incorrect program behaviour or a security vulnerability, whereas a `warning` or `recommendation` might have less severe / immediate consequences.

The default code scanning suite I believe only shows queries with the `security` tag (I think), a `precision` of `high` or `very-high` and a severity of `warning` or `error` (this is a very high standard).  The security-extended suite is the same but also shows `medium` precision queries, so users expect to get more results at the cost of sorting through more false positives.

Security-severity: there's a (internal) script to generate these.

geoffw0, line 10: For this query I think we could upgrade the `severity` to `error`, because I can't see breaking this rule ever being a good idea.  For `precision`, we need to run the query on a fairly large sample of real world projects (either on lgtm.com or through the DCA job) and see what we get.

geoffw0 (commented):
LGTM apart from a few nitpicks, the meta-data, and the missing change note.

MathiasVP:
> When calling `open`/`openat` with the `O_CREAT` flag, the mode argument must be supplied. The query introduced in this PR checks for this.

When I read this paragraph [here](https://man7.org/linux/man-pages/man2/open.2.html):
> The mode argument must be supplied if O_CREAT or O_TMPFILE is specified in flags; if it is not supplied, some arbitrary bytes from the stack will be applied as the file mode.

it sounds like we should also be highlighting cases where `O_TMPFILE` is used?

(I should add that I'm not really familiar with this API at all.)

cpp/ql/test/query-tests/Security/CWE/CWE-732/OpenCallMissingModeArgument.c
```diff
@@ -0,0 +1,16 @@
+typedef unsigned int mode_t;
+
+#define O_CREAT 0100
+
+int open(const char *pathname, int flags, ...);
+
+int openat(int dirfd, const char *pathname, int flags, ...);
```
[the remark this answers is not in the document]
jketema: Yeah, I actually started out with 4 prototypes (two for `open` and two for `openat`), as you might expect, but of course there's no overloading in C, so that's not possible. So I started looking at actual implementations, and this seems to way this is done in C standard libraries. See, e.g., [OpenBSD](https://github.com/openbsd/src/blob/2207c4325726fdc5c4bcd0011af0fdf7d3dab137/sys/sys/fcntl.h#L208-L216).

Alternative would be to switch to C++.

cpp/ql/src/Security/CWE/CWE-732/OpenCallMissingModeArgument.ql
```diff
@@ -0,0 +1,19 @@
...
+ */
+
+import cpp
+import FilePermissions
+
+from FileCreationWithOptionalModeExpr fc
+where not fc.hasModeArgument()
+select fc,
+  "A file is created here without providing a mode argument, which may leak bits from the stack"
```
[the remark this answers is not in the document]
jketema: Might there be a similar problem in the meta-data of the query? It wasn't completely clear to me where there should be a period or not.

cpp/ql/src/Security/CWE/CWE-732/OpenCallMissingModeArgument.ql
MathiasVP: > The default code scanning suite I believe only shows queries with the security tag (I think), a precision of high or very-high and a severity of warning or error (this is a very high standard). The security-extended suite is the same but also shows medium precision queries, so users expect to get more results at the cost of sorting through more false positives.

This is mostly correct (the security-extended suite doesn't show the medium precision queries by default, though). Here's how you can find the formalization of these rules:

Code Scanning uses the `.qls` file from [here](https://github.com/github/codeql/blob/main/cpp/ql/src/codeql-suites/cpp-code-scanning.qls). That one applies `code-scanning-selectors.yml` from the suite named `codeql/suite-helpers`. Looking that one up we find this [this file](https://github.com/github/codeql/blob/main/misc/suite-helpers/code-scanning-selectors.yml) which has the requirements.

Similarly, the formalization of the rules for the extended security suite can be found [here](https://github.com/github/codeql/blob/main/misc/suite-helpers/security-extended-selectors.yml).

I really hope we can bring this query to `@ precision high` so that it appears in code scanning by default.

cpp/ql/test/query-tests/Security/CWE/CWE-732/OpenCallMissingModeArgument.c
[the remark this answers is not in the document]
geoffw0: Ah, I hadn't thought of that.  If its the way OpenBSD does it, I'm happy.

cpp/ql/src/Security/CWE/CWE-732/OpenCallMissingModeArgument.ql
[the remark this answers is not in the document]
geoffw0: I think just the `@description` has a full stop, but check what other queries do.

cpp/ql/test/query-tests/Security/CWE/CWE-732/OpenCallMissingModeArgument.c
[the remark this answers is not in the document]
jketema: Neither had I. It does explain why you can get random bytes from the stack though, as that's what variadic functions do when you don't provide enough arguments.

jketema:
Observations without knowing how we might handle similar cases: `O_TMPFILE` is Linux specific as far as I can tell, and temp files are dealt with differently on different platforms. For, example, on Windows you need to bitwise-or a constant with `O_CREAT`. This makes me somewhat hesitant, as the value of `O_TMPFILE` maybe used for different purposes elsewhere. `O_CREAT` on the other hand is pretty well standardised.

cpp/ql/test/query-tests/Security/CWE/CWE-732/OpenCallMissingModeArgument.c
```diff
@@ -0,0 +1,16 @@
...
+
+const char *a_file = "/a_file";
+
+void test_open() {
+  open(a_file, O_CREAT);
+  open(a_file, O_CREAT, 0);
+  openat(0, a_file, O_CREAT);
+  openat(0, a_file, O_CREAT, 0);
+}
```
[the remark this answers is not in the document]
jketema: Done. Totally forgot about `// GOOD` and `// BAD`. Thanks for reminding me.

cpp/ql/src/Security/CWE/CWE-732/OpenCallMissingModeArgument.ql
jketema: I changed it to error.

MathiasVP:
> Observations without knowing how we might handle similar cases: `O_TMPFILE` is Linux specific as far as I can tell, and temp files are dealt with differently on different platforms. For, example, on Windows you need to bitwise-or a constant with `O_CREAT`. This makes me somewhat hesitant, as the value of `O_TMPFILE` maybe used for different purposes elsewhere. `O_CREAT` on the other hand is pretty well standardised.

Good point. We could check if there exists an `O_TMPFILE` macro in the database, and that would maybe be good enough to conclude that it's compiled for a Linux environment.

In any case, we don't need to include that in this PR. If we're able to find interesting results with the query as-is there's no need to go down that route for now 👍.

cpp/ql/src/Security/CWE/CWE-732/FilePermissions.qll
```diff
@@ -1,6 +1,29 @@
...
+  input.prefix(2) = "0x" and
+  result =
+    strictsum(int ix |
+      ix in [2 .. input.length()]
+    |
+      16.pow(input.length() - (ix + 1)) * "0123456789abcdef".indexOf(input.charAt(ix).toLowerCase())
+    )
+}
+
+int o_creat_new() {
```
jketema, line 15: This should be named `o_creat`, but that conflicts with `semmle.code.cpp.commons.unix.Constants`. Should this predicate be moved to `Constants` and replace the one there?

cpp/ql/src/Security/CWE/CWE-732/FilePermissions.qll
```diff
@@ -1,6 +1,29 @@
...
+  exists(Macro m | m.getName() = "O_CREAT" | result = parseHex(m.getBody()))
+  or
+  exists(Macro m | m.getName() = "O_CREAT" | result = parseOctal(m.getBody()))
+}
+
+int o_tmpfile() {
```
jketema, line 21: Similarly to `o_creat_new`, should this be moved to `Constants`?

jketema:
> In any case, we don't need to include that in this PR. If we're able to find interesting results with the query as-is there's no need to go down that route for now 👍.

I went down the rabbit hole after all, as it turns out `O_CREAT` is also defined differently on different OSes 🙁 @geoffw0 and I figured this out after running a query on LGTM and finding a false positive.

cpp/ql/lib/change-notes/2022-02-02-detect-missing-mode-argument-for-open.md
```diff
@@ -0,0 +1,4 @@
+---
+category: newQuery
+---
+* Added a new query, `open-call-with-mode-argument`, to detect when `open` or `openat` is called with the `O_CREAT` flag but when the `mode` argument is omitted.
```
geoffw0, line 4:
```suggestion
* Added a new query, `cpp/open-call-with-mode-argument`, to detect when `open` or `openat` is called with the `O_CREAT` flag but when the `mode` argument is omitted.
```

cpp/ql/src/Security/CWE/CWE-732/FilePermissions.qll
geoffw0: I'm a bit reluctant about (though not strongly against) changing `semmle.code.cpp.commons.unix.Constants` because its clearly specific to the Unix implementation, where the value given presumably is always correct.

You can solve the name clash by replacing
```
import semmle.code.cpp.commons.unix.Constants
```
with
```
import semmle.code.cpp.commons.unix.Constants as UnixConstants
```
and then accessing it via `UnixConstants::o_creat` (or whatever).

cpp/ql/src/Security/CWE/CWE-732/FilePermissions.qll
```diff
@@ -1,6 +1,29 @@
 import cpp
 import semmle.code.cpp.commons.unix.Constants
 
+bindingset[input]
+int parseHex(string input) {
+  input.prefix(2) = "0x" and
```
MathiasVP: CodeQL is correct here 😂. `input.prefix(2)` creates a new string at runtime whereas `input.matches("0x%")` won't.

MathiasVP: Ah, you fixed it already. Excellent!

cpp/ql/src/Security/CWE/CWE-732/FilePermissions.qll
jketema: Resolved as suggested.

cpp/ql/lib/change-notes/2022-02-02-detect-creat-in-openat.md
```diff
@@ -0,0 +1,4 @@
+---
+category: minorAnalysis
+---
+* The `cpp/world-writable-file-creation` query now only detects `open` and `openat` calls with both the `O_CREAT` and `O_TMPFILE` flag, irrespective of whether the `mode` argument is present.
```
geoffw0, line 4: This is getting difficult to understand.  Is the following still accurate?
```suggestion
* The `cpp/world-writable-file-creation` query now only detects `open` and `openat` calls with the `O_CREAT` or `O_TMPFILE` flag.
```

geoffw0 (commented):
I think we're nearly there.  We do still need to decide on a `precision`.

MathiasVP:
After we've assigned it a precision we should also run the security-extended or the code scanning suite (depending on which precision we assign it)

cpp/ql/lib/change-notes/2022-02-02-detect-creat-in-openat.md
jketema: Close enough. I'll change it.

cpp/ql/lib/change-notes/2022-02-02-detect-creat-in-openat.md
```diff
@@ -0,0 +1,4 @@
+---
+category: minorAnalysis
+---
+* he `cpp/world-writable-file-creation` query now only detects `open` and `openat` calls with the `O_CREAT` or `O_TMPFILE` flag.
```
MathiasVP, line 4:
```suggestion
* The `cpp/world-writable-file-creation` query now only detects `open` and `openat` calls with the `O_CREAT` or `O_TMPFILE` flag.
```

cpp/ql/lib/semmle/code/cpp/commons/unix/Constants.qll
```diff
@@ -11,10 +11,10 @@ import cpp
  */
 bindingset[input]
 int parseOctal(string input) {
-  input.charAt(0) = "0" and
+  input.regexpMatch("0[0-7]+") and
   result =
-    strictsum(int ix |
-      ix in [0 .. input.length()]
+    sum(int ix |
+      ix in [1 .. input.length()]
```
MathiasVP, line 17: If the input matches the regexp `0[0-7]+` the length of the string must be at least 2, and so the range `[1 .. input.length()]` must be non-empty, right? So can't this still be a strictsum?

cpp/ql/src/Security/CWE/CWE-732/FilePermissions.qll
```diff
@@ -15,15 +15,21 @@ int parseHex(string input) {
 }
 
 int o_creat() {
-  exists(Macro m | m.getName() = "O_CREAT" |
-    result = parseHex(m.getBody()) or result = UnixConstants::parseOctal(m.getBody())
-  )
+  result =
+    unique(int v |
```
MathiasVP, line 19: Is this `unique` actually required? Is it ever the case that the body can both be valid hex and valid octal?

cpp/ql/src/Security/CWE/CWE-732/FilePermissions.qll
```diff
@@ -15,15 +15,21 @@ int parseHex(string input) {
...
+      )
+    )
 }
 
 int o_tmpfile() {
-  exists(Macro m | m.getName() = "O_TMPFILE" |
-    result = parseHex(m.getBody()) or result = UnixConstants::parseOctal(m.getBody())
-  )
+  result =
+    unique(int v |
```
MathiasVP, line 28: Same question here.

cpp/ql/lib/semmle/code/cpp/commons/unix/Constants.qll
jketema: It can be, but it's not necessary. Is there a reason to prefer `strictsum`?

cpp/ql/src/Security/CWE/CWE-732/FilePermissions.qll
jketema: There's on codebase that that defines `O_CREAT` with two different values. So, not so much hex vs. oct. Maybe worth a comment?

cpp/ql/lib/semmle/code/cpp/commons/unix/Constants.qll
MathiasVP: `strictsum` is more efficient 😍. `sum` needs to handle the case where the `[1 .. input.length()]` range is empty, while `strictsum` doesn't. I've included the gory details if you're curious:

This query:
```ql
from int n
where n = sum(Expr e | exists(e.getValue().toInt()) | e.getValue().toInt())
select n
```
compiles to the following DIL code:
```
noinline project##select#query#f#nullary() :-
  exists(cached string call_result#2 |
    exists(/* Expr::Expr */ cached dontcare entity _ |
      Expr::Expr::getValue_dispred#ff(_, call_result#2)
    ),
    exists(dontcare int _ | toInt(call_result#2, _))
  )
.

#select#query#f(unique int n) :-
  n = sum(int sum#project#)(/* Expr::Expr */ cached entity e, int sum#expr# |
    exists(cached string call_result#4 |
      Expr::Expr::getValue_dispred#ff(e, call_result#4),
      toInt(call_result#4, sum#expr#)
    ) |
    sum#project# = sum#expr#
  );
  (not(project##select#query#f#nullary()), n = 0)
.
```
The `(not(project##select#query#f#nullary()), n = 0)` disjunct is the special case that needs to be there in case the range is empty (if you look closely you'll see that `project##select#query#f#nullary()` is a predicate that holds if the range is non-empty).

Compare that to the DIL for this query:
```ql
from int n
where n = strictsum(Expr e | exists(e.getValue().toInt()) | e.getValue().toInt())
select n
```
here you get:
```
#select#query#f(unique int n) :-
  n = sum(int strictsum#project#)(/* Expr::Expr */ cached entity e,
                                  int strictsum#expr# |
    exists(/* @value */ cached entity v |
      exists(cached string call_result#4 |
        values(v, call_result#4), toInt(call_result#4, strictsum#expr#)
      ),
      valuebind(v, e)
    ) |
    strictsum#project# = strictsum#expr#
  )
.
```
which is basically the same piece of code, but with no checking if the range is non-empty.

cpp/ql/src/Security/CWE/CWE-732/FilePermissions.qll
MathiasVP: Ah, I see! It would be great to have a comment on that, indeed :)

jketema: I tried to address this. Not sure if I have been clear enough, so please have a look.

MathiasVP: Awsome. That's exactly how I would phrase it as well :)

cpp/ql/src/Security/CWE/CWE-732/OpenCallMissingModeArgument.ql
jketema: I did a run on LTGM and all instances seem to be true positives now, so setting `@precision high` seems reasonable.

cpp/ql/src/Security/CWE/CWE-732/FilePermissions.qll
```diff
@@ -1,5 +1,44 @@
 import cpp
-import semmle.code.cpp.commons.unix.Constants
+import semmle.code.cpp.commons.unix.Constants as UnixConstants
+
+bindingset[input]
+int parseHex(string input) {
```
geoffw0, line 5: Can we have a QLDoc comment for this predicate, similarly to `parseOctal`?  (your comments are otherwise excellent)

cpp/ql/src/Security/CWE/CWE-732/OpenCallMissingModeArgument.ql
geoffw0: I agree, and that's great news as this query will have higher visibility. :+1:

geoffw0 (dismissed):
Looks great, results are great, I have one nitpicky comment about QLDoc but this is ready to merge.

geoffw0 (approved):
Thanks. :+1:

PR #7799, opened 2022-02-01
Ruby: use ruby specific cache key

aibaars:
The ql-for-ql workflow copied the original Ruby ones and likely caused clashes with the cache-key names.


nickrolfe:
Looks fine. Do you want to change the ql-for-ql workflows as well? Particularly the `**/Cargo.lock` part.

nickrolfe (approved):

PR #7801, opened 2022-02-01
JS: Migrate CodeQL tests for ML-powered queries

henrymercer:
This PR migrates our internal CodeQL tests for ML-powered queries to the `github/codeql` repo.

## Generated test data

The `endpoint_large_scale` tests run over test data used in the standard JavaScript libraries to test NoSQL injection, SQL injection, path injection, and XSS queries.

The `github/codeql` repo contains a `sync-files.py` script which is used to keep duplicates up to date. However upon examination it looks like it might not be that easy to adapt it to keep globs in sync rather than just files in sync (syncing the current list of files isn't sufficient since we want to capture new test files being added too).

Therefore I have kept our existing script to sync the files.  I've also added the generated paths to `.gitattributes` so that they don't pollute the PR diff view.

## CI

I've added new "JS ML-powered queries tests" CI jobs to verify the formatting and compilation of our QL code and to run the QL tests.  I added a separate job rather than integrating with the JS language tests job to avoid increasing the runtime of that job, since it's already pretty slow.  Furthermore a separate job only needs to run when the ML-powered codebase is changed, and gives us quicker feedback on PRs.

There is some duplication since the JS language tests job will verify that everything in `javascript/ql` formats correctly, so the separate formatting check isn't necessary.  However I've kept it so that we can get a quicker turnaround time if formatting is incorrect on one of our PRs.


javascript/ql/experimental/adaptivethreatmodeling/test/modeled_apis/package.json
```diff
@@ -0,0 +1,9 @@
+{
+  "name": "",
+  "description": "",
+  "version": "0.0.0",
+  "dependencies": {
+    "mongodb": "2.2.33"
+  },
+  "devDependencies": {}
+}
```
Z80coder, line 9: What's the purpose of this file in the context of the tests?

Z80coder (dismissed):
LGTM, but I'm not qualified to comment on whether this test framework is consistent with the existing testing conventions of `github/codeql`.

I wonder also whether it would be good to add a `README.md` in the `test` folder that explains how to run these tests locally?

javascript/ql/experimental/adaptivethreatmodeling/test/modeled_apis/package.json
henrymercer: This specifies the version of the `mongodb` package that the tests depend on, since the package's API depends on the version of the package.  It's not strictly speaking necessary but helps us write and understand the tests.

henrymercer:
> I wonder also whether it would be good to add a `README.md` in the `test` folder that explains how to run these tests locally?

I would usually be in favour of documentation like this, but I think it's superfluous here since the process is identical to the rest of the packs in this repo — you can run the tests by opening the repo in VS Code and triggering them from the Test Explorer.  The process to run the ML-powered library tests from the CLI is also standard.

.github/workflows/js-ml-tests.yml
```diff
@@ -0,0 +1,70 @@
...
+    branches:
+      - main
+      - "rc/*"
+  pull_request:
+    paths:
+      - "javascript/ql/experimental/adaptivethreatmodeling/**"
+      - .github/workflows/js-ml-tests.yml
+    branches:
+      - main
+      - "rc/*"
```
aeisenberg, line 17: Maybe we still want to run these tests if they are PRs against an unprotected branch.

javascript/ql/experimental/adaptivethreatmodeling/test/qlpack.yml
```diff
@@ -0,0 +1,5 @@
+name: codeql/javascript-experimental-atm-tests
+version: 0.0.0
```
aeisenberg, line 2: If this pack is not going to be published, best to remove the version (prevents accidental publishing).

```suggestion
```

javascript/ql/experimental/adaptivethreatmodeling/test/update_endpoint_test_files.py
```diff
@@ -0,0 +1,49 @@
+#!/usr/bin/env python3
+
+# This script updates the JavaScript test data used by the endpoint CodeQL tests.
```
aeisenberg, line 3: How often should this script be run? How are we going to be reminded to run it? We _could_ create an action that does this on a schedule, but that might be overkill.

aeisenberg (dismissed):
Looks sensible. I have a few questions and suggestions inline, but nothing that is strictly required before it can be merged.

javascript/ql/experimental/adaptivethreatmodeling/test/update_endpoint_test_files.py
henrymercer: Given that being out of date just means having potentially slightly worse test data, I don't think it's worth investing that much in automating it. I was thinking of simply setting up a Slack reminder.

aeisenberg (approved):

aeisenberg:
Looks good from my perspective. It should go in as long as @Z80coder agrees.

Z80coder (approved):

PR #7802, opened 2022-02-01
C++: Recognize password struct fields.

geoffw0:
This is an enhancement to `cpp/cleartext-transmission` that allows it to spot passwords that are struct or class fields (much more reliably).  I've weeded out a lot of false positive results in previous PRs, but I wanted to address one of the causes of _missing_ results so that we don't just end up having very few results overall.

Note that the `SensitiveNode` isn't a perfect implementation; it tries to avoid matching all accesses to a `SensitiveVariable` to reduce duplicate results where a variable is accessed multiple times, but this is not done in an especially principled or reliable manner.  Happy to take suggestions on how to pick the right `Node` better.

New results: https://lgtm.com/query/6627472974229539370/
All results: https://lgtm.com/query/7344851133823789124/


MathiasVP (approved):
LGTM!

PR #5429, opened 2021-03-17
C#: Add data flow tests for various pattern types

tamasvajk:

PR #7809, opened 2022-02-02
C#: Add flow test cases for undetected value flow, when making variable bindings in pattern matching.

michaelnebel:


tamasvajk:
A while back I also collected some unhandled tuple related patterns, but I never actually got to merging it: https://github.com/github/codeql/pull/5429. Maybe we could merge these efforts now.

Also, I think your test file shouldn't be in the `library-tests/dataflow/fields/` folder. Maybe a new `records` folder would be more appropriate.

michaelnebel:
That is a good point!
Not sure, what the best strategy is for structuring the tests.
Even though records are used here (it could just as well had been classes or structs), it is only the case, because it was the shortest form of type declarations I could think of, where we could make use of pattern matching with property patterns and variable bindings.
Maybe we need a folder *pattern* for all pattern related tests?

michaelnebel:
Should we just stick to the existing strategy?
Then I will create a *records* folder and move the pattern test here.

tamasvajk:
> Should we just stick to the existing strategy? Then I will create a _records_ folder and move the pattern test here.

I don't know if there's an explicit existing strategy here. The `field` folder seems to contain tests for flows through fields. So these pattern related tests don't seem to belong there. We could add them to the `local` folder, or to a new `pattern` folder. I would probably go with the latter just to be explicit. (Although the `tuples` folder already contains test cases with pattern matching.) What do you think @hvitved?

hvitved:
Let's add this to a new `patterns` folder (along with a new test).

hvitved (approved):

PR #7811, opened 2022-02-02
Python: refactor API-graph labels to an IPA type

erik-krogh:
Converts the labels in API graphs to an IPA type.  
A similar refactoring helped performance in JS, but I doubt it'll do much here (because you're not using API-graphs that much yet).   
However, the refactoring should help keep performance good in the future. 

This should be a simple easy to review PR before I open [the big one](https://github.com/github/codeql/pull/7806) for review. 

See the backref for a performance evaluation (not done yet, but I won't merge unless it turns out OK).

### DIFF
--- a/python/ql/lib/semmle/python/ApiGraphs.qll
+++ b/python/ql/lib/semmle/python/ApiGraphs.qll
@@ -114,13 +114,13 @@ module API {
      * Gets a node such that there is an edge in the API graph between this node and the other
      * one, and that edge is labeled with `lbl`.
      */
-    Node getASuccessor(string lbl) { Impl::edge(this, lbl, result) }
+    Node getASuccessor(Label::ApiLabel lbl) { Impl::edge(this, lbl, result) }
 
     /**
      * Gets a node such that there is an edge in the API graph between that other node and
      * this one, and that edge is labeled with `lbl`
      */
-    Node getAPredecessor(string lbl) { this = result.getASuccessor(lbl) }
+    Node getAPredecessor(Label::ApiLabel lbl) { this = result.getASuccessor(lbl) }
 
     /**
      * Gets a node such that there is an edge in the API graph between this node and the other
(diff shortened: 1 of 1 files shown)


erik-krogh:
[An evaluation](https://github.com/github/codeql-dca-main/tree/data/erik-krogh/pr-7811-e06f652__nightly__code-scanning/reports) shows a slight hint of a performance improvement.

python/ql/lib/semmle/python/ApiGraphs.qll
```diff
@@ -514,36 +514,126 @@ module API {
...
+    LabelModule mod(string m) { result.getMod() = m }
+
+    /** Gets the `member` edge label for member `m`. */
+    LabelMember member(string m) { result.getMember() = m }
+
+    /** Gets the `member` edge label for the unknown member. */
+    LabelUnknownMember unknownMember() { any() }
+
+    /** Gets the `member` edge label for the given attribute reference. */
+    ApiLabel memberFromRef(DataFlow::AttrRef pr) {
```
yoff, line 623: The variable is probably called `pr` because it once referred to a property reference. But fixing it is unrelated to this PR.

yoff (approved):
LGTM

yoff:
I will give @tausbn a chance to look this over, but it looks ready to merge to me.

erik-krogh:
> it looks ready to merge to me.

I'll just merge it. I can do corrections in the next PR.

python/ql/lib/semmle/python/ApiGraphs.qll
```diff
@@ -514,36 +514,126 @@ module API {
...
+        MkLabelUnknownMember() or
+        MkLabelParameter(int i) {
+          none() // TODO: Fill in when adding def nodes
+        } or
+        MkLabelReturn() or
+        MkLabelSubclass() or
+        MkLabelAwait()
+
+      /** A label for a module. */
+      class LabelModule extends ApiLabel {
```
tausbn, line 549: Why is it `LabelModule` and not `ModuleLabel`? The latter reads much better to me.

erik-krogh: It was to get some very clear and consistent names of what the classes represent.
All the newtypes start with `MkLabel`, and all the classes start with `Label`.   
Those names are all private inside the `Label` module, so they're not ever exposed outside of `ApiGraphs.qll`.

PR #7819, opened 2022-02-02
Ruby: add def-nodes and separate method/return steps to API graphs

asgerf:
Following up on a discussion with @hvitved and @aibaars, this makes some API graph changes in preparation for MaD.

Happy to wait until after https://github.com/github/codeql/pull/7663 and fix the merge conflicts here (though there are surprisingly few conflicts)

### Changes to API graphs
- The `getReturn("foo")` edge is now split into `getMethod("foo").getReturn()` through an intermediate node representing the method access itself. This node is always unique to the call site, so it can also also be used as a representative for the call itself (this differs from JS, where we need `API::CallNode` to work with calls in the API graph).
- Adds edges going from the method access to its parameters. These have labels:
  - `getParameter(n)` for the `n`th positional parameter
  - `getKeywordParameter(name)` for the keyword parameter with that `name`
  - `getBlock()` for the block parameter
- The `instance` edge label has been replaced with the path `getMethod("new").getReturn()`. We need a way to refer to the arguments of a constructor call, just like any other call, so we need the intermediate method/constructor access node to exist, and here it seemed better to unify the two representations rather than invent a "constructor access node". The `getInstance()` method in QL is still there, it's just the underlying graph that has changed.

NOTE: API graphs labels do not distinguish between "argument" and "parameter". This might look a bit confusing at first, and possibly made more sense when consuming an API graph from a library, where access paths in the library needed to match up with those in the application. Just mentioning it up front that this was deliberate and matches how we do it for JS.

### Example
```rb
X.foo do |x|
  x.bar
end
```
You can get to `x.bar` via the path `getMember("X").getMethod("foo").getBlock().getParameter(0).getMethod("bar").getReturn()`.

Explaining each step of the path:
- `getMember("X")` gets the `X`
- `.getMethod("foo")` gets the `X.foo` call (the call itself, not its return value)
- `.getBlock()` gets the entire block argument (`|x| ... end`)
- `.getParameter(0)` gets the first parameter of the block (`x`)
- `.getMethod("bar")` gets the `x.bar` call
- `.getReturn()` get the return value of the `x.bar` call

### Drive-by fixes

There are two fixes to the data-flow libraries, found while debugging/pairing with @hvitved. I'm happy to split them out into a separate PR if necessary.

### DIFF
--- a/ruby/ql/lib/codeql/ruby/ApiGraphs.qll
+++ b/ruby/ql/lib/codeql/ruby/ApiGraphs.qll
@@ -44,6 +44,11 @@ module API {
     DataFlow::LocalSourceNode getAnImmediateUse() { Impl::use(this, result) }
 
     /**
+     * Gets a data-flow node corresponding the value flowing into this API component.
+     */
+    DataFlow::Node getARhs() { Impl::def(this, result) }
+
+    /**
      * Gets a call to a method on the receiver represented by this API component.
      */
     DataFlow::CallNode getAMethodCall(string method) {
@@ -96,6 +101,30 @@ module API {
     Node getReturn(string method) {
       result = this.getASubclass().getASuccessor(Label::return(method))
     }
+
+    private predicate hasParameterIndex(int n) {
+      exists(string str |
+        exists(this.getASuccessor(Label::parameterByStr(str))) and
+        n = str.toInt()
+      )
+    }
+
+    /** Gets an API node representing the `n`th positional parameter. */
+    Node getParameter(int n) {
+      result = this.getASuccessor(Label::parameter(n)) and this.hasParameterIndex(n)
+    }
+
+    private predicate hasKeywordParameter(string name) {
+      exists(this.getASuccessor(Label::keywordParameter(name)))
+    }
+
+    /** Gets an API node representing the given keyword parameter. */
+    Node getKeywordParameter(string name) {
+      result = this.getASuccessor(Label::keywordParameter(name)) and this.hasKeywordParameter(name)
+    }
+
+    /** Gets an API node representing the block parameter. */
+    Node getBlock() { result = this.getASuccessor(Label::blockParameter()) }
 
     /**
      * Gets a `new` call to the function represented by this API component.
@@ -260,7 +289,9 @@ module API {
       /** The root of the API graph. */
       MkRoot() or
       /** A use of an API member at the node `nd`. */
-      MkUse(DataFlow::Node nd) { isUse(nd) }
+      MkUse(DataFlow::Node nd) { isUse(nd) } or
+      /** A value that escapes into an API at the node `nd` */
+      MkDef(DataFlow::Node nd) { isDef(nd) }
 
     private string resolveTopLevel(ConstantReadAccess read) {
       TResolved(result) = resolveConstantReadAccess(read) and
@@ -319,6 +350,8 @@ module API {
         useCandFwd().flowsTo(node) and
         useStep(_, node, nd)
       )
+      or
+      parameterStep(_, defCand(), nd)
     }
 
     /**
@@ -327,6 +360,13 @@ module API {
     cached
     predicate use(TApiNode nd, DataFlow::Node ref) { nd = MkUse(ref) }
 
+    /**
+     * Holds if `ref` is a RHS of node `nd`.
+     */
+    cached
+    predicate def(TApiNode nd, DataFlow::Node rhs) { nd = MkDef(rhs) }
+
+    /** Gets a node reachable from a use-node. */
     private DataFlow::LocalSourceNode useCandFwd(TypeTracker t) {
       t.start() and
       isUse(result)
@@ -334,6 +374,7 @@ module API {
       exists(TypeTracker t2 | result = useCandFwd(t2).track(t2, t))
     }
 
+    /** Gets a node reachable from a use-node. */
     private DataFlow::LocalSourceNode useCandFwd() { result = useCandFwd(TypeTracker::end()) }
 
     private DataFlow::Node useCandRev(TypeBackTracker tb) {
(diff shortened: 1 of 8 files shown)


ruby/ql/lib/codeql/ruby/ApiGraphs.qll
```diff
@@ -43,6 +43,11 @@ module API {
      */
     DataFlow::LocalSourceNode getAnImmediateUse() { Impl::use(this, result) }
 
+    /**
+     * Gets a data-flow node corresponding the value flowing into this API component.
+     */
+    DataFlow::Node getARhs() { Impl::def(this, result) }
```
erik-krogh, line 49: Shouldn't you also add `getAValueReachingRhs()`?

asgerf:
Force-pushed to resolve conflicts with https://github.com/github/codeql/pull/7663.

The `getMethod(name)` predicate now performs `getASubclass()`. This affects both `getReturn(name)` and `getInstance()` so both are subclass-aware, preserving the behaviour from the previous PR.

ruby/ql/lib/codeql/ruby/ApiGraphs.qll
```diff
@@ -240,8 +303,12 @@ module API {
     newtype TApiNode =
       /** The root of the API graph. */
       MkRoot() or
+      /** The method accessed at `call`, synthetically treated as a separate object. */
+      MkMethodAccessNode(DataFlow::CallNode call) { isUse(call) } or
       /** A use of an API member at the node `nd`. */
-      MkUse(DataFlow::Node nd) { isUse(nd) }
+      MkUse(DataFlow::Node nd) { isUse(nd) } or
+      /** A value that escapes into an API at the node `nd` */
+      MkDef(DataFlow::Node nd) { isDef(nd) }
```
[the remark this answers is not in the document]
asgerf: I can see that "def node" is a confusing name without more context, so I've opened a discussion internally about whether to rename it.

But briefly: it's a place where a value flows _from_ this codebase _into_ another codebase.

> What does it mean for a value to `escape into an API` as the comment says?

I've rephrased this as: A value that escapes into an external library at the node `nd`. Does that make more sense?

ruby/ql/test/library-tests/dataflow/api-graphs/callbacks.rb
```diff
@@ -0,0 +1,39 @@
+Something.foo.withCallback do |a, b| #$ use=getMember("Something").getMethod("foo").getReturn().getMethod("withCallback").getReturn()
+    a.something #$ use=getMember("Something").getMethod("foo").getReturn().getMethod("withCallback").getBlock().getParameter(0).getMethod("something").getReturn()
+    b.somethingElse #$ use=getMember("Something").getMethod("foo").getReturn().getMethod("withCallback").getBlock().getParameter(1).getMethod("somethingElse").getReturn()
+end
```
hvitved, line 4: Perhaps we should update the test to also cover `def`s. So we would have `def=getMember("Something").getMethod("foo").getReturn().getMethod("withCallback").getBlock()` here (right?)

hvitved (changes requested):
Looks really great! Thanks for helping out with this Asger.

ruby/ql/lib/codeql/ruby/ApiGraphs.qll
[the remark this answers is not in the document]
hmac: Thanks! This (plus the internal discussion) really helps me to get my head around it.

ruby/ql/lib/codeql/ruby/ApiGraphs.qll
```diff
@@ -43,6 +45,16 @@ module API {
      */
     DataFlow::LocalSourceNode getAnImmediateUse() { Impl::use(this, result) }
 
+    /**
+     * Gets a data-flow node corresponding the value flowing into this API component.
+     */
+    DataFlow::Node getARhs() { Impl::def(this, result) }
```
hmac, line 51: Is this `Rhs` as in right-hand side? Right-hand side of what? This predicate seems to be `getInducingNode` specialised to just def nodes, but we don't use it in this PR. Is it necessary? Are there later MaD changes that rely on it maybe?

erik-krogh: It is necessary for MaD, and you are likely going to use it for future library modelling.  

You can see some examples of how `getARhs()` is used [in my Python PR](https://github.com/github/codeql/pull/7806/commits/e434f075fa00123d2d9104e3ec2d58df250ce538).

ruby/ql/lib/codeql/ruby/ApiGraphs.qll
```diff
@@ -289,46 +356,33 @@ module API {
      * Holds if `ref` is a use of a node that should have an incoming edge labeled `lbl`,
      * from a use node that flows to `node`.
      */
-    private predicate useStep(string lbl, ExprCfgNode node, DataFlow::Node ref) {
+    private predicate useStep(string lbl, DataFlow::Node node, DataFlow::Node ref) {
```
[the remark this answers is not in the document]
asgerf: It's an edge between nodes, I don't see why we'd want to restrict the predecessor to be `ExprNode`.

We've had our fair share of bugs in the JS libraries due to use of the `ValueNode` in places where something could only be a value node at the time it was written, but then silently changed to become an unwanted filter after a refactoring. Personally I'd avoid using `ExprNode` in a predicate signature unless the predicate only makes sense for that type of node.

ruby/ql/lib/codeql/ruby/ApiGraphs.qll
```diff
@@ -88,15 +100,49 @@ module API {
...
+    private predicate hasParameterIndex(int n) {
+      exists(string str |
+        exists(this.getASuccessor(Label::parameterByStr(str))) and
+        n = str.toInt()
+      )
+    }
+
+    /** Gets an API node representing the `n`th positional parameter. */
+    Node getParameter(int n) {
+      result = this.getASuccessor(Label::parameter(n)) and this.hasParameterIndex(n)
```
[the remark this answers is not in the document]
asgerf: 👍 

The point was to avoid `InverseAppend` leaking out to the callers of `getParameter` and `getKeywordParameter`, which caused some problems in the JS libraries, so I've also added `pragma[nomagic]` on the two predicates for now.

@erik-krogh's change to IPA type labels was proven to be the most robust fix

ruby/ql/test/library-tests/dataflow/api-graphs/callbacks.rb
asgerf: Added `def` tags as optional. It seemed a bit much to require them everywhere.

hvitved (approved):
Let's do a DCA run before merging.

ruby/ql/lib/codeql/ruby/ApiGraphs.qll
hmac: Ah I see. The name makes sense in the case of `foo.bar = x` as in the example in your PR. For method call parameters it is a bit of a weird name.

asgerf:
The [evaluation](https://github.com/github/codeql-dca-main/tree/data/asgerf/ruby-def-nodes__Differences__code-scanning__2/reports) shows OK performance, although it isn't without some cost. I've investigated the performance in opal and didn't see any bad join orders. There are roughly twice as many API nodes to track (2M up from 1M), so there's just more work to do.

The 8 new results were caused by https://github.com/github/codeql/pull/7819/commits/9a496e647f658d14dcec414dcfe90e2226561eb9

PR #7821, opened 2022-02-02
QL: upload sarif as part of the QL-for-QL workflow

erik-krogh:
Code-scanning doesn't provide a nice way of splitting the results based on which run produced them.  

So now we keep a .sarif file for each language implementation as an artifact.  

If you open the .sarif file in VSCode (and install the recommended extension), then you can nicely view and directly fix the results.


tausbn (approved):
Lovely!

PR #7824, opened 2022-02-02
Ruby: Model calls to `constantize` as code executions

hmac:
[`constantize`](https://api.rubyonrails.org/classes/String.html#method-i-constantize) is an ActiveSupport extension on `String` that attempts to look up a constant with the same name as the receiver. It has similar issues to [`Module#const_get`](https://docs.ruby-lang.org/en/3.1/Module.html#method-i-const_get).


nickrolfe:
Looks very sensible, but can you do a DCA run?

nickrolfe (approved):
LGTM

PR #7825, opened 2022-02-02
Python: adjust downgrade script location and format

cklin:
This PR updates Python database downgrade scripts to the expected location and format.


tausbn (approved):
To the extent that I understand what's going on here, this looks fine to me.

yoff:
This seems to create an asymmetry with where we store our upgrade scripts, should those also be moved?

Issue #7803, opened 2022-02-02
Missing model for `.get()` function for `Map` in Unvalidated Dynamic Call 


Naman-ntc:
I have multiple questions/reports.
1.) The following code has the Unvalidated Dynamic Call vulnerability however it is missed by codeql. 
```js
var actions = new Map();
actions.put("play", function(data) {
    // ...
});
actions.put("pause", function(data) {
    // ...
});

app.get('/perform/:action/:payload2', function(req, res) {
    let action = actions.get(req.params.action);
    res.end(action(req.params.payload));
});
```
This seems to be because of the missing model for the `.get` function of `Map`. Based on the occurrence of `PropRead`, the `isAdditionalTaintStep`, the `tgtlabel` is modified as `MaybeFromProto` or `MaybeNonFunction`. However, a similar case is missing for Map `.get` method. The following is a (perhaps dirty and potentially wrong) way to implement the missing case that needs to be added to the `isAdditionalTaintStep`
```codeql
 ("get" = dst.getAstNode().(MethodCallExpr).getMethodName() and
    src.asExpr() = dst.getAstNode().(MethodCallExpr).getAnArgument() and 
    srclabel.isTaint() and dstlabel instanceof MaybeNonFunction)
```

2.) I wanted to clarify whether the following code is expected to be safe without the `FunctionCheck` sanitizer guard. Even after adding the `.get` model, the following code still doesn't report the vulnerability (even though there are two sanitizers that are required?).
```js
var express = require('express');
var app = express();

var actions = new Map();
actions.put("play", function play(data) {
  // ...
});
actions.put("pause", function pause(data) {
  // ...
});

app.get('/perform/:action/:payload', function(req, res) {
  if (actions.has(req.params.action)) {
    let action = actions.get(req.params.action);
    // GOOD: `action` is either the `play` or the `pause` function from above
    res.end(action(req.params.payload));
  } else {
    res.end("Unsupported action.");
  }
});
```

I suspect it might be because `WhitelistContainmentCallSanitizer` below sanitizes both the labels. Is that correct? 
https://github.com/github/codeql/blob/15c1ce722a4bcaa892e982c34ac28cd3d77e4a11/javascript/ql/lib/semmle/javascript/dataflow/TaintTracking.qll#L1065

PR #7828, opened 2022-02-03
JS: Adding model for `.get` function of `Map` in Unvalidated Dynamic Method Call

Naman-ntc:
Fixes the first part of #7803 
Thanks to @max-schaefer for suggesting the dataflow alternative of enhancement I proposed
Summary of changes -
* Introduced a model for `.get` function of `Map` 
* Added more test files


javascript/ql/test/query-tests/Security/CWE-754/UnvalidatedDynamicMethodCallGood4.js
```diff
@@ -0,0 +1,20 @@
...
+actions.put("play", function play(data) {
+    // ...
+});
+actions.put("pause", function pause(data) {
+    // ...
+});
+
+app.get('/perform/:action/:payload', function(req, res) {
+    if (typeof actions.get(req.params.action) === 'function') {
+        let action = actions.get(req.params.action); // OK but flagged [INCONSISTENCY]
```
Naman-ntc, line 14: This is flagged by our query. Is this expected?

javascript/ql/test/query-tests/Security/CWE-754/UnvalidatedDynamicMethodCallGood3.js
```diff
@@ -0,0 +1,20 @@
...
+    // ...
+});
+actions.put("pause", function pause(data) {
+    // ...
+});
+
+app.get('/perform/:action/:payload', function(req, res) {
+    let action = actions.get(req.params.action);
+    // GOOD: `action` is either the `play` or the `pause` function from above
+    if (typeof action === 'function') {
```
Naman-ntc, line 15: This should not be flagged because `action` cannot be `proto` right?

javascript/ql/test/query-tests/Security/CWE-754/UnvalidatedDynamicMethodCallGood4.js
max-schaefer: It's a limitation of the query, I think. In general it's hard to tell that these two calls to `get` must return the same value and hence the first one sanitises the second.

Naman-ntc: Yep, that sounds reasonable.

javascript/ql/test/query-tests/Security/CWE-754/UnvalidatedDynamicMethodCallGood3.js
erik-krogh: That sounds right, a call to `get` on a map shouldn't be able to return something from the prototype.   

I've seen a bunch of `get` implementations that are implemented like `function get(x) {return obj[x];}`, but if that's the case, then the query should pick up on the dynamic property read inside the `get` function.

javascript/ql/test/query-tests/Security/CWE-754/UnvalidatedDynamicMethodCallGood4.js
```diff
@@ -0,0 +1,20 @@
...
+        // GOOD: `action` is either the `play` or the `pause` function from above
+        res.end(action(req.params.payload));
```
erik-krogh, line 16:
```suggestion
        let action = actions.get(req.params.action);
        res.end(action(req.params.payload));  // OK but flagged [INCONSISTENCY]. `action` is either the `play` or the `pause` function from above
```

We usually put the `//OK ` or `//NOT OK` comments on the line that is flagged.  
For path-queries it's the sink that gets flagged, so the comment should go there.

javascript/ql/test/query-tests/Security/CWE-754/UnvalidatedDynamicMethodCallGood3.js
```diff
@@ -0,0 +1,20 @@
...
+        res.end(action(req.params.payload));
```
erik-krogh, line 16:
```suggestion
    if (typeof action === 'function') {
        res.end(action(req.params.payload)); // GOOD: `action` is either the `play` or the `pause` function from above
```

erik-krogh (commented):
LGTM :+1: 
Good work! 

Just two comments about the placement of the `// OK` comments.

javascript/ql/test/query-tests/Security/CWE-754/UnvalidatedDynamicMethodCallGood3.js
erik-krogh: That sounds right, a call to `get` on a map shouldn't be able to return something from the prototype.

I've seen a bunch of get implementations that are implemented like `function get(x) {return obj[x];}`, but if that's the case, then the query should pick up on the dynamic property read inside the `get` function.

erik-krogh:
And my suggestions made it so the `UnvalidatedDynamicMethodCall.qlref` test is failing.  
You need to run the test and update the expected output.

erik-krogh (approved):
:+1:  

I'll do a small evaluation to check if there is a performance regression (I highly doubt that), and then I'll merge it.

PR #7829, opened 2022-02-03
Ruby: add Parameter.getLogicalPosition()

asgerf:
Adds `Parameter.getLogicalPosition()` as a way to get the index of a positional argument, while taking into account that it may be declared after a splat parameter (and therefore have a non-constant logical index).

Commit-by-commit review recommended.

### DIFF
--- a/ruby/ql/lib/codeql/ruby/ast/Parameter.qll
+++ b/ruby/ql/lib/codeql/ruby/ast/Parameter.qll
@@ -19,6 +19,27 @@ class Parameter extends AstNode, TParameter {
   /** Gets the zero-based position of this parameter. */
   final int getPosition() { this = any(Callable c).getParameter(result) }
 
+  /** Holds if this parameter is declared after a positional `*splat` parameter. */
+  private predicate isAfterSplatParameter() {
+    exists(Callable c, int n |
+      c.getParameter(n) instanceof SplatParameter and
+      this = c.getParameter([n .. c.getNumberOfParameters() - 1])
+    )
+  }
+
+  /**
+   * Gets the zero-based position of this parameter, if it is a plain positional parameter.
+   *
+   * This corresponds to the index of a positional argument that can flow into this parameter.
+   *
+   * Has no result for non-positional parameters, splat parameters, or parameters declared after a splat parameter.
+   */
+  final int getLogicalPosition() {
+    (this instanceof SimpleParameter or this instanceof OptionalParameter) and
+    result = getPosition() and
+    not isAfterSplatParameter()
+  }
+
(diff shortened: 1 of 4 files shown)


asgerf:
Not sure if the AST node is the correct place for this, please advice.

Exposing the `ArgumentPosition` and `ParameterPosition` classes could also make for a more general interface. However, I believe the intention is to add synthetic argument/parameter positions in the future, which we'd probably want to keep internal.

ruby/ql/lib/codeql/ruby/ast/Parameter.qll
```diff
@@ -19,6 +19,27 @@ class Parameter extends AstNode, TParameter {
...
+  }
+
+  /**
+   * Gets the zero-based position of this parameter, if it is a plain positional parameter.
+   *
+   * This corresponds to the index of a positional argument that can flow into this parameter.
+   *
+   * Has no result for non-positional parameters, splat parameters, or parameters declared after a splat parameter.
+   */
+  final int getLogicalPosition() {
```
aibaars, line 37: I think this predicate should not be in the public API .

ruby/ql/lib/codeql/ruby/ast/Parameter.qll
```diff
@@ -19,6 +19,27 @@ class Parameter extends AstNode, TParameter {
...
+      this = c.getParameter([n .. c.getNumberOfParameters() - 1])
+    )
+  }
+
+  /**
+   * Gets the zero-based position of this parameter, if it is a plain positional parameter.
+   *
+   * This corresponds to the index of a positional argument that can flow into this parameter.
+   *
+   * Has no result for non-positional parameters, splat parameters, or parameters declared after a splat parameter.
```
aibaars, line 35: What about 
```
def foo (a,b=1,c)
end
```
I think in this case `c` does not have a "logical" position either.

ruby/ql/lib/codeql/ruby/ast/Parameter.qll
hvitved: I also think this should just be a private predicate in `DataFlowPrivate`.

ruby/ql/lib/codeql/ruby/ast/Parameter.qll
asgerf: I see. I was under the impression that the splat parameter actually was the only thing that could offset the index of other positional parameters by a variable amount.

asgerf:
As discussed, I'll close this PR and use `ParameterPosition` directly in the API graph PR instead, so we don't have to commit to anything in the public API.

PR #7830, opened 2022-02-03
C++: Fix re-evaluation in `cpp/return-stack-allocated-memory`

MathiasVP:
We were, once again, hit by the IR re-evaluation problem because we didn't import `semmle.code.cpp.valuenumbering.GlobalValueNumbering` before we imported the IR.

### DIFF
--- a/cpp/ql/src/Likely Bugs/Memory Management/ReturnStackAllocatedMemory.ql
+++ b/cpp/ql/src/Likely Bugs/Memory Management/ReturnStackAllocatedMemory.ql
@@ -14,6 +14,9 @@
  */
 
 import cpp
+// We don't actually use the global value numbering library in this query, but without it we end up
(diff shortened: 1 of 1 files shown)


geoffw0 (approved):
I'm a little sad this is necessary, but thanks for fixing it!

MathiasVP:
> I'm a little sad this is necessary, but thanks for fixing it!

Yeah, same :(

intrigus-lgtm:
OOI: could this problem be detected by a QL-for-QL query automatically?

MathiasVP:
> OOI: could this problem be detected by a QL-for-QL query automatically?

Maybe it could. As far as I understand the underlying problem, a query for this would need to model a lot of implementation details related to the order in which the QL compiler traverses files on disk.

Issue #6964, opened 2021-10-26
Horrible performance of Python PrintAst query on large databases


kjcolley7:
I'm working on a large Python database that has >10,000 Python source files. I'm using the VSCode extension to write and run queries as I'm developing them. I'd like to be able to use the AST Viewer. However, I have discovered (after hours of debugging) a horrible performance issue that results in this query attempting to compute global flow and points-to analysis, which shouldn't be required to simply print out the AST graph.

The reason this happens is due to the feature that displays the AST of regex strings. Here is the flow of references that causes this to happen:

printAst.ql:
```ql
import semmle.python.PrintAst
```

PrintAst.qll:
```ql
import semmle.python.RegexTreeView
```

RegexTreeView.qll:
```ql
private import semmle.python.regex
```

regex.qll:
```ql
/** A StrConst used as a regular expression */
class Regex extends RegexString {
  Regex() { used_as_regex(this, _) }
  //...
}
```

```ql
/**
 * Holds if `s` is used as a regex with the `re` module, with the regex-mode `mode` (if known).
 * If regex mode is not known, `mode` will be `"None"`.
 */
predicate used_as_regex(Expr s, string mode) {
  //...
  // Implementation uses DataFlow and the API graph
  //...
}
```

The impact of this issue is that attempting to view the AST of a small Python file doesn't complete running after more than 12 hours. By simply patching out the regex support from PrintAST.qll, the same query on the same database completes in under 1 minute.

I suggest simply removing the functionality for dumping the AST of regex strings. I feel that it is completely unnecessary, and it is certainly not worth the performance impact.

kjcolley7:
Maybe users who want this functionality could enable it in Customizations.qll by extending an abstract class and overriding a predicate like `predicate shouldPrintAstForRegexString(<string literal type>)`. Alternatively, it could be on by default but automatically disable when the database is large, maybe by counting the number of Python source files or lines of code.

tausbn:
We discussed this internally in the CodeQL Python team, and have agreed that the best approach for now is to disable the printing of regex ASTs.

I'm a bit curious about your use-case, however. If you're unable to run most of our analyses on your database, then you're really missing out on most of the power of our libraries.

Does this database contain multiple copies of the same files? (e.g. one or more copies of the Python standard library.)

This is one case where our analysis gets a lot slower, since it may not be able to disambiguate these copies properly, leading to a lot of apparent cross-talk between them.

kjcolley7:
[...]

PR #7831, opened 2022-02-03
Python: Remove `RegExpTerm` from PrintAST

RasmusWL:
Since this caused bad performance (as we had to evaluate points-to).

Fixes https://github.com/github/codeql/issues/6964

This approach was motivated by the comment on the issue from @tausbn:

> We discussed this internally in the CodeQL Python team, and have
> agreed that the best approach for now is to disable the printing of
> regex ASTs.

I tried to keep our RegExpTerm logic, but doing the fix below did not
work, and still evaluated RegExpTerm :| I guess we will just have to
revert this PR if we want it back

```diff
   TRegExpTermNode(RegExpTerm term) {
+    none() and
     exists(StrConst str | term.getRootTerm() = getParsedRegExp(str) and shouldPrint(str, _))
   }
```

### DIFF
--- a/python/ql/src/change-notes/2021-01-19-remove-regexp-from-printast.md
+++ b/python/ql/src/change-notes/2021-01-19-remove-regexp-from-printast.md
@@ -1,4 +1,4 @@
 ---
 category: fix
 ---
-* The [View AST functionality](https://codeql.github.com/docs/codeql-for-visual-studio-code/exploring-the-structure-of-your-source-code/) no longer prints detailed information about regular expression, greatly improving performance.
+* The [View AST functionality](https://codeql.github.com/docs/codeql-for-visual-studio-code/exploring-the-structure-of-your-source-code/) no longer prints detailed information about regular expressions, greatly improving performance.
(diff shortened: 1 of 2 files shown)


python/ql/src/change-notes/2021-01-19-remove-regexp-from-printast.md
```diff
@@ -0,0 +1,4 @@
+---
+category: fix
+---
+* The [View AST functionality](https://codeql.github.com/docs/codeql-for-visual-studio-code/exploring-the-structure-of-your-source-code/) no longer prints detailed information about regular expression, greatly improving performance.
```
nickrolfe, line 4:
```suggestion
* The [View AST functionality](https://codeql.github.com/docs/codeql-for-visual-studio-code/exploring-the-structure-of-your-source-code/) no longer prints detailed information about regular expressions, greatly improving performance.
```

tausbn (approved):
Thanks for doing this! Hopefully someday we can revive this code...

PR #7832, opened 2022-02-03
Java: Simplify model generator query using flow state.

aschackmull:


bmuskalla (approved):
That looks really good, thank you for doing this. Do you know how this affects performance (👀 guava and jdk)

btw: this is still a draft PR

aschackmull:
> Do you know how this affects performance (👀 guava and jdk)

Haven't checked yet.

java/ql/src/utils/model-generator/ModelGeneratorUtils.qll
```diff
@@ -151,11 +151,29 @@ predicate isRelevantTaintStep(DataFlow::Node node1, DataFlow::Node node2) {
...
+}
+
+string parameterNodeAsInput(DataFlow::ParameterNode p) {
+  result = parameterAccess(p.asParameter())
+  or
+  result = "Argument[-1]" and p instanceof DataFlow::InstanceParameterNode
+}
+
 string returnNodeAsOutput(TargetAPI api, ReturnNodeExt node) {
   if node.getKind() instanceof ValueReturnKind
```
atorralba, line 169: Isn't `api` unbound in this branch of the `if` statement?

java/ql/src/utils/model-generator/CaptureSummaryModels.ql
```diff
@@ -40,21 +37,51 @@ string captureQualifierFlow(TargetAPI api) {
...
+      isRelevantContent(c) and
+      state1 instanceof TaintRead and
+      state2 instanceof TaintRead
+    )
   }
 
-  override predicate isAdditionalTaintStep(DataFlow::Node node1, DataFlow::Node node2) {
-    isRelevantTaintStep(node1, node2)
+  override predicate isSanitizer(DataFlow::Node n) {
+    exists(Type t | t = n.getType() and not isRelevantType(t))
```
atorralba, line 84: Would it make sense to simplify this?

```suggestion
    not isRelevantType(n.getType()))
```

aschackmull: Not all nodes have types, so that's not the same thing. You can argue that all nodes ought to have types, and I'd agree, but that's not where we are at.

aschackmull: (I wrote that first, and it lost a lot of flow, which took me some time to debug)

java/ql/src/utils/model-generator/ModelGeneratorUtils.qll
aschackmull: Indeed it is. We're probably being saved by magic at the moment. But yeah, that ought to be fixed.

java/ql/src/utils/model-generator/CaptureSummaryModels.ql
aschackmull: Negating `isRelevantType` to `isIrrelevantType` could probably simplify things.

atorralba: Oh, I see. Thanks for explaining! Nothing against keeping it as it is then.

java/ql/src/utils/model-generator/ModelGeneratorUtils.qll
michaelnebel: From the looks of it, there is nothing that explicitly ties the returnnode to the TargetApi in captureSource (I am unfamiliar with the code, so this might be wrong). Will try and push a change that makes this explicit as also removes the TargetApi from the returnNodeAsOutput predicate.

michaelnebel:
@hvitved : Could you take a look at the last commit? It would be great to get this PR merged, such that we can start porting.

michaelnebel (dismissed):
Looks good to me!
All test pass and the output produced by the workflow actions also seems reasonable.

java/ql/src/utils/model-generator/CaptureSourceModels.ql
```diff
@@ -44,7 +44,8 @@ string captureSource(TargetAPI api) {
     config.hasFlow(source, sink) and
     sourceNode(source, kind) and
     api = source.getEnclosingCallable() and
```
hvitved, line 46: @bmuskalla : Do you know if it was deliberate to restrict source models to those where the existing source is directly in the callable that we generate the summary for? That will exclude wrapper methods, i.e.
```java
string source() { // will be included
  return knownSource();
}

string sourceWrapped() { // will not be included
  return source();
}
```

If not, I would expect that `api = source.getEnclosingCallable()` should be removed.

hvitved (commented):
LGTM, just one question for Benny.

java/ql/src/utils/model-generator/CaptureSourceModels.ql
bmuskalla: No, that was not intentional. But also not covered [by the tests](https://github.com/github/codeql/blob/main/java/ql/test/utils/model-generator/p/Sources.java) so it might be good to have a case of that

hvitved: @michaelnebel : Could you add a test that verifies my suspicion? And if I am correct, then remove the `api = source.getEnclosingCallable()` line?

michaelnebel: @hvitved: That sounds like an improvement we can try and make, when the implementation is ported.
Should we just merge this PR now?

hvitved: OK, as long as we also backport the improvement to Java.

michaelnebel: Looks like we wrote stuff at the same time :-) I will try and make the improvement.

java/ql/src/utils/model-generator/ModelGeneratorUtils.qll
```diff
@@ -151,11 +151,29 @@ predicate isRelevantTaintStep(DataFlow::Node node1, DataFlow::Node node2) {
   )
 }
 
-string returnNodeAsOutput(TargetAPI api, ReturnNodeExt node) {
+predicate isRelevantContent(DataFlow::Content f) {
+  isRelevantType(f.(DataFlow::FieldContent).getField().getType()) or
```
joefarebrother, line 155: Should synthetic fields also be covered?

hvitved (dismissed):
(Needs to be rebased to resolve merge conflict.)

michaelnebel:
> (Needs to be rebased to resolve merge conflict.)

Hmm, I am not allowed to force push after the rebase. Does this require to make a new PR against my own fork?

java/ql/src/utils/model-generator/ModelGeneratorUtils.qll
michaelnebel: Yes, you are right. Will add it.

michaelnebel:
Thank you for all the reviews. Unless anyone finds any critical mistakes, it will be a good time to merge the content of this PR and make improvements on follow up PRs.

hvitved (approved):

aschackmull:
Thanks for finishing this PR in my absence.

PR #7839, opened 2022-02-03
C++: fix IR generation for constructor base inits when no constructor is present.

rdmarsh2:
When a class has no virtual members and no constructor, but an explicitly written destructor, constructors for classes that inherit from it will have a generated call to the nonexistent constructor, leading to any later entries in the initializer list being dropped in the IR. This PR works around the issue by adding a minimal translation for the broken call, allowing later entries in the list to be included in the IR.


MathiasVP (approved):
LGTM! We should probably create an extractor issue for this.

PR #7841, opened 2022-02-04
C++: Update C++ variable hiding test

jketema:
Structured bindings are now handled better, so the false negative related to structured bindings is now a true positive.


MathiasVP (approved):
LGTM!

PR #7842, opened 2022-02-04
Misc: Streamline `consistency-queries/qlpack.yml`

RasmusWL:
Wanted to add this for Python, so that was the day I learned about `.codeqlmanifest.json` :open_mouth: 

I looked at the `qlpack.yml` for the other languages, and since they were all slightly different, I decided to streamline them. I also made them all use the preferred `dependencies` instead of `libraryPathDependencies`.

### DIFF
--- a/csharp/ql/consistency-queries/qlpack.yml
+++ b/csharp/ql/consistency-queries/qlpack.yml
@@ -1,4 +1,5 @@
 name: codeql-csharp-consistency-queries
+groups: [csharp, test, consistency-queries]
 dependencies:
   codeql/csharp-all: "*"
 extractor: csharp
--- a/csharp/ql/consistency-queries/qlpack.yml
+++ b/csharp/ql/consistency-queries/qlpack.yml
@@ -1,5 +1,4 @@
 name: codeql-csharp-consistency-queries
 dependencies:
   codeql/csharp-all: "*"
-  codeql/csharp-queries: "*"
 extractor: csharp
--- a/csharp/ql/consistency-queries/qlpack.yml
+++ b/csharp/ql/consistency-queries/qlpack.yml
@@ -1,6 +1,5 @@
 name: codeql-csharp-consistency-queries
-version: 0.0.0
-libraryPathDependencies:
-  - codeql/csharp-all
-  - codeql/csharp-queries
+dependencies:
+  codeql/csharp-all: "*"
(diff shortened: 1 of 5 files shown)


csharp/ql/consistency-queries/qlpack.yml
```diff
@@ -1,6 +1,5 @@
 name: codeql-csharp-consistency-queries
-version: 0.0.0
-libraryPathDependencies:
-  - codeql/csharp-all
-  - codeql/csharp-queries
+dependencies:
+  codeql/csharp-all: "*"
+  codeql/csharp-queries: "*"
```
hvitved, line 4: Hmm, I am not sure that this line is needed.

python/ql/consistency-queries/qlpack.yml
```diff
@@ -0,0 +1,5 @@
+name: codeql/python-consistency-queries
+groups: [python, consistency-queries]
```
hvitved, line 2: Should we add this to the other packs as well?

aibaars: Should we also add the `test` group? If I'm not mistaken QL packs in the `test` group are not published.  @aeisenberg is that right?

RasmusWL: Good point :+1: Truth be told, I made the commits for streamlining the other languages and the one for adding the Python consistency queries on two separate days (originally in 2 separate branches), so I had forgotten about this detail.

csharp/ql/consistency-queries/qlpack.yml
RasmusWL: I'm not sure either. We can commit this suggestion if you would like to try and remove it?

```suggestion
```

hvitved: Yes, let's give it a try.

python/ql/consistency-queries/qlpack.yml
aeisenberg: Yes. Since they are now part of the workspace, they will be picked up when publishing.

Note that if you remove the `groups` property, the they will be ignored. So, either add the `test` group here and to all consistency query packs. Or remove this group.

I'd prefer adding the test group here and else where since we may want to specify this in the future

hvitved (approved):

PR #7844, opened 2022-02-04
Python: Update `.expected` to support Python 3.10

RasmusWL:
This PR should be merged together with internal PR that changes the test to use Python 3.10 (so the tests on this PR will fail).

The changes required _does_ mean that 3 tests won't pass when using previous python versions, but I don't see any way around it.


RasmusWL:
superseded by https://github.com/github/codeql/pull/14791

PR #7846, opened 2022-02-04
C# 10: Tuple deconstruction.

michaelnebel:
In this PR, we
* Refactor existing tuple data flow test cases to use the default taint configuration.
* Introduce data flow tests for mixed assignment and declaration tuples.
* Introduce tests for the mixed assignment and declaration tuples in general.


hvitved (approved):
Thanks for converting the existing test to use inline expectations.

PR #7849, opened 2022-02-04
C/C++: Useless Test : verification of "Fully converted" Type

Yonah125:
This fix false positive caused by the left operand of the equal and non-equals being the same, but the fully converted left operand not.

For example, based on Packet++

curLayer != NULL && dynamic_cast<TLayer*>(curLayer) == NULL

The query used to detect this as a false positive, because the left operand type of equal and non-equals are the same, but the fully converted type are not.


cpp/ql/src/experimental/Best Practices/UselessTest.ql
```diff
@@ -38,6 +39,8 @@ where
   ) and
   eq.getRightOperand() instanceof Literal and
   ne.getRightOperand() instanceof Literal and
+  eq.getLeftOperand().getFullyConverted().getType() =
+    ne.getLeftOperand().getFullyConverted().getType() and
```
MathiasVP, line 43: I think you'd want to call `getUnspecifiedType` instead of `getType` to resolve `typedef`s and ignore specifiers like `const` and friends.

cpp/ql/src/experimental/Best Practices/UselessTest.ql
```diff
@@ -1,3 +1,4 @@
+
```
MathiasVP, line 1: Looks like the file needs to be autoformatted 😄.

Yonah125: Should be better now

cpp/ql/src/experimental/Best Practices/UselessTest.ql
Yonah125: DIdn't know about that, thanks !

MathiasVP:
Autoformat check still fails:
```
2022-02-04T18:19:38.9841194Z Formatting ql/cpp/ql/src/experimental/Best Practices/UselessTest.ql.
2022-02-04T18:19:38.9859546Z ##[error]ql/cpp/ql/src/experimental/Best Practices/UselessTest.ql would change by autoformatting.
```

Yonah125:
Very strange, in order to auto format I use this command : 

codeql query format UselessTest.ql --output=UT.ql

Any problem in it ?

MathiasVP:
> Very strange, in order to auto format I use this command :
> 
> codeql query format UselessTest.ql --output=UT.ql
> 
> Any problem in it ?

That does look correct. Stnrage. In any case, to my feeble human eyes, it looks like https://github.com/github/codeql/pull/7849/commits/61dc9ef12e4cd96698d58f9e89362526c269563a fixes the issue. I've restarted the PR checks. Let's see what it says 🤞. If that passes I'll merge the fix.

MathiasVP (approved):
LGTM!

Yonah125:
A little question : is there any way to test this request on, for example, a thousand of project ? 
is there any easy way to adapt this query for other languages ?

PR #7867, opened 2022-02-06
Java: Timing attacks while comparing the headers value

ahmed532009:
A constant-time algorithm should be used for checking the value of headers. In other words, the comparison time should not depend on the content of the input, Otherwise, an attacker may be able to implement a timing attacks that may reveal the value of sensitive headers like <code>x-auth-token</code>


java/ql/src/experimental/Security/CWE/CWE-208/TimingAttackAgainstHeader.qhelp
```diff
@@ -0,0 +1,28 @@
+<!DOCTYPE qhelp PUBLIC "-//Semmle//qhelp//EN" "qhelp.dtd">
+<qhelp>
+
+<overview>
+<p>
+A constant-time algorithm should be used for checking the value of headers.
```
pwntester, line 6:
```suggestion
A constant-time algorithm should be used for checking the value of sensitive headers.
```

intrigus-lgtm:
Out of interest, do you know of any successful timing attack against a real application or an academic paper that describes a possible attack?

ahmed532009:
Sorry, but i can not find any real application or an academic paper that describes this type of attacks.

java/ql/src/experimental/Security/CWE/CWE-208/TimingAttackAgainstHeader.qhelp
```diff
@@ -0,0 +1,28 @@
...
+<recommendation>
+<p>
+Use <code>MessageDigest.isEqual()</code> method to check the value of headers.
+If this method is used, then the calculation time depends only on the length of input byte arrays,
+and does not depend on the contents of the arrays.
+</p>
+</recommendation>
+<example>
+<p>
+The following example uses <code>Arrays.equals()</code> method for validating a csrf token .
```
smowton, line 22:
```suggestion
The following example uses <code>Arrays.equals()</code> method for validating a csrf token.
```

java/ql/src/experimental/Security/CWE/CWE-208/TimingAttackAgainstHeader.qhelp
```diff
@@ -0,0 +1,28 @@
...
+This method implements a non-constant-time algorithm.
+Both the message and the signature come from an untrusted HTTP request:
+</p>
+<sample src="UnsafecsrfComparison.java" />
```
smowton, line 26: Include a fixed example

java/ql/src/experimental/Security/CWE/CWE-208/TimingAttackAgainstHeader.qhelp
```diff
@@ -0,0 +1,28 @@
+<!DOCTYPE qhelp PUBLIC "-//Semmle//qhelp//EN" "qhelp.dtd">
+<qhelp>
+
+<overview>
+<p>
+A constant-time algorithm should be used for checking the value of headers.
+In other words, the comparison time should not depend on the content of the input. 
+An attacker may be able to forge the value of the header.
```
smowton, line 8:
```suggestion
Otherwise timing information could be used to infer the header's expected, secret value.
```

java/ql/src/experimental/Security/CWE/CWE-208/TimingAttackAgainstHeader.ql
```diff
@@ -0,0 +1,66 @@
+/**
+ * @name Timing attack against headers value
+ * @description A constant-time algorithm should be used for checking the value of headers. 
+ *              In other words, the comparison time should not depend on the content of the input
+ *              Otherwise, an attacker may be able to implement a timing attacks that may reveal the value of sensitive headers
```
smowton, line 5:
```suggestion
 * @description Use of a non-constant-time verification routine to check the value of an HTTP header,
 *              possibly allowing a timing attack to infer the header's expected value.
```

java/ql/src/experimental/Security/CWE/CWE-208/TimingAttackAgainstHeader.ql
```diff
@@ -0,0 +1,66 @@
+/**
+ * @name Timing attack against headers value
```
smowton, line 2:
```suggestion
 * @name Timing attack against header value
```

java/ql/src/experimental/Security/CWE/CWE-208/TimingAttackAgainstHeader.ql
```diff
@@ -0,0 +1,66 @@
...
+}
+
+private predicate isNonConstantTimeEqualsCall(Expr firstObject, Expr secondObject) {
+  exists(NonConstantTimeEqualsCall call |
+    firstObject = call.getQualifier() and
+    secondObject = call.getAnArgument()
+    or
+    firstObject = call.getAnArgument() and
+    secondObject = call.getQualifier()
+  )
```
smowton, line 34: If you're always expecting argument 0, use `getArgument(0)` instead of `getAnArgument()`.

If the expected position of the thing being compared varies from method to method, check which method is being called, otherwise you'll produce false positives regarding the method's other arguments.

java/ql/src/experimental/Security/CWE/CWE-208/TimingAttackAgainstHeader.ql
```diff
@@ -0,0 +1,66 @@
...
+}
+class NonConstantTimeComparisonSink extends DataFlow::Node {
+  Expr anotherParameter;
+  NonConstantTimeComparisonSink() {
+      isNonConstantTimeEqualsCall(this.asExpr(), anotherParameter)
```
smowton, line 39:
```suggestion
  NonConstantTimeComparisonSink() {
      isNonConstantTimeEqualsCallArgument(this.asExpr())
```

java/ql/src/experimental/Security/CWE/CWE-208/TimingAttackAgainstHeader.ql
```diff
@@ -0,0 +1,66 @@
...
+  }
+}
+
+private predicate isNonConstantTimeEqualsCall(Expr firstObject, Expr secondObject) {
+  exists(NonConstantTimeEqualsCall call |
+    firstObject = call.getQualifier() and
+    secondObject = call.getAnArgument()
+    or
+    firstObject = call.getAnArgument() and
+    secondObject = call.getQualifier()
```
smowton, line 33:
```suggestion
private predicate isNonConstantEqualsCallArgument(Expr e) {
  exists(NonConstantTimeEqualsCall call |
    e = [call.getQualifier(), call.getAnArgument()]
```

smowton:
Please add tests under https://github.com/github/codeql/tree/main/java/ql/test/experimental/query-tests/security

You could use or adapt the qhelp example as the test. Then, abbreviate the example so it contains just the minimum code required to show what we're illustrating -- in this case, something like `if(request.getHeader("x-csrf-token").equals(expectedToken)) { ... } // Bad` is probably enough to illustrate the bad case, for example -- the example doesn't need to include null checks, iterating over cookies, etc.

Marcono1234:
`NonConstantTimeCheckOnSignatureQuery.qll` already [defines logic for non-constant times checks](https://github.com/github/codeql/blob/a448db11b57e7bac553af29fd625c39dd86e45e6/java/ql/src/experimental/Security/CWE/CWE-208/NonConstantTimeCheckOnSignatureQuery.qll#L291). Would it make sense to reuse this (possibly by moving it to a new shared library file)?

(Please only consider this a suggestion; I am not a member of this project.)

java/ql/src/experimental/Security/CWE/CWE-208/ComparingValueOfSensetiveHeader.java
```diff
@@ -0,0 +1,20 @@
+import javax.servlet.http.HttpServletRequest;
+import java.nio.charset.StandardCharsets;
+import java.security.MessageDigest;
+
+
```
smowton, line 5:
```suggestion
```

java/ql/src/experimental/Security/CWE/CWE-208/ComparingValueOfSensetiveHeader.java
```diff
@@ -0,0 +1,20 @@
...
+private boolean UnsafecsrfComparison(String csrfTokenInCookie) {
+         if(csrfTokenInCookie == null || !csrfTokenInCookie.equals(request.getHeader("X-CSRF-TOKEN"))) { // BAD
+                 return false;
+         }
+}
+
```
smowton, line 11:
```suggestion
```

smowton:
The test should have a `.qlref` file indicating the query to run, and a `.expected` file (`codeql test run your/test/dir` runs the test and creates a `.actual` file with the real test results, which can be renamed to `.expected` with `codeql test run your/test/dir --learn` or simply by manually renaming it. After doing this `codeql test run your/test/dir` should succed.

ahmed532009:
Hi [Marcono1234](https://github.com/Marcono1234),
<code>NonConstantTimeCheckOnSignatureQuery.qll</code> defines a sink that compares two objects or arrays or strings, but i need only a sink that compare strings value because the headers above require this type.

ahmed532009:
Hi , 
i get some trouble in running the test, can you help me pls?

atorralba:
> i get some trouble in running the test, can you help me pls?

How are you trying to run the test? What's the error you're getting?

ahmed532009:
i tried but i encounter this,
Capture

smowton:
This indicates your Java doesn't compile.

The suggest `--show-extractor-output` doesn't fully work for Java, but you can see the errors with

```
$ cat java/ql/test/experimental/query-tests/security/CWE-208/CWE-208.testproj/log/javac-output*.log

[2022-02-25 10:29:44] [javac-output-5066] private boolean UnsafecsrfComparison(String csrfTokenInCookie) {
[2022-02-25 10:29:44] [javac-output-5066]         ^
[2022-02-25 10:29:44] [javac-output-5066] ComparingValueOfSensetiveHeader.java:9: error: class, interface, or enum expected
[2022-02-25 10:29:44] [javac-output-5066]          }
[2022-02-25 10:29:44] [javac-output-5066]          ^
[2022-02-25 10:29:44] [javac-output-5066] ComparingValueOfSensetiveHeader.java:15: error: class, interface, or enum expected
[2022-02-25 10:29:44] [javac-output-5066]           if (csrfTokenInRequest == null || !MessageDigest.isEqual(
[2022-02-25 10:29:44] [javac-output-5066]           ^
[2022-02-25 10:29:44] [javac-output-5066] ComparingValueOfSensetiveHeader.java:19: error: class, interface, or enum expected
[2022-02-25 10:29:44] [javac-output-5066]          }
[2022-02-25 10:29:44] [javac-output-5066]          ^
[2022-02-25 10:29:44] [javac-output-5066] TimingAttackAgainstSignagure/Test.java:13: error: duplicate class: Test
[2022-02-25 10:29:44] [javac-output-5066] public class Test {
[2022-02-25 10:29:44] [javac-output-5066]              ^
```

smowton:
Basically the java code / qlref file in the `CWE-208` directory is causing the test runner to treat what are really two individual tests like one large one, meaning the two different Test.java files clash. Then the `ComparingValueOfSensetiveHeader.java` file doesn't compile since it's a snippet not a complete .java file. Deleting files in the root of the test directory fixes things; I've pushed a commit doing that.

ahmed532009:
@smowton thanks for your help

smowton:
Looks like this has lost the qlref and java file for the test?

ahmed532009:
Can we fix that ?
In spite of that,  i have added this files tell me if there is a problem
with that

Le ven. 25 fév. 2022 à 15:48, Chris Smowton ***@***.***> a
écrit :

smowton:
Looks reasonable now. I pushed a cosmetic fix renaming the qlref file.

smowton:
Applied more fixes:
* qhelp used the wrong sample file name
* autoformat
* test expectations were out of date

smowton:
Notes from the final commits, for future reference--

1. Use `codeql query format` to get QL passing the autoformat checks
2. Use `codeql generate query-help` to check qhelp is valid
3. Watch out with `path-problem` queries, you have to provide a description for each `$@` placeholder. Example: `select source, source.getNode(), sink.getNode(), "Main problem message with placeholder1: $@ and placeholder2: $@", placeholderElement1, "foo", placeholderElement2, "bar"` will render as `Main problem message with placeholder1: foo and placeholder2: bar`, with `foo` and `bar` linking to placeholderElement1 and 2 respectively.

smowton (approved):

Marcono1234:
@smowton, what is your and the other maintainers' opinion on squashing commits when merging (see also https://github.com/github/codeql/discussions/6484)?
For example this PR included 39 new commits into `main`. I am not trying to blame you or ahmed532009 here, during development of a PR it is (in my opinion) totally fine to have as many commits as needed. But my concern is whether these commits really have to be preserved in the history of `main`. Especially as this repository evolves this will become a bigger problem because the Git history keep increasing, making it more expensive to check out this repository and making it more difficult to find relevant commits.

smowton:
Personally I'd say commits like that don't matter much, because it's easy to view the history of `main` showing only first parents (by convention, the previous commit on `main`) if you want to.

PR #7872, opened 2022-02-07
Collect framework coverage on demand

bmuskalla:
This also turns the relevant query into a metric query to avoid problematic analysis results as the counts are interpreted as diagnostic results:
```
| Diagnostics for framework coverage | 132 results (101 unknowns, 10 errors, 21 warnings) |
```

Collecting framework coverage should not happen on every CodeQL analysis. It should rather happen on-demand as it is not tied to a specific codebase. This PR excludes the query from all our suites and sets up an adhoc workflow to run it daily and feed the metric data into the sarif upload (for later post processing)


henrymercer (commented):
Couple of suggestions.  Does this still belong in the `Diagnostics` directory or should we move to `Metrics` or something else?

bmuskalla:
>  Does this still belong in the Diagnostics directory or should we move to Metrics or something else?
Yes, indeed. I applied you suggestions and going to move the file to the LOC query.

henrymercer (approved):
LGTM

PR #7875, opened 2022-02-07
Ruby: put AST node locations in a single table

nickrolfe:
This should make `getLocation()` on generated nodes faster.

I generated the upgrade/downgrade scripts with some quick hacks to the generator program, and only hand-edited the results for formatting. That means reviewers should feel confident that, for the upgrades/downgrades, they only need to check:
- `{ruby,erb}_ast_node_parent` <-> `{ruby,erb}_ast_node_info` upgrades/downgrades.
- `{ruby,erb}_tokeninfo` modifications
- and just a small sample of the `*_def` table modifications.

I tested them by:
1. Running all the qltests with `--search-path=main-extractor-pack:ql` so that they would use an extractor from `main` and upgrade the dbs on the fly.
2. Building a large Ruby database with an extractor from `main`, upgrading it, then downgrading it, and checking that all the relations in the original and downgraded databases were byte-for-byte identical.


aibaars (approved):
LGTM

PR #7876, opened 2022-02-07
JS: recognize more startswith sanitizers for path-injection queries

erik-krogh:
Gets a TN for CVE-2021-23484 

[An evaluation seems clean](https://github.com/github/codeql-dca-main/tree/data/erik-krogh/pr-7876-ca5f91e__nightly__CustomSuite/reports).

### DIFF
--- a/javascript/ql/lib/semmle/javascript/security/dataflow/TaintedPathCustomizations.qll
+++ b/javascript/ql/lib/semmle/javascript/security/dataflow/TaintedPathCustomizations.qll
@@ -513,12 +513,23 @@ module TaintedPath {
 
     override predicate blocks(boolean outcome, Expr e) {
       member = "relative" and
-      e = pathCall.getArgument(1).asExpr() and
+      e = maybeGetJoinArg(pathCall.getArgument(1)).asExpr() and
       outcome = startsWith.getPolarity().booleanNot()
       or
       not member = "relative" and
-      e = pathCall.getArgument(0).asExpr() and
+      e = maybeGetJoinArg(pathCall.getArgument(0)).asExpr() and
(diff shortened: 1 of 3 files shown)


javascript/ql/lib/semmle/javascript/security/dataflow/TaintedPathCustomizations.qll
```diff
@@ -513,13 +513,24 @@ module TaintedPath {
...
       outcome = startsWith.getPolarity().booleanNot()
       or
       not member = "relative" and
-      e = pathCall.getArgument(0).asExpr() and
+      e = maybeGetJoinArg(pathCall.getArgument(0)).asExpr() and
       outcome = startsWith.getPolarity()
     }
+
+    bindingset[e]
+    private DataFlow::Node maybeGetJoinArg(DataFlow::Node e) {
```
esbena, line 525: I'm having trouble relating the name and the semantics of this predicate. Could you add a bit of qldoc? Alternative names, based on semantics: `maybeGetLastJoinArg`, `getPathSuffix`?

erik-krogh: I ended up with `maybeGetPathSuffix`.  
I want the `maybe` in there to capture that the predicate also acts as the identity function.

esbena (dismissed):
(do you want to deal with the ql-for-ql alerts in this PR?)

erik-krogh:
> (do you want to deal with the ql-for-ql alerts in this PR?)

(We could use some automatic suggestions on that).

esbena (approved):

PR #7878, opened 2022-02-07
Shared: Switch to dot-separated access paths in summary specs

asgerf:
This changes the syntax of input/output specifications to the dot-separated `A.B` style instead of `B of A`.

It briefly retains support for both syntaxes so the tests pass at each commit, but support for the `of`-style is removed entirely at the end.

`AccessPathSyntax.qll` is shared between JS and the shared data-flow libraries. Since parsing is no longer entirely trivial, it made sense to share it. I also think we should also share other aspects of parsing (such as the `n1..n2` argument ranges), but I'd like to separate that from the mega-PR.

One change from the previous parsing is that an empty access path now has zero tokens; previously, the empty path consisted of a single empty token. This mattered for how `Annotated` summaries were implemented in Java, where the empty input/output spec must be interpreted in a certain way. This was preserved by special-casing the empty access path in `interpretInput` and `interpretOutput`. This doesn't feel great, but it seems like this special case was already there, it's just more explicit now.

I've started DCA evaluations for all four affected languages (against Differences code-scanning.qls).


asgerf:
The evaluations aren't done yet, but in the interest of time, it might be worth starting reviews before that.

I'd like to nominate some reviewers, but others are obviously welcome to join in:
- @aschackmull to review `AccessPathSyntax.qll`, `FlowSummaryImpl.qll` and the Java-specific changes
- @erik-krogh to review JS-specific changes
- @hvitved to review the Ruby and C#-specific changes

erik-krogh (dismissed):
JS :+1: (But that's the easy review).  

For the rest of you: I've added a backref for how the `x of y -> y.x` rewrite was made.

asgerf:
Evaluations:
- [Java](https://github.com/github/codeql-dca-main/tree/data/asgerf/dot-separated-access-paths-java/reports) appears to show a failure and a slowdown, but both of these can be observed in nightly runs as well, so probably not related to this PR. I've manually inspected the tuple counts for the greatest slowdown and could not find anything amiss.
- [C#](https://github.com/github/codeql-dca-main/tree/data/asgerf/dot-separated-a__Differences__code-scanning__3/reports) looks fine, albeit with a slight bias towards a slowdown. I investigated the tuple counts from the greatest slowdown, and again, could not find anything amiss.
- [Ruby](https://github.com/github/codeql-dca-main/tree/data/asgerf/dot-separated-a__Differences__code-scanning/reports) looks fine, uneventful
- [JS](https://github.com/github/codeql-dca-main/tree/data/asgerf/dot-separated-a__Differences__code-scanning__1/reports) also fine, uneventful

hvitved (changes requested):
LGTM, thanks for making this change Asger. Could I please ask you to also update `getComponentStackCsv` in `FlowSummaryImpl.qll`(and consequently the expected test output in `ql/test/library-tests/dataflow/library` and `ql/test/library-tests/frameworks/EntityFramework`)?

javascript/ql/lib/semmle/javascript/frameworks/data/internal/Shared.qll
```diff
@@ -57,6 +57,7 @@
  */
 
 private import Impl as Impl
+import AccessPathSyntax
```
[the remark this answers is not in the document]
hvitved: 👍 Let's just keep it as is.

java/ql/lib/semmle/code/java/dataflow/internal/AccessPathSyntax.qll
```diff
@@ -0,0 +1,82 @@
...
+  }
+
+  /** Gets the number of tokens on the path (if there are no syntax errors). */
+  int getNumToken() {
+    result = count(int n | exists(getRawToken(this, n))) and
+    not hasSyntaxError()
+  }
+
+  /** Gets the `n`th-last token, with 0 being the last token. */
+  AccessPathToken getLastToken(int n) { result = getToken(getNumToken() - 1 - n) }
```
smowton, line 52: Noting ql-for-ql warnings re: implicit `this`

java/ql/lib/semmle/code/java/dataflow/internal/AccessPathSyntax.qll
```diff
@@ -0,0 +1,82 @@
...
+  string getName() { result = this.getPart(1) }
+
+  /**
+   * Gets the argument list, such as `1,2` from `Member[1,2]`,
+   * or has no result if there are no arguments.
+   */
+  string getArgumentList() { result = this.getPart(2) }
+
+  /** Gets the `n`th argument to this token, such as `x` or `y` from `Member[x,y]`. */
+  string getArgument(int n) { result = this.getArgumentList().splitAt(",", n) }
```
smowton, line 75: Should we be flexible on whitespace while we have the chance? Should `Token[arg1, arg2]` mean the same as `Token[arg1,arg2]`?

java/ql/lib/semmle/code/java/dataflow/internal/FlowSummaryImpl.qll
```diff
@@ -811,69 +812,46 @@ module Private {
...
-        or
-        result = interpretComponentSpecific(c)
+    private SummaryComponent interpretComponent(AccessPathToken token) {
+      exists(ParameterPosition pos |
+        parseArg(token, pos) and result = SummaryComponent::argument(pos)
       )
+      or
+      exists(ArgumentPosition pos |
+        parseParam(token, pos) and result = SummaryComponent::parameter(pos)
+      )
```
smowton, line 838: These ParameterPosition / ArgumentPosition uses seem backwards

java/ql/lib/semmle/code/java/dataflow/internal/FlowSummaryImpl.qll
```diff
@@ -811,69 +812,46 @@ module Private {
...
-    private predicate interpretSpec(string spec, int idx, SummaryComponentStack stack) {
-      exists(string c |
-        relevantSpec(spec) and
-        specLength(spec, idx + 1) and
-        specSplit(spec, c, idx) and
-        stack = SummaryComponentStack::singleton(interpretComponent(c))
-      )
+    private predicate interpretSpec(AccessPath spec, int idx, SummaryComponentStack stack) {
+      idx = spec.getNumToken() - 1 and
+      stack = SummaryComponentStack::singleton(interpretComponent(spec.getLastToken(idx)))
```
smowton, line 854: getNumToken - 1 *and* a use of getLastToken? Do these cancel out?

smowton:
Also would be interested to hear arguments for a regexpFind (which might bind and find matches other than at the start of an access path specifier) plus a cross-check based on match lengths, vs. a stateful parse starting at the LHS, with a predicate argument tracking the offset where the previous access path element parse finished, and so where the next one should begin. I imagine the combination of greed and lookaheads in the regexp are supposed to produce exactly the same behaviour, but my regex-fu isn't sufficient to know that for sure.

java/ql/lib/semmle/code/java/dataflow/internal/AccessPathSyntax.qll
asgerf: I'm on the fence here.

On one hand I don't think the spaces actually improve readability, and I also wouldn't want to need an auto-formatter to resolve arguments about CSV rows should be formatted.

On the other hand, allowing spaces is probably the least surprising behaviour; could save someone a lot of time if they simply assumed spaces would work. However, this can also be handled by emitting a CSV validation error when an argument has trailing spaces.

Ultimately I'm leaning towards not allowing spaces, but I'm open to changing it if someone feels strongly about changing it.

java/ql/lib/semmle/code/java/dataflow/internal/FlowSummaryImpl.qll
asgerf: I tried to preserve the original indexing scheme, but yeah it ended up looking weird.

I've pushed a commit that replaces the whole thing with indexes that start at the beginning and use `getToken` instead of `getLastToken`. Could you take another look?

java/ql/lib/semmle/code/java/dataflow/internal/FlowSummaryImpl.qll
asgerf: It's supposed to be backwards (see the diff). An `Argument[n]` token results in a synthesized parameter, and `Parameter[n]` results in a synthesized argument (to a callback).

asgerf:
> Also would be interested to hear arguments for a regexpFind [...]

The `regexpFind` call parses the whole string with a single string operation, without producing unneeded intermediate string values.

If you try to parse one token at a time, you can't use regexps efficiently, because you can't tell the engine where to start matching (it will start the beginning and tell _you_ where each match began, that's different). So to parse the `n+1`th token you'd have to strip off the first `n` tokens and parse from the remaining string; that's not as efficient. You could manually parse one character at a time, or rely on `indexOf`, but that seems quite fiddly to me.

Also, the `regexpFind` solution can be factored it into a predicate with a `bindingset` (which you can't do with a recursive predicate), though in the current formulation we don't actually need this.

asgerf:
> Could I please ask you to also update getComponentStackCsv in FlowSummaryImpl.qll(and consequently the expected test output in ql/test/library-tests/dataflow/library and ql/test/library-tests/frameworks/EntityFramework)?

Argh, I overlooked this request first time around. Should be fixed in https://github.com/github/codeql/pull/7878/commits/caf5931f152f0c6fd5f4c86143fb62f391a3c660

java/ql/lib/semmle/code/java/dataflow/internal/FlowSummaryImpl.qll
smowton: Thanks, looks good! AFAICT getLastToken is now dead code, and I'd like to see whitespace significance avoided wherever possible; otherwise lgtm.

java/ql/lib/semmle/code/java/dataflow/internal/AccessPathSyntax.qll
asgerf: Added support for surrounding spaces

java/ql/lib/semmle/code/java/dataflow/internal/FlowSummaryImpl.qll
asgerf: Removed `getLastToken`

asgerf:
Force-pushed to resolve conflicts in the Ruby libraries. Only the commit `Ruby: update CSV rows to dot-separated syntax` was affected.

smowton (approved):

erik-krogh (approved):

PR #7879, opened 2022-02-07
Docs: Note codeql-go needs an install step before use

smowton:


felicitymay (approved):
Many thanks for updating the docs to address this user feedback 💖

PR #7880, opened 2022-02-07
QL for QL: sync changes from Ruby

nickrolfe:
In particular, the changes from #7875 that modify the dbscheme to put location columns in a single table, making `getLocation()` faster.

As discussed on Slack, I won't write/generate the upgrade/downgrade scripts, since the primary use case of QL for QL is CI checks that build fresh databases.


erik-krogh (approved):
Thanks :+1:

PR #7881, opened 2022-02-07
CPP: Fix performance for cpp/cleartext-transmission

geoffw0:
There was a moderate performance explosion on some projects (such as `nlohmann__json`), caused by the `Encrypted` class.  Tested locally, DCA run to follow.

### DIFF
--- a/cpp/ql/src/Security/CWE/CWE-311/CleartextTransmission.ql
+++ b/cpp/ql/src/Security/CWE/CWE-311/CleartextTransmission.ql
@@ -168,15 +168,11 @@ class NetworkRecv extends NetworkSendRecv {
   override Recv target;
 }
 
-predicate encryptionFunction(Function f)
-{
-  f.getName()
-  .toLowerCase()
-  .regexpMatch(".*(crypt|encode|decode|hash|securezero).*")
+predicate encryptionFunction(Function f) {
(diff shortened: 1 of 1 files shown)


cpp/ql/src/Security/CWE/CWE-311/CleartextTransmission.ql
```diff
@@ -168,6 +168,14 @@ class NetworkRecv extends NetworkSendRecv {
   override Recv target;
 }
 
+predicate encryptionFunction(Function f) {
+  f.getName().toLowerCase().regexpMatch(".*(crypt|encode|decode|hash|securezero).*")
+}
+
+predicate encryptionType(UserType t) {
+  t.getName().toLowerCase().regexpMatch(".*(crypt|encode|decode|hash|securezero).*")
+}
```
MathiasVP, line 177: If this is for a join-order fix you probably want to mark them as `pragma[noinline]`.

geoffw0: Done (and improves performance a bit further)!

MathiasVP: Perfect! Do you happen to have tuple counts before and after this change?

geoffw0:
DCA shows a significant speedup and no change in results. :)

PR #7884, opened 2022-02-07
C++: fix hasImplicitCopyConstructor for templates

rdmarsh2:
Fixes some cases in instantiations of templates with manually written copy constructors or copy assignment operators where hasImplicitCopyConstructor would incorrectly hold.

### DIFF
--- a/cpp/ql/lib/change-notes/2022-03-10-template-implicit-copy.md
+++ b/cpp/ql/lib/change-notes/2022-03-10-template-implicit-copy.md
@@ -1,0 +1,4 @@
+---
+category: minorAnalysis
+---
+* `hasImplicitCopyConstructor` and `hasImplicitCopyAssignmentOperator` now correctly handle implicitly-deleted operators in templates.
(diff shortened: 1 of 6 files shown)


rdmarsh2:
We're not at the point where all non-deleted copy constructors have an instantiation, and I don't fully understand what the current state is, so it's hard to trust any changes I might make. @igfoo and @nickrolfe, do you know more about what's been changed with implicit copy constructors/assignment operators since that query was originally written?

jbj:
If we extract all _deleted_ copy constructors, then perhaps the classes with an _implicit_ copy constructors are exactly the ones where it's not _deleted_ and not _user-supplied_. But I'm not sure if extraction is precise enough to rely on that, especially around templates.

igfoo:
I'm not certain OTTOMH, but I would imagine that we only extract non-deleted copy constructors in an instantiation if that constructor is actually called (`instantiations_entity_set` tracks which instantiations are actually reachable).

rdmarsh2:
It looks like there's at least [one non-template case](https://github.com/github/codeql/blob/c81f2661a3fc8f7b08f2f5947cd760029ebf142d/cpp/ql/test/query-tests/Best%20Practices/RuleOfTwo/RuleOfTwo.cpp#L174) where no implicit copy constructor can be generated but no deleted copy constructor is extracted. I don't think we'll be able to delete all the QL logic but we can simplify it a fair bit by relying on what's actually extracted.

MathiasVP (approved):
LGTM!

PR #7890, opened 2022-02-08
Ruby/QL: add `unique` annotation on `node` column

nickrolfe:
@alexet pointed out that the `node` column in the `*_ast_node_info` tables should be a keyset, since each node has exactly one row in that table.

This may allow the optimiser/evaluator to make better decisions.


aibaars (approved):

hvitved (approved):

nickrolfe:
I'll merge once DCA finishes, just in case there are any surprises.

nickrolfe:
DCA results seem unremarkable, so merging.

PR #7892, opened 2022-02-08
JS: Add a `isNaN` sanitizer, and use it in queries that already had a typeof check

erik-krogh:
Gets a TN for CVE-2021-45459 

[Evaluation looks clean](https://github.com/github/codeql-dca-main/tree/data/erik-krogh/nanSan__nightly__security-extend/reports) (ignore the backrefs).  

In `UnsafeShellCommandConstructionQuery.qll` and `UnvalidatedDynamicMethodCallQuery.qll` I also added a missing sanitizierGuard to the `isSanitizerGuard` predicate.

### DIFF
--- a/javascript/ql/lib/semmle/javascript/security/dataflow/UnsafeShellCommandConstructionQuery.qll
+++ b/javascript/ql/lib/semmle/javascript/security/dataflow/UnsafeShellCommandConstructionQuery.qll
@@ -25,7 +25,7 @@ class Configuration extends TaintTracking::Configuration {
   override predicate isSanitizerGuard(TaintTracking::SanitizerGuardNode guard) {
     guard instanceof PathExistsSanitizerGuard or
     guard instanceof TaintTracking::AdHocWhitelistCheckSanitizer or
-    guard instanceof NaNGuard or
+    guard instanceof NumberGuard or
     guard instanceof TypeOfSanitizer
   }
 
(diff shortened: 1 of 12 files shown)


javascript/ql/lib/semmle/javascript/security/dataflow/UnsafeShellCommandConstructionQuery.qll
```diff
@@ -24,7 +24,9 @@ class Configuration extends TaintTracking::Configuration {
 
   override predicate isSanitizerGuard(TaintTracking::SanitizerGuardNode guard) {
     guard instanceof PathExistsSanitizerGuard or
-    guard instanceof TaintTracking::AdHocWhitelistCheckSanitizer
+    guard instanceof TaintTracking::AdHocWhitelistCheckSanitizer or
+    guard instanceof NumberGuard or
+    guard instanceof TypeOfSanitizer
```
esbena, line 29: Is this `TypeOfSanitizer` supposed to be a part of this PR?

esbena (commented):
LGTM. I'm surprised we haven't needed this sanitizer earlier.

javascript/ql/lib/semmle/javascript/security/dataflow/UnsafeShellCommandConstructionQuery.qll
erik-krogh: Yes, I noticed that we defined the guard in the Customizations file, but we never included it in the `isSanitizerGuard` predicate.   
I did a similar thing in `UnvalidatedDynamicMethodCallQuery`.

esbena (approved):

PR #7894, opened 2022-02-08
Python: Normalise string prefixes

tausbn:
This is the external part of a change to the extractor that normalises the `UBRF` prefix characters to their lower-case parts.

This PR is expected to fail. Please see the internal PR for the successful test run.


yoff (approved):
LGTM

PR #7895, opened 2022-02-08
Upgrade scripts testing: set initial dbschemes

cklin:
This commit sets initial dbschemes for `cpp`, `csharp`, `java`, `javascript`, and `python` so that automated testing for upgrade scripts would also cover legacy upgrades.

Sources of the new "initial" dbschemes (paths are within CodeQL CLI releases):
- C++: `legacy-upgrades/upgrades/454f1e15151422355049dc4f1f0486a03baeffef/old.dbscheme`
- C#: `legacy-upgrades/upgrades/b1166f9a16534e138c00b4384890af55e26073cd/old.dbscheme`
- Java: `legacy-upgrades/upgrades/57bb881e645fa5759e6fb186a4fe211eb8f1dba7/old.dbscheme`
- JS: `legacy-upgrades/upgrades/75fcf6336d7ced8405d54b504857b525a97ba8c2/old.dbscheme`
- Python: `legacy-upgrades/upgrades/3f9d0d3de95633ed3a937c8ace752dd99fd91a94/old.dbscheme`


csharp/ql/lib/upgrades/initial/semmlecode.csharp.dbscheme
```diff
@@ -1,1707 +1,752 @@
...
- * C# dbscheme
- */
-
-/** ELEMENTS **/
-
-@element = @declaration | @stmt | @expr | @modifier | @attribute | @namespace_declaration
-         | @using_directive | @type_parameter_constraints | @external_element
-         | @xmllocatable | @asp_element | @namespace;
-
-@declaration = @callable | @generic | @assignable;
```
michaelnebel, line 91: This looks like a significant change to the "initial" dbscheme. 
(1) Does the replacement initial "match" the old dbscheme in one of the upgrade folders? 
(2) Does it require that the hash value is re-calculated for the naming purposes of that particular upgrade folder.
(3) Or was the initial db scheme here just plain wrong 😄 ?

hvitved: I would also like to know what was wrong with the original initial DB scheme for C#?

erik-krogh (approved):

michaelnebel (approved):
LGTM!

MathiasVP (approved):
C++ 👍

smowton (approved):

cklin:
> The Java initial dbscheme doesn't match `query-cli/upgrades/initial-dbschemes/semmlecode.dbscheme`. I think I saw something about 5 early upgrades having to be dropped, but I can't find it now; is that the reason? Should `query-cli/upgrades/initial-dbschemes` be updated or removed, or will that happen in a future PR?

Yes, that is correct.

java/ql/lib/upgrades/initial/semmlecode.dbscheme
```diff
@@ -0,0 +1,648 @@
...
+			int location: @location_default ref);
+
+@xmllocatable = @xmlcharacters | @xmlelement | @xmlcomment | @xmlattribute | @xmldtd | @file | @xmlnamespace;
+
+@top = @element | @locatable | @folder;
+
+/* *** Well-Founded Orders *** */
+
+/* The type hierarchy order */
+order:typeOrder(3);
```
Marcono1234, line 615: It looks like this is causing an error for the QL extractor:
> [build-stdout] ERROR /home/runner/work/codeql/codeql/java/ql/lib/upgrades/initial/semmlecode.dbscheme:615: parse error

See https://github.com/github/codeql/runs/5160181021?check_suite_focus=true#step:8:37
But I am not sure if that is relevant.

tausbn: Ah, this is not relevant to the present PR, but it _is_ very relevant to the QL extractor. Thanks for bringing this to our attention!

PR #7912, opened 2022-02-09
JS: convert more type-trackers to API-graphs

erik-krogh:
I was looking at some of our TypeTrackers and I realized that we had some that could be expressed using API-graphs.


asgerf:
Cool! LGTM if the evaluation works out.

erik-krogh:
> Cool! LGTM if the evaluation works out.

[It looks fine](https://github.com/github/codeql-dca-main/tree/data/erik-krogh/pr-7912-12d31d7__nightly__security-extend/reports).

asgerf (approved):

PR #2252, opened 2019-11-04
JS: Parse regular expressions from string literals

asger-semmle:
This is an experiment to see what we can do if we parse all string literals as regular expressions.

The overhead in extraction time and database size is about 5%.

I've adapted the security-relevant regexp queries to use ASTs. Preserving the exact behaviour of these queries would be possible, but also defeat the purpose of the experiment, so I've tried to take full advantage of the AST to see how it works out.

To be fair, some of these adjustments could possibly be back-ported to the original version using meta-regexps, like the fact that we now flag `^foo|bar$` as two misleading anchor precedence alerts.

[Results so far](https://git.semmle.com/asger/dist-compare-reports/tree/regexp_1572633946781).

PR #7917, opened 2022-02-09
Ruby: IncompleteHostnameRegExp.ql

aibaars:
This pull request is a port of JavaScripts [IncompleteHostnameRegExp.ql](https://github.com/github/codeql/blob/main/javascript/ql/src/Security/CWE-020/IncompleteHostnameRegExp.ql) query. Most of the query and documentation are a shameless [copy](https://github.com/github/codeql/pull/7917/commits/1ad6e96fc80cff9d583f8d14cc13cc4543b1b774) of the original.

The pull request is split into:
* a commit that copies the Javascript implementation: https://github.com/github/codeql/pull/7917/commits/1ad6e96fc80cff9d583f8d14cc13cc4543b1b774
* a commit that adapts the implementation for Ruby:  https://github.com/github/codeql/pull/7917/commits/a221e28faada9bdd047f2fcc16a864ec472f3eac


ruby/ql/src/queries/security/cwe-020/IncompleteHostnameRegExp.ql
```diff
@@ -0,0 +1,55 @@
...
+bindingset[pattern]
+predicate isIncompleteHostNameRegExpPattern(string pattern, string hostPart) {
+  hostPart =
+    pattern
+        .regexpCapture("(?i).*" +
+            // an unescaped single `.`
+            "(?<!\\\\)[.]" +
+            // immediately followed by a sequence of subdomains, perhaps with some regex characters mixed in, followed by a known TLD
+            "([():|?a-z0-9-]+(\\\\)?[.](" + commonTopLevelDomainRegex() + "))" + ".*", 1)
+}
```
hmac, line 32: I wonder, is it worth adding a shared `HostnameRegExpUtils.qll` file to share this with the python implementation? I notice the [JS version](https://github.com/aibaars/codeql/blob/main/javascript/ql/src/Security/CWE-020/HostnameRegexpShared.qll) has a different implementation. I don't know whether we (the dynamic teams) want to consider consolidating them?

aibaars: Good idea.

ruby/ql/src/queries/security/cwe-020/IncompleteHostnameRegExp.ql
```diff
@@ -0,0 +1,55 @@
...
+
+predicate regularExpression(Expr n, string pattern) {
+  (
+    n instanceof RegExpLiteral
+    or
+    exists(API::Node regexp, DataFlow::CallNode callNode |
+      regexp = API::getTopLevelMember("Regexp") and
+      (callNode = regexp.getAnInstantiation() or callNode = regexp.getAMethodCall("compile")) and
+      n = callNode.getArgument(0).asExpr().getExpr()
+    )
```
nickrolfe, line 42: I notice this disjunct is almost identical to [this charpred](https://github.com/github/codeql/blob/1a91a79b5b503f51ddc6ed2eb53b70b8189e897f/ruby/ql/lib/codeql/ruby/security/performance/RegExpInjectionCustomizations.qll#L56-L60). I can't quite decide whether or not it's complicated enough to warrant having a shared predicate between the two.

ruby/ql/src/change-notes/2022-02-10-incomplete-hostname-regexp.md
```diff
@@ -0,0 +1,5 @@
+---
+category: newQuery
+---
+lgtm,codescanning
+* Added a new query, `rb/incomplete-hostname-regexp`. The query finds instances where a hostname is incompletely sanitized due to enescaped character in a regular expression.
```
nickrolfe, line 5:
```suggestion
* Added a new query, `rb/incomplete-hostname-regexp`. The query finds instances where a hostname is incompletely sanitized due to an unescaped character in a regular expression.
```

ruby/ql/src/queries/security/cwe-020/IncompleteHostnameRegExp.qhelp
```diff
@@ -0,0 +1,69 @@
...
+            when it accidentally succeeds.
+
+        </p>
+    </overview>
+
+    <recommendation>
+        <p>
+
+            Escape all meta-characters appropriately when constructing
+            regular expressions for security checks, pay special attention to the
```
nickrolfe, line 33:
```suggestion
            regular expressions for security checks, and pay special attention to the
```

ruby/ql/src/change-notes/2022-02-10-incomplete-hostname-regexp.md
```diff
@@ -0,0 +1,5 @@
+---
+category: newQuery
+---
+lgtm,codescanning
```
hvitved, line 4: Remove

hvitved (changes requested):

ruby/ql/src/queries/security/cwe-020/IncompleteHostnameRegExp.ql
aibaars: Yes, I wondered the same.

ruby/ql/src/queries/security/cwe-020/IncompleteHostnameRegExp.ql
asgerf: A bit of history: The one you copied originated in JS, written before we parsed string literals as RegExps. When we [started extracting strings](https://github.com/github/codeql/pull/2252) as RegExps, it was rewritten to its current form.

I'd say the new version is better, if you have ASTs for regexps.

tausbn (approved):
:+1: for the Python docs changes.

python/ql/lib/semmle/python/RegexTreeView.qll
```diff
@@ -48,6 +49,19 @@ newtype TRegExpParent =
...
   TRegExpBackRef(Regex re, int start, int end) { re.backreference(start, end) }
 
+/**
+ * Provides regular expression patterns.
+ */
+module RegExpPatterns {
+  /**
+   * Gets a pattern that matches common top-level domain names in lower case.
+   */
+  string commonTLD() {
```
tausbn, line 59: This name is not really in line with our style guide. At the very least, it should be `Tld` not `TLD`. Also, since this function returns a result, it should perhaps start with `get`.

python/ql/lib/semmle/python/RegexTreeView.qll
```diff
@@ -751,6 +767,9 @@ class RegExpGroup extends RegExpTerm, TRegExpGroup {
    */
   int getNumber() { result = re.getGroupNumber(start, end) }
 
+  /** Holds if this is a capture group. */
+  predicate isCapture() { not exists(this.getNumber()) }
```
tausbn, line 771: I may be completely misunderstanding this predicate, but is that `not` correct? Is this not expressing "not an unnamed capture group" instead?

python/ql/lib/semmle/python/RegexTreeView.qll
```diff
@@ -1009,3 +1028,24 @@ class RegExpBackRef extends RegExpTerm, TRegExpBackRef {
 
 /** Gets the parse tree resulting from parsing `re`, if such has been constructed. */
 RegExpTerm getParsedRegExp(StrConst re) { result.getRegex() = re and result.isRootTerm() }
+
+/**
+ * A node whose value may flow to a position where it is interpreted
+ * as a part of a regular expression.
+ */
+class RegExpPatternSource extends DataFlow::CfgNode {
```
tausbn, line 1036: I don't think this class should be implemented in this file. I would rather that it live somewhere closer to the rest of the data-flow implementation.

python/ql/src/Security/CWE-020/HostnameRegexpShared.qll
```diff
@@ -0,0 +1,202 @@
...
+
+private import HostnameRegexpSpecific
+
+/**
+ * Holds if the given constant is unlikely to occur in the origin part of a URL.
+ */
+predicate isConstantInvalidInsideOrigin(RegExpConstant term) {
+  // Look for any of these cases:
+  // - A character that can't occur in the origin
+  // - Two dashes in a row
```
tausbn, line 14: This one surprises me. What about URLs with [Punycode](https://en.wikipedia.org/wiki/Punycode) such as https://xn--wrdle-vua.dk?

tausbn (changes requested):
A few comments and suggestions. I did not review the query in detail. I have no doubt that the original JavaScript query is excellent. :slightly_smiling_face: 
My only lingering concern here is the fact that the string `"\."` in JavaScript gets interpreted as just a string containing a period (the backslash disappears) whereas in Python, it is interpreted as the string `"\\."` (i.e. the backslash is interpreted as just a literal backslash, since `\.` is not a valid escape). However, I'm guessing that this detail is abstracted away by relying on the regex library anyway, and so it should be fine.

python/ql/lib/semmle/python/RegexTreeView.qll
erik-krogh: :heart: :+1: 

> it should perhaps start with get.   

And since it has multiple results it should probably start with `getA`. So `getACommonTld()`.

erik-krogh (approved):
JS changes :+1:

python/ql/lib/semmle/python/RegexTreeView.qll
```diff
@@ -48,6 +49,19 @@ newtype TRegExpParent =
   /** A back reference */
   TRegExpBackRef(Regex re, int start, int end) { re.backreference(start, end) }
 
+/**
+ * Provides regular expression patterns.
```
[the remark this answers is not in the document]
aibaars: Would `Provides utility predicates related to regular expressions.` be any better?

python/ql/lib/semmle/python/RegexTreeView.qll
aibaars: I believe all capture groups have a number, but only some have a name. Is that assumption incorrect?

aibaars: Oh wait, the `not` makes no sense at all. You're right.

python/ql/lib/semmle/python/RegexTreeView.qll
aibaars: For Javascript it is in javascript/ql/lib/semmle/javascript/Regexp.qll and this file looks like the Python variant of that. I don't mind moving it to another place, but could you tell me where you'd like me to move it to?

python/ql/src/Security/CWE-020/HostnameRegexpShared.qll
aibaars: @erik-krogh , do you know ^ ?

erik-krogh: I don't know...  
My best guess is that's it's sufficiently uncommon that we don't need to consider it.  
 
@asgerf introduced the JS implementation as part of [this PR](https://github.com/github/codeql/pull/2252).  
Maybe he knows something?

asgerf: I know double dashes are valid in the hostname, but in practice it's an indicator that this regexp is not used to check a hostname string, at least not in a security-relevant context.

javascript/ql/lib/semmle/javascript/Regexp.qll
```diff
@@ -999,13 +999,13 @@ predicate isInterpretedAsRegExp(DataFlow::Node source) {
...
 /**
- * Provides regular expression patterns.
+ * Provides utility predicates related to regular expressions.
  */
 module RegExpPatterns {
   /**
    * Gets a pattern that matches common top-level domain names in lower case.
    */
-  string commonTLD() {
+  string getACommonTld() {
```
erik-krogh, line 1008: Perhaps a deprecated alias for the old name?

ruby/ql/src/queries/security/cwe-020/IncompleteHostnameRegExp.qhelp
```diff
@@ -0,0 +1,72 @@
...
+			The check is however easy to bypass because the unescaped
+			<code>.</code> allows for any character before
+			<code>example.com</code>, effectively allowing the redirect to go to
+			an attacker-controlled domain such as <code>wwwXexample.com</code>.
+
+		</p>
+		<p>
+
+			Address this vulnerability by escaping <code>.</code>
+			appropriately: <code>regex = /((www|beta)\.)?example\.com/</code>.
```
nickrolfe, line 62: I think it still needs the caret:
```suggestion
			appropriately: <code>regex = /^((www|beta)\.)?example\.com/</code>.
```

nickrolfe (commented):
Otherwise, LGTM.

ruby/ql/src/queries/security/cwe-020/IncompleteHostnameRegExp.qhelp
aibaars: You're right. The example snippet does have the `^` . I'll make the suggested change for both Javascript and Ruby.

nickrolfe: It's important because, without it, the code is still unsafe:

```
irb(main):001:0> "wwwXexample.com".match?(/((www|beta)\.)?example\.com/)
=> true
```

aibaars: Indeed, same for Javascript.

nickrolfe (approved):

nickrolfe (approved):

erik-krogh (approved):

hvitved (approved):

alexrford (approved):

PR #7918, opened 2022-02-09
[Java]: Improving ZipSlip query

ahmed532009:
Zip Slip is a widespread critical archive extraction vulnerability, allowing attackers to write arbitrary files on the system, typically resulting in remote command execution, in this PR i try to improve the zipslip query.


smowton:
I'm afraid the example you've added is already recognised by the ZipSlip query without modification. The reason is line 32,  `this.getDeclaringType().getASupertype*() = archiveEntry`, which says that any type derived from either Java's `ZipEntry` or Apache's `ArchiveEntry` will be covered. Since all the types you add in this PR are derived from one of those two, there is no need to explicitly list them.

I will however update the qhelp to clarify that other kinds of archive, including Jar and Tar, are indeed already handled.

If you find a test case which is not covered by the existing ZipSlip query, please let us know.

PR #7919, opened 2022-02-10
Ruby: recognise additional form for OpenURI

hmac:
This gets us one step closer to https://github.com/github/codeql-ruby-CVE-coverage/issues/43


nickrolfe (approved):
Nice

PR #7920, opened 2022-02-10
Ruby: Add String flow summaries

hmac:
This seemed fairly straightforward, so I thought I'd give it a go.

This PR is branched from #7886, which I think is why not all the CI tests are running on it. Once that PR is merged, we should hopefully get a full CI run.

### DIFF
--- a/ruby/ql/lib/codeql/ruby/frameworks/core/String.qll
+++ b/ruby/ql/lib/codeql/ruby/frameworks/core/String.qll
@@ -1,0 +1,533 @@
+/** Provides flow summaries for the `String` class. */
+
+private import codeql.ruby.AST
+private import codeql.ruby.ApiGraphs
+private import codeql.ruby.DataFlow
+private import codeql.ruby.dataflow.FlowSummary
+private import codeql.ruby.dataflow.internal.DataFlowDispatch
+
+// TODO: the way we interpret `preservesValue` in this module may not be
+// correct: we assume that if the input string appears intact in the output,
+// then value is preserves. This means that we consider appending or prepending
+// characters to the string to be value-preserving.
+// We may want to be stricter here, and define value-preserving as when the
+// output string exactly matches the input string.
+/**
+ * Provides flow summaries for the `String` class.
+ *
+ * The summaries are ordered (and implemented) based on
+ * https://docs.ruby-lang.org/en/3.1/String.html.
+ */
+module String {
+  /**
+   * Value-preserving flow from the receiver to the return value.
+   */
+  private predicate valueIdentityFlow(string input, string output, boolean preservesValue) {
+    input = "Receiver" and
+    output = "ReturnValue" and
+    preservesValue = true
+  }
+
+  /**
+   * Taint-preserving (but not value-preserving) flow from the receiver to the return value.
+   */
+  private predicate taintIdentityFlow(string input, string output, boolean preservesValue) {
+    input = "Receiver" and
+    output = "ReturnValue" and
+    preservesValue = false
+  }
+
+  private class NewSummary extends SummarizedCallable {
+    NewSummary() { this = "String.new" }
+
+    override MethodCall getACall() {
+      result = API::getTopLevelMember("String").getAnInstantiation().getExprNode().getExpr()
+    }
+
+    override predicate propagatesFlowExt(string input, string output, boolean preservesValue) {
+      input = "Argument[0]" and
+      output = "ReturnValue" and
+      preservesValue = true
+    }
+  }
+
+  private class TryConvertSummary extends SummarizedCallable {
+    TryConvertSummary() { this = "String.try_convert" }
+
+    override MethodCall getACall() {
+      result =
+        API::getTopLevelMember("String").getAMethodCall("try_convert").getExprNode().getExpr()
+    }
+
+    override predicate propagatesFlowExt(string input, string output, boolean preservesValue) {
+      input = "Argument[0]" and
+      output = "ReturnValue" and
+      preservesValue = true
+    }
+  }
+
+  /**
+   * A flow summary for the `String#%` method.
+   */
+  private class FormatSummary extends SummarizedCallable {
+    private MethodCall mc;
+
+    FormatSummary() { this = "%" and mc.getMethodName() = this }
+
+    override MethodCall getACall() { result = mc }
+
+    override predicate propagatesFlowExt(string input, string output, boolean preservesValue) {
+      input = ["Receiver", "Argument[0]", "ArrayElement of Argument[0]"] and
+      output = "ReturnValue" and
+      preservesValue = true
+    }
+  }
+
+  // TODO: * + << <=>
+  /**
+   * A flow summary for the `String#b` method.
+   */
+  private class BSummary extends SummarizedCallable {
+    private MethodCall mc;
+
+    BSummary() { this = "b" and mc.getMethodName() = this }
+
+    override MethodCall getACall() { result = mc }
+
+    override predicate propagatesFlowExt(string input, string output, boolean preservesValue) {
+      valueIdentityFlow(input, output, preservesValue)
+    }
+  }
+
+  /**
+   * A flow summary for the `String#byteslice` method.
+   */
+  private class BytesliceSummary extends SummarizedCallable {
+    private MethodCall mc;
+
+    BytesliceSummary() { this = "byteslice" and mc.getMethodName() = this }
+
+    override MethodCall getACall() { result = mc }
+
+    override predicate propagatesFlowExt(string input, string output, boolean preservesValue) {
+      valueIdentityFlow(input, output, preservesValue)
+    }
+  }
+
+  /**
+   * A flow summary for `String#capitalize(!)`.
+   */
+  private class CapitalizeSummary extends SummarizedCallable {
+    private MethodCall mc;
+
+    CapitalizeSummary() { this = ["capitalize", "capitalize!"] and mc.getMethodName() = this }
+
+    override MethodCall getACall() { result = mc }
+
+    override predicate propagatesFlowExt(string input, string output, boolean preservesValue) {
+      input = "Receiver" and
+      preservesValue = false and
+      output = "ReturnValue"
+    }
+  }
+
+  /**
+   * A flow summary for `String#center`, `String#ljust` and `String#rjust`.
+   */
+  private class CenterSummary extends SimpleSummarizedCallable {
+    CenterSummary() { this = ["center", "ljust", "rjust"] }
+
+    override predicate propagatesFlowExt(string input, string output, boolean preservesValue) {
+      (
+        input = "Receiver" and
+        output = "ReturnValue"
+        or
+        input = "Argument[1]" and
+        output = "ReturnValue"
+      ) and
+      preservesValue = false
+    }
+  }
+
+  /**
+   * A flow summary for the `String#chomp`, `String#chomp!`, `String#chop` and `String#chop!` methods.
+   */
+  private class ChompSummary extends SummarizedCallable {
+    private MethodCall mc;
+
+    ChompSummary() { this = ["chomp", "chomp!", "chop", "chop!"] and mc.getMethodName() = this }
+
+    override MethodCall getACall() { result = mc }
+
+    override predicate propagatesFlowExt(string input, string output, boolean preservesValue) {
+      input = "Receiver" and
+      preservesValue = false and
+      (
+        output = "ReturnValue"
+        or
+        this = ["chomp!", "chop!"] and output = "Receiver"
+      )
+    }
+  }
+
+  // TODO: we already have a summary for Array#clear. Check that it applies correctly to String#clear.
+  /**
+   * A flow summary for `String#concat` and `String#prepend`.
+   */
+  private class ConcatSummary extends SimpleSummarizedCallable {
+    ConcatSummary() { this = ["concat", "prepend"] }
+
+    override predicate propagatesFlowExt(string input, string output, boolean preservesValue) {
+      input = ["Receiver", "Argument[_]"] and
+      output = ["ReturnValue", "Receiver"] and
+      preservesValue = true
+    }
+  }
+
+  /**
+   * A flow summary for `String#delete(!)`, `String#delete_prefix(!)` and `String#delete_suffix(!)`.
+   */
+  private class DeleteSummary extends SimpleSummarizedCallable {
+    DeleteSummary() { this = ["delete", "delete_prefix", "delete_suffix"] + ["", "!"] }
+
+    override predicate propagatesFlowExt(string input, string output, boolean preservesValue) {
+      taintIdentityFlow(input, output, preservesValue)
+    }
+  }
+
+  /**
+   * A flow summary for `String#downcase(!)`, `String#upcase` and `String#swapcase(!)`.
+   */
+  private class DowncaseSummary extends SimpleSummarizedCallable {
+    DowncaseSummary() { this = ["downcase", "upcase", "swapcase"] + ["", "!"] }
+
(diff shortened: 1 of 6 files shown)


ruby/ql/lib/codeql/ruby/frameworks/core/String.qll
```diff
@@ -0,0 +1,542 @@
...
+   * A flow summary for `String#each_line` and `String#lines`.
+   */
+  private class EachLineSummary extends SummarizedCallable {
+    MethodCall mc;
+
+    EachLineSummary() { this = ["each_line", "lines"] and mc.getMethodName() = this }
+
+    override MethodCall getACall() { result = mc }
+
+    predicate hasBlock() { exists(Block b | b = mc.getBlock()) }
```
asgerf, line 235: There's a common misunderstanding about how fields work in CodeQL, which ultimately is why this predicate won't work as expected.

Declaring a field in the class is basically syntactic sugar for wrapping every single predicate body with an extra `exists` clause with some conditions attached to it. It doesn't mean the value of `this` is associated with unique field values.

The `hasBlock()` predicate holds if _any_ call to `each_line` or `lines` has a block argument. Doing `this.hasBlock()` in a member predicate does in any way guarantee that the value of `mc`, as seen by that predicate, actually has a block argument.

The only way to handle this is to make sure the value of `this` encodes knowledge of whether the block argument exists. Then `getACall()` should only find the calls where the block argument respectively exists or doesn't exist.

If you want to use a field, it could look like this:
```codeql
  private class EachLineSummary extends SummarizedCallable {
    string hasBlock;

    EachLineSummary() { this = "each_line_" + hasBlock and hasBlock = ["block", "noblock"] }

    override MethodCall getACall() {
      result.getMethodName() = ["each_line", "lines"] and
      if exists(result.getBlock()) then
        hasBlock = "block"
      else
        hasBlock = "noblock"
    }

    override predicate propagatesFlowExt(string input, string output, boolean preservesValue) {
      preservesValue = false and
      input = "Receiver" and
      (
        output = "Parameter[0] of BlockArgument"
        or
        hasBlock = "noblock" and output = "ArrayElement[?] of ReturnValue"
        or
        hasBlock = "block" and output = "ReturnValue"
      )
    }
```

It may also be nicer to split this into two classes, but importantly, the two classes still need to have distinct `this` values.

Note that the `getACall()` and `propagatesFlowExt` predicate are separate because, by design, the `propagatesFlowExt` should not depend on the precise call-site. This is because `propagatesFlowExt` must be computed an an earlier stage than `getACall()`. (`getACall()` can in general depend on the data-flow IR, and the data-flow IR depends on `propagatesFlowExt`).

asgerf (commented):
I haven't done a detailed review of the individual summaries, as there are two underlying issues to address first -- see inline comments.

I'm happy to discuss more in person if you like. I feel like this review is not the ideal place to explain how fields work.

asgerf:
@hmac The syntax change for MaD access paths just merged. I've generated a [commit](https://github.com/github/codeql/commit/1766c90d69f9ad1daef05df4abfbb82df9e07c35) that updates this PR to the new syntax as well. Would you like me to push it straight to this PR?

ruby/ql/lib/codeql/ruby/frameworks/core/String.qll
hmac: Ugh, yeah. This is not the first time I've made this mistake and probably won't be the last. Splitting into multiple classes sounds like the way to go.

ruby/ql/lib/codeql/ruby/frameworks/core/String.qll
```diff
@@ -0,0 +1,542 @@
...
+private import codeql.ruby.DataFlow
+private import codeql.ruby.dataflow.FlowSummary
+private import codeql.ruby.dataflow.internal.DataFlowDispatch
+
+// TODO: the way we interpret `preservesValue` in this module may not be
+// correct: we assume that if the input string appears intact in the output,
+// then value is preserves. This means that we consider appending or prepending
+// characters to the string to be value-preserving.
+// We may want to be stricter here, and define value-preserving as when the
+// output string exactly matches the input string.
```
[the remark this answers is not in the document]
hmac: > As a rule of thumb, value-preservation is mainly about whether the flow can be used to connect a store and load operation. In other words, whether the fields (or "contents") of an incoming value should be allowed to propagate through.

I don't really follow this. But if we define "value-preserving" to mean that the input value and output value are identical, then I agree that probably all of these summaries want to set it to false. I'll update them accordingly.

hmac:
@asgerf thanks for taking a look and for the explanations. I've (hopefully) fixed the fields issue, ensured that `preservesValue` is only true for methods that are close to the identity function, and rewritten the access paths to use dot syntax.

hmac:
@alexrford adding summaries for `String#(g)sub` has caused some extra results in the `CleartextLogging` tests which look to be duplicates of existing results - https://github.com/github/codeql/pull/7920/commits/07c70adde50733f2b8838c4f0b7703318e535def has more detail. Do you think this is something we can/should fix? I'm not sure how much of a problem it would be in the real world.

asgerf:
> `preservesValue` is only true for methods that are close to the identity function, and rewritten the access paths to use dot syntax.

When dealing with strings, I'd say it's still better to set `preservesValue=false` everywhere. Even if it's _possible_ for the string value to propagate through unchanged, it should be false. That aspect of data flow simply isn't what value-preservation is supposed to handle.

The primary purpose of `preservesValue` is propagation of fields (or "contents" as the general concept is called in the shared data flow library). As long as strings don't have any associated fields, there's no reason to use value-preserving steps; it will only enable spurious flow to propagate further.

Some queries may care about strings propagating entirely unchanged from A to B, but keep in mind that it's a fairly niche use-case. It's best to let the query deal with that on its own, as there's usually other domain-specific rules to factor in anyway. Usually this is done by either (1) sticking to value-preserving flow and manually adding the string steps that are believed to preserve the value (e.g. `to_str` calls), or (2) using taint-flow and adding sanitizers for the string operations that don't preserve enough of the original string value.

alexrford:
> @alexrford adding summaries for `String#(g)sub` has caused some extra results in the `CleartextLogging` tests which look to be duplicates of existing results - [07c70ad](https://github.com/github/codeql/commit/07c70adde50733f2b8838c4f0b7703318e535def) has more detail. Do you think this is something we can/should fix? I'm not sure how much of a problem it would be in the real world.

This is due to https://github.com/github/codeql/blob/31d214d5ee8881164a830878b785f9daf11fe915/ruby/ql/lib/codeql/ruby/security/CleartextLoggingCustomizations.qll#L222 considering the RHS of each matching assignment as a fresh `Source` - we do this because the variable name in the assignment is part of the criteria for if the RHS is a potential password value, but it does make these cases a bit awkward if the RHS already flows from some other `Source` (in this case another potential password assignment).

I don't see many great solutions to this without significantly restructuring the query:

 - We could make it s.t. the `Source` is any `Expr e | not e instanceof NonCleartextPassword` and convert the existing `Source`s instead into `Sanitizer`s if they appear to have non-suspicious names (e.g. `foo = "8c4f0b770331"` would be a sanitizer, but `passwd = "8c4f0b770331"` would not be). However, I'd be nervous about performance and ending up with a lot of FPs due to the large number of sources.
 - We could amend the existing sources to not include cases where the value is likely to flow from another source (e.g. if the RHS of this assignment is a call to `sub`/`gsub`/`freeze` etc. and the receiver is a variable access with a password like name), but it feels like a hack and probably won't be very robust against further flow summary improvements.
 - We could exclude sources that have flow from other sources explicitly, but may run into problems with non-monotonic recursion.

On balance, leaving this as-is seems okay to me. The test example is a little bit contrived and I think that the code scanning UI would display this sensibly as a single alert at `stdout_logger.info password_masked_ineffective_sub` that has multiple paths (where the paths to that alert happen to overlap).

hmac:
> The primary purpose of preservesValue is propagation of fields (or "contents" as the general concept is called in the shared data flow library). As long as strings don't have any associated fields, there's no reason to use value-preserving steps; it will only enable spurious flow to propagate further.

> Some queries may care about strings propagating entirely unchanged from A to B, but keep in mind that it's a fairly niche use-case. It's best to let the query deal with that on its own, as there's usually other domain-specific rules to factor in anyway. Usually this is done by either (1) sticking to value-preserving flow and manually adding the string steps that are believed to preserve the value (e.g. to_str calls), or (2) using taint-flow and adding sanitizers for the string operations that don't preserve enough of the original string value.

Sorry, I'm afraid I'm still not understanding you completely. When would it be problematic to have value-preserving flow for `String#freeze` or `String#to_s`? Conversely, when would it be useful to _not_ have that flow?

Adding special-cases in every dataflow query for these methods seems a strange approach to me.

asgerf:
> [...] I think it would be consistent to imagine that strings are like arrays but with a single field containing the string data. Then, obviously, flow through String#to_s would preserve the value of that field. I think it might even be useful to make this explicit, since having a DataFlow::Content subclass for strings would allow us to use clearsContent to model things like String#replace.

Yes, this is spot on, and it might be a sensible thing to do for Ruby's mutable strings. It doesn't come without trade-offs; there are limits to how many nested fields the data flow library can track through; and most `isSource` and `isSink` predicates would need to specify that flow starts/end in the `StringContent` field (or the field needs to be special in some way), and all the taint steps would work very different from other languages. It's not obviously the right choice, but it could be done.

Assuming this field was being used, then `to_s` could be modelled as a step `Receiver.SyntheticField[StringContent]` to `ReturnValue.SyntheticField[StringContent]`. This could be a value-preserving step if you like, it wouldn't matter really. 

Modelling `to_s` as `Receiver -> ReturnValue` as a value-preserving step would also propagate the `StringContent` field, along any other fields that the receiver may have (in case it's not a string). This last part is important, because it allows unrelated fields to continue through, such as `ArrayElement`. Since the model applies to any method call named `to_s`, there is no reason to believe the receiver can't have other fields. Even for methods with a name that only occurs on the string type, spurious flow could have caused a non-string value to reach the receiver, and the value-preserving step would allow the spurious flow to continue unrestricted.

What's the harm in propagating unrelated fields, such as `ArrayElement`? You've seen how it caused spurious flow for the `concat` method. This occurred because all the `ArrayElement[i]` fields from the first argument are allowed to propagate directly to the result (this is in addition to the steps added by the array model, which change the `i` index). A taint step from the first argument of `concat` to its result would not allow `ArrayElement[i]` to get through. For the `to_s` method, it's more complicated to see why it's a problem, so I'll get back to that in a bit.

When the same method name occurs on different library types, there are basically two approaches to handling calls with that method name:
1. Determine the type of the receiver, and only apply the summary relevant for that type.
2. Use a single set of summaries for all calls with that method name, but design them in a way that they are harmless when applied to the wrong type.

For (1) we use API graphs, but this relies on having a known set of endpoints that produce the type in question (what library models provide, essentially). For pervasively used types like strings and arrays, we have to rely on (2), as we can't realistically identify all the ways to obtain a string or an array.

For (2) to work, flow through fields is useful for designing summaries that implicitly check the receiver type. A summary of form `Receiver.ArrayElement[i] -> ReturnValue.ArrayElement[i]` is harmless if the receiver isn't an array, because non-array types don't have the `ArrayElement` field. But this is where modeling `to_s` as a value-preserving steps gets problematic, because suddenly you have strings that _do_ appear to have the `ArrayElement` field on it, and all the models you thought would only apply to arrays suddenly apply to the return value of `Array#to_s` as well.

With all that said, it helps to have some real-world examples to justify decisions like this. It's hard to construct small examples involving spurious flow without the example looking absurd.

> Adding special-cases in every dataflow query for these methods seems a strange approach to me.

As I said, I'd expect this to be a niche and most dataflow queries won't need to special-case them. If the query is tracking non-string values like numbers, the `to_s` conversion is actually an unwanted data-flow step, and then such queries would have to special-case these calls to get rid of them. AFAIK there just isn't a precedent for having lots of queries that care about "value-preserving string-only flow".

asgerf:
> As an aside, doing type inference here would massively increase the sensitivity of API graphs - I don't know if that was the idea when we were talking as a team about using RBS signatures? I'm thinking of simple stuff like identifying that Time.now.zone is a getTopLevelMember("String").getInstance() because you know about Time.now and Time#zone. Is there a worry that it would make the API graph too large, and/or are there other problems?

The (2) method is still more robust than trying to perform global type inference, even if you have type signatures for all dependencies.

Global type inference is notoriously problematic. Not only is it intractable, it's the kind of intractable problem that looks solvable at first, but where early successes don't generalize. It can lure you into thinking that, if you just tweak it for a few more years it will reach a point where it "just works", but it will never get there.

If you design everything on the assumption that good-quality type information is available, you're putting all your eggs in one very fragile basket. The more robust system is one that requires the least amount of type information up front in order to work.

There are differences between how `DataFlow::Configuration` and type-tracking perform data flow analysis, and sometimes `DataFlow::Configuration` will be able to track flow to a point where type-tracking couldn't (and vice versa, mainly because type-tracking uses some heuristics that work well for API objects, like database connection handles and such). With the (2) method, we can leverage the added precision from data-flow configurations. With the (1) method, we are relying on _both_ analyses being able to track the value from the source to the receiver of a method call -- you sort of get the worst of both worlds.

Another difference is that `DataFlow::Configuration` only tracks some specific sources; whereas trying to type-track everything with a known type is going to track a lot of stuff we might never care about.

But wait, can't you just type-track strings and arrays, and leave out other types? No, you have to track any type from which you could possibly get to a string, so if `Foo#bar -> Bar` and `Bar#baz -> str`, you have to track instances of `Foo` and `Bar`, because otherwise you won't know that `foo.bar.baz` is a string. You end up tracking virtually everything.

That doesn't mean type-tracking/API graphs can't be helpful most of the time. We use type-tracking for call graphs in JS after all, except we try to focus on tracking flows that might generate call edges, rather than tracking anything that has known type.

So what I advise for is a hybrid approach where tracking of pervasive types like strings and arrays are given a precision boost while relieving us from the burden of tracking all sources of strings/arrays. And then we rely on API graphs, possibly derived from RBS files, for modelling less pervasive types.

ruby/ql/lib/codeql/ruby/frameworks/core/String.qll
```diff
@@ -0,0 +1,595 @@
...
+
+    override MethodCall getACall() {
+      result =
+        API::getTopLevelMember("String").getAMethodCall("try_convert").getExprNode().getExpr()
+    }
+
+    override predicate propagatesFlowExt(string input, string output, boolean preservesValue) {
+      input = "Argument[0]" and
+      output = "ReturnValue" and
+      preservesValue = true
```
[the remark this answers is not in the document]
hmac: Ah yeah good point, I'll update that.

asgerf:
No other comments from my end 👍

ruby/ql/lib/codeql/ruby/frameworks/core/String.qll
```diff
@@ -151,7 +131,15 @@ module String {
     }
   }
 
-  // TODO: we already have a summary for Array#clear. Check that it applies correctly to String#clear.
+  /**
+   * This is a placeholder for `String#clear`.
+   * We can't currently write this summary because there is no `DataFlow::Content` node to refer to (unlike with `Array#clear`).
+   * We need a `DataFlow::Content` node in order to override `clearsContent`.
+   */
+  private class ClearSummary extends SimpleSummarizedCallable {
```
hvitved, line 139: We don't need to do it on this PR, but I believe some other languages have defined default barriers, and I guess this would be a good candidate for that.

hvitved (approved):

ruby/ql/lib/codeql/ruby/frameworks/core/String.qll
hmac: That seems like a good idea!

PR #7, opened 2018-08-03
C++: resolveElement

ian-semmle:
This should make it easier to experiment with ideas such as linker awareness in QL.

PR #7928, opened 2022-02-10
C++: Add table that identifies C++ structured bindings

jketema:
This introduces a new table `is_structured_binding` that identifies all variables that are declared as part of structured binding declarations. The naming follows the C++17 standard, which calls something like `auto [x, y] = ...` a _structured binding declaration_ and the `x` and `y` in that declaration _structured bindings_. Not sure how clear this is, so I'm open to other name suggestions.

The table is exposed by adding a predicate to the `Variable` class. Not sure if this is the preferred approach, or whether subclassing `Variable` would be better.


cpp/ql/lib/semmle/code/cpp/Variable.qll
```diff
@@ -169,6 +169,12 @@ class Variable extends Declaration, @variable {
     variable_instantiation(underlyingElement(this), unresolveElement(v))
   }
 
+  /**
+   * Holds if this variable is declated as part of a structured binding
+   * declaration. For example, `x` in `auto [x, y] = ...`.
+   */
+  predicate isStructuredBinding() { is_structured_binding(unresolveElement(this)) }
```
MathiasVP, line 176: I don't know much about the `unresolvelement` and `underlyingElement` (except for the fact that I know it's related to linker awareness). But all other predicates seem to use `underlyingElement(this)` instead of `unresolveElement(this)`. Is there a reason `unresolveElement(this)` is the correct thing to do here?

MathiasVP:
I think this PR also deserves a change note.

cpp/ql/lib/semmle/code/cpp/Variable.qll
jketema: I honestly just guessed based on what I saw happening in the surrounding code. Do we have any documentation on `unresolvElement` and `underlyingElement` explaining their meaning?

geoffw0: There's _some_ documentation in the predicates defined near the top of `Element.qll`, but its far from great.

MathiasVP: Unfortunately, no. And I'm not sure that we'll be able to get proper documentation on these predicates since they precede most people on the C team (as can be witnessed by this very low PR number: https://github.com/github/codeql/pull/7). The best possible source of documentation is probably the QLDoc around these predicates: https://github.com/github/codeql/blob/main/cpp/ql/lib/semmle/code/cpp/Element.qll#L44.

jketema: Turns out `Element.qll` is very explicit on when to use `underlyingElement` and `unresolveElement`: use the former with `this` and the latter with anything that's not `this`. So this should be `underlyingElement`.

jketema: Looking at the implementations that also makes sense to me, as `unresolveElement` does some special handling for incomplete classes, which is not something that is relevant here.

jketema: Changed to `underlyingElement`. Thanks a lot for pointing this out!

jketema:
> I think this PR also deserves a change note.

Added a change note.

PR #7930, opened 2022-02-10
RB: convert the ruby ApiGraphs to use IPA labels

erik-krogh:
[Evaluation looks clean](https://github.com/github/codeql-dca-main/tree/data/erik-krogh/pr-7930-9739929__Differences__code-scanning/reports).

### DIFF
--- a/ruby/ql/lib/codeql/ruby/ApiGraphs.qll
+++ b/ruby/ql/lib/codeql/ruby/ApiGraphs.qll
@@ -305,7 +305,7 @@ module API {
    * shortest length.
    */
   cached
-  module Impl {
+  private module Impl {
     cached
     newtype TApiNode =
       /** The root of the API graph. */
--- a/ruby/ql/lib/codeql/ruby/ApiGraphs.qll
+++ b/ruby/ql/lib/codeql/ruby/ApiGraphs.qll
@@ -631,13 +631,13 @@ module API {
         } or
         MkLabelBlockParameter()
 
-      /** A label for the member named `prop`. */
+      /** A label for a member, for example a constant. */
       class LabelMember extends ApiLabel {
-        string member;
+        private string member;
 
         LabelMember() { this = MkLabelMember(member) }
 
-        /** Gets the property associated with this label. */
+        /** Gets the member name associated with this label. */
         string getMember() { result = member }
 
         override string toString() { result = "getMember(\"" + member + "\")" }
@@ -650,9 +650,9 @@ module API {
         override string toString() { result = "getUnknownMember()" }
       }
 
-      /** A label for the method named `method`. */
+      /** A label for a method. */
       class LabelMethod extends ApiLabel {
-        string method;
+        private string method;
 
         LabelMethod() { this = MkLabelMethod(method) }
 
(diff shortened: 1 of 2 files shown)


ruby/ql/lib/codeql/ruby/ApiGraphs.qll
```diff
@@ -306,7 +305,7 @@ module API {
    * shortest length.
    */
   cached
-  private module Impl {
+  module Impl {
```
hvitved, line 308: Why this change?

ruby/ql/lib/codeql/ruby/ApiGraphs.qll
```diff
@@ -602,48 +601,135 @@ module API {
...
+          any(DataFlowDispatch::ParameterPosition arg).isKeyword(name)
+        } or
+        MkLabelParameter(int n) {
+          any(DataFlowDispatch::ArgumentPosition c).isPositional(n)
+          or
+          any(DataFlowDispatch::ParameterPosition c).isPositional(n)
+        } or
+        MkLabelBlockParameter()
+
+      /** A label for the member named `prop`. */
```
hvitved, line 634: `A label for a member, for example a constant.`

ruby/ql/lib/codeql/ruby/ApiGraphs.qll
```diff
@@ -602,48 +601,135 @@ module API {
...
+      class LabelMember extends ApiLabel {
+        string member;
+
+        LabelMember() { this = MkLabelMember(member) }
+
+        /** Gets the property associated with this label. */
```
hvitved, line 640: property -> member name

ruby/ql/lib/codeql/ruby/ApiGraphs.qll
```diff
@@ -602,48 +601,135 @@ module API {
...
+        MkLabelParameter(int n) {
+          any(DataFlowDispatch::ArgumentPosition c).isPositional(n)
+          or
+          any(DataFlowDispatch::ParameterPosition c).isPositional(n)
+        } or
+        MkLabelBlockParameter()
+
+      /** A label for the member named `prop`. */
+      class LabelMember extends ApiLabel {
+        string member;
```
hvitved, line 636: I prefer private fields.

hvitved (changes requested):
Thanks a lot for making this change Erik.

ruby/ql/lib/codeql/ruby/ApiGraphs.qll
erik-krogh: I needed that change in the other languages so I could access private predicates.  
I just copy-pasted the change, but it's not needed here, so I'll revert it.

ruby/ql/lib/codeql/ruby/ApiGraphs.qll
```diff
@@ -602,48 +601,135 @@ module API {
...
+    /** A label in the API-graph */
+    class ApiLabel extends TLabel {
+      /** Gets a string representation of this label. */
+      string toString() { result = "???" }
+    }
+
+    private import LabelImpl
+
+    private module LabelImpl {
+      newtype TLabel =
```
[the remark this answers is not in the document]
erik-krogh: You are suggesting I move `TLabel` into the `Impl` module?  

If we want them to be cached in the same stage we can also just make the module `cached` and add dependencies between the modules. (Basically the cached stages pattern).

hvitved (approved):

PR #7946, opened 2022-02-10
Workflows: Augment workflow to ensure failure with invalid change notes

aeisenberg:


aeisenberg:
Example of a failing workflow: https://github.com/github/codeql/runs/5148220083?check_suite_focus=true

.github/workflows/check-change-note.yml
```diff
@@ -8,6 +8,7 @@ on:
       - "*/ql/src/**/*.qll"
       - "!**/experimental/**"
       - "!ql/**"
+      - ".github/workflows/check-change-not.yml"
```
dbartol, line 11:
```suggestion
      - ".github/workflows/check-change-note.yml"
```

.github/workflows/check-change-note.yml
```diff
@@ -8,6 +8,7 @@ on:
...
```
dbartol, line 12: Should change `*/ql/src/**/*.qll` to `*/ql/**/*.qll` (and same for `.ql`). Otherwise we're not running on changes that only affect the library packs.

.github/workflows/validate-change-notes.yml
```diff
@@ -0,0 +1,31 @@
...
+
+      - name: Setup CodeQL
+        uses: ./.github/actions/fetch-codeql
+
+      - name: Fail if there are any errors with existing change notes
+        if: |
+          github.event.pull_request.draft == false
+
+        run: |
+          codeql pack release --groups cpp,csharp,java,javascript,python,ruby,-examples,-test,-experimental
```
dbartol, line 31: Does this actually fail the step if `codeql` returns non-zero? I thought you needed `set -e` to make that happen.

dbartol (commented):
A couple minor things worth looking at.

aeisenberg:
Fixes https://github.com/github/codeql-core/issues/2058

.github/workflows/check-change-note.yml
aeisenberg: This was committed accidentally. I think we should exclude tests and example packs from here. So, I will add a line like this: `*/ql/lib/**/*.qll`. Same for `*.ql`. It's not really related to this change, but I'll add it anyway.

.github/workflows/validate-change-notes.yml
```diff
@@ -0,0 +1,31 @@
...
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v2
+
+      - name: Setup CodeQL
+        uses: ./.github/actions/fetch-codeql
+
+      - name: Fail if there are any errors with existing change notes
+        if: |
+          github.event.pull_request.draft == false
```
aibaars, line 28: I think there is no need for skipping this check in draft mode. Personally, I like to be notified of simple errors as early as possible.

```suggestion
        if: |
          github.event.pull_request.draft == false
```

aibaars (commented):
Looks great to me. Thanks for adding this check.

.github/workflows/validate-change-notes.yml
aeisenberg: That's fine. I was just following the example of the check-change-note job. I have no strong feelings either way.

```suggestion
```

aibaars (approved):

PR #7968, opened 2022-02-11
JS: Taint analysis for win paths

kaeluka:
This contributes support for tracking win32 paths in tainted path queries.

The main difference is, that you can now extend query labels by extending the `Platform` class (as in `TaintedWindowsPaths.ql`) and path will then be tracked using 8 labels, instead of four.

By default, there's the four labels as before:

 1. `normalized-relative-posix-path`
 2. `normalized-absolute-posix-path`
 3. `raw-relative-posix-path`
 4. `raw-absolute-posix-path`

When extending the `Platform` class by a `win32` variant, you'll additionally get:

 5. `normalized-relative-win32-path`
 6. `normalized-absolute-win32-path`
 7. `raw-relative-win32-path`
 8. `raw-absolute-win32-path`.

I also refactored the `isAdditionalTaintStep` predicate that would have — IMHO — become too large after this change by pulling out much of the logic into an abstract class and extensions thereof.

I ran an experiment, and looked at it with @esbena looking over my shoulder. It's not affecting the results of pre-existing queries. The results it finds for the new queries lgtm (I've looked at a subset of results from various projects).

### DIFF
--- a/javascript/ql/lib/semmle/javascript/security/dataflow/TaintedPathCustomizations.qll
+++ b/javascript/ql/lib/semmle/javascript/security/dataflow/TaintedPathCustomizations.qll
@@ -371,7 +371,7 @@ module TaintedPath {
         Path::PathLabel inputLabel, DataFlow::Node inputNode, Path::PathLabel outputLabel,
         DataFlow::Node outputNode
       ) {
-        getInput() = inputNode and
+        this.getInput() = inputNode and
         this = outputNode and
         outputLabel = inputLabel.toNormalized()
       }
--- a/javascript/ql/lib/semmle/javascript/security/dataflow/TaintedPathCustomizations.qll
+++ b/javascript/ql/lib/semmle/javascript/security/dataflow/TaintedPathCustomizations.qll
@@ -89,6 +89,12 @@ module TaintedPath {
     }
 
     /**
+     * Holds if `s` is a drive path, eg.: `C:\\test.txt`.
+     */
+    bindingset[s]
+    predicate isDrivePath(string s) { s.regexpMatch("^[a-zA-Z]:\\\\*") }
+
+    /**
      * A flow label representing a file path.
      *
      * By default, there are four flow labels, representing the different combinations of
@@ -299,15 +305,34 @@ module TaintedPath {
         Path::PathLabel inputLabel, DataFlow::Node inputNode, Path::PathLabel outputLabel,
         DataFlow::Node outputNode
       ) {
-        exists(int n | inputNode = this.getArgument(n) and outputNode = this |
-          n = 0 and outputLabel = inputLabel.toNormalized()
+        outputNode = this and
+        inputNode = input and
+        (
+          inputLabel.getPlatform() = Path::platformPosix() and
+          exists(int n | inputNode = this.getArgument(n) |
+            // If the initial argument is tainted, just normalize it. It can be relative or absolute.
+            n = 0 and outputLabel = inputLabel.toNormalized()
+            or
+            n > 0 and
+            // For later arguments, the flow label depends on whether the first argument is absolute or relative.
+            // If in doubt, we assume it is absolute.
+            inputLabel.canContainDotDotSep() and
+            outputLabel.isNormalized() and
+            if Path::isRelative(this.getArgument(0).getStringValue())
+            then outputLabel.isRelative()
+            else outputLabel.isAbsolute()
+          )
           or
-          n > 0 and
-          inputLabel.canContainDotDotSep() and
-          outputLabel.isNormalized() and
-          if Path::isRelative(this.getArgument(0).getStringValue())
-          then outputLabel.isRelative()
-          else outputLabel.isAbsolute()
+          inputLabel.getPlatform() = Path::platformWin32() and
+          // on windows, when one input is absolute, the output may also be
+          // absolute, no matter the position. This is due to drive paths:
+          // path.win32.join('foo\\bar', 'D:\\fnord') => absolute ('D:\\test.txt')
+          // path.win32.join('D:\\fnord', 'foo\\bar') => absolute ('D:\\fnord\\foo\\bar')
+          if
+            inputLabel.isAbsolute() or
+            exists(int n | Path::isDrivePath(this.getArgument(n).getStringValue()))
+          then outputLabel = inputLabel.toAbsolute().toNormalized()
+          else outputLabel = inputLabel
         )
       }
     }
--- a/javascript/ql/lib/semmle/javascript/security/dataflow/TaintedPathCustomizations.qll
+++ b/javascript/ql/lib/semmle/javascript/security/dataflow/TaintedPathCustomizations.qll
@@ -35,6 +35,13 @@ module TaintedPath {
 
   module Label {
     /**
+     * A flow label representing an array of path elements that may include "..".
+     */
+    abstract class SplitPath extends DataFlow::FlowLabel {
+      SplitPath() { this = "splitPath" }
+    }
+
+    /**
      * A string indicating if a path is normalized, that is, whether internal `../` components
      * have been removed.
      */
@@ -50,9 +57,15 @@ module TaintedPath {
     }
 
     /**
-     * A flow label representing a Posix path.
+     * Holds if `s` is a relative path.
+     */
+    bindingset[s]
+    predicate isRelative(string s) { not s.charAt(0) = "/" }
+
+    /**
+     * A flow label representing a file path.
      *
-     * There are currently four flow labels, representing the different combinations of
+     * By default, there are four flow labels, representing the different combinations of
      * normalization and absoluteness.
      */
     abstract class PosixPath extends DataFlow::FlowLabel {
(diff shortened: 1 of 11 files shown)


javascript/ql/lib/semmle/javascript/security/dataflow/TaintedPathCustomizations.qll
```diff
@@ -80,499 +122,608 @@ module TaintedPath {
...
-   * An expression of form `x.matches(/\.\./)` or similar.
-   */
-  class ContainsDotDotRegExpSanitizer extends BarrierGuardNode instanceof StringOps::RegExpTest {
-    ContainsDotDotRegExpSanitizer() { super.getRegExp().getAMatchedString() = [".", "..", "../"] }
-
-    override predicate blocks(boolean outcome, Expr e, DataFlow::FlowLabel label) {
-      e = super.getStringOperand().asExpr() and
-      outcome = super.getPolarity().booleanNot() and
-      label.(Label::PosixPath).canContainDotDotSlash() // can still be bypassed by normalized absolute path
+        label.(Path::PathLabel).isAbsolute()
```
erik-krogh, line 585: You forgot to use `e` in the second part of this `blocks` predicate.  
(Code-scanning have a warning about it).

kaeluka:
@erik-krogh has asked me to split the first commit into 3:
 1. with the refactoring into a module
 2. adding the platform class that creates new paths
 3. adding the logic

javascript/ql/lib/semmle/javascript/security/dataflow/TaintedPathCustomizations.qll
```diff
@@ -109,470 +122,552 @@ module TaintedPath {
...
-      output = this
-      or
-      this = NodeJSLib::FS::moduleMember("realpathSync").getACall() and
-      input = this.getArgument(0) and
-      output = this
-      or
-      this = NodeJSLib::FS::moduleMember("realpath").getACall() and
-      input = this.getArgument(0) and
-      output = this.getCallback(1).getParameter(1)
+      DataFlow::Node getOutput() { result = output }
```
erik-krogh, line 196: I remember how `getInput` and `getOutput` was used previously.  
(In the `step` predicate).  

Now you have the `mightReturnPath` predicate, so I don't see what the `getInput()`/`getOutput()` predicates are used for.   
I can see that you use the predicates in some places where you use one transformer to define another, but in those cases you could define `getInput`/`getOutput` on those classes. 

Some of the implementations of `PathProducingNode` could also be simpler by not using the `input`/`output` fields.  
(E.g. in `RelativePathCall` you could trivially remove `input`/`output`).  

I think you should delete the fields from this class and the `getInput()` / `getOutput()` methods, and then only introduce them in the implementation-classes where they are needed.

javascript/ql/lib/semmle/javascript/security/dataflow/TaintedPathCustomizations.qll
```diff
@@ -109,470 +122,552 @@ module TaintedPath {
...
+      PathProducingNode() { this = this }
 
-    /**
-     * Gets the input path to be normalized.
-     */
-    DataFlow::Node getInput() { result = input }
+      abstract predicate mightReturnPath(
+        Label::PosixPath inputlabel, DataFlow::Node inputNode, Label::PosixPath outputlabel,
+        DataFlow::Node outputNode
+      );
```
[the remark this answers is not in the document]
kaeluka: renaming to `step` 👍

javascript/ql/lib/semmle/javascript/security/dataflow/TaintedPathCustomizations.qll
```diff
@@ -109,470 +122,552 @@ module TaintedPath {
...
 
   /**
-   * A call that normalizes a path.
+   * Provides logic for tracking taint flow through `DataFlow::CallNode`s
+   * that return paths.
    */
-  class NormalizingPathCall extends DataFlow::CallNode {
-    DataFlow::Node input;
-    DataFlow::Node output;
+  module PathNodes {
```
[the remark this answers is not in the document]
kaeluka: > But: I'm also not sure we should have a module at all. The module wasn't there before, and if we want to keep as much backwards-compatibility as possible, then we shouldn't change how to use these classes.

The module is just used inside this file to logically group those classes. It's not referenced in other files. Are you happy with making it `private module PathTransformations`?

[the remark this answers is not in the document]
kaeluka: > (If you change this initial refactoring commit, but you squash it onto the initial refactoring commit?).

You mean modifying the initial refactoring commit, right?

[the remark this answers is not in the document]
kaeluka: Good point about the near-name-clash 👍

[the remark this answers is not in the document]
erik-krogh: > The module is just used inside this file to logically group those classes. It's not referenced in other files. Are you happy with making it private module PathTransformations? 

I'm thinking that previous you could import e.g. `NormalizingPathCall` using `TaintedPathCustomizations::TaintedPath::NormalizingPathCall`, and now you need an extra module in that path.  
A solution could be to add a `import PathNodes`, that way you still get the nice grouping of the code, without affecting import-paths.  

> You mean modifying the initial refactoring commit, right? 

Yes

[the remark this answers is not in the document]
kaeluka: > A solution could be to add a import PathNodes, that way you still get the nice grouping of the code, without affecting import-paths.

Awesome!

javascript/ql/lib/semmle/javascript/security/dataflow/TaintedPathCustomizations.qll
kaeluka: This was good advice, code is cleaner now and still does the same thing.

kaeluka: 👍

javascript/ql/lib/semmle/javascript/security/dataflow/TaintedPathCustomizations.qll
```diff
@@ -109,470 +122,554 @@ module TaintedPath {
...
+    class JoinPathCall extends PathTransformationNode {
+      JoinPathCall() {
+        this = NodeJSLib::Path::moduleMember("join").getACall() and
+        input = this.getAnArgument() and
+        output = this
+      }
 
-    PreservingPathCall() {
-      this =
-        NodeJSLib::Path::moduleMember(["dirname", "toNamespacedPath", "parse", "format"]).getACall() and
```
erik-krogh, line 216: I'm confused by what you've done with this class. 

It seems to me that you've changed the behavior of the first case (pasted below) in this "refactoring" commit.  
```CodeQL
NodeJSLib::Path::moduleMember(["dirname", "toNamespacedPath", "parse", "format"]).getACall() and
input = this.getAnArgument() and
output = this
```

This class was used in the `isPosixPathStep` predicate to just forward a step while preserving the flow-labels.  

It seems to me that you've changed it. 

The `"dirname"` and `"format"` case is in the `PreservingPathCall` class, with the same behavior. 
The `"toNamespacedPath"` case is in `ToNamespacedPathCall`, with the same behavior.  

But the `"parse"` case is in a new class `ParseCall` with different behavior.   
You've added some property checks based on `["root", "dir"]` 

I don't think that check works, but that's another matter. (`CallNode::getALocalUse()` can never be a `PropRead`).

kaeluka: Sorry this reply took so long, I got really bogged down with other work and then time off.

What I tried to do here was to track through `parse` more precisely (it gives you an object which contains the `root`, `dir`, `extension`, `baseName` of a path). In principle, we may add more precise tracking of those individual properties — but that's neither a refactoring, nor what this PR really needs to be adding.

As the PR big enough on its own, **I've amend the refactoring PR to remove the class `ParseCall` altogether** (https://github.com/github/codeql/compare/7193fd691a391fb2f27db708b7877c648f797589..8b817fe555a0fdee7caed60ba44be1fc11ac39b7) and moved `parse` back to its friends `dirname` and `format`.

Thank you 👍

javascript/ql/lib/semmle/javascript/security/dataflow/TaintedPathCustomizations.qll
```diff
@@ -80,499 +132,596 @@ module TaintedPath {
...
-      this = NodeJSLib::FS::moduleMember("realpathSync").getACall() and
-      input = this.getArgument(0) and
-      output = this
-      or
-      this = NodeJSLib::FS::moduleMember("realpath").getACall() and
-      input = this.getArgument(0) and
-      output = this.getCallback(1).getParameter(1)
+  module PathTransformations {
+    /** An abstract description of what kind of paths a call may return. */
+    abstract class PathTransformationNode extends DataFlow::CallNode {
```
erik-krogh, line 186:
```suggestion
    abstract class PathTransformationCall extends DataFlow::CallNode {
```
?

javascript/ql/lib/semmle/javascript/security/dataflow/TaintedPathCustomizations.qll
```diff
@@ -850,97 +1001,32 @@ module TaintedPath {
...
-   * standard taint step `src -> dst` should be suppresesd.
+   * standard taint step `src -> dst` should be suppressed.
    */
-  private predicate isPosixPathStep(
-    DataFlow::Node src, DataFlow::Node dst, Label::PosixPath srclabel, Label::PosixPath dstlabel
+  predicate isPathStep(
+    DataFlow::Node src, DataFlow::Node dst, Path::PathLabel srclabel, Path::PathLabel dstlabel
+  ) {
+    exists(PathTransformationNode ptn | ptn.step(srclabel, src, dstlabel, dst))
+  }
```
erik-krogh, line 1010: I think you should just inline this predicate into the place it's used.  
With your refactoring the predicate has become a predicate for taint-steps that originate from some call, which is not reflected in the name/qldoc.

javascript/ql/lib/semmle/javascript/security/dataflow/TaintedPathCustomizations.qll
```diff
@@ -50,23 +84,41 @@ module TaintedPath {
...
+      forall(Platform p | not s.substring(0, p.getSep().length()) = p.getSep())
+    }
+
+    /**
+     * Holds if `s` is a drive path, eg.: `C:\\test.txt`.
+     */
+    bindingset[s]
+    predicate isDrivePath(string s) { s.regexpMatch("^[a-zA-Z]:\\\\*") }
+
+    /**
```
erik-krogh, line 100:
```suggestion
    /** DEPRECATED: Use `PathLabel`. */
    abstract deprecated class PosixLabel extends PathLabel {
      PosixLabel() { platform = platformPosix() }
    }

    /**
```

javascript/ql/lib/semmle/javascript/security/dataflow/TaintedPathCustomizations.qll
```diff
@@ -33,7 +36,38 @@ module TaintedPath {
    */
   abstract class BarrierGuardNode extends DataFlow::LabeledBarrierGuardNode { }
 
-  module Label {
+  /**
```
erik-krogh, line 39:
```suggestion
  /** DEPRECATED: Use `Path`. */
  deprecated module Label = Path;

  /**
```

javascript/ql/lib/semmle/javascript/security/dataflow/ZipSlipQuery.qll
```diff
@@ -11,11 +11,11 @@ import javascript
 import ZipSlipCustomizations::ZipSlip
 
 // Materialize flow labels
-private class ConcretePosixPath extends TaintedPath::Label::PosixPath {
+private class ConcretePosixPath extends TaintedPath::Path::PathLabel {
```
erik-krogh, line 14: When I saw this I thought you were materializing the labels for all platforms here.  
You are, but the only platform that is activated by this file is the Posix platform.  

I think you could add in the comment of these "label-materializations" that only the Posix platform exists for this query, and add in the QLDoc for `abstract class Platform extends string` that only the Posix platform exists by default.

javascript/ql/lib/semmle/javascript/security/dataflow/TaintedPathCustomizations.qll
```diff
@@ -33,7 +36,38 @@ module TaintedPath {
...
+
+    /** Gets the string used for tracking Posix paths. */
+    string platformPosix() { result = "posix" }
+
+    /** Gets the string used for tracking Windows paths. */
+    string platformWin32() { result = "win32" }
+
+    /** A platform on which we track paths, eg.: `posix`. */
+    abstract class Platform extends string {
+      Platform() { this = [platformPosix(), platformWin32()] }
```
erik-krogh, line 58:
```suggestion
      bindingset[this]
      Platform() { any() }
```

javascript/ql/lib/semmle/javascript/security/dataflow/TaintedPathCustomizations.qll
```diff
@@ -80,499 +132,596 @@ module TaintedPath {
...
+          inputLabel.getPlatform() = Path::platformPosix() and
+          outputLabel = inputLabel.toRelative().toNormalized()
+          or
+          // relative, on win32, does not always return a normalized path, eg.,
+          // when both arguments are equal, relative will return ''
+          // (normalized would be '.').
+          // It also does not always return a relative path, eg.,
+          // relative('C:\\', 'A:\\') returns 'A:\\'.
+          inputLabel.getPlatform() = Path::platformWin32() and
+          outputLabel = inputLabel
```
erik-krogh, line 261: I think you could move all of these windows specific paths into a windows specific file (`TaintedWindowsPathQuery.qll`?).    
In all cases you can do something like this (the name is bad):   
```CodeQL
class RelativePathCallWindows extends RelativePathCall {
  override predicate step(...) {
    super.step(...)
    or
    this.getAnArgument() = inputNode and 
    this = outputNode and
    outputLabel.getPlatform() = inputLabel.getPlatform() and
    inputLabel.getPlatform() = Path::platformWin32() and
    outputLabel = inputLabel
  }
}
```

If too much logic ends up getting duplicated, then you can move some if it into a utility predicate on the class.

javascript/ql/lib/semmle/javascript/security/dataflow/TaintedPathCustomizations.qll
```diff
@@ -80,499 +132,596 @@ module TaintedPath {
       predicate isAbsolute() { relativeness = "absolute" }
 
       /** Gets the path label with normalized flag set to true. */
-      PosixPath toNormalized() {
+      PathLabel toNormalized() {
```
erik-krogh, line 135: I think all of these utility predicates should preserve the platform 
(add `this.getPlatform() = result.getPlatform()` to `toNormalized`, `toNonNormalized`, `toAbsolute`, toRelative`).

erik-krogh: I think your `NormalizingPathCall` doesn't preserve the platform due to you not preserving the platform in these utility predicates.  
(The rest look OK).

javascript/ql/src/Security/CWE-022/TaintedWindowsPath.ql
```diff
@@ -0,0 +1,33 @@
...
+ *       external/cwe/cwe-036
+ *       external/cwe/cwe-073
+ *       external/cwe/cwe-099
+ */
+
+import javascript
+import semmle.javascript.security.dataflow.TaintedPathQuery
+import semmle.javascript.security.dataflow.TaintedPathCustomizations
+import DataFlow::PathGraph
+import semmle.javascript.frameworks.NodeJSLib
```
erik-krogh, line 22:
```suggestion
import javascript
import semmle.javascript.security.dataflow.TaintedPathQuery
import DataFlow::PathGraph
```

Note to self: add this as an example to the `ql/redundant-import` ticket.

javascript/ql/src/Security/CWE-022/TaintedWindowsPath.ql
```diff
@@ -0,0 +1,33 @@
...
+
+class Win32Platform extends Path::Platform {
```
erik-krogh, line 24:
```suggestion
/** Materializes the Win32 platform, which enables the Win32 tainted-path steps */
class Win32Platform extends Path::Platform {
```

javascript/ql/test/query-tests/Security/CWE-022/TaintedWindowsPath/TaintedWindowsPath.qlref
```diff
@@ -0,0 +1 @@
+Security/CWE-022/TaintedWindowsPath.ql
```
erik-krogh, line 1: I would like a test of which taint-flows are only an issue on Windows.  

I've written it for you here:  
```CodeQL
import javascript
import semmle.javascript.security.dataflow.TaintedPathQuery

// TODO: Move this to TaintedWindowsPathQuery.qll?
class Win32Platform extends Path::Platform {
  Win32Platform() { this = Path::platformWin32() }

  override string getSep() { result = "\\" }
}

from Configuration cfg, DataFlow::PathNode source, DataFlow::PathNode sink
where
  cfg.hasFlowPath(source, sink) and
  sink.getFlowLabel().(Path::PathLabel).getPlatform() = Path::platformWin32() and
  not sink.getFlowLabel().(Path::PathLabel).getPlatform() = Path::platformPosix()
select source, sink
```

It detects that `express.js:27` is only an issue on windows (as expected).  
But it doesn't flag `express.js:35`. I haven't looked into why.

javascript/ql/test/query-tests/Security/CWE-022/TaintedPath/express.js
```diff
@@ -4,6 +4,16 @@ var express = require("express"),
...
+
+app.get("/some/path/2", function (req, res) {
+  const p = path.join('.', req.query.filename);
+  if (!p.startsWith('..')) {
+    req.files.foo.mv(p); // OK (but not on windows)
+    res.status(200);
+  } else {
+    res.status(400);
+  }
+});
```
erik-krogh, line 19: Could you move this test to the Windows-path test folder?  
With the test I suggested we can actually test that `req.files.foo.mv(p)` is only flagged by the windows query.

kaeluka:
Talked this over with Calum; I'll drop this for the time being.

PR #7969, opened 2022-02-11
Docs: remove mention of 'filter queries'

aibaars:
Filter queries are a historic feature that is no longer supported.


jf205 (approved):
Thanks @aibaars. This change will make it to the live site when the docs team do the next udpate.

PR #7980, opened 2022-02-11
QL: fix pack name for ql-for-ql

erik-krogh:


tausbn (approved):

nickrolfe:
How about this while you're there?

```diff
diff --git a/.github/workflows/ql-for-ql-build.yml b/.github/workflows/ql-for-ql-build.yml
index 16a675427d..e3570d38ae 100644
--- a/.github/workflows/ql-for-ql-build.yml
+++ b/.github/workflows/ql-for-ql-build.yml
@@ -3,8 +3,14 @@ name: Run QL for QL
 on:
   push:
     branches: [main]
+    paths:
+      - "ql/**"
+      - .github/workflows/ql-for-ql-build.yml
   pull_request:
     branches: [main]
+    paths:
+      - "ql/**"
+      - .github/workflows/ql-for-ql-build.yml
 
 env:
   CARGO_TERM_COLOR: always
```

erik-krogh:
> How about this while you're there?

As it is the workflow should always run, and that is how it should be (it does the analysis for java, python, js, etc.).  
I don't know when it broke, but I suspect it has something to do with the recent renaming the qlpack.

erik-krogh:
This PR broke things: https://github.com/github/codeql/pull/7913 
But the build queries were just fetched from the cache, no new build was attempted (that should have been the case). 

The cache got evicted when Taus merged a PR recently, and then the build broke.  
The cache should have been evicted by the PR that broke things. 
Basically a race condition with two PRs being merged.

PR #7982, opened 2022-02-11
C++: Remove some unused legacy relations from the DB scheme

jketema:

* `externalData`
* `snapshotDate`
* `duplicateCode`
* `similarCode`
* `tokens`

This is similar to the following C# change https://github.com/github/codeql/pull/7746


MathiasVP:
There are a couple of compile errors in `CodeDuplication.qll` now. For example:
```
ERROR: Could not resolve type @duplication_or_similarity (/home/runner/work/semmle-code/semmle-code/ql/cpp/ql/src/external/CodeDuplication.qll:14,20-46)
```

I think we should redefine the `Copy` class as:
```ql

private newtype TCopy = MkCopy() // Or whatever

/**
 * DEPRECATED: This class is no longer used.
 * 
 * A token block used for detection of duplicate and similar code.
 */
class Copy extends TCopy { ... }
```
and replace body of the predicates that refer to non-existing tables with `none()`. Hopefully, that means that existing code that uses these classes will still compile.

How does that sound?

jketema:
Sound like a plan.

cpp/ql/lib/semmlecode.cpp.dbscheme.stats
```diff
@@ -2634,2445 +2395,6 @@
 <dependencies/>
 </relation>
 <relation>
```
MathiasVP, line 2397: I think you want to delete this line, but keep the last `<relation>` on line 5057? Currently it doesn't look like well-bracketed XML to me.

MathiasVP:
Almost LGTM! The QLDoc Check is complaining because we don't have QLDoc on:
- The `external/CodeDuplication.qll` file, and
- the `CodeDuplication::SimilarBlock` class.

I think we can ignore the file QLDoc, but maybe we should add a QLDoc on the `SimilarBlock` just saying `DEPRECATED: This class is no longer used.` or something similar.

And finally, we need to include a change note in https://github.com/github/codeql/tree/main/cpp/ql/src/change-notes.

jketema:
I accidentally added an empty line between `CodeDuplication::SimilarBlock` and its documentation. That should be fixed now.

MathiasVP:
> I accidentally added an empty line between `CodeDuplication::SimilarBlock` and its documentation. That should be fixed now.

Oh, perfect.

jketema:
The change note should be about the deprecation of the classes?

MathiasVP:
> The change note should be about the deprecation of the classes?

Yeah, exactly.

MathiasVP (dismissed):
LGTM!

MathiasVP (dismissed):

MathiasVP (approved):

jketema:
Discussion on code duplication queries: https://github.com/github/codeql-c-team/issues/560

PR #7984, opened 2022-02-11
JS: fix most ql-for-ql warnings

erik-krogh:
[Fixes 40 QL-for-QL warnings](https://github.com/github/codeql/pull/7984/checks?check_run_id=5159495402). 
Mostly benign qldoc changes.  

But there was a real bug in [UnsafeCodeConstructionCustomizations.qll](https://github.com/github/codeql/pull/7984/files#diff-aa9eaa3a4406dd3324fe81e65c1bb8ab5ef2229f14a368730aa3929075559e40) :scream:  
(The evaluation is on that query).  

[Evaluation looks fine](https://github.com/github/codeql-dca-main/tree/data/erik-krogh/pr-7984-a1c5724__default__CustomSuite/reports).  
(There are some new results in some minified files that are hard to triage).


annarailton (approved):
Happy with the doc changes to the one CodeML file `javascript/ql/experimental/adaptivethreatmodeling/lib/experimental/adaptivethreatmodeling/BaseScoring.qll`

javascript/ql/lib/semmle/javascript/dataflow/TaintTracking.qll
```diff
@@ -635,39 +635,18 @@ module TaintTracking {
           pred.asExpr() = succ.getAstNode().(MethodCallExpr).getReceiver() and
           (
             // sorted, interesting, properties of String.prototype
-            name = "anchor" or
```
kaeluka, line 638: for really large sets, readability suffers — like in this case. I like the consistency, though. But maybe the real issue here is the auto-formatter that won't allow having one line per item for large sets?

kaeluka (approved):
This looks good to me. Glad to see that documentation is coming out looking very consistent 👍 
I added one observation — but I think that's mostly about the analysis, rather than this specific PR.

javascript/ql/lib/semmle/javascript/dataflow/TaintTracking.qll
asgerf: +💯

One workaround we commonly is to put an empty line comment at the end of each item

PR #7985, opened 2022-02-11
Ruby: separate constant propagation of regexps from strings

nickrolfe:
This lets us distinguish constant regexp values from regular strings, which will be useful for a query I'm working on. It's similar to the way that symbols are already distinguished from strings.

As part of the change, I've deprecated the names `getStringOrSymbol` and `isStringOrSymbol`, and replaced them with `getStringlikeValue` and `isStringlikeValue`.

N.B. I checked that there were no changes to `ValueText.expected` besides the new column.


nickrolfe:
This needs some work, for example to track the regexp flag string alongside its value. I'll update the PR once I've validated the approach by finishing the query I'm working on.

ruby/ql/test/library-tests/ast/escape_sequences/EscapeSequences.ql
```diff
@@ -21,4 +21,6 @@ query predicate stringlikeLiterals(StringlikeLiteral l, string value, string kin
   value = l.getConstantValue().getString() and kind = "string"
```
hvitved, line 21: Can now be simplified to
```ql
exists(ConstantValue v | v = l.getValueType() and value = v.getString() and kind = v.getValueType())
```

hvitved (changes requested):
Looks great! We should run a DCA job before merging.

ruby/ql/test/library-tests/ast/escape_sequences/EscapeSequences.ql
nickrolfe: You mean
```ql
exists(ConstantValue v |
  v = l.getConstantValue() and value = v.getStringlikeValue() and kind = v.getValueType()
)
```
But yes, good point.

hvitved: Yes, indeed.

nickrolfe:
I merged in `main` to resolve the conflict with renamed `ParseRegExp.qll` (git was able to merge it cleanly).

nickrolfe:
DCA performance looks fine to me.

PR #8012, opened 2022-02-14
QL: allow raw db types in upgrade/downgrade scripts

erik-krogh:
Removes FPs like [this one](https://github.com/github/codeql/security/code-scanning/7316?query=ref%3Arefs%2Fheads%2Fmain).


MathiasVP (approved):
LGTM!

PR #7757, opened 2022-01-26
C#: Remove stats for removed relations

hvitved:
I forgot to remove a few entries on https://github.com/github/codeql/pull/7746.

PR #8017, opened 2022-02-14
C#: add externalData back to dbscheme

nickrolfe:
This is a partial revert of https://github.com/github/codeql/pull/7746 and https://github.com/github/codeql/pull/7757.

It turns out that the `externalData` relation is still used (e.g. by the field team), but it's populated by the CSV extractor, not the C# extractor.


csharp/downgrades/a696c8bae067f69ab3208e98ce35f4fdf7efb68b/upgrade.properties
```diff
@@ -0,0 +1,2 @@
+description: Add externalData relation
+compatibility: full
```
aibaars, line 2: should we add a `delete:` line, or is it better to just leave the table if it exists?

nickrolfe: That was an oversight on my part. I'll add a `delete` line.

tamasvajk (approved):
LGTM

csharp/ql/lib/semmlecode.csharp.dbscheme.stats
```diff
@@ -12,6 +12,10 @@
             <k>@extractor_message</k>
             <v>41551</v>
         </e>
+        <e>
+            <k>@externalDataElement</k>
```
MathiasVP, line 16: As far as I know, `codeql test` will throw an exception if you have a `typesizes` with a `0` entry. I _think_ the motivation for the now-reverted PR is that the stats generated by DCA don't include stats from tests, so we cannot provide a non-zero `typesizes` entry for `externalDataElement`.

What we're doing in C++ to address this issue is to add an entry in the `static-stats/cpp` folder in the DCA repo. I imagine C# could do the same if it's needed.

nickrolfe: There are several other, existing `typesizes` entries with a `0` value. I'm a bit confused about why they don't cause a problem.

MathiasVP: Good point. I guess this shouldn't block the revert. But I'd definitely be interested in knowing why `codeql test` doesn't complain too 🤔.

PR #8022, opened 2022-02-14
JS: Add url-parse.qs as an alias for the querystringify library

asgerf:
The `querystringify` package is re-exported as `qs` from `url-parse`.


esbena (approved):

asgerf:
[Evaluation](https://github.com/github/codeql-dca-main/tree/data/asgerf/url-parse-qs__smoke-test__smoke-test/reports) looks fine

PR #8028, opened 2022-02-14
Python: Fix bad `fastTC` in `ASTNode::contains`

tausbn:
Between 2.8.0 and 2.8.1, the performance of `ASTNode::contains` degraded considerably for some queries (and some databases), one such example being the `py/comparison-of-identical-expressions` query.

In 2.8.0, the `contains` relation was inlined into the helper predicates for that query, whereas in 2.8.1, the `contains` relation was evaluated in isolation using a `boundedFastTC`:
```
(4s) Starting to evaluate predicate boundedFastTC(AstExtended::AstNode::getAChildNode_dispred#ff,AstExtended::AstNode#class#f)/2@c4bc3flu = HOP boundedFastTC(3646779,2616172)
(887s)  >>> Relation boundedFastTC(AstExtended::AstNode::getAChildNode_dispred#ff,AstExtended::AstNode#class#f): 12307755 rows using 115 MB
```

I tried rewriting the `contains` predicate to use a manual TC, but this also had poor performance (though still better than `fastTC`). Because `contains` is used in a lot of places, it's inconvenient to manually fix up and magic all of these places, and so I opted to `inline` this predicate instead.

### DIFF
--- a/python/ql/lib/semmle/python/AstExtended.qll
+++ b/python/ql/lib/semmle/python/AstExtended.qll
@@ -47,6 +47,7 @@ abstract class AstNode extends AstNode_ {
   AstNode getParentNode() { result.getAChildNode() = this }
 
   /** Whether this contains `inner` syntactically */
+  pragma[inline]
   predicate contains(AstNode inner) { this.getAChildNode+() = inner }
 
   pragma[noinline]
--- a/python/ql/lib/semmle/python/Patterns.qll
+++ b/python/ql/lib/semmle/python/Patterns.qll
@@ -10,7 +10,8 @@ class Pattern extends Pattern_, AstNode {
   override Scope getScope() { result = this.getCase().getScope() }
 
   /** Gets the case statement containing this pattern */
-  Case getCase() { result = this.getParent() or result = this.getParent().(Pattern).getCase() }
+  pragma[nomagic]
+  Case getCase() { result.contains(this) }
 
   override string toString() { result = "Pattern" }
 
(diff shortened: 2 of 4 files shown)


yoff:
Performance test not looking too promising..?

yoff:
> I have a further fix for one of the queries that did this. I don't know if it's sufficient, but I'll push it and rerun the performance tests.

Ok, that fix looks sensible, let's see what effect is has.. :-)

tausbn:
Hm. Seems my `getCase` fix broke the tests. I'll see if I can rewrite it to work correctly again.

tausbn:
I think the latest evaluation looks good. There are a few individual queries that got slower overall, but the slowdown there is on the order of 10 seconds _in total_ (whereas the overall speedup is between 5% and 9%).

tausbn:
I am tempted, however, to just do a PR where I add the `nomagic` to `getCase`. I wonder if that's the bit that made the most difference here. What do you think?

yoff:
> I am tempted, however, to just do a PR where I add the `nomagic` to `getCase`. I wonder if that's the bit that made the most difference here. What do you think?

Yes, that could well be a simple fix, avoiding splitting out predicates.

RasmusWL:
Was https://github.com/github/codeql/pull/8255 a sufficient fix for this performance problem?

tausbn:
> Was #8255 a sufficient fix for this performance problem?

I think so. I'll close this for now. If we start seeing degradation in the calculation of `contains`, we can always revive this PR.

PR #8030, opened 2022-02-14
C++: Cache `getLocation` on `Instruction` and `Operand`

MathiasVP:
Currently in draft while I test on DCA.


dbartol (approved):
LGTM assuming the DCA run shows the expected improvement

MathiasVP:
Hm. DCA doesn't show the expected improvements. I'll keep this one in draft for now and investigate why that is (at some point later point).

PR #8032, opened 2022-02-14
Java: Add Guard Classes for checking OS & unify System Property Access

JLLeitschuh:
This is useful when you have a bit of logic that is only vulnerable on specific OSes. Additionally, it unifies the access of all system properties (from all libraries and JDK APIs) into one method call `getSystemProperty(_)`.


java/ql/test/query-tests/security/CWE-200/semmle/tests/TestSafe.java
```diff
@@ -0,0 +1,88 @@
...
+                    ); // GOOD: Directory has permissions `-rw-------`
+                } else {
+                    Files.setPosixFilePermissions(
+                            tempDirChild,
+                            posixFilePermissions
+                    ); // GOOD: Good has permissions `-rw-------`, or will throw an exception if this fails
+                }
+            } else if (!Files.exists(tempDirChild)) {
+                // On Windows, we still need to create the directory, when it doesn't already exist.
+                Files.createDirectory(tempDirChild); // GOOD: Windows doesn't share the temp directory between users
```
JLLeitschuh, line 45: Without the OS guard changes, this gets flagged as vulnerable

java/ql/test/query-tests/security/CWE-200/semmle/tests/TestSafe.java
```diff
@@ -0,0 +1,88 @@
...
+                    ); // GOOD: Directory has permissions `drwx------`
+                } else {
+                    Files.setPosixFilePermissions(
+                            tempDirChild,
+                            posixFilePermissions
+                    ); // GOOD: Good has permissions `drwx------`, or will throw an exception if this fails
+                }
+            } else if (!Files.exists(tempDirChild)) {
+                // On Windows, we still need to create the directory, when it doesn't already exist.
+                Files.createDirectories(tempDirChild); // GOOD: Windows doesn't share the temp directory between users
```
JLLeitschuh, line 82: Without the OS guard changes, this also gets flagged as vulnerable

java/ql/src/Security/CWE/CWE-200/TempDirUsageSafe.java
```diff
@@ -30,7 +34,7 @@ void exampleSafeWithWindowsSupportFile() {
         createTempFile(tempChildFile.toPath()); // GOOD: Good has permissions `-rw-------`
     }
 
-    static void createTempFile(Path tempDir) {
+    static void createTempFile(Path tempDirChild) {
```
JLLeitschuh, line 37: Simple fix because this wouldn't have compiled earlier

java/ql/lib/semmle/code/java/frameworks/apache/Lang.qll
```diff
@@ -37,3 +37,12 @@ private class ApacheStrBuilderFluentMethod extends FluentMethod {
     this.getReturnType().(RefType).hasQualifiedName("org.apache.commons.lang3.text", "StrBuilder")
   }
 }
+
+/**
+ * The class ``org.apache.commons.lang.SystemUtils` or `org.apache.commons.lang3.SystemUtils`.
```
intrigus-lgtm, line 42: Double backticks.

java/ql/lib/semmle/code/java/frameworks/apache/Lang.qll
```diff
@@ -37,3 +37,12 @@ private class ApacheStrBuilderFluentMethod extends FluentMethod {
...
+ */
+class ApacheSystemUtilis extends Class {
```
intrigus-lgtm, line 44: Typo: Utilis

java/ql/lib/semmle/code/java/os/OSCheck.qll
```diff
@@ -0,0 +1,93 @@
...
+ */
+
+import java
+import semmle.code.java.controlflow.Guards
+private import semmle.code.java.frameworks.apache.Lang
+private import semmle.code.java.dataflow.DataFlow
+private import semmle.code.java.StringCheck
+
+/**
+ * A complimentatry guard that checks if the current platform is Windows.
```
intrigus-lgtm, line 12: complimentatry

java/ql/lib/semmle/code/java/os/OSCheck.qll
```diff
@@ -0,0 +1,93 @@
...
+ */
+abstract class IsWindowsGuard extends Guard { }
+
+/**
+ * A complimentatry guard that checks if the current platform is unix or unix-like.
```
intrigus-lgtm, line 17: Dito

JLLeitschuh:
Thanks @intrigus-lgtm for the feedback! 🙂

java/ql/lib/semmle/code/java/os/OSCheck.qll
```diff
@@ -0,0 +1,93 @@
...
+ */
+abstract class IsUnixGuard extends Guard { }
+
+/**
+ * Holds when the MethodAccess is a call to check the current OS using either the upper case `osUpperCase` or lower case `osLowerCase` string constants.
+ */
+bindingset[osUpperCase, osLowerCase]
+private predicate isOsFromSystemProp(MethodAccess ma, string osUpperCase, string osLowerCase) {
```
smowton, line 25: Only one of `osUpperCase` or `osLowerCase` is bound in disjuncts 1 and 2 respectively. Thanks to the bindingset annotation this doesn't try to compute the product of the bound parameter and all possible strings, but it does compute the product with whatever strings are specified at the various invocation sites.

In general you should make sure a parameter is bound in all disjuncts.

java/ql/lib/semmle/code/java/os/OSCheck.qll
```diff
@@ -0,0 +1,93 @@
...
+  exists(MethodAccessSystemGetProperty sgpMa |
+    sgpMa.hasCompileTimeConstantGetPropertyName("os.name")
+  |
+    DataFlow::localExprFlow(sgpMa, ma.getQualifier()) and // Call from System.getProperty to some partial match method
```
smowton, line 29: Why the asymmetry between how isUpper and isLowerCase are used?

java/ql/lib/semmle/code/java/os/OSCheck.qll
```diff
@@ -0,0 +1,93 @@
...
+    ma.getAnArgument().(CompileTimeConstantExpr).getStringValue().matches(osUpperCase)
+    or
+    exists(MethodAccess toLowerCaseMa |
+      toLowerCaseMa.getMethod() =
+        any(Method m | m.getDeclaringType() instanceof TypeString and m.hasName("toLowerCase"))
+    |
+      DataFlow::localExprFlow(sgpMa, toLowerCaseMa.getQualifier()) and // Call from System.getProperty to toLowerCase
```
smowton, line 36: They could also check the OS string without forcing the case, right? How about simply following through any toUpper or toLower calls we find without regard to which one it is, and make a case-insensitive check ourselves in QL (e.g. `comparedAgainstConstant.toLowerCase().matches("linux%")`)

java/ql/lib/semmle/code/java/os/OSCheck.qll
```diff
@@ -0,0 +1,93 @@
...
+
+private predicate isOsFromApacheCommons(FieldAccess fa, string fieldName) {
+  exists(Field f | f = fa.getField() |
+    f.getDeclaringType() instanceof ApacheSystemUtilis and
+    f.hasName(fieldName)
+  )
+}
+
+private class IsWindowsFromApacheCommons extends IsWindowsGuard instanceof FieldAccess {
+  IsWindowsFromApacheCommons() { isOsFromApacheCommons(this, "IS_OS_WINDOWS%") }
```
smowton, line 62: Looks like matching syntax, but `matches` not used. Check if this is tested?

java/ql/lib/semmle/code/java/os/OSCheck.qll
JLLeitschuh: I'm looking at the various ways that people describe to check the OS here:
https://stackoverflow.com/questions/228477/how-do-i-programmatically-determine-operating-system-in-java

java/ql/lib/semmle/code/java/os/OSCheck.qll
JLLeitschuh: Sorry, the above is going a bit over my head, can you try the ELI5 version? Or something maybe a bit higher than that, but still simpler 😆

smowton: Take for example this predicate:

```
bindingset[x,y]
predicate xOrYIsOne(int x, int y) {
  x = 1 or y = 1
}
```

This is made of two disjuncts, each of which only binds one of x or y.

Since these are `int`s, the only way this will compile at all is with a bindingset annotation. Let's say the use site constrains them to be either 1 or 2.

Then this will be computed by evaluating `x = 1`, then `y = 1`, then adding the results together.

That means we'll get `(1, 1), (1, 2)` + `(1, 1), (2, 1)`

If the surrounding environment constrained them to be between 1 and 1000 then we'd get (1, 1), (1, 2) ... (1, 1000) and similarly the other way around.

If they were naturally finite types, like database types, then we could remove the bindingset annotation and we'd get  a hit for every possible value of the unbound parameter.

In the example given here, the use sites limit us to a small range of possible parameter values so the result duplication isn't too bad, but it's bad practice in general to have a disjunct that leaves a parameter unspecified like this.

Instead you could do:

```
bindingset[osUpperCase, osLowerCase]
private predicate isOsFromSystemProp(MethodAccess ma, string osString) {
  exists(MethodAccessSystemGetProperty sgpMa, Expr sgpFlowsToExpr |
    sgpMa.hasCompileTimeConstantGetPropertyName("os.name")
  |
    DataFlow::localExprFlow(sgpMa, sgpFlowsToExpr) and // Call from System.getProperty to some partial match method
    ma.getAnArgument().(CompileTimeConstantExpr).getStringValue().toLowerCase().matches(osString) and
    (
      sgpFlowsToExpr = ma.getQualifier() 
      or
      exists(MethodAccess caseChangeMa |
        caseChangeMa.getMethod() =
          any(Method m | m.getDeclaringType() instanceof TypeString and m.hasName(["toLowerCase", "toUpperCase"]))
        |
        sgpFlowsToExpr = caseChangeMa.getQualifier() and // Call from System.getProperty to case-switching method
        DataFlow::localExprFlow(caseChangeMa, ma.getQualifier()) and // Call from case-switching method to some partial match method
    )
  ) and
  isStringPartialMatch(ma)
}
```

smowton: That also regularises the use of the case-switching methods. `osString` should be lower-case in this implementation.

java/ql/lib/semmle/code/java/os/OSCheck.qll
smowton: I think it's safe to assume that if people are using `get("os.name").toLowerCase().startsWith(...)` then someone out there will be using `toUpperCase` too (but see comment above for a proposed fix)

java/ql/lib/semmle/code/java/os/OSCheck.qll
JLLeitschuh: It wasn't tested. I was lazy. Let me fix that

java/ql/lib/semmle/code/java/StringCheck.qll
```diff
@@ -0,0 +1,14 @@
...
+
+import java
+
+/**
+ * Holds if `ma` is a call to a method that checks a partial string match.
+ */
+predicate isStringPartialMatch(MethodAccess ma) {
+    ma.getMethod().getDeclaringType() instanceof TypeString and
+    ma.getMethod().getName() =
+      ["contains", "startsWith", "matches", "regionMatches", "indexOf", "lastIndexOf"]
```
Marcono1234, line 13: Should this also consider `endsWith`? Otherwise the predicate name and documentation might be a bit misleading.

java/ql/lib/semmle/code/java/os/OSCheck.qll
```diff
@@ -0,0 +1,93 @@
...
+}
+
+private class IsUnixFromApacheCommons extends IsUnixGuard instanceof FieldAccess {
+  IsUnixFromApacheCommons() { isOsFromApacheCommons(this, "IS_OS_UNIX") }
+}
+
+/**
+ * A guard that checks if the `java.nio.file.FileSystem` supports posix file permissions.
+ * This is often used to infer if the OS is unix-based.
+ * Looks for calls to `contains("poxix")` on the `supportedFileAttributeViews` method returned by `FileSystem`.
```
Marcono1234, line 72:
```suggestion
 * Looks for calls to `contains("posix")` on the `supportedFileAttributeViews` method returned by `FileSystem`.
```

java/ql/src/Security/CWE/CWE-200/TempDirLocalInformationDisclosure.ql
```diff
@@ -102,6 +103,21 @@ private class FileCreateTempFileSink extends FileCreationSink {
   }
 }
 
+/**
+ * A guard that checks what OS the program is running on.
```
Marcono1234, line 107: Isn't this only checking if the OS is Windows (based on the subclasses below), and not in general checking for any OS?

Marcono1234 (commented):
Would it make sense to also consider [`OperatingSystemMXBean.getName()`](https://docs.oracle.com/en/java/javase/17/docs/api/java.management/java/lang/management/OperatingSystemMXBean.html#getName())?

Sometimes the OS path separator (`File.pathSeparator`, `File.pathSeparatorChar`) and OS file path separator (`File.separator`, `File.separatorChar`) are used for OS detection; examples:
- https://github.com/openjdk/jdk/blob/d7f31d0d53bfec627edc83ceb75fc6202891e186/src/jdk.compiler/share/classes/com/sun/tools/javac/file/JavacFileManager.java#L673-L674
- https://github.com/openjdk/jdk/blob/d7f31d0d53bfec627edc83ceb75fc6202891e186/src/jdk.compiler/share/classes/com/sun/tools/sjavac/JavacState.java#L913

Would it make sense to consider these as well? (That is, comparison in `EqualityTest`, and maybe `EqualsMethod` call (qualifier or argument), where the compared value is a compile time constant)

java/ql/lib/semmle/code/java/os/OSCheck.qll
```diff
@@ -0,0 +1,93 @@
...
+    f.hasName(fieldName)
+  )
+}
+
+private class IsWindowsFromApacheCommons extends IsWindowsGuard instanceof FieldAccess {
+  IsWindowsFromApacheCommons() { isOsFromApacheCommons(this, "IS_OS_WINDOWS%") }
+}
+
+private class IsUnixFromApacheCommons extends IsUnixGuard instanceof FieldAccess {
+  IsUnixFromApacheCommons() { isOsFromApacheCommons(this, "IS_OS_UNIX") }
```
Marcono1234, line 66: Should this also consider Linux (`IS_OS_LINUX`) and Mac (`IS_OS_MAC%`) (and possibly some of the other Unix-like constants)?

java/ql/lib/semmle/code/java/StringCheck.qll
JLLeitschuh: I would be inclined to agree given the method name, but I think that `endsWith` isn't something that would work with either of the current uses of this utility method.

@smowton what do you think? Change the name? I was also thinking about a query being more flexible to include/exclude end checks.

java/ql/src/Security/CWE/CWE-200/TempDirLocalInformationDisclosure.ql
JLLeitschuh: Yes, good call, I'll update this

java/ql/lib/semmle/code/java/StringCheck.qll
smowton: If you need a rather irregular predicate like this make it `private`, name it something like `isOsCheckMatchingMethod` and note in the qldoc why it only considers particular methods

java/ql/lib/semmle/code/java/os/OSCheck.qll
JLLeitschuh: Added a test

java/ql/lib/semmle/code/java/os/OSCheck.qll
JLLeitschuh: I've been thinking about this over the past few days, and I don't know if it can be in all cases. I'm trying to think through the logic. Take this example:

```java
    /**
     * Should only be called on windows
     */
    private void onlyOnWindows() {}
    
    /**
     * Should only be called on unix like systems
     */
    private void onlyOnUnix() {}


    void test() {
        if (System.getProperty("os.name").contains("Windows")) {
            onlyOnWindows();
        }

        if (SystemUtils.IS_OS_UNIX) {
            onlyOnUnix();
        } else {
            onlyOnWindows();
        }

        if (!SystemUtils.IS_OS_LINUX) {
            // Might be another different unix-like system, so this can't be `onlyOnWindows()`.
        } else {
            onlyOnUnix();
        }
    }
```

I think the same issue exists with `"IS_OS_WINDOWS%"` above too.  If `SystemUtils.IS_OS_WINDOWS_2000` evaluates to `false`, it doesn't mean that you're on a unix system.

I need to fix this more generally. Trying to wrap my head around how to do so. @smowton, any suggestions you have here would be appreciated.

smowton: Sounds right to me: all the more specific tests are meaningful in the context of a positive finding but a negative one is uninformative. I would recommend doing this by categorising the various flags into those which indicates is-any-unix (which is roughly the negation of is-any-windows), vs. those that indicate is-specific-unix (and similarly is-specific-windows). It's then up to the configuration incorporating these into an isBarrierGuard predicate to only use is-specific-unix positively, while is-any-unix can be used in either direction.

JLLeitschuh:
@Marcono1234 ended up adding your OS check proposal here: [`3c53a05` (#8032)](https://github.com/github/codeql/pull/8032/commits/3c53a05e1622de7c599bf3eb61c661f600031cc0)

java/ql/lib/semmle/code/java/JDK.qll
```diff
@@ -244,6 +280,8 @@ class MethodAccessSystemGetProperty extends MethodAccess {
   /**
    * Holds if this call has a compile-time constant first argument with the value `propertyName`.
    * For example: `System.getProperty("user.dir")`.
+   *
+   * Note: Better to use `semmle.code.java.environment.SystemProperty#getSystemProperty` instead.
    */
   predicate hasCompileTimeConstantGetPropertyName(string propertyName) {
```
JLLeitschuh, line 286: @smowton wondering if I should deprecate this in favor of the new way of finding all expressions that access system properties

java/ql/lib/semmle/code/java/environment/SystemProperty.qll
```diff
@@ -0,0 +1,123 @@
...
+    or
+    f.getName() = "USER_HOME" and propertyName = "user.home"
+    or
+    f.getName() = "USER_LANGUAGE" and propertyName = "user.language"
+    or
+    f.getName() = "USER_NAME" and propertyName = "user.name"
+    or
+    f.getName() = "USER_TIMEZONE" and propertyName = "user.timezone"
+  )
+}
```
JLLeitschuh, line 99: GitHub copilot wrote 85% of this for me. I'm very impressed!

java/ql/lib/semmle/code/java/environment/SystemProperty.qll
```diff
@@ -0,0 +1,123 @@
...
+private MethodAccess getSystemPropertyFromApacheFileUtils(string propertyName) {
+  exists(Method m |
+    result.getMethod() = m and
+    m.getDeclaringType().hasQualifiedName("org.apache.commons.io", "FileUtils")
+  |
+    m.hasName(["getTempDirectory", "getTempDirectoryPath"]) and propertyName = "java.io.tmpdir"
+    or
+    m.hasName(["getUserDirectory", "getUserDirectoryPath"]) and propertyName = "user.home"
+  )
+}
```
JLLeitschuh, line 110: Both `getTempDirectory` and `getUserDirectory` return `File` instead of `String`:

 - https://github.com/apache/commons-io/blob/39b61a422e6bf87c47f15883a5b2f0d785e560fb/src/main/java/org/apache/commons/io/FileUtils.java#L1459-L1468
  - https://github.com/apache/commons-io/blob/39b61a422e6bf87c47f15883a5b2f0d785e560fb/src/main/java/org/apache/commons/io/FileUtils.java#L1481-L1490

I figure that this is still fine as it's essentially equivalent to the string variant, but wrapped in a different type

java/ql/lib/semmle/code/java/StringFormat.qll
```diff
@@ -312,27 +313,7 @@ private predicate formatStringValue(Expr e, string fmtvalue) {
...
-      e = f.getAnAccess() and
-      f.getDeclaringType() instanceof TypeFile and
-      fmtvalue = "x" // dummy value
-    |
-      f.hasName("pathSeparator") or
-      f.hasName("pathSeparatorChar") or
-      f.hasName("separator") or
-      f.hasName("separatorChar")
-    )
+    e = getSystemProperty(["line.separator", "file.separator", "path.separator"]) and fmtvalue = "x" // dummy value
```
JLLeitschuh, line 316: Simpler, and more comprehensive! 😄

java/ql/lib/semmle/code/java/environment/SystemProperty.qll
```diff
@@ -0,0 +1,231 @@
...
+    or
+    f.hasName("USER_HOME") and propertyName = "user.home"
+    or
+    f.hasName("USER_LANGUAGE") and propertyName = "user.language"
+    or
+    f.hasName("USER_NAME") and propertyName = "user.name"
+    or
+    f.hasName("USER_TIMEZONE") and propertyName = "user.timezone"
+  )
+}
```
JLLeitschuh, line 134: ~85% of this was written by GitHub copilot. Made it significantly easier!

java/ql/lib/semmle/code/java/os/OSCheck.qll
```diff
@@ -0,0 +1,158 @@
...
+ * Holds when the Guard is an equality check between the system property with the name `propertyName`
+ * and the string or char constant `compareToLiteral`.
+ */
+private Guard isOsFromSystemPropertyEqualityCheck(string propertyName, string compareToLiteral) {
+  result
+      .isEquality(getSystemProperty(propertyName),
+        any(Literal literal |
+          (literal instanceof CharacterLiteral or literal instanceof StringLiteral) and
+          literal.getValue() = compareToLiteral
+        ), _)
```
JLLeitschuh, line 80: @smowton I could probably simplify this logic with the following diff to `CompileTimeConstantExpr`. Do I have your approval to make the following change?

```diff
diff --git a/java/ql/lib/semmle/code/java/Expr.qll b/java/ql/lib/semmle/code/java/Expr.qll
index 5a991d8814..d98cdbb6b0 100755
--- a/java/ql/lib/semmle/code/java/Expr.qll
+++ b/java/ql/lib/semmle/code/java/Expr.qll
@@ -168,6 +168,8 @@ class CompileTimeConstantExpr extends Expr {
   string getStringValue() {
     result = this.(StringLiteral).getValue()
     or
+    result = this.(CharacterLiteral).getValue()
+    or
     result =
       this.(AddExpr).getLeftOperand().(CompileTimeConstantExpr).getStringValue() +
         this.(AddExpr).getRightOperand().(CompileTimeConstantExpr).getStringValue()

```

java/ql/lib/semmle/code/java/environment/SystemProperty.qll
```diff
@@ -0,0 +1,230 @@
...
+ */
+Expr getSystemProperty(string propertyName) {
+  result = getSystemPropertyFromSystem(propertyName) or
+  result = getSystemPropertyFromFile(propertyName) or
+  result = getSystemPropertyFromApacheSystemUtils(propertyName) or
+  result = getSystemPropertyFromApacheFileUtils(propertyName) or
+  result = getSystemPropertyFromGuava(propertyName) or
+  result = getSystemPropertyFromOperatingSystemMXBean(propertyName) or
+  result = getSystemPropertyFromSpringProperties(propertyName)
+}
```
JLLeitschuh, line 15: This API has been kicking around my head for the past few weeks. Trying to unify all the system property accessor `FieldAccess` & `MethodAccess` into a common API has been something I've been trying to think through how to do well.

I don't think this could be done with a class (given that it needs to take an argument), but if there's a better way to do this, I'm happy to hear it.

java/ql/lib/semmle/code/java/environment/SystemProperty.qll
```diff
@@ -0,0 +1,230 @@
...
+    f.hasName("JAVA_RUNTIME_NAME") and propertyName = "java.runtime.name"
+    or
+    f.hasName("JAVA_RUNTIME_VERSION") and propertyName = "java.runtime.version"
+    or
+    f.hasName("JAVA_SPECIFICATION_NAME") and propertyName = "java.specification.name"
+    or
+    f.hasName("JAVA_SPECIFICATION_VENDOR") and propertyName = "java.specification.vendor"
+    or
+    f.hasName("JAVA_UTIL_PREFS_PREFERENCES_FACTORY") and
+    propertyName = "java.util.prefs.PreferencesFactory"
```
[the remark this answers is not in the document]
JLLeitschuh: I'll leave a comment, that's a good thing to note in the code IMHO

java/ql/lib/semmle/code/java/JDK.qll
JLLeitschuh: As of my changes, it's only ever now used in my code, it isn't used anywhere else in this repository.

java/ql/lib/semmle/code/java/os/OSCheck.qll
JLLeitschuh: To make it easier, I broke this out into an independent PR: https://github.com/github/codeql/pull/8325

java/ql/lib/semmle/code/java/JDK.qll
JLLeitschuh: Also, at your request, I've improved my comment

java/ql/lib/semmle/code/java/JDK.qll
```diff
@@ -4,6 +4,7 @@
 
 import Member
 import semmle.code.java.security.ExternalProcess
+private import semmle.code.java.dataflow.FlowSteps
```
JLLeitschuh, line 7: I presume this is a pretty large import, I hope it's fine and won't break anything. Should I do this differently?

java/ql/lib/semmle/code/java/dataflow/FlowSources.qll
```diff
@@ -182,6 +183,8 @@ class EnvInput extends LocalUserInput {
     // Results from various specific methods.
     this.asExpr().(MethodAccess).getMethod() instanceof EnvReadMethod
     or
+    this.asExpr() = getSystemProperty(_)
+    or
```
JLLeitschuh, line 187: Found another really good place to put my new predicate! 😄

java/ql/src/change-notes/2022-02-14-os-guards.md
```diff
@@ -1,6 +1,8 @@
 ---
 category: minorAnalysis
 ---
- * Add new guards `IsWindowsGuard` and `IsUnixGuard` to detect OS specific guards.
+ * Add new guards `IsWindowsGuard`, `IsSpecificWindowsVariant`, `IsUnixGuard`, and `IsSpecificUnixVariant` to detect OS specific guards.
+ * Add new predicate `getSystemProperty` that gets all expressions that retrieve system properties from a variety of sources (eg. alternative JDK API's, Google Guava, Apache Commons, Apache IO, ect..).
  * Update "Local information disclosure in a temporary directory" (`java/local-temp-file-or-directory-information-disclosure`) to remove false-positives when OS is properly used as logical guard.
+ * Update "Local information disclosure in a temporary directory" (`java/local-temp-file-or-directory-information-disclosure`) to use `getSystemProperty` to resolve more
```
intrigus-lgtm, line 7: This sentence seems to be incomplete or at least missing a period.

JLLeitschuh: Indeed, it is incomplete. My bad

java/ql/lib/semmle/code/java/JDK.qll
```diff
@@ -228,11 +250,13 @@ class MethodSystemGetenv extends Method {
 /**
  * Any method named `getProperty` on class `java.lang.System`.
  */
-class MethodSystemGetProperty extends Method {
+class MethodSystemGetProperty extends ValuePreservingMethod {
```
smowton, line 253: This isn't a value-preserving method. A value-preserving method is `return this;` or `return param1;` or something else that conveys an exact reference across a function.

Do you mean to say you want `System.getProperty(x)` to generally conserve *taint*, such that if `x` is user-controlled then so is the result of `System.getProperty`? If so I'm not sure we'd want that as a general rule, since in order to be exploited it needs the user to know the name of a property containing sensitive information, or to have a way to get an arbitrary string *into* that environment.

java/ql/lib/semmle/code/java/frameworks/Properties.qll
```diff
@@ -1,15 +1,25 @@
...
+library class PropertiesGetPropertyMethod extends ValuePreservingMethod {
   PropertiesGetPropertyMethod() {
     getDeclaringType() instanceof TypeProperty and
     hasName("getProperty")
   }
+
+  override predicate returnsValue(int arg) { arg = 1 }
+}
+
+library class PropertiesGetMethod extends Method {
```
smowton, line 18: Don't use `library` in new code

JLLeitschuh: I did it for consistency with the rest of the file, should I drop `library` from the other classes in this file? I'd like them to be consistent.

java/ql/lib/semmle/code/java/JDK.qll
JLLeitschuh: `System.getProperty(propertyName, defaultValue)` is a value preserving method. Here's the code of this method:

```java
    /**
     * Searches for the property with the specified key in this property list.
     * If the key is not found in this property list, the default property list,
     * and its defaults, recursively, are then checked. The method returns the
     * default value argument if the property is not found.
     *
     * @param   key            the hashtable key.
     * @param   defaultValue   a default value.
     *
     * @return  the value in this property list with the specified key value.
     * @see     #setProperty
     * @see     #defaults
     */
    public String getProperty(String key, String defaultValue) {
        String val = getProperty(key);
        return (val == null) ? defaultValue : val;
    }
```

java/ql/lib/semmle/code/java/os/OSCheck.qll
```diff
@@ -0,0 +1,150 @@
...
+          .hasQualifiedName("java.util", "Set") and
+      m.hasName("contains")
+    ) and
+    this.getArgument(0).(CompileTimeConstantExpr).getStringValue() = "posix" and
+    exists(Method supportedFileAttribtueViewsMethod |
+      supportedFileAttribtueViewsMethod.hasName("supportedFileAttributeViews") and
+      supportedFileAttribtueViewsMethod.getDeclaringType() instanceof TypeFileSystem
+    |
+      DataFlow::localExprFlow(any(MethodAccess ma |
+          ma.getMethod() = supportedFileAttribtueViewsMethod
```
Marcono1234, line 146: Typo: Attrib<del>tu</del><ins>ut</ins>e
```suggestion
    exists(Method supportedFileAttributeViewsMethod |
      supportedFileAttributeViewsMethod.hasName("supportedFileAttributeViews") and
      supportedFileAttributeViewsMethod.getDeclaringType() instanceof TypeFileSystem
    |
      DataFlow::localExprFlow(any(MethodAccess ma |
          ma.getMethod() = supportedFileAttributeViewsMethod
```

java/ql/src/Security/CWE/CWE-200/TempDirLocalInformationDisclosure.ql
```diff
@@ -102,11 +103,36 @@ private class FileCreateTempFileSink extends FileCreationSink {
...
+  override predicate checks(Expr e, boolean branch) {
+    this.controls(e.getBasicBlock(), branch.booleanNot())
+  }
+}
+
+private class IsWindowsBarrierGuard extends WindowsOsBarrierGuard instanceof IsWindowsGuard {
+  override predicate checks(Expr e, boolean branch) { this.controls(e.getBasicBlock(), branch) }
+}
+
+private class IsAnyWindowsBarrierGuard extends WindowsOsBarrierGuard instanceof IsSpecificWindowsVariant {
```
Marcono1234, line 121:
```suggestion
private class IsSpecificWindowsBarrierGuard extends WindowsOsBarrierGuard instanceof IsSpecificWindowsVariant {
```

java/ql/src/change-notes/2022-02-14-os-guards.md
```diff
@@ -0,0 +1,8 @@
+---
+category: minorAnalysis
+---
+ * Add new guards `IsWindowsGuard`, `IsSpecificWindowsVariant`, `IsUnixGuard`, and `IsSpecificUnixVariant` to detect OS specific guards.
+ * Add new predicate `getSystemProperty` that gets all expressions that retrieve system properties from a variety of sources (eg. alternative JDK API's, Google Guava, Apache Commons, Apache IO, ect..).
```
Marcono1234, line 5: > Apache Commons, Apache IO

The library names here might be a misleading / incomplete because their full names are "Apache Commons Lang" and "Apache Commons IO".

Marcono1234, line 5:
```suggestion
 * Add new predicate `getSystemProperty` that gets all expressions that retrieve system properties from a variety of sources (e.g. alternative JDK API's, Google Guava, Apache Commons, Apache IO, etc.).
```

java/ql/src/change-notes/2022-02-14-os-guards.md
```diff
@@ -0,0 +1,8 @@
...
+ * Update "Local information disclosure in a temporary directory" (`java/local-temp-file-or-directory-information-disclosure`) to remove false-positives when OS is properly used as logical guard.
+ * Update "Local information disclosure in a temporary directory" (`java/local-temp-file-or-directory-information-disclosure`) to use `getSystemProperty` to use new `getSystemProperty` predicate.
```
Marcono1234, line 7: Duplicate text:
```suggestion
 * Update "Local information disclosure in a temporary directory" (`java/local-temp-file-or-directory-information-disclosure`) to use new `getSystemProperty` predicate.
```

java/ql/lib/semmle/code/java/frameworks/Properties.qll
smowton: Just don't add new code using it; we'll get around to a bulk cleanup at some point.

java/ql/lib/semmle/code/java/JDK.qll
smowton: Doh right, I'd forgotten those overloads existed.

java/ql/lib/semmle/code/java/environment/SystemProperty.qll
```diff
@@ -0,0 +1,249 @@
...
+  result = getSystemPropertyFromApacheFileUtils(propertyName) or
+  result = getSystemPropertyFromGuava(propertyName) or
+  result = getSystemPropertyFromOperatingSystemMXBean(propertyName) or
+  result = getSystemPropertyFromSpringProperties(propertyName)
+}
+
+private MethodAccess getSystemPropertyFromSystem(string propertyName) {
+  result.(MethodAccessSystemGetProperty).hasCompileTimeConstantGetPropertyName(propertyName)
+  or
+  exists(Method m | result.getMethod() = m | m.hasName("lineSeparator")) and
```
Marcono1234, line 23: Could probably simplify this:
```suggestion
  result.getMethod().hasName("lineSeparator") and
```

java/ql/src/change-notes/2022-02-14-os-guards.md
smowton, line 7:
```suggestion
 * Update "Local information disclosure in a temporary directory" (`java/local-temp-file-or-directory-information-disclosure`) to use new `getSystemProperty` predicate.
```

java/ql/src/change-notes/2022-02-14-os-guards.md
smowton, line 5:
```suggestion
 * Add new predicate `getSystemProperty` that gets all expressions that retrieve system properties from a variety of sources (eg. alternative JDK API's, Google Guava, Apache Commons, Apache IO, etc..).
```

java/ql/lib/semmle/code/java/environment/SystemProperty.qll
Marcono1234: Maybe it should be documented for `getSystemProperty` then, that the type of the expression is in some cases something other than a `String` (e.g. here a `File`, or for the separators a `char`).

java/ql/lib/semmle/code/java/JDK.qll
smowton: This is ok -- note that in FlowSteps.qll's `Frameworks` module you should private-import this file back. This is to ensure all queries using FlowSteps see the same set of standard value-preserving methods etc, and so the related QL can be evaluated once for the whole query suite, not re-evaluated per query as it would need to be if each one defined extra flow steps.

java/ql/lib/semmle/code/java/environment/SystemProperty.qll
```diff
@@ -0,0 +1,249 @@
...
+    ec.hasName("USER_HOME") and propertyName = "user.home"
+    or
+    ec.hasName("USER_DIR") and propertyName = "user.dir"
+  )
+}
+
+private MethodAccess getSystemPropertyFromOperatingSystemMXBean(string propertyName) {
+  exists(Method m |
+    m = result.getMethod() and
+    m.getDeclaringType().hasQualifiedName("java.lang.management", "OperatingSystemMXBean")
```
Marcono1234, line 232: Might be good to consider subtypes as well, because there is at least [`com.sun.management.OperatingSystemMXBean`](https://docs.oracle.com/en/java/javase/17/docs/api/jdk.management/com/sun/management/OperatingSystemMXBean.html)

Marcono1234 (commented):
As always, feel free to consider my comments only as suggestions; I am not a member of this project.

java/ql/src/change-notes/2022-02-14-os-guards.md
Marcono1234: Maybe also the following?
- "eg." -> "e.g."
- "etc.." -> "etc."

java/ql/lib/semmle/code/java/environment/SystemProperty.qll
JLLeitschuh: Indeed! Thanks

java/ql/test/library-tests/environment/SystemPropertyAccess.java
```diff
@@ -0,0 +1,25 @@
...
+public class SystemPropertyAccess {
+    private static final Properties SYSTEM_PROPERTIES = System.getProperties();
+
+    void test() {
+        System.getProperty("os.name");
+        System.getProperty("os.name", "default");
+        System.getProperties().getProperty("os.name");
+        System.getProperties().get("java.io.tmpdir");
+        SYSTEM_PROPERTIES.getProperty("java.home");
+        SYSTEM_PROPERTIES.get("file.encoding");
```
JLLeitschuh, line 14: @smowton is there a way to make these tests pass?

java/ql/lib/semmle/code/java/environment/SystemProperty.qll
Marcono1234: Ah you are right; `com.sun.management.OperatingSystemMXBean` does not override them, and there are no other subtypes which override these either. You could of course consider declarations by subtypes nonetheless in case the JDK overrides them at some point (e.g. to adjust Javadoc), but it is probably unlikely that this will happen.

java/ql/test/library-tests/environment/SystemPropertyAccess.java
Marcono1234: The problem might be that you are using local flow, but `SYSTEM_PROPERTIES` is a field defined outside of the method, therefore local flow does not consider it (if I understand it correctly). Maybe making it a local variable works.

(I hope it is alright that I answered this)

java/ql/lib/semmle/code/java/os/OSCheck.qll
```diff
@@ -0,0 +1,150 @@
...
+ * Holds when the Guard is an equality check between the system property with the name `propertyName`
+ * and the string or char constant `compareToLiteral`.
+ */
+private Guard isOsFromSystemPropertyEqualityCheck(string propertyName, string compareToLiteral) {
+  result
+      .isEquality(getSystemProperty(propertyName),
+        any(Literal literal |
+          (literal instanceof CharacterLiteral or literal instanceof StringLiteral) and
+          literal.getValue() = compareToLiteral
+        ), _)
```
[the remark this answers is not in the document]
JLLeitschuh: This is my understanding of guards from someone who's recently been introduced to them. So, this is my lay-person's explanation:

This predicate is is establishing that a guard exists, not whether or not the guard guarantees that a specific case is true or false.

IE. Both of these are 'windows guards':
 - `File.separatorChar == '\\'`
 - `File.separatorChar != '\\'
 
However, their end-use utility as guards is only in the larger context of code being guarded is only possible when they are used with `DataFlow::BarrierGuard`.

TL;DR: `Guard` merely establishes the existence of a guard. `BarrierGuard` establishes whether or not the Guard is relevant with respect to the true/false state of it's evaluation.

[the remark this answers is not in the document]
JLLeitschuh: Looks like my assessment here was kinda wrong. 🤔

java/ql/lib/semmle/code/java/environment/SystemProperty.qll
JLLeitschuh: Done

java/ql/lib/semmle/code/java/JDK.qll
JLLeitschuh: Done. Thanks!

java/ql/lib/semmle/code/java/os/OSCheck.qll
```diff
@@ -0,0 +1,150 @@
...
+  exists(StringPartialMatchMethod m, CompileTimeConstantExpr matchedStringConstant |
+    m = ma.getMethod() and
+    matchedStringConstant.getStringValue().toLowerCase().matches(osString)
+  |
+    DataFlow::localExprFlow(matchedStringConstant, ma.getArgument(m.getMatchParameterIndex()))
+  )
+}
+
+private class IsWindowsFromSystemProp extends IsWindowsGuard instanceof MethodAccess {
+  IsWindowsFromSystemProp() { isOsFromSystemProp(this, "window%") }
```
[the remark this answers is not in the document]
JLLeitschuh: I've updated the code to use a regular expression instead

java/ql/test/library-tests/environment/SystemPropertyAccess.java
JLLeitschuh: > (I hope it is alright that I answered this)

Yea, totally fine. Your feedback so far has been really helpful.

>  Maybe making it a local variable works.

It would, but I'd appreciate if something like `localDataFlow` supported static variable declarations like this

smowton: It shouldn't really because this flow is not local. I think I answered this in a different forum already, but you'd need to make a predicate that incorporates both local flow and initializer -> read steps.

java/ql/lib/semmle/code/java/environment/SystemProperty.qll
```diff
@@ -0,0 +1,272 @@
...
+ * Holds if data can flow from `e1` to `e2` in zero or more
+ * local (intra-procedural) steps or via local variable intializers
+ * for final variables.
+ */
+private predicate localExprFlowPlusInitializers(Expr e1, Expr e2) {
+  localFlowPlusInitializers(DataFlow::exprNode(e1), DataFlow::exprNode(e2))
+}
+
+/**
+ * Holds if data can flow from `node1` to `node2` in zero or more
```
smowton, line 262:
```suggestion
 * Holds if data can flow from `pred` to `succ` in zero or more
```

java/ql/lib/semmle/code/java/environment/SystemProperty.qll
```diff
@@ -0,0 +1,272 @@
...
+ * local (intra-procedural) steps or via local variable intializers
```
smowton, line 263:
```suggestion
 * local (intra-procedural) steps or via instance or static variable intializers
```

java/ql/lib/semmle/code/java/dataflow/FlowSources.qll
```diff
@@ -182,6 +183,8 @@ class EnvInput extends LocalUserInput {
     // Results from various specific methods.
     this.asExpr().(MethodAccess).getMethod() instanceof EnvReadMethod
     or
+    this.asExpr() = getSystemProperty(_)
```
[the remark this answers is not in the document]
JLLeitschuh: `line.separator` is something the user can control

https://stackoverflow.com/a/22681891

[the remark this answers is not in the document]
JLLeitschuh:
```java
/**
 * System Property initialization for internal use only
 * Retrieves the platform, JVM, and command line properties,
 * applies initial defaults and returns the Properties instance
 * that becomes the System.getProperties instance.
 */
public final class SystemProps {

    // no instances
    private SystemProps() {}

    /**
     * Create and initialize the system properties from the native properties
     * and command line properties.
     * Note:  Build-defined properties such as versions and vendor information
     * are initialized by VersionProps.java-template.
     *
     * @return a Properties instance initialized with all of the properties
     */
    public static Map<String, String> initProperties() {

        // Initially, cmdProperties only includes -D and props from the VM
        Raw raw = new Raw();
        HashMap<String, String> props = raw.cmdProperties();

        String javaHome = props.get("java.home");
        assert javaHome != null : "java.home not set";

        putIfAbsent(props, "user.home", raw.propDefault(Raw._user_home_NDX));
        putIfAbsent(props, "user.dir", raw.propDefault(Raw._user_dir_NDX));
        putIfAbsent(props, "user.name", raw.propDefault(Raw._user_name_NDX));

        // Platform defined encoding cannot be overridden on the command line
        put(props, "sun.jnu.encoding", raw.propDefault(Raw._sun_jnu_encoding_NDX));
        var nativeEncoding = ((raw.propDefault(Raw._file_encoding_NDX) == null)
                ? raw.propDefault(Raw._sun_jnu_encoding_NDX)
                : raw.propDefault(Raw._file_encoding_NDX));
        put(props, "native.encoding", nativeEncoding);

        // Add properties that have not been overridden on the cmdline
        putIfAbsent(props, "file.encoding", nativeEncoding);

        // Use platform values if not overridden by a commandline -Dkey=value
        // In no particular order
        putIfAbsent(props, "os.name", raw.propDefault(Raw._os_name_NDX));
        putIfAbsent(props, "os.arch", raw.propDefault(Raw._os_arch_NDX));
        putIfAbsent(props, "os.version", raw.propDefault(Raw._os_version_NDX));
        putIfAbsent(props, "line.separator", raw.propDefault(Raw._line_separator_NDX));
        putIfAbsent(props, "file.separator", raw.propDefault(Raw._file_separator_NDX));
        putIfAbsent(props, "path.separator", raw.propDefault(Raw._path_separator_NDX));
        putIfAbsent(props, "java.io.tmpdir", raw.propDefault(Raw._java_io_tmpdir_NDX));
        putIfAbsent(props, "http.proxyHost", raw.propDefault(Raw._http_proxyHost_NDX));
        putIfAbsent(props, "http.proxyPort", raw.propDefault(Raw._http_proxyPort_NDX));
        putIfAbsent(props, "https.proxyHost", raw.propDefault(Raw._https_proxyHost_NDX));
        putIfAbsent(props, "https.proxyPort", raw.propDefault(Raw._https_proxyPort_NDX));
        putIfAbsent(props, "ftp.proxyHost", raw.propDefault(Raw._ftp_proxyHost_NDX));
        putIfAbsent(props, "ftp.proxyPort", raw.propDefault(Raw._ftp_proxyPort_NDX));
        putIfAbsent(props, "socksProxyHost", raw.propDefault(Raw._socksProxyHost_NDX));
        putIfAbsent(props, "socksProxyPort", raw.propDefault(Raw._socksProxyPort_NDX));
        putIfAbsent(props, "http.nonProxyHosts", raw.propDefault(Raw._http_nonProxyHosts_NDX));
        putIfAbsent(props, "ftp.nonProxyHosts", raw.propDefault(Raw._ftp_nonProxyHosts_NDX));
        putIfAbsent(props, "socksNonProxyHosts", raw.propDefault(Raw._socksNonProxyHosts_NDX));
        putIfAbsent(props, "sun.arch.abi", raw.propDefault(Raw._sun_arch_abi_NDX));
        putIfAbsent(props, "sun.arch.data.model", raw.propDefault(Raw._sun_arch_data_model_NDX));
        putIfAbsent(props, "sun.os.patch.level", raw.propDefault(Raw._sun_os_patch_level_NDX));
        putIfAbsent(props, "sun.stdout.encoding", raw.propDefault(Raw._sun_stdout_encoding_NDX));
        putIfAbsent(props, "sun.stderr.encoding", raw.propDefault(Raw._sun_stderr_encoding_NDX));
        putIfAbsent(props, "sun.io.unicode.encoding", raw.propDefault(Raw._sun_io_unicode_encoding_NDX));
        putIfAbsent(props, "sun.cpu.isalist", raw.propDefault(Raw._sun_cpu_isalist_NDX));
        putIfAbsent(props, "sun.cpu.endian", raw.propDefault(Raw._sun_cpu_endian_NDX));

        /* Construct i18n related options */
        fillI18nProps(props,"user.language", raw.propDefault(Raw._display_language_NDX),
                raw.propDefault(Raw._format_language_NDX));
        fillI18nProps(props,"user.script",   raw.propDefault(Raw._display_script_NDX),
                raw.propDefault(Raw._format_script_NDX));
        fillI18nProps(props,"user.country",  raw.propDefault(Raw._display_country_NDX),
                raw.propDefault(Raw._format_country_NDX));
        fillI18nProps(props,"user.variant",  raw.propDefault(Raw._display_variant_NDX),
                raw.propDefault(Raw._format_variant_NDX));

        return props;
    }
```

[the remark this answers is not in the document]
smowton: It's hard to say without experimenting at length, but it looks like props that come from Raw might not let the user override them? In that function at least `native.,encoding` doesn't let the user override it.

At the very least this is surely a different PR.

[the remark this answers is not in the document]
JLLeitschuh: Can do

atorralba:
This has some QLDoc errors:

```
The following CodeQL elements are lacking documentation:
semmle/code/java/environment/SystemProperty.qll  SystemProperty                   file
semmle/code/java/frameworks/Properties.qll       Properties::PropertiesGetMethod  class
```

smowton:
Having to revert https://github.com/github/codeql/pull/8360 because of apparent performance problems. Remind me why you needed concatenation of string + char again?

If you still want it, a few possibilities to investigate:

1. Using test project https://lgtm.com/projects/g/geogebra/geogebra, why do we end up performing millions of string concatenations at considerable cost? Are there really millions of string + non-string-constant operations in the codebase?
2. If there really are some long concatenation sequences in the codebase causing this to cost a lot, consider a recursion that finds the parts of the concatenation and assigns them integer indices, before finally concatenating them in one go using QL's `concat` aggregate.

JLLeitschuh:
> Remind me why you needed concatenation of string + char again? 

For this particular change, I don't really, I wrote that new predicate to be more extensive than I needed so that it was generally more useful. In this particular PR, I just need to do equality checking between both string and characters because the new system property predicate sometimes returns a expression that has the type `char`.

Would adding the `cached` annotation to `getStrigifiedValue` and `getString` resolve the performance issues?

smowton:
No, `cached` is only useful if we're recomputing a particular predicate repeatedly. Here one evaluation of it is expensive.

JLLeitschuh:
> If there really are some long concatenation sequences in the codebase causing this to cost a lot, consider a recursion that finds the parts of the concatenation and assigns them integer indices, before finally concatenating them in one go using QL's concat aggregate.

I have no idea what this would look like unfortunately. I'd need some pointers on what the code would look like here.

smowton:
Probably not worth the difficulty just for this. I've made https://github.com/github/codeql/pull/8407 to revert the original addition of CharacterLiteral to getStringValue per Marcono's objections in #8325, and recommend you simply special-case CharacterLiteral in this PR as you did originally for the time being. We can investigate how to do this in a performant manner as and when the need arises.

smowton:
I also made a minor commit above to improve some class names.

JLLeitschuh:
Anything else required here?

smowton:
Just running a new performance evaluation as I realised the first one only considered queries with precision >= high, whereas the temp-dir-local-info query has precision = medium

smowton (approved):

JLLeitschuh:
Yeay! Very happy this is finally merged!

java/ql/src/Security/CWE/CWE-200/TempDirLocalInformationDisclosure.ql
```diff
@@ -102,11 +103,36 @@ private class FileCreateTempFileSink extends FileCreationSink {
...
+
+private class IsWindowsBarrierGuard extends WindowsOsBarrierGuard instanceof IsWindowsGuard {
+  override predicate checks(Expr e, boolean branch) { this.controls(e.getBasicBlock(), branch) }
+}
+
+private class IsSpecificWindowsBarrierGuard extends WindowsOsBarrierGuard instanceof IsSpecificWindowsVariant {
+  override predicate checks(Expr e, boolean branch) {
+    branch = true and this.controls(e.getBasicBlock(), branch)
+  }
+}
```
aschackmull, line 125: These three barrier-guards don't really make sense. Using `Guard.controls` in the `checks` predicate of a barrier-guard is almost guaranteed to be completely wrong, as the barrier-guard abstraction is explicitly designed to handle the `.controls`-aspect of guards.
The expression `e` should be the variable access being checked by the guard, and the `branch` should be the result of the guard that indicates that `e` is safe. So the basic implementation hint for the `checks` predicate should just be to use the AST and nothing more. If you really want to use `Guard.controls` to pick out the nodes that are sanitised then you're really just defining a regular sanitizer and not a sanitizer-guard.

JLLeitschuh: So, I partially understand what you're saying. That being said, I don't know how to fix it.

Also, could you suggest a test case that this doesn't work for that I could use as a unit test

aschackmull: A barrierguard is just syntactic sugar for a sanitizer with a specific pattern. To see how a barrier-guard gets desugared into a sanitizer, look at the `getAGuardedNode()` predicate:
```
  final Node getAGuardedNode() {
    exists(SsaVariable v, boolean branch, RValue use |
      this.checks(v.getAUse(), branch) and
      use = v.getAUse() and
      this.controls(use.getBasicBlock(), branch) and
      result.asExpr() = use
    )
  }
```
You end up just joining `this.controls` with itself, which semantically is mostly harmless. But you can simplify things by just defining the sanitizers directly. E.g. replace `IsSpecificWindowsBarrierGuard` with the following:
```
abstract private class WindowsOsSanitizer extends DataFlow::Node { }

private class IsSpecificWindowsSanitizer extends WindowsOsSanitizer instanceof IsSpecificWindowsVariant {
  IsSpecificWindowsSanitizer() {
    any(IsSpecificWindowsVariant guard).control(this.asExpr().getBasicBlock(), true)
  }
}

...

    node instanceof WindowsOsSanitizer
  }
```
This is mostly equivalent to your `IsSpecificWindowsBarrierGuard` except it drops the restriction that only uses of ssa variables are sanitized.
The other two, `IsNotUnixBarrierGuard` and `IsWindowsBarrierGuard`, can be similarly rewritten, but note that they are somewhat broken as-is, since they currently guard both of their branches. In
```
if (<IsWindowsBarrierGuard>) {
  foo
} else {
  bar
}
```
you're sanitizing both `foo` and `bar`.

java/ql/lib/semmle/code/java/os/OSCheck.qll
```diff
@@ -0,0 +1,161 @@
...
+  IsUnixFromSystemProp() {
+    isOsFromSystemProp(this, any(string s | s.regexpMatch(["mac.*", "linux.*"])))
+  }
+}
+
+bindingset[fieldNamePattern]
+private predicate isOsFromApacheCommons(FieldAccess fa, string fieldNamePattern) {
+  exists(Field f | f = fa.getField() |
+    f.getDeclaringType() instanceof TypeApacheSystemUtils and
+    f.getName().matches(fieldNamePattern)
```
aschackmull, line 110: I guess it is unlikely that this will give spurious matches, but do note that `_` is a wildcard for the `.matches` predicate, so all of the underscores in the names below could be matched with anything.

PR #8036, opened 2022-02-15
C++: Remove some unused legacy relations from the DB scheme - Take 2

jketema:

* `snapshotDate`
* `duplicateCode`
* `similarCode`
* `tokens`

This is similar to the reverted https://github.com/github/codeql/pull/7982, except that `externalData` is kept, as it is used by the field team.


MathiasVP (approved):
LGTM! :shipit:

PR #8038, opened 2022-02-15
C#: Refactor Structural Comparison for Control Flow Elements.

michaelnebel:
In this PR we re-factor the existing implementation of the structural comparison for controls flow elements.
The new approach is based on a *global value number* approach.

The idea is strongly linked to the global value numbering used in the type unification.
Please note that the implementation is still hidden under the existing interface, but we plan to make a follow up PR where this is removed.


csharp/ql/test/library-tests/structuralcomparison/structuralComparison.ql
```diff
@@ -0,0 +1,25 @@
...
+    e2.fromSource() and
+    e1 != e2 and
+    e1.getEnclosingCallable() = e2.getEnclosingCallable() and
+    not e1 instanceof ThisAccess
+  }
+}
+
+query predicate same(ControlFlowElement e1, ControlFlowElement e2) {
+  exists(StructuralComparisonTest sct | sct.same(e1, e2))
+}
```
tamasvajk, line 23: You could remove the duplicates by enforcing some location based order between `e1` and `e2`, something along these lines:
```
exists(StructuralComparisonTest sct, Location l1, Location l2 | 
  sct.same(e1, e2) and 
  l1 = e1.getLocation() and 
  l2 = e2.getLocation() and 
  (l1.getStartLine() < l2.getStartLine() or
   l1.getStartLine() = l2.getStartLine() and l1.getStartColumn() < l2.getStartColumn())
)
```

tamasvajk (commented):
Added a couple of comments. This is a low confidence first review.

BTW, what's the benefit of doing the comparison with GVN vs the previous structural comparison?

michaelnebel:
> Added a couple of comments. This is a low confidence first review.
> 
> BTW, what's the benefit of doing the comparison with GVN vs the previous structural comparison?

Thank you for looking into this!
The benefit should be better performance (and maybe also simpler code). Unfortunately, the DCA run shows a performance regression. I suspect that that the cause of this is due to some missing pragmas (some `noinline` and `nomagic`).

csharp/ql/lib/semmle/code/csharp/commons/StructuralComparison.qll
```diff
@@ -14,17 +53,146 @@ private Declaration referenceAttribute(Expr e) {
...
+  /** Gets the string representation of this global value number. */
+  abstract string toString();
+}
+
+private class ConstantGvn extends Gvn, TConstantGvn {
+  override string toString() { this = TConstantGvn(result) }
+}
+
+private class ListGvn extends Gvn, TListGvn {
+  private GvnList l;
```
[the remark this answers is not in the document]
michaelnebel: Yes, you are absolutely right. I have flattened the ADT type(s), which means that they now look a bit more general, but they still only contain the same values as the "restrictions" has been changed.

csharp/ql/test/library-tests/structuralcomparison/structuralComparison.ql
michaelnebel: That is a good idea!

michaelnebel:
@hvitved: The DCA run now completed and it looks like there are no overall performance regressions. Some queries themselves might be a bit slower now, but maybe that could be linked to building the cache? There was identified a single "join order badness", but that is not side effect of this change.

hvitved (changes requested):
Looks very good to me! I have made some (mostly trivial) suggestions [here](https://github.com/michaelnebel/codeql/pull/6).

hvitved (approved):

PR #8041, opened 2022-02-15
Add version of `prepare-db-upgrade.sh` supporting multiple languages

jketema:


misc/scripts/prepare-db-upgrade.sh
```diff
@@ -0,0 +1,168 @@
...
+--lang <LANG>
+        Language to update the schema for.
+
+Must be run within the git repo needing an update.
+EOF
+  exit "${exit_code}"
+}
+
+prev_hash="origin/main"
+lang="cpp"
```
nickrolfe, line 33: Could we remove this, and make the `--lang` flag mandatory? If you forget to provide it after making a change to, say, the Ruby dbscheme, it just tells you there's no work to do, because it was looking at the C++ dbscheme.

misc/scripts/prepare-db-upgrade.sh
```diff
@@ -0,0 +1,168 @@
...
+    ;;
+  ruby)
+    scheme_file="${lang}/ql/lib/${lang}.dbscheme"
+    ;;
+  *)
+    usage 2 "Unrecognised language: ${lang}"
+    ;;
+esac
+
+qldir="${top_level}/ql"
```
nickrolfe, line 100: This should make it possible to run from either repository:
```suggestion
qldir="${top_level}"                                                                        
if [ ! -f "${qldir}/${scheme_file}" ]; then
  qldir="${top_level}/ql"
  if [ ! -f "${qldir}/${scheme_file}" ]; then                                               
    echo "Cannot find ${qldir}/${scheme_file}"                                             
    exit 2                                                                                  
  fi
fi
```

nickrolfe (commented):
Thanks for doing this!

misc/scripts/prepare-db-upgrade.sh
jketema: Agreed, and done.

misc/scripts/prepare-db-upgrade.sh
jketema: Is the aim to support both `codeql` being checked out as a submodule and independently?

nickrolfe: That's right. I imagine someone working on C/C++ would likely run this from the submodule, while someone working on Ruby (where everything, including the extractor, is self-contained in the `codeql` repo) would run it from an independent checkout of `codeql`.

jketema: I wonder if we should go for something simpler based on the script location:
```
app_dir="$(dirname "$0")"
cd ${app_dir}
qldir="$(git rev-parse --show-toplevel)"

if [ "x${qldir}" = "x" ]; then
  usage 2 "Script not in a code repository."
fi
```

nickrolfe: Yeah, that's certainly better.

jketema: Done.

nickrolfe (approved):
LGTM 👍

PR #8044, opened 2022-02-16
Language-agnostic document on db up-/downgrades

nickrolfe:
And make the old Python and Ruby helper scripts forward to the new language-independent version added in https://github.com/github/codeql/pull/8041.


nickrolfe:
Python/Ruby folks: I could also just delete the old `prepare-db-upgrade.sh` files, if you prefer, instead of making them forward to the language-independent script.

aibaars:
LGTM, I'd just delete the scripts instead of forwarding.

docs/prepare-db-upgrade.md
```diff
@@ -0,0 +1,135 @@
...
+### Testing your scripts
+
+Although we have some automated testing of the scripts (e.g. to test that you can upgrade databases all the way from the an initial dbscheme to the newest, and back), it's essential that you apply some more rigorous testing any non-trivial upgrade or downgrade. You might do so as follows:
+
+#### Running qltests
+
+To test the upgrade script, run:
+
+```
+codeql test run --search-path=<old-extractor-pack>;ql <test-dir>
```
aibaars, line 72: `;` is the windows path separator. In addition, I think you no longer need to specify `ql` in the search-path
```suggestion
codeql test run --search-path=<old-extractor-pack> <test-dir>
```

docs/prepare-db-upgrade.md
```diff
@@ -0,0 +1,135 @@
...
+```
+
+Where `<old-extractor-pack>` is an extractor pack containing the old extractor and dbscheme that pre-date your changes, and `<test-dir>` is the directory containing the qltests for your language. This will run the tests using an old extractor, and the test databases will all be upgraded in place using your new upgrade script.
+
+To test the downgrade script, create an extractor pack that includes your new dbscheme and extractor changes. Then checkout the `main` branch of `codeql` (i.e. a branch that does not include your changes), and run:
+
+```
+codeql test run --search-path=<new-extractor-pack>;ql <test-dir>
```
aibaars, line 80:
```suggestion
codeql test run --search-path=<new-extractor-pack> <test-dir>
```

tausbn (dismissed):
Looks good to me! I would favour just deleting instead of forwarding.

docs/prepare-db-upgrade.md
nickrolfe: When I try without `:ql`, I get

```
Could not evaluate queries in .../ql/test/library-tests/ast/control: The CodeQL database is not compatible with a QL library  referenced by the query you are trying to run.
	 CodeQL database location: ql/test/library-tests/ast/control/control.testproj
	Query location: ql/test/library-tests/ast/control/CaseExpr.ql
The database may be too new for the QL libraries the query is using; try upgrading them.
```

aibaars: Ah ok, that's strange. `codeql` is able to compile the test query which (transitively) depends on `codeql/ruby-all`, so I wonder why it cannot find the upgrade scripts that are inside that pack without adding a search path. @aeisenberg do you understand this?

Instead of using `;` I think you can also repeat the `--search-path` flag .

docs/prepare-db-upgrade.md
```diff
@@ -69,15 +69,15 @@ Although we have some automated testing of the scripts (e.g. to test that you ca
...
+codeql test run --search-path=<old-extractor-pack> --search-path=ql <test-dir>
 ```
 
 Where `<old-extractor-pack>` is an extractor pack containing the old extractor and dbscheme that pre-date your changes, and `<test-dir>` is the directory containing the qltests for your language. This will run the tests using an old extractor, and the test databases will all be upgraded in place using your new upgrade script.
 
 To test the downgrade script, create an extractor pack that includes your new dbscheme and extractor changes. Then checkout the `main` branch of `codeql` (i.e. a branch that does not include your changes), and run:
 
 ```
-codeql test run --search-path=<new-extractor-pack>;ql <test-dir>
+codeql test run --search-path=<new-extractor-pack> --search-path=ql <test-dir>
```
aibaars, line 80: I couldn't get this command to run somehow. Does it work for you?

nickrolfe: Yes, it works for me, although it appears the `--search-path=ql` is not needed in this case.

aibaars: ok, not sure what I was doing wrong. I did expect that `--search-path=ql` wasn't needed, but failed to validate that. Thanks for confirming. Shall we remove it ?

aibaars (approved):
LGTM, perhaps remove the `--search-path=ql` flag from the command to test downgrade scripts.

aibaars (approved):

PR #8054, opened 2022-02-16
JS: split request forgery query into server-side and client-side variants

asgerf:
The `RequestForgery` query was previously `@precision medium` and I'd like to raise this to `high`.

However, it flags both server-side request forgery (SSRF) and client-side request forgery (which is the same, except the browser/client is sending the request) and I only want to promote the actual SSRF variant.

The client-side variant tends to be much less severe than SSRF, and as far as I can tell, it doesn't have a dedicated CWE number like SSRF. Even the name is something I had to make up.

Note that what I call "client-side request forgery" should not to be confused with *cross*-site request forgery, which is what CSRF and XSRF typically stand for.  We could also call it `FrontendRequestForgery` if we want to avoid confusion with CSRF. 🤷 

The query has been split as follows:
- `js/request-forgery` is now specific to SSRF and promoted to `@precision high`, which means being shown by default.
- To avoid regressing on overall coverage, a new query `js/client-side-request-forgery` covers the client-side cases, and remains in `security-extended`.
- The new query has a security-severity of `5.0`. There's no real CWE number to derive this from, so I chose it to be 1 point below XSS, as it can _sometimes_ lead to XSS, and it is unlikely to escalate to anything worse than that.

The old qhelp and metadata was a bit unclear on what exactly causes request forgery to happen. I rewrote parts of it to emphasize how the different parts of the the URL play a role in request forgery.

The metadata changes were also ported to other languages, and involve a full `@name` change to `Server-side request forgery`. The only non-JS changes are in https://github.com/github/codeql/commit/38afb0c40823fef9edbd65c11d82dc0540d1ca4d. Other language teams: please review and let me know what you think.

[Evaluation](https://github.com/github/codeql-dca-main/tree/data/asgerf/split-request-f__default__security-extend__1/reports) looks fine. 5 results were moved from `js/request-forgery` to `js/client-side-request-forgery`.


michaelnebel (approved):
C#. LGTM.

erik-krogh (commented):
Looks good :+1:  
I just got a few minor comments.

erik-krogh (approved):

javascript/ql/src/Security/CWE-918/ClientSideRequestForgery.qhelp
```diff
@@ -0,0 +1,64 @@
...
+
+        <p>
+            For example, given the query string <code>message_id=../pastebin/123</code>, the request will
+            end up targeting the <code>/api/pastebin</code> endpoint. Or if there is an open redirect on the login page,
+            a query string like <code>message_id=../../login?redirect_url=https://evil.com</code> could give
+            the attacker full control over the response as well.
+        </p>
+
+        <p>
+            In example below, the input has been restricted to a number so the endpoint cannot be altered:
```
hubwriter, line 51:
```suggestion
            In example below, the input has been restricted to a number so that the endpoint cannot be altered:
```

hubwriter (approved):

javascript/ql/src/Security/CWE-918/ClientSideRequestForgery.qhelp
asgerf: Looks like the PR was merged a bit early. I'll open a new PR with this change

asgerf: Opened in https://github.com/github/codeql/pull/8186

PR #8059, opened 2022-02-16
C++: new query for insufficient key strength

rdmarsh2:
Add a new query for insufficient key strength with prime-based RSA, DSA, and Diffie-Hellman key exchange. This query currently covers the OpenSSL/LibreSSL/BoringSSL APIs.

### DIFF
--- a/cpp/ql/src/Security/CWE/CWE-326/InsufficientKeySize.ql
+++ b/cpp/ql/src/Security/CWE/CWE-326/InsufficientKeySize.ql
@@ -1,0 +1,50 @@
+/**
+ * @name Use of a cryptographic algorithm with insufficient key size
+ * @description Using cryptographic algorithms with too small a key size can
+ *              allow an attacker to compromise security.
+ * @kind problem
+ * @problem.severity error
+ * @precision high
+ * @id cpp/insufficient-key-size
+ * @tags security
+ *       external/cwe/cwe-326
+ */
+
+import cpp
+import semmle.code.cpp.ir.dataflow.DataFlow
+import semmle.code.cpp.ir.IR
+
+int getMinimumKeyStrength(string func) {
+  func = "EVP_PKEY_CTX_set_dsa_paramgen_bits" and result = 2048
+  or
+  func = "EVP_PKEY_CTX_set_dh_paramgen_prime_len" and result = 2048
+  or
+  func = "EVP_PKEY_CTX_set_rsa_keygen_bits" and result = 2048
+}
+
+class KeyStrengthFlow extends DataFlow::Configuration {
+  KeyStrengthFlow() {
+    this = "KeyStrengthFlow"
+  }
+
+  override predicate isSource(DataFlow::Node node) {
+    node.asInstruction() instanceof IntegerConstantInstruction
+  }
+
+  override predicate isSink(DataFlow::Node node) {
+    exists(FunctionCall fc, string name|
+      node.asExpr() = fc.getArgument(1) and
+      fc.getTarget().hasGlobalName(name) and
+      exists(getMinimumKeyStrength(name))
+    )
+  }
+}
+
+from DataFlow::PathNode source, DataFlow::PathNode sink, KeyStrengthFlow conf, FunctionCall fc, string name, int bits
+where
+  conf.hasFlowPath(source, sink) and
+  sink.getNode().asExpr() = fc.getArgument(1) and
+  fc.getTarget().hasGlobalName(name) and
+  bits = getMinimumKeyStrength(name) and
+  source.getNode().asInstruction().(ConstantValueInstruction).getValue().toInt() <  bits
+select fc, source, sink, "The key size $@ is insufficient for security", source, source.toString()
--- a/cpp/ql/src/Security/CWE/CWE-326/InsufficientKeySize.ql
+++ b/cpp/ql/src/Security/CWE/CWE-326/InsufficientKeySize.ql
@@ -31,7 +31,10 @@ class KeyStrengthFlow extends DataFlow::Configuration {
   KeyStrengthFlow() { this = "KeyStrengthFlow" }
 
   override predicate isSource(DataFlow::Node node) {
-    node.asInstruction().(IntegerConstantInstruction).getValue() != "0"
+    exists(int bits |
+    node.asInstruction().(IntegerConstantInstruction).getValue().toInt() = bits and
+    bits < getMinimumKeyStrength(_, _) and
+    bits > 0 // exclude sentinel values
   }
 
   override predicate isSink(DataFlow::Node node) {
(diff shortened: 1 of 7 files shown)


cpp/ql/test/query-tests/Security/CWE/CWE-326/InsufficientKeySize.expected
```diff
@@ -0,0 +1,3 @@
+| test.cpp:32:5:32:38 | call to EVP_PKEY_CTX_set_dsa_paramgen_bits | test.cpp:32:45:32:48 | 1024 | test.cpp:32:45:32:48 | 1024 | The key size $@ is insufficient for security | test.cpp:32:45:32:48 | 1024 | 1024 |
+| test.cpp:33:5:33:42 | call to EVP_PKEY_CTX_set_dh_paramgen_prime_len | test.cpp:33:49:33:52 | 1024 | test.cpp:33:49:33:52 | 1024 | The key size $@ is insufficient for security | test.cpp:33:49:33:52 | 1024 | 1024 |
+| test.cpp:35:5:35:36 | call to EVP_PKEY_CTX_set_rsa_keygen_bits | test.cpp:35:43:35:46 | 1024 | test.cpp:35:43:35:46 | 1024 | The key size $@ is insufficient for security | test.cpp:35:43:35:46 | 1024 | 1024 |
```
geoffw0, line 3: I think the `.expected` needs regenerating.  These line numbers don't seem to align.

cpp/ql/src/Security/CWE/CWE-326/InsufficientKeySize.ql
```diff
@@ -0,0 +1,56 @@
...
+ * @id cpp/insufficient-key-size
+ * @tags security
+ *       external/cwe/cwe-326
+ */
+
+import cpp
+import semmle.code.cpp.ir.dataflow.DataFlow
+import semmle.code.cpp.ir.IR
+
+// Holds if `func` is the name of an encryption function that accepts a key size as parameter `paramIndex`
```
geoffw0, line 17:
```suggestion
// Gets the recommended minimum key size (in bits) of `func`, the name of an encryption function that accepts a key size as parameter `paramIndex`
```

cpp/ql/src/Security/CWE/CWE-326/InsufficientKeySize.ql
```diff
@@ -0,0 +1,56 @@
...
+    exists(FunctionCall fc, string name, int param |
+      node.asExpr() = fc.getArgument(param) and
+      fc.getTarget().hasGlobalName(name) and
+      exists(getMinimumKeyStrength(name, param))
+    )
+  }
+}
+
+from
+  DataFlow::PathNode source, DataFlow::PathNode sink, KeyStrengthFlow conf, FunctionCall fc,
```
geoffw0, line 46: You're using `DataFlow::PathNode`s here, but it is not a `path-problem`.

cpp/ql/src/Security/CWE/CWE-326/InsufficientKeySize.ql
```diff
@@ -0,0 +1,56 @@
...
+  string name, int bits
+where
+  conf.hasFlowPath(source, sink) and
+  sink.getNode().asExpr() = fc.getArgument(1) and
```
geoffw0, line 50: I think the `1` here should be the parameter index from `getMinimumKeyStrength` rather than a hardcoded value.

geoffw0 (commented):
This looks fantastic!

I've made a few nitpicks.  I'd also like to see some real-world results and a change note.

saragraber08:
https://1drv.ms/u/s!AsVxiwxAk0yxgQx0-lGo_vltbJ1C

saragraber08:
> https://1drv.ms/u/s!AsVxiwxAk0yxgQx0-lGo_vltbJ1C

Please helps me

geoffw0:
I've started an all-LGTM run (linked from the issue).

saragraber08:
https://1drv.ms/u/s!AsVxiwxAk0yxgQx0-lGo_vltbJ1C hey omgash thank you

saragraber08:
Hey do you mind walking me through the steps I'm new to this and I really
need the information

cpp/ql/src/Security/CWE/CWE-326/InsufficientKeySize.ql
```diff
@@ -0,0 +1,56 @@
...
+ * @problem.severity error
+ * @precision high
+ * @id cpp/insufficient-key-size
+ * @tags security
+ *       external/cwe/cwe-326
+ */
+
+import cpp
+import semmle.code.cpp.ir.dataflow.DataFlow
+import semmle.code.cpp.ir.IR
```
MathiasVP, line 15: Given the number of re-evaluation issues we've had in the past, I wouldn't be surprised if we need to import GVN before importing the IR here. But let's revisit that once we have a DCA run.

saragraber08: Thank you so much

rdmarsh2: DCA doesn't look like there's reevaluation. I see a 2-4% slowdown depending on project. 4% is pretty high, I'll see if I can make some small changes to improve it.

geoffw0:
I'm happy with this PR.  Are there still any performance concerns / improvements to be made?

MathiasVP:
Before we merge this PR we should also add the ready-for-docs-review label.

rdmarsh2:
I'm still not really happy with performance but the run time is almost entirely dataflow evaluation, so I'm not sure there's anything to do.

MathiasVP:
> I'm still not really happy with performance but the run time is almost entirely dataflow evaluation, so I'm not sure there's anything to do.

Yeah, this is one of those places where it would be really beneficial to start stage 1 of dataflow at the sinks instead of at the sources. I imagine there are _many_ `IntegerConstantInstruction` instructions with a value that's not `0`, but very few of those actually flow to one of the sinks. So stage 1 of dataflow is probably quite expensive.

MathiasVP:
Here's an idea: How about defining sources as only those `IntegerConstantInstruction ` whose value is lower than any result returned by `getMinimumKeyStrength`? It might not do a lot, but it'll rule out some integer constants.

ethanpalm (approved):
From a docs point of view, this looks great ⚡

rdmarsh2:
> Here's an idea: How about defining sources to also only include those IntegerConstantInstruction whose value is lower than some result returned by getMinimumKeyStrength? It might not do a lot, but it'll rule out some integer constants.

Good suggestion. I've made the change and I'll do a fresh DCA run.

rdmarsh2:
Performance looks good in the new DCA run, and I've just fixed the tests. I think we're ready to merge now.

MathiasVP (approved):
LGTM!

PR #8061, opened 2022-02-17
Python: Add data-flow through Django ORM models

RasmusWL:
There was one big item I had almost left out :flushed: namely whether these new data-flow steps should always be active, or whether they should only be active when the right imports are made. After looking at what is done in C#, I have made them always be active.


python/ql/lib/semmle/python/frameworks/Django.qll
```diff
@@ -645,6 +805,166 @@ module PrivateDjango {
...
+
+            override predicate jumpStep(DataFlow::Node nodeFrom, DataFlow::Node nodeTo) {
+              // save -> synthetic
+              exists(API::Node modelClass, DataFlow::MethodCallNode saveCall |
+                // TODO: The `nodeTo` should be restricted more, such that flow to
+                // base-classes are only for the fields that are defined in the
+                // base-class... but only passing on flow for a specific attribute requires flow-summaries,
+                // so we can do
+                // `obj (in obj.save call) ==read of attr==> synthetic attr on base-class ==store of attr==> synthetic for base-class`
+                nodeTo = nodeToStoreIn(modelClass, _) and
```
yoff, line 954: I believe that `nodeToStoreIn(modelClass, _)` does indeed restrict to fields defined by the base class?

python/ql/lib/semmle/python/dataflow/new/internal/DataFlowPrivate.qll
```diff
@@ -3,6 +3,16 @@ private import DataFlowPublic
...
+// Since we allow extra data-flow steps from modeled frameworks, we import these
+// up-front, to ensure these are included. This provides a more seamless experience from
+// a user point of view, since they don't need to know they need to import a specific
+// set of .qll files to get the same data-flow steps as they are used to seeing. This
+// also ensures that we don't end up re-evaluating data-flow because it has different
+// global steps in some configurations.
+//
+// This matches behavior in C#.
+private import semmle.python.Frameworks
+private import semmle.python.Concepts
```
yoff, line 15: If we import `Concepts`, do we also need to import `Frameworks`?

yoff: I am a bit wary of this, but only because I vaguely remember a debug sessions with @tausbn where we needed to avoid it. Perhaps he can recall some details..

yoff (changes requested):
Excellent, thorough testing. And the thoughtful comments explained most of the questions that I had right away.

 In general, I think the modelling quite comprehensive, and the bits that are left out quite reasonable (and well argued).

python/ql/lib/semmle/python/dataflow/new/internal/DataFlowPrivate.qll
tausbn: What you mention _is_ stirring some disused neurons at the back of my brain, but unfortunately I can't remember off-hand what the issue was.

I think the use here is probably fine. I guess the only objection would be that you can't turn them off again. No matter what, you'll get the ORM-based data-flow steps, even if you don't want them. 

(I guess this _might_ matter if someone were to, say, define a local version of `django` that doesn't use the same way of passing data through models. I think this is very unlikely, though.)

python/ql/test/library-tests/frameworks/django-orm/dummy_import.py
```diff
@@ -0,0 +1,5 @@
+# to force extractor to see files. since we use `--max-import-depth=1`, we use this
+# "fake" import that doesn't actually work, but tricks the python extractor to look at
+# all the files
```
yoff, line 3: Is this something either we or customers need to be aware of when analysing projects in the wild?

python/ql/lib/semmle/python/dataflow/new/internal/DataFlowPrivate.qll
RasmusWL: > If we import Concepts, do we also need to import Frameworks?

Technically no, since we know that Concepts imports Frameworks... if you would like me to change that part, I can certainly do that :+1:

> I think the use here is probably fine. I guess the only objection would be that you can't turn them off again. No matter what, you'll get the ORM-based data-flow steps, even if you don't want them.

Yes, agreed that there _are_ downsides to this. I think this won't be a problem for 99.999999% of the cases, so I would much rather have the positives effects mentioned in the comment from this :blush:

python/ql/test/library-tests/frameworks/django-orm/dummy_import.py
RasmusWL: no. we run the extractor in a different way when used in tests and with the codeql cli -- this is sort of annoying, but the way it is :neutral_face: :shrug: (I looked into changing this some while back, but it was not easy)

python/ql/lib/semmle/python/frameworks/Django.qll
```diff
@@ -565,45 +565,205 @@ module PrivateDjango {
...
+              //
+              // The big benefit is for how we could then model `values`/`values_list`. For example,
+              // `Model.objects.value_list(name, description)` would result in (for the attribute description)
+              // 0: <Synthetic node for Model> -- [attr description]
+              // 1: ==> Model.objects [ListElement, attr description]
+              // 2: ==> .value_list(...) [ListElement, TupleIndex 1]
+              //
+              // but for now, we just model a store step directly from the synthetic
+              // node to the method call.
+              // extra method on query-set/manager that does _not_ return a query-set
```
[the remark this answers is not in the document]
RasmusWL: Thanks, will just change it to highlight that this is a collection of instances

```suggestion
              // node to the method call.
              //
              // extra method on query-set/manager that does _not_ return a query-set,
              // but a collection of instances.
```

python/ql/lib/semmle/python/frameworks/Django.qll
RasmusWL: Not quite. The problem is https://github.com/github/codeql/pull/8061/files/27d5349a745afb6da1fb516276b838ee75afddd5#diff-4899b0abafadf88280bc9f62bb280ebc9c4369071b008e853464b0b184539b3eR162

Since there had been a store of a different subclass with the same field-name

```
class PolyBook(PolymorphicModel):
    title = models.CharField(max_length=256)


class PolyPhysicalBook(PolyBook):
    physical_location = models.CharField(max_length=256)
    same_name_different_value = models.CharField(max_length=256)


class PolyEBook(PolyBook):
    download_link = models.CharField(max_length=256)
    same_name_different_value = models.CharField(max_length=256)

def poly_save_physical_book():
    return PolyPhysicalBook.objects.create(
        title=SOURCE,
        physical_location=SOURCE,
        same_name_different_value=SOURCE,
    )
```

we think there is flow from the SOURCE to the `same_name_different_value` field _on the base class `PolyBook`_ (although that field doesn't exist on the base-class), and therefore also to the subclass `PolyEBook` -- but there is no such flow.

python/ql/lib/semmle/python/dataflow/new/internal/DataFlowPrivate.qll
yoff: Alright, let's ship to learn on that one. (And I hope I will not regret this when doing flow summaries...).

For the import, I think I would prefer to be minimal, since seeing both confused me. I do see the situation where we no longer need concepts and forget to import frameworks...but that should be caught quickly...

RasmusWL:
```suggestion
private import semmle.python.Frameworks
```

RasmusWL, line 15:
```suggestion
private import semmle.python.Frameworks
```

python/ql/test/library-tests/frameworks/django-orm/dummy_import.py
yoff: Cool, thanks.

yoff (approved):
LGTM

RasmusWL:
Nice, tests looks good (expect for some QL warnings that looks like FPs to me), so will merge this :+1:

PR #8138, opened 2022-02-20
Ruby: Implement `FileSystemWriteAccess` concept

alexrford:
e.g. `File.write`, `IO.print`, etc.

This necessitated some restructuring in `Files.qll` to clear things up a bit. `IO`/`File` read/write calls in general are mostly modelled in the abstract `IOOrFileMethodCall` and its implementations `IOOrFileReadMethodCall` and `IOOrFileWriteMethodCall`. The `IO::FileReader` and similar classes are then restrictions of those classes.


hmac (approved):
This seems like a nice improvement!

PR #8142, opened 2022-02-21
Ruby: Add IncompleteMultiCharSanitization query

hmac:


ruby/ql/lib/codeql/ruby/regexp/RegExpTreeView.qll
```diff
@@ -440,6 +456,24 @@ class RegExpSequence extends RegExpTerm, TRegExpSequence {
   }
 
   override string getAPrimaryQlClass() { result = "RegExpSequence" }
+
+  override predicate matchesEmptyString() {
+    forall(RegExpTerm child | child = this.getAChild() | child.matchesEmptyString())
+  }
+
+  // Why can't we use concat(...) with language[monotonicAggregates] here instead?
```
aibaars, line 464: This predicate looks like it could have very many results. I couldn't find an equivalent in the JS library, perhaps for a good reason.

ruby/ql/lib/codeql/ruby/regexp/RegExpTreeView.qll
```diff
@@ -268,6 +268,12 @@ class RegExpTerm extends RegExpParent {
 
   /** Gets the primary QL class for this term. */
   override string getAPrimaryQlClass() { result = "RegExpTerm" }
+
+  /** Holds if this regular expression term can match the empty string. */
+  predicate matchesEmptyString() { none() }
+
+  /** Gets a string matched by this regular expression. */
+  string getAMatch() { none() }
```
aibaars, line 276: In the JS library these methods are called `isNullable` and `getAMatchedString`. I actually like your names more, but perhaps maintaining consistency across languages is more important.

ruby/ql/lib/codeql/ruby/security/IncompleteMultiCharacterSanitizationQuery.qll
```diff
@@ -0,0 +1,209 @@
+/**
```
aibaars, line 1: I think it should be possible to share most of this file with the Javascript variant.

ruby/ql/src/queries/security/cwe-116/IncompleteMultiCharacterSanitization.ql
```diff
@@ -0,0 +1,27 @@
...
+ */
+
+import ruby
+import codeql.ruby.frameworks.core.String
+import codeql.ruby.DataFlow
+import codeql.ruby.security.IncompleteMultiCharacterSanitizationQuery
+import codeql.ruby.security.IncompleteMultiCharacterSanitizationSpecific as Specific
+
+from
+  StringSubstitutionCall replace, Specific::EmptyReplaceRegExpTerm dangerous, string prefix,
```
aibaars, line 23: You could reduce the number of imports a bit by declaring more the types like this:
```suggestion
  DataFlow::Node replace, RegExpTerm dangerous, string prefix,
```
The `hasResult` predicate takes care of restricting the types.

Alternatively you can remove the entire `from .. where .. select` and write:

```
query predicate problems = hasResult/4;
```
An perhaps even better, you can rename `hasResult` to `problems` and make it a `query` predicate in the QLL file. That way all you have to do in the QL file is `import codeql.ruby.security.IncompleteMultiCharacterSanitizationQuery`

ruby/ql/lib/codeql/ruby/security/IncompleteMultiCharacterSanitizationQuery.qll
```diff
@@ -0,0 +1,37 @@
...
+private import codeql.ruby.frameworks.core.String
+private import codeql.ruby.dataflow.internal.DataFlowDispatch
+private import codeql.ruby.security.IncompleteMultiCharacterSanitization
+
+/**
+ * Holds if `replace` has a pattern argument containing a regular expression
+ * `dangerous` which matches a dangerous string beginning with `prefix`, in
+ * attempt to avoid a vulnerability of kind `kind`.
+ */
+predicate hasResult(
```
aibaars, line 15: Let's try :
```suggestion
query predicate problems(
```
This is the name of the "alerts" result set.

aibaars (commented):
This looks good to me. Let's ask the JavaScript team for approval too.

ruby/ql/src/queries/security/cwe-116/IncompleteMultiCharacterSanitization.ql
aibaars: The predicate should be problems(replace, message, dangerous, prefix). Just drop the kind parameter and insert the message string with placeholder as the second parameter.

javascript/ql/lib/semmle/javascript/security/IncompleteMultiCharacterSanitization.qll
```diff
@@ -0,0 +1,202 @@
+/**
```
erik-krogh, line 1: Should this file instead be named `IncompleteMultiCharacterSanitizationQuery.qll`? Such that we use the `Query.qll` suffix. 

Otherwise LGTM.

erik-krogh (dismissed):
JS 👍

erik-krogh (approved):
JS still 👍

alexrford (approved):

PR #8149, opened 2022-02-21
Shared: use shared access path syntax to parse arguments in CSV rows

asgerf:
Previously it was up to each language to parse the numeric arguments in an access path token, such as `Argument[1..3]`. This has lead to some duplication and inconsistency, with plenty of `regexpCapture` calls doing similar things.

This PR places some common parsing primitives into the `AccessPathSyntax.qll` file and uses those where possible.

Note that for the _identifying_ access path (in dynamic languages), the syntax `Argument[N-1]` means the last argument of a call, and `Argument[N-2]` the second-last, etc. For the time being at least, this syntax is specific to the identifying access path and therefore to dynamic languages, but I still think the shared `AccessPathSyntax.qll` file is the best place for this code to live. The predicate for parsing it, `parseIntWithExplicitArity`, can only be invoked with a known arity, and the call-site arity is not generally known when interpreting an input/output spec (for static langauges, it should be known to the author writing the spec, so there is no need to use the `N-1` syntax). I'm mentioning this here in case you're wondering why this code appears to be unused for your language.

Also replaces a few `regexpCapture` calls for extracting a field name from a token such as `Field[x]`. The qualified name of a field can contain commas (at least for C#), so to avoid splitting it, we extract it with `getArgumentList()` rather than `getAnArgument()`.

### DIFF
--- a/java/ql/lib/semmle/code/java/dataflow/internal/AccessPathSyntax.qll
+++ b/java/ql/lib/semmle/code/java/dataflow/internal/AccessPathSyntax.qll
@@ -6,12 +6,110 @@
  * (which does not use the shared data flow libraries).
  */
 
+/**
+ * Convenience-predicate for extracting two capture groups at once.
+ */
+bindingset[input, regexp]
+private predicate regexpCaptureTwo(string input, string regexp, string capture1, string capture2) {
+  capture1 = input.regexpCapture(regexp, 1) and
+  capture2 = input.regexpCapture(regexp, 2)
+}
+
 /** Companion module to the `AccessPath` class. */
 module AccessPath {
   /** A string that should be parsed as an access path. */
   abstract class Range extends string {
     bindingset[this]
     Range() { any() }
+  }
+
+  /**
+   * Parses an integer constant `n` or interval `n1..n2` (inclusive) and gets the value
+   * of the constant or any value contained in the interval.
+   */
+  bindingset[arg]
+  int parseInt(string arg) {
+    result = arg.toInt()
+    or
+    // Match "n1..n2"
+    exists(string lo, string hi |
+      regexpCaptureTwo(arg, "(-?\\d+)\\.\\.(-?\\d+)", lo, hi) and
+      result = [lo.toInt() .. hi.toInt()]
+    )
+  }
+
+  /**
+   * Parses a lower-bounded interval `n..` and gets the lower bound.
+   */
+  bindingset[arg]
+  private int parseLowerBound(string arg) {
+    result = arg.regexpCapture("(-?\\d+)\\.\\.", 1).toInt()
+  }
+
+  /**
+   * Parses an integer constant or interval (bounded or unbounded) that explicitly
+   * references the arity, such as `N-1` or `N-3..N-1`.
+   *
+   * Note that expressions of form `N-x` will never resolve to a negative index,
+   * even if `N` is zero (it will have no result in that case).
+   */
+  bindingset[arg, arity]
+  private int parseIntWithExplicitArity(string arg, int arity) {
+    result >= 0 and // do not allow N-1 to resolve to a negative index
+    exists(string lo |
+      // N-x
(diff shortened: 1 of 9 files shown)


java/ql/lib/semmle/code/java/dataflow/internal/AccessPathSyntax.qll
```diff
@@ -6,13 +6,111 @@
...
+      lo = arg.regexpCapture("N-(\\d+)\\.\\.", 1) and
+      result = [arity - lo.toInt(), arity - 1]
+    )
+    or
+    exists(string lo, string hi |
+      // x..N-y
+      regexpCaptureTwo(arg, "(-?\\d+)\\.\\.N-(\\d+)", lo, hi) and
+      result = [lo.toInt() .. arity - hi.toInt()]
+      or
+      // N-x..Ny
```
smowton, line 74:
```suggestion
      // N-x..N-y
```

smowton (approved):
Java version of this looks fine

ruby/ql/lib/codeql/ruby/dataflow/internal/FlowSummaryImplSpecific.qll
```diff
@@ -76,15 +76,10 @@ SummaryComponent interpretComponentSpecific(string c) {
   result = FlowSummary::SummaryComponent::arrayElementUnknown()
   or
   exists(int i |
-    c.regexpCapture("ArrayElement\\[([0-9]+)\\]", 1).toInt() = i and
```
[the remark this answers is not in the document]
nickrolfe: This looks fine – I can see that you've moved the restriction to nonnegative indices up to `arrayElementKnown`.

ruby/ql/lib/codeql/ruby/dataflow/internal/FlowSummaryImplSpecific.qll
```diff
@@ -76,15 +76,10 @@ SummaryComponent interpretComponentSpecific(string c) {
...
+    c.getName() = "ArrayElement" and
+    i = AccessPath::parseInt(c.getAnArgument()) and
     result = FlowSummary::SummaryComponent::arrayElementKnown(i)
   )
-  or
-  exists(int i1, int i2 |
-    c.regexpCapture("ArrayElement\\[([-0-9]+)\\.\\.([0-9]+)\\]", 1).toInt() = i1 and
-    c.regexpCapture("ArrayElement\\[([-0-9]+)\\.\\.([0-9]+)\\]", 2).toInt() = i2 and
-    result = FlowSummary::SummaryComponent::arrayElementKnown([i1 .. i2])
-  )
```
nickrolfe, line 87: It took me a second to realise why this could be deleted, but it looks like this is now handled as part of the call to `parseInt` above, on line 80.

nickrolfe (approved):

tamasvajk (commented):
Are there any performance implications of these changes?

michaelnebel:
> Are there any performance implications of these changes?

Maybe run DCA before merging?

asgerf:
DCA evaluations were quite flaky, but overall doesn't indicate a slow-down:
- [Ruby](https://github.com/github/codeql-dca-main/tree/data/asgerf/use-shared-acce__Differences__code-scanning__3/reports)
- [Java](https://github.com/github/codeql-dca-main/tree/data/asgerf/use-shared-acce__Differences__code-scanning__2/reports) - this was particularly noisy, but symmetric on both failures and timings.
- [JavaScript](https://github.com/github/codeql-dca-main/tree/data/asgerf/use-shared-acce__Differences__code-scanning__1/reports#query-run-timings-per-query-single--multi-threaded)
- [C#](https://github.com/github/codeql-dca-main/tree/data/asgerf/use-shared-acce__Differences__code-scanning/reports)

michaelnebel (approved):
Looks like a really nice improvement on both readability and sharing!

PR #8150, opened 2022-02-21
JS: preparation for sharing MaD library based on API graphs

asgerf:
Some preparation to make the MaD prototype in JavaScript usable from Ruby. Only JS code is actually affected by this PR.

- Renames `{Shared,Impl}.qll` to `ApiGraphModels` and `ApiGraphModelsSpecfic`. The `Specific` suffix matches the convention for files providing language-specific details in the shared data flow library.
- Makes the `API::InvokeNode` type be provided by the language, as not all languages use the term `InvokeNode`.
- Makes the base case of access path resolution be language-specific. That is, the `moduleImport` part is now JS-specific. In Ruby, the base case looks quite different as members of an imported package are just flooded into the local scope where it is imported (or something to that effect), and you can't really see locally which access paths refer to something from the imported package.
- Finally, `Member[x]` is now treated as a language-specific token. Ruby has both `obj.foo` and `obj::foo` and it's a bit ambiguous which of these is a "member". It might still have a token named `Member` but I'm not convinced the implementation can be shared (i.e. the corresponding predicate on an API graph node might be called something different).


erik-krogh (approved):
:+1:

PR #8151, opened 2022-02-21
JS: Separate the ML model for ML-powered queries into its own pack

henrymercer:
This has a number of benefits, including enabling users to more easily get started with development: running `codeql pack install` from the `-queries` pack will now install the ML model.

Commit-by-commit review recommended: one of the commits documents why the ML model pack isn't part of the default workspace.


annarailton (approved):
Really positive change, thank you!

Issue #8069, opened 2022-02-19
Useless upcast - false positive


sungam3r:
https://github.com/graphql-dotnet/graphql-dotnet/security/code-scanning/417?query=ref%3Arefs%2Fheads%2Fdevelop
Whole file: https://github.com/graphql-dotnet/graphql-dotnet/blob/e0470b2f461edfba651ad4976770c341fe8e1aab/src/GraphQL/TaskExtensions.cs#L31-L31

sungam3r:
One more example: https://github.com/graphql-dotnet/graphql-dotnet/security/code-scanning/768?query=ref%3Arefs%2Fheads%2Fdevelop

sungam3r:
https://github.com/graphql-dotnet/graphql-dotnet/security/code-scanning/712?query=ref%3Arefs%2Fheads%2Fdevelop

I get CS0411 if remove this cast.

RasmusWL:
Hi @sungam3r, thanks for reporting this, I've forwarded it to the the C# team, that are looking at this :+1:

tamasvajk:
@sungam3r Thank you for reporting this issue. It is being fixed in https://github.com/github/codeql/pull/8152.

PR #8152, opened 2022-02-21
C# Exclude dynamic casts from useless casts check

tamasvajk:
This PR removes all casts to `dynamic` from the useless upcast check. The issue was reported in https://github.com/github/codeql/issues/8069.

### DIFF
--- a/csharp/ql/src/Language Abuse/UselessUpcast.ql
+++ b/csharp/ql/src/Language Abuse/UselessUpcast.ql
@@ -207,6 +207,17 @@ class ExplicitUpcast extends ExplicitCast {
   }
 }
 
+// Stop external filepaths from appearing in the results
+class ValueOrRefTypeLocation extends ValueOrRefType {
+  override predicate hasLocationInfo(string path, int sl, int sc, int el, int ec) {
+    exists(string fullPath | super.hasLocationInfo(fullPath, sl, sc, el, ec) |
+      if exists(this.getFile().getRelativePath())
+      then path = fullPath
+      else path = fullPath.regexpReplaceAll(".*/", "<external>/")
+    )
+  }
+}
+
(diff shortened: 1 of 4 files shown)


csharp/ql/src/Language Abuse/UselessUpcast.ql
```diff
@@ -204,6 +204,19 @@ class ExplicitUpcast extends ExplicitCast {
     this = any(LocalVariableDeclAndInitExpr decl | decl.isImplicitlyTyped()).getInitializer()
     or
     exists(LambdaExpr c | c.canReturn(this))
+    or
+    dest instanceof DynamicType
+  }
+}
+
+// Stop external filepaths from appearing in the results
+class ValueOrRefTypeLocation extends ValueOrRefType {
```
michaelnebel, line 213: Why is this needed?

michaelnebel (approved):
Looks plausible to me!

csharp/ql/src/Language Abuse/UselessUpcast.ql
tamasvajk: Ahh, thanks. This should not be added, especially not here (it would only have a place in tests). In the test we're listing types, which are not necessarily `fromSource`. This means that their path would point to framework locations, which is platform and installation dependent. 

This change is not needed, because we exclude the only framework type that would show up in the tests. See [here](https://github.com/github/codeql/pull/8152/commits/7cba3c59515017d322e2256c0ce8a0d90338deae#diff-afbeb8fdd27e6e411446f8b13e44a33c1f08a527498d1cf1aa96ad84bc502cebL9).

hvitved (approved):

michaelnebel (approved):
Looks good to me!

PR #8155, opened 2022-02-21
Java: Exclude files with errors from modulus analysis

smowton:
A user pointed out a case where partial extraction results including an SSA variable graph with no initialiser (`while(1) { int i = i - 1; }`) causes effectively-infinite recursion in computing modulus candidates.


aschackmull:
This does not look like the right fix to me - ad-hoc-fixing one specific library isn't the way to go. There are lots of places where we assume that the code we analyse has actually compiled successfully, so if this assumption is somehow wrong then we should either make sure it's right or revisit the entire stack of QL libraries.

PR #8164, opened 2022-02-22
Ruby: interpret string escape sequences in getConstantValue()

nickrolfe:
Motivation: being able to write queries that can tell the difference between the string values `'\n'` (U+005C, U+006E) and `"\n"` (U+000A).

I've added a new test in `ruby/ql/test/library-tests/ast/escape_sequences` that I hope covers all the edge cases, and I think all the other test changes make sense.

I'm not sure if any of the new unescape predicates need to be cached. I'll run DCA and see if anything stands out.


ruby/ql/lib/codeql/ruby/ast/internal/Literal.qll
```diff
@@ -3,6 +3,7 @@ private import AST
 private import Constant
 private import TreeSitter
 private import codeql.ruby.controlflow.CfgNodes
+private import codeql.NumberUtils
 
 int parseInteger(Ruby::Integer i) {
```
aibaars, line 8: The implementation of `parseInteger` should probably use `NumberUtils`

nickrolfe: OK, I'll have a go at that.

ruby/ql/lib/codeql/NumberUtils.qll
```diff
@@ -0,0 +1,97 @@
...
+  result = 13 and hex = ["d", "D"]
+  or
+  result = 14 and hex = ["e", "E"]
+  or
+  result = 15 and hex = ["f", "F"]
+}
+
+/** Gets the integer value of the `octal` char. */
+private int toOctal(string octal) {
+  octal = "0" and result = 0
```
aibaars, line 70: If I recall correctly the `parseInteger` implementation uses something like `result = "012345678".indexOf(octal)` . That's shorter, not sure if it is more efficient .

nickrolfe: OK, let's go with that.

ruby/ql/lib/codeql/ruby/ast/internal/Literal.qll
nickrolfe: I've done this locally, but my implementation is more conservative. My new `parseHexInt` returns no value for `0xdeadbeef`, while the current `parseInteger` implementation returns `-559038737` (i.e. wraps to a negative number). Ruby treats `0xdeadbeef` as the positive/unsigned integer `3735928559`.

Should I attempt to update my implementation of `parseHexInt` to allow wrapping, even if it's not correct?

aibaars: There is no problem changing the meaning of `parseInteger` if the old definition is incorrect. Returning no value is probably better than incorrect wrapping.

nickrolfe:
The DCA run showed a significant slowdown on one project in particular. I'm investigating.

hvitved (commented):
LGTM, but as you say, performance needs to be investigated.

hvitved (dismissed):

PR #8165, opened 2022-02-22
JS: support more property writes in js/prototype-pollution-utility

erik-krogh:
and generalize `ObjectDefinePropertyAsPropWrite`

Gets a TP/TN for CVE-2021-23470


asgerf (approved):
LGTM

This makes `PropWrite.getRhs()` recursive, but I guess that's fine.

PR #8166, opened 2022-02-22
Ruby/Python: regex parser: group sequences of 'normal' characters

aibaars:
The Javascript library groups sequences of normal characters, it would be good to make the behaviour of the regex libraries behave as similarly to each other as possible.

### DIFF
--- a/python/ql/lib/semmle/python/regex.qll
+++ b/python/ql/lib/semmle/python/regex.qll
@@ -427,6 +427,7 @@ abstract class RegexString extends Expr {
   }
 
   predicate normalCharacter(int start, int end) {
+    end = start + 1 and
     this.character(start, end) and
     not this.specialCharacter(start, end, _)
   }
@@ -447,16 +448,16 @@ abstract class RegexString extends Expr {
   }
 
   /**
-   * A sequence of 'simple' characters.
+   * Holds if the range [start:end) consists of only 'normal' characters.
    */
-  predicate simpleCharacterSequence(int start, int end) {
-    // a simple character inside a character set is interpreted on its own
-    this.simpleCharacter(start, end) and
+  predicate normalCharacterSequence(int start, int end) {
+    // a normal character inside a character set is interpreted on its own
+    this.normalCharacter(start, end) and
     this.inCharSet(start)
     or
-    // a maximal run of simple characters is considered as one constant
+    // a maximal run of normal characters is considered as one constant
     exists(int s, int e |
-      e = max(int i | simpleCharacterRun(s, i)) and
+      e = max(int i | this.normalCharacterRun(s, i)) and
       not this.inCharSet(s)
     |
       // 'abc' can be considered one constant, but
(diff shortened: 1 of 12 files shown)


aibaars:
@yoff Could you have a look at this?

python/ql/lib/semmle/python/regex.qll
```diff
@@ -446,6 +446,45 @@ abstract class RegexString extends Expr {
...
+   * A sequence of 'normal' characters.
+   */
+  predicate normalCharacterSequence(int start, int end) {
+    this.normalCharacter(start, end) and
+    end = start + 1 and
+    exists(int x, int y | this.charSet(x, y) and x <= start and y >= end)
+    or
+    exists(int s, int e |
+      e = max(int i | normalCharacterSub(s, i)) and
+      not exists(int x, int y | this.charSet(x, y) and x <= s and y >= e)
```
tausbn, line 459: Having recently fixed some unfortunate uses of `not exists` in this file, I was a bit wary of this construction. I think it's probably fine, though, since we're materialising the `charSet` relation anyway.

python/ql/test/library-tests/regex/FirstLast.expected
```diff
@@ -86,30 +86,30 @@
 | ^[A-Z_]+$(?<!not-this) | last | 1 | 7 |
 | ^[A-Z_]+$(?<!not-this) | last | 1 | 8 |
 | ^[A-Z_]+$(?<!not-this) | last | 8 | 9 |
-| ^[A-Z_]+$(?<!not-this) | last | 20 | 21 |
+| ^[A-Z_]+$(?<!not-this) | last | 13 | 21 |
 | ax{01,3} | first | 0 | 1 |
 | ax{01,3} | last | 1 | 2 |
 | ax{01,3} | last | 1 | 8 |
-| ax{01,3} | last | 7 | 8 |
```
yoff, line 93: This (and others like it) looks like an existing bug. I think it is due to [this line](https://github.com/github/codeql/blob/main/python/ql/lib/semmle/python/regex.qll#L414) in `simpleCharacter` not considering qualifiers longer than one character (such as the multiples notation `{n, m}`).

yoff (changes requested):
Great to see this, thanks for doing it!
I believe that what you are doing, is changing `character` from being either a `simpleCharacter` or an `escapedCharacter` to now being either a maximal `simpleCharacter` run or an `escapedCharacter`. I have therefor suggested to rename some of the predicates.

I also wonder if the computation of maximal runs is standard or if a more efficient one might exist, but if the performance looks fine, then we can punt this question.

python/ql/lib/semmle/python/regex.qll
tausbn, line 459:
```suggestion
      not exists(int x, int y | this.charSet(x, y) and x <= s and e <= y)
```
As above.

tausbn (commented):
A few comments. Otherwise this looks good to me. :+1:

python/ql/lib/semmle/python/regex.qll
```diff
@@ -446,6 +446,45 @@ abstract class RegexString extends Expr {
     )
   }
 
+  /**
+   * A sequence of 'normal' characters.
+   */
+  predicate normalCharacterSequence(int start, int end) {
+    this.normalCharacter(start, end) and
+    end = start + 1 and
+    exists(int x, int y | this.charSet(x, y) and x <= start and y >= end)
```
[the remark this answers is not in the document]
aibaars: Replaced with `this.inCharSet(start)` as suggested by `@yoff`

python/ql/lib/semmle/python/regex.qll
aibaars: Replaced with `this.inCharSet(start)` as suggested by `@yoff`

aibaars:
> I see from the failed tests that something is off with using simple chars instead of normal chars. I would like to understand what that is, but it does not have to be during this PR.

I think `simpleCharacter` does not exclude characters that are part of escape sequences.

aibaars:
@yoff @tausbn I think I addressed your comments.

yoff (approved):
LGTM, is the DCA run from earlier still valid?

yoff:
> I think `simpleCharacter` does not exclude characters that are part of escape sequences.
Ah, yes, that might be it :-)

aibaars:
> LGTM, is the DCA run from earlier still valid?

Yes, the [`force-push`](https://github.com/github/codeql/compare/ca0e26f275e44246a35fc0328667930a73be6a7d..0c23f5815fbe221222592f5670037462ed9d901c) only fixed a test case.

PR #8168, opened 2022-02-22
JS: model hapi handler returns as reflected-xss sinks

esbena:
According to [a hapi documentation example](https://hapi.dev/tutorials/routing?lang=en_US#path), route handler responses should return escaped HTML, so I have added those returns as sinks for the `js/reflected-xss` query.


asgerf (approved):
👍  LGTM assuming evaluation works out

esbena:
The evaluation is uninteresting.

PR #8171, opened 2022-02-22
update ATM NosqlInjection and SqlInjection query docs

kaeluka:
The docs from the ATM NoSqlInjection query were about SQL injections. This PR fixes that, and also polishes the ATM SQL Injection query docs.


henrymercer:
Thanks!

hubwriter (approved):

PR #8172, opened 2022-02-22
add pre-commit configuration

redsun82:
This enables use of the `pre-commit` framework to run quick pre-commit
checks. In particular this allows to automatically fix:
* trailing white spaces
* absence or multiple newlines at the end of files
* QL code formatting
* file sync

More could be added in the future: anything that can be checked fast
can be added in the configuration (for example well-formedness of
`qldoc` files).

This is a purely opt-in feature. Instructions for enabling it and
possibly configuring its behaviour are in `pre-commit-hook-setup.md`.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -22,7 +22,7 @@ repos:
         entry: python3 config/sync-files.py --latest
         pass_filenames: false
 
-    -   id: qlhelp
+    -   id: qhelp
         name: Check query help generation
         files: \.qhelp$
         language: system
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -7,6 +7,7 @@ repos:
     hooks:
     -   id: trailing-whitespace
     -   id: end-of-file-fixer
+
 -   repo: local
     hooks:
     -   id: codeql-format
@@ -14,8 +15,15 @@ repos:
         files: \.qll?$
         language: system
         entry: codeql query format --in-place
+
     -   id: sync-files
         name: Fix files required to be identical
         language: system
         entry: python3 config/sync-files.py --latest
         pass_filenames: false
+
+    -   id: qlhelp
+        name: Check query help generation
+        files: \.qhelp$
+        language: system
+        entry: python3 misc/scripts/check-qhelp.py
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,0 +1,21 @@
+# See https://pre-commit.com for more information
+# See https://pre-commit.com/hooks.html for more hooks
+exclude: /test/.*$(?<!\.ql)(?<!\.qll)(?<!\.qlref)
+repos:
+-   repo: https://github.com/pre-commit/pre-commit-hooks
+    rev: v3.2.0
+    hooks:
+    -   id: trailing-whitespace
+    -   id: end-of-file-fixer
+-   repo: local
+    hooks:
+    -   id: codeql-format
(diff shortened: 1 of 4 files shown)


criemen (approved):
LGTM, but please wait with merging for approval of somebody who's actually writing QL code.

intrigus-lgtm:
I have written a custom pre-commit script some time ago that includes checking query help generation.
Feel free to integrate it either into https://github.com/github/codeql/blob/main/misc/scripts/pre-commit or the `pre-commit` framework. (I don't have time right now to do this myself).
```bash
#!/bin/bash

exec 1>&2
exitVal=0
while read -r f
do
  filename="${f##*/}"
  extension="${filename##*.}"
  p="$PWD/$f";
  if [[ -f "$p" ]] && { [ "$extension" == "ql" ] || [ "$extension" == "qll" ]; }
  then
    if ! codeql query format --check-only "$p"
    then
      exitVal=1
    fi
  fi
  if [[ -f "$p" ]] && { [ "$extension" == "qhelp" ]; }
  then
    if ! codeql generate query-help -vvv --log-to-stderr --format markdown -- "$p"
    then
      exitVal=1
    fi
  fi
done <<<"$(git diff --cached --relative --name-only)"
exit $exitVal
```

.pre-commit-config.yaml
```diff
@@ -0,0 +1,21 @@
...
+    hooks:
+    -   id: codeql-format
+        name: Fix QL file formatting
+        files: \.qll?$
+        language: system
+        entry: codeql query format --in-place
+    -   id: sync-files
+        name: Fix files required to be identical
+        language: system
+        entry: python3 config/sync-files.py --latest
```
MathiasVP, line 20: I'm a bit reluctant to suggest scripts that automatically rewrites your commits (I personally run my pre-commit hook for `codeql query format` without `--in-place`), but if people like automatic fixes like this I won't stand in their way 😄.

MathiasVP (approved):
LGTM. I've only added a very opinionated comment that's totally not a blocker for this PR.

.pre-commit-config.yaml
redsun82: Tomorrow I will have a look at the `pre-commit install --config` option, because that could provide an easy way to give the two flavours of hooks (auto-fixing vs only checking) already pre-made :slightly_smiling_face:

redsun82: yes, `--config` is definitely the way to go, I've updated `pre-commit-hook-setup.md` instructions accordingly.

.pre-commit-config.yaml
```diff
@@ -7,15 +7,23 @@ repos:
...
         language: system
         entry: codeql query format --in-place
+
     -   id: sync-files
         name: Fix files required to be identical
         language: system
         entry: python3 config/sync-files.py --latest
         pass_filenames: false
+
+    -   id: qlhelp
```
MathiasVP, line 25: This should probably be named `qhelp` instead of `qlhelp`?

redsun82: yep!

redsun82:
@intrigus-lgtm I built upon your suggestion and integrated a qhelp check. I just hit a bit of a can of worms with how `.inc.qhelp` files should be handled :sweat_smile: which is the main reason `check-qhelp.py` looks a little bit more complicated than you would expect. Nothing too fancy, it mainly has to build an include map when `.inc.qhelp` files are involved and work its way from there, still quite fast to run.

MathiasVP (approved):
LGTM!

PR #8173, opened 2022-02-22
C++: Add another `CWE-825` query

MathiasVP:
This is a new IR-based query that can hopefully serve as a high-precision version of `cpp/stack-address-escape`. Instead of raising an alert on every assignment that writes the address of a local variable in a non-local variable, the query only raises an alert if we find a subsequent load of that global variable after the local variable's function has returned.

The query doesn't find a lot of results currently. But all the results that it _does_ find look like true positives to me.


cpp/ql/src/Likely Bugs/Memory Management/UsingExpiredStackAddress.cpp
```diff
@@ -0,0 +1,25 @@
...
+  int x = 0;
+  xptr = &x;
+  // ...
+  // use `xptr`
+  // ...
+  xptr = nullptr;
+}
+
+void example2() {
+  localAddressEscapes();
```
jketema, line 23: I think this should be:
```suggestion
  localAddressDoesNotEscape();
```
otherwise this example is identical to `example1`.

cpp/ql/src/Likely Bugs/Memory Management/UsingExpiredStackAddress.qhelp
```diff
@@ -0,0 +1,49 @@
+<!DOCTYPE qhelp PUBLIC
+  "-//Semmle//qhelp//EN"
+  "qhelp.dtd">
+<qhelp>
+<overview>
+<p>
+This rule finds uses of pointers that likely point to local variables in
+expired stack frames. Such pointers to local variables is only valid
+until the function returns, after which it becomes a dangling pointer.
```
jketema, line 9:
```suggestion
expired stack frames. A pointer to a local variable is only valid
until the function returns, after which it becomes a dangling pointer.
```

cpp/ql/test/query-tests/Likely Bugs/Memory Management/UsingExpiredStackAddress/test.cpp
```diff
@@ -0,0 +1,183 @@
...
+
+int test_escapes_through_reference() {
+  escape_through_reference();
+  return *s101.p; // BAD
+}
+
+struct S300 {
+  int a1[15];
+  int a2[14][15];
+  int a3[13][14][15];
```
jketema, line 117: These do not seem to be used below?

jketema (commented):
Review of the non-query part. I'll look at the query separately.

cpp/ql/src/Likely Bugs/Memory Management/UsingExpiredStackAddress.ql
```diff
@@ -0,0 +1,244 @@
...
+  not instr instanceof LoadInstruction and
+  result = globalValueNumber(instr.(CopyInstruction).getSourceValue())
+  or
+  exists(LoadInstruction load | instr = load |
+    result = TLoad(globalValueNumber(load.getSourceAddress()))
+  )
+  or
+  exists(ConvertInstruction convert, Type fromType, Type toType | instr = convert |
+    uncheckedConversionTypes(convert, fromType, toType) and
+    result = TConversion("unchecked", globalValueNumber(convert.getUnary()), fromType, toType)
```
jketema, line 118: Again maybe a beginner question: I know here I have a `ConvertInstruction` which is information that is then "converted" to a string `unchecked` in this case, and then in `TConversion` I need to reconstruct the `ConvertInstruction` using an `exists`. Couldn't you just pass the instruction as the first argument instead of a string? And, if, so why isn't that the taken approach?

cpp/ql/src/Likely Bugs/Memory Management/UsingExpiredStackAddress.ql
```diff
@@ -0,0 +1,244 @@
...
+  exists(IRBlock block1, IRBlock block2 |
+    store.getBlock() = block1 and
+    result.getBlock() = block2 and
+    block1.getASuccessor+() = block2
+  )
+}
+
+/**
+ * Holds if `store` copies the address of `f`'s local variable `var`
+ * into th address `globalAddress`.
```
jketema, line 156:
```suggestion
 * into the address `globalAddress`.
```

cpp/ql/src/Likely Bugs/Memory Management/UsingExpiredStackAddress.ql
```diff
@@ -0,0 +1,244 @@
...
+  IRBlock block, boolean isCallBlock, TGlobalAddress address, boolean isStoreBlock
+where
+  globalAddressPointsToStack(store, var, call, block, address, isCallBlock, isStoreBlock) and
+  block.getAnInstruction() = load and
+  globalValueNumber(load.getSourceAddress()) = address and
+  (
+    // We know that we have a sequence:
+    // (1) store to `address` -> (2) return from `f` -> (3) load from `address`.
+    // But if (2) and (3) happen in the sam block we need to check the
+    // block indices to nsure that (3) happens after (2).
```
jketema, line 222:
```suggestion
    // block indices to ensure that (3) happens after (2).
```

cpp/ql/src/Likely Bugs/Memory Management/UsingExpiredStackAddress.ql
```diff
@@ -0,0 +1,244 @@
...
+    if isCallBlock = true
+    then
+      // If so, the load must happen after the call.
+      exists(int callIndex, int loadIndex |
+        blockLoadsFromAddress(_, loadIndex, load, _) and
+        block.getInstruction(callIndex) = call and
+        callIndex < loadIndex
+      )
```
jketema, line 230: If I understand this correctly, this seems rather "coarse" in that it ignores any case in which there is a call after the load, even if the call is unrelated (similarly for the store case below)? I'm not suggestion this should be fixed here and now, if I'm correct 😃

jketema:
Someone with more knowledge of CodeQL than me probably also needs to have a look at this.

cpp/ql/src/Likely Bugs/Memory Management/UsingExpiredStackAddress.ql
```diff
@@ -0,0 +1,244 @@
...
+predicate stackPointerFlowsToUse(Instruction address, VariableAddressInstruction source) {
+  exists(VariableAddressInstruction var |
+    var = address and
+    var = source and
+    var.getASTVariable() instanceof StackVariable and
+    // Pointer-to-member types aren't properly handled in the dbscheme.
+    not var.getResultType() instanceof PointerToMemberType and
+    // Rule out FPs caused by extraction errors.
+    not any(ErrorExpr e).getEnclosingFunction() = var.getEnclosingFunction()
+  )
```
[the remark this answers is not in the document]
MathiasVP: Oh, excellent comments! I think the reason is that `source` at some point wasn't a `VariableAddressInstruction`, so I had to introduce an `exists` and use that one everywhere instead. I can totally rewrite it to what you're suggesting, yes 👍.

cpp/ql/src/Likely Bugs/Memory Management/UsingExpiredStackAddress.ql
MathiasVP: We cannot use the instruction itself because then these two occurrences of `(long)x`:

```cpp
static int x;
long test1() {
  return (long)x; // `convertInstruction1`
}
long test2() {
  return (long)x; // `convertInstruction2`
}
```

will have different `TGlobalAddress` representations (because one of them will have a `TConversion(convertInstruction1)` component and the other will have a `TConversion(convertInstruction2)` component. Does that make sense?

jketema: Yes, that makes perfect sense.

cpp/ql/src/Likely Bugs/Memory Management/UsingExpiredStackAddress.ql
MathiasVP: That's a good point. I think your conclusion is correct. I have an idea for how to fix this, but I think I'll delay that fix to a later PR.

cpp/ql/src/Likely Bugs/Memory Management/UsingExpiredStackAddress.cpp
MathiasVP: Oh, right. Thanks a lot!

cpp/ql/src/Likely Bugs/Memory Management/UsingExpiredStackAddress.qhelp
MathiasVP: Indeed, this reads much cleaner 👍.

MathiasVP:
> Review of the non-query part. I'll look at the query separately.

Thanks a lot for all the comments. I think my `e` key is starting to wear off >.<

cpp/ql/test/query-tests/Likely Bugs/Memory Management/UsingExpiredStackAddress/test.cpp
MathiasVP: Oops. I've added those tests in 53299d61eb2b7fd61a705da6b3703ec07c622842 now.

cpp/ql/src/Likely Bugs/Memory Management/UsingExpiredStackAddress.ql
```diff
@@ -0,0 +1,245 @@
+/**
+ * @name Use of expired stack-address.
```
geoffw0, line 2: We don't usually put punctuation in query `@name`s.
```suggestion
 * @name Use of expired stack-address
```

cpp/ql/test/query-tests/Likely Bugs/Memory Management/UsingExpiredStackAddress/test.cpp
```diff
@@ -0,0 +1,199 @@
...
+}
+
+void store_address_of_argument(int x) {
+  s101.p = &x;
+}
+
+int store_argument_address() {
+  int x;
+  store_address_of_argument(x);
+  return *s101.p; // BAD
```
geoffw0, line 58: It took me a moment to see what's going on here - we've stored the address of a parameter not of the local variable `x` in this function.  Perhaps the parameter should be called `int y` to improve readability of the test?

cpp/ql/test/query-tests/Likely Bugs/Memory Management/UsingExpiredStackAddress/test.cpp
```diff
@@ -0,0 +1,199 @@
...
+}
+
+void address_escapes_through_pointer_arith() {
+  int x = 0;
```
geoffw0, line 62: This ought really be an array, so that `p0 + 1` etc are valid addresses.

cpp/ql/test/query-tests/Likely Bugs/Memory Management/UsingExpiredStackAddress/test.cpp
```diff
@@ -0,0 +1,199 @@
...
+int test_pointer_arith_bad() {
+  address_escapes_through_pointer_arith();
+  return *s101.p; // BAD
+}
+
+int test_pointer_arith_good_1() {
+  int x;
+  address_escapes_through_pointer_arith();
+  s101.p = &x;
+  return *s101.p; // GOOD [FALSE POSITIVE]
```
geoffw0, line 80: I don't see a false positive result in the `.expected` for this line (perhaps you fixed it?)

cpp/ql/src/Likely Bugs/Memory Management/UsingExpiredStackAddress.ql
```diff
@@ -0,0 +1,245 @@
...
+ * @description Accessing the stack-allocated memory of a function
+ *              after it has returned can lead to memory corruption.
+ * @kind problem
+ * @problem.severity warning
```
geoffw0, line 6: I'd probably say this is an error?

geoffw0 (approved):
This looks promising, I have only nitpicks.

I appreciate thorough tests!

cpp/ql/test/query-tests/Likely Bugs/Memory Management/UsingExpiredStackAddress/test.cpp
MathiasVP: Oh, indeed I did, yes. Thanks.

cpp/ql/src/Likely Bugs/Memory Management/UsingExpiredStackAddress.ql
MathiasVP: Hm, yes. I think it should be an error, indeed.

MathiasVP: Fixed in 033edc24f46c2eca7753891564c8557ececd0d84.

cpp/ql/test/query-tests/Likely Bugs/Memory Management/UsingExpiredStackAddress/test.cpp
MathiasVP: Fixed in 033edc24f46c2eca7753891564c8557ececd0d84.

cpp/ql/test/query-tests/Likely Bugs/Memory Management/UsingExpiredStackAddress/test.cpp
MathiasVP: Good point. Fixed in 033edc24f46c2eca7753891564c8557ececd0d84.

cpp/ql/test/query-tests/Likely Bugs/Memory Management/UsingExpiredStackAddress/test.cpp
MathiasVP: I've renamed it in 033edc24f46c2eca7753891564c8557ececd0d84.

cpp/ql/src/Likely Bugs/Memory Management/UsingExpiredStackAddress.ql
```diff
@@ -0,0 +1,247 @@
...
+  not vai.getResultType() instanceof PointerToMemberType and
+  // Rule out FPs caused by extraction errors.
+  not any(ErrorExpr e).getEnclosingFunction() = f
+}
+
+/**
+ * Holds if `source` is the base address of an address computation whose
+ * result is stored in `address`.
+ */
+predicate stackPointerFlowsToUse(Instruction address, VariableAddressInstruction source) {
```
rdmarsh2, line 34: Could this just be `DataFlow::localFlow` or `TaintTracking::localTaint`?

MathiasVP: It could, but that would also give us flow through phi nodes. I've seen multiple FPs from results like:
```cpp
int* global1;
void f(bool b) {
  int local;
  int *p1, *p2;
  if(b) {
    p1 = &local;
  } else {
    p1 = p2;
  }
  // ...
  if(!b) {
    global = p1;
  }
}
```
where the address of the local doesn't actually escape due to complex path conditions.

MathiasVP:
I've been super unlucky with the last two DCA runs:
- The first one failed because there was a null-pointer issue in the CLI's structured logging. This has been fixed so I restarted DCA with a more recent internal repo SHA.
- So now I accidentally included the structured binding stuff from the internal repo without pulling in the changes to the dbscheme from https://github.com/github/codeql/pull/7928 🤦. So the DCA run failed because of a dbscheme mismatch.

The third time's the charm (I hope!) 🤞

jketema (approved):
LGTM, assuming third time is a charm when it comes to DCA runs.

cpp/ql/test/query-tests/Likely Bugs/Memory Management/UsingExpiredStackAddress/test.cpp
geoffw0: This fix is missing.

MathiasVP: Oops. _Actually_ fixed in e4af34253aa10b598778e3949b02e46d27ef62cb. Sorry about that!

geoffw0: No problem!

geoffw0 (approved):
I'm also happy with this PR now (DCA pending...)

hubwriter (approved):
Looks good from a docs PoV. 👍

MathiasVP:
DCA looks good 🎉.

PR #8174, opened 2022-02-22
C++: Simplify `cpp/declaration-hides-variable`

jketema:
The check for `(unnamed local variable)` is no longer needed, because these variables are now identified as being compiler generated.


MathiasVP (approved):
LGTM!

PR #8185, opened 2022-02-23
JS: recognize modules imported by AMD imports as library inputs

erik-krogh:
Recognizes the source for CVE-2020-7792 

[Evaluation was uneventful](https://github.com/github/codeql-dca-main/tree/data/erik-krogh/pr-8185-2032126__nightly__CustomSuite/reports).


asgerf (approved):
The change LGTM, though it does seem like that CVE ought to be flagged by `js/prototype-pollution-utility` which doesn't currently rely on this notion of library inputs.

PR #8200, opened 2022-02-23
Python: Add `debug partial flow` snippet

RasmusWL:
I've been writing that out by hand quite a few times, so wanted to save it as a snippet :blush:


yoff (approved):
Sure, why not :-) It does assume a specific setup, where `Configuration` is defined, but I think it codifies our current development practices well enough.

I usually rewrite the query (or use [this](https://github.com/github/codeql/blob/main/python/ql/test/experimental/meta/debug/dataflowTestPaths.ql)), but the snippet does look easier than rewriting :-)

PR #8209, opened 2022-02-23
C++: Add IR structured binding tests

jketema:
Tests for the three structured binding cases (arrays, data members, and tuple-like classes).


MathiasVP (approved):
LGTM!

PR #8221, opened 2022-02-23
JS: recognize more module exports from the factory pattern

erik-krogh:
Recognizes a source for CVE-2021-23574

### DIFF
--- a/javascript/ql/lib/semmle/javascript/PackageExports.qll
+++ b/javascript/ql/lib/semmle/javascript/PackageExports.qll
@@ -120,14 +120,21 @@ private DataFlow::Node getAValueExportedByPackage() {
   exists(ImmediatelyInvokedFunctionExpr func, DataFlow::ParameterNode prev, int i |
     prev.getName() = "factory" and
     func.getParameter(i) = prev.getParameter() and
-    result = func.getInvocation().getArgument(i).flow().getAFunctionValue().getAReturn() and
-    DataFlow::globalVarRef("define").getACall().getArgument(1) = prev.getALocalUse() and
+    DataFlow::globalVarRef("define").getACall().getArgument(any(int a | a >= 1)) =
+      prev.getALocalUse() and
     func.getFile() =
       min(int j, File f |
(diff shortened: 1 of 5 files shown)


javascript/ql/lib/semmle/javascript/PackageExports.qll
```diff
@@ -120,14 +120,21 @@ private DataFlow::Node getAValueExportedByPackage() {
   exists(ImmediatelyInvokedFunctionExpr func, DataFlow::ParameterNode prev, int i |
     prev.getName() = "factory" and
     func.getParameter(i) = prev.getParameter() and
-    result = func.getInvocation().getArgument(i).flow().getAFunctionValue().getAReturn() and
-    DataFlow::globalVarRef("define").getACall().getArgument(1) = prev.getALocalUse() and
+    DataFlow::globalVarRef("define").getACall().getArgument(any(int a | a >= 1)) =
```
kaeluka, line 123: can `int a` ever be larger than 2? and couldn't it also be `0`? like in 1.3.2 here: https://requirejs.org/docs/api.html#define Perhaps, this should simply be `getAnArgument`

javascript/ql/lib/semmle/javascript/PackageExports.qll
```diff
@@ -120,14 +120,21 @@ private DataFlow::Node getAValueExportedByPackage() {
   exists(ImmediatelyInvokedFunctionExpr func, DataFlow::ParameterNode prev, int i |
     prev.getName() = "factory" and
```
kaeluka, line 121: as you're naming the parameternode below `exports` after its name, could you also rename `prev` to `factory`?

kaeluka (commented):
overall, lgtm!

javascript/ql/lib/semmle/javascript/PackageExports.qll
erik-krogh: You seem to be right, it can be any argument.

kaeluka (approved):
LGTM, thanks!

PR #8225, opened 2022-02-24
C++: Update the IR translation for structured bindings

jketema:
Update the IR translation for structured bindings. Observe that in the before situation the IR is only incorrect in the array and data member cases, and not in the tuple case.

<strike>What I realised is missing wrt to tests, is test that involve rvalue references. I'm not sure if it's worth adding those. I will check locally whether these cause any problems.</strike>


cpp/ql/lib/semmle/code/cpp/ir/internal/IRUtilities.qll
```diff
@@ -30,7 +39,10 @@ Type getVariableType(Variable v) {
         result = v.getInitializer().getExpr().getType()
         or
         not exists(v.getInitializer()) and result = v.getType()
-      else result = v.getType()
+      else
+        if isNonReferenceStructuredBinding(v)
+        then exists(ReferenceType r | r.getBaseType() = v.getUnderlyingType() | result = r)
```
jketema, line 44: I'm not sure whether this `exists` will always work. Also I'm not sure whether to use `v.getUnderlyingType()` or `v.getUnspecifiedType()`. I did already see I cannot use `v.getType()` as that doesn't work in the case of `typedefs`.

cpp/ql/test/library-tests/ir/ir/raw_consistency.expected
```diff
@@ -17,6 +17,7 @@ containsLoopOfForwardEdges
 lostReachability
 backEdgeCountMismatch
 useNotDominatedByDefinition
+| ir.cpp:1486:8:1486:8 | Unary | Operand 'Unary' is not dominated by its definition in function '$@'. | ir.cpp:1486:8:1486:8 | void StructuredBindingDataMemberStruct::StructuredBindingDataMemberStruct() | void StructuredBindingDataMemberStruct::StructuredBindingDataMemberStruct() |
```
jketema, line 20: This seems to happen irrespective of my changes and is related to the `StructuredBindingDataMemberMemberStruct m` member I introduced in the tests.

cpp/ql/lib/semmle/code/cpp/ir/internal/IRUtilities.qll
jketema: This question is still relevant, although the code here has changed slightly.

MathiasVP (commented):
Awesome work! Here are my first superficial review comments. The consistency failure is definitely something we need to fix before we can merge this, I think. I'll let @dbartol or @rdmarsh2 do the deep reviewing on it.

cpp/ql/test/library-tests/ir/ir/raw_ir.expected
```diff
@@ -7985,401 +7985,431 @@ ir.cpp:
...
+# 1486|     v1486_1(void)                                      = EnterFunction                : 
+# 1486|     mu1486_2(unknown)                                  = AliasedDefinition            : 
+# 1486|     mu1486_3(unknown)                                  = InitializeNonLocal           : 
+# 1486|     r1486_4(glval<unknown>)                            = VariableAddress[#this]       : 
+# 1486|     mu1486_5(glval<StructuredBindingDataMemberStruct>) = InitializeParameter[#this]   : &:r1486_4
+# 1486|     r1486_6(glval<StructuredBindingDataMemberStruct>)  = Load[#this]                  : &:r1486_4, ~m?
+# 1486|     mu1486_7(StructuredBindingDataMemberStruct)        = InitializeIndirection[#this] : &:r1486_6
+#-----|   Goto -> Block 2
+
+# 1486|   Block 1
```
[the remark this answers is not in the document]
jketema: Yeah, see my second comment above.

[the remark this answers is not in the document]
MathiasVP: Ah, right. I didn't see your comment about this not being caused by your PR. All good, then!

dbartol (commented):
Initial comments. Still looking at IR diffs.

cpp/ql/lib/semmle/code/cpp/ir/internal/IRUtilities.qll
```diff
@@ -30,7 +39,10 @@ Type getVariableType(Variable v) {
         result = v.getInitializer().getExpr().getType()
         or
         not exists(v.getInitializer()) and result = v.getType()
-      else result = v.getType()
+      else
+        if isNonReferenceStructuredBinding(v)
+        then exists(LValueReferenceType r | r.getBaseType() = v.getUnderlyingType() | result = r)
```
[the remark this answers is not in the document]
jketema: I was actually wondering about that. For every case I could come up with `T&` was there, but that might just have been by accident.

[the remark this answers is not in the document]
jketema: So it's not guaranteed by the extractor. Consider the following in isolation:
```c++
struct S {
    int x;
} s;

void f() {
    auto [x] = s;
}
```
then
```codeql
import cpp

from LValueReferenceType r
select r
```
yields:
```
| file://:0:0:0:0 | S & |
| file://:0:0:0:0 | __va_list_tag & |
| file://:0:0:0:0 | const S & |
| file://:0:0:0:0 | const __va_list_tag & |
```
and the IR becomes:
```
#    5| void f()
#    5|   Block 0
#    5|     v5_1(void)     = EnterFunction                             : 
#    5|     mu5_2(unknown) = AliasedDefinition                         : 
#    5|     mu5_3(unknown) = InitializeNonLocal                        : 
#    6|     r6_1(glval<S>) = VariableAddress[(unnamed local variable)] : 
#    6|     r6_2(glval<S>) = VariableAddress[s]                        : 
#    6|     r6_3(S)        = Load[s]                                   : &:r6_2, ~m?
#    6|     mu6_4(S)       = Store[(unnamed local variable)]           : &:r6_1, r6_3

#    6|   Block 1
#    6|     r6_5(glval<S>)   = VariableAddress[(unnamed local variable)] : 
#    6|     r6_6(glval<int>) = FieldAddress[x]                           : r6_5

#    7|   Block 2
#    7|     v7_1(void) = NoOp         : 
#    5|     v5_4(void) = ReturnVoid   : 
#    5|     v5_5(void) = AliasedUse   : ~m?
#    5|     v5_6(void) = ExitFunction : 
```
So I'll need to come up with a different solution here.

dbartol:
The IR for the structured binding initializations and access look correct. Figure out whether you want to make some of the changes I proposed in my review, and I'll approve.

cpp/ql/lib/semmle/code/cpp/ir/internal/IRUtilities.qll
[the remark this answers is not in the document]
jketema: I could generate the relevant type in the extractor, but I'm not sure whether that's the correct approach.

cpp/ql/lib/semmle/code/cpp/ir/internal/IRUtilities.qll
jketema: Changed to `getUnspecifiedType`.

cpp/ql/lib/semmle/code/cpp/ir/internal/IRUtilities.qll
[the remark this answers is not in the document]
jketema: Opened a PR that ensures the required type is output by the extractor.

cpp/ql/lib/change-notes/2022-02-24-structured-bindings-in-ir.md
```diff
@@ -0,0 +1,4 @@
+---
+category: minorAnalysis
+---
+* Structured bindings are handled in the IR translation.
```
intrigus-lgtm, line 4: What does this mean for a user? Or are these change note not intended for users?

jketema: The IR is visible to the user. Does that answer your question.

MathiasVP: Maybe we should also write something along the lines of: `As a result, many queries now support structured bindings.`?

jketema: Done in a slightly different form, so please check.

MathiasVP: Looks good. Thanks!

cpp/ql/lib/semmle/code/cpp/ir/internal/IRUtilities.qll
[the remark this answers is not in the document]
dbartol: Yeah, I think generating the type in the extractor is the best way to do this, although you should add a comment here explaining why we can rely on that type being present.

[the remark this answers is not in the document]
jketema: Added a comment. This now depends on the internal PR for the extractor to be merged.

MathiasVP (approved):
LGTM!

PR #8229, opened 2022-02-24
JS: step through parentheses in barrier functions

erik-krogh:
Gets a TN for CVE-2020-28503 

[Evaluation was uneventful](https://github.com/github/codeql-dca-main/tree/data/erik-krogh/parenSan__nightly__security-extend__1/reports). (Ignore the backref).

### DIFF
--- a/javascript/ql/lib/semmle/javascript/dataflow/Configuration.qll
+++ b/javascript/ql/lib/semmle/javascript/dataflow/Configuration.qll
@@ -2043,11 +2043,9 @@ private class BarrierGuardFunction extends Function {
         // ad hoc support for conjunctions:
         getALogicalAndParent(guard) = returnExpr and guardOutcome = true
         or
-        // getALogicalAndOperand+(returnExpr) = guard.asExpr() and guardOutcome = true
         // ad hoc support for disjunctions:
         getALogicalOrParent(guard) = returnExpr and guardOutcome = false
       |
-        // getALogicalOrOperand+(returnExpr) = guard.asExpr() and guardOutcome = false
         exists(SsaExplicitDefinition ssa |
           ssa.getDef().getSource() = returnExpr and
           ssa.getVariable().getAUse() = getAReturnedExpr()
--- a/javascript/ql/lib/semmle/javascript/dataflow/Configuration.qll
+++ b/javascript/ql/lib/semmle/javascript/dataflow/Configuration.qll
@@ -1992,20 +1992,26 @@ module PathGraph {
 }
 
 /**
- * Gets an operand of the given `&&` operator.
- *
- * We use this to construct the transitive closure over a relation
- * that does not include all of `BinaryExpr.getAnOperand`.
+ * Gets a logical `and` expression, or parenthesized expression, that contains `guard`.
  */
-private Expr getALogicalAndOperand(LogAndExpr e) { result = e.getAnOperand() }
+private Expr getALogicalAndParent(BarrierGuardNode guard) {
+  barrierGuardIsRelevant(guard) and result = guard.asExpr()
+  or
+  result.(LogAndExpr).getAnOperand() = getALogicalAndParent(guard)
+  or
+  result.getUnderlyingValue() = getALogicalAndParent(guard)
+}
 
 /**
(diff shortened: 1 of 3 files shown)


javascript/ql/lib/semmle/javascript/dataflow/Configuration.qll
```diff
@@ -2035,11 +2041,13 @@ private class BarrierGuardFunction extends Function {
         returnExpr = guard.asExpr()
         or
         // ad hoc support for conjunctions:
-        getALogicalAndOperand+(returnExpr) = guard.asExpr() and guardOutcome = true
+        getALogicalAndParent(guard) = returnExpr and guardOutcome = true
         or
+        // getALogicalAndOperand+(returnExpr) = guard.asExpr() and guardOutcome = true
```
kaeluka, line 2046: did you mean to remove this line?

javascript/ql/lib/semmle/javascript/dataflow/Configuration.qll
```diff
@@ -2035,11 +2041,13 @@ private class BarrierGuardFunction extends Function {
...
         // ad hoc support for disjunctions:
-        getALogicalOrOperand+(returnExpr) = guard.asExpr() and guardOutcome = false
+        getALogicalOrParent(guard) = returnExpr and guardOutcome = false
       |
+        // getALogicalOrOperand+(returnExpr) = guard.asExpr() and guardOutcome = false
```
kaeluka, line 2050: did you mean to remove this line?

javascript/ql/lib/semmle/javascript/dataflow/Configuration.qll
```diff
@@ -1992,20 +1992,26 @@ module PathGraph {
...
- * that does not include all of `BinaryExpr.getAnOperand`.
+ * Gets a logical `or` expression, or parenthesized expression, that contains `guard`.
  */
-private Expr getALogicalOrOperand(LogOrExpr e) { result = e.getAnOperand() }
+private Expr getALogicalOrParent(BarrierGuardNode guard) {
+  barrierGuardIsRelevant(guard) and result = guard.asExpr()
+  or
+  result.(LogOrExpr).getAnOperand() = getALogicalOrParent(guard)
+  or
+  result.getUnderlyingValue() = getALogicalOrParent(guard)
```
kaeluka, line 2013: can you confirm: this line is the reason you made this new predicate, rather than continuing to use `getALogicalOrOperand+`? It makes it so that even barrier guard nodes that are wrapped in (parens) are handled properly?

kaeluka:
> Gets a TN for https://github.com/advisories/GHSA-897m-rjf5-jp39

Which true negative is that?

erik-krogh:
> > Gets a TN for [GHSA-897m-rjf5-jp39](https://github.com/advisories/GHSA-897m-rjf5-jp39)
> 
> Which true negative is that?

It's the `js/prototype-polluting-assignment` query.   
We already got a TP on the vulnerable version, but we also got an FP on the fixed version.  
With this change we now get a TN on the fix.

javascript/ql/lib/semmle/javascript/dataflow/Configuration.qll
erik-krogh: Yes.   

I could also have written a simpler step predicate like: 
```CodeQL
Expr getLogicalAndThing(Expr prev) {
  result = prev.(LogAndExpr).getAnOperand()
  or
  result = prev.getUnderlyingValue()
}
```

And then used that like `getLogicalAndThing+(returnExpr) = guard`.   
Which would be more in line with the previous implementation.  

--- 

However, the approach I have chosen gives better performance, because I'm limiting the base-case, which means that the transitive closure gets way smaller.  
(Using the `+`/`'*` operators is generally faster, but it sometimes computes some very large intermediate results if it doesn't limit the base case).

kaeluka: Thanks for the thorough explanation <3

kaeluka (approved):
LGTM! Thanks!

PR #8230, opened 2022-02-24
C#: Buildless extractor option.

michaelnebel:
In this PR we introduce a new C# extractor option, which indicates, whether a tracing or buildless (standalone) extraction should be performed, when using the autobuilder.

Closes https://github.com/github/codeql-csharp-team/issues/205

```bash
codeql database create --language=csharp -Obuildless=true db
codeql database create --language=csharp --extractor-option=buildless=true db
```

### DIFF
--- a/csharp/codeql-extractor.yml
+++ b/csharp/codeql-extractor.yml
@@ -40,3 +40,4 @@ options:
       and should only be used in cases where it is not possible to build
       the code (for example if it uses inaccessible dependencies).
     type: string
+    pattern: "^(false|true)$"
--- a/csharp/codeql-extractor.yml
+++ b/csharp/codeql-extractor.yml
@@ -33,7 +33,10 @@ options:
     title: Whether to use buildless (standalone) extraction.
     description: >
       A value indicating, which type of extraction the autobuilder should perform.
-      If 'yes', then the standalone extractor will be used, otherwise tracing extraction
+      If 'true', then the standalone extractor will be used, otherwise tracing extraction
       will be performed.
-      The default is 'no'.
+      The default is 'false'.
+      Note that buildless extraction will generally yield less accurate analysis results,
+      and should only be used in cases where it is not possible to build
+      the code (for example if it uses inaccessible dependencies).
     type: string
--- a/csharp/codeql-extractor.yml
+++ b/csharp/codeql-extractor.yml
@@ -29,3 +29,11 @@ options:
           (to write uncompressed TRAP).
         type: string
         pattern: "^(none|gzip|brotli)$"
+  buildless:
+    title: Whether to use buildless (standalone) extraction.
+    description: >
+      A value indicating, which type of extraction the autobuilder should perform.
+      If 'yes', then the standalone extractor will be used, otherwise tracing extraction
+      will be performed.
+      The default is 'no'.
+    type: string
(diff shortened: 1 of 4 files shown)


tamasvajk (dismissed):

csharp/codeql-extractor.yml
```diff
@@ -16,3 +16,12 @@ file_types:
     extensions:
       - .cs
 legacy_qltest_extraction: true
+options:
+  buildless:
+    title: Whether to use buildless (standalone) extraction.
+    description: >
+      A value indicating, which type of extraction the autobuilder should perform.
+      If 'yes', then the standalone extractor will be used, otherwise tracing extraction
```
hvitved, line 24: I think we should use `true` instead of `yes` (and `false` below instead of `no`).

csharp/codeql-extractor.yml
```diff
@@ -16,3 +16,12 @@ file_types:
...
+      will be performed.
+      The default is 'no'.
```
hvitved, line 26: Add:
```
Note that buildless extraction will generally yield less accurate analysis results, and should only be
used in cases where it is not possible to build the code (for example if it uses inaccessible dependencies).
```

csharp/ql/src/change-notes/2022-02-24-buildless-extractor-option.md
```diff
@@ -0,0 +1,4 @@
+---
+category: minorAnalysis
+---
+* The C# extractor now accepts an extractor option `buildless`, which is used to decide what type of extraction that should be performed. If `yes` then buildless (standalone) extraction will be performed. Otherwise tracing extraction will be performed (default).
```
hvitved, line 4: yes -> true

hvitved (changes requested):
LGTM, some small remarks.

michaelnebel:
> LGTM, some small remarks.

Sure, I will make the changes in the description.
The "parser" of the environment variable will still accept `yes` and `no` (like for other `boolean` environment variables). So this is only a change to the documentation.

csharp/codeql-extractor.yml
```diff
@@ -16,3 +16,15 @@ file_types:
...
+    title: Whether to use buildless (standalone) extraction.
+    description: >
+      A value indicating, which type of extraction the autobuilder should perform.
+      If 'true', then the standalone extractor will be used, otherwise tracing extraction
+      will be performed.
+      The default is 'false'.
+      Note that buildless extraction will generally yield less accurate analysis results,
+      and should only be used in cases where it is not possible to build
+      the code (for example if it uses inaccessible dependencies).
+    type: string
```
hvitved, line 30: I think we should add `pattern: "^(false|true)$"`

michaelnebel: Yes, that is a good idea - will add that.

hvitved (dismissed):

csharp/ql/src/change-notes/2022-02-24-buildless-extractor-option.md
```diff
@@ -0,0 +1,4 @@
+---
+category: minorAnalysis
+---
+* The C# extractor now accepts an extractor option `buildless`, which is used to decide what type of extraction that should be performed. If `true` then buildless (standalone) extraction will be performed. Otherwise tracing extraction will be performed (default).
```
hvitved, line 4: Perhaps add: ``The option is added via `codeql database create --language=csharp -Obuildless=true ...` ``

hvitved (dismissed):

michaelnebel:
@hvitved Requested review again, as a rebase was needed. Otherwise no changes.

hvitved (approved):

PR #8231, opened 2022-02-24
QL: fix ql-for-ql errors inside ql-for-ql

erik-krogh:
I downloaded [the latest .sarif files from the workflow](https://github.com/github/codeql/actions/workflows/ql-for-ql-build.yml?query=branch%3Amain), and fixed everything.


tausbn (approved):
Nice!

PR #8232, opened 2022-02-24
Add new xss queries to result counting query

TomBolton:
Add the new `Xss` queries to the `CountAlertsAndEndpoints.ql` query.


henrymercer (approved):
Looks OK, but keep in mind the performance implications of this change — as per the deprecation comment this query does not scale well with the number of security queries we extract alerts for.

TomBolton:
Thank you @henrymercer and yes, these two "cheap" `Xss` queries will be the final modification to the `CountAlertsAndEndpoints.ql` query before it gets replaced.

Rawannkm (commented):
Screenshot_٢٠٢٢٠٢٢٢-٠٩٥٩٣٢_ABSHER~2

PR #8234, opened 2022-02-24
Ruby: add meta-queries for measuring sources, sinks, and call edges

asgerf:
Adds three meta-queries for measuring the number of taint sources, sinks, and call edges.

The idea is to include these in a DCA evaluation (as meta-queries) so it's possible to see the impact of your changes at a more fine-grained level. Often, a change generates no new alerts, but produces sources or sinks that give an indication of its impact. The idea was initially discussed in [this issue](https://github.com/github/codeql-javascript-team/issues/191). As for JS, I didn't include logging-related sinks as they tend to dominate too much compared to their security-relevance.

As an example, [this diff](https://github.com/github/codeql-dca-main/blob/data/asgerf/meta-queries__Differences__smoke-test/reports/alert-meta-comparison.md) is comparing `main` to some earlier version from about a week ago, showing plenty of new sinks.

### DIFF
--- a/ruby/ql/src/queries/meta/internal/TaintMetrics.qll
+++ b/ruby/ql/src/queries/meta/internal/TaintMetrics.qll
@@ -1,0 +1,38 @@
+private import codeql.files.FileSystem
+private import codeql.ruby.DataFlow
+private import codeql.ruby.dataflow.RemoteFlowSources
+private import codeql.ruby.security.CodeInjectionCustomizations
+private import codeql.ruby.security.CommandInjectionCustomizations
+private import codeql.ruby.security.XSS
+private import codeql.ruby.security.PathInjectionCustomizations
+private import codeql.ruby.security.ServerSideRequestForgeryCustomizations
+private import codeql.ruby.security.UnsafeDeserializationCustomizations
+private import codeql.ruby.security.UrlRedirectCustomizations
+
+class RelevantFile extends File {
+  RelevantFile() { not getRelativePath().regexpMatch(".*/test(case)?s?/.*") }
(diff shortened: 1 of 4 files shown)


ruby/ql/src/queries/meta/internal/TaintMetrics.qll
```diff
@@ -0,0 +1,38 @@
...
+
+RemoteFlowSource relevantTaintSource(string kind) {
+  result.getLocation().getFile() instanceof RelevantFile and
+  kind = result.getSourceType()
+}
+
+DataFlow::Node relevantTaintSink(string kind) {
+  result.getLocation().getFile() instanceof RelevantFile and
+  (
+    kind = "CodeInjection" and result instanceof CodeInjection::Sink
```
nickrolfe, line 24: Any reason not to include `CleartextLogging::Sink`?

nickrolfe (commented):
Thanks for doing this.

I have one question, but otherwise it looks good.

ruby/ql/src/queries/meta/internal/TaintMetrics.qll
asgerf: From the PR description,
> As for JS, I didn't include logging-related sinks as they tend to dominate too much compared to their security-relevance.

If you want to measure the number of logging calls, I'd probably put them in a separate meta-query.

nickrolfe: Sorry for not reading more closely.

nickrolfe (approved):

Issue #5595, opened 2021-04-03
Java: `RefType.getAnAncestor()` is error-prone


Marcono1234:
The predicate [`RefType.getAnAcestor()`](https://codeql.github.com/codeql-standard-libraries/java/semmle/code/java/Type.qll/predicate.Type$RefType$getAnAncestor.0.html) is error-prone when one of the supertypes is a generic type parameterized with `Object` as type argument. In that case `getASupertype()` and `getAnAcestor()` return all types which have a lower bound (but are not actually relevant), e.g.:
```ql
import java

from RefType t
where t.hasName("ObjectToStringComparator")
select t, t.getAnAncestor()
```
[Query Console link](https://lgtm.com/query/7884599397805363842/)

<details>
<summary>Result hierarchy (click to expand)</summary>

- `ObjectToStringComparator`
  - `Serializable`
  - `Object`
  - `Comparator<Object>`
    - `Comparator<>`
    - `Comparator<? super Boolean>`
    - `Comparator<? super Callable<Integer>>`
    - `Comparator<? super CharSequence>`
    - `Comparator<? super Class<?>>`
    - `Comparator<? super Diff<?>>`
    - `Comparator<? super Double>`
    - `Comparator<? super E>`
    - `Comparator<? super E>`
    - `Comparator<? super E>`
    - `Comparator<? super E>`
    - `Comparator<? super E>`
    - `Comparator<? super E>`
    - `Comparator<? super E>`
    - `Comparator<? super E>`
    - `Comparator<? super E>`
    - `Comparator<? super E>`
    - `Comparator<? super E>`
    - `Comparator<? super E>`
    - `Comparator<? super Field>`
    - `Comparator<? super Format>`
    - `Comparator<? super Hobby>`
    - `Comparator<? super ImmutablePair<Integer,Integer>>`
    - `Comparator<? super ImmutableTriple<Integer,Integer,Integer>>`
    - `Comparator<? super Integer>`
    - `Comparator<? super K>`
    - `Comparator<? super K>`
    - `Comparator<? super K>`
    - `Comparator<? super K>`
    - `Comparator<? super L>`
    - `Comparator<? super Locale>`
    - `Comparator<? super Long>`
    - `Comparator<? super Method>`
    - `Comparator<? super MethodDescriptor>`
    - `Comparator<? super Method[]>`
    - `Comparator<? super O>`
    - `Comparator<? super O>`
    - `Comparator<? super O>`
    - `Comparator<? super O>`
[...]

Issue #5521, opened 2021-03-24
Java: RawType getASourceSupertype() is not generic type / getting source declaration or supertype is cumbersome


Marcono1234:
It appears that for a `RawType` the result of `getASourceSupertype()` will not be the generic type. (This is also described by the `getASourceSupertype()` [documentation](https://codeql.github.com/codeql-standard-libraries/java/semmle/code/java/Type.qll/predicate.Type$RefType$getASourceSupertype.0.html).)

In addition to that there are the following issues:
- CodeQL raw types have a `<>` at the end of their name, therefore a check for the class name such as `getDeclaringType().getASourceSupertype*().hasQualifiedName("java.util", "List")` would not hold due to the trailing `<>`.
- Access of static fields and methods declared on a generic class are treated like accesses on a raw type (`fieldAccess.getQualifier().getType()` and `methodAccess.getDeclaringType()` will have a raw type as result).
Whether that is the correct behavior might be a different story.

This all combined causes some false negatives for the pattern `getDeclaringType().getASourceSupertype*()` (which is also used a few times in the CodeQL codebase).
Is there a more 'correct' predicate for this task? A workaround might be `getDeclaringType().getASourceSupertype*().getErasure()`.
If there is no alternative currently, would it make sense to change the behavior of `getASourceSupertype()` or introduce a new predicate?

aschackmull:
These are good and difficult questions. The `getASourceSupertype` predicate was itself introduced to work around some similar problems and has mostly been doing its job.  It would be nice to rework some of these things, but it's not clear what the right solutions are and how to get there without a lot of potentially breaking changes.

PR #8241, opened 2022-02-24
Java: Update stats and make some performance tweaks

igfoo:


java/ql/lib/change-notes/2022-02-21-type-hierarchy.md
```diff
@@ -0,0 +1,6 @@
+---
+category: feature
+---
+* Added `hasDescendant(RefType anc, Type sub)`
+* Added `RefType.getADescendant()`
+* Added `RefType.getAStrictAncestor()`
```
smowton, line 6:
```suggestion
* Added `RefType.getAnAncestor()`
* Added `RefType.getAStrictAncestor()`
```

smowton (commented):
LGTM if DCA results are good

java/ql/lib/change-notes/2022-02-21-type-hierarchy.md
igfoo: `RefType.getAnAncestor()` already exists; this just changes its definition.

Rawannkm:
Thanks for the first time since we have 😘😘🤑😎💯

yo-h (approved):

java/ql/lib/semmle/code/java/Type.qll
```diff
@@ -37,6 +37,16 @@ predicate hasSubtype(RefType t, Type sub) {
...
 }
 
+/**
+ * Holds if reference type `anc` is a direct or indirect supertype of `sub`, including itself.
+ */
+cached
+predicate hasDescendant(RefType anc, Type sub) {
+  anc = sub
+  or
+  exists(RefType mid | hasSubtype(anc, mid) and hasDescendant(mid, sub))
```
Marcono1234, line 47: Usage of `hasSubtype` has its issues with generic and parameterized types, most notably:
- `getAnAcestor()` never gets for a parameterized type (e.g. `java.util.List<String>`) the generic source declaration (e.g. `java.util.List`)
- `hasSubtype` seems to be unable to return subtypes for generic types, e.g. for `java.util.List` it does not get `java.util.ArrayList` ([example query](https://lgtm.com/query/5617555852827933522/)); maybe that is actually a bug? (can report that separately if you want)

By adding more predicates which suffer from these issues, it will become more likely that users will use these predicates by accident for generic or parameterized types and run into false negatives or not working queries. Though maybe not considering generic and parameterized types here is acceptable / desired for performance?
I just wanted to point that out, in case you were not aware of it.

Related to this might be #5521 and #5595.

igfoo: I take your point about people being more likely to use these predicates if we make it easier for them, but I don't think the new predicates are much easier to use than e.g. `hasSubtype*`. So as this PR doesn't make the behaviour worse, I think I'll go ahead and merge it.

The `List`/`ArrayList` example does look suspicious at first glance, though. Could you go ahead and file a ticket for that please, so we can take a proper look into it.

Marcono1234: Have created #8339 for this.

aschackmull:
The caching of the transitive closure of `hasSubtype` provides an interesting cache-size/analysis-time trade-off. I think it's mostly an optimiser short-coming that we have to make this trade-off at all - better code-generation ought to be able to handle the transitive closures without caching the entire thing.
As for the remaining performance tweaks, I think the QL changes look somewhat dubious, so I'd like to follow up on those.
(It would btw. have been nice if these two parts had been separated into two commits)

PR #8245, opened 2022-02-25
CPP: Add query for CWE-476: NULL Pointer Dereference when using exception handling blocks

ihsinme:
this query looks for access situations for pointer data types in situations where they might not have been allocated. the main situation is the transition to the exception handler, before the allocation of memory. an additional query looks for memory re-freeing situations during exception handling.

CVE-2020-15304

I am currently working on making a change to the actual project.


cpp/ql/src/experimental/Security/CWE/CWE-476/DangerousUseOfExceptionBlocks.ql
```diff
@@ -0,0 +1,164 @@
+/**
+ * @name Dangerous use of exception blocks.
+ * @description When clearing the data in the catch block, you must be sure that the memory was allocated before the exception.
+ * @kind problem
+ * @id cpp/dangerous-use-of-exception-blocks
+ * @problem.severity warning
+ * @precision medium
+ * @tags correctness
+ *       security
+ *       external/cwe/cwe-476
```
geoffw0, line 10: And perhaps CWE-415 (Double Free)?

cpp/ql/test/experimental/query-tests/Security/CWE/CWE-476/semmle/tests/test.cpp
```diff
@@ -0,0 +1,252 @@
...
+	allocData(valData);
+	cleanFunction();
+	 
+  } 
+  catch (...) 
+  {
+    for (size_t i = 0; i < valData->sizeInt; i++)
+    {
+	if (valData->bufMyData[i]) 
+       	    delete[] valData->bufMyData[i]->buffer; // GOOD
```
geoffw0, line 92: `valData->bufMyData[i]` exists due to line 78, but (I think) may contain uninitialized (non-NULL) data if we didn't reach line 83.  I would say this code is still not safe.

cpp/ql/test/experimental/query-tests/Security/CWE/CWE-476/semmle/tests/test.cpp
```diff
@@ -0,0 +1,252 @@
...
+    }
+    delete [] valData->bufMyData;
+    delete valData;
+  }
+}
+
+
+void funcWork4() {
+  int a;
+  myGlobalData *valData;
```
geoffw0, line 154:
```suggestion
  myGlobalData *valData = 0;
```
because the `new` on line 156 could fail, taking us straight to the `catch` block.

cpp/ql/test/experimental/query-tests/Security/CWE/CWE-476/semmle/tests/test.cpp
```diff
@@ -0,0 +1,252 @@
...
+  } 
+  catch (...) 
+  {
+    delete valData; // GOOD
+  }
+}
+
+void funcWork4b() {
+  int a;
+  myGlobalData *valData;
```
geoffw0, line 170:
```suggestion
  myGlobalData *valData = 0;
```

cpp/ql/test/experimental/query-tests/Security/CWE/CWE-476/semmle/tests/test.cpp
```diff
@@ -0,0 +1,252 @@
...
+	throwFunction(a);	 
+  } 
+  catch (...) 
+  {
+    delete valData; // BAD
+  }
+}
+void funcWork5() {
+  int a;
+  myGlobalData *valData;
```
geoffw0, line 184:
```suggestion
  myGlobalData *valData = 0;
```

cpp/ql/test/experimental/query-tests/Security/CWE/CWE-476/semmle/tests/test.cpp
```diff
@@ -0,0 +1,252 @@
...
+  }
+  catch (...) 
+  {
+        delete valData; // GOOD
+  }
+}
+
+void funcWork5b() {
+  int a;
+  myGlobalData *valData;
```
geoffw0, line 205:
```suggestion
  myGlobalData *valData = 0;
```

cpp/ql/test/experimental/query-tests/Security/CWE/CWE-476/semmle/tests/test.cpp
```diff
@@ -0,0 +1,252 @@
...
+  }
+  catch (...) 
+  {
+        delete valData; // BAD
+  }
+}
+void funcWork6() {
+  int a;
+  int flagB = 0;
+  myGlobalData *valData;
```
geoffw0, line 223:
```suggestion
  myGlobalData *valData = 0;
```

cpp/ql/src/experimental/Security/CWE/CWE-476/DangerousUseOfExceptionBlocks.cpp
```diff
@@ -0,0 +1,48 @@
...
+...
+  catch (const exception &) {
+	  delete valData;
+    valData = NULL;
+	  throw;
+  }
+  catch (...) 
+  {
+    delete valData; // GOOD  
+...
```
geoffw0, line 48: I struggled to follow these examples.  I feel like they (and most of the tests) could probably contain fewer levels of pointer indirection and still get the same meaning across.

cpp/ql/src/experimental/Security/CWE/CWE-476/DangerousUseOfExceptionBlocks.qhelp
```diff
@@ -0,0 +1,23 @@
+<!DOCTYPE qhelp PUBLIC
+  "-//Semmle//qhelp//EN"
+  "qhelp.dtd">
+<qhelp>
+<overview>
+<p>When clearing the data in the catch block, you must be sure that the memory was allocated before the exception.</p>
```
geoffw0, line 6:
```suggestion
<p>When releasing memory in a catch block, be sure that the memory was allocated and has not already been released.</p>
```

geoffw0 (commented):
My initial thoughts.  I will need to review the QL in a bit more detail.

cpp/ql/src/experimental/Security/CWE/CWE-476/DangerousUseOfExceptionBlocks.ql
ihsinme: Yes, you are right.
if you don't mind, I'll add it on a separate line.

cpp/ql/test/experimental/query-tests/Security/CWE/CWE-476/semmle/tests/test.cpp
ihsinme: I need some time to think about this place.

cpp/ql/src/experimental/Security/CWE/CWE-476/DangerousUseOfExceptionBlocks.cpp
ihsinme: I agree with you.
but this is a forced complexity.
I need to repeat difficulty level CVE.

ihsinme:
Thanks for corrections.

cpp/ql/test/experimental/query-tests/Security/CWE/CWE-476/semmle/tests/test.cpp
ihsinme: initialization `new` is a pretty vague thing.
```
Creates and initializes objects with dynamic storage duration, that is, objects whose lifetime is not necessarily limited by the scope in which they were created.
``` 
but at the same time 
```
The initializer is not optional if

type or new-type is an array of unknown bound
```
but in practice I was not able to get uninitialized data.  
https://www.onlinegdb.com/7SoYOoODS

if you agree, I will add a loop that resets the data after the memory allocation procedure?

cpp/ql/src/experimental/Security/CWE/CWE-476/DangerousUseOfExceptionBlocks.cpp
geoffw0: Its good to have simple and complicated cases in the tests.  However in the example, we need to be as simple as possible so that the reader can immediately see what's going on.

cpp/ql/test/experimental/query-tests/Security/CWE/CWE-476/semmle/tests/test.cpp
ihsinme: good example. thanks.

cpp/ql/src/experimental/Security/CWE/CWE-476/DangerousUseOfExceptionBlocks.cpp
ihsinme: I understand.
I will definitely make the tests and examples simpler.

ihsinme: good afternoon @geoffw0.
I tried to simplify.
Please take a look, maybe something else needs to be simplified.

cpp/ql/src/experimental/Security/CWE/CWE-476/DangerousUseOfExceptionBlocks.ql
```diff
@@ -0,0 +1,165 @@
...
+ *       external/cwe/cwe-415
+ */
+
+import cpp
+
+/** Holds if the release can occur twice. in the current block of catch and above in the block of try or other block catch. */
```
geoffw0, line 16:
```suggestion
/** Holds if `vr` may be released in the `try` block associated with `cb`, or in a `catch` block prior to `cb`. */
```

cpp/ql/src/experimental/Security/CWE/CWE-476/DangerousUseOfExceptionBlocks.ql
```diff
@@ -0,0 +1,165 @@
...
+pragma[inline]
+predicate doubleCallDelete(CatchAnyBlock cb, Variable vr) {
+  // Search for exceptions after freeing memory.
+  exists(Expr e1 |
+    (
```
geoffw0, line 21:
```suggestion
    // `e1` is a delete of `vr`
    (
```

cpp/ql/src/experimental/Security/CWE/CWE-476/DangerousUseOfExceptionBlocks.ql
```diff
@@ -0,0 +1,165 @@
...
+      e1 = vr.getAnAccess().getEnclosingStmt().(ExprStmt).getExpr().(DeleteArrayExpr) or
+      e1 = vr.getAnAccess().getEnclosingStmt().(ExprStmt).getExpr().(DeleteExpr)
+    ) and
+    e1.getEnclosingFunction() = cb.getEnclosingFunction() and
+    (
+      e1.getEnclosingStmt().getParentStmt*() = cb.getTryStmt().getStmt() and
```
geoffw0, line 27:
```suggestion
      // `e1` occurs in the `try` block associated with `cb`
      e1.getEnclosingStmt().getParentStmt*() = cb.getTryStmt().getStmt() and
```

cpp/ql/src/experimental/Security/CWE/CWE-476/DangerousUseOfExceptionBlocks.ql
```diff
@@ -0,0 +1,165 @@
...
+      exists(Expr e2, ThrowExpr th |
```
geoffw0, line 28:
```suggestion
      // `e2` is a `throw` (or a function call that may throw) that occurs in the `try` block after `e1`
      exists(Expr e2, ThrowExpr th |
```

cpp/ql/src/experimental/Security/CWE/CWE-476/DangerousUseOfExceptionBlocks.ql
```diff
@@ -0,0 +1,165 @@
...
+        (
+          e2 = th or
+          e2 = th.getEnclosingFunction().getACallToThisFunction()
+        ) and
+        e2.getEnclosingStmt().getParentStmt*() = cb.getTryStmt().getStmt() and
+        e1.getASuccessor+() = e2
+      ) and
+      not exists(AssignExpr ae |
```
geoffw0, line 36:
```suggestion
      // there is no assignment `vr = 0` in the `try` block after `e1`
      not exists(AssignExpr ae |
```

cpp/ql/src/experimental/Security/CWE/CWE-476/DangerousUseOfExceptionBlocks.ql
```diff
@@ -0,0 +1,165 @@
...
+        ae.getRValue().getValue() = "0" and
+        e1.getASuccessor+() = ae and
+        ae.getEnclosingStmt().getParentStmt*() = cb.getTryStmt().getStmt()
+      )
+      or
+      // Search for a situation when there is a higher catch block that also frees memory.
+      exists(CatchBlock cbt, Expr e2, ThrowExpr th |
+        e1.getEnclosingStmt().getParentStmt*() = cbt and
+        exists(cbt.getParameter()) and
+        (
```
geoffw0, line 47:
```suggestion
        // `e2` is a `throw` (or a function call that may throw) that occurs in the `catch` block after `e1`
        (
```

cpp/ql/src/experimental/Security/CWE/CWE-476/DangerousUseOfExceptionBlocks.ql
```diff
@@ -0,0 +1,165 @@
...
+          e2 = th or
+          e2 = th.getEnclosingFunction().getACallToThisFunction()
+        ) and
+        e2.getEnclosingStmt().getParentStmt*() = cbt and
+        e1.getASuccessor+() = e2 and
+        not exists(AssignExpr ae |
```
geoffw0, line 53:
```suggestion
        // there is no assignment `vr = 0` in the `catch` block after `e1`
        not exists(AssignExpr ae |
```

cpp/ql/src/experimental/Security/CWE/CWE-476/DangerousUseOfExceptionBlocks.ql
```diff
@@ -0,0 +1,165 @@
...
+          ae.getEnclosingStmt().getParentStmt*() = cbt
+        )
+      )
+    ) and
+    // Exclude the presence of a check in catch block.
+    not exists(IfStmt ifst |
+      ifst.getEnclosingStmt().getParentStmt*() = cb.getAStmt()
+    )
+  )
+}
```
geoffw0, line 66: I think there are several paths to simplifying this function and reducing the amount of duplicate code in it.  One would be to add an additional parameter to the predicate, for the block `b` where `vr` is deleted:
```
/** Holds if `vr` may be released in the block `b`, which is the `try` block associated with `cb` or a `catch` block prior to `cb`. */
predicate doubleCallDelete(Block b, CatchAnyBlock cb, Variable vr) {
```
In the first case we can set it with:
```
b = cb.getTryStmt().getStmt()
````
in the second it replaces the variable `cbt`, we just have to cast it to `CatchBlock` in the appropriate place:
```
exists(b.(CatchBlock).getParameter()) and
```
Having done this, both copies of the code for spotting assignments `vr = 0` can be combined into one.  I think the logic around `e2` can also be combined into one copy as well.

File: cpp/ql/src/experimental/Security/CWE/CWE-476/DangerousUseOfExceptionBlocks.ql [hunk 18]
```diff
@@ -0,0 +1,165 @@
...
+
+/** Holds if an exception can be thrown before the memory is allocated, and when the exception is handled, an attempt is made to access unallocated memory in the catch block. */
+pragma[inline]
+predicate pointerDereference(CatchAnyBlock cb, Variable vr, Variable vro) {
+  // Search exceptions before allocating memory.
+  exists(Expr e0, Expr e1 |
+    (
+      exists(AssignExpr ase |
```
geoffw0, line 74:
```suggestion
      // `e0` is a `new` expression (or equivalent function call) assigned to `vro`
      exists(AssignExpr ase |
```

cpp/ql/src/experimental/Security/CWE/CWE-476/DangerousUseOfExceptionBlocks.ql
```diff
@@ -0,0 +1,165 @@
...
+        ase = vro.getAnAccess().getEnclosingStmt().(ExprStmt).getExpr().(AssignExpr) and
+        (
+          e0 = ase.getRValue().(NewOrNewArrayExpr) or
+          e0 = ase.getRValue().(NewOrNewArrayExpr).getEnclosingFunction().getACallToThisFunction()
+        ) and
+        vro = ase.getLValue().(VariableAccess).getTarget()
+      )
+      or
+      exists(AssignExpr ase |
```
geoffw0, line 83: I'm not exactly sure what this part is for, but please could you add a comment explaining it?

cpp/ql/src/experimental/Security/CWE/CWE-476/DangerousUseOfExceptionBlocks.ql
```diff
@@ -0,0 +1,165 @@
...
+        ase = vro.getAnAccess().(Qualifier).getEnclosingStmt().(ExprStmt).getExpr().(AssignExpr) and
+        (
+          e0 = ase.getRValue().(NewOrNewArrayExpr) or
+          e0 = ase.getRValue().(NewOrNewArrayExpr).getEnclosingFunction().getACallToThisFunction()
+        ) and
+        not ase.getLValue() instanceof VariableAccess and
+        vro = ase.getLValue().getAPredecessor().(VariableAccess).getTarget()
+      )
+    ) and
+    exists(AssignExpr ase |
```
geoffw0, line 93:
```suggestion
    // `e1` is a `new` expression (or equivalent function call) assigned to `vr`
    exists(AssignExpr ase |
```

cpp/ql/src/experimental/Security/CWE/CWE-476/DangerousUseOfExceptionBlocks.ql
```diff
@@ -0,0 +1,165 @@
...
+      (
+        e1 = ase.getRValue().(NewOrNewArrayExpr) or
+        e1 = ase.getRValue().(NewOrNewArrayExpr).getEnclosingFunction().getACallToThisFunction()
+      ) and
+      vr = ase.getLValue().(VariableAccess).getTarget()
+    ) and
+    e0.getASuccessor*() = e1 and
+    e0.getEnclosingStmt().getParentStmt*() = cb.getTryStmt().getStmt() and
+    e1.getEnclosingStmt().getParentStmt*() = cb.getTryStmt().getStmt() and
+    exists(Expr e2, ThrowExpr th |
```
geoffw0, line 104:
```suggestion
    // `e2` is a `throw` (or a function call that may throw) that occurs in the `try` block before `e0`
    exists(Expr e2, ThrowExpr th |
```

cpp/ql/src/experimental/Security/CWE/CWE-476/DangerousUseOfExceptionBlocks.ql
```diff
@@ -0,0 +1,165 @@
...
+          va = exp.(DeleteExpr).getExpr().getAPredecessor+().(Qualifier).(VariableAccess) or
+          va = exp.(DeleteExpr).getExpr().getAPredecessor+().(VariableAccess)
+        ) and
+        vr = exp.(DeleteExpr).getExpr().(VariableAccess).getTarget()
+      ) and
+      va.getEnclosingStmt() = exp.getEnclosingStmt() and
+      vro = va.getTarget() and
+      vr != vro
+    ) and
+    pointerDereference(cb, vr, vro) and
```
geoffw0, line 145: My understanding of what's going on here is that we have a `try` block containing `vr = new ...`, and a `catch` block containing `delete vr`, but its possible to `throw` before the `vr = new ...` and reach the `delete vr`.

What is `vro` for?  Do we really need to track two `new` allocations to catch this bug?

geoffw0 (commented):
Some suggestions and questions about the query itself.

ihsinme:
thanks for your fixes.
I'm a little late with the answer, I hope to answer all your questions soon

cpp/ql/src/experimental/Security/CWE/CWE-476/DangerousUseOfExceptionBlocks.ql
ihsinme: this is necessary to account for indirect access to a variable, e.g.
```cpp
vro[i] = new vroType;
pointVro->vro[i] = new vroType;
pointVro.vro[i] = new vroType;
```
do you think the following comment is acceptable
// `e0` is a `new expression` (or equivalent function call) assigned to the array element `vro`

cpp/ql/src/experimental/Security/CWE/CWE-476/DangerousUseOfExceptionBlocks.ql
ihsinme: You are absolutely right if we are talking about deletion.
but I wanted to add detection of a problem with access to unallocated memory, as it happened in CVE-2020-15304.
```cpp
   for (size_t i = 0; i < _data->tileBuffers.size(); i++)
   {
        delete [] _data->tileBuffers[i]->buffer;
   }
```

cpp/ql/src/experimental/Security/CWE/CWE-476/DangerousUseOfExceptionBlocks.ql
ihsinme: thanks for the idea.
I hope I did as you intended?

geoffw0: Yep, the change looks good to me!

cpp/ql/src/experimental/Security/CWE/CWE-476/DangerousUseOfExceptionBlocks.ql
geoffw0: Yes I think so.  The more chunks of code that have comments, the easier it is to understand what's going on.

geoffw0:
Thanks for making the changes - we've definitely made the QL easier to follow!

I tried this query [on LGTM](https://lgtm.com/query/7394893711500709996/) but I didn't find any results (a few still to finish as I write this).  It might be a good idea to start thinking of ways we could widen the query?

cpp/ql/src/experimental/Security/CWE/CWE-476/DangerousUseOfExceptionBlocks.ql
geoffw0: :+1:

ihsinme:
> Thanks for making the changes - we've definitely made the QL easier to follow!
> 
> I tried this query [on LGTM](https://lgtm.com/query/7394893711500709996/) but I didn't find any results (a few still to finish as I write this). It might be a good idea to start thinking of ways we could widen the query?

my analysis showed several detections.
in projects using `openexr` for example.
https://github.com/mnaydenov/FreeImage-Adv.git

This project has already fixed a similar problem.
https://github.com/RTEMS/rtems-tools/pull/2
https://github.com/RTEMS/rtems-tools.git

I am currently evaluating how to expand the detect if in the first part look for access without operator `delete`.

For the second part, I don't see extensions yet. I'll be glad for a hint.

geoffw0:
For the benefit of my future self and other readers, here's an LGTM run of the query on the two projects mentioned above: https://lgtm.com/query/410819352512416178/

cpp/ql/src/experimental/Security/CWE/CWE-476/DangerousUseOfExceptionBlocks.cpp
ihsinme: can I make this item resolved?

geoffw0: Yeah, I think so.

ihsinme:
good afternoon @geoffw0 

Sorry for the delay in expanding this PR.
https://lgtm.com/query/4145429960886415578/

I suggest adding this version of detection, I like the preliminary results. although there are still a number of false positives associated with null initialization.

The idea behind the query is that an exception can be thrown when memory is allocated, and if it is freed during processing, then this behavior is undefined.

if you don't mind, I'll add that part to the main request.

geoffw0:
The new case seems to be finding good results, though the message will need to explain the problem clearly so that a user can make sense of what they're being shown (it needs to say that if the allocation in the try block fails, then an unallocated pointer will be freed in the catch block).

---

As an improvement, you should be able to spot variables that have been initialized to zero in a constructor initializer list using the `ConstructorFieldInit` class.

geoffw0 (approved):
Great!

geoffw0:
It looks like there are some new results in the test, and the QL needs to be autoformatted.

ihsinme:
@geoffw0 
I made changes now you can run checks.

ihsinme:
@geoffw0 
Are there any other errors during checks on this PR?

geoffw0 (approved):
Looks good to me and checks are all passing now. :+1:

PR #8246, opened 2022-02-25
CPP: Add query for CWE-754: Improper Check for Unusual or Exceptional Conditions when using functions scanf 

ihsinme:
the query looks for situations of working with `scanf` functions in situations where the developer has not left the possibility of controlling the correctness of work. There is no check of the returned value, the filled variable is not initialized and its value is not evaluated after the function has run. this leads to a situation where, continuing to work with a filled variable, the developer runs the risk of working with a randomly filled value at the declaration stage.

CVE-2019-15900

I'm working on a real fix for this issue.


cpp/ql/src/experimental/Security/CWE/CWE-754/ImproperCheckReturnValueScanf.ql
```diff
@@ -0,0 +1,108 @@
...
+  not exists(Expr e0, int i |
+    i in [n .. fc.getNumberOfArguments() - 1] and
+    fc.getASuccessor+() = e0 and
+    e0.getEnclosingElement() instanceof ComparisonOperation and
+    (
+      e0.(VariableAccess).getTarget() = fc.getArgument(i).(VariableAccess).getTarget() or
+      e0.(VariableAccess).getTarget() =
+        fc.getArgument(i).(AddressOfExpr).getOperand().(VariableAccess).getTarget()
+    )
+  )
```
jketema, line 107: Could you clarify what this achieves? Like everything starting from line 71, this seems to also look at how the argument is being used, so I'm not sure why this separate from what I see there.

jketema (commented):
Hi @ihsinme. Thanks for your contribution. Some comments below.

ihsinme:
thanks for the comments.
I will try to fix everything as soon as possible.

cpp/ql/src/experimental/Security/CWE/CWE-754/ImproperCheckReturnValueScanf.ql
```diff
@@ -0,0 +1,108 @@
...
+      )
+      or
+      not exists(Expr e1 |
+        fc.getASuccessor+() = e1 and
+        (
+          e1.(VariableAccess).getTarget() =
+            fc.getArgument(i).(AddressOfExpr).getOperand().(VariableAccess).getTarget() or
+          e1.(VariableAccess).getTarget() = fc.getArgument(i).(VariableAccess).getTarget()
+        )
+      )
```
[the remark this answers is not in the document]
ihsinme: you are right this part is artificially set separately.
the criterion for the lack of evaluation of the use of the result in the comparison operation will reduce the number of detections.

this criterion is not an error condition, rather, on the contrary, if you did not initialized a variable, did not check the correctness of the call, then the subsequent check is also an indicator of your incorrect work. therefore it stands separately and in the future it will be easier to remove it.

but if you do not agree with my reasoning, I will remove it

cpp/ql/src/experimental/Security/CWE/CWE-754/ImproperCheckReturnValueScanf.qhelp
```diff
@@ -0,0 +1,22 @@
+<!DOCTYPE qhelp PUBLIC
+  "-//Semmle//qhelp//EN"
+  "qhelp.dtd">
+<qhelp>
+<overview>
+<p>Working with reading data without validation procedures and with uninitialized arguments can lead to unpredictable consequences.</p>
```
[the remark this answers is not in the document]
ihsinme: what do you think about it?

```
Use of functions that write data to uninitialized variables without assessing the correctness of the work of these functions. Can have unexpected consequences when you take random data as the correct result.
```

cpp/ql/src/experimental/Security/CWE/CWE-754/ImproperCheckReturnValueScanf.qhelp
```diff
@@ -0,0 +1,22 @@
...
+
+</recommendation>
+<example>
+<p>The following example demonstrates erroneous and corrected work with a function call.</p>
```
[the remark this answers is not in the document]
ihsinme: what do you think about it?

```
The following example demonstrates the erroneous and corrected work with a function call that fills uninitialized variables without checking the correctness of its operation.
```

```
The following example demonstrates erroneous and corrected work with a function call scanf.
```

cpp/ql/src/experimental/Security/CWE/CWE-754/ImproperCheckReturnValueScanf.qhelp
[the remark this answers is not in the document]
jketema: I think you need to be very explicit that this query is about the `scanf` family of functions and the values assigned to its var args.

cpp/ql/src/experimental/Security/CWE/CWE-754/ImproperCheckReturnValueScanf.qhelp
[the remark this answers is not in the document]
jketema: You want to state explicitly what is erroneous.

cpp/ql/src/experimental/Security/CWE/CWE-754/ImproperCheckReturnValueScanf.ql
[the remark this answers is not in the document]
jketema: It should probably not be removed. Does one of your test cases cover this? This would help me to better understand what is going on.

cpp/ql/test/experimental/query-tests/Security/CWE/CWE-754/semmle/tests/test.cpp
```diff
@@ -0,0 +1,101 @@
...
+  scanf("%i", p); // BAD
+  return i;
+}
+
+int functionWork2() {
+  int i = 0;
+  char a[10] = {0, 0, 0, 0, 0, 0, 0, 0, 0, 0};
+  int b = 1;
+  int *p = &b;
+  scanf("%i", &i); // GOOD:the error can be determined by examining the initial value.
```
jketema, line 54:
```
// GOOD:the error can be determined by examining the initial value.
```
Technically this is not quite the case, as one of two things might have happened:

1. `scanf` succeeded and read a `0` value.
2. `scanf` failed, but this is inconsequential, because `i` was initialized.

So maybe write something like: "Argument initialized even when `scanf` fails"

cpp/ql/test/experimental/query-tests/Security/CWE/CWE-754/semmle/tests/test.cpp
```diff
@@ -0,0 +1,101 @@
...
+  int *p = &b;
+  scanf("%i", &i); // BAD
+  scanf("%s", a); // BAD
+  scanf("%i", p); // BAD
+  return i;
+}
+
+int functionWork2() {
+  int i = 0;
+  char a[10] = {0, 0, 0, 0, 0, 0, 0, 0, 0, 0};
```
jketema, line 51: Nitpicking:
```suggestion
  char a[10] = "";
```

cpp/ql/test/experimental/query-tests/Security/CWE/CWE-754/semmle/tests/test.cpp
```diff
@@ -0,0 +1,101 @@
...
+  scanf("%s", a); // GOOD:the error can be determined by examining the initial value.
+  scanf("%i", p); // GOOD:the error can be determined by examining the initial value.
+  return i;
+}
+
+int functionWork2_() {
+  int i;
+  i = 0;
+  char a[10];
+  a[0] = 0;
```
jketema, line 64: Nitpicking:
```suggestion
  a[0] = '\0';
```

cpp/ql/test/experimental/query-tests/Security/CWE/CWE-754/semmle/tests/test.cpp
```diff
@@ -0,0 +1,101 @@
...
+  int r;
+  r = scanf("%i", &i);
+  if (r != 1) // GOOD
+    return -1;
+  r = scanf("%s", a);
+  if (r == 1) // GOOD
+    return -1;
+  r = scanf("%i", p);
+  if (r != 1) // GOOD
+    return -1;
```
jketema, line 34: Technically these are good because `a` and `p` are not being used. Maybe do something like:
```suggestion
  r = scanf("%s", a);
  if (r == 1)
    char a0 = a[0]; // GOOD
  r = scanf("%i", p);
  if (r == 1)
    int pv = *p;  // GOOD
```

cpp/ql/test/experimental/query-tests/Security/CWE/CWE-754/semmle/tests/test.cpp
```diff
@@ -0,0 +1,101 @@
...
+}
+
+int functionWork1b() {
+  int i;
+  char a[10];
+  int b;
+  int *p = &b;
+  scanf("%i", &i); // BAD
+  scanf("%s", a); // BAD
+  scanf("%i", p); // BAD
```
jketema, line 45: `a` and `p` are not being used, so these are good at the moment?

cpp/ql/test/experimental/query-tests/Security/CWE/CWE-754/semmle/tests/test.cpp
```diff
@@ -0,0 +1,101 @@
...
+}
+
+int functionWork2() {
+  int i = 0;
+  char a[10] = {0, 0, 0, 0, 0, 0, 0, 0, 0, 0};
+  int b = 1;
+  int *p = &b;
+  scanf("%i", &i); // GOOD:the error can be determined by examining the initial value.
+  scanf("%s", a); // GOOD:the error can be determined by examining the initial value.
+  scanf("%i", p); // GOOD:the error can be determined by examining the initial value.
```
jketema, line 56: Again `a` and `p` are not being used.

cpp/ql/test/experimental/query-tests/Security/CWE/CWE-754/semmle/tests/test.cpp
```diff
@@ -0,0 +1,101 @@
...
+  int i;
+  i = 0;
+  char a[10];
+  a[0] = 0;
+  int b;
+  b=1;
+  int *p = &b;
+  scanf("%i", &i); // GOOD:the error can be determined by examining the initial value.
+  scanf("%s", a); // GOOD:the error can be determined by examining the initial value.
+  scanf("%i", p); // GOOD:the error can be determined by examining the initial value.
```
jketema, line 70: `a` and `p` are not being used.

cpp/ql/test/experimental/query-tests/Security/CWE/CWE-754/semmle/tests/test.cpp
```diff
@@ -0,0 +1,101 @@
...
+  return 0;
+}
+int functionWork2b_() {
+  char a[10];
+  scanf("%s", a); // BAD
+  globalVala2 = a[0];
+  return 0;
+}
+
+void functionRunner() {
```
jketema, line 93: This shouldn't be needed for testing, as the query does not involve any inter-procedural analyses.

jketema (commented):
Thanks for the updates, especially making the tests cover more cases.

cpp/ql/src/experimental/Security/CWE/CWE-754/ImproperCheckReturnValueScanf.qhelp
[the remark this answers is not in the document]
ihsinme: Using the scanf family of functions that write data to previously uninitialized argument variables without evaluating the correct operation of these functions. Can have unexpected consequences when you take random data as the correct result.

cpp/ql/src/experimental/Security/CWE/CWE-754/ImproperCheckReturnValueScanf.qhelp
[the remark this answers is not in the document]
ihsinme: at this point, I usually point out the example without describing it in detail. so it's hard for me to understand your suggestion. I would like to leave it like that.

```
The following example demonstrates erroneous and corrected work with a function call scanf.
```

cpp/ql/test/experimental/query-tests/Security/CWE/CWE-754/semmle/tests/test.cpp
ihsinme: i added usage via branched return.
if you do not mind?

cpp/ql/src/experimental/Security/CWE/CWE-754/ImproperCheckReturnValueScanf.ql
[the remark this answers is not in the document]
ihsinme: it's a good idea.
I added a simple example of `functionWork1_`, which the developer knows for sure that his return number is greater than 10. This does not give him a guarantee that the code will work correctly, but it improves it.

cpp/ql/test/experimental/query-tests/Security/CWE/CWE-754/semmle/tests/test.cpp
jketema: That's fine, as long as there's a use.

cpp/ql/src/experimental/Security/CWE/CWE-754/ImproperCheckReturnValueScanf.qhelp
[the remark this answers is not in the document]
jketema: How about:
```
The `scanf` family functions does not require the memory pointed to by
its additional pointer arguments to be initialized before calling. The user is
required to check the return value of `scanf` and similar functions to
establish how many of the additional arguments were assigned values. Not
checking the return value and reading one of the arguments not assigned a
value is undefined behavior and may have unexpected consequences.
```

cpp/ql/src/experimental/Security/CWE/CWE-754/ImproperCheckReturnValueScanf.qhelp
```diff
@@ -0,0 +1,22 @@
+<!DOCTYPE qhelp PUBLIC
+  "-//Semmle//qhelp//EN"
+  "qhelp.dtd">
+<qhelp>
+<overview>
+<p>Working with reading data without validation procedures and with uninitialized arguments can lead to unpredictable consequences.</p>
+
+</recommendation>
```
jketema, line 8:
```suggestion
</overview>

<recommendation>
<p>
The user should check the return value of `scanf` and related functions and check that any additional argument was assigned a value before reading the additional argument.
</p>
</recommendation>
```

cpp/ql/src/experimental/Security/CWE/CWE-754/ImproperCheckReturnValueScanf.qhelp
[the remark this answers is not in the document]
jketema: How about:
```
The first first example below is correct, as value of `i` is only read once it is checked
that `scanf` has read one item. The second example is incorrect, as the return value
of `scanf` is not checked, and as `scanf` might have failed to read any item before
returning.
```

cpp/ql/src/experimental/Security/CWE/CWE-754/ImproperCheckReturnValueScanf.ql
```diff
@@ -0,0 +1,103 @@
+/**
+ * @name Improper check return value scanf
+ * @description Using a function call without the ability to evaluate the correctness of the work can lead to unexpected results.
```
jketema, line 3:
```suggestion
 * @description Not checking the return value of `scanf` and related functions may lead to undefined behavior.
```

cpp/ql/src/experimental/Security/CWE/CWE-754/ImproperCheckReturnValueScanf.ql
```diff
@@ -0,0 +1,103 @@
+/**
+ * @name Improper check return value scanf
```
jketema, line 2:
```suggestion
 * @name Improper check of return value of scanf
```

cpp/ql/src/experimental/Security/CWE/CWE-754/ImproperCheckReturnValueScanf.ql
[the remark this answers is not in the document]
jketema: Thanks.

cpp/ql/src/experimental/Security/CWE/CWE-754/ImproperCheckReturnValueScanf.ql
```diff
@@ -0,0 +1,103 @@
...
+      )
+    ) and
+    not exists(AssignExpr ae |
+      ae.getLValue() = vt.getAnAccess().getParent() and
+      ae.getASuccessor+() = fc
+    ) and
+    not exists(FunctionCall f0 |
+      f0.getAnArgument().getAChild() = vt.getAnAccess() and
+      f0.getASuccessor+() = fc
+    ) and
```
jketema, line 53: As nothing in the first half of this predicate refers to `e0`:
```suggestion
  // Fillable argument was not initialized.
  vt instanceof LocalScopeVariable and
  not vt.getAnAssignment().getASuccessor+() = fc and
  (
    not vt.hasInitializer()
    or
    exists(Expr e, Variable v |
      e = vt.getInitializer().getExpr() and
      v = e.(AddressOfExpr).getOperand().(VariableAccess).getTarget() and
      (
        not v.hasInitializer() and
        not v.getAnAssignment().getASuccessor+() = fc
      )
    )
  ) and
  not exists(AssignExpr ae |
    ae.getLValue() = vt.getAnAccess().getParent() and
    ae.getASuccessor+() = fc
  ) and
  not exists(FunctionCall f0 |
    f0.getAnArgument().getAChild() = vt.getAnAccess() and
    f0.getASuccessor+() = fc
  ) and
  exists(Expr e0 |
```

cpp/ql/src/experimental/Security/CWE/CWE-754/ImproperCheckReturnValueScanf.ql
```diff
@@ -0,0 +1,103 @@
...
+        e0.(Assignment).getRValue().(ArrayExpr).getArrayBase().(VariableAccess).getTarget() = vt
+      )
+      or
+      e0.getEnclosingStmt() instanceof ReturnStmt and
+      e0.(VariableAccess).getTarget() = vt
+      or
+      not exists(Expr e1 |
+        fc.getASuccessor+() = e1 and
+        e1.(VariableAccess).getTarget() = vt
+      )
```
jketema, line 68: What does this do? If I remove this, all tests still pass.

cpp/ql/test/experimental/query-tests/Security/CWE/CWE-754/semmle/tests/test.cpp
```diff
@@ -0,0 +1,117 @@
...
+    return (int)*a;
+  if(retIndex == 1)
+    return *p;
+  return i;
+}
+int functionWork1_() {
+  int i;
+  scanf("%i",&i); // GOOD
+  if(i<10)
+    return -1;
```
jketema, line 64: This is actually `BAD`, as `i` is not initialized, `scanf` might not have scanned any arguments, and as `i` is accessed after `scanf`.

cpp/ql/src/experimental/Security/CWE/CWE-754/ImproperCheckReturnValueScanf.ql
ihsinme: Yes, you are right.
this section is for finding cases where the variable is not used inside functions that call 1111.
we removed this case from the test cases, replacing it with a return from the function.
but I would like to point out that I would like to keep this detection zone.

cpp/ql/test/experimental/query-tests/Security/CWE/CWE-754/semmle/tests/test.cpp
ihsinme: Yes, you are right.
but I still see two sources of false detection.
1. when the developer knows exactly what is in the source string or fills in the string just before reading. (the situation is strange, but it takes place in reality)
2. when the developer knows what to return, for example, an even number or a maximum of 3 or a code word. and this is a way to control the correct operation of scanf.
I don't know yet how to remove those two FP. when we clear them we will lose about TP.
that is why I said above that this detector is temporary.

ihsinme:
I want to thank you for your corrections.
especially the English text.

cpp/ql/src/experimental/Security/CWE/CWE-754/ImproperCheckReturnValueScanf.ql
jketema: Please add a test case.

cpp/ql/test/experimental/query-tests/Security/CWE/CWE-754/semmle/tests/test.cpp
jketema: For (1) can you point me to an actual case? For (2) if `scanf` fails, reading `i` is undefined behavior, which C/C++ don't allow, and the code in which this is detected should really be fixed.

ihsinme:
```cpp
2734:  assert(text[0] == '\\');
2735:
2736:  switch(text[1])
2737:  {
2738:  case 'x':
2739:    if (!isxdigit(text[2]) || !isxdigit(text[3]))
2740:      return 0;
2741:    hex[0] = text[2];
2742:    hex[1] = text[3];
2743:    hex[2] = '\0';
2744:    sscanf(hex, "%x", &hex_value);
2745:    *value = (uint8_t) hex_value;
2746:    break;
2747:
2748:  case 'n':
2749:    *value = '\n';
2750:    break;
2751:
2752:  case 't':
2753:    *value = '\t';
2754:    break;
```

https://github.com/VirusTotal/yara/blob/master/libyara/re_lexer.c#L2744-L2744

ihsinme: On the second point, do you think that it should be removed?

jketema: It can be left in, but it should be noted in the test that it's it's bad and good good, but is also not detected. SO change the comment by `// BAD [NOT DETECTED]`. Also a test that shows a good case should be added (this should absolutely *not* be derived from the virus total example)

ihsinme: I have a few questions about the first point.

1. At your request, I gave a real example and the essence of `(this should absolutely not be derived from the virus total example)` is not very clear to me.
Should I remove the example from the discussions?
2. It is very difficult to add this situation to the test.
2.1. if not this `if (!isxdigit(text[2]) || !isxdigit(text[3]))` is `BAD`
2.2. if instead of `sscanf(hex, "%x", &hex_value);` this is `sscanf(hex, "%x-%x", &hex_value,&hex_value);` this is `BAD`
so please tell me again what example you would like me to add.

jketema: > Should I remove the example from the discussions?

No. That example should just not end up in the CodeQL codebase.

jketema: > It is very difficult to add this situation to the test.

How about something like:
```c
char number[] = "42";
int d;
sscanf(number, "%d", &d); // GOOD: sscanf always succeeds
return d;
```

ihsinme: > > Should I remove the example from the discussions?
> 
> No. That example should just not end up in the CodeQL codebase.

We have already closed this issue by making changes to the test file.

ihsinme: > > It is very difficult to add this situation to the test.
> 
> How about something like:
> 
> ```c
> char number[] = "42";
> int d;
> sscanf(number, "%d", &d); // GOOD: sscanf always succeeds
> return d;
> ```

if we make this sentence a test file, I don't see how we can distinguish it from one like.
```cpp
char number[] = "a42";
int d;
sscanf(number, "%d", &d); // GOOD: sscanf always succeeds
return d;
```

```cpp
char number[] = "42";
int d;
char c;
sscanf(number, "%c %d", &c,&d); // GOOD: sscanf always succeeds
return d;
```

jketema: Apologies. I completely misremembered what the related piece of CodeQL was doing. How about:
```
char number[] = "42";
int d;
sscanf(number, "%d", &d); // GOOD: sscanf always succeeds
if (d < 16)
  return -1;
```

cpp/ql/src/experimental/Security/CWE/CWE-754/ImproperCheckReturnValueScanf.ql
jketema: Presumably the test case for this is `functionWork3b`?

ihsinme: yes, sorry for not specifying that explicitly.

cpp/ql/test/experimental/query-tests/Security/CWE/CWE-754/semmle/tests/test.cpp
jketema: The `sscanf` prototype is missing now, and the `.expected` file is out-of-date

jketema (approved):
Thanks for the contribution and addressing my comments.

Not for this PR, but you might want to have a look at the data flow library https://codeql.github.com/docs/codeql-language-guides/analyzing-data-flow-in-cpp/. I think that a rewrite based on the library might help to improve the quality of the analysis of what happens after a `scanf` call.

PR #8247, opened 2022-02-25
CPP: Add query for CWE-190: Integer Overflow or Wraparound when using transform after operation

ihsinme:
This query looks for integer overflow situations that are not covered by the `cpp/integer-multiplication-cast-to-long query`. the first part is belated and meaningless transformations like `s=(size_t)(i*i)`, which would still happen according to the law of expression. It's possible that the developer didn't use the conversion location correctly. the second part is the conversion of types that are the same size but different in sign, and then use the result in pointer arithmetic.

CVE-2021-43618

https://github.com/facebook/zstd/pull/2424


MathiasVP:
Hi @ihsinme,

Thanks for another contribution. I really like this one :) Will review it very soon!

cpp/ql/src/experimental/Security/CWE/CWE-190/DangerousUseOfTransformationAfterOperation.ql
```diff
@@ -0,0 +1,94 @@
...
+    e0.(FunctionCall)
+        .getTarget()
+        .getParameter(argumentPosition(e0.(FunctionCall), mexp, _))
+        .getType()
+        .getSize() = mexp.getConversion().getConversion().getType().getSize()
+  )
+}
+
+/** Holds if the situation of a possible signed overflow used in pointer arithmetic is found. */
+predicate signSmallerWithEqualSizes(MulExpr mexp, Expr e1, Expr e2) {
```
MathiasVP, line 47: This predicate suffers from the same performance issue as I highlighted in the `conversionDoneLate` predicate, and it needs a similar performance fix (inline the predicate or bind `e1` and `e2` using `mexp.hasOperands(e1, e2)`). Otherwise it'll likely time out on many real-world projects.

cpp/ql/src/experimental/Security/CWE/CWE-190/DangerousUseOfTransformationAfterOperation.ql
```diff
@@ -0,0 +1,94 @@
...
+      )
+    ) and
+    exists(PointerAddExpr pa |
+      ae.getASuccessor+() = pa and
+      pa.getAnOperand().(VariableAccess).getTarget() = ae.getLValue().(VariableAccess).getTarget()
+    )
+  )
+}
+
+from MulExpr mexp, string msg
```
MathiasVP, line 75: Can you add a comment around here that explains why this query catches things that `cpp/integer-multiplication-cast-to-long` doesn't catch?

cpp/ql/src/experimental/Security/CWE/CWE-190/DangerousUseOfTransformationAfterOperation.ql
```diff
@@ -0,0 +1,94 @@
...
+    not e1.isConstant() and
+    not e1.hasConversion() and
+    not e1.hasConversion() and
+    (
+      e2.isConstant() or
+      not e2.hasConversion()
+    ) and
+    (
+      conversionDoneLate(mexp, e1, e2) and
+      msg = "this transformation is applied after multiplication"
```
MathiasVP, line 88:
```suggestion
      msg = "This transformation is applied after multiplication."
```

cpp/ql/src/experimental/Security/CWE/CWE-190/DangerousUseOfTransformationAfterOperation.ql
```diff
@@ -0,0 +1,94 @@
...
+      or
+      signSmallerWithEqualSizes(mexp, e1, e2) and
+      msg = "possible signed overflow followed by offset of the pointer out of bounds"
```
MathiasVP, line 91:
```suggestion
      msg = "Possible signed overflow followed by offset of the pointer out of bounds."
```

MathiasVP, line 91: We have a [range analysis library](https://codeql.github.com/codeql-standard-libraries/cpp/semmle/code/cpp/rangeanalysis/SimpleRangeAnalysis.qll/module.SimpleRangeAnalysis.html) that can rule out cases that obviously doesn't overflow. See how it's used here: https://github.com/github/codeql/blob/main/cpp/ql/src/Likely%20Bugs/Arithmetic/SignedOverflowCheck.ql#L30 (there's only a `convertedExprMightOverflow` predicate that checks for both overflow and underflow at the same time). Would that be useful to rule out false positives in some cases?

cpp/ql/test/experimental/query-tests/Security/CWE/CWE-190/DangerousUseOfTransformationAfterOperation/test.cpp
```diff
@@ -0,0 +1,23 @@
+void testCall (unsigned long);
+void functionWork() {
+  unsigned long aL;
+  char aA[10],*aP;
+  unsigned char aUC;
+  int aI;
+  unsigned int aUI;
+  aI = (aUI*8)/10; // GOOD
```
MathiasVP, line 8: This is using unitialized memory (which is undefined behavior). Could you take `aUI` as a parameter instead (or initialize it)?

Similarly, some of the other expressions also use uninitialized memory. Please initialize those or take them as parameters instead :)

cpp/ql/src/experimental/Security/CWE/CWE-190/DangerousUseOfTransformationAfterOperation.qhelp
```diff
@@ -0,0 +1,24 @@
+<!DOCTYPE qhelp PUBLIC
+  "-//Semmle//qhelp//EN"
+  "qhelp.dtd">
+<qhelp>
+<overview>
+<p>Searching for places where data type conversion occurs late and does not affect the computational result.</p>
```
MathiasVP, line 6: This describes what the query is doing, but not actually what is wrong. Could you rephrase this to explain why it's bad that a type conversion occurs late and does not affect the computational result?

cpp/ql/src/experimental/Security/CWE/CWE-190/DangerousUseOfTransformationAfterOperation.ql
```diff
@@ -0,0 +1,94 @@
+/**
+ * @name Dangerous use of transformation after operation.
+ * @description By using the transformation after the operation, you are doing a pointless and dangerous action.
+ * @kind problem
+ * @id cpp/dangerous-use-of-transformation-after-operation.
```
MathiasVP, line 5:
```suggestion
 * @id cpp/dangerous-use-of-transformation-after-operation
```
Query ids aren't supposed to have dots in them. That's also why the CI checks are failing currently.

MathiasVP (commented):
Here are my initial comments.

ihsinme:
thanks for the comments.
I will try to fix everything as soon as possible.

separately sorry for repeated mistakes.

cpp/ql/src/experimental/Security/CWE/CWE-190/DangerousUseOfTransformationAfterOperation.ql
```diff
@@ -0,0 +1,94 @@
...
+}
+
+/** Holds if a nonsensical type conversion situation is found. */
+predicate conversionDoneLate(MulExpr mexp, Expr e1, Expr e2) {
+  mexp.getConversion().hasExplicitConversion() and
+  mexp.getConversion() instanceof ParenthesisExpr and
+  mexp.getConversion().getConversion() instanceof CStyleCast and
+  mexp.getConversion().getConversion().getType().getSize() > mexp.getType().getSize() and
+  mexp.getConversion().getConversion().getType().getSize() > e2.getType().getSize() and
+  mexp.getConversion().getConversion().getType().getSize() > e1.getType().getSize() and
```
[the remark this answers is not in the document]
ihsinme: https://github.com/github/codeql/blob/bc22b9b2083cca1be6cc8a85eb385ed911f1b701/cpp/ql/src/experimental/Security/CWE/CWE-190/DangerousUseOfTransformationAfterOperation.ql#L78

Please advise.
doesn't the presence of this restriction in the main block solve the problem?

cpp/ql/src/experimental/Security/CWE/CWE-190/DangerousUseOfTransformationAfterOperation.qhelp
ihsinme: how do you view this text?

```
Search for places where the result of the multiplication is subjected to explicit conversion, not the arguments. Therefore, during the multiplication period, you can lose meaningful data.
```

cpp/ql/src/experimental/Security/CWE/CWE-190/DangerousUseOfTransformationAfterOperation.ql
ihsinme: how do you view this text?

```
// Search for an explicit conversion of the result of multiplication in the situation of multiplying types of a smaller size than the result 
// or casting the result of multiplication of comparable types to signed ones.
// In these situations, it is necessary to use the conversion not of the result, but of the arguments.
```

cpp/ql/src/experimental/Security/CWE/CWE-190/DangerousUseOfTransformationAfterOperation.ql
ihsinme: I am considering this idea.
in my experience of using `Range` there are too many surprises of uncertainty.

in the first part, I hope for a difference in the sizes of the types of arguments and results, in the second time, that the signed result is 1 bit less than the arguments.

I propose to return to this point if other cleaning options are exhausted?

cpp/ql/src/experimental/Security/CWE/CWE-190/DangerousUseOfTransformationAfterOperation.qhelp
MathiasVP: That sounds more appropriate, yes. :)

cpp/ql/src/experimental/Security/CWE/CWE-190/DangerousUseOfTransformationAfterOperation.ql
MathiasVP: > in my experience of using Range there are too many surprises of uncertainty.

I'd be interested in hearing what you mean by this.

> I propose to return to this point if other cleaning options are exhausted?

That's fine.

cpp/ql/src/experimental/Security/CWE/CWE-190/DangerousUseOfTransformationAfterOperation.ql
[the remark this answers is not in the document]
MathiasVP: Remember that QL is evaluated bottom-up. So the evaluator _starts_ by evaluating `conversionDoneLate` and `signSmallerWithEqualSizes`. Then, once both of those predicates are evaluated, the main block is evaluated.

cpp/ql/src/experimental/Security/CWE/CWE-190/DangerousUseOfTransformationAfterOperation.ql
MathiasVP: That still doesn't explain (at least not to me) what the difference between this query and `cpp/integer-multiplication-cast-to-long` is. As far as I remember, the goal of `cpp/integer-multiplication-cast-to-long` is also to detect cases like:
```cpp
int i1 = ...;
int i2 = ...;
long j = i * i;
```
where there's an implicit cast on the result of the multiplication, but where the cast should really be on the arguments.

Is the difference just that this query requires the cast to be explicit, while `cpp/integer-multiplication-cast-to-long` requires the cast to be implicit?

ihsinme: yes, the idea of the request is that the developer uses an explicit override, which would have happened anyway, therefore, this is either stylistically incorrect or overflow.

MathiasVP:
Here are the CI failures:
```
Executing tests in /home/runner/work/semmle-code/semmle-code/ql/cpp/ql/test/experimental/query-tests/Security/CWE/CWE-190/DangerousUseOfTransformationAfterOperation.
2022-03-03T16:33:32.9863889Z --- expected
2022-03-03T16:33:32.9864202Z +++ actual
2022-03-03T16:33:32.9864550Z @@ -1,4 +1,4 @@
2022-03-03T16:33:32.9865409Z -| test.cpp:9:8:9:12 | ... * ... | possible signed overflow followed by offset of the pointer out of bounds |
2022-03-03T16:33:32.9866305Z -| test.cpp:13:24:13:28 | ... * ... | this transformation is applied after multiplication |
2022-03-03T16:33:32.9876401Z -| test.cpp:16:28:16:32 | ... * ... | this transformation is applied after multiplication |
2022-03-03T16:33:32.9877278Z -| test.cpp:19:22:19:26 | ... * ... | this transformation is applied after multiplication |
2022-03-03T16:33:32.9878102Z +| test.cpp:9:8:9:12 | ... * ... | Possible signed overflow followed by offset of the pointer out of bounds. |
2022-03-03T16:33:32.9878849Z +| test.cpp:13:24:13:28 | ... * ... | This transformation is applied after multiplication. |
2022-03-03T16:33:32.9879567Z +| test.cpp:16:28:16:32 | ... * ... | This transformation is applied after multiplication. |
2022-03-03T16:33:32.9880271Z +| test.cpp:19:22:19:26 | ... * ... | This transformation is applied after multiplication. |
2022-03-03T16:33:32.9891524Z ##[error][2999/5383 comp 6.6s eval 171ms] FAILED(RESULT) /home/runner/work/semmle-code/semmle-code/ql/cpp/ql/test/experimental/query-tests/Security/CWE/CWE-190/DangerousUseOfTransformationAfterOperation/DangerousUseOfTransformationAfterOperation.qlref
```

cpp/ql/src/experimental/Security/CWE/CWE-190/DangerousUseOfTransformationAfterOperation.ql
MathiasVP: I see. Thanks for clearing that up. Could you extend the description you wrote [here](https://github.com/github/codeql/pull/8247#discussion_r816512144) to say that this query finds cases similar to the ones found by `cpp/integer-multiplication-cast-to-long`, but requires the cast to be explicit?

MathiasVP (approved):
LGTM!

PR #8254, opened 2022-02-25
Ruby: initial prototype of models-as-data

asgerf:
Adds support for the API-graph based models-as-data format, discussed [here](https://github.com/github/codeql-team/discussions/619).

The model for `Regexp.escape` and `ActiveStorage::Filename` have been ported.

For example, a slightly simplified version of `ActiveStorage::Filename`:
```codeql
private class Summaries extends ModelInput::SummaryModelCsv {
  row =
  "activestorage;;Member[ActiveStorage].Member[Filename].Method[new];Argument[0];ReturnValue;taint",
  }
}
```
The row for summaries has the form:
```
package;type;path;input;output;kind
```
- The `package` is the name of a Ruby gem, or the empty string for the standard library. As discussed, the package name is not currently used for anything. This is because we can't reliably see what transitive dependencies exist (only the direct ones), but it may be used in the future. If nothing else, it serves as documentation.
- The `type` can be:
  - an empty string to refer to the global scope,
  - the string `any`, to refer to any expression,
  - the name of a type defined by a `TypeModelCsv` row.
- The `path` is similar to an input/output spec, but is evaluated as a path through the API graph, starting at any node identified by the `(package,type)` pair.
- The `kind` is `value` for value-preserving flow, and `taint` for taint flow.

----

Currently, very few of the existing models would actually benefit from being ported, because:
- The model is dynamically generated, sometimes in a database-dependent way.
- The model contributes to a concept, rather than being a source, sink, or summary.

The former of these has to do with these being standard library models, which wasn't the best use-case for MaD in the first place. The issue of how to contribute to a concept is something to investigate later, possibly after sharing `Concepts.qll`.

----

[Evaluation](https://github.com/github/codeql-dca-main/tree/data/asgerf/mad-prototype__Differences__code-scanning__1/reports) looks fine


ruby/ql/lib/codeql/ruby/ApiGraphs.qll
```diff
@@ -266,6 +266,29 @@ module API {
...
+   */
+  abstract class EntryPoint extends string {
+    bindingset[this]
+    EntryPoint() { any() }
+
+    /** Gets a data-flow node corresponding to a use-node for this entry point. */
+    DataFlow::LocalSourceNode getAUse() { none() }
+
+    /** Gets a data-flow node corresponding to a def-node for this entry point. */
+    DataFlow::Node getARhs() { none() }
```
aibaars, line 285: It's not really clear what right-hand-side means; the right-hand of what? The name suggest there should also be a left-hand-side. Wouldn't `getADef` be a better name, since there is already a `getAUse` method.

ruby/ql/lib/codeql/ruby/frameworks/data/internal/ApiGraphModelsSpecific.qll
```diff
@@ -0,0 +1,141 @@
...
+
+private import ruby
+private import codeql.ruby.DataFlow
+private import codeql.ruby.dataflow.internal.DataFlowPrivate as DataFlowPrivate
+private import ApiGraphModels
+
+class Unit = DataFlowPrivate::Unit;
+
+import codeql.ruby.ApiGraphs
+import codeql.ruby.dataflow.internal.AccessPathSyntax as AccessPathSyntax
```
aibaars, line 26: Use `private import` if possible to avoid pulling all transitively imported names into scope.

```suggestion
private import codeql.ruby.ApiGraphs
private import codeql.ruby.dataflow.internal.AccessPathSyntax as AccessPathSyntax
```

ruby/ql/lib/codeql/ruby/dataflow/FlowSummary.qll
```diff
@@ -2,6 +2,8 @@
 
 import ruby
 import codeql.ruby.DataFlow
```
aibaars, line 4:
```suggestion
private import ruby
private import codeql.ruby.DataFlow
```

aibaars (commented):
Looks good to me.

ruby/ql/lib/codeql/ruby/frameworks/data/internal/ApiGraphModelsSpecific.qll
```diff
@@ -0,0 +1,98 @@
...
+}
+
+/**
+ * Gets a Ruby-specific API graph successor of `node` reachable by resolving `token`.
+ */
+bindingset[token]
+API::Node getExtraSuccessorFromInvoke(InvokeNode node, AccessPathToken token) {
+  token.getName() = "Instance" and
+  result = node.getInstance()
+}
```
hmac, line 120: I'm not sure what we are trying to match here. `InvokeNode` corresponds to a method call, which doesn't have a concept of an instance, as far as I know. The closest thing would be if it returns a class which can be instantiated e.g. `foo().new`, but that should be handled in `getExtraSuccessorFromNode` since the path would be `Method[foo].ReturnValue.Instance`.

Is this maybe a JS-ism that doesn't translate to Ruby, like `new foo()` or something?

ruby/ql/test/query-tests/security/cwe-094/CodeInjection.expected
```diff
@@ -5,6 +5,8 @@ edges
 | CodeInjection.rb:3:12:3:24 | ...[...] :  | CodeInjection.rb:21:21:21:24 | code |
 | CodeInjection.rb:3:12:3:24 | ...[...] :  | CodeInjection.rb:27:15:27:18 | code |
 | CodeInjection.rb:3:12:3:24 | ...[...] :  | CodeInjection.rb:30:19:30:22 | code |
+| CodeInjection.rb:3:12:3:24 | ...[...] :  | CodeInjection.rb:36:24:36:27 | code :  |
```
hmac, line 8: (very minor nit): this commit message has a typo: `s/flwo/flow`.

ruby/ql/lib/codeql/ruby/ApiGraphs.qll
hmac: I had a similar question [here](https://github.com/github/codeql/pull/7819#discussion_r801210664) about `API::Node.getARhs()`. I still believe it's a confusing name in the context of Ruby.

ruby/ql/lib/codeql/ruby/ApiGraphs.qll
```diff
@@ -271,6 +271,26 @@ module API {
     DataFlow::CallNode getCallNode() { this = Impl::MkMethodAccessNode(result) }
   }
 
+  /**
+   * An API entry point.
```
hmac, line 275: What is an entry point? Could we add an example or a bit of explanation here?

hmac:
Oh, I had one other question: do we have/can we add any consistency queries or something equivalent to report when a MaD row is malformed? I spent a long time debugging problems arising from having an extra semicolon in a row string.

ruby/ql/lib/codeql/ruby/frameworks/data/internal/ApiGraphModelsSpecific.qll
asgerf: @hmac is right. I'll add a comment to explain, since deliberate re-export can be hard to distinguish from accidental re-export.

ruby/ql/lib/codeql/ruby/ApiGraphs.qll
erik-krogh: EntryPoints are used when we want `API::Node`s for something that is not connected to the root of the API-graph.   
In JS that is usually references to global variables for specific library models (including all global variables in the API-graph would blow up).    
[Here is an example of how we've used an EntryPoint](https://github.com/github/codeql/blob/bcdbfefb2b49b5aa6d483fdb703097780056b070/javascript/ql/lib/semmle/javascript/frameworks/History.qll#L7-L21).

ruby/ql/lib/codeql/ruby/frameworks/data/ModelsAsData.qll
```diff
@@ -0,0 +1,33 @@
...
+ *
+ * The `package` part of a CSV row should be the name of a Ruby gem, or the empty
+ * string if it's referring to the standard library.
+ *
+ * The `type` part can be one of the following:
+ *   - the empty string, referring to the global scope,
+ *   - the string `any`, referring to any expression, or
+ *   - the name of a type definition from `ModelInput::TypeModelCsv`
+ *
+ * The `path` part is a dot
```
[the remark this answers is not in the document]
asgerf: Guess I forgot distracted and forgot to finish it. Will fix it

ruby/ql/lib/codeql/ruby/frameworks/data/ModelsAsData.qll
```diff
@@ -0,0 +1,33 @@
...
+import Shared::ModelOutput as ModelOutput
+private import codeql.ruby.dataflow.RemoteFlowSources
+
+/**
+ * A remote flow source originating from a CSV source row.
+ */
+private class RemoteFlowSourceFromCsv extends RemoteFlowSource::Range {
+  RemoteFlowSourceFromCsv() { this = ModelOutput::getASourceNode("remote").getAnImmediateUse() }
+
+  override string getSourceType() { result = "Remote flow" }
```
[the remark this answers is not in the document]
asgerf: Add `(from model)` for now.

In general this is something that will also come up a lot when contributing to concepts, and the same solution ought to be used here. I'd rather not try to solve that in this PR.

ruby/ql/lib/codeql/ruby/ApiGraphs.qll
asgerf: Elaborated a bit

asgerf:
> Oh, I had one other question: do we have/can we add any consistency queries or something equivalent to report when a MaD row is malformed? I spent a long time debugging problems arising from having an extra semicolon in a row string.

Absolutely. I was planning to do this later, but on second thought things like this should be added as soon as possible. I've gone ahead and pushed a few more commits to add checks for this.

ruby/ql/lib/codeql/ruby/ApiGraphs.qll
```diff
@@ -266,6 +266,36 @@ module API {
...
+    DataFlow::LocalSourceNode getAUse() { none() }
+
+    /** Gets a data-flow node corresponding to a def-node for this entry point. */
+    DataFlow::Node getARhs() { none() }
+
+    /** Gets a call corresponding to a method access node for this entry point. */
+    DataFlow::CallNode getACall() { none() }
+
+    /** Gets an API-node for this entry point. */
+    API::Node getNode() { result = root().getASuccessor(Label::entryPoint(this)) }
```
hvitved, line 298: `getANode`?

ruby/ql/lib/codeql/ruby/dataflow/FlowSummary.qll
```diff
@@ -165,3 +167,33 @@ private class SummarizedCallableAdapter extends Impl::Public::SummarizedCallable
...
 class RequiredSummaryComponentStack = Impl::Public::RequiredSummaryComponentStack;
+
+private class SummarizedCallableFromModel extends SummarizedCallable {
+  string input;
+  string output;
+  string kind;
+
+  SummarizedCallableFromModel() {
+    ModelOutput::summaryModel(input, output, kind) and
+    this = input + ";" + output + ";" + kind
```
hvitved, line 178: I would expect that this would instead be the concatenation of `package`, `type`, and `path`, since that is what identifies (calls to) the method. This also means that if a method has multiple summaries, then we will still only synthesize one method.

ruby/ql/lib/codeql/ruby/frameworks/ActiveStorage.qll
```diff
@@ -17,43 +18,13 @@ class ActiveStorageFilenameSanitizedCall extends Path::PathSanitization::Range,
...
-    input = "Argument[-1]" and
-    output = "ReturnValue" and
-    preservesValue = false
+/** Taint related to `ActiveStorage::Filename`. */
+private class Summaries extends ModelInput::SummaryModelCsv {
+  override predicate row(string row) {
+    row =
+      [
+        "activestorage;;Member[ActiveStorage].Member[Filename].Method[new];Argument[0];ReturnValue;taint",
+        "activestorage;;Member[ActiveStorage].Member[Filename].Instance.Method[sanitized];Argument[-1];ReturnValue;taint",
```
hvitved, line 27: (not your PR, but `Argument[-1]` does not work; should be `Receiver` instead (actually, it should be `Argument[Receiver]`, but it is the latter for now.)

ruby/ql/lib/codeql/ruby/frameworks/data/internal/ApiGraphModels.qll
```diff
@@ -0,0 +1,515 @@
...
+ *    It can also be a synthetic type name defined by a type definition (see type definitions below).
+ * 3. The `path` column is a `.`-separated list of "access path tokens" to resolve, starting at the node selected by `package` and `type`.
+ *    The possible access path tokens are:
+ *     - Member[x] : a property named `x`. May be a comma-separated list of named.
+ *     - Argument[n]: the n-th argument to a call. May be a range of form `x..y` (inclusive) and/or a comma-separated list.
+ *                    Additionally, `N-1` refers to the last argument, `N-2` refers to the second-last, and so on.
+ *     - Parameter[n]: the n-th parameter of a callback. May be a range of form `x..y` (inclusive) and/or a comma-separated list.
+ *     - ReturnValue: the value returned by a function call
+ *     - Instance: the value returned by a constructor call
+ *     - Awaited: the value from a resolved promise/future-like object
```
hvitved, line 32: JS specific

ruby/ql/lib/codeql/ruby/frameworks/data/internal/ApiGraphModelsSpecific.qll
```diff
@@ -0,0 +1,165 @@
...
+bindingset[package, type, path]
+API::Node getExtraNodeFromPath(string package, string type, AccessPath path, int n) {
+  isRelevantFullPath(package, type, path) and
+  package = any(string s) and // Allow any package name, see `isPackageUsed`.
+  type = "" and
+  n = 0 and
+  result = API::root()
+  or
+  // A row of form `;any;Method[foo]` should match any method named `foo`.
+  package = any(string s) and
```
hvitved, line 59: same as above

ruby/ql/lib/codeql/ruby/frameworks/data/internal/ApiGraphModelsSpecific.qll
```diff
@@ -0,0 +1,165 @@
...
+    methodName = token.getAnArgument()
+  )
+}
+
+/**
+ * An API graph entry point corresponding to a method name such as `foo` in `;any;Method[foo]`.
+ *
+ * This ensures that the API graph rooted in that method call is materialized.
+ */
+private class EntryPointFromAnyType extends API::EntryPoint {
```
hvitved, line 86: Shouldn't we make sure to import this file, or `ApiGraphModels.qll` into `ApiGraph.qll` to avoid recomputation?

ruby/ql/lib/codeql/ruby/frameworks/data/internal/ApiGraphModelsSpecific.qll
```diff
@@ -0,0 +1,165 @@
...
+  int getNumArgument() { result = getCallNode().getNumberOfArguments() }
+}
+
+/** Gets the `InvokeNode` corresponding to a specific invocation of `node`. */
+InvokeNode getAnInvocationOf(API::Node node) { result = node }
+
+/**
+ * Holds if `name` is a valid name for an access path token in the identifying access path.
+ */
+bindingset[name]
```
hvitved, line 144: Why this `bindingset`?

hvitved (changes requested):
Looks great, Asger!

ruby/ql/lib/codeql/ruby/ApiGraphs.qll
hvitved: `getASource`?

ruby/ql/lib/codeql/ruby/dataflow/FlowSummary.qll
asgerf: > This also means that if a method has multiple summaries, then we will still only synthesize one method.

But if there are different ways to access the same method, you end up synthesizing multiple identical methods.

For example, if you had these summaries, would you like to synthesize one or two methods?
```
foo;;Member[Bar].Method[m];Argument[0];ReturnValue;taint
foo;;Member[Bar].Member[Impl].Method[m];Argument[0];ReturnValue;taint
```

From our perspective it's not well-defined if these two are "the same method".

I guess it comes down to whether or not there are adverse effects to:
1. sharing callables for unrelated methods that happen to have the same input/output/kind spec, and
2. a call having multiple call targets, due to having multiple input/output/kind specs

I can see some problems with (1) actually, as it can confuse an interprocedural analysis that has imperfect context sensitivity. The monomorphic API use assumption, for example, would not hold because these callables end up being much more polymorphic than real-world code would be. And sometimes an ad-hoc analysis does a bit of interprocedural reasoning without bothering with context sensitivity.

OK, I've convinced myself and will instead canonicalize the callables based on `package,type,path`.

ruby/ql/lib/codeql/ruby/frameworks/ActiveStorage.qll
asgerf: I'd like to update it in this PR so these models can serve as a reference how to do things.

In the process of doing this I came to few realizations:
- Ruby API graphs don't have a label for `Receiver`. I initially wanted to sync these which is easy enough, but due to how the receiver argument works in Ruby, I don't think this will ever be needed in the identifying access path or in API graphs even. So I deliberately left that out. (In JS one can pass an arbitrary value as `this` and some libraries take advantage of that, so the situation is different there).
- `BlockArgument` was not implemented in the identifying access path. Added that in a separate commit.
- There is no support for keyword arguments in neither the identifying access path nor in input/output specs. It's straightforward to add, but we'll need to agree on whether to name it `Argument[name]` or `KeywordArgument[name]`. In Python there is a desire to use `Argument[name]` because almost any positional argument has a corresponding named argument, which we'll want to express with a single token. For now I've left it out.

ruby/ql/lib/codeql/ruby/frameworks/data/internal/ApiGraphModels.qll
asgerf: After trying to update this documentation, I came to the conclusion that it is too prone to falling out of date, at least at this stage in development. I've rephrased it, PTAL.

ruby/ql/lib/codeql/ruby/frameworks/data/internal/ApiGraphModelsSpecific.qll
asgerf: Sure, if you want.

In JS we try to ensure all files that aren't query-specific are transitively included from `javascript.qll` (but not necessarily made visible). Is there a reason you're not doing this in Ruby? Seems more robust IMO.

ruby/ql/lib/codeql/ruby/frameworks/data/internal/ApiGraphModelsSpecific.qll
asgerf: It would be part of the predicate's signature, if there was a way to specify the signature expected from `ApiGraphModels.qll`. Since we can't specify the signature, the predicate has the exact the signature expected by `ApiGraphModels.qll`, rather than the most precise signature it could have.

ruby/ql/lib/codeql/ruby/frameworks/data/internal/ApiGraphModelsSpecific.qll
```diff
@@ -0,0 +1,165 @@
...
+ */
+predicate isExtraValidNoArgumentTokenInIdentifyingAccessPath(string name) {
+  name = ["Instance", "WithBlock", "WithoutBlock"]
+}
+
+/**
+ * Holds if `argument` is a valid argument to an access path token with the given `name`, occurring
+ * in an identifying access path.
+ */
+bindingset[name, argument]
```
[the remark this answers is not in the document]
asgerf: Same answer as above

asgerf:
Thanks for the thorough review @hvitved, I've addressed all comments, PTAL

ruby/ql/lib/codeql/ruby/ApiGraphs.qll
asgerf: It's a metaphor, relating to a concept that's pretty well-established in with compilers/program analysis. The right-hand side of an assignment is the value "flowing into" something.

I think you'd come to regret the names you're suggesting if you actually tried to used the predicate, rather than just looking at its definition. Please keep in mind that you can't always evaluate an API just by looking at its definition; you have to consider what it looks like at the use site.

`getADefNode`:
1. Same problem as `getADefinition`, a _definition_ of a parameter is an argument?!
2. API graphs have a concept called "def-nodes", and the predicate doesn't actually return a def-node, it returns a data-flow node (when the receiver is a def-node).

`getASource`:
1. It's not clear if it involves transitive flow. Does it correspond to `getAValueReachingRhs` or just `getARhs`? With `getARhs` it's pretty obvious that transitive flow is not implied. `getAnImmediateSource` might be OK though.
2. It doesn't return `LocalSourceNode`, though the name suggests that it might.
3. You end up using `getASource` when defining a sink, and you'll get confused review comments suggesting that you made a mistake there. For example:
```codeql
  node = API::moduleImport("foo").getMember("bar").getParameter(0).getASource() // wait, source or sink?
}
```

----

With that being said, I recognise there's a lot of backlash against the current name, and I'm open to finding a better name. But the new name has to actually be better.

Also, it's a little awkward having the discussion in this PR since API graphs have been around for a long time already, it's just that Python and Ruby didn't have the full version. It would be nice if you'd also recognize that you are suggesting a rename of an existing predicate with lots of uses. It's not some new thing I invented on the spot here.

ruby/ql/lib/codeql/ruby/frameworks/data/internal/ApiGraphModelsSpecific.qll
hvitved: As long as they are not made visible that should work too. However, I don't know if that makes compilation slower.

hvitved (approved):
Thanks for addressing all my comments.

PR #8255, opened 2022-02-25
Python: Prevent magic/inlining in `getCase`

tausbn:
This is a simplified version of https://github.com/github/codeql/pull/8028 consisting of just the `nomagic` fix.


yoff (approved):
This seems like an effective minimal change, let us start with this.

PR #8272, opened 2022-02-28
Ruby: Add rb/tainted-format-string query

hmac:
This shares as much as possible with the JS version.

### DIFF
--- a/javascript/ql/lib/semmle/javascript/security/dataflow/TaintedFormatStringSpecific.qll
+++ b/javascript/ql/lib/semmle/javascript/security/dataflow/TaintedFormatStringSpecific.qll
@@ -1,0 +1,6 @@
+/**
+ * Provides JS-specific imports needed for `TaintedFormatStringQuery` and `TaintedFormatStringCustomizations`.
+ */
+
+import javascript
+import semmle.javascript.security.dataflow.DOM
--- a/ruby/ql/lib/codeql/ruby/security/TaintedFormatStringSpecific.qll
+++ b/ruby/ql/lib/codeql/ruby/security/TaintedFormatStringSpecific.qll
@@ -21,7 +21,7 @@ abstract class PrintfStyleCall extends DataFlow::CallNode {
   /**
    * Gets then `n`th formatted argument of this call.
    */
-  DataFlow::Node getFormatArgument(int n) { n > 0 and result = this.getArgument(n) }
+  DataFlow::Node getFormatArgument(int n) { n >= 0 and result = this.getArgument(n + 1) }
 }
 
 /**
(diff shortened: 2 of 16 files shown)


erik-krogh (dismissed):
JS :+1:

alexrford (changes requested):
Looks good, only minor comments.

ruby/ql/lib/codeql/ruby/security/TaintedFormatStringSpecific.qll
```diff
@@ -0,0 +1,61 @@
...
+    this = API::getTopLevelMember("Kernel").getAMethodCall("printf")
+    or
+    this.asExpr().getExpr() instanceof UnknownMethodCall and
+    this.getMethodName() = "printf"
+  }
+
+  // Kernel#printf supports two signatures:
+  //   printf(io, string, ...)
+  //   printf(string, ...)
+  override DataFlow::Node getFormatString() { result = this.getArgument([0, 1]) }
```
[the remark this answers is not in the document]
hmac: Good point! I've made this change, though instead of `instanceof StringLiteralCfgNode` I've used `.getConstantValue().isString(_)`, to take advantage of local flow, e.g. in
```ruby
fmt = "hello %s %s"
printf(fmt, params[:format], "arg2")
```

alexrford (approved):
👍

javascript/ql/lib/semmle/javascript/security/dataflow/TaintedFormatStringSpecific.qll
```diff
@@ -0,0 +1,6 @@
+/**
+ * Provides JS-specific imports needed for `TaintedFormatStringQuery` and `TaintedFormatStringCustomizations`.
+ */
+
+import javascript
+import semmle.javascript.security.dataflow.DOM
```
erik-krogh, line 6: I think this `import semmle.javascript.security.dataflow.DOM` is redundant.  
But that's fine.  
(`ql/redundant-import` anyone?)

erik-krogh (approved):

javascript/ql/lib/semmle/javascript/security/dataflow/TaintedFormatStringSpecific.qll
hmac: A query for redundant imports would be great! Even better would be a compiler warning delivered via LSP.

erik-krogh: I don't think it belongs as a compiler warning.  
In the above case both of the imports eventually import each other, so deleting any one of them will solve it.  

And do you want a compiler warning on both for that? I think that feels wrong, and so does the arbitrary choice of warning on the second.  
And sometimes (redundant) imports and their order can be important (I'm not sure why, but I think the CPP team can tell stories).   

In QL-for-QL: arbitary choices and lingering warnings are fine, so I think it should be a query.  
(But I want to figure out how we'll get support for parameterized modules before I start more QL-for-QL query writing).

PR #8289, opened 2022-02-28
Python: Fix syntax error in `with` test output

tausbn:
Depends on an internal PR. The two lines in question were caused by
the insertion of an extra node due to the failure to parse a trailing
comma corrcetly.


yoff (approved):
LGTM

PR #8293, opened 2022-02-28
Ruby: parse more string literals as regular expressions

aibaars:
This pull request changes the `Regex` class to be `abstract` and extend `StringlikeLiteral`. This allows adding custom subclasses to parse more string literals as regular expressions if needed. The default subclasses are `RegExpLiteral` and string literals that somehow flow into functions like `Regex.new`, `Regex.compile`, `String.match` and `String.match?`.

### DIFF
--- a/.github/workflows/ruby-qltest.yml
+++ b/.github/workflows/ruby-qltest.yml
@@ -63,6 +63,7 @@ jobs:
   qltest:
     runs-on: ubuntu-latest
     strategy:
+      fail-fast: false
       matrix:
         slice: ["1/2", "2/2"]
     steps:
--- a/ruby/ql/lib/change-notes/2022-02-28-regex-string-literals.md
+++ b/ruby/ql/lib/change-notes/2022-02-28-regex-string-literals.md
@@ -1,0 +1,4 @@
+---
+category: minorAnalysis
+---
+* The `Regex` class now an abstract class that extends `StringlikeLiteral` with implementations for `RegExpLiteral` and string literals that 'flow' into functions that are known to interpret string arguments as regular expressions such `Regex.new` and `String.match`.
--- a/ruby/ql/lib/change-notes/2022-02-28-regex-string-literals.md
+++ b/ruby/ql/lib/change-notes/2022-02-28-regex-string-literals.md
@@ -1,4 +1,4 @@
 ---
 category: minorAnalysis
 ---
-* The `Regex` class now an abstract class that extends `StringlikeLiteral` with implementations for `RegExpLiteral` and string literals that 'flow' into functions that are known to interpret string arguments as regular expressions such `Regex.new` and `String.match`.
+* The `Regex` class is now an abstract class that extends `StringlikeLiteral` with implementations for `RegExpLiteral` and string literals that 'flow' into functions that are known to interpret string arguments as regular expressions such as `Regex.new` and `String.match`.
--- a/ruby/ql/lib/codeql/ruby/security/performance/ParseRegExp.qll
+++ b/ruby/ql/lib/codeql/ruby/security/performance/ParseRegExp.qll
@@ -14,7 +14,7 @@ private import codeql.ruby.ApiGraphs
 
 /**
  * A `StringlikeLiteral` containing a regular expression term, that is, either
- * a regular expression literal, a string literal used in a context where
+ * a regular expression literal, or a string literal used in a context where
  * it is parsed as regular expression.
  */
 abstract class RegExp extends AST::StringlikeLiteral {
--- a/ruby/ql/lib/codeql/ruby/security/performance/ParseRegExp.qll
+++ b/ruby/ql/lib/codeql/ruby/security/performance/ParseRegExp.qll
@@ -7,8 +7,32 @@
 
 private import codeql.ruby.ast.Literal as AST
 private import codeql.Locations
+private import codeql.ruby.DataFlow
+private import codeql.ruby.TaintTracking
+private import codeql.ruby.typetracking.TypeTracker
+private import codeql.ruby.ApiGraphs
 
-class RegExp extends AST::RegExpLiteral {
+/**
+ * A `StringlikeLiteral` containing a regular expression term, that is, either
+ * a regular expression literal, a string literal used in a context where
+ * it is parsed as regular expression.
+ */
+abstract class RegExp extends AST::StringlikeLiteral {
+  /**
+   * Holds if this `RegExp` has the `s` flag for multi-line matching.
+   */
+  predicate isDotAll() { none() }
+
+  /**
+   * Holds if this `RegExp` has the `i` flag for case-insensitive matching.
+   */
+  predicate isIgnoreCase() { none() }
+
+  /**
+   * Gets the flags for this `RegExp`, or the empty string if it has no flags.
+   */
+  string getFlags() { result = "" }
+
   /**
    * Helper predicate for `charSetStart(int start, int end)`.
    *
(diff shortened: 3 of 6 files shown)


ruby/ql/lib/change-notes/2022-02-28-regex-string-literals.md
```diff
@@ -0,0 +1,4 @@
+---
+category: minorAnalysis
+---
+* The `Regex` class now an abstract class that extends `StringlikeLiteral` with implementations for `RegExpLiteral` and string literals that 'flow' into functions that are known to interpret string arguments as regular expressions such `Regex.new` and `String.match`.
```
intrigus-lgtm, line 4:
```suggestion
* The `Regex` class is now an abstract class that extends `StringlikeLiteral` with implementations for `RegExpLiteral` and string literals that 'flow' into functions that are known to interpret string arguments as regular expressions such as `Regex.new` and `String.match`.
```

aibaars: Thanks!

ruby/ql/lib/codeql/ruby/security/performance/ParseRegExp.qll
```diff
@@ -7,8 +7,32 @@
...
 private import codeql.ruby.ast.Literal as AST
 private import codeql.Locations
+private import codeql.ruby.DataFlow
+private import codeql.ruby.TaintTracking
+private import codeql.ruby.typetracking.TypeTracker
+private import codeql.ruby.ApiGraphs
+
+/**
+ * A `StringlikeLiteral` containing a regular expression term, that is, either
+ * a regular expression literal, a string literal used in a context where
```
alexrford, line 17:
```suggestion
 * a regular expression literal, or a string literal used in a context where
```

.github/workflows/ruby-qltest.yml
```diff
@@ -63,6 +63,7 @@ jobs:
   qltest:
     runs-on: ubuntu-latest
     strategy:
+      fail-fast: false
```
alexrford, line 66: Was this change prompted by anything in particular?

ruby/ql/lib/codeql/ruby/security/performance/ParseRegExp.qll
```diff
@@ -933,3 +957,63 @@ class RegExp extends AST::RegExpLiteral {
...
+ * Holds if `source` may be interpreted as a regular expression.
+ */
+cached
+private predicate isInterpretedAsRegExp(DataFlow::Node source) {
+  // The first argument to an invocation of `Regexp.new` or `Regexp.compile`.
+  source = API::getTopLevelMember("Regexp").getAMethodCall(["compile", "new"]).getArgument(0)
+  or
+  // The argument of a call that coerces the argument to a regular expression.
+  exists(DataFlow::CallNode mce |
+    mce.getMethodName() = ["match", "match?"] and
```
alexrford, line 993: I guess that this is mostly for cases where the receiver is a string, and we're not too concerned about other `match` methods giving us false positive sources?

alexrford (commented):
Looks nice, some very minor comments.

ruby/ql/lib/codeql/ruby/security/performance/ParseRegExp.qll
aibaars: That's right. In addition the value flowing into the first argument should be some form of string literal. So it's quite restrictive already. The worst that can happen is that we wrongly consider some string literals to be potential regular expressions.

aibaars:
@intrigus-lgtm and @alexrford Thanks for the reviews!

alexrford (approved):

ruby/ql/lib/codeql/ruby/security/performance/ParseRegExp.qll
```diff
@@ -933,3 +957,63 @@ class RegExp extends AST::RegExpLiteral {
...
+    source = mce.getArgument(0)
+  )
+}
+
+/**
+ * Gets a node whose value may flow (inter-procedurally) to `re`, where it is interpreted
+ * as a part of a regular expression.
+ */
+private DataFlow::Node regExpSource(DataFlow::Node re, TypeBackTracker t) {
```
hvitved, line 1002: Couldn't this instead be a taint-tracking configuration where the sources are all nodes that have a constant string value, and the sinks are `isInterpretedAsRegExp`?

ruby/ql/lib/codeql/ruby/security/performance/ParseRegExp.qll
```diff
@@ -933,3 +957,63 @@ class RegExp extends AST::RegExpLiteral {
...
+
+  override string getFlags() { result = this.getFlagString() }
+}
+
+private class ParsedStringRegExp extends RegExp {
+  private DataFlow::Node parse;
+
+  ParsedStringRegExp() { this = regExpSource(parse).asExpr().getExpr() }
+
+  DataFlow::Node getAParse() { result = parse }
```
hvitved, line 974: This appears to not be called?

ruby/ql/lib/codeql/ruby/security/performance/RegExpTreeView.qll
```diff
@@ -795,3 +795,47 @@ class RegExpNamedCharacterProperty extends RegExpTerm, TRegExpNamedCharacterProp
...
+/**
+ * A node whose value may flow to a position where it is interpreted
+ * as a part of a regular expression.
+ */
+abstract class RegExpPatternSource extends DataFlow::Node {
+  /**
+   * Gets a node where the pattern of this node is parsed as a part of
+   * a regular expression.
+   */
+  abstract DataFlow::Node getAParse();
```
hvitved, line 808: This appears to not be called?

ruby/ql/lib/codeql/ruby/security/performance/RegExpTreeView.qll
```diff
@@ -795,3 +795,47 @@ class RegExpNamedCharacterProperty extends RegExpTerm, TRegExpNamedCharacterProp
...
+
+  /**
+   * Gets the root term of the regular expression parsed from this pattern.
+   */
+  abstract RegExpTerm getRegExpTerm();
```
hvitved, line 813: This appears to not be called?

hvitved (commented):
Sorry for the late review. I also noted a significant slowdown on `opal`, is that acceptable, or something we can prevent?

ruby/ql/lib/codeql/ruby/security/performance/ParseRegExp.qll
aibaars: These methods are called in a follow-up pull request.

ruby/ql/lib/codeql/ruby/security/performance/ParseRegExp.qll
```diff
@@ -7,8 +7,32 @@
...
+private import codeql.ruby.TaintTracking
+private import codeql.ruby.typetracking.TypeTracker
+private import codeql.ruby.ApiGraphs
+
+/**
+ * A `StringlikeLiteral` containing a regular expression term, that is, either
+ * a regular expression literal, or a string literal used in a context where
+ * it is parsed as regular expression.
+ */
+abstract class RegExp extends AST::StringlikeLiteral {
```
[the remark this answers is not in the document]
aibaars: This makes it possible for queries (or libraries) to contribute additional regular expression "sinks" and have those parsed as well.

ruby/ql/lib/codeql/ruby/security/performance/ParseRegExp.qll
aibaars: I adopted the same approach as JavaScript. See also: https://github.com/github/codeql/blob/d3da790191fc84a6061fd7a8f688db5e6961f398/javascript/ql/lib/semmle/javascript/Regexp.qll#L1018

ruby/ql/lib/codeql/ruby/security/performance/ParseRegExp.qll
```diff
@@ -933,3 +957,63 @@ class RegExp extends AST::RegExpLiteral {
...
+  override predicate isIgnoreCase() { none() }
+
+  override string getFlags() { none() }
+}
+
+/**
+ * Holds if `source` may be interpreted as a regular expression.
+ */
+cached
+private predicate isInterpretedAsRegExp(DataFlow::Node source) {
```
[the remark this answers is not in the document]
aibaars: Good idea.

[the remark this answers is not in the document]
aibaars: https://github.com/github/codeql/pull/8330

ruby/ql/lib/codeql/ruby/security/performance/ParseRegExp.qll
hvitved: Yes, more light weight, but less precise. But with the pruning of the data flow library I would not expect it to actually be much slower. And the computation will be cached anyway.

ruby/ql/lib/codeql/ruby/security/performance/ParseRegExp.qll
hvitved: OK.

PR #8298, opened 2022-03-01
C++: Fix `GVN` performance on more invalid IR

MathiasVP:
As we've seen in the past, the performance of GVN degrades exponentially when it's run on invalid IR.

In this case, a project had some `InheritanceConversionInstruction`s with 10 base classes and 10 derived classes, which meant that some instructions had 100 value numbers. To make things worse, this could lead to an instruction getting `100^2` value numbers if one invalid `InheritanceConversionInstruction` depended on another invalid `InheritanceConversionInstruction`.

The fix is to exclude `InheritanceConversionInstruction`s with multiple base or derived classes from the GVN loop. This is also how we handle other cases of invalid IR.

While we're at it, I also decided to replace the uses of `strictcount(...) = 1` with `unique( | | ...)`. This code predates the introduction of `unique`, which is why this wasn't done initially.

### DIFF
--- a/cpp/ql/lib/semmle/code/cpp/ir/implementation/aliased_ssa/gvn/internal/ValueNumberingInternal.qll
+++ b/cpp/ql/lib/semmle/code/cpp/ir/implementation/aliased_ssa/gvn/internal/ValueNumberingInternal.qll
@@ -106,6 +106,12 @@ private predicate filteredNumberableInstruction(Instruction instr) {
   or
   instr instanceof FieldAddressInstruction and
   count(instr.(FieldAddressInstruction).getField()) != 1
+  or
+  instr instanceof InheritanceConversionInstruction and
+  (
+    count(instr.(InheritanceConversionInstruction).getBaseClass()) != 1 or
+    count(instr.(InheritanceConversionInstruction).getDerivedClass()) != 1
+  )
 }
 
 private predicate variableAddressValueNumber(
@@ -115,8 +121,7 @@ private predicate variableAddressValueNumber(
   // The underlying AST element is used as value-numbering key instead of the
   // `IRVariable` to work around a problem where a variable or expression with
   // multiple types gives rise to multiple `IRVariable`s.
-  instr.getIRVariable().getAST() = ast and
-  strictcount(instr.getIRVariable().getAST()) = 1
+  unique( | | instr.getIRVariable().getAST()) = ast
 }
 
 private predicate initializeParameterValueNumber(
@@ -133,8 +138,7 @@ private predicate constantValueNumber(
   ConstantInstruction instr, IRFunction irFunc, IRType type, string value
 ) {
   instr.getEnclosingIRFunction() = irFunc and
-  strictcount(instr.getResultIRType()) = 1 and
-  instr.getResultIRType() = type and
+  unique( | | instr.getResultIRType()) = type and
   instr.getValue() = value
 }
 
@@ -152,7 +156,7 @@ private predicate fieldAddressValueNumber(
 ) {
   instr.getEnclosingIRFunction() = irFunc and
   instr.getField() = field and
-  strictcount(instr.getField()) = 1 and
+  unique( | | instr.getField()) = field and
   tvalueNumber(instr.getObjectAddress()) = objectAddress
 }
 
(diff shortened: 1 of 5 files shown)


cpp/ql/lib/semmle/code/cpp/ir/implementation/aliased_ssa/gvn/internal/ValueNumberingInternal.qll
```diff
@@ -152,7 +156,7 @@ private predicate fieldAddressValueNumber(
 ) {
   instr.getEnclosingIRFunction() = irFunc and
   instr.getField() = field and
```
geoffw0, line 158: I believe this line is redundant now, like the cases above.

cpp/ql/lib/semmle/code/cpp/ir/implementation/aliased_ssa/gvn/internal/ValueNumberingInternal.qll
```diff
@@ -106,6 +106,12 @@ private predicate filteredNumberableInstruction(Instruction instr) {
   or
   instr instanceof FieldAddressInstruction and
   count(instr.(FieldAddressInstruction).getField()) != 1
+  or
+  instr instanceof InheritanceConversionInstruction and
+  (
+    count(instr.(InheritanceConversionInstruction).getBaseClass()) != 1 or
+    count(instr.(InheritanceConversionInstruction).getDerivedClass()) != 1
```
geoffw0, line 113: Could/should this be `not unique(...)`?

geoffw0 (approved):
LGTM.

I'd like to see the performance improvement on `RainerKuemmerle/g2o`, but it might be a bit of a pain to test this in full - I'll settle for the evidence performance hasn't regressed (which DCA appears to have provided) and wait for the next dist build to look at `RainerKuemmerle/g2o`.

cpp/ql/lib/semmle/code/cpp/ir/implementation/aliased_ssa/gvn/internal/ValueNumberingInternal.qll
MathiasVP: Oops. You're right.

cpp/ql/lib/semmle/code/cpp/ir/implementation/aliased_ssa/gvn/internal/ValueNumberingInternal.qll
MathiasVP: It could, but I remember cases where `not unique(...)` didn't always compile to the DIL code we expected.

cpp/ql/lib/semmle/code/cpp/ir/implementation/aliased_ssa/gvn/internal/ValueNumberingInternal.qll
MathiasVP: Fixed in 4acae4a2d1b560017ac10ff61043b23e48136eef

MathiasVP:
> LGTM.
> 
> I'd like to see the performance improvement on `RainerKuemmerle/g2o`, but it might be a bit of a pain to test this in full - I'll settle for the evidence performance hasn't regressed (which DCA appears to have provided) and wait for the next dist build to look at `RainerKuemmerle/g2o`.

Yeah, I considered adding `RainerKuemmerle/g2o` to DCA, but it seems to be more of a library than an application. So I don't think we'd find useful security results in that project. So if it's only for monitoring performance I don't think it's worth adding such a giant project to the suite.

cpp/ql/lib/semmle/code/cpp/ir/implementation/aliased_ssa/gvn/internal/ValueNumberingInternal.qll
geoffw0: OK, lets not worry about it now.

geoffw0 (approved):
:+1: from me.

MathiasVP:
DCA shows no performance changes on our small set of databases (which is expected). The new DCA result is spurious. It has appeared in other runs as well. Merging! :shipit:

PR #8302, opened 2022-03-01
Ruby: TypeTracker: add smallstep for functions that return their arguments

aibaars:
Depends on https://github.com/github/codeql/pull/8317

### DIFF
--- a/ruby/ql/lib/codeql/ruby/typetracking/TypeTrackerSpecific.qll
+++ b/ruby/ql/lib/codeql/ruby/typetracking/TypeTrackerSpecific.qll
@@ -11,12 +11,31 @@ class Node = DataFlowPublic::Node;
 
 class TypeTrackingNode = DataFlowPublic::LocalSourceNode;
 
+/** Holds if there is a simple local flow step from `nodeFrom` to `nodeTo` */
 predicate simpleLocalFlowStep = DataFlowPrivate::localFlowStepTypeTracker/2;
 
+/**
+ * Holds if data can flow from `node1` to `node2` in a way that discards call contexts.
+ */
 predicate jumpStep = DataFlowPrivate::jumpStep/2;
 
-/** Holds if there is a level step from `pred` to `succ`. */
-predicate levelStep(Node pred, Node succ) { none() }
+/**
+ * Holds if there is a summarized local flow step from `nodeFrom` to `nodeTo`,
+ * because there is direct flow from a parameter to a return. That is, summarized
+ * steps are not applied recursively.
+ */
+pragma[nomagic]
+private predicate summarizedLocalStep(Node nodeFrom, Node nodeTo) {
+  exists(DataFlowPublic::ParameterNode param, DataFlowPrivate::ReturningNode returnNode |
+    DataFlowPrivate::LocalFlow::getParameterDefNode(param.getParameter())
+        .(TypeTrackingNode)
+        .flowsTo(returnNode) and
+    callStep(nodeTo.asExpr(), nodeFrom, param)
+  )
+}
+
+/** Holds if there is a level step from `nodeFrom` to `nodeTo`. */
+predicate levelStep(Node nodeFrom, Node nodeTo) { summarizedLocalStep(nodeFrom, nodeTo) }
 
 /**
  * Gets the name of a possible piece of content. This will usually include things like
@@ -48,6 +67,13 @@ private predicate viableParam(
   )
 }
 
+private predicate callStep(ExprNodes::CallCfgNode call, Node nodeFrom, Node nodeTo) {
+  exists(DataFlowDispatch::ParameterPosition pos |
+    argumentPositionMatch(call, nodeFrom, pos) and
+    viableParam(call, nodeTo, pos)
+  )
+}
+
 /**
  * Holds if `nodeFrom` steps to `nodeTo` by being passed as a parameter in a call.
  *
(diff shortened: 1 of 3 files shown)


ruby/ql/lib/codeql/ruby/typetracking/TypeTrackerSpecific.qll
```diff
@@ -13,7 +13,18 @@ class TypeTrackingNode = DataFlowPublic::LocalSourceNode;
 
 predicate simpleLocalFlowStep = DataFlowPrivate::localFlowStepTypeTracker/2;
 
-predicate jumpStep = DataFlowPrivate::jumpStep/2;
+predicate jumpStep(Node nodeFrom, TypeTrackingNode nodeTo) {
```
asgerf, line 16: I don't think this should be a jump step, it should be local.

Please correct me if I'm wrong, but as I understand it, a "jump" step is one that resets the calling context (allowing flow out of a different caller than where it came in). Any reason to allow that here?

aibaars: I just (ab)used the `jumpStep` predicate because that has a `LevelStep` summary in `smallstepNoCall`. Perhaps it would be better to introduce a new predicate (`localStep` ?) in the shared TypeTracking library.

hvitved: Actually, it should perhaps be part of `localFlowStepTypeTracker` instead, since then it will work recursively (i.e., a method that returns one of its arguments via a call to another method that also does). This means such summarized steps will also be included in `LocalSourceNode::flowsTo`, which I guess we want?

Note that if we do this, we need to add a new disjunct to `isLocalSourceNode` which includes all nodes that correspond to a `ExprNodes::CallCfgNode`.

hvitved (changes requested):

ruby/ql/lib/codeql/ruby/typetracking/TypeTrackerSpecific.qll
aibaars: @hvitved  I added the new step to `simpleLocalFlowStep`. Somehow this changes the expected output back to the previous version. Not sure if that is desirable or not. I am not sure if/how this can be added to `localFlowStepTypeTracker` instead. I ran into some non-monotonic recursion errors.

ruby/ql/lib/codeql/ruby/typetracking/TypeTrackerSpecific.qll
```diff
@@ -11,8 +11,23 @@ class Node = DataFlowPublic::Node;
...
+/** Holds if there is a simple local flow step from `nodeFrom` to `nodeTo` */
+predicate simpleLocalFlowStep(Node nodeFrom, Node nodeTo) {
+  DataFlowPrivate::localFlowStepTypeTracker(nodeFrom, nodeTo)
+  or
+  // Summarize calls with flow directly from a parameter to a return.
+  exists(DataFlowPublic::ParameterNode param, DataFlowPrivate::ReturningNode returnNode |
+    DataFlowPrivate::LocalFlow::getParameterDefNode(param.getParameter())
+        .(TypeTrackingNode)
+        .flowsTo(returnNode) and
+    // Step from captured parameter to local call sites
```
hvitved, line 23: I don't understand this comment; it has nothing to do with variable capturing, right?

ruby/ql/lib/codeql/ruby/typetracking/TypeTrackerSpecific.qll
hvitved: I have added a commit that adds summarized local steps to `LocalSourceNode::flowsTo`, which in turn means that summarized steps will be constructed recursively. @asgerf : Does JS do this as well?

ruby/ql/lib/codeql/ruby/typetracking/TypeTrackerSpecific.qll
aibaars: That's just a left-over originating from https://github.com/github/codeql/blob/main/javascript/ql/lib/semmle/javascript/dataflow/internal/StepSummary.qll#L154

ruby/ql/lib/codeql/ruby/typetracking/TypeTrackerSpecific.qll
asgerf: No we don't contribute these to the `flowsTo` relation. It would make `flowsTo` depend on the call graph which is a pretty bad thing to do IMO.

In JS the dependency layers here roughly are:
- Local flow, including `flowsTo`
- Import resolution (establish flow between export and import). There are many module systems in JS, and their models may depend on local flow. Other languages with a less chaotic history of module systems may want to put this stage higher up.
- Call resolution and type-tracking steps (mutually recursive)
- API graphs

asgerf: When I said "local" earlier I guess I should have said "level".

hvitved: OK, I have given this some more thought, and I now understand why @aibaars used `jumpStep`, namely because it uses `LevelStep`. So unless we change the interface to the shared type tracking library, I think that is the best we can do for now. I will revert my changes and go back to using `jumpStep`.

hvitved: With https://github.com/github/codeql/pull/8317 we should change it to using `levelStep` instead.

aibaars: @hvitved Shall we close this pull request, or rebase it onto your work in https://github.com/github/codeql/pull/8317 ?

hvitved: Let's rebase once that PR is merged.

hvitved: @aibaars : That PR has now been merged, so let's rebase this PR and use the newly introduced `levelStep`.

hvitved (approved):
LGTM, but let's wait for DCA to confirm that performance is not worsened.

PR #8306, opened 2022-03-01
Ruby: add rb/clear-text-storage-sensitive-data query

alexrford:


ruby/ql/lib/codeql/ruby/security/internal/CleartextSources.qll
```diff
@@ -0,0 +1,273 @@
...
+        exists(CfgNodes::ExprNodes::PairCfgNode p |
+          this.asExpr() = lit and p = lit.getAKeyValuePair()
+        |
+          p.getKey().getConstantValue().getStringOrSymbol() = name and
+          p.getValue() = val.asExpr()
+        )
+      )
+    }
+
+    override string describe() { result = "an write to " + name }
```
hmac, line 212:
```suggestion
    override string describe() { result = "a write to " + name }
```

ruby/ql/src/queries/security/cwe-312/CleartextStorage.ql
```diff
@@ -0,0 +1,24 @@
...
+
+import ruby
+import codeql.ruby.security.CleartextStorageQuery
+import codeql.ruby.DataFlow
+import DataFlow::PathGraph
+
+from Configuration config, DataFlow::PathNode source, DataFlow::PathNode sink
+where config.hasFlowPath(source, sink)
+select sink.getNode(), source, sink, "Sensitive data returned by $@ is stored here.",
+  source.getNode(), source.getNode().(Source).describe()
```
hmac, line 24: Looking at some of the results in the DCA evaluation, I wonder if we should report these alerts at the source location rather than the sink location. There seem to be a lot of cases in `diaspora/diaspora` where a password value flows from a test to a call to `User.create` or similar. I worry that, since the alert is located at the sink, in `models/user.rb`, any code-scanning filtering to remove alerts in test files will let it pass through.

ruby/ql/test/query-tests/security/cwe-312/app/controllers/users_controller.rb
```diff
@@ -0,0 +1,66 @@
...
+  def fileWrites
+    new_password = "0157af7c38cbdd24f1616de4e5321861"
+
+    # BAD: plaintext password stored to disk
+    IO.write("foo.txt", "password: #{new_password}\n")
+
+    # BAD: plaintext password stored to disk
+    File.new("bar.txt", "a").puts("password: #{new_password}")
+  end
+end
```
hmac, line 66: There's a FP in the evaluation that is this:
```rb
random_password = SecureRandom.hex(20)
self.password = random_password
```

I don't know if this is due to over-sensitivity in `CleartextSources` or something else, but I bet this is a common pattern and it would be good to not flag it.

hmac (commented):
This looks good - just a few minor comments and a couple of potential improvements we could make.

ruby/ql/src/queries/security/cwe-312/CleartextStorage.ql
alexrford: This seems like a good idea overall. The case of a sink in test code and source in production code seems far less likely to occur in practice.

ruby/ql/test/query-tests/security/cwe-312/app/controllers/users_controller.rb
alexrford: I guess that the intention in the flagged case here is to set the value to a random hash value where the password that hashes to it is unknown, in which case I agree that it's a FP.

My first approximation to fixing this would is to augment to `notSensitiveRegexp()` expression to include `random` as a "safe" term. I have some slight concerns that there could be cases where a random plaintext password is generated and assigned to a field (e.g. for initial signup or password reset) - but it seems unlikely that the field that it's written to would work for both hashed and unhashed passwords. Consequently, in this theoretical case I'd expect there to be the same vulnerability when setting a real password that we could hopefully flag up.

PR #8308, opened 2022-03-02
Java: Add query to detect clickjacking

ahmed-farid-dev:
Clickjacking is an attack that tricks a user into clicking a webpage element which is invisible or disguised as another element. This can cause users to unwittingly download malware, visit malicious web pages, provide credentials or sensitive information, transfer money, or purchase products online.


smowton:
Not as yet, it's in the queue for the security lab to make an initial assessment

java/ql/src/experimental/Security/CWE/CWE-451/clickjacking.qhelp
```diff
@@ -0,0 +1,45 @@
...
+unintentionally clicking a button or link on the vulnerable site.
+</p>
+
+</overview>
+<recommendation>
+
+<p>
+Set the <code>X-Frame-Options</code> HTTP header to <code>DENY</code>, to instruct web browsers to
+block attempts to load the site in a frame. Alternatively, if framing is needed in certain
+circumstances, specify <code>ALLOW FROM: ...</code> to limit the ability
```
kevinbackhouse, line 20: `ALLOW FROM` is obsolete according to the Mozilla [documentation](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Frame-Options).

java/ql/test/experimental/query-tests/security/Clickjacking/Test.java
```diff
@@ -0,0 +1,14 @@
...
+import javax.servlet.ServletResponse;
+
+public class Test {
+
+   public void safeXframeOption(HttpServletResponse response) {
+        response.addHeader("X-Frame-Options", "DENY");  //GOOD
+    }
+
+    public void unsafeXframeOption(HttpServletResponse res) {
+        res.setHeader("X-Frame-Options", "SAMEORIGIN"); //BAD
```
kevinbackhouse, line 11: Why is this bad? I couldn't find a recommendation that you shouldn't use it in either the [OWASP](https://cheatsheetseries.owasp.org/cheatsheets/Clickjacking_Defense_Cheat_Sheet.html) or [Mozilla](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Frame-Options) documentation.

java/ql/src/experimental/Security/CWE/CWE-451/clickjacking.ql
```diff
@@ -0,0 +1,38 @@
...
+    header.getMethod() instanceof ResponseAddHeaderMethod
+  ) and
+  header.getArgument(0).(CompileTimeConstantExpr).getStringValue().toLowerCase() = "x-frame-options" and
+  header.getArgument(1).(CompileTimeConstantExpr).getStringValue().toLowerCase() = "sameorigin"
+}
+
+from MethodAccess call
+where
+  not hasCodeXFrameOptions(call) and
+  hasCodeXFramecheck(call)
```
kevinbackhouse, line 37: This doesn't look right to me. `hasCodeXFrameOptions` is true if arg 0 is `"x-frame-options"`. `hasCodeXFramecheck` is true if arg 0 is `"x-frame-options"` and arg 1 is `"sameorigin"`. So surely this query never has any results?

java/ql/src/experimental/Security/CWE/CWE-451/clickjacking.ql
```diff
@@ -0,0 +1,27 @@
...
+  (
+    header.getMethod() instanceof ResponseSetHeaderMethod or
+    header.getMethod() instanceof ResponseAddHeaderMethod
+  ) and
+  header.getArgument(0).(CompileTimeConstantExpr).getStringValue().toLowerCase() = "x-frame-options" 
+}
+
+from MethodAccess call
+where not hasCodeXFrameOptions(call) 
+select call, "Configuration missing the X-Frame-Options setting."
```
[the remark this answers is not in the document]
kevinbackhouse: @ahmed-farid-dev: Here's an example of running this query on Apache Hadoop: it has over 1 million results. I'm afraid we can only accept queries that have a low number of false positive results, so you will need to improve the query if you want us to accept it.

https://lgtm.com/query/8531091682539568277/

PR #8309, opened 2022-03-02
C++: Fix the join order in two places in the IR dataflow library

jketema:
Not having these fixed caused problems when updating the database stats file.


MathiasVP (dismissed):
LGTM!

jketema:
@MathiasVP I fixed a typo in the commit message that was annoying me, otherwise no changes.

MathiasVP (approved):

PR #8310, opened 2022-03-02
C++: Update the DB scheme stats file

jketema:


MathiasVP:
This is quite a scary PR: 39,547 additions and 39,130 deletions. But we've verified that it behaves as expected on DCA. I'm wondering whether we should do a full test on all of LGTM, though before merging this, though.

jketema:
> This is quite a scary PR: 39,547 additions and 39,130 deletions.

Because the formatting and order in the stats file changed. Also when I looked at the actual contents a lot if the numbers changed, because of the discussed difference between old and new set of projects we use to generate the stats file.

MathiasVP:
Oh yeah, sorry. It wasn't meant as a criticism! I wasn't expecting the diff to be smaller than it is, at all.

jketema:
> Oh yeah, sorry. It wasn't meant as a criticism! I wasn't expecting the diff to be smaller than it is, at all.

I did not see it as criticism at all, not worries. I was just stating the obvious reason why this happened.

MathiasVP:
I'll do a test on all of LGTM for this PR.

MathiasVP (approved):
Everything looks fine on LGTM 🎉!

PR #8318, opened 2022-03-03
C++: New query cpp/potential-system-data-exposure

geoffw0:
This is a new query `cpp/potential-system-data-exposure` closely related to the existing `cpp/system-data-exposure`, both providing coverage of [CWE-497](https://cwe.mitre.org/data/definitions/497.html).  Roughly speaking, each result of the old `cpp/system-data-exposure` is now either:
 - still a result of `cpp/system-data-exposure`, if the information is exposed to a remote sink (this change was made in a previous PR).
 - a result of this new query `cpp/potential-system-data-exposure`, if the above does not hold but the information exposed is considered especially sensitive (e.g. a password being output to `stdout`).
 - not a result any more (the old query used to be very noisy).

This part is likely to remain `@precision medium`, because some projects will consider standard output and similar channels to be safe from attackers.  `cpp/system-data-exposure` is likely to be upgraded to `@precision high`, as most of those results will carry some real risk.

Note that the diff is larger than the true number of changes because:
 - I've moved some existing QL code into a `SystemData.qll` library, shared by both queries.
 - I've migrated most of our internal tests for CWE-497 to the public repo (there's an internal PR removing them there).

---

LGTM results: https://lgtm.com/query/7993974152184061577/
 - results look reasonable to me
 - there are a few cases being picked up where we read into one field of a struct and output from another; this isn't ideal, but in most cases the result is reasonable regardless.

I've also started a DCA run.


MathiasVP (commented):
This looks great! Here are my initial comments.

geoffw0:
DCA shows:
 - 9 results for the new query; all are in `kamailio__kamailio`, matching results we'd already found via LGTM.
 - some performance wobbles, including analysis 3% slower overall; I don't believe this is significant.
 - `cpp/potential-system-data-exposure` is probably in the slower half of queries for performance, but far from the slowest, I don't think its causing expensive re-evaluations or anything like that.

cpp/ql/src/Security/CWE/CWE-497/PotentiallyExposedSystemData.ql
```diff
@@ -0,0 +1,219 @@
...
+/**
+ * Data originating from a call to `mysql_get_client_info()`.
+ */
+class SQLClientInfo extends SystemData {
+  SQLClientInfo() { this.(FunctionCall).getTarget().hasName("mysql_get_client_info") }
+
+  override Expr getAnExpr() { result = this }
+}
+
+private predicate sqlConnectInfo(FunctionCall source, VariableAccess use) {
```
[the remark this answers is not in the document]
geoffw0: I agree, this is just left over from the old design (assuming it was even justified there).

Updated to `Expr`.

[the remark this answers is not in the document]
geoffw0: The other two changes deserve a little bit of care, to be sure we do [still] catch real cases.  I'll have to make them next week.

cpp/ql/src/Security/CWE/CWE-497/SystemData.qll
```diff
@@ -236,15 +236,26 @@ class LogonUser extends SystemData {
   override predicate isSensitive() { any() }
 }
 
-private predicate regQuery(FunctionCall source, VariableAccess use) {
+private newtype TRegQueryParameter =
+  TSubKeyName(Expr e) or
```
MathiasVP, line 240: Could we restrict `e` to be only those arguments that are actually keys? Something like:
```ql
e = any(Call call | call.getTarget.hasName(/* ... */)).getArgument(1)
```
and similarly for the other branches?

cpp/ql/src/Security/CWE/CWE-497/PotentiallyExposedSystemData.ql
```diff
@@ -0,0 +1,219 @@
...
+    regQuery(this, result) and
+    this.(FunctionCall).getAnArgument().getValue().toLowerCase().regexpMatch(".*(pass|token|key).*")
+  }
+}
+
+class PotentiallyExposedSystemDataConfiguration extends TaintTracking::Configuration {
+  PotentiallyExposedSystemDataConfiguration() { this = "PotentiallyExposedSystemDataConfiguration" }
+
+  override predicate isSource(DataFlow::Node source) {
+    source.asExpr() = any(SystemData sd).getAnExpr()
```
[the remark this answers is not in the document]
geoffw0: I see your point, and I could improve it by having the `SystemData.qll` library return `DataFlow::Node`s rather than `Expr`s. 
 I've failed to find an example where it makes a [positive] difference though - perhaps because we're using taint tracking so it flows from the pre- to post-update nodes anyway?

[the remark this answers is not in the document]
geoffw0: (might clean up and push the change anyway though...)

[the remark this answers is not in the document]
geoffw0: Done.

cpp/ql/src/Security/CWE/CWE-497/SystemData.qll
geoffw0: > That's a lot of unnecessary data!

Do you have any reason to believe the full set of values of the `newtype` would be evaluated?  I would (perhaps naively) expect it not to be, but rather the 'characteristic predicate' (in this case rather trivial) of this type would be added to the various `RegQuery` classes.

I must admit I haven't experimented in this area very much.  It appears you can't create an 'infinite' `newtype` such as:
```
newtype MaybeInt =
  TInt(int i) or
  TNone()
```
which perhaps supports your interpretation.

MathiasVP: > Do you have any reason to believe the full set of values of the newtype would be evaluated? I would (perhaps naively) expect it not to be, but rather the 'characteristic predicate' (in this case rather trivial) of this type would be added to the various RegQuery classes.

The semantics of `newtype` branches follows the exact same steps as with predicates. `TSubKeyName(Expr e)` is the same as `TSubKeyName(Expr e) { any() }` and contains an `Expr e` for any expression that satisfy the branch's body (which in this case is `any()`). You can see an example of this when you run it:

```cpp
import cpp

newtype TRegQueryParameter =
  TSubKeyName(Expr e) or
  TValueName(Expr e) or
  TReturnData(Expr e)

from int n
where n = strictcount(TRegQueryParameter e)
select n
```
if you look in the log you'll see something like:
```ql
Tuple counts for num#quick-query::TSubKeyName#ff/2@2d8bc2co after 999ms:
  13321297 ~3%     {2} r1 = INVOKE HIGHER-ORDER RELATION construct<quick-query::TRegQueryParameter,0> ON <dom#quick-query::TReturnData#f>
                    return r1
Tuple counts for num#quick-query::TValueName#ff/2@1b66508b after 2s:
  13321297 ~0%     {2} r1 = INVOKE HIGHER-ORDER RELATION construct<quick-query::TRegQueryParameter,1> ON <dom#quick-query::TReturnData#f>
                    return r1
 uple counts for num#quick-query::TReturnData#ff/2@b5e6e6qk after 4s:
  13321297 ~0%     {2} r1 = INVOKE HIGHER-ORDER RELATION construct<quick-query::TRegQueryParameter,2> ON <dom#quick-query::TReturnData#f>
                    return r1
Starting to evaluate predicate #select#query#f#count_range/1@131d37qk
Tuple counts for #select#query#f#count_range/1@131d37qk after 1.2s:
  13321297 ~0%     {1} r1 = SCAN num#quick-query::TSubKeyName#ff OUTPUT In.1 'arg0'
  
  13321297 ~0%     {1} r2 = SCAN num#quick-query::TValueName#ff OUTPUT In.1 'arg0'
  
  13321297 ~2%     {1} r3 = SCAN num#quick-query::TReturnData#ff OUTPUT In.1 'arg0'
  
  26642594 ~0%     {1} r4 = r2 UNION r3
  39963891 ~0%     {1} r5 = r1 UNION r4
                    return r5
```

i.e., each branch will have `13321297` elements.

> I must admit I haven't experimented in this area very much. It appears you can't create an 'infinite' newtype such as:
> ```ql
> newtype MaybeInt =
>   TInt(int i) or
>   TNone()
> ```
> which perhaps supports your interpretation.

Indeed, this would create an infinite branch which is not currently possible.

geoffw0: ... I've just found and pushed an embarrassingly simple solution.  It isn't _quite_ as clean as before, but its close and still prevents duplication of data in the QL.

MathiasVP: > ... yep, best I can tell it really is constructed unless we get lucky with the optimizer.

Yeah, I don't know if the optimizer could push in some magic in these branches. I'm not sure I've ever seen that before, though.

> I'll see if I can find a better way to do this.

Great! Do take a look at https://github.com/MathiasVP/ql/commit/b853824aa16b119a3909ed86036695f2710a769f for some inspiration if that helps!

geoffw0:
I think this is PR is ready, apart from merge conflicts and submodule stuff which I'll address tomorrow.

cpp/ql/src/Security/CWE/CWE-497/SystemData.qll
```diff
@@ -252,14 +252,14 @@ class LogonUser extends SystemData {
  * the `regQuery` predicate concisely.
  */
 private newtype TRegQueryParameter =
-  TSubKeyName(Expr e) or
-  TValueName(Expr e) or
-  TReturnData(Expr e)
+  TSubKeyName() or
```
MathiasVP, line 255: Ah. Excellent!

geoffw0: Yeah, I don't know why it took me so long to realize I could do it this way!

Thanks for the help and discussion.

MathiasVP: No problem!

MathiasVP:
LGTM! Do we want a docs review of these changes?

MathiasVP (dismissed):

geoffw0:
> Do we want a docs review of these changes?

Yeah, for the two `.qhelp` files.

@github/docs-content-codeql

skedwards88:
Docs first responder here :wave: I've added this PR to our board for writer review.

mchammer01:
Not sure I'll get to reviewing this today. If not, I'll look at this tomorrow 👍🏻

cpp/ql/src/Security/CWE/CWE-497/ExposedSystemData.ql
```diff
@@ -13,284 +13,15 @@
  */
```
mchammer01, line 13: `adversary` -> `malicious user` ?

cpp/ql/src/Security/CWE/CWE-497/PotentiallyExposedSystemData.qhelp
```diff
@@ -0,0 +1,28 @@
+<!DOCTYPE qhelp PUBLIC
+  "-//Semmle//qhelp//EN"
+  "qhelp.dtd">
+<qhelp>
+<overview>
+<p>Exposing system data or debugging information may help an adversary to learn about the system and form an attack plan.  An attacker can use error messages that reveal technologies, operating systems, and product versions to tune their attack against known vulnerabilities in these technologies.</p>
```
mchammer01, line 6: I'd make the same changes to to this qhelp file that I suggested for the other qhelp file `adversary` -> `malicious user`, remove the `to` in `may help an adversary to learn about`, and possibly find a synonym for `technologies`.
I am a bit confused because the content for both seems to be exactly the same 😕

mchammer01 (approved):
@geoffw0 - LGTM ✨ 
Approving this so that I don't block this work.
I've left a few suggestions and comments for your consideration. Feel free to ignore them if you don't agree 🙂 

Also, I didn't see a change note about the new query, but I'm not sure whether all CodeQL teams are adding change notes for new queries.

MathiasVP:
> Also, I didn't see a change note about the new query, but I'm not sure whether all CodeQL teams are adding change notes for new queries.

All language teams should add a change-note for a new query. It's also done in this PR. See [this commit](https://github.com/github/codeql/pull/8318/commits/9e193f624c740b99181a5efbfe14214ea6e45d9d).

cpp/ql/src/Security/CWE/CWE-497/ExposedSystemData.qhelp
```diff
@@ -5,11 +5,11 @@
 <overview>
 <p>Exposing system data or debugging information may help an adversary to learn about the system and form an attack plan.  An attacker can use error messages that reveal technologies, operating systems, and product versions to tune their attack against known vulnerabilities in these technologies.</p>
```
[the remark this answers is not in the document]
geoffw0: Done (I went with "known vulnerabilities in the software.").

cpp/ql/src/Security/CWE/CWE-497/PotentiallyExposedSystemData.qhelp
geoffw0: Done.

We've essentially forked one query into two variations here, and ended up with two versions of the `qhelp` as well.  There are slight differences reflecting the difference in focus between the two queries, most of the text remains identical.

geoffw0:
I think I've addressed all concerns / suggestions now.

MathiasVP (approved):
LGTM!

cpp/ql/src/Security/CWE/CWE-497/PotentiallyExposedSystemData.qhelp
mchammer01: Thanks for explaining, it now makes more sense to me 🙂

mchammer01:
@MathiasVP - thanks for pointing the change note to me, I didn't see it at all 😞 
LGTM ✨

PR #8321, opened 2022-03-03
C++: More TPs from `cpp/using-expired-stack-address`

MathiasVP:
This PR changes `cpp/using-expired-stack-address` so that the must-flow in the query follows flow into predicates. This should allow us to catch more true positive results (in particular, it fixes a missing result in our tests). I suggest reviewing the PR on a commit-by-commit basis.

- 9df923a7c82c81d4a9a6d87889936325f693ddf8 fixes the missing must-flow
- bf10456bf53170514767516bd8c86c669e9f4486 adds a path explanation to the query. This was a huge help when debugging the query results on LGTM, and I expect users will have the same opinion.

<s>I've marked this as a draft while I run DCA.</s> DCA looks good.

### DIFF
--- a/cpp/ql/src/Likely Bugs/Memory Management/UsingExpiredStackAddress.ql
+++ b/cpp/ql/src/Likely Bugs/Memory Management/UsingExpiredStackAddress.ql
@@ -178,13 +178,10 @@ predicate blockStoresToAddress(
   globalAddress = globalAddress(store.getDestinationAddress())
 }
 
-predicate blockLoadsFromAddress(
-  IRBlock block, int index, LoadInstruction load, TGlobalAddress globalAddress
-) {
-  block.getInstruction(index) = load and
-  globalAddress = globalAddress(load.getSourceAddress())
-}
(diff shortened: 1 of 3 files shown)


geoffw0 (approved):
Happy with the new results.  Its definitely nice to have paths from this query as well (e.g. https://lgtm.com/query/4938718690567093881/).  :+1:

The QL is difficult to review though - feels like in an ideal world most of it would be in a separate library with a clear spec and tests - but I suppose there's only one place we'd be using it at present.

MathiasVP:
> The QL is difficult to review though - feels like in an ideal world most of it would be in a separate library with a clear spec and tests - but I suppose there's only one place we'd be using it at present.

Indeed. There is _some_ similarity between this new query, the `cpp/return-stack-allocated-memory` query, and the `unsafe-use-of-this` (i.e., they all use their own version of a must-flow relation). I'll create an issue that describes this in more detail. Hopefully, we can merge some if this code.

PR #8325, opened 2022-03-03
[Java] Add CharacterLiteral to CompileTimeConstantExpr.getStringValue

JLLeitschuh:
Simplifies end-user logic when looking for charater literals.

### DIFF
--- a/java/ql/test/library-tests/constants/constants/Constants.java
+++ b/java/ql/test/library-tests/constants/constants/Constants.java
@@ -18,6 +18,7 @@ class Constants {
 		int paren = (12);
 		String string = "a string";
 		int ternary = (3 < 5) ? 1 : 2;
+		char charLiteral = 'a';
 		
 		return;
 	}
--- a/java/ql/test/library-tests/constants/constants/Values.java
+++ b/java/ql/test/library-tests/constants/constants/Values.java
@@ -88,7 +88,7 @@ class Values {
         int var_local = final_local; //42
         int var_param = notConstant; //Not constant
         int var_nonfinald_local = var_field; //Not constant
-        String concatinatedString = "a" + "b"; //ab
-        String concatinatedChar = "ab" + 'c'; //abc
+        String concatenatedString = "a" + "b"; //ab
+        String concatenatedChar = "ab" + 'c'; //abc
     }
 }
--- a/java/ql/test/library-tests/constants/constants/Values.java
+++ b/java/ql/test/library-tests/constants/constants/Values.java
@@ -88,5 +88,7 @@ class Values {
         int var_local = final_local; //42
         int var_param = notConstant; //Not constant
         int var_nonfinald_local = var_field; //Not constant
+        String concatinatedString = "a" + "b"; //ab
+        String concatinatedChar = "ab" + 'c'; //abc
     }
 }
--- a/java/ql/test/library-tests/constants/getStringValue.ql
+++ b/java/ql/test/library-tests/constants/getStringValue.ql
@@ -1,0 +1,9 @@
+import semmle.code.java.Variable
+
+from Variable v, Expr init, RefType enclosing, string constant
+where
+  v.getInitializer() = init and
+  init.getEnclosingCallable().getDeclaringType() = enclosing and
+  enclosing.hasQualifiedName("constants", "Values") and
+  constant = init.(CompileTimeConstantExpr).getStringValue()
+select init, constant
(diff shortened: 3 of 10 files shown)


java/ql/test/library-tests/constants/constants/Constants.java
```diff
@@ -18,6 +18,7 @@ void constants(final int notConstant) {
 		int paren = (12);
 		String string = "a string";
 		int ternary = (3 < 5) ? 1 : 2;
+		char charLiteral = 'a';
 		
 		return;
 	}
```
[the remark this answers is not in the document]
JLLeitschuh: I don't want to fix this as a part of this PR 😆  someone might want to make a sweep though, just my 10c

java/ql/test/library-tests/constants/constants/Values.java
```diff
@@ -88,5 +88,7 @@ void values(final int notConstant) {
         int var_local = final_local; //42
         int var_param = notConstant; //Not constant
         int var_nonfinald_local = var_field; //Not constant
+        String concatinatedString = "a" + "b"; //ab
+        String concatinatedChar = "ab" + 'c'; //abc
```
atorralba, line 92:
```suggestion
        String concatenatedString = "a" + "b"; //ab
        String concatenatedChar = "ab" + 'c'; //abc
```

java/ql/test/library-tests/constants/getStringValue.ql
```diff
@@ -0,0 +1,9 @@
+import semmle.code.java.Variable
+
+from Variable v, Expr init, RefType enclosing, string constant
+where
+  v.getInitializer() = init and
+  init.getEnclosingCallable().getDeclaringType() = enclosing and
+  enclosing.hasQualifiedName("constants", "Values") and
+  constant = init.(CompileTimeConstantExpr).getStringValue()
```
atorralba, line 8:
```suggestion
from Variable v, CompileTimeConstantExpr init, RefType enclosing, string constant
where
  v.getInitializer() = init and
  init.getEnclosingCallable().getDeclaringType() = enclosing and
  enclosing.hasQualifiedName("constants", "Values") and
  constant = init.getStringValue()
```

atorralba (commented):
Hey @JLLeitschuh, thanks for your contribution. This mostly LGTM, I added a couple of inline comments.

smowton (approved):

JLLeitschuh:
This PR is done and can be merged whenever you're comfortable doing so. 😄

Marcono1234:
This change breaks string concatenation logic for `CompileTimeConstantExpr` and might also negatively affect custom user queries. Consider for example `'a' + 'b'`; with this change `getStringValue()` erroneously has `"ab"` as result, while the correct result is an `int` representing `((int) 'a') + ((int) 'b')` (unfortunately this does not seem to be covered by any tests).
Currently the logic in `CompileTimeConstantExpr` is also incomplete, but once it gets closer to the JLS, this would affect more cases, such as `'a' + 1`.
Similarly a user who is using `getStringValue()` in their query will likely expect that the expression is really of type `String`, and not of type `char` / `Character`, which is incompatible with `String`.

It looks like the reason why this change to `getStringValue()` was proposed, was to simplify a very specific case where a CodeQL expression represents a read of a field of type `char` or `String`, see https://github.com/github/codeql/pull/8032/files#r818222618.
Would it therefore make sense to revert the changes done to `getStringValue()`, and instead modify the code for that specific use case? To me that looks like a saner solution.

smowton:
Hm, sure, the concat case clearly needs fixing. Concat of string + char is nice to have though for example. @JLLeitschuh how about amending to have a separate getStringifiedValue, and use that to handle concatenation in the particular case where the resulting expression is of type String? Thus `'a'` wouldn't have a `getStringValue` (but would have a `getStringifiedValue`), but `"a" + 'b'` would have both.

`getStringifiedValue` is hard to implement in general of course since e.g. floating point numbers are likely to stringify differently to how the user wrote them in source (e.g. `final double d = 1.0e1` will stringify as `10.0`), but it would be acceptable in my view to give a partial implementation so long as that is clearly noted in `getStringifiedValue`'s qldoc.

Marcono1234:
> `getStringifiedValue` is hard to implement in general of course since e.g. floating point numbers are likely to stringify differently to how the user wrote them in source (e.g. `final double d = 1.0e1` will stringify as `10.0`), but it would be acceptable in my view to give a partial implementation so long as that is clearly noted in `getStringifiedValue`'s qldoc

To me that looks acceptable as well (if the qldoc describes that the value is converted to string), because in Java for example `"a" + 1.0e1` is `"a10.0"`, so this would not be that different from what one might expect.

JLLeitschuh:
Thanks for the feedback! Resolved here: https://github.com/github/codeql/pull/8360

PR #8332, opened 2022-03-04
Ruby: Use taint tracking instead of type tracking to define `regExpSource`

hvitved:
Follow-up to https://github.com/github/codeql/pull/8293.

### DIFF
--- a/ruby/ql/lib/codeql/ruby/security/performance/ParseRegExp.qll
+++ b/ruby/ql/lib/codeql/ruby/security/performance/ParseRegExp.qll
@@ -8,9 +8,9 @@
 private import codeql.ruby.ast.Literal as AST
 private import codeql.Locations
 private import codeql.ruby.DataFlow
-private import codeql.ruby.TaintTracking
-private import codeql.ruby.typetracking.TypeTracker
+private import codeql.ruby.controlflow.CfgNodes
 private import codeql.ruby.ApiGraphs
+private import codeql.ruby.dataflow.internal.tainttrackingforlibraries.TaintTrackingImpl
 
 /**
  * A `StringlikeLiteral` containing a regular expression term, that is, either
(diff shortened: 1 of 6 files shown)


aibaars (commented):
This looks fine to me. How does this approach compare to using a type tracker in terms of precision and efficiency?

hvitved:
> How does this approach compare to using a type tracker in terms of precision and efficiency?

Precision: The data-flow library has much higher precision, e.g. we are tracking contents (fields) and call contexts precisely.
Performance: While it may a bit slower, it doesn't matter much in practice (as witnessed by DCA). Unlike type-tracking, the data-flow library applies sophisticated pruning of nodes before tracking flow from a given source to a given node, which helps a lot on performance.

RasmusWL:
Since we discussed these two approaches in architecture meeting, I would be very interested to see whether you have any concrete examples where the new taint-tracking based modeling is able to give a result for `regExpSource` where the old type-tracking based approach wasn't. I couldn't see any changes to tests, or any new results in alerts/alert-meta in the performance test.

hmac:
As I understand it, we have to be careful to not re-use the same dataflow library multiple times (hence DataFlow2,3 etc.) If we want to use taint tracking for another library concept, will we have to duplicate this library again in case someone writes a query that uses both `regExpSource` and the other library concept? In other words, would this copy be better named something like `DataFlowImplForRegExpSource`, if we can't re-use it for anything else?

hvitved:
> As I understand it, we have to be careful to not re-use the same dataflow library multiple times (hence DataFlow2,3 etc.) If we want to use taint tracking for another library concept, will we have to duplicate this library again in case someone writes a query that uses both `regExpSource` and the other library concept? In other words, would this copy be better named something like `DataFlowImplForRegExpSource`, if we can't re-use it for anything else?

That is correct. There are cases where reusing the same copy of the data-flow library in multiple places in other libraries is OK (specifically when those two libraries are known to always be in scope at the same time), but in general it is safer to always just use a separate copy. However, in the not-too-distant future we will be able to get rid of all the copies using a parameterized module instead.

ruby/ql/lib/codeql/ruby/security/performance/ParseRegExp.qll
```diff
@@ -990,30 +990,42 @@ private predicate isInterpretedAsRegExp(DataFlow::Node source) {
...
+    exists(DataFlow::CallNode mce |
+      mce.getMethodName() = ["match", "match?"] and
+      node = mce.getReceiver() and
+      mce.getArgument(0).asExpr().getExpr() instanceof AST::RegExpLiteral
+    )
+  }
 }
 
 /**
  * Gets a node whose value may flow (inter-procedurally) to `re`, where it is interpreted
```
asgerf, line 1025:
```suggestion
 * Gets a string constant node whose value may flow (inter-procedurally) to `re`, where it is interpreted
```
It might be worth mentioning that this only finds string constants/literals now (which is fine).

ruby/ql/lib/codeql/ruby/security/performance/ParseRegExp.qll
```diff
@@ -990,30 +990,42 @@ private predicate isInterpretedAsRegExp(DataFlow::Node source) {
...
-  )
+private class RegExpConfiguration extends Configuration {
+  RegExpConfiguration() { this = "RegExpConfiguration" }
+
+  override predicate isSource(DataFlow::Node source) {
+    source.asExpr() =
+      any(ExprCfgNode e |
+        e.getConstantValue().isString(_) and
+        not e instanceof ExprNodes::VariableReadAccessCfgNode and
+        not e instanceof ExprNodes::ConstantReadAccessCfgNode
```
asgerf, line 1007: Is there a reason not simply to restrict this to string literals?

asgerf (commented):
Overall I approve of switching to taint-tracking here. In JS I believe it was a case of "all we have is a hammer" so we used type-tracking because we don't have multiple copies of the data-flow configuration.

ruby/ql/lib/codeql/ruby/security/performance/ParseRegExp.qll
hvitved: No, I guess not.

hvitved: Actually, I just realized that we need that above for cases where a regular expression is constructed as the concatenation of multiple strings.

nickrolfe (dismissed):
The changes in this PR look good, and I think it should be merged.

But I think there is more work to do in this area. What I want to be able to do is take a `DataFlow::Node` for some arbitrary expression (that our analysis says has a regexp value), and get the resulting root `RegExpTerm`. It's not clear to me how to do that. That's what motivated me to make https://github.com/github/codeql/pull/7985, but I'm not sure if that change is still needed after this PR and https://github.com/github/codeql/pull/8293.

aibaars:
> The changes in this PR look good, and I think it should be merged.
> 
> But I think there is more work to do in this area. What I want to be able to do is take a `DataFlow::Node` for some arbitrary expression (that our analysis says has a regexp value), and get the resulting root `RegExpTerm`. It's not clear to me how to do that. That's what motivated me to make #7985, but I'm not sure if that change is still needed after this PR and #8293.

Would that be solved by the `RegExpPatternSource` defined in #7917 ?

hvitved:
Rebased again to sync in latest `DataFlowImpl.qll`/`TaintTrackingImpl.qll` changes.

nickrolfe (approved):

PR #8333, opened 2022-03-04
JS: Use `TypeTracker::continue` when doing taint steps

hvitved:


hvitved:
@asgerf : It seems wrong to me to allow taint steps when data is stored in a field, but I can see that my change breaks a test. What do you think?

asgerf:
Hm, I agree it seems wrong, but based on that test it also seems useful 🤔. I guess it's unlikely that the property name would find a spurious match after using taint steps.

If anyone cares to investigate further, I'd say the next step is to add a meta-query for regexp string literals, and use that to evaluate how much of a difference it makes in practice.

PR #8345, opened 2022-03-06
Java: Add `MyBatis`' `Providers` sinks

jorgectf:
See https://mybatis.org/mybatis-3/apidocs/reference/org/apache/ibatis/jdbc/AbstractSQL.html

FYI: This PR is not tied to a Bug Bounty submission.


java/ql/test/experimental/query-tests/security/CWE-089/src/main/MybatisSqlInjectionService.java
```diff
@@ -51,6 +52,17 @@ public void bad9(String name) {
 		sqlInjectionMapper.bad9(hashMap);
 	}
 
+	public String bad10(String name) {
+		// TODO: make useful tests
```
jorgectf, line 56: How should proper tests for these sinks be like? Would a couple of real world examples (select, update, insert and delete) be enough?

jorgectf:
This PR has been inspired by a recent CTF challenge (D3CTF - ezsql) in which a GET parameter was injected into `AbstractSQL.WHERE(here)`. Our solution (cc: @TheGrandPew) involves achieving RCE through a JNDI lookup using OGNL (meaning there is a OGNL injection abusing [Dynamic SQL](https://mybatis.org/mybatis-3/dynamic-sql.html)). Annotations seem not to be vulnerable, but this PR's methods seem to.

Should [OgnlInjection.ql](/github/codeql/blob/main/java/ql/src/Security/CWE/CWE-917/OgnlInjection.ql) cover these sinks?

// cc: @pwntester

java/ql/test/experimental/query-tests/security/CWE-089/src/main/MybatisSqlInjectionService.java
atorralba: Ideally, you'd have (at least) a test line for each new sink being added.

atorralba:
> Should [OgnlInjection.ql](https://github.com/github/codeql/blob/main/java/ql/src/Security/CWE/CWE-917/OgnlInjection.ql) cover these sinks?

Judging by your explanation, these sinks are both SQLi and OGNL injection sinks. I'm not sure whether we have other cases of this kind of "dual" sinks in the codebase that we could get inspiration from, but one thing we could do is labeling these differently (so, instead of `sql`, something like `sql-ognl`) and adding them to both the SQLi and the OGNL injection queries.

java/ql/lib/semmle/code/java/frameworks/MyBatis.qll
```diff
@@ -20,7 +20,45 @@ private class SqlSinkCsv extends SinkModelCsv {
...
+        "org.apache.ibatis.jdbc;AbstractSQL;false;HAVING;(String);;Argument[0];sql",
+        "org.apache.ibatis.jdbc;AbstractSQL;false;GROUP_BY;(String[]);;Argument[0];sql",
+        "org.apache.ibatis.jdbc;AbstractSQL;false;GROUP_BY;(String);;Argument[0];sql",
+        "org.apache.ibatis.jdbc;AbstractSQL;false;WHERE;(String[]);;Argument[0];sql",
+        "org.apache.ibatis.jdbc;AbstractSQL;false;WHERE;(String);;Argument[0];sql",
+        "org.apache.ibatis.jdbc;AbstractSQL;false;OUTER_JOIN;(String[]);;Argument[0];sql",
+        "org.apache.ibatis.jdbc;AbstractSQL;false;OUTER_JOIN;(String);;Argument[0];sql",
+        "org.apache.ibatis.jdbc;AbstractSQL;false;RIGHT_OUTER_JOIN;(String[]);;Argument[0];sql",
+        "org.apache.ibatis.jdbc;AbstractSQL;false;RIGHT_OUTER_JOIN;(String);;Argument[0];sql",
+        "org.apache.ibatis.jdbc;AbstractSQL;false;LEFT_OUTER_JOIN;(String[]);;Argument[0];sql",
```
smowton, line 41: For array arguments use `ArrayElement of Argument[0]` for example.

atorralba:
> We should sink at the return statement of the query providers

Interesting. That means that we'd also need taint steps from the `AbstractSQL` methods' arguments to the returned `AbstractSQL` object.

pwntester:
> Interesting. That means that we'd also need taint steps from the AbstractSQL methods' arguments to the returned AbstractSQL object.

Si, AbstractSQL parece ser un helper builder para construir queries

jorgectf:
* Would this be interesting enough to add as sink? https://mybatis.org/mybatis-3/apidocs/org/apache/ibatis/annotations/CacheNamespace.html (`@Property`'s `value`)

jorgectf:
> Ideally, you'd have (at least) a test line for each new sink being added.

Tests are still WIP.

pwntester:
> I doubt so, but would this be interesting enough to add as a sink for ognli? https://mybatis.org/mybatis-3/apidocs/org/apache/ibatis/annotations/CacheNamespace.html (`@Property`'s `value`)


I dont think user controlled data can flow to an annotation value

java/ql/lib/semmle/code/java/frameworks/MyBatis.qll
```diff
@@ -102,3 +102,40 @@ class MyBatisSqlOperationAnnotationMethod extends Method {
...
+
+  class MyBatisInjectionSink extends DataFlow::Node {
+    MyBatisInjectionSink() {
+      exists(Annotation a, Method m, TypeLiteral type, Class c |
+        a.getType()
+            .hasQualifiedName("org.apache.ibatis.annotations",
+              ["Select", "Delete", "Insert", "Update"] + "Provider") and
+        type = a.getValue(["type", "value"]) and
+        c.hasMethod(m, type.getTypeName().getType()) and
+        m.hasName(a.getTarget().getName()) and
```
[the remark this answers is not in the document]
jorgectf: You are right, while `a.getTarget()` takes the annotated method and that one and the one the *Provider is referencing should be named the same, I guess they could not 👍

java/ql/lib/semmle/code/java/frameworks/MyBatis.qll
```diff
@@ -102,3 +102,40 @@ class MyBatisSqlOperationAnnotationMethod extends Method {
...
+        this.asExpr() = m.getBody().getAStmt().(ReturnStmt).getResult()
+      )
+    }
+  }
+
+  class MyBatisAbstractSQLStep extends Unit {
```
atorralba, line 123: I don't think `Abstract` is a good name here, because this class is not abstract. Also, instead of extending `Unit` and having to add this to each relevant query, I'd recommend one of the following:

* Just extend `AdditionalTaintStep`.
* Implement this as rows of `SummaryModelCsv` (argument to return value is a common step for a CSV summary).

java/ql/lib/semmle/code/java/frameworks/MyBatis.qll
```diff
@@ -102,3 +102,40 @@ class MyBatisSqlOperationAnnotationMethod extends Method {
...
+    predicate step(DataFlow::Node node1, DataFlow::Node node2) {
+      exists(MethodAccess ma |
+        ma.getMethod()
+            .getDeclaringType()
+            .hasQualifiedName("org.apache.ibatis.jdbc", ["AbstractSQL", "AbstractSQL<SQL>"]) and
```
atorralba, line 128: In case you decide to go the `AdditionalTaintStep` route: instead of having to write the type parameter here, you can just use `getSourceDeclaration`:

```suggestion
            .getDeclaringType()
            .getSourceDeclaration()
            .hasQualifiedName("org.apache.ibatis.jdbc", "AbstractSQL") and
```

java/ql/lib/semmle/code/java/frameworks/MyBatis.qll
```diff
@@ -102,3 +102,40 @@ class MyBatisSqlOperationAnnotationMethod extends Method {
 class TypeParam extends Interface {
   TypeParam() { this.hasQualifiedName("org.apache.ibatis.annotations", "Param") }
 }
+
+module ProviderInjection {
```
atorralba, line 106: What's the reason for creating this module?

java/ql/lib/semmle/code/java/frameworks/MyBatis.qll
```diff
@@ -102,3 +102,40 @@ class MyBatisSqlOperationAnnotationMethod extends Method {
...
+  private import semmle.code.java.dataflow.DataFlow
```
atorralba, line 107: Conventionally, imports go at the top of the file.

java/ql/lib/semmle/code/java/frameworks/MyBatis.qll
jorgectf: Importing just the needed stuff in OGNL and SQL Injection queries :)

java/ql/lib/semmle/code/java/frameworks/MyBatis.qll
jorgectf: > I don't think Abstract is a good name here

`Abstract` stands for the name of the class, but I agree this would mislead. How should the class name be then?

> Just extend AdditionalTaintStep.

Done 👍 

> Implement this as rows of SummaryModelCsv (argument to return value is a common step for a CSV summary)

Thank you! I will look into it :)

jorgectf:
Pending looking into refactoring using `SummaryModelCsv`.

See https://github.com/github/codeql/pull/8345#discussion_r823458641

java/ql/lib/semmle/code/java/frameworks/MyBatis.qll
```diff
@@ -102,3 +104,95 @@ class MyBatisSqlOperationAnnotationMethod extends Method {
...
+        a.getType() instanceof MyBatisProvider and
+        type = a.getValue(["type", "value"]) and
+        c.hasMethod(m, type.getTypeName().getType()) and
+        m.hasName(a.getValue("method").(StringLiteral).getValue()) and
+        this.asExpr() = m.getBody().getAStmt().(ReturnStmt).getResult()
+      )
+    }
+  }
+
+  class MyBatisAdditionalTaintStep extends TaintTracking::AdditionalTaintStep {
```
atorralba, line 143: I think it's easier if you just remove this class and make `MyBatisProviderStep` and `MyBatisAbstractSQLToStringStep` extend `AdditionalTaintStep`. You don't need to add them to the `step` predicate of the specific configurations, since `AdditionalTaintStep` subclasses are applied globally (i.e. to all configurations).

java/ql/lib/semmle/code/java/frameworks/MyBatis.qll
jorgectf: I believe my last comment is wrong, since the whole framework is already imported while using some other libraries like `ExternalFlow`, and so making a module would not impact the queries at all (?).

java/ql/lib/semmle/code/java/frameworks/MyBatis.qll
```diff
@@ -105,94 +105,88 @@ class TypeParam extends Interface {
...
-    abstract override predicate step(DataFlow::Node node1, DataFlow::Node node2);
+class MyBatisInjectionSink extends DataFlow::Node {
+  MyBatisInjectionSink() {
+    exists(Annotation a, Method m, TypeLiteral type, Class c |
+      a.getType() instanceof MyBatisProvider and
+      type = a.getValue(["type", "value"]) and
+      c.hasMethod(m, type.getTypeName().getType()) and
+      m.hasName(a.getValue("method").(StringLiteral).getValue()) and
+      this.asExpr() = m.getBody().getAStmt().(ReturnStmt).getResult()
+    )
```
atorralba, line 138: I think this can be simplified:

```suggestion
    exists(Annotation a, Method m |
      a.getType() instanceof MyBatisProvider and
      m.getDeclaringType() = a.getValue(["type", "value"]).(TypeLiteral).getTypeName().getType() and
      m.hasName(a.getValue("method").(StringLiteral).getValue()) and
      this.asExpr() = m.getBody().getAStmt().(ReturnStmt).getResult()
    )
```

java/ql/lib/semmle/code/java/frameworks/MyBatis.qll
```diff
@@ -105,94 +105,88 @@ class TypeParam extends Interface {
...
+    |
+      a.getType() instanceof MyBatisProvider and
+      annotatedMethod.getAnAnnotation() = a and
+      ma.getMethod() = annotatedMethod and
+      ma.getAnArgument() = n1.asExpr() and
+      type = a.getValue(["type", "value"]) and
+      providerMethod.hasName(a.getValue("method").(StringLiteral).getValue()) and
+      c.hasMethod(providerMethod, type.getTypeName().getType()) and
+      providerMethod.getAParameter() = n2.asParameter()
+    )
```
atorralba, line 156: Similarly. Also, you need to wire arguments with parameters based on the index, otherwise you are creating steps for each argument witch each parameter (arg0 -> param0, arg0 -> param1, arg1 -> param0, arg1 -> param1, and so on):

```suggestion
    exists(MethodAccess ma, Annotation a, Method providerMethod |
      exists(int i |
        ma.getArgument(i) = n1.asExpr() and
        providerMethod.getParameter(i) = n2.asParameter()
      )
    |
      a.getType() instanceof MyBatisProvider and      
      ma.getMethod().getAnAnnotation() = a and
      providerMethod.getDeclaringType() =
        a.getValue(["type", "value"]).(TypeLiteral).getTypeName().getType() and
      providerMethod.hasName(a.getValue("method").(StringLiteral).getValue())
    )
```

java/ql/lib/semmle/code/java/frameworks/MyBatis.qll
```diff
@@ -105,94 +105,88 @@ class TypeParam extends Interface {
...
-        annotatedMethod.getAnAnnotation() = a and
-        ma.getMethod() = annotatedMethod and
-        ma.getAnArgument() = n1.asExpr() and
-        type = a.getValue(["type", "value"]) and
-        providerMethod.hasName(a.getValue("method").(StringLiteral).getValue()) and
-        c.hasMethod(providerMethod, type.getTypeName().getType()) and
-        providerMethod.getAParameter() = n2.asParameter()
-      )
-    }
+private class MyBatisProviderStep extends TaintTracking::AdditionalTaintStep {
```
atorralba, line 142: Now that I think of it, this is actually an `AdditionalValueStep`, since `n1` and `n2` are the same value.

java/ql/lib/semmle/code/java/frameworks/MyBatis.qll
```diff
@@ -105,94 +105,88 @@ class TypeParam extends Interface {
...
+        "GROUP_BY", "WHERE", "OUTER_JOIN", "RIGHT_OUTER_JOIN", "LEFT_OUTER_JOIN", "INNER_JOIN",
+        "JOIN", "FROM", "DELETE_FROM", "SELECT_DISTINCT", "SELECT", "INTO_VALUES", "INTO_COLUMNS",
+        "VALUES", "INSERT_INTO", "SET", "UPDATE"
+      ]
   }
+}
 
-  class MyBatisAdditionalTaintStep extends TaintTracking::AdditionalTaintStep {
-    abstract override predicate step(DataFlow::Node node1, DataFlow::Node node2);
+class MyBatisInjectionSink extends DataFlow::Node {
```
atorralba, line 130: This will need QLDoc.

java/ql/lib/semmle/code/java/security/OgnlInjection.qll
```diff
@@ -3,6 +3,7 @@
 import java
 private import semmle.code.java.dataflow.DataFlow
 private import semmle.code.java.dataflow.ExternalFlow
+private import semmle.code.java.frameworks.MyBatis
```
atorralba, line 6: `MyBatis` is already imported in `ExternaFlow`, which means it's already in the global scope. You can remove this.

```suggestion
```

java/ql/lib/semmle/code/java/security/QueryInjection.qll
```diff
@@ -3,6 +3,7 @@
 import java
 import semmle.code.java.dataflow.DataFlow
 import semmle.code.java.frameworks.javaee.Persistence
+private import semmle.code.java.frameworks.MyBatis
```
atorralba, line 6: Same as above.

```suggestion
```

java/ql/lib/semmle/code/java/frameworks/MyBatis.qll
atorralba: Yes, I don't think the module is necessary.

java/ql/lib/semmle/code/java/security/OgnlInjection.qll
jorgectf: `MyBatisInjectionSink` can't be found if I delete that import. It is privately imported in 
https://github.com/github/codeql/blob/main/java/ql/lib/semmle/code/java/dataflow/ExternalFlow.qll#L137, but I guess since it is private, we can't access it in this context.

java/ql/lib/semmle/code/java/frameworks/MyBatis.qll
jorgectf: You are right! Thank you :)

java/ql/lib/semmle/code/java/frameworks/MyBatis.qll
jorgectf: Done https://github.com/github/codeql/pull/8345/commits/c683b48af714814ddf44c80be1a0f72b6ee5b5b8

java/ql/lib/semmle/code/java/frameworks/MyBatis.qll
jorgectf: Done https://github.com/github/codeql/pull/8345/commits/8482c019595595e407f713e9f121f6514d336ce4

java/ql/lib/semmle/code/java/frameworks/MyBatis.qll
```diff
@@ -102,3 +104,88 @@ class MyBatisSqlOperationAnnotationMethod extends Method {
...
+  }
+}
+
+private class MyBatisAbstractSQLMethodsStep extends SummaryModelCsv {
+  override predicate row(string row) {
+    row =
+      [
+        "org.apache.ibatis.jdbc;AbstractSQL;true;" + any(MyBatisAbstractSQLMethodNames m) +
+          ";;;Argument[0..1];ReturnValue;taint"
+      ]
```
jorgectf, line 177: Is this too wide? I think adding each method would be a bit noisy, we would loose correlation with `MyBatisAbstractSQLMethodNames`, but CSV's `input` field would be more accurate.

jorgectf:
Regardless pending comments (https://github.com/github/codeql/pull/8345#discussion_r826076472 https://github.com/github/codeql/pull/8345#discussion_r826076783 https://github.com/github/codeql/pull/8345#discussion_r826373098), I think this PR is ready for a final polishing review.

java/ql/lib/semmle/code/java/security/OgnlInjection.qll
atorralba: You're absolutely right. I was only thinking about the taint steps, but if you need to directly use a class from `MyBatis` here (in this case, `MyBatisInjectionSink`), you need the import indeed.

java/ql/lib/semmle/code/java/frameworks/MyBatis.qll
atorralba: I'd favor being explicit here and wiring only the exact arguments we know propagate taint, even if it means repeating the method names.

If you wanted, you could always do that mapping in `MyBatisAbstractSQLMethodNames` (also, I think this should be singular) and create a predicate to obtain the tainted arguments depending on the method name, so something like:

```ql
private class MyBatisAbstractSQLMethodName extends string {
  string taintedArgs;

  MyBatisAbstractSQLMethodName() {
    // I'm making this up, I didn't check which method propagate taint from which arguments
    this in ["SELECT", "SELECT_DISTINCT"] and taintedArgs = "Argument[0]"
    or
    this in ["VALUES", "INSERT_INTO", "SET", "UPDATE"] and taintedArgs = "Argument[0..1]"
    // or ...
  }

  /** Gets the tainted arguments for this method name. */
  string getTaintedArgs() { result = taintedArgs }
}

```

and then:

```ql
private class MyBatisAbstractSQLMethodsStep extends SummaryModelCsv {
    exists(MyBatisAbstractSQLMethodName m |
      row =
        "org.apache.ibatis.jdbc;AbstractSQL;true;" + m + ";;;" + m.getTaintedArgs() +
          ";ReturnValue;taint"
    )
  }
}
```

This impacts readability though, so there isn't a perfect solution.

atorralba: It could be even more natural if `MyBatisAbstractSQLMethodName` becomes `MyBatisAbstractSQLMethod`:

```ql
private class MyBatisAbstractSQLMethod extends Method {
  string taintedArgs;

  MyBatisAbstractSQLMethod() {
    this.getDeclaringType().getSourceDeclaration() instanceof MyBatisAbstractSQL and
    (
      // I'm making this up, I didn't check which method propagate taint from which arguments
      this.hasName(["SELECT", "SELECT_DISTINCT"]) and taintedArgs = "Argument[0]"
      or
      this.hasName(["VALUES", "INSERT_INTO", "SET", "UPDATE"]) and taintedArgs = "Argument[0..1]"
      // or ...
    )
  }

  string getTaintedArgs() { result = taintedArgs }
}

// ...

private class MyBatisAbstractSQLMethodsStep extends SummaryModelCsv {
    exists(MyBatisAbstractSQLMethod m |
      row =
        "org.apache.ibatis.jdbc;AbstractSQL;true;" + m.getName() + ";;;" + m.getTaintedArgs() +
          ";ReturnValue;taint"
    )
  }
}

// ...

private class MyBatisAbstractSQLAnonymousClassStep extends TaintTracking::AdditionalTaintStep {
  override predicate step(DataFlow::Node node1, DataFlow::Node node2) {
    exists(MethodAccess ma, ClassInstanceExpr c |
      ma.getMethod() instanceof MyBatisAbstractSQLMethod and
      c.getAnonymousClass().getACallable() = ma.getCaller() and
      node1.asExpr() = ma and
      node2.asExpr() = c
    )
  }
}
```

jorgectf: This is really cool!

jorgectf: Done in https://github.com/github/codeql/pull/8345/commits/9aa440e5b6e9fd47d9ca12c749146be3a522deca

atorralba:
Almost there. This will need a change note: https://github.com/github/codeql/blob/main/docs/change-notes.md

atorralba:
The test `java/ql/test/experimental/query-tests/security/CWE-089/src/main/MyBatisAnnotationSqlInjection.qlref` is failing @jorgectf.

jorgectf:
> The test `java/ql/test/experimental/query-tests/security/CWE-089/src/main/MyBatisAnnotationSqlInjection.qlref` is failing @jorgectf.

Fixed 👍

java/ql/lib/semmle/code/java/frameworks/MyBatis.qll
```diff
@@ -102,3 +104,97 @@ class MyBatisSqlOperationAnnotationMethod extends Method {
...
+  }
+}
+
+private class MyBatisAbstractSqlMethodsStep extends SummaryModelCsv {
+  override predicate row(string row) {
+    exists(MyBatisAbstractSqlMethod m |
+      row =
+        "org.apache.ibatis.jdbc;AbstractSQL;true;" + m.getName() + ";(" + m.getCsvSignature() +
+          ");;" + m.getTaintedArgs() + ";Argument[-1];taint"
+    )
```
aschackmull, line 198: This isn't really the way we want to add csv-rows to our models-as-data coverage. As-is this is an awkward mix of traditional QL-modelling and a csv string bits.
We would like our models-as-data to be database independent, such that these rows don't depend on the existence of the `MyBatisAbstractSql` type.

aschackmull (changes requested):

java/ql/lib/semmle/code/java/frameworks/MyBatis.qll
jorgectf: Done in https://github.com/github/codeql/pull/8345/commits/9e1b98e5a4da22c60e36871dc2486d16f91555df. Would you prefer having the rows like https://github.com/github/codeql/commit/f9ae1efa2c3848a4ddf711c4902d28ad598d87c7#diff-34610297bc46b266af677a9ce54f188299bf011b8b4ec4b04264fab2231ded21R162?

java/ql/lib/semmle/code/java/frameworks/MyBatis.qll
```diff
@@ -102,3 +104,94 @@ class MyBatisSqlOperationAnnotationMethod extends Method {
...
+  }
+}
+
+private class MyBatisAbstractSqlToStringStep extends SummaryModelCsv {
+  override predicate row(string row) {
+    row = "org.apache.ibatis.jdbc;AbstractSQL;true;toString;;;Argument[-1];ReturnValue;taint"
+  }
+}
+
+private class MyBatisAbstractSqlMethod extends string {
```
aschackmull, line 160: I'm not a fan of this QL style

aschackmull, line 160: I'm not entirely a fan of this QL style, could we inline this class instead?

java/ql/lib/semmle/code/java/frameworks/MyBatis.qll
```diff
@@ -102,3 +104,94 @@ class MyBatisSqlOperationAnnotationMethod extends Method {
 class TypeParam extends Interface {
   TypeParam() { this.hasQualifiedName("org.apache.ibatis.annotations", "Param") }
 }
+
+private class MyBatisAbstractSql extends RefType {
```
aschackmull, line 108: This is indeed dead code as per the Code Scanning warning.

java/ql/lib/semmle/code/java/frameworks/MyBatis.qll
```diff
@@ -102,3 +104,94 @@ class MyBatisSqlOperationAnnotationMethod extends Method {
...
+ * - `MyBatisProvider`
+ * - https://mybatis.org/mybatis-3/apidocs/org/apache/ibatis/annotations/package-summary.html
+ */
+class MyBatisInjectionSink extends DataFlow::Node {
+  MyBatisInjectionSink() {
+    exists(Annotation a, Method m |
+      a.getType() instanceof MyBatisProvider and
+      m.getDeclaringType() = a.getValue(["type", "value"]).(TypeLiteral).getTypeName().getType() and
+      m.hasName(a.getValue("method").(StringLiteral).getValue()) and
+      this.asExpr() = m.getBody().getAStmt().(ReturnStmt).getResult()
```
aschackmull, line 132: This restricts the return statement to occur in the top-level block in the method, which seems like an arbitrary restriction. Use `ReturnStmt::getEnclosingCallable()` instead.

java/ql/lib/semmle/code/java/frameworks/MyBatis.qll
```diff
@@ -102,3 +104,94 @@ class MyBatisSqlOperationAnnotationMethod extends Method {
...
+    )
+  }
+}
+
+private class MyBatisProviderStep extends TaintTracking::AdditionalValueStep {
+  override predicate step(DataFlow::Node n1, DataFlow::Node n2) {
+    exists(MethodAccess ma, Annotation a, Method providerMethod |
+      exists(int i |
+        ma.getArgument(i) = n1.asExpr() and
+        providerMethod.getParameter(i) = n2.asParameter()
```
aschackmull, line 142: A direct join on `i` ought to be pre-emptively prohibited:
```suggestion
        ma.getArgument(pragma[only_bind_into](i)) = n1.asExpr() and
        providerMethod.getParameter(pragma[only_bind_into](i)) = n2.asParameter()
```

java/ql/lib/semmle/code/java/frameworks/MyBatis.qll
aschackmull: I would prefer the latter.

java/ql/lib/semmle/code/java/frameworks/MyBatis.qll
```diff
@@ -102,3 +102,116 @@
...
+      )
+    |
+      a.getType() instanceof MyBatisProvider and
+      ma.getMethod().getAnAnnotation() = a and
+      providerMethod.getDeclaringType() =
+        a.getValue(["type", "value"]).(TypeLiteral).getTypeName().getType() and
+      providerMethod.hasName(a.getValue("method").(StringLiteral).getValue())
+    )
+  }
+}
```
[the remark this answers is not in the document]
jorgectf: https://github.com/github/codeql/issues/8953 :)

jorgectf:
All changes done @aschackmull :D

java/ql/lib/semmle/code/java/frameworks/MyBatis.qll
```diff
@@ -102,3 +104,116 @@ class MyBatisSqlOperationAnnotationMethod extends Method {
...
+ * - `MyBatisProvider`
+ * - https://mybatis.org/mybatis-3/apidocs/org/apache/ibatis/annotations/package-summary.html
+ */
+class MyBatisInjectionSink extends DataFlow::Node {
+  MyBatisInjectionSink() {
+    exists(Annotation a, Method m |
+      a.getType() instanceof MyBatisProvider and
+      m.getDeclaringType() = a.getValue(["type", "value"]).(TypeLiteral).getTypeName().getType() and
+      m.hasName(a.getValue("method").(StringLiteral).getValue()) and
+      this.getEnclosingCallable() = m.getBody().getAStmt().(ReturnStmt).getEnclosingCallable()
```
aschackmull, line 128: This change has now horribly broken the code. Take a moment to think why. Here's the proper fix:
```suggestion
      exists(ReturnStmt ret | this.asExpr() = ret.getResult() and ret.getEnclosingCallable() = m)
```

jorgectf: 🤯 My bad! I shouldn't make fast changes at night. Thanks for pointing it out! :)

aschackmull (approved):

PR #8347, opened 2022-03-07
remove all deprecations that are over a year old

erik-krogh:
We tend to add deprecated predicates to keep backwards compatibility.  
But I've yet to see us remove deprecated predicates after the 1 year deprecation period.  

I did a bit of scripting to detect and remove all deprecated predicates/classes/modules that are more than 1 year old (according to `git blame`).  

300+ deprecated things are deleted by this PR.


MathiasVP (dismissed):
I haven't verified that all of the deleted stuff has been deprecated for > 1 year. But assuming that this is the case (and we don't break any tests), this LGTM for C/C++! It should definitely have a change note though.

jketema:
Is there any reason something like `https://github.com/github/codeql/blob/main/cpp/ql/src/external/MostlyDuplicateClass.ql` wasn't removed? Which had an `@deprecated` 17 months ago.

hvitved (dismissed):
C# 👍

yoff:
Great initiative!

RasmusWL:
Very nice :muscle: I've often also wanted this :+1: 

> I did a bit of scripting to detect and remove all deprecated predicates/classes/modules that are more than 1 year old (according to `git blame`).

Did you consider that a change might have been made significantly earlier than the PR was merged? Such that the git-blame for the line might be 12 months old, but the PR was only merged 10 months ago, and therefore we can't actually delete it yet?

erik-krogh:
> Did you consider that a change might have been made significantly earlier than the PR was merged? Such that the git-blame for the line might be 12 months old, but the PR was only merged 10 months ago, and therefore we can't actually delete it yet?

No I didn't. Doing it right would require finding the commit and associated PR with the GitHub API, and that sounds tricky.  
I could add a few months on top of the 12 months to guard against that case?

RasmusWL:
> I could add a few months on top of the 12 months to guard against that case?

I think that sounds fine for an automated solution. My gut feeling would be that 14 months should cover all cases :+1:

smowton (dismissed):
Java :+1:

python/ql/lib/semmle/python/Flow.qll
```diff
@@ -507,18 +483,6 @@ class ImportStarNode extends ControlFlowNode {
 class SubscriptNode extends ControlFlowNode {
   SubscriptNode() { toAst(this) instanceof Subscript }
 
-  /**
-   * DEPRECATED: Use `getObject()` instead.
-   * This will be formally deprecated before the end 2018 and removed in 2019.
```
tausbn, line 512: Oh, I remember this comment. :laughing: 

It's a pity this didn't say "removed in 2020", seeing as we're still in March 2020 according to some people.

tausbn (dismissed):
Nice! :+1:

cpp/ql/lib/semmle/code/cpp/File.qll
```diff
@@ -308,13 +288,6 @@ class File extends Container, @file {
    */
   override predicate fromSource() { numlines(underlyingElement(this), _, _, _) }
 
-  /**
-   * Holds if this file may be from a library.
-   *
-   * DEPRECATED: For historical reasons this is true for any file.
-   */
-  deprecated override predicate fromLibrary() { any() }
```
geoffw0, line 316: Oooh, I think this one's existed in a deprecated state since around the time I started at Semmle.  About time!

python/ql/lib/semmle/python/types/Extensions.qll
```diff
@@ -30,40 +30,7 @@ abstract class PointsToExtension extends @py_flow_node {
...
-
-abstract deprecated class CustomPointsToOriginFact extends CustomPointsToFact {
-  abstract predicate pointsTo(Object value, ClassObject cls);
-
-  override predicate pointsTo(Context context, Object value, ClassObject cls, ControlFlowNode origin) {
-    this.pointsTo(value, cls) and origin = this and context.appliesTo(this)
-  }
-}
-
 /* Custom points-to fact with inferred class */
```
jketema, line 33: This seems a left-over comment that should be removed.

javascript/ql/lib/change-notes/2022-02-07-deleted-deprecations.md
```diff
@@ -0,0 +1,4 @@
+---
+category: minorAnalysis
+---
+* All deprecated predicates/classes/modules that have been deprecated for over a year has been deleted.
```
kaeluka, line 4:
```suggestion
* All deprecated predicates/classes/modules that have been deprecated for over a year have been deleted.
```

python/ql/lib/change-notes/2022-02-07-deleted-deprecations.md
```diff
@@ -0,0 +1,4 @@
+---
+category: minorAnalysis
+---
+* All deprecated predicates/classes/modules that have been deprecated for over a year has been deleted.
```
kaeluka, line 4:
```suggestion
* All deprecated predicates/classes/modules that have been deprecated for over a year have been deleted.
```

cpp/ql/lib/change-notes/2022-02-07-deleted-deprecations.md
```diff
@@ -0,0 +1,4 @@
+---
+category: minorAnalysis
+---
+* All deprecated predicates/classes/modules that have been deprecated for over a year has been deleted.
```
kaeluka, line 4:
```suggestion
* All deprecated predicates/classes/modules that have been deprecated for over a year have been deleted.
```

csharp/ql/lib/change-notes/2022-02-07-deleted-deprecations.md
```diff
@@ -0,0 +1,4 @@
+---
+category: minorAnalysis
+---
+* All deprecated predicates/classes/modules that have been deprecated for over a year has been deleted.
```
kaeluka, line 4:
```suggestion
* All deprecated predicates/classes/modules that have been deprecated for over a year have been deleted.
```

java/ql/lib/change-notes/2022-02-07-deleted-deprecations.md
```diff
@@ -0,0 +1,4 @@
+---
+category: minorAnalysis
+---
+* All deprecated predicates/classes/modules that have been deprecated for over a year has been deleted.
```
kaeluka, line 4:
```suggestion
* All deprecated predicates/classes/modules that have been deprecated for over a year have been deleted.
```

kaeluka (approved):
The changes to javascript LGTM. Just a typo in the changenotes.

cpp/ql/lib/semmle/code/cpp/controlflow/LocalScopeVariableReachability.qll
```diff
@@ -4,125 +4,6 @@
 
 import cpp
```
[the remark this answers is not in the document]
MathiasVP: `bbSuccessorEntryReachesLoopInvariant` is definitely just an implementation detail that should have never been public. I think this predicate is public because:
- We have a file `StackVariableReachability.qll` that is pretty much identical to file this, except that its API uses `StackVariable` instead of `LocalScopeVariable`.
- `StackVariableReachability.qll`'s `bbSuccessorEntryReachesLoopInvariant` predicate is used internally by [DefinitionsAndUses.qll](https://github.com/github/codeql/blob/main/cpp%2Fql%2Flib%2Fsemmle%2Fcode%2Fcpp%2Fcontrolflow%2FDefinitionsAndUses.qll#L211)

All in all, no one in their right mind would use `bbSuccessorEntryReachesLoopInvariant` from this file. I say we just delete it. I think the entire file can be replaced with:
```ql
import semmle.code.cpp.controlflow.StackVariableReachability
class LocalScopeVariableReachability = StackVariableReachability
```
As a bonus, this will also ensure that the `bbSuccessorEntryReachesLoopInvariant` predicate from `StackVariableReachability.qll` is still available when you import `LocalScopeVariableReachability.qll`.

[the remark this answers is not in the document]
erik-krogh: The `LocalScopeVariableReachability` class a deleted deprecation anyway, so we don't need that alias.  
I think the conclusion is that I can just delete the file?

erik-krogh:
I did a rebase to resolve a merge conflict.  
I reapplied the patch in the `delete all deprecations that are over 14 months old`-commit, and everything else is the same.

MathiasVP (approved):

sins921 (commented):
****

PR #8348, opened 2022-03-07
C#: ExternalAPI implementation for Telemetry.

michaelnebel:
In this PR we introduce C# Telemetry queries for
* All usages of libraries (summarised).
* Usages of unsupported APIs (ie. missing flow summaries).
* Usages of supported APIs (ie. flow summaries)
* Usages of known Sources.
* Usages of known Sinks.

The structure implementation is inspired by the corresponding java implementation, but there is only limited re-use of the java implementation itself.
Furthermore, the there are some language specific differences between C# and Java.
Some explicit sharing between C# and Java has been made.

The following concepts has been translated from Java to C# (which is reflected in the C# implementation).
* Java Package -> C# Namespace.
* Java Jar container -> C# Assembly

Other relevant information:
* The *Telemetry* queries have *kind* `metric` and is tagged as `summary`. This means that the queries will be executed as a part of the `csharp-code-scanning.qls` suite (as this suite includes all `metric` queries with the tag `summary`). Also see the discussion here: `https://github.com/github/codeql/pull/7417`.
* The query [meta data](https://github.com/github/codeql/blob/main/docs/query-metadata-style-guide.md) for `metric` queries has been extended to allow the result to be of type `string`,`int` (see this [ADR](https://github.com/github/code-scanning/blob/main/docs%2Fadrs%2F0014-metrics-for-external-libraries.md).
* The result message (ie. string results of the query) will be stored in GitHubs datawarehouse in the table `ts_metric_results` and should be accessible via *datadot*.

### DIFF
--- a/csharp/ql/src/Telemetry/ExternalApi.qll
+++ b/csharp/ql/src/Telemetry/ExternalApi.qll
@@ -5,6 +5,7 @@ private import semmle.code.csharp.dataflow.DataFlow
 private import semmle.code.csharp.dataflow.ExternalFlow
 private import semmle.code.csharp.dataflow.FlowSummary
 private import semmle.code.csharp.dataflow.internal.DataFlowPrivate
+private import semmle.code.csharp.dataflow.internal.DataFlowDispatch as DataFlowDispatch
 private import semmle.code.csharp.dataflow.TaintTracking
 private import semmle.code.csharp.dataflow.internal.TaintTrackingPrivate
 private import semmle.code.csharp.security.dataflow.flowsources.Remote
@@ -23,7 +24,7 @@ class TestLibrary extends RefType {
 /**
  * An external API from either the C# Standard Library or a 3rd party library.
  */
-class ExternalApi extends Callable {
+class ExternalApi extends DataFlowDispatch::DataFlowCallable {
   ExternalApi() { this.fromLibrary() }
 
   /**
@@ -32,13 +33,13 @@ class ExternalApi extends Callable {
   private string getSignature() {
     result =
       this.getDeclaringType().getUnboundDeclaration() + "." + this.getName() + "(" +
-        this.parameterTypesToString() + ")"
+        parameterQualifiedTypeNamesToString(this) + ")"
   }
 
   /**
    * Gets the namespace of this API.
    */
-  private string getNamespace() { result = this.getDeclaringType().getNamespace().toString() }
+  private string getNamespace() { this.getDeclaringType().hasQualifiedName(result, _) }
 
   /**
    * Gets the assembly file name containing this API.
@@ -74,7 +75,7 @@ class ExternalApi extends Callable {
 
   /** Holds if this API has a supported summary. */
   predicate hasSummary() {
-    this.getUnboundDeclaration() = any(SummarizedCallable sc) or
+    this instanceof SummarizedCallable or
     defaultAdditionalTaintStep(this.getAnInput(), _)
   }
 
--- a/csharp/ql/src/Telemetry/ExternalApi.qll
+++ b/csharp/ql/src/Telemetry/ExternalApi.qll
@@ -58,15 +58,18 @@ class ExternalApi extends DataFlowDispatch::DataFlowCallable {
    */
   string getInfo() { result = this.getInfoPrefix() + "#" + this.getSignature() }
 
+  /** Gets a call to this API callable. */
+  DispatchCall getACall() {
+    exists(DataFlowDispatch::NonDelegateDataFlowCall call | call.getDispatchCall() = result |
+      this = result.getADynamicTarget().getUnboundDeclaration()
+      or
+      this = result.getAStaticTarget().getUnboundDeclaration()
+    )
+  }
+
   /** Gets a node that is an input to a call to this API. */
   private ArgumentNode getAnInput() {
-    exists(DispatchCall call |
-      result.getCall().(DataFlowDispatch::NonDelegateDataFlowCall).getDispatchCall() = call
-    |
-      this = call.getADynamicTarget().getUnboundDeclaration()
-      or
-      this = call.getAStaticTarget().getUnboundDeclaration()
-    )
+    result.getCall().(DataFlowDispatch::NonDelegateDataFlowCall).getDispatchCall() = this.getACall()
   }
 
   /** Gets a node that is an output from a call to this API. */
@@ -74,9 +77,7 @@ class ExternalApi extends DataFlowDispatch::DataFlowCallable {
     exists(DataFlowDispatch::NonDelegateDataFlowCall call, DataFlowImplCommon::ReturnKindExt ret |
       result = ret.getAnOutNode(call)
     |
-      this = call.getDispatchCall().getADynamicTarget().getUnboundDeclaration()
-      or
-      this = call.getDispatchCall().getAStaticTarget().getUnboundDeclaration()
+      this.getACall() = call.getDispatchCall()
     )
   }
 
--- a/csharp/ql/src/Telemetry/ExternalApi.qll
+++ b/csharp/ql/src/Telemetry/ExternalApi.qll
@@ -10,21 +10,21 @@ private import semmle.code.csharp.dataflow.internal.TaintTrackingPrivate
 private import semmle.code.csharp.security.dataflow.flowsources.Remote
 
 /**
+ * A test library.
+ */
+class TestLibrary extends RefType {
+  TestLibrary() {
+    this.getNamespace()
+        .getName()
+        .matches(["NUnit.Framework%", "Xunit%", "Microsoft.VisualStudio.TestTools.UnitTesting%"])
+  }
+}
+
+/**
  * An external API from either the C# Standard Library or a 3rd party library.
  */
-class ExternalAPI extends Callable {
-  ExternalAPI() { this.fromLibrary() }
-
-  /** Holds if this API is not worth supporting */
-  predicate isUninteresting() { this.isTestLibrary() or this.isParameterlessConstructor() }
-
-  /** Holds if this API is is a constructor without parameters */
-  private predicate isParameterlessConstructor() {
-    this instanceof Constructor and this.getNumberOfParameters() = 0
-  }
-
-  /** Holds if this API is part of a common testing library or framework */
-  private predicate isTestLibrary() { this.getDeclaringType() instanceof TestLibrary }
+class ExternalApi extends Callable {
+  ExternalApi() { this.fromLibrary() }
 
   /**
    * Gets the unbound type, name and parameter types of this API.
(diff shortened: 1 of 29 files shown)


csharp/ql/src/Telemetry/ExternalApi.qll
```diff
@@ -0,0 +1,30 @@
+/** Provides classes and predicates related to handling APIs from external libraries. */
+
+private import ExternalApiSpecific
+
+/**
+ * An external API from either the C# Standard Library or a 3rd party library.
+ */
+class ExternalApi extends ExternalApiSpecific {
```
hvitved, line 8: I think the level of sharing here is not enough that we should actually have a shared implementation.

csharp/ql/src/Telemetry/ExternalApiSpecific.qll
```diff
@@ -0,0 +1,80 @@
...
+   */
+  string getInfo() { result = this.getInfoPrefix() + "#" + this.getSignature() }
+
+  /** Gets a node that is an input to a call to this API. */
+  DataFlow::Node getAnInput() {
+    exists(Call call | call.getTarget().getUnboundDeclaration() = this |
+      result.asExpr() = call.getAnArgument()
+    )
+    or
+    result.(ArgumentNode).getCall().getEnclosingCallable() = this
```
hvitved, line 53: This looks wrong; in this case
```csharp
void M() {
  Foo(x);
}
```
`x` will be considered an input to `M`.

csharp/ql/src/Telemetry/ExternalApiSpecific.qll
```diff
@@ -0,0 +1,80 @@
...
+import semmle.code.csharp.dataflow.FlowSummary
+import semmle.code.csharp.dataflow.TaintTracking
+import semmle.code.csharp.security.dataflow.flowsources.Remote
+private import semmle.code.csharp.dataflow.internal.DataFlowPrivate
+private import semmle.code.csharp.dataflow.internal.TaintTrackingPrivate
+
+/**
+ * An external API from either the C# Standard Library or a 3rd party library.
+ */
+class ExternalApiSpecific extends Callable {
```
hvitved, line 15: Extends `DataFlowCallable` (i.e., only unbounds)?

csharp/ql/src/Telemetry/ExternalApiSpecific.qll
```diff
@@ -0,0 +1,80 @@
...
+   */
+  string getInfoPrefix() { result = this.getAssembly() + "#" + this.getNamespace() }
+
+  /**
+   * Gets the assembly file name, namespace and signature of this API.
+   */
+  string getInfo() { result = this.getInfoPrefix() + "#" + this.getSignature() }
+
+  /** Gets a node that is an input to a call to this API. */
+  DataFlow::Node getAnInput() {
```
hvitved, line 48: The current implementation will not include `this` arguments. I think something like this should work


```ql
private import semmle.code.csharp.dataflow.internal.DataFlowDispatch
  DataFlow::ArgumentNode getAnInput() {
    exists(NonDelegateDataFlowCall call |
      result.getCall() = call and
      this = [call.getDispatchCall().getADynamicTarget(), call.getDispatchCall().getAStaticTarget()).getUnboundDeclaration()
    )
  }
```

csharp/ql/src/Telemetry/ExternalApiSpecific.qll
```diff
@@ -0,0 +1,80 @@
...
+    result.(ArgumentNode).getCall().getEnclosingCallable() = this
+  }
+
+  /** Gets a node that is an output from a call to this API. */
+  DataFlow::Node getAnOutput() {
+    exists(Call call | call.getTarget().getUnboundDeclaration() = this | result.asExpr() = call)
+    or
+    result.(PostUpdateNode).getPreUpdateNode().(ArgumentNode).getCall().getEnclosingCallable() =
+      this
+  }
```
hvitved, line 62: This will not include `out`/`ref` outputs. Something like this perhaps, which should also handle post-update nodes:
```ql
private import DataFlowImplCommon

  DataFlow::Node getAnOutput() {
    exists(NonDelegateDataFlowCall call, ReturnKindExt ret |
      this = [call.getDispatchCall().getADynamicTarget(), call.getDispatchCall().getAStaticTarget()).getUnboundDeclaration() and
      result = ret.getAnOutNode(call)
    )
  }
```

csharp/ql/src/Telemetry/ExternalApiSpecific.qll
```diff
@@ -0,0 +1,80 @@
...
+
+  /** Holds if this API has a supported summary. */
+  predicate hasSummary() {
+    this.getUnboundDeclaration() = any(SummarizedCallable sc) or
```
hvitved, line 66:
```ql
this.getUnboundDeclaration() instanceof SummarizedCallable
```

hvitved: And you can get rid of the `getUnboundDeclaration()` call if the base class is changed to `DataFlowCallable`.

csharp/ql/src/Telemetry/SupportedExternalSinks.ql
```diff
@@ -0,0 +1,17 @@
+/**
+ * @name Supported sinks in external libraries
+ * @description A list of 3rd party APIs detected as sinks. Excludes test and generated code.
```
hvitved, line 3: Do we exclude generated code?

hvitved (commented):
Approach looks good to me. I mainly have some comments about how to identify input/output.

csharp/ql/src/Telemetry/ExternalApi.qll
michaelnebel: Alright. I will remove the sharing!

csharp/ql/src/Telemetry/ExternalApiSpecific.qll
michaelnebel: Yes, you are right. This should also be aligned with the TargetApi implementation, which also now extends DatFlowCallable. Also; We are only interested in unbound version as these are the ones we make summaries for.

csharp/ql/src/Telemetry/ExternalApiSpecific.qll
michaelnebel: Yes

csharp/ql/src/Telemetry/SupportedExternalSinks.ql
michaelnebel: We do not (copy paste error). Description will be updated to reflect this.

csharp/ql/src/Telemetry/ExternalApiSpecific.qll
michaelnebel: hmm yes, you are right - this is completely nonsense! Good catch!

csharp/ql/src/Telemetry/ExternalApiSpecific.qll
michaelnebel: Could we do a knowledge sharing session on some of this? I have previously stared at IPA type TDataFlowCall and not understood, what was going on. It is not clear to me, why a solution (like the one outlined above) will work.

csharp/ql/src/Telemetry/ExternalApiSpecific.qll
michaelnebel: Same comment as above.

csharp/ql/src/Telemetry/ExternalApiSpecific.qll
hvitved: Sure thing!

java/ql/src/Telemetry/ExternalApi.qll
```diff
@@ -35,17 +42,17 @@ class ExternalApi extends Callable {
         "#" + this.getName() + paramsString(this)
   }
 
+  private string containerAsJar(Container container) {
```
hvitved, line 45: This was already a problem before, but it should not be a member predicate, since `this` is unbound.

java/ql/src/Telemetry/ExternalApi.qll
```diff
@@ -79,22 +97,3 @@ class ExternalApi extends Callable {
   /** Holds if this API is supported by existing CodeQL libraries, that is, it is either a recognized source or sink or has a flow summary. */
   predicate isSupported() { this.hasSummary() or this.isSource() or this.isSink() }
 }
-
-/** DEPRECATED: Alias for ExternalApi */
-deprecated class ExternalAPI = ExternalApi;
```
hvitved, line 84: This should not be removed.

csharp/ql/src/Telemetry/ExternalApi.qll
```diff
@@ -0,0 +1,110 @@
...
+   * Gets the assembly file name and namespace of this API.
+   */
+  string getInfoPrefix() { result = this.getAssembly() + "#" + this.getNamespace() }
+
+  /**
+   * Gets the assembly file name, namespace and signature of this API.
+   */
+  string getInfo() { result = this.getInfoPrefix() + "#" + this.getSignature() }
+
+  /** Gets a node that is an input to a call to this API. */
```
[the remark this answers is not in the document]
michaelnebel: That makes sense! Will make it so.

java/ql/src/Telemetry/ExternalApi.qll
michaelnebel: Right. Will fix.

java/ql/src/Telemetry/ExternalApi.qll
michaelnebel: Alright I will put it back. Took the chance of removing it as it is a very recent deprecation (Erik Krogh made lots of refactoring to rename classes to use PascalCasing for classes).

hvitved: It is exactly there because, in principle, someone could be referring to the old name. So instead of breaking their code, they will get a warning.

michaelnebel: Oh right. Don't know what I was thinking.

csharp/ql/src/Telemetry/ExternalApi.qll
```diff
@@ -58,25 +58,26 @@ class ExternalApi extends DataFlowDispatch::DataFlowCallable {
...
 
-  /** Gets a node that is an input to a call to this API. */
-  private ArgumentNode getAnInput() {
-    exists(DispatchCall call |
-      result.getCall().(DataFlowDispatch::NonDelegateDataFlowCall).getDispatchCall() = call
-    |
-      this = call.getADynamicTarget().getUnboundDeclaration()
+  /** Gets a call to this API callable. */
+  DispatchCall getACall() {
+    exists(DataFlowDispatch::NonDelegateDataFlowCall call | call.getDispatchCall() = result |
```
hvitved, line 63: This can be removed.

hvitved (approved):

michaelnebel:
Regarding DCA
* The workflow run failure of the Powershell project is unrelated.
* There has been identified 5 new metric queries as a part of the run (the telemetry queries).
* The overall analysis time is increased with 3% for the code scanning query suite (which indicates that the time consumption of the new queries are insignificant).

Merging the changes now.

PR #8354, opened 2022-03-07
Incomplete url string sanitization

aibaars:
Added a new query, `rb/incomplete-url-substring-sanitization`. The query finds instances where a URL is incompletely sanitized due to insufficient checks.


javascript/ql/src/Security/CWE-020/IncompleteUrlSubstringSanitization.qll
```diff
@@ -0,0 +1,62 @@
+/**
```
erik-krogh, line 1: Shouldn't this file be named `*Query.qll` and be moved into the `lib/` pack?

aibaars: It's not really a general purpose library and does not define any clear extension points. Is there a good reason to make this code part of the public API ?

erik-krogh: Hmmm, good point. Keep it.

erik-krogh (approved):
JS :+1:

ruby/ql/lib/codeql/ruby/security/performance/RegExpTreeView.qll
```diff
@@ -59,6 +59,19 @@ module RegExpFlags {
...
 }
 
+/**
+ * Provides regular expression patterns.
+ */
+module RegExpPatterns {
+  /**
+   * Gets a pattern that matches common top-level domain names in lower case.
+   */
+  string commonTLD() {
```
nickrolfe, line 69: I saw you got suggestions about this predicate for the other languages; we should apply them here as well.

ruby/ql/src/queries/security/cwe-020/examples/IncompleteUrlSubstringSanitization_BAD1.rb
```diff
@@ -0,0 +1,8 @@
+class AppController < ApplicationController
+    def index
+        url = params[:url]
+        # BAD: the host of `url` may be controlled by an attacker
+        if url.include?("example.com")
+            redirect_to url
```
nickrolfe, line 6:
```suggestion
            redirect_to url
        end
```

ruby/ql/lib/codeql/ruby/InclusionTests.qll
```diff
@@ -0,0 +1,133 @@
...
+        value = index.getConstantValue().getInt() and
+        value = [0, -1]
+      |
+        value = -1 and polarity = false and comparison.getExpr() instanceof CaseEqExpr
+        or
+        value = -1 and polarity = false and comparison.getExpr() instanceof EqExpr
+        or
+        value = -1 and polarity = true and comparison.getExpr() instanceof NEExpr
+        or
+        value = 0 and polarity = false and comparison.getExpr() instanceof NEExpr
```
asgerf, line 100: Why this case? I'd say that `value = 0` is only relevant for relational comparisons.

ruby/ql/lib/codeql/ruby/InclusionTests.qll
```diff
@@ -0,0 +1,133 @@
...
+ * or is a substring of another string.
+ *
+ * Examples:
+ * ```
+ * A.include?(B)
+ * A.index(B) !== -1
+ * A.index(B) >= 0
+ * ```
+ */
+class InclusionTest extends DataFlow::Node instanceof InclusionTest::Range {
```
hvitved, line 20: Why is this using data-flow nodes instead of CFG nodes?

ruby/ql/lib/codeql/ruby/StringOps.qll
```diff
@@ -0,0 +1,180 @@
...
+
+private import ruby
+private import codeql.ruby.DataFlow
+private import codeql.ruby.controlflow.CfgNodes
+private import InclusionTests
+
+/**
+ * Provides classes for reasoning about string-manipulating expressions.
+ */
+module StringOps {
```
hvitved, line 13: Is this module wrapper needed?

ruby/ql/lib/codeql/ruby/StringOps.qll
```diff
@@ -0,0 +1,180 @@
...
+  /**
+   * A expression that is equivalent to `A.start_with?(B)` or `!A.start_with?(B)`.
+   */
+  class StartsWith extends DataFlow::Node instanceof StartsWith::Range {
```
hvitved, line 17: Why is this using data-flow nodes instead of CFG nodes?

ruby/ql/lib/codeql/ruby/StringOps.qll
aibaars: No, this just matches what JavaScript does.

ruby/ql/lib/codeql/ruby/security/performance/RegExpTreeView.qll
aibaars: That's right. That should be taken care of by the other pull request which shares the same commit.

ruby/ql/lib/codeql/ruby/InclusionTests.qll
aibaars: The JavaScript library does the same, so keeping it this way makes it easier to share the implementation of the query.

ruby/ql/src/queries/security/cwe-020/IncompleteUrlSubstringSanitizationSpecific.qll
```diff
@@ -0,0 +1,8 @@
+import codeql.ruby.DataFlow
+import codeql.ruby.StringOps
+import codeql.ruby.security.performance.RegExpTreeView::RegExpPatterns as RegExpPatterns
+
+/** Holds if `node` may evaluate to `value` */
+predicate mayHaveStringValue(DataFlow::Node node, string value) {
+  node.asExpr().getExpr().getConstantValue().getString() = value
```
hvitved, line 7: You can remove `getExpr()`

hvitved (changes requested):

PR #8358, opened 2022-03-07
C++: Upgrade cpp/system-data-exposure to high precision

geoffw0:
Upgrade `cpp/system-data-exposure` from `@precision medium` to `@precision high`, following recent work improving the query.

I've run the query on all LGTM projects (linked from the internal issue):
 - 38 projects have results (i.e. not many; `cpp/potential-system-data-exposure` exists as well now as is somewhat noisier).
 - in many cases the sources, sinks and paths are difficult to understand without a deeper understanding of each project.  I've done my best to review all results.
 - some fairly strong results, some weak/questionable, no clear false positives identified.
	- in the cases where the sink is a call to `write`, its hard to tell whether a socket or file is being written to, and if it is a file, whether that file could leak to an attacker.  For example, `uyvg/valgrind-preload` is probably writing to a log file and that's probably OK for most users; the first result in `kosh04/newlisp` it is explicitly writing to a log file, but I've no idea the security implications in that case.
		- we *could* migrate these results to `cpp/potential-system-data-exposure`, but I feel on average they're stronger than the results of that query, and not numerous enough to frustrate anyone.
	- the results in `bitnine-oss/agensgraph` might be (weakly?) encrypted?
		- we *could* add an encryption heuristic that blocks these paths, but given the small volume of these results I'm not inclined to at this time.


MathiasVP (approved):
I agree with everything you said! Should we run a DCA with the Code Scanning suite to make sure the performance of that suite is still good after we've included this query?

geoffw0:
Good idea.  Started.

geoffw0:
DCA results:
 - no alert changes.
 - new query `cpp/system-data-exposure` (which is correct, for the code scanning suite).
 - lots of wobble, including in build times which changing a query `@precision` cannot cause.  Wobble might tend to be higher on this smaller suite compared to the larger 'security extended' one we normally use ... or we've just been unlucky in this instance.
 - overall analysis time 2% slower.

I'm not particularly worried by what I'm seeing, but I'm not confident either.  I'm going to run DCA again and see if we get similar or different results.

jketema:
Note that you're now comparing your branch with one that has an updated stats file which is creating additional differences.

geoffw0:
> Note that you're now comparing your branch with one that has an updated stats file which is creating additional differences.

That's odd - I just told DCA to run on the PR - and trusted that it would do the right thing (i.e. either run on main + a merge of my branch into main; or on the last common commit + my branch).

I guess I'll merge main myself and start it over.

MathiasVP:
> > Note that you're now comparing your branch with one that has an updated stats file which is creating additional differences.
> 
> That's odd - I just told DCA to run on the PR - and trusted that it would do the right thing (i.e. either run on main + a merge of my branch into main; or on the last common commit + my branch).
> 
> I guess I'll merge main myself and start it over.

I'm not sure what you mean @jketema. The DCA run says it's comparing:

[f706dee](https://github.com/github/semmle-code/commit/f706deeeb3080066760647fcdf81fe29226d4494)@[08ce128](https://github.com/github/codeql/commit/08ce128d640b013e9ea5081b38a49f640a715733) with [f706dee](https://github.com/github/semmle-code/commit/f706deeeb3080066760647fcdf81fe29226d4494)@[cwe497c=c793699562](https://github.com/github/codeql/commit/c793699562af3ff7010e3830e22061e180586c50) which looks like the right thing to me? i.e., the same internal SHA on both runs, and the CodeQL SHA is just the base this PR + @geoffw0's changes. I don't think the updated stats PR can affect this?

jketema:
> 'm not sure what you mean @jketema. The DCA run says it's comparing: [f706dee](https://github.com/github/semmle-code/commit/f706deeeb3080066760647fcdf81fe29226d4494)@[08ce128](https://github.com/github/codeql/commit/08ce128d640b013e9ea5081b38a49f640a715733) with [f706dee](https://github.com/github/semmle-code/commit/f706deeeb3080066760647fcdf81fe29226d4494)@[cwe497c=c793699562](https://github.com/github/codeql/commit/c793699562af3ff7010e3830e22061e180586c50). This looks like the right thing to me? i.e., the same internal SHA on both runs, and the CodeQL SHA is just the base this PR + @geoffw0's changes. I don't think the updated stats PR can affect this?

How is the ql submodule handled? Is it just set to `cwe497c` or is it merged with whatever the submodule is set to in semmle-code?  In the former case you'll be using two different stats files. In the latter case things should be ok.

jketema:
I'm stupid. `08ce128` is the commit from just before the stats file merge, so everything should be ok.

geoffw0:
There's a lot of noise on DCA, but given the nature of the changes I think I should only be looking at analysis and query run times.  Here are the total analysis times for the two runs:

<html>
<body>


source | a targets | b targets | weight | conclusion | a_m | a_std | b_m | b_std | diff | relative
-- | -- | -- | -- | -- | -- | -- | -- | -- | -- | --
Overall (excl. partials) | - | - |   |   | 13781 |   | 14014 |   | 232.7 | 0.0169


</body>
</html>

<html>
<body>


source | a targets | b targets | weight | conclusion | a_m | a_std | b_m | b_std | diff | relative
-- | -- | -- | -- | -- | -- | -- | -- | -- | -- | --
Overall (excl. partials) | - | - |   |   | 14391 |   | 14493 |   | 102 | 0.00709


</body>
</html>

Annoying there's a slowdown on both, but I suspect it isn't statistically significant.  I don't see consistency in which projects / queries are faster / slower either.

I think this is ready to merge.

MathiasVP:
CI seems stuck. But obviously, no test results are changed by this PR. So I'm just gonna go ahead and merge this :shipit:.

PR #8360, opened 2022-03-08
[Java] Add CompileTimeConstantExpr.getStringified method

JLLeitschuh:
Removes CharacterLiteral from CompileTimeConstantExpr.getStringValue

Resolves:
 - https://github.com/github/codeql/pull/8325#issuecomment-1060470279
 - https://github.com/github/codeql/pull/8325#issuecomment-1060587205

### DIFF
--- a/java/ql/lib/change-notes/2022-03-07-char-literal-in-compile-time-constant.md
+++ b/java/ql/lib/change-notes/2022-03-07-char-literal-in-compile-time-constant.md
@@ -1,5 +1,5 @@
 ---
 category: minorAnalysis
 ---
- * Add new method `CompileTimeConstantExpr.getStringifiedValue` which attempts to compute the
+ * Add new predicate `CompileTimeConstantExpr.getStringifiedValue` which attempts to compute the
    `String.valueOf` string rendering of a constant expression.
--- a/java/ql/lib/change-notes/2022-03-07-char-literal-in-compile-time-constant.md
+++ b/java/ql/lib/change-notes/2022-03-07-char-literal-in-compile-time-constant.md
@@ -1,5 +1,0 @@
----
-category: minorAnalysis
----
- * Add new predicate `CompileTimeConstantExpr.getStringifiedValue` which attempts to compute the
-   `String.valueOf` string rendering of a constant expression.
--- a/java/ql/lib/change-notes/2022-03-07-char-literal-in-compile-time-constant.md
+++ b/java/ql/lib/change-notes/2022-03-07-char-literal-in-compile-time-constant.md
@@ -1,5 +1,5 @@
 ---
 category: minorAnalysis
 ---
- * Add new method `CharacterLiteral.getStringifiedValue` which attempts to compute the
-   `String.valueOf` compile time constant value of the expression.
+ * Add new method `CompileTimeConstantExpr.getStringifiedValue` which attempts to compute the
+   `String.valueOf` string rendering of a constant expression.
--- a/java/ql/lib/change-notes/2022-03-07-char-literal-in-compile-time-constant.md
+++ b/java/ql/lib/change-notes/2022-03-07-char-literal-in-compile-time-constant.md
@@ -1,0 +1,5 @@
+---
+category: minorAnalysis
+---
+ * Add new method `CharacterLiteral.getStringifiedValue` which attempts to compute the
+   `String.valueOf` compile time constant value of the expression.
--- a/java/ql/lib/semmle/code/java/Expr.qll
+++ b/java/ql/lib/semmle/code/java/Expr.qll
@@ -201,17 +201,9 @@ class CompileTimeConstantExpr extends Expr {
     result = this.(StringLiteral).getValue()
     or
     (
-      if this.(AddExpr).getAnOperand().getType() instanceof TypeString // If either operand type is already a String:
-      then
-        // Then the stringified version of the expression can be safely used as a String will be created.
-        result =
-          this.(AddExpr).getLeftOperand().(CompileTimeConstantExpr).getStringifiedValue() +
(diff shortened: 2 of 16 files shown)


java/ql/lib/semmle/code/java/Expr.qll
```diff
@@ -161,18 +161,58 @@ class CompileTimeConstantExpr extends Expr {
...
 
+  /**
+   * Gets the stringified value of this expression, where possible.
+   *
+   * The stringified version of a compile-time constant expression is the equivalent to
+   * the result of calling `String.valueOf(expr)` on the expression.
+   * 
+   * Note that this does not handle the following cases:
+   *
+   * - mathematical computations of type `long`.
```
JLLeitschuh, line 172: This also doesn't work with float computations, anything else I should call out explicitly?

java/ql/lib/change-notes/2022-03-07-char-literal-in-compile-time-constant.md
```diff
@@ -0,0 +1,5 @@
+---
+category: minorAnalysis
+---
+ * Add new method `CharacterLiteral.getStringifiedValue` which attempts to compute the
+   `String.valueOf` compile time constant value of the expression.
```
smowton, line 5:
```suggestion
 * Add new method `CompileTimeConstantExpr.getStringifiedValue` which attempts to compute the
   `String.valueOf` string rendering of a constant expression.
```

java/ql/lib/semmle/code/java/Expr.qll
```diff
@@ -161,18 +161,58 @@ class CompileTimeConstantExpr extends Expr {
...
+   */
+  pragma[nomagic]
+  string getStringifiedValue() {
+    result = getStringValue()
+    or
+    result = this.(Literal).getValue()
```
smowton, line 178: I don't think this is right re: floating-point expressions? Also check what happens re: int literals expressed in hex for example?

java/ql/lib/semmle/code/java/Expr.qll
```diff
@@ -161,18 +161,58 @@ class CompileTimeConstantExpr extends Expr {
...
+      then
+        // Then the stringified version of the expression can be safely used as a String will be created.
+        result =
+          this.(AddExpr).getLeftOperand().(CompileTimeConstantExpr).getStringifiedValue() +
+            this.(AddExpr).getRightOperand().(CompileTimeConstantExpr).getStringifiedValue()
+      else
+        // Adding two literals of primitive type will not result in a String.
+        result =
+          this.(AddExpr).getLeftOperand().(CompileTimeConstantExpr).getStringValue() +
+            this.(AddExpr).getRightOperand().(CompileTimeConstantExpr).getStringValue()
```
smowton, line 214:
```suggestion
      this.getType() instanceof TypeString and
      this.(AddExpr).getLeftOperand().(CompileTimeConstantExpr).getStringifiedValue() +
        this.(AddExpr).getRightOperand().(CompileTimeConstantExpr).getStringifiedValue()
```

Since if the result is a string, any non-string is getting `toString`'d, right?

java/ql/lib/change-notes/2022-03-07-char-literal-in-compile-time-constant.md
```diff
@@ -0,0 +1,5 @@
+---
+category: minorAnalysis
+---
+ * Add new method `CharacterLiteral.getStringifiedValue` which attempts to compute the
```
Marcono1234, line 4:
```suggestion
 * Add new predicate `CharacterLiteral.getStringifiedValue` which attempts to compute the
```

java/ql/test/library-tests/constants/getStringified.ql
```diff
@@ -0,0 +1,9 @@
+import semmle.code.java.Variable
```
Marcono1234, line 1: Shouldn't this be `import java`, or is there any advantage in importing specifically `Variable` (and the other classes transitively I assume)?
(Though it appears the other older tests in this directory are doing the same.)

java/ql/lib/semmle/code/java/Expr.qll
smowton: I was; it's probably fine then.

java/ql/test/library-tests/constants/getStringified.ql
smowton: Yeah, conventionally if whatever you're after is brought in by `import java` then just do that.

java/ql/lib/semmle/code/java/Expr.qll
JLLeitschuh: Much appreciated @Marcono1234 for following up here. I thought what I had was right, but I really appreciate you verifying!

java/ql/test/library-tests/constants/getStringified.ql
JLLeitschuh: Done!

java/ql/lib/semmle/code/java/Expr.qll
JLLeitschuh: I think this test does a nice job of confirming your assessment:

https://github.com/github/codeql/pull/8360/files#diff-5c5952823fbe7d5baaa07b97ee5b43627e904a86112101461d6be2e8eb6e4b06

smowton (approved):

JLLeitschuh:
I'm done with this PR, anyone who's able/interested, feel free to merge it! 😄

atorralba (approved):

smowton:
Should've measured performance, is very bad at least on Geogebra. Revert: https://github.com/github/codeql/pull/8402

PR #8366, opened 2022-03-08
C++: Mark everything in CodeDuplication.qll as deprecated

jketema:
Although we earlier added a comment to the classes in noting that they are deprecated, we did not properly mark the classes as actually being deprecated in https://github.com/github/codeql/pull/8036.

All predicates - except for 3 - depend on the classes being functional, which they no longer are, so mark those a deprecated as well. The three remaining predicates (`FunctionDeclarationEntry`, `numberOfSourceMethods`, and `whitelistedLineForDuplication`) seem to be helpers, and are likely not used when the library is not used, so mark those as deprecated as well.


cpp/ql/src/external/CodeDuplication.qll
```diff
@@ -74,29 +74,37 @@ class DuplicateBlock extends Copy {
...
+deprecated class SimilarBlock extends Copy {
   override string toString() {
     result = "Similar code: " + this.sourceLines() + " almost duplicated lines."
   }
 }
 
-/** Gets a function with a body and a location. */
-FunctionDeclarationEntry sourceMethod() {
+/**
+ * DEPRECATED: The CodeDuplication library will be removed in a future release.
```
MathiasVP, line 84:
```suggestion
 * DEPRECATED: The `CodeDuplication` library will be removed in a future release.
```

cpp/ql/src/external/CodeDuplication.qll
```diff
@@ -74,29 +74,37 @@ class DuplicateBlock extends Copy {
...
+deprecated FunctionDeclarationEntry sourceMethod() {
   result.isDefinition() and
   exists(result.getLocation()) and
   numlines(unresolveElement(result.getFunction()), _, _, _)
 }
 
-/** Gets the number of member functions in `c` with a body and a location. */
-int numberOfSourceMethods(Class c) {
+/**
+ * DEPRECATED: The CodeDuplication library will be removed in a future release.
```
MathiasVP, line 95:
```suggestion
 * DEPRECATED: The `CodeDuplication` library will be removed in a future release.
```

MathiasVP (dismissed):
Thanks for doing this!

jketema:
> Thanks for doing this!

Unfortunately CI is now failing because of this. It's failing on the query files in the same directory; which have been marked as deprecated for 17 months. I guess I could remove them.

MathiasVP:
> Unfortunately CI is now failing because of this. It's failing on the query files in the same directory; which have been marked as deprecated for 17 months. I guess I could remove them.

Ah, I see. If it's indeed been deprecated for so long feel free to remove them!

MathiasVP (approved):
:shipit:

jketema:
For completeness. The commit marking the queries as deprecated: https://github.com/github/codeql/commit/fc45b6cd3cee747044469669fc3680bf2ec79f64

jketema:
Removing one more deprecated query made me touch [config/identical-files.json](https://github.com/github/codeql/pull/8366/files#diff-3abab5de80197372a3cc8d192e4e5abf1d026265b819d08b2dd1c406d691b9ff). Not sure whether this means other people now also need to review this.

MathiasVP (approved):

PR #8368, opened 2022-03-08
C++: Factor must-flow predicates out of two queries

MathiasVP:
This PR factors out the common functionality from `cpp/unsafe-use-of-this` and `cpp/return-stack-allocated-memory`. This shouldn't change any behavior[^1], but it should mean that we re-evaluate a _tiny_ little bit less stuff in the Code Scanning suite.

[^1]: Apart from the path explanation for `cpp/unsafe-use-of-this` which is now a lot better as it doesn't just use the `toString` given by the IR.

### DIFF
--- a/cpp/ql/lib/semmle/code/cpp/ir/dataflow/MustFlow.qll
+++ b/cpp/ql/lib/semmle/code/cpp/ir/dataflow/MustFlow.qll
@@ -1,0 +1,265 @@
+/**
+ * This file provides a library for inter-procedural must-flow data flow analysis.
+ * Unlike `DataFlow.qll`, the analysis provided by this file checks whether data _must_ flow
+ * from a source to a _sink_.
+ */
+
+private import cpp
+import semmle.code.cpp.ir.dataflow.DataFlow
+private import semmle.code.cpp.ir.IR
+
+/**
+ * A configuration of a data flow analysis that performs must-flow analysis. This is different
+ * from `DataFlow.qll` which performs may-flow analysis (i.e., it finds paths where the source _may_
+ * flow to the sink).
+ *
+ * Like in `DataFlow.qll`, each use of the `MustFlow.qll` library must define its own unique extension
+ * of this abstract class. To create a configuration, extend this class with a subclass whose
+ * characteristic predicate is a unique singleton string and override `isSource`, `isSink` (and
+ * `isAdditionalFlowStep` if additional steps are required).
+ */
+abstract class MustFlowConfiguration extends string {
+  bindingset[this]
+  MustFlowConfiguration() { any() }
+
+  /**
+   * Holds if `source` is a relevant data flow source.
+   */
+  abstract predicate isSource(DataFlow::Node source);
+
+  /**
+   * Holds if `sink` is a relevant data flow sink.
+   */
+  abstract predicate isSink(DataFlow::Node sink);
+
+  /**
+   * Holds if the additional flow step from `node1` to `node2` must be taken
+   * into account in the analysis.
+   */
+  predicate isAdditionalFlowStep(DataFlow::Node node1, DataFlow::Node node2) { none() }
+
+  /**
+   * Holds if data must flow from `source` to `sink` for this configuration.
+   *
+   * The corresponding paths are generated from the end-points and the graph
+   * included in the module `PathGraph`.
+   */
+  final predicate hasFlowPath(MustFlowPathNode source, MustFlowPathSink sink) {
+    this.isSource(source.getNode()) and
+    source.getASuccessor+() = sink
+  }
+}
+
+/** Holds if `node` flows from a source. */
+pragma[nomagic]
+private predicate flowsFromSource(DataFlow::Node node, MustFlowConfiguration config) {
+  config.isSource(node)
+  or
+  exists(DataFlow::Node mid |
+    step(mid, node, config) and
+    flowsFromSource(mid, pragma[only_bind_into](config))
+  )
+}
+
+/** Holds if `instr` flows to a sink. */
+pragma[nomagic]
+private predicate flowsToSink(DataFlow::Node node, MustFlowConfiguration config) {
+  flowsFromSource(node, pragma[only_bind_into](config)) and
+  (
+    config.isSink(node)
+    or
+    exists(DataFlow::Node mid |
+      step(node, mid, config) and
+      flowsToSink(mid, pragma[only_bind_into](config))
+    )
+  )
+}
+
+cached
+private module Cached {
+  /** Holds if `p` is the `n`'th parameter of the non-virtual function `f`. */
+  private predicate parameterOf(Parameter p, Function f, int n) {
+    not f.isVirtual() and f.getParameter(n) = p
+  }
+
+  /**
+   * Holds if `instr` is the `n`'th argument to a call to the non-virtual function `f`, and
+   * `init` is the corresponding initiazation instruction that receives the value of `instr` in `f`.
+   */
+  private predicate flowIntoParameter(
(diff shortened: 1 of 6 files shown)


cpp/ql/lib/semmle/code/cpp/ir/dataflow/MustFlow.qll
```diff
@@ -0,0 +1,265 @@
...
+private predicate flowsFromSource(DataFlow::Node node, MustFlowConfiguration config) {
+  config.isSource(node)
+  or
+  exists(DataFlow::Node mid |
+    step(mid, node, config) and
+    flowsFromSource(mid, pragma[only_bind_into](config))
+  )
+}
+
+/** Holds if `instr` flows to a sink. */
```
jketema, line 64:
```suggestion
/** Holds if `node` flows to a sink. */
```

cpp/ql/lib/semmle/code/cpp/ir/dataflow/MustFlow.qll
```diff
@@ -0,0 +1,265 @@
...
+  ) {
+    exists(int n |
+      parameterOf(_, f, n) and
+      flowIntoParameter(f, pragma[only_bind_into](pragma[only_bind_out](n)), call, instr, init)
+    )
+  }
+
+  /**
+   * Holds if `instr` is the qualifier to a call to the non-virtual function `f`, and
+   * `init` is the corresponding initiazation instruction that receives the value of
```
jketema, line 115:
```suggestion
   * `init` is the corresponding initialization instruction that receives the value of
```

cpp/ql/src/Likely Bugs/Memory Management/ReturnStackAllocatedMemory.ql
```diff
@@ -18,157 +18,70 @@ import cpp
...
-  or
-  operandToInstructionStep(node1.asOperand(), node2.asInstruction())
-}
-
-predicate instructionToOperandStep(Instruction instr, Operand operand) { operand.getDef() = instr }
-
-/**
- * Holds if `operand` flows to the result of `instr`.
- *
- * This predicate ignores flow through `PhiInstruction`s to create a 'must flow' relation. It also
```
jketema, line 69: This remark about `PhiInstruction`s is not in the new version, maybe add it to `MustFlow.qll`?

cpp/ql/src/Likely Bugs/Memory Management/ReturnStackAllocatedMemory.ql
```diff
@@ -18,157 +18,70 @@ import cpp
...
+  MustFlowPathNode source, MustFlowPathNode sink, VariableAddressInstruction var,
+  ReturnStackAllocatedMemoryConfig conf
 where
-  hasFlow(source, sink) and
-  source.getNode().asInstruction() = var
+  conf.hasFlowPath(source, sink) and
+  source.getNode().asInstruction() = var and
+  // Only raise an alert if we're returning from the _same_ callable as the on that
+  // declared the stack variable.
+  var.getEnclosingFunction() = sink.getNode().getEnclosingCallable()
```
jketema, line 85: This seems new, or what this somewhere else before in a disguise that I didn't pick up on?

jketema (commented):
Only a few small comments.

cpp/ql/src/Likely Bugs/Memory Management/ReturnStackAllocatedMemory.ql
jketema: It think it's important to have this, as this is what locally makes this "must" flow.

cpp/ql/src/Likely Bugs/Memory Management/ReturnStackAllocatedMemory.ql
MathiasVP: Yeah, it's new. The old must-flow predicates in `cpp/return-stack-allocated-memory` didn't have flow into function calls, but that was implemented (and indeed required) for `cpp/unsafe-use-of-this`. So I've ported flow into function calls to this new library.

This means that we now track flow into function calls in `cpp/return-stack-allocated-memory` as well, and we could now get a result from this snippet without this extra condition:
```
void id(int* px) {
  return p; // this returns the local variable `x`, but it's fine as the local variable isn't declared in this scope.
}
void f() {
  int x;
  int* px = id(&x);
}
```

Hm, I should actually add this test to the test cases for the `cpp/return-stack-allocated-memory` query. Let me go ahead and do that!

cpp/ql/src/Likely Bugs/Memory Management/ReturnStackAllocatedMemory.ql
MathiasVP: Good point. I'll add that comment back in there 👍.

cpp/ql/src/Likely Bugs/Memory Management/ReturnStackAllocatedMemory.ql
MathiasVP: 69417e150ae7bbb6ad1f909fbb46e95bb39c1dd2 adds the above test.

cpp/ql/src/Likely Bugs/Memory Management/ReturnStackAllocatedMemory.ql
MathiasVP: 69417e150ae7bbb6ad1f909fbb46e95bb39c1dd2 re-adds this comment.

cpp/ql/src/Likely Bugs/Memory Management/ReturnStackAllocatedMemory.ql
```diff
@@ -18,157 +18,70 @@ import cpp
...
+   * This configuration intentionally conflates addresses of fields and their object, and pointer offsets
+   * with their base pointer as this allows us to detect cases where an object's address flows to a
+   * return statement via a field. For example:
+   *
+   * ```cpp
+   * struct S { int x, y };
+   * int* test() {
+   *   S s;
+   *   return &s.x; // BAD: &s.x is an address of a variable on the stack.
+   * }
```
jketema, line 68: There's no closing triple quotes after snippet.

cpp/ql/src/Likely Bugs/Memory Management/ReturnStackAllocatedMemory.ql
jketema: Is this something that needs wider testing? LGTM or at least DCA?

MathiasVP: I don't think this particular change warrants extended testing, but this PR as a whole definitely needs to be tested with DCA. I've already started a run which should hopefully be backlinked from this PR.

cpp/ql/lib/semmle/code/cpp/ir/dataflow/MustFlow.qll
```diff
@@ -0,0 +1,270 @@
...
+  ) {
+    not f.isVirtual() and
+    init.getEnclosingFunction() = f
+  }
+
+  /**
+   * Holds if `instr` is an argument (or argument indirection) to a call, and
+   * `succ` is the corresponding initialization instruction in the call target.
+   */
+  private predicate flowThroughCallable(Instruction argument, Instruction parameter) {
```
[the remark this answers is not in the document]
MathiasVP: Good point. I would like to add flow out at some point, but I'm not exactly sure how to do that in a must-flow way.

MathiasVP:
DCA looks good. Merging 🎉

PR #8372, opened 2022-03-08
Bump ATM pack version numbers to 0.1.0

henrymercer:
Perform a minor version bump since we're shipping quite a few changes in the next release.


tiferet:
OoI, what are the high-level changes being shipped? Are they all optimization changes, or do some of them change the results?

A related question: What happens if some of the results are changed by the new version? Will alerts that were created by the old version but are no longer detected by the new version close automatically? IIRC there was some problem with that a couple of months ago, when Ian was testing different models before the public launch -- do you know if that's working as expected now?

esbena (approved):

henrymercer:
Superseded by #8404

PR #8373, opened 2022-03-08
Ruby: Fix incorrect parsing of ranges

hvitved:
Avoids parsing the `4,8}` bit in `a{4,8}` as a `RegExpNormalChar`.

cc @yoff and @joefarebrother


nickrolfe (approved):
Nice catch. It's a bit embarrassing that it slipped through before, with all those spurious entries that were in `parse.expected`.

PR #8374, opened 2022-03-08
QL: add query detecting block comments in a position where a QLDoc should be

erik-krogh:
Inspired by [this block comment](https://github.com/github/codeql/blob/86121164c5b0695a8c12fa43677f55db62b34dc7/python/ql/lib/semmle/python/types/Extensions.qll#L51) that should have been a QLDoc.  

That comment not being a QLDoc caused my "remove all deprecated things and their QLDoc"-patch to not delete that comment.  

There appears to be some results, mostly in Python. (The results I haven't fixed look benign).


MathiasVP (approved):

michaelnebel (approved):
C#: LGTM.

tausbn (approved):
Argh. Just when I fixed a whole bunch of Python alerts. :sob: 
Anyway, I think this looks reasonable. :+1:

geoffw0 (commented):
CPP :+1:

PR #8380, opened 2022-03-09
JS: support that the base is not a method-call in getAChainedMethodCall

erik-krogh:
Gets a TP for CVE-2021-46704 

[Evaluation was uneventful](https://github.com/github/codeql-dca-main/tree/data/erik-krogh/pr-8380-cebd241__nightly__security-extend/reports).  

--- 

Consider the below. 

```CodeQL
result = DataFlow::moduleImport("foo").getAnInstantiation().getAMethodCall*().getAMethodCall("bar");
```

This will not flag the below: 
```JavaScript
const Foo = require("foo");
const inst = new Foo();
const result = inst.bar();
```

Because `DataFlow::moduleImport("foo").getAnInstantiation().getAMethodCall*()` is restricted to the type `DataFlow::CallNode`, even if we take 0 steps through the transitive closure.

### DIFF
--- a/javascript/ql/lib/semmle/javascript/dataflow/Sources.qll
+++ b/javascript/ql/lib/semmle/javascript/dataflow/Sources.qll
@@ -147,7 +147,7 @@ class SourceNode extends DataFlow::Node {
    */
   DataFlow::CallNode getAChainedMethodCall(string methodName) {
     // the direct call to `getAMethodCall` is needed in case the base is not a `DataFlow::CallNode`.
-    result = [getAMethodCall*().getAMethodCall(methodName), getAMethodCall(methodName)]
+    result = [getAMethodCall+().getAMethodCall(methodName), getAMethodCall(methodName)]
   }
 
   /**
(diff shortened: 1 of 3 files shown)


javascript/ql/lib/semmle/javascript/dataflow/Sources.qll
```diff
@@ -146,7 +146,8 @@ class SourceNode extends DataFlow::Node {
    * that is, `o.m(...)` or `o[p](...)`.
    */
   DataFlow::CallNode getAChainedMethodCall(string methodName) {
-    result = getAMethodCall*().getAMethodCall(methodName)
+    // the direct call to `getAMethodCall` is needed in case the base is not a `DataFlow::CallNode`.
+    result = [getAMethodCall*().getAMethodCall(methodName), getAMethodCall(methodName)]
```
kaeluka, line 150:
```suggestion
    result = [getAMethodCall+().getAMethodCall(methodName), getAMethodCall(methodName)]
```
this should work just as well and express intention clearer — right?

kaeluka (approved):
So, I'm understanding this correctly — the problem in your example is that in the third line below, `inst` is not a `CallNode` and therefore the `bar` call is not found? That's quite an easy mistake to make. As you're on a roll, this might be worth doing a ql-for-ql query for?

```
const Foo = require("foo");
const inst = new Foo();
const result = inst.bar();
```

Also, the check asks for a change note, but I think you can disable that.

esbena (approved):

PR #8382, opened 2022-03-09
C++: Use a `TaintTracking::Configuration` in three more queries

MathiasVP:
Initially, I was also going to modernize the sanitizers used in the three queries, but I realized that this code should be changed anyway once we merged the ongoing work on range analysis. So for now I've copied the ad-hoc sanitizers into these three queries to not change their behavior.

### DIFF
--- a/cpp/ql/src/Critical/OverflowDestination.ql
+++ b/cpp/ql/src/Critical/OverflowDestination.ql
@@ -14,7 +14,10 @@
  */
 
 import cpp
-import semmle.code.cpp.security.TaintTracking
+import semmle.code.cpp.ir.dataflow.TaintTracking
+import semmle.code.cpp.controlflow.IRGuards
+import semmle.code.cpp.security.FlowSources
+import DataFlow::PathGraph
 
 /**
  * Holds if `fc` is a call to a copy operation where the size argument contains
@@ -45,9 +48,49 @@ predicate sourceSized(FunctionCall fc, Expr src) {
   )
 }
 
-from FunctionCall fc, Expr vuln, Expr taintSource
+predicate readsVariable(LoadInstruction load, Variable var) {
+  load.getSourceAddress().(VariableAddressInstruction).getASTVariable() = var
+}
+
+predicate hasUpperBoundsCheck(Variable var) {
+  exists(RelationalOperation oper, VariableAccess access |
+    oper.getAnOperand() = access and
+    access.getTarget() = var and
+    // Comparing to 0 is not an upper bound check
+    not oper.getAnOperand().getValue() = "0"
+  )
+}
+
+predicate nodeIsBarrierEqualityCandidate(DataFlow::Node node, Operand access, Variable checkedVar) {
+  readsVariable(node.asInstruction(), checkedVar) and
+  any(IRGuardCondition guard).ensuresEq(access, _, _, node.asInstruction().getBlock(), true)
+}
+
+class OverflowDestinationConfig extends TaintTracking::Configuration {
+  OverflowDestinationConfig() { this = "OverflowDestinationConfig" }
+
+  override predicate isSource(DataFlow::Node source) { source instanceof FlowSource }
+
+  override predicate isSink(DataFlow::Node sink) { sourceSized(_, sink.asExpr()) }
+
+  override predicate isSanitizer(DataFlow::Node node) {
+    exists(Variable checkedVar |
+      readsVariable(node.asInstruction(), checkedVar) and
+      hasUpperBoundsCheck(checkedVar)
+    )
+    or
+    exists(Variable checkedVar, Operand access |
+      readsVariable(access.getDef(), checkedVar) and
+      nodeIsBarrierEqualityCandidate(node, access, checkedVar)
+    )
+  }
+}
+
+from
+  FunctionCall fc, OverflowDestinationConfig conf, DataFlow::PathNode source,
+  DataFlow::PathNode sink
 where
-  sourceSized(fc, vuln) and
-  tainted(taintSource, vuln)
+  conf.hasFlowPath(source, sink) and
+  sourceSized(fc, sink.getNode().asExpr())
 select fc,
   "To avoid overflow, this operation should be bounded by destination-buffer size, not source-buffer size."
--- a/cpp/ql/src/Security/CWE/CWE-129/ImproperArrayIndexValidation.ql
+++ b/cpp/ql/src/Security/CWE/CWE-129/ImproperArrayIndexValidation.ql
@@ -3,7 +3,7 @@
  * @description Accessing an array without first checking
  *              that the index is within the bounds of the array can
  *              cause undefined behavior and can also be a security risk.
- * @kind problem
+ * @kind path-problem
  * @id cpp/unclear-array-index-validation
  * @problem.severity warning
  * @security-severity 8.8
@@ -12,9 +12,11 @@
  */
 
 import cpp
-import semmle.code.cpp.controlflow.Guards
-private import semmle.code.cpp.rangeanalysis.RangeAnalysisUtils
-import semmle.code.cpp.security.TaintTracking
+import semmle.code.cpp.controlflow.IRGuards
+import semmle.code.cpp.security.FlowSources
+import semmle.code.cpp.ir.dataflow.TaintTracking
+import semmle.code.cpp.rangeanalysis.RangeAnalysisUtils
+import DataFlow::PathGraph
 
 predicate hasUpperBound(VariableAccess offsetExpr) {
   exists(BasicBlock controlled, StackVariable offsetVar, SsaDefinition def |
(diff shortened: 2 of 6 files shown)


cpp/ql/src/Critical/OverflowDestination.ql
```diff
@@ -45,9 +48,49 @@ predicate sourceSized(FunctionCall fc, Expr src) {
...
+  override predicate isSanitizer(DataFlow::Node node) {
+    exists(Variable checkedVar |
+      readsVariable(node.asInstruction(), checkedVar) and
+      hasUpperBoundsCheck(checkedVar)
+    )
+    or
+    exists(Variable checkedVar, Operand access |
+      readsVariable(access.getDef(), checkedVar) and
+      nodeIsBarrierEqualityCandidate(node, access, checkedVar)
+    )
```
geoffw0, line 85: This is a copy of `nodeIsBarrier` from `DefaultTaintTracking.qll`, right?  If we're emulating that, do `nodeIsBarrierIn` and `additionalTaintStep` from `DefaultTaintTracking.qll` matter as well?

cpp/ql/src/Critical/OverflowDestination.ql
```diff
@@ -14,7 +14,10 @@
  */
 
 import cpp
-import semmle.code.cpp.security.TaintTracking
+import semmle.code.cpp.ir.dataflow.TaintTracking
+import semmle.code.cpp.controlflow.IRGuards
+import semmle.code.cpp.security.FlowSources
+import DataFlow::PathGraph
```
geoffw0, line 20: You haven't changed this query to `path-problem`.

cpp/ql/src/Critical/OverflowDestination.ql
geoffw0: Also it might be good to have a note explaining their history and that they may or may not actually be helping the query at this point - so that future maintainers can potentially clean things up without being overly confused/paranoid.

cpp/ql/src/Critical/OverflowDestination.ql
geoffw0: Doing this should affect the test in `cpp\ql\test\query-tests\Security\CWE\CWE-119\semmle\tests`.

cpp/ql/src/Security/CWE/CWE-129/ImproperArrayIndexValidation.ql
```diff
@@ -32,11 +34,60 @@ predicate linearBoundControls(BasicBlock controlled, SsaDefinition def, StackVar
...
+
+predicate isFlowSource(FlowSource source, string sourceType) { sourceType = source.getSourceType() }
+
+class ImproperArrayIndexValidationConfig extends TaintTracking::Configuration {
+  ImproperArrayIndexValidationConfig() { this = "ImproperArrayIndexValidationConfig" }
+
+  override predicate isSource(DataFlow::Node source) { isFlowSource(source, _) }
+
+  override predicate isSanitizer(DataFlow::Node node) {
+    hasUpperBound(node.asExpr())
```
geoffw0, line 68: This duplicates the check in `isUnboundedArrayIndex`.  Do you expect having it here will affect results or performance?

geoffw0 (commented):
Looks good, I've made a few comments and questions.  I will also do some kind of run on LGTM of each query before approving this to merge.

cpp/ql/src/Critical/OverflowDestination.ql
MathiasVP: > This is a copy of `nodeIsBarrier` from `DefaultTaintTracking.qll`, right? If we're emulating that, do `nodeIsBarrierIn` and `additionalTaintStep` from `DefaultTaintTracking.qll` matter as well?

Yep, it's a copy of `nodeIsBarrier` from `DefaultTaintTracking.qll`. `isAdditionalTaintStep` shouldn't matter as those taint steps should already be provided by `semmle.code.cpp.ir.dataflow.internal.TaintTrackingUtil.qll`.

`nodeIsBarrierIn` is more complicated. I hope we don't need them as I don't like the whole concept of `unpredictable` instructions. Let's see what DCA tells us about the result changes (if any).

MathiasVP: > Also it might be good to have a note explaining their history and that they may or may not actually be helping the query at this point - so that future maintainers can potentially clean things up without being overly confused/paranoid.

It definitely is changing the query results. It's blocking a lot of flow. Most of them are flows we do want to block, but sometimes it's also the cause of `[NOT DETECTED]` results.

cpp/ql/src/Critical/OverflowDestination.ql
MathiasVP: Well spotted! Fixed in f2676968f01eca4a1f89f9187fa197dbeb9cd0c4. I had to change add a couple of `getFullyConverted` to remove some duplicated results.

cpp/ql/src/Security/CWE/CWE-129/ImproperArrayIndexValidation.ql
MathiasVP: It'll not affect performance. I added this sanitizer because I found it odd that we were handling sanitizers at the sink node, but not in any of the intermediary nodes. If it affects any results it's for better.

geoffw0:
LGTM runs...

`cpp/overflow-destination`
 - before: https://lgtm.com/query/7063513232700019020/
 - after: https://lgtm.com/query/5954408493628011040/

`cpp/unclear-array-index-validation`
 - before: https://lgtm.com/query/7161980796863928292/
 - after: https://lgtm.com/query/1344744053190000681/

`cpp/uncontrolled-allocation-size`
 - before: https://lgtm.com/query/8782095103494040182/
 - after: https://lgtm.com/query/5576401620571102812/

MathiasVP:
> LGTM runs...
> 
> `cpp/overflow-destination`
> 
> * before: https://lgtm.com/query/7063513232700019020/
> * after: https://lgtm.com/query/5954408493628011040/
> 
> `cpp/unclear-array-index-validation`
> 
> * before: https://lgtm.com/query/7161980796863928292/
> * after: https://lgtm.com/query/1344744053190000681/
> 
> `cpp/uncontrolled-allocation-size`
> 
> * before: https://lgtm.com/query/8782095103494040182/
> * after: https://lgtm.com/query/5576401620571102812/

Thanks a lot! I was just in the process of creating diff queries myself.

geoffw0:
They're not diff queries - I decided that viewing the difference in presentation (`path-problem`) was more important in this case.

MathiasVP:
> They're not diff queries - I decided that viewing the difference in presentation (`path-problem`) was more important in this case.

Yeah, that's fair. I stripped out the path-problem stuff in `cpp/overflow-destination` and created a diff like this: https://lgtm.com/query/2343590706647795193/, but I'm totally okay with just using two runs and comparing manually.

geoffw0:
On `cpp/overflow-destination`:
 - paths show where taint comes from, and for this query taint is only used to 'filter' results to the most dangerous cases - so it isn't as relevant as for other queries.  Still nice to have.
 - 2 new results for `gpac/gpac`; LGTM.
 - 3 lost results for `alibaba/AliSQL` (the last 3 in `page0zip.cc`)
 - 6 new results for `arvidn/libtorrent`; they're displaying as a table not as alerts, I think they're outside the source location?
 - 1 new result for `kamailio/kamalio` (the one in `parser_carrierroute.c`); LGTM.
 - 1 new result for `haiku/haiku` (the one in `pool.c`); LGTM.
 - 1 new result for `pure-data`; LGTM.
 - 2 new results for `libretro-uae` (in `dr_flac.h` and `stdstring.c`); this first is a little weak IMO but not wrong, the second LGTM.
 - 1 lost result for `lsds/sgx-lkl` (the one in `symbol.c`)
 - overall the new results look good, I suspect they're due to wider sources; the lost results are sparse and don't concern me.
 - this requires a change note, and I think we can claim wider results as well as paths.

geoffw0:
On `cpp/uncontrolled-allocation-size`:
 - 1 new result in `vim/vim` (line 741); very long path but looks plausible.
 - 1 new result in `tesseract-ocr/tesseract` (in `bitvec.h`); similar to the above.
 - 2 lost results in `aws/s2n-tls`; perhaps we don't support / model `fscanf` as a source?
	- TODO: if this is true, create an issue for it and note the results we expect to probably recover.
 - 2 new results in `gpac/gpac`; look OK I think.  `str2ulong` is terrifying, perhaps we didn't track taint through that previously???
 - 10 new results in `alibaba/AliSQL` (all but the one in `my_malloc.c`); seem plausible.
 - 1 new result in `apache/trafficserver` (`CacheTool.cc`); the path goes through a bunch of largely irrelevant template code; the `reinterpret_cast` on line 739 is the key; looks correct and impressive!
 - I stopped reading at this point; there are quite a lot of new results but they seem good to me.

geoffw0:
> It may also be due to the removal of the barriers on "unpredictable" instructions. If we like the results I don't think we should include that barrier to remove these results.

You might be right, specifically the case:
```
// don't use dataflow through calls to pure functions if two or more operands
// are unpredictable
```
which might mean we now get taint flowing through things like pointer differences (`endOfBuffer - pointerIntoBuffer`) where we might not have done before.  I think I'm OK with the results like that I saw for `cpp/overflow-destination`.

geoffw0:
On `cpp/unclear-array-index-validation`:
 - this is the nosiest of the three queries and has got noisier (results in 83 -> 143 of 184 projects)
	- that's a bit of a warning sign that something may be wrong with the query.
	- though it lacks a `@precision` tag, and is probably not used all that much.
		- I'd be happy if you want to give it `@precision low`.  That's neater and I don't _think_ will mean it suddenly gets run in one of our standard suites?
 - 2 new results in `bitcoin/bitcoin`
	- in `uint256.cpp`, I think taint is flowing through a `char` type that may not be appropriate
		- were there any barriers based on `char` / small types previously?
	- in `system.cpp`, are we tainting `i` from `i < argc`?  If not, I'm not sure what's going on.
 - I'd like to improve this one a bit.

MathiasVP:
> I'd be happy if you want to give it @precision low. That's neater and I don't think will mean it suddenly gets run in one of our standard suites?

Indeed, I think we should give it a `@precision` low. I can do that in this PR.

> - 2 new results in bitcoin/bitcoin
>   - in uint256.cpp, I think taint is flowing through a char type that may not be appropriate
>     - were there any barriers based on char / small types previously?
>   - in system.cpp, are we tainting i from i < argc? If not, I'm not sure what's going on.
> - I'd like to improve this one a bit.

There are no barriers based on small types in `DefaultTaintTracking`, no. With that said, that might be a reasonable approach to filtering out some FPs from this query.

I might start by porting over the unpredictable instruction barrier to _just_ this one query. I don't want to do too much work on improving the FP rate of a `@precision low` query, but I think we can improve it a lot with reasonable little work.

MathiasVP:
> * I'd be happy if you want to give it `@precision low`.  That's neater and I don't _think_ will mean it suddenly gets run in one of our standard suites?

Fixed in 693eca21797a695362c944a70c79d6ea9491eba0.

MathiasVP:
@geoffw0 all the new results on `cpp/unclear-array-index-validation` happened because I incorrectly changed the sink in the query to be the entire array expression (instead of the offset expression) [here](https://github.com/github/codeql/pull/8382/commits/2328898b1973fe5e9c60ffaae8a030ec6e4a9779#diff-89cde0214c18fcc281f9b01da00e1e1581a2f251963252eee1c26b1c7eea0772R81)). I've fixed this in 0d3e47bcae8c3d44fb5214a11fffd71025413688. 

We now get a much more manageable difference. [Here's](https://lgtm.com/query/1297864307648970310/) a diff query and [here](https://lgtm.com/query/289967874546147215/) are all the results. I think they look good!

geoffw0 (dismissed):
Yep, looks much better. :+1:

There are still false positives for that query to do with characters being checked against (presumably) 256-long arrays, but I don't think we need to fix that as part of this work.

MathiasVP:
DCA showed a giant number of new results on `cpp/unclear-array-index-validation` (likely due to the bug I introduced in https://github.com/github/codeql/pull/8382/commits/2328898b1973fe5e9c60ffaae8a030ec6e4a9779, but which I then fixed in https://github.com/github/codeql/pull/8382/commits/0d3e47bcae8c3d44fb5214a11fffd71025413688). I'll start another DCA run on the HEAD of this PR and merge it if the results look better.

geoffw0 (dismissed):
Still :+1:

MathiasVP:
I've created the internal PR with the required changes.

cpp/ql/src/Security/CWE/CWE-129/ImproperArrayIndexValidation.ql
```diff
@@ -32,11 +36,92 @@ predicate linearBoundControls(BasicBlock controlled, SsaDefinition def, StackVar
...
+      readsVariable(node.asInstruction(), checkedVar) and
+      hasUpperBoundsCheck(checkedVar)
+    )
+    or
+    exists(Variable checkedVar, Operand access |
+      readsVariable(access.getDef(), checkedVar) and
+      nodeIsBarrierEqualityCandidate(node, access, checkedVar)
+    )
+    or
+    // Don't use dataflow into binary instructions if both operands are unpredictable
```
geoffw0, line 88: I do wonder what case this is supposed to cover besides subtraction ... but the existence of the case below as well suggests there's more to it.

geoffw0 (approved):
Changes LGTM.

cpp/ql/src/Security/CWE/CWE-129/ImproperArrayIndexValidation.ql
MathiasVP: Yeah, these two barriers block flow in ~100 examples on `vim/vim`. For this first case, it's most often subtraction/addition. The next case also handles stuff like getting the index of some character in a non-constant string (which they know is bounded due to invariants on the string). I imagine we could do something much more clever here if we want to, but that should probably be once we decide to invest effort in rewriting this query to something much more modern.

PR #8385, opened 2022-03-09
QL: add query detecting ordering by a constant

erik-krogh:
This was a mistake I made myself recently.  
I never committed the mistake, but that could have happened.  

No results in `github/codeql` :tada:


MathiasVP (approved):
LGTM!

PR #8390, opened 2022-03-10
C++: Remove uniqueness constraint from uuid

redsun82:
Different class definitions can have the same uuid. This happens for
example when using `#import <msxml6.dll>`: there will be several C++
classes generated in `msxml6.tlh` which will share uuids with
`extern "C"` struct declarations in the system header `msxml.h`.

Notice that as far as the standard cpp QL library and queries go, we
expose `getUuid()` on `UserType` and we never try to invert it, so we
only rely on uniqueness of the `id` column in the `usertype_uuid` table,
not the `uuid` column.

Closes github/codeql-c-team#893


jketema (approved):
LGTM. I also checked if have any queries that use `getUuid()` but that doesn't seem to be the case. So this won't break anything from that perspective.

PR #8391, opened 2022-03-10
C#: Retire the StructuralComparisonConfiguration interface

michaelnebel:
In this PR we remove the StructuralComparisonConfiguration interface as it is no longer needed.
Instead one can directly compare the global value numbers of the control flow expressions.


michaelnebel:
What is the default approach in such scenarios, when it comes to change notes? Should this be documented by a change note or is the deprecation description enough in itself?

hvitved:
> What is the default approach in such scenarios, when it comes to change notes? Should this be documented by a change note or is the deprecation description enough in itself?

I would think that getting deprecation warnings should be enough.

csharp/ql/src/Language Abuse/MissedTernaryOpportunity.ql
```diff
@@ -12,25 +12,12 @@
...
+private Expr getAssignedExpr(Stmt stmt) {
+  result = stmt.stripSingletonBlocks().(ExprStmt).getExpr().(AssignExpr).getLValue()
+}
 
-  IfStmt getIfStmt() {
-    exists(AssignExpr ae | ae = result.getThen().stripSingletonBlocks().(ExprStmt).getExpr() |
-      same(ae.getLValue(), _)
-    )
-  }
+private predicate ifStmt(IfStmt is) {
```
hvitved, line 19: I would just inline this below

hvitved (commented):
Looks really great. Just one minor suggestion.

hvitved (approved):

PR #8393, opened 2022-03-10
Improve join order for RangeAnalysis::boundedPhi

smowton:


java/ql/lib/semmle/code/java/dataflow/RangeAnalysis.qll
```diff
@@ -734,7 +744,11 @@ private predicate boundedPhi(
   SsaPhiNode phi, Bound b, int delta, boolean upper, boolean fromBackEdge, int origdelta,
   Reason reason
 ) {
-  forex(SsaVariable inp, SsaReadPositionPhiInputEdge edge | edge.phiInput(phi, inp) |
+  // Forcibly split into an existence and a forall stage instead of using forex because otherwise
```
hvitved, line 747: I am surprised to see that this doesn't use the "ranking approach" as in e.g. the [modulus library](https://github.com/github/codeql/blob/main/java/ql/lib/semmle/code/java/dataflow/ModulusAnalysis.qll#L162-L184). Replacing recursion through `forall`/`forex` with recursion over ranked indices typically performs much better.

smowton: Like https://github.com/github/codeql/pull/8405 ? Which do you prefer?

hvitved: Yes. I prefer that PR over this one.

hvitved: (this one also appears to have a significant slow-down on `palatable/lambda`).

smowton:
Dropped in favour of #8405

PR #8395, opened 2022-03-10
Ruby: add `rb/clear-text-storage-sensitive-data` query

alexrford:
This is https://github.com/github/codeql/pull/8306 again, which I'm unable to retarget against main and reopen.

The majority of this query is identical to `rb/clear-text-logging-sensitive-data`, with the main differences being in the sinks (file writes and other persistent writes rather than logging calls), and in reporting alerts at the sink rather than the source to make it easier to classify FPs with flow from a test file. The latter change may also be useful for the logging version of this query, but it appears to be a more common case with this query.

As this changes the shared [`SensitiveDataHeuristics.qll`](https://github.com/github/codeql/compare/main...alexrford:ruby/clear-text-storage?expand=1#diff-d462865e69b774f672162b8189402c7c07d8f0dafd08473c8e4640e5affdf1a2) library, it may have a minor impact on results for JS/Python queries relying on the `notSensitiveRegexp/0` predicate. See the comments at https://github.com/github/codeql/pull/8306#discussion_r822207620 for some context on this change.


yoff:
> As this changes the shared [`SensitiveDataHeuristics.qll`](https://github.com/github/codeql/compare/main...alexrford:ruby/clear-text-storage?expand=1#diff-d462865e69b774f672162b8189402c7c07d8f0dafd08473c8e4640e5affdf1a2) library, it may have a minor impact on results for JS/Python queries relying on the `notSensitiveRegexp/0` predicate. See the comments at [#8306 (comment)](https://github.com/github/codeql/pull/8306#discussion_r822207620) for some context on this change.

I guess a suitable DCA run could give some visibility into potential changes. Otherwise Python looks fine to me. @RasmusWL  has worked on sensitive data and may have an opinion.

erik-krogh (approved):
The only JS/Python change seems to be adding `"random"` to `notSensitiveRegexp()`, which seems fine to me.  

JS :+1:

alexrford:
> I guess a suitable DCA run could give some visibility into potential changes. Otherwise Python looks fine to me. `@RasmusWL` has worked on sensitive data and may have an opinion.

We lose one result in test code for `py/clear-text-storage-sensitive-data` in the standard code scanning set. It's not the kind of FP that this change is directly intended to avoid, where this random data represents some hash value. However, the result is low value, being in test code, and doesn't look to me like it would be a common pattern for production code.

RasmusWL (approved):
:+1: for Python

> > I guess a suitable DCA run could give some visibility into potential changes. Otherwise Python looks fine to me. `@RasmusWL` has worked on sensitive data and may have an opinion.
> 
> We lose one result in test code for `py/clear-text-storage-sensitive-data` in the standard code scanning set. It's not the kind of FP that this change is directly intended to avoid, where this random data represents some hash value. However, the result is low value, being in test code, and doesn't look to me like it would be a common pattern for production code.

Agreed, that result was not interesting :+1:

hmac (approved):

PR #8396, opened 2022-03-10
Ruby: resolve `ql/field-only-used-in-charpred` alerts

alexrford:
The first commit replaces these fields with `exists` or `don't-care`s in the charpred. The second commit simplifies a few of these charpreds into ones that should be equivalent in practice, with `KernelMethodCall` needing the most scrutiny.

### DIFF
--- a/ruby/ql/lib/codeql/ruby/ast/internal/Scope.qll
+++ b/ruby/ql/lib/codeql/ruby/ast/internal/Scope.qll
@@ -184,7 +184,7 @@ abstract class ScopeImpl extends AstNode, TScopeType {
 }
 
 private class ScopeRealImpl extends ScopeImpl, TScopeReal {
-  ScopeRealImpl() { exists(Scope::Range range | range = toGenerated(this)) }
+  ScopeRealImpl() { toGenerated(this) instanceof Scope::Range }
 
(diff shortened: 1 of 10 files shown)


yoff (dismissed):
Python 👍

erik-krogh (approved):
JS :+1:

ruby/ql/lib/codeql/ruby/ast/internal/Scope.qll
```diff
@@ -184,9 +184,7 @@ abstract class ScopeImpl extends AstNode, TScopeType {
 }
 
 private class ScopeRealImpl extends ScopeImpl, TScopeReal {
-  private Scope::Range range;
-
-  ScopeRealImpl() { range = toGenerated(this) }
+  ScopeRealImpl() { exists(Scope::Range range | range = toGenerated(this)) }
```
nickrolfe, line 187: As the QL-for-QL warning suggests, this can be
```suggestion
  ScopeRealImpl() { toGenerated(this) instanceof Scope::Range }
```

nickrolfe (commented):
Just one suggestion. Otherwise, LGTM.

erik-krogh (approved):

nickrolfe (approved):

PR #8399, opened 2022-03-11
Ruby: implement getComponent(n) for simple and hash-key symbols

nickrolfe:
This fixes a bug where `StringlikeLiteralCfgNode::getConstantValue()` would return both the correct symbol `:foo` and the empty symbol `:` for simple symbols and hash-key symbols. That's because it assumes `getComponent(n)` is implemented for every string-like literal type, and it concatenates those child components' constant values. (It uses `concat`, rather than `strictconcat`, so that it correctly returns the empty string for empty string literals (`""`) that have zero child components).

I recommend reviewing the two commits separately, to better see the effect this has.

~One unintended side-effect is that we lost some results for `Expr::getConstantValue()` where the expressions are not in the CFG. It's expected that stringlike literals that are not reachable will not have results for `getConstantValue()`, but it might indicate a bug in the CFG for pattern matching (unrelated to this PR).~ See comment below.


nickrolfe:
I discussed the CFG issue with @aibaars. It's expected that keys in pattern matching expressions aren't in the CFG (since they must be simple literals).

We also agreed that even symbols that are not reachable in the CFG should have results for `getConstantValue()`, so I've added back the overrides `SimpleSymbolLiteral::getConstantValue()` and HashKeySymbolLiteral::getConstantValue()`.

hvitved (dismissed):

hvitved (approved):

PR #8411, opened 2022-03-11
QL: add restrictive transitive closure query

erik-krogh:
Identifies the mistake fixed in this PR: https://github.com/github/codeql/pull/8380 

Basically, a transitive closure might restrict the type when you do zero steps through it, which might not be intentional.  
I think the results found by the query are benign (except for the one already fixed).  

---- 

Drive-by fix to the callgraph, example below:   

```CodeQL
module Foo {
  predicate bar() {none()}
}
class C {
  predicate bar() {none()}
  predicate test() {
    Foo::bar() // <- would spuriously resolve to C::bar() (on top of Foo::bar()). 
  }
}
```

### DIFF
--- a/ql/ql/src/queries/bugs/RestrictiveClosure.ql
+++ b/ql/ql/src/queries/bugs/RestrictiveClosure.ql
@@ -1,0 +1,53 @@
+/**
+ * @name Restrictive transitive closure
+ * @description The transitive closure operation might restrict the type even when taking zero steps.
+ * @kind problem
+ * @problem.severity warning
+ * @id ql/restrictive-transitive-closure
+ * @tags correctness
+ *       maintainability
+ * @precision high
+ */
+
+import ql
+
+Class base(Call c) {
+  result = c.(MemberCall).getBase().getType().getDeclaration()
+  or
+  not c instanceof MemberCall and
+  result = c.getTarget().(ClassPredicate).getParent()
+}
+
+Call closureCall() { result.isClosure("*") }
+
+Class return(Call c) { result = c.getTarget().getReturnType().getDeclaration() }
+
+Class superClass(Class c) { result = c.getASuperType().getResolvedType().getDeclaration() }
+
+from Call c, Class return, Class base
+where
+  c = closureCall() and
+  return = return(c) and
+  base = base(c) and
+  return != base and
+  // We aren't restricted anyway from the sorounding code.
+  not superClass*(base) = return and
+  not exists(InstanceOf inst |
+    inst.getExpr() = c and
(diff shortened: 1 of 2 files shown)


ql/ql/src/queries/bugs/RestrictiveClosure.ql
```diff
@@ -0,0 +1,53 @@
...
+
+Class superClass(Class c) { result = c.getASuperType().getResolvedType().getDeclaration() }
+
+from Call c, Class return, Class base
+where
+  c = closureCall() and
+  return = return(c) and
+  base = base(c) and
+  return != base and
+  // We aren't restricted anyway from the sorounding code.
```
tausbn, line 33:
```suggestion
  // We aren't restricted anyway from the surrounding code.
```

ql/ql/src/queries/bugs/RestrictiveClosure.ql
```diff
@@ -0,0 +1,53 @@
...
+  not superClass*(base) = return and
```
tausbn, line 34: Am I missing something subtle here, or does this not imply `return != base` (from line 32)?

ql/ql/src/queries/bugs/RestrictiveClosure.ql
```diff
@@ -0,0 +1,53 @@
...
+    operand = comp.getAnOperand() and
+    superClass*(operand.getType().getDeclaration()) = return
+  ) and
+  // If the result is used in a call, then we only flag if the "closure in the middle" could be removed.
+  not exists(MemberCall memberCall | memberCall.getBase() = c |
+    not exists(ClassPredicate pred |
+      pred = superClass*(base).getClassPredicate(memberCall.getMemberName()) and
+      memberCall.getNumberOfArguments() = pred.getArity()
+    )
+  )
```
tausbn, line 51: `not exists ... not` ... Is this just a `forall`?

tausbn (changes requested):
A few comments, otherwise this looks good to me.

Out of interest, is there a "canonical" fix for the case where you _don't_ want the type to be restricted?

ql/ql/src/queries/bugs/RestrictiveClosure.ql
erik-krogh: That escalated quickly... 
https://github.com/github/codeql/pull/9832

erik-krogh:
> Out of interest, is there a "canonical" fix for the case where you don't want the type to be restricted? 

The fix I went for was to replace: `stepPred*(x)` with `[x, stepPred+(x)]`.   
That fix should work in all cases.

PR #8421, opened 2022-03-13
Ruby: Add `rb/weak-cryptographic-algorithm` query

alexrford:
Ports `py/weak-cryptographic-algorithm` to Ruby. Currently supports `OpenSSL` ciphers.

This is generally similar, using the `Cryptography::CryptographicOperation` concept. There is a slight tweak in moving the `isWeak()` predicate into `CryptographicOperation` rather than only having it on `CryptographicAlgorithm`, which allows us to account for operations with ciphers that use a weak block mode (in practice, this means `ECB`). The `CryptographicAlgorithm` only includes the block encryption algorithm itself, not the block mode in the context where the algorithm is used in a stream cipher.

The side-effect of this is that the alert messages can currently be misleading, e.g. suggesting that the block encryption algorithm is weak when it is actually the block mode.

The best solutions to this that I can see are:
 1. Keeping `CryptographicAlgorithm` and the block mode separate, but introducing a proper `BlockMode` class with e.g. an `isWeak()` predicate of its own. `CryptographicOperation` could then have a `BlockMode getBlockMode()` predicate rather than its own `isWeak()` predicate, and the query could check both the encryption algorithm and the block mode.
 2. Include the block mode in the `CryptographicAlgorithm`, and add 2 predicates - `isWeakEncryptionAlgorithm` and `isWeakBlockMode`, to that class. Then `CryptographicOperation#isWeak()` could be removed.


ruby/ql/lib/codeql/ruby/Concepts.qll
```diff
@@ -762,3 +762,58 @@ module Logging {
     abstract DataFlow::Node getAnInput();
   }
 }
+
+/**
+ * Provides models for cryptographic things.
```
aibaars, line 767: Perhaps `things` -> `concepts`.

ruby/ql/lib/codeql/ruby/security/OpenSSL.qll
```diff
@@ -308,4 +312,189 @@ class OpenSSLCipher extends MkOpenSSLCipher {
...
+          else
+            // `OpenSSL::Cipher::<blockAlgo>.new` uses CBC by default
+            blockMode = "CBC"
+        |
+          cipherName = blockAlgo + "-" + blockMode and
+          cipherMode.isBlockMode(blockMode)
+        )
+      )
+    |
+      cipher.matchesName(cipherName)
```
aibaars, line 440: I think this can be moved to line 381 . 

```
exists(string cipherName |  cipher.matchesName(cipherName) |
    ...
    or
    ...
)

File: ruby/ql/lib/codeql/ruby/security/OpenSSL.qll [hunk 3]
```diff
@@ -308,4 +312,189 @@ class OpenSSLCipher extends MkOpenSSLCipher {
...
+
+  CipherInstantiation() {
+    exists(string cipherName |
+      // `OpenSSL::Cipher.new('<cipherName>')`
+      this = cipherApi().getAnInstantiation() and
+      cipherName = this.getStringArgument(0) and
+      // CBC is used by default
+      cipherMode.isBlockMode("CBC")
+      or
+      // `OpenSSL::Cipher::AES` instantiations
```
aibaars, line 388: This is quite a large block of text and I wonder if it would improve readability if split into smaller predicates.

aibaars (commented):
This looks good to me.

ruby/ql/lib/codeql/ruby/security/OpenSSL.qll
alexrford: I've split this into 4 separate predicates that each cover a subset of cases, I think it's slightly clearer.

ruby/ql/lib/codeql/ruby/Concepts.qll
```diff
@@ -762,3 +762,58 @@ module Logging {
...
+  import security.CryptoAlgorithms
+
+  /**
+   * A data-flow node that is an application of a cryptographic algorithm. For example,
+   * encryption, decryption, signature-validation.
+   *
+   * Extend this class to refine existing API models. If you want to model new APIs,
+   * extend `CryptographicOperation::Range` instead.
+   */
+  class CryptographicOperation extends DataFlow::Node {
```
nickrolfe, line 784: I think you can make this `instanceof CryptographicOperation::Range`, and remove the `range` field.

ruby/ql/test/query-tests/security/cwe-327/BrokenCryptoAlgorithm.expected
```diff
@@ -0,0 +1,18 @@
+| broken_crypto.rb:4:8:4:34 | call to new | The cryptographic algorithm DES is broken or weak, and should not be used. |
+| broken_crypto.rb:8:1:8:18 | call to update | The cryptographic algorithm DES is broken or weak, and should not be used. |
+| broken_crypto.rb:12:8:12:43 | call to new | The cryptographic algorithm AES is broken or weak, and should not be used. |
```
nickrolfe, line 3: It's a bit of a shame that this message doesn't mention the specific cipher mode that makes it weak. Would that be difficult to add?

nickrolfe (commented):
Sorry, I know you already had a review and were just looking for re-approval, but I have a couple of comments.

ruby/ql/lib/codeql/ruby/security/OpenSSL.qll
```diff
@@ -308,4 +312,241 @@ class OpenSSLCipher extends MkOpenSSLCipher {
...
+   */
+  bindingset[s]
+  predicate isBlockMode(string s) { this.getBlockMode() = s.toUpperCase() }
+
+  /** Holds if this cipher mode is a weak block mode. */
+  predicate isWeak() { isWeakBlockMode(this.getBlockMode()) }
+}
+
+private string getStringArgument(DataFlow::CallNode call, int i) {
+  result = call.getArgument(i).asExpr().getConstantValue().getStringOrSymbol()
```
nickrolfe, line 367: Here's why you're getting a CI failure: I deprecated `getStringOrSymbol` and renamed it `getStringlikeValue`.

ruby/ql/test/query-tests/security/cwe-327/BrokenCryptoAlgorithm.expected
alexrford: Yeah, the inaccurate message here is not great. At the level of `CryptographicOperation`, we have no concept of the block mode, so the weakness/strength of the operation is just a kind of black box. The solutions that I've considered are:
 1. Keeping `CryptographicAlgorithm` and the block mode separate, but introducing a proper `BlockMode` class with e.g. an `isWeak()` predicate of its own. `CryptographicOperation` could then have a `BlockMode getBlockMode()` predicate rather than its own `isWeak()` predicate, and the query could check both the encryption algorithm and the block mode.
 2. Include the block mode in the `CryptographicAlgorithm`, and add 2 predicates - `isWeakEncryptionAlgorithm` and `isWeakBlockMode`, to that class. Then `CryptographicOperation#isWeak()` could be removed.

I think I prefer `1` overall as it seems like the intention of the `CryptographicAlgorithm` class (shared with JS and Python) is to represent the algorithm modulo any block mode. Do you have any preferences between these?

nickrolfe (approved):
LGTM. I'll leave it up to you whether you address my comment about the block mode in the alert message.

ruby/ql/test/query-tests/security/cwe-327/BrokenCryptoAlgorithm.expected
nickrolfe: 1 sounds like a nice solution, if it's not too difficult to implement.

PR #8323, opened 2022-03-03
Enforcing consistent casing of acronyms

erik-krogh:
[Acronyms should be PascalCase/camelCase](https://github.com/github/codeql/blob/main/docs/ql-style-guide.md#naming).   

I made a patch for converting the casing of acronyms to PascalCase.  
At first I thought it would just be a fun day-of-learning exercise that shouldn't be merged.  
However, naming conflicts started showing up (see list below), and now I think this PR should be merged to avoid e.g. having both a `XmlComment` and `XMLComment` class in the same scope. 

## Conflicts detected from this: 

There are a few places where only the casing on names were different, and those started conflicting with this patch (because they got imported into the same scope).     
They are: 

1) JS: Two (effectively identical) implementations of `isDomRootType(..)`. ([one](https://github.com/github/codeql/blob/9a91e66714864eedbc8b71873875cd5678115945/javascript/ql/lib/Expressions/DOMProperties.qll#L8-L10), [two](https://github.com/github/codeql/blob/9a91e66714864eedbc8b71873875cd5678115945/javascript/ql/lib/semmle/javascript/security/dataflow/DOM.qll#L8-L10)). I simply deleted one of them. 
2) Java: Two identical implementations of `UrlOpenStreamMethod`. ([one](https://github.com/github/codeql/blob/9a91e66714864eedbc8b71873875cd5678115945/java/ql/lib/semmle/code/java/frameworks/Networking.qll#L145-L150), [two](https://github.com/github/codeql/blob/9a91e66714864eedbc8b71873875cd5678115945/java/ql/src/experimental/Security/CWE/CWE-036/OpenStream.ql#L30-L35)). One of those were on a `.ql` file, so that one just got deleted.  
3) C#: Two `XmlComment` classes that did very different things. ([one](https://github.com/github/codeql/blob/9a91e66714864eedbc8b71873875cd5678115945/csharp/ql/lib/semmle/code/csharp/Comments.qll#L69), [two](https://github.com/github/codeql/blob/9a91e66714864eedbc8b71873875cd5678115945/csharp/ql/lib/semmle/code/csharp/XML.qll#L310)). One of them could be renamed to `XmlCommentLine`, which I think is a more fitting name.  

## The code in this PR:  

- This PR first introduces a QL-for-QL query for detecting use of upper-case acronyms in names.  
- Then some of those got automatically patched (see backref)
   - a deprecated alias for the old name were added.   
   - QLDoc that mentions one of the renaming got updated. 
[...]

PR #8422, opened 2022-03-13
JS: Address some code that weren't affecting any query result

erik-krogh:
As a followup to https://github.com/github/codeql/pull/8347 I checked if some code had become dead.  
There was no **new** dead code, but I found some interesting things regardless.  

1) Some code doesn't affect query result.
    That is partly expected (e.g. a lot of classes in `Expr.qll`/`TypeScript.qll` are not used anywhere).  
    But there were also some unexpected result, like a few `Sanitizer` classes that weren't referenced anywhere.  
2) Code that was already dead. That is, code that doesn't affect any query, and isn't referenced by any public API. 
    (Not this PR, I'll address that in another PR). 

This query partially addresses 1) in JS.   
There's still plenty of code that doesn't affect any query, but all the remaining seems benign.


esbena (commented):
Interesting findings. Thanks once again for doing this kind of cleanup!

One nit: Should all of the deprecations have a qldoc string pointing the users to the alternative?

---

Also, please confirm that https://github.com/github/codeql/blob/4fc85a791d45d931ec6dac0304e8c314eb0ba23a/javascript/ql/test/library-tests/Classes/tests.ql contains the inlined query predicate version of the files deleted in https://github.com/github/codeql/pull/8422/commits/5e52a710912b244bd4195b0ebc9de67a058ba04a.

erik-krogh:
> One nit: Should all of the deprecations have a qldoc string pointing the users to the alternative?

:+1: 

> Also, please confirm that https://github.com/github/codeql/blob/4fc85a791d45d931ec6dac0304e8c314eb0ba23a/javascript/ql/test/library-tests/Classes/tests.ql contains the inlined query predicate version of the files deleted in [5e52a71](https://github.com/github/codeql/commit/5e52a710912b244bd4195b0ebc9de67a058ba04a).

Confirmed :+1:

esbena:
We need a change note for the same reason as in https://github.com/github/codeql/pull/8323#issuecomment-1060462937: 

> If people are going to start seeing deprecated warnings they weren't seeing before, this definitely requires a change note.

esbena (approved):

PR #8425, opened 2022-03-14
C#: Avoid combinatorial explosion in structural comparison library

hvitved:
In cases where the target of a call/access has multiple values (which is a DB inconsistency), the GVN construction underlying the structural comparision library may run into a combinatorial explosion. This change excludes such expressions from the GVN construction.


michaelnebel:
How was this discovered?

michaelnebel (approved):
LGTM

hvitved:
> How was this discovered?

It was discovered during the current distribution upgrade.

PR #8428, opened 2022-03-14
C++: Handle C11 _Noreturn in DefaultOptions

jketema:


cpp/ql/lib/semmle/code/cpp/Specifier.qll
```diff
@@ -38,7 +38,7 @@ class FunctionSpecifier extends Specifier {
 
 /**
  * A C/C++ storage class specifier: `auto`, `register`, `static`, `extern`,
- * or `mutable".
+ * or `mutable`.
```
geoffw0, line 41: Good spot.

geoffw0 (dismissed):
LGTM.

cpp/ql/lib/semmle/code/cpp/Specifier.qll
jketema: Took me a while to realize why it was not highlighting in VS Code 😄

geoffw0 (dismissed):

geoffw0 (approved):

PR #8430, opened 2022-03-14
Js/CVE 2022 24718

kaeluka:


javascript/ql/lib/semmle/javascript/frameworks/Handlebars.qll
```diff
@@ -27,3 +27,86 @@ module Handlebars {
...
+module TaintStep {
+  /**
+   * Gets a `SourceNode` tracked from a compilation of a Handlebars template.
+   */
+  private DataFlow::SourceNode compiledHandlebarsTemplate(DataFlow::Node originalCall) {
+    result = compiledHandlebarsTemplate(DataFlow::TypeTracker::end(), originalCall)
+  }
+
+  private DataFlow::SourceNode compiledHandlebarsTemplate(
+    DataFlow::TypeTracker t, DataFlow::Node originalCall
```
erik-krogh, line 41:
```suggestion
  private DataFlow::SourceNode compiledTemplate(DataFlow::CallNode compileCall) {
    result = compiledHandlebarsTemplate(DataFlow::TypeTracker::end(), compileCall)
  }

  private DataFlow::SourceNode compiledTemplate(
    DataFlow::TypeTracker t, DataFlow::CallNode compileCall
```
You might as well have a more specific type (`CallNode` instead of just `Node`).  
And you can rename `originalCall` to something more descriptive in the same go, I went for `compileCall`. 
We're in a `Handlebars` module in a `Handlebars` file, so I don't think we need to call this predicate something with `Handlebars`.

javascript/ql/lib/semmle/javascript/frameworks/Handlebars.qll
```diff
@@ -27,3 +27,86 @@ module Handlebars {
...
+      exists(DataFlow::CallNode templatingCall, DataFlow::CallNode compileCall |
+        templatingCall = compiledHandlebarsTemplate(compileCall).getACall() and
+        pred = templatingCall.getAnArgument().getALocalSource().getAPropertyWrite().getRhs() and
+        (
+          compileCall
+              .getArgument(0)
+              .mayHaveStringValue(any(string templateText |
+                  // an approximation: if the template contains the
+                  // helper name, it's probably a helper call
+                  templateText.matches("%" + helperName + "%")
```
erik-krogh, line 84: Would something more precise work here? 
I'm wondering whether we could match both "loud" and "name" in `{{loud name}}` (probably using a regexpCapture).  
That way we could match precisely which helper is invoked with which argument. 

(You should probably split some of this predicate out into a helper predicate).

javascript/ql/lib/semmle/javascript/frameworks/Handlebars.qll
```diff
@@ -27,3 +27,86 @@ module Handlebars {
...
+          // When we don't have a string value, we can't be sure
+          // and will assume a step.
+          not exists(string s | compileCall.getArgument(0).mayHaveStringValue(s))
+        )
+      ) and
+      exists(DataFlow::CallNode registerHelperCall, DataFlow::FunctionNode helperFunc |
+        registerHelperCall = any(Handlebars::Handlebars hb).getAMethodCall("registerHelper") and
+        registerHelperCall.getArgument(0).mayHaveStringValue(helperName) and
+        helperFunc = registerHelperCall.getArgument(1).getAFunctionValue() and
+        helperFunc.getAParameter() = succ
```
erik-krogh, line 96:
```suggestion
        succ = registerHelperCall.getArgument(1).getAFunctionValue().getAParameter()
```
You don't need `helperFunc`.

javascript/ql/lib/semmle/javascript/frameworks/Handlebars.qll
```diff
@@ -27,3 +27,86 @@ module Handlebars {
...
+  /**
+   * A shared flow step from passing data to a handlebars template with
+   * helpers registered.
+   */
+  class HandlebarsStep extends DataFlow::SharedFlowStep {
+    DataFlow::CallNode templatingCall;
+
+    override predicate step(DataFlow::Node node1, DataFlow::Node node2) {
+      isHandlebarsArgStep(node1, node2)
+    }
```
erik-krogh, line 110:
```suggestion
    override predicate step(DataFlow::Node pred, DataFlow::Node succ) {
      isHandlebarsArgStep(pred, succ)
    }
```

erik-krogh (commented):
Generally looks OK, but I got some comments.

javascript/ql/lib/semmle/javascript/frameworks/Handlebars.qll
kaeluka: I thought about that, too.

My thinking: the proposed implementation might lead to a problem (FP alert) when:

* there's a sink in a helper
* a template uses, in its text, the precise name of the helper (in this example, a template that contains the term `loud` as text) — leading to a false step.

I had originally decided against it because I thought that handlebar templates are rare enough and such a name is rare enough for an issue to happen.

But as you're also asking yourself this, I'll increase precision.

javascript/ql/lib/semmle/javascript/frameworks/Handlebars.qll
```diff
@@ -27,3 +27,86 @@ module Handlebars {
     SafeString() { this = any(Handlebars h).getAConstructorInvocation("SafeString") }
   }
 }
+
+/** Provides logic for taint steps for the handlebars library. */
+module TaintStep {
```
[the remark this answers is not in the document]
kaeluka: 👍

javascript/ql/lib/semmle/javascript/frameworks/Handlebars.qll
```diff
@@ -27,3 +27,86 @@ module Handlebars {
...
+  /**
+   * Gets a `SourceNode` tracked from a compilation of a Handlebars template.
```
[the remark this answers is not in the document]
kaeluka: 👍

javascript/ql/lib/semmle/javascript/frameworks/Handlebars.qll
kaeluka: 👍

javascript/ql/lib/semmle/javascript/frameworks/Handlebars.qll
kaeluka: well spotted :)

javascript/ql/lib/semmle/javascript/frameworks/Handlebars.qll
kaeluka: 👍

javascript/ql/lib/semmle/javascript/frameworks/Handlebars.qll
```diff
@@ -27,3 +27,126 @@ module Handlebars {
...
+   * In the example, the predicate will hold for:
+   *
+   * * helperName="catFile", argIdx=1, arg="path"
+   * * helperName="echo", argIdx=1, arg="p1"
+   * * helperName="echo", argIdx=2, arg="p2"
+   *
+   * The initial `{{path}}` will not be considered, as it has no arguments.
+   */
+  bindingset[templateText]
+  private predicate isTemplateHelperCallArg(
```
erik-krogh, line 85:
```suggestion
  private predicate isTemplateWithHelperCall(
``` 

A fully descriptive name might be something like: `isTemplateWithHelperCallThatHasArgument`, but that would be unwieldy.

javascript/ql/lib/semmle/javascript/frameworks/Handlebars.qll
```diff
@@ -27,3 +27,126 @@ module Handlebars {
...
+          compileCall.getArgument(0).mayHaveStringValue(templateText)
+        |
+          pred =
+            templatingCall.getAnArgument().getALocalSource().getAPropertyWrite(paramName).getRhs() and
+          isTemplateHelperCallArg(templateText, helperName, argIdx, paramName) and
+          isHelperParam(helperName, helperFunction, succ, argIdx)
+        )
+        or
+        // When we don't have a string value, we can't be sure
+        // and will assume a step.
```
erik-krogh, line 133:
```suggestion
        // and we assume a step to all parameters of all helpers. 
```

javascript/ql/lib/semmle/javascript/frameworks/Handlebars.qll
```diff
@@ -27,3 +27,126 @@ module Handlebars {
...
+    DataFlow::TypeTracker t, DataFlow::CallNode compileCall
+  ) {
+    t.start() and
+    result = any(Handlebars::Handlebars hb).getAMethodCall(["compile", "template"]) and
+    result = compileCall
+    or
+    exists(DataFlow::TypeTracker t2 | result = compiledTemplate(t2, compileCall).track(t2, t))
+  }
+
+  private predicate isHelperParam(
```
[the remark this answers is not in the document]
kaeluka: :+1:

javascript/ql/lib/semmle/javascript/frameworks/Handlebars.qll
```diff
@@ -27,3 +27,126 @@ module Handlebars {
...
+    string helperName, DataFlow::FunctionNode helperFunction, DataFlow::ParameterNode param,
+    int paramIndex
+  ) {
+    exists(DataFlow::CallNode registerHelperCall |
+      registerHelperCall = any(Handlebars::Handlebars hb).getAMethodCall("registerHelper") and
```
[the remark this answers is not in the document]
kaeluka: 🤯 👍

javascript/ql/lib/semmle/javascript/frameworks/Handlebars.qll
```diff
@@ -27,3 +27,126 @@ module Handlebars {
...
+      registerHelperCall.getArgument(0).mayHaveStringValue(helperName) and
+      helperFunction = registerHelperCall.getArgument(1).getAFunctionValue() and
+      param = helperFunction.getParameter(paramIndex)
+    )
+  }
+
+  /** Holds if `call` is a block wrapped inside curly braces inside the template `templateText`. */
+  bindingset[templateText]
+  predicate templateHelperParamBlock(string templateText, string call) {
```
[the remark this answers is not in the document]
kaeluka: 👍

javascript/ql/lib/semmle/javascript/frameworks/Handlebars.qll
kaeluka: 👍

javascript/ql/lib/semmle/javascript/frameworks/Handlebars.qll
```diff
@@ -27,3 +27,126 @@ module Handlebars {
...
+      string helperName, DataFlow::CallNode templatingCall, DataFlow::CallNode compileCall,
+      DataFlow::FunctionNode helperFunction
+    |
+      templatingCall = compiledTemplate(compileCall).getACall() and
+      (
+        exists(string templateText, string paramName, int argIdx |
+          compileCall.getArgument(0).mayHaveStringValue(templateText)
+        |
+          pred =
+            templatingCall.getAnArgument().getALocalSource().getAPropertyWrite(paramName).getRhs() and
```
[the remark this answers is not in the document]
kaeluka: you're right, this can be limited to `getArgument(0)`. But I'd rather not restrict the query to only handle object literals. 👍

javascript/ql/lib/semmle/javascript/frameworks/Handlebars.qll
kaeluka: true, a bit too unwieldy imo. I'll leave it. 👍

javascript/ql/lib/semmle/javascript/frameworks/Handlebars.qll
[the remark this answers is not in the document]
erik-krogh: You haven't limited to `getArgument(0)`, did you intend to do that?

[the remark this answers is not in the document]
kaeluka: Ha! This was the case at two locations and i had fixed the other! Now fixed in both cases 😜 what a day

javascript/ql/lib/semmle/javascript/frameworks/Handlebars.qll
```diff
@@ -27,3 +27,142 @@ module Handlebars {
...
+            templatingCall.getArgument(0).getALocalSource().getAPropertyWrite(paramName).getRhs() and
+          isTemplateHelperCallArg(templateText, helperName, argIdx, paramName) and
+          succ = getRegisteredHelperParam(helperName, helperFunction, argIdx)
+        )
+        or
+        // When we don't have a string value, we can't be sure
+        // and we assume a step to all parameters of all helpers.
+        not exists(string s | compileCall.getArgument(0).mayHaveStringValue(s)) and
+        pred = templatingCall.getArgument(0).getALocalSource().getAPropertyWrite().getRhs() and
+        succ = getRegisteredHelperParam(helperName, helperFunction, _)
```
erik-krogh, line 152: I'm not too sure if we should include this.  
There is a lot of potential for false flow here, because any `.compile()` call with an unknown string can cause flow to all helpers.  

Is the string a known constant in CVE-2022-24718?  

Looking at [some `handlebars.compile(..` calls](https://cs.github.com/?scopeName=All+repos&scope=&q=handlebars.compile%28+path%3A*.js) it's clear that there are plenty of calls where the string is unknown for various reasons.

erik-krogh (commented):
I did [an evaluation](https://github.com/github/codeql-dca-main/tree/data/erik-krogh/pr-8430-2cbb25a__handlebars-and-__code-scanning/reports).  
(`nightly-old` + a bunch of random projects that use `handlebars.compile(..`).  

I'm a bit worried about false flow in the case where we don't know the template string, but otherwise LGTM.

javascript/ql/lib/semmle/javascript/frameworks/Handlebars.qll
kaeluka: OK, removing this will be easy and have the added benefit to simplify the implementation. Will remove it ASAP after the vacation 👍

kaeluka: done 👍

javascript/ql/test/query-tests/Security/CWE-022/TaintedPath/handlebars.js
```diff
@@ -34,7 +34,7 @@ app.get('/some/path2', function (req, res) {
 });
 
 app.get('/some/path3', function (req, res) {
-    res.send(data.compiledUnknown({ name: req.params.name })); // NOT ALLOWED (could be using a vulnerable helper)
+    res.send(data.compiledUnknown({ name: req.params.name })); // ALLOWED (could be using a vulnerable helper, but we'll assume it's ok)
```
erik-krogh, line 37: We tend to use `// NOT OK` / `// OK`, or sometimes `// GOOD` / `// BAD`.  
(That also works with the consistency checking when we use that).

kaeluka: The alert location is in a different line (11), adding `NOT OK` here will break the existing consistency test. This comment is really only intended for a human

kaeluka:
@erik-krogh I think I've addressed all concerns.

PR #8431, opened 2022-03-14
Delete dead code

erik-krogh:
Deletes all dead code.  
As found by a QL-for-QL query that attempts to overapproximate what code is alive.  

Code is alive if: 
- It can affect a query result (the `from-where-select` or a `query predicate` imported by a `.ql` file). 
- It is part of a public API that can be imported from another `.qll` file. 
- Can affect or be referenced by any of the above (recursively). 
- (Other edge cases, see `DeadCodeQuery.qll`). 

Any code not marked as alive is dead, and is deleted by this PR.  
No test or external code should be affected by this PR. 

This PR is a followup to https://github.com/github/codeql/pull/8347, because I wanted to check if I had introduced dead code by deleting deprecations.    
(I had not, the code was dead already). 

---- 

There were some benign cases.  
E.g. in the shared DataFlow library a lot of code is copy-pasted between stages, and not all the predicates are used in all the stages. 

---- 

The dead code wasn't marked as unused by the CodeQL compiler for various reasons.   
- Recursive predicates that depend on themself. (Or mutually recursive predicates).  
- Public classes/predicates in private modules. 
- Public classes in `.ql` files that aren't used. 

---- 

The QL-for-QL code also contains predicate for detecting "unqueryable" code.   
That is: code that cannot affect any query result (but it might be part of the public API).   
I already found a bunch of [interesting results](https://github.com/github/codeql/pull/8422) using that code, but a QL-for-QL query needs more work to weed out benign results.


smowton (dismissed):
:+1: Java

esbena (dismissed):
JS :+1:

intrigus-lgtm:
Regarding this part of the PR:
https://github.com/github/codeql/pull/8431/files#diff-ae052bd5f05c791a2d3f7fbb99ab3a5a7fadac51704cceefef602db27a00686c

Why is `ConcreteSensitiveInsecureUrl` dead code?
https://github.com/github/codeql/blob/69353bb014fcbb3587da3bf7c2273ffc7d033bc6/javascript/ql/lib/semmle/javascript/security/dataflow/InsecureDownloadQuery.qll#L12-L15
A search for it only shows two results:
https://cs.github.com/github/codeql?q=%22ConcreteSensitiveInsecureUrl%22


But for `ConcretePosixPath` there are also only two results in the code search!
https://github.com/github/codeql/blob/6aad8d6897e8b24790a8be1338656c81e80f10f5/javascript/ql/lib/semmle/javascript/security/dataflow/ZipSlipQuery.qll#L13-L16
https://cs.github.com/github/codeql?q=%22ConcretePosixPath%22

**Why is the first considered dead code, while the last is not? What am I missing?**

erik-krogh:
> Why is the first considered dead code, while the last is not? What am I missing? 

The first extends a non-concrete class, the second extends an abstract class.  
[The documentation has more information about abstract classes.](https://codeql.github.com/docs/ql-language-reference/types/#abstract-classes).    
Relevant snippet: `an abstract class is defined as the union of its subclasses.`

hvitved (approved):
C#/`SsaImplCommon.qll` 👍

ruby/ql/lib/codeql/ruby/frameworks/Files.qll
```diff
@@ -189,12 +189,6 @@ module IO {
     }
   }
 
-  // "Direct" `IO` instances, i.e. cases where there is no more specific
-  // subtype such as `File`
-  private class IOInstanceStrict extends IOInstance {
```
nickrolfe, line 194: @alexrford you added this class originally. Is its deadness unintentional?

MathiasVP (approved):
C++ and QL 👍.

ruby/ql/lib/codeql/ruby/frameworks/Files.qll
alexrford: Yeah, I should have removed this as part of the refactoring in https://github.com/github/codeql/pull/8138

cpp/ql/lib/semmle/code/cpp/ir/dataflow/internal/SsaImplCommon.qll
```diff
@@ -287,20 +287,6 @@ private module SsaDefReaches {
     )
   }
 
-  /**
-   * Holds if the SSA definition of `v` at `def` reaches uncertain SSA definition
-   * `redef` in the same basic block, without crossing another SSA definition of `v`.
-   */
-  predicate ssaDefReachesUncertainDefWithinBlock(
```
geoffw0, line 294: Does this change preserve identical files?

MathiasVP: It does because all the identical files were modified :)

erik-krogh: Yes, the code is dead in all copies.

PR #8433, opened 2022-03-14
JS: Remove `isEffectiveSinkWithOverridingScore` from ML-powered libraries

henrymercer:
This was previously used in the ATM external API query, but is now dead code.


annarailton (approved):
Happy to MOG with this one.

henrymercer:
Evaluation shows no change in alerts and flat runtime, so merging.

PR #8435, opened 2022-03-14
Add flow state versions of isBarrierIn, isBarrierOut, and isBarrierGuard

jketema:
I've a use-case for `isBarrierOut` in my rewrite of `cpp/command-line-injection`, which needs a `isSanitizerOut` for a particular flow state. Others added for completeness.

### DIFF
--- a/cpp/ql/lib/semmle/code/cpp/dataflow/internal/DataFlowImpl.qll
+++ b/cpp/ql/lib/semmle/code/cpp/dataflow/internal/DataFlowImpl.qll
@@ -87,11 +87,29 @@ abstract class Configuration extends string {
   /** Holds if data flow into `node` is prohibited. */
   predicate isBarrierIn(Node node) { none() }
 
+  /**
+   * Holds if data flow into `node` is prohibited when the flow state is
+   * `state`
+   */
+  predicate isBarrierIn(Node node, FlowState state) { none() }
+
   /** Holds if data flow out of `node` is prohibited. */
   predicate isBarrierOut(Node node) { none() }
 
+  /**
+   * Holds if data flow out of `node` is prohibited when the flow state is
+   * `state`
+   */
+  predicate isBarrierOut(Node node, FlowState state) { none() }
+
   /** Holds if data flow through nodes guarded by `guard` is prohibited. */
   predicate isBarrierGuard(BarrierGuard guard) { none() }
+
+  /**
+   * Holds if data flow through nodes guarded by `guard` is prohibited when
+   * the flow state is `state`
+   */
+  predicate isBarrierGuard(BarrierGuard guard, FlowState state) { none() }
 
   /**
    * Holds if the additional flow step from `node1` to `node2` must be taken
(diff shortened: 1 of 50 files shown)


cpp/ql/lib/semmle/code/cpp/dataflow/internal/DataFlowImpl.qll
```diff
@@ -348,6 +384,17 @@ private predicate stateBarrier(NodeEx node, FlowState state, Configuration confi
   exists(Node n |
     node.asNode() = n and
```
MathiasVP, line 385: Code Scanning is not wrong here, I think 😂. I think this is the change you want to make?
```suggestion
    node.asNode() = n |
```

jketema: Indeed. Thanks.

MathiasVP (commented):
I think the changes LGTM! Let's wait and hear what @hvitved says, and maybe we should also do another round of DCA? I imagine the slowdowns are fixed by https://github.com/github/codeql/pull/8435/commits/e8a857d7c382d3f963e7c23297589c57130b0997, but it would be good to make sure!

jketema:
Not sure if it's worth doing all the DCA experiments again. A will re-do the Java one, as it's the only one that showed a slow down involving a query that uses taint tracking and/or data flow. And I'll do another attempt at a Python DCA run, which failed before it ran any queries.

hvitved (approved):
LGTM. I have started a single-threaded DCA run for just `apache__commons-compress`, so let's wait for that before merging.

hvitved:
> I have started a single-threaded DCA run for just `apache__commons-compress`, so let's wait for that before merging.

The run looks good, so go ahead and merge.

aschackmull:
There are a couple of (corner-)cases where this PR neglects to account for stateful in-/out-barriers. These are only checked on local value and taint steps and on jump steps. These barriers are not checked on read/store steps, nor on regular interprocedural steps. Unfortunately, fixing this won't be a clean addition, so I'm not sure whether it's worth fixing, or whether perhaps the use-case should rather be checked to see if the regular out-barrier would have been sufficient (which would allow us to revert stateful in-/out-barriers).
Another relatively minor omission is that this PR incorrectly handles the interplay between regularly defined sources/sinks (which implicitly use the default state), and partial barriers at those nodes using exactly the default state.

jketema:
@aschackmull Thanks. Is this easy for you to address, or should we have a look together?

aschackmull:
The easy way for me to address this is to remove `predicate isBarrierIn(Node node, FlowState state)` and `predicate isBarrierOut(Node node, FlowState state)`, so the question is whether you can do without those two predicates? Fully supporting those two predicates is going to be messy so that's my preferred solution.

jketema:
Unfortunately `predicate isBarrierOut(Node node, FlowState state)` was my main reason for adding this in the first place. See [here](https://github.com/jketema/codeql/blob/a84ee50af0a3226696f9f179eb56cee230444fb7/cpp/ql/src/Security/CWE/CWE-078/ExecTainted.ql#L120). If you think there's another solution that might work there, I'm up for it.

aschackmull:
Is it important for flow in the `ConcatState` to be able to continue after reaching a sink? Because otherwise you could simply use
```
  override predicate isSanitizerOut(DataFlow::Node node) {
    isSink(node, _) // Prevent duplicates along a call chain, since `shellCommand` will include wrappers
  }
```
instead.

jketema:
> Is it important for flow in the ConcatState to be able to continue after reaching a sink?

That might lead to false negatives, but I haven't looked if those occur in practise (I can do that if you'd like).

PR #8444, opened 2022-03-15
renaming more upper-case acronyms to PascalCase

erik-krogh:
The three first commits are auto-generated, and the remaining commits are various fixups. 

---- 

A bunch of the code is used internally, and thus some internal changes are also requires to avoid the CI running into deprecation errors. 
Thus this PR depends on an internal PR.  

---- 

`getURL()` is changed to `getUrl()`.  
That change requires changing the core and the documentation as `getURL()` is seen by the core as an indication that a value has a location.  

---- 

~~The change-notes from [the previous patch](https://github.com/github/codeql/pull/8323) also covers this patch (assuming I get this merged before the next release is cut).~~   
TODO: New change-note required.


smowton:
Java :+1:

MathiasVP (dismissed):

MathiasVP:
^ My approval was for C++. Not sure why that wasn't picked up by GitHub 🤔.

erik-krogh:
> ^ My approval was for C++. Not sure why that wasn't picked up by GitHub thinking.

Because it's still a draft PR. Your approvals don't count on draft PRs.

PR #8447, opened 2022-03-15
C++: Add missing `security-severity` tags

MathiasVP:
As identified by the new ql-for-ql query we added [here](https://github.com/github/codeql/pull/8437).


jketema (approved):
LGTM Assuming these were added with the scripting we have for this.

MathiasVP:
> LGTM Assuming these were added with the scripting we have for this.

Oh yeah, sorry. I should have said that. These were indeed added automatically by the scripting we have for this. Nothing controversial going on here!

PR #8449, opened 2022-03-15
C#: Re-use asPartialModel in utility and test code.

michaelnebel:
In this PR we
* Push the declaration of asPartialModel downwards and into DataFlowCallable to allow sharing between test code and utility code.
* Move RelevantSummarizedCallable out of the shared flow summary implementation (it was only used by C#). This simplifies the implementation as we can reduce the height of the inheritance hierarchy.
* Base the TargetApi for C# on DataFlowCallable (ie. only Callables with unbound declarations), which we should have done in the first place.

### DIFF
--- a/csharp/ql/lib/semmle/code/csharp/dataflow/internal/DataFlowDispatch.qll
+++ b/csharp/ql/lib/semmle/code/csharp/dataflow/internal/DataFlowDispatch.qll
@@ -249,6 +249,45 @@ class JumpReturnKind extends ReturnKind, TJumpReturnKind {
 
 class DataFlowCallable extends DotNet::Callable {
   DataFlowCallable() { this.isUnboundDeclaration() }
+
+  private string parameterQualifiedTypeNamesToString() {
+    result =
+      concat(Parameter p, int i |
+        p = this.getParameter(i)
+      |
+        p.getType().getQualifiedName(), "," order by i
+      )
+  }
+
+  /** Holds if a summary should apply for all overrides of this. */
+  predicate isBaseCallableOrPrototype() {
+    this.getDeclaringType() instanceof Interface
+    or
+    exists(Modifiable m | m = [this.(Modifiable), this.(Accessor).getDeclaration()] |
+      m.isAbstract()
+      or
+      this.getDeclaringType().(Modifiable).isAbstract() and m.(Virtualizable).isVirtual()
+    )
+  }
+
+  /** Gets a string representing whether a summary should apply for all overrides of this. */
+  private string getCallableOverride() {
+    if this.isBaseCallableOrPrototype() then result = "true" else result = "false"
+  }
+
+  /** Computes the first 6 columns for CSV rows. */
+  string asPartialModel() {
+    exists(string namespace, string type |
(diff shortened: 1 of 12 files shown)


csharp/ql/lib/semmle/code/csharp/dataflow/internal/DataFlowDispatch.qll
```diff
@@ -249,6 +249,45 @@ class JumpReturnKind extends ReturnKind, TJumpReturnKind {
...
+    )
+  }
+
+  /** Gets a string representing whether a summary should apply for all overrides of this. */
+  private string getCallableOverride() {
+    if this.isBaseCallableOrPrototype() then result = "true" else result = "false"
+  }
+
+  /** Computes the first 6 columns for CSV rows. */
+  string asPartialModel() {
```
hvitved, line 279: This predicate and `isBaseCallableOrPrototype` will now be public (since `DataFlowCallable` is), which I don't think we want. Instead it would be better to add them as top-level predicates in `DataFlowPrivate.qll`.

csharp/ql/lib/semmle/code/csharp/dataflow/internal/FlowSummaryImpl.qll
```diff
@@ -1022,46 +1022,6 @@ module Private {
     }
   }
 
-  /** Provides a query predicate for outputting a set of relevant flow summaries. */
```
hvitved, line 1025: I still think it makes sense to have here, in case other languages want to use it later.

java/ql/src/utils/model-generator/ModelGeneratorUtilsSpecific.qll
```diff
@@ -58,38 +58,38 @@ class TargetApi extends Callable {
     ) and
     isRelevantForModels(this)
   }
-}
 
-private string isExtensible(RefType ref) {
-  if ref.isFinal() then result = "false" else result = "true"
-}
+  private string isExtensible(RefType ref) {
```
hvitved, line 62: These predicates should not be members of `TargetApi`, as that means there is a Cartesian product with the implicit `this` parameter.

java/ql/src/utils/model-generator/ModelGeneratorUtilsSpecific.qll
```diff
@@ -58,38 +58,38 @@ class TargetApi extends Callable {
...
-      + api.getName() + ";" //
-      + paramsString(api) + ";" //
-      + /* ext + */ ";" //
+  /**
+   * Computes the first 6 columns for CSV rows.
+   */
+  string asPartialModel() {
+    result =
+      this.typeAsSummaryModel() + ";" //
+        + isExtensible(this.bestTypeForModel()) + ";" //
```
hvitved, line 88: As QL4QL says: `this.isExtensible`

hvitved (changes requested):

michaelnebel:
Thank you very much for the feedback! Considering the comments, I think it is easier to start over, instead of fixing the comments here as most of the changes needs to be reverted :-)

PR #8451, opened 2022-03-15
C#: Model generator improvements and more tests

michaelnebel:
In this PR we
* Improve the model generation for callables that uses collection like types for parameters (or just types that has collection like types as base class)
* Improve the model generation for callables that uses IEnumerable types as parameter types
* Introduce more test cases.

### DIFF
--- a/csharp/ql/lib/semmle/code/csharp/commons/Collections.qll
+++ b/csharp/ql/lib/semmle/code/csharp/commons/Collections.qll
@@ -66,6 +66,17 @@ class CollectionType extends RefType {
   }
 }
 
+/** An IEnumerable type. */
+class IEnumerableType extends RefType {
+  IEnumerableType() {
+    this.hasQualifiedName("System.Collections", "IEnumerable")
+    or
+    this.(ConstructedType)
+        .getUnboundGeneric()
+        .hasQualifiedName("System.Collections.Generic", "IEnumerable<>")
+  }
+}
+
 /** An object creation that creates an empty collection. */
 class EmptyCollectionCreation extends ObjectCreation {
   EmptyCollectionCreation() {
--- a/csharp/ql/lib/semmle/code/csharp/commons/Collections.qll
+++ b/csharp/ql/lib/semmle/code/csharp/commons/Collections.qll
@@ -1,6 +1,8 @@
 /** Provides classes for collections. */
 
 import csharp
+import semmle.code.csharp.frameworks.system.Collections
+import semmle.code.csharp.frameworks.system.collections.Generic
 
 private string modifyMethodName() {
   result =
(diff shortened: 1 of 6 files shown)


csharp/ql/lib/semmle/code/csharp/commons/Collections.qll
```diff
@@ -66,6 +66,17 @@ class CollectionType extends RefType {
   }
 }
 
+/** An IEnumerable type. */
```
hvitved, line 69: `` ` `` around `IEnumerable`.

csharp/ql/src/utils/model-generator/ModelGeneratorUtilsSpecific.qll
```diff
@@ -69,9 +70,14 @@ private predicate isPrimitiveTypeUsedForBulkData(Type t) {
   t.getName().regexpMatch("byte|char|Byte|Char")
 }
 
+private predicate isContainerType(Type t) {
```
hvitved, line 73: The old `LibraryTypeDataFlow.qll` used to have
```ql
private import semmle.code.csharp.frameworks.system.Collections

private predicate isCollectionType(ValueOrRefType t) {
  t.getABaseType*() instanceof SystemCollectionsIEnumerableInterface and
  not t instanceof StringType
}
```
which should cover all sorts of collection types (including arrays).

michaelnebel: Hah, that is interesting. I actually tried a solution similar to this one, but dropped it because a string would identify as a collection like type and I thought it would be a dit dirty to use an "exception" list (ie. t is a collection type if ..... unless t is a string, ...). I will re-introduce such a solution again :-)

csharp/ql/lib/semmle/code/csharp/commons/Collections.qll
```diff
@@ -66,15 +68,13 @@ class CollectionType extends RefType {
...
-    or
-    this.(ConstructedType)
-        .getUnboundGeneric()
-        .hasQualifiedName("System.Collections.Generic", "IEnumerable<>")
-  }
+/** Holds if `t` is a collection type. */
+predicate isCollectionType(ValueOrRefType t) {
+  not t instanceof StringType and
+  exists(ValueOrRefType base | base = t.getABaseType*() |
+    base instanceof SystemCollectionsGenericIEnumerableTInterface or
```
hvitved, line 75: This line should not be needed, since [`IEnumerable<T>`](https://docs.microsoft.com/en-us/dotnet/api/system.collections.generic.ienumerable-1?view=net-6.0) extends `IEnumerable`.

hvitved (changes requested):

csharp/ql/src/utils/model-generator/ModelGeneratorUtilsSpecific.qll
```diff
@@ -71,7 +72,7 @@ private predicate isPrimitiveTypeUsedForBulkData(Type t) {
 
 private string parameterAccess(Parameter p) {
   if
-    p.getType() instanceof ArrayType and
+    isCollectionType(p.getType()) and
     not isPrimitiveTypeUsedForBulkData(p.getType().(ArrayType).getElementType())
```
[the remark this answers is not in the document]
michaelnebel: Good question. I decided to copy the logic from Java and "say" that char[] and byte[] are now "bulk" like types but not any bound generic (with byte or char) as this appears to be the case for Java.
This is also a new approach compared to our existing summaries (which we have manually made) as we don't perceive char[] and byte[] as "bulk" types in our existing summaries.
The effect of something being bulk is that the "entire" datastructure is tainted (and not just ".Element").
We could also decide to completely eliminate the concept of bulk data for C# (@hvitved : What do you think?)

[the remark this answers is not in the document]
michaelnebel: Alright. We remove it.
Does the same thing hold for Java?

hvitved (approved):

PR #8454, opened 2022-03-15
QL: detect unqueryable code

erik-krogh:
Similar to [`ql/dead-code`](https://github.com/github/codeql/pull/8431), but does not consider public APIs as live.  

That means any code that doesn't affect any (known) query is flagged.   
The query is therefore good at detecting untested code or code that could be deprecated (:eyes: Python). 

An earlier iteration of this QL-for-QL query found some obvious bugs in JS: https://github.com/github/codeql/pull/8422  
The remaining results look interesting, but I haven't acted much on them.  

Way too much code got flagged in the `cpp` folder.  
I suspect that's from a bunch of code only being used outside `github/codeql`.  
So I decided to limit the query to the dynamic languages, the results seem reasonable for those.


cpp/ql/lib/semmle/code/cpp/ir/dataflow/internal/SsaInternals.qll
```diff
@@ -267,7 +267,9 @@ Instruction getSourceAddressFromNode(Node node) {
 }
 
 /** Gets the source value of `instr` if it's an instruction that behaves like a `LoadInstruction`. */
-Instruction getSourceValue(Instruction instr) { result = getSourceValueOperand(instr).getDef() }
+deprecated Instruction getSourceValue(Instruction instr) {
```
MathiasVP, line 270: This is in an internal directory, so this might as well just be deleted.

MathiasVP (commented):
I don't think we should deprecate code just because it doesn't have a use in our own set of queries. For instance, `localExprFlow` isn't something we're likely to use ourselves, but I've frequently seen QL code "in the wild" that uses it.

cpp/ql/lib/semmle/code/cpp/ir/dataflow/internal/DataFlowUtil.qll
```diff
@@ -1017,7 +1017,7 @@ predicate localInstructionFlow(Instruction e1, Instruction e2) {
  * local (intra-procedural) steps.
  */
 pragma[inline]
-predicate localExprFlow(Expr e1, Expr e2) { localFlow(exprNode(e1), exprNode(e2)) }
+deprecated predicate localExprFlow(Expr e1, Expr e2) { localFlow(exprNode(e1), exprNode(e2)) }
```
jbj, line 1020: This predicate exists for consistency with the data-flow libraries of other languages. Example: https://github.com/github/codeql/blob/b45f56ac08dd60cf53b2a873b74eb6ec44ef20e7/java/ql/lib/semmle/code/java/dataflow/internal/DataFlowUtil.qll#L80-L85

cpp/ql/lib/semmle/code/cpp/ir/dataflow/internal/DataFlowUtil.qll
```diff
@@ -731,15 +731,15 @@ DefinitionByReferenceNode definitionByReferenceNodeFromArgument(Expr argument) {
 }
 
 /** Gets the `VariableNode` corresponding to the variable `v`. */
-VariableNode variableNode(Variable v) { result.getVariable() = v }
+deprecated VariableNode variableNode(Variable v) { result.getVariable() = v }
```
jbj, line 734: This predicate, and the other unused `somethingNode` predicates, exist for consistency: if there's a public `SomethingNode` class, then there's a public `somethingNode` predicate.

jbj (commented):
I don't think we want the `DataFlowUtil` changes for C++.

erik-krogh:
I reverted the `DataFlowUtil.qll` deprecations, and deleted some stuff in various `/internal/` folder (where it doesn't seem to be exposed elsewhere).

ql/ql/src/codeql_ql/style/DeadCodeQuery.qll
```diff
@@ -208,8 +210,9 @@ private AstNode benign() {
   result instanceof Comment or
   not exists(result.toString()) or // <- invalid code
   // cached-stages pattern
-  result.(Module).getAMember().(ClasslessPredicate).getName() = "forceStage" or
-  result.(ClasslessPredicate).getName() = "forceStage" or
+  result.(Module).getAMember().(ClasslessPredicate).getName() =
+    ["forceStage", "forceCachingInSameStageforceCachingInSameStage"] or
```
tausbn, line 214: I feel like I'm seeing double. Should this be simply `forceCachingInSameStage`?

erik-krogh: And the case was also redundant.  
The below case in the disjunction already covered it.

PR #8457, opened 2022-03-16
Python: Add dataflow consistency query

RasmusWL:
As a real consistency query, so it will be run as part of ALL tests. (which might make CI take longer, but the value is nice I think)

I've made a dummy consistency query in https://github.com/github/codeql/pull/8458 to convince reviewers that these consistency queries are actually run :blush:


RasmusWL:
AHA, a few inconsistencies uncovered :detective: @yoff maybe we can work together on fixing these?

yoff:
Interesting. There are a few instances of `Node has multiple PostUpdateNodes.` The rest are missing `toString`s.

yoff:
It seems we have very few failure modes:
- `Call should have one enclosing callable but has 0.` (Lots)
- `Node steps to itself` (Lots)
- `Store step does not preserve enclosing callable.` (Just a few)

I wonder what is going on [here](https://github.com/github/semmle-code/actions/runs/5938080871/job/16101800809#step:27:7790)? We should not have semantic changes, should we? Is this to do with the missing file?

yoff:
> same problem as #1403

Ah, so updating should fix it.

RasmusWL:
woops, git merge did things to C++ which was certainly not intended

python/ql/consistency-queries/DataFlowConsistency.ql
```diff
@@ -0,0 +1,57 @@
...
+  predicate uniqueCallEnclosingCallableExclude(DataFlowCall call) {
+    not exists(call.getLocation().getFile().getRelativePath())
+  }
+
+  predicate identityLocalStepExclude(Node n) {
+    not exists(n.getLocation().getFile().getRelativePath())
+  }
+
+  predicate multipleArgumentCallExclude(ArgumentNode arg, DataFlowCall call) {
+    isArgumentNode(arg, call, _)
```
hvitved, line 53: Could this be strengthened, to make it clear where you expect multiple arguments?

python/ql/consistency-queries/DataFlowConsistency.ql
```diff
@@ -0,0 +1,57 @@
...
+private import codeql.dataflow.internal.DataFlowImplConsistency
+
+private module Input implements InputSig<PythonDataFlow> {
+  private import Private
+  private import Public
+
+  predicate argHasPostUpdateExclude(ArgumentNode n) {
+    exists(ArgumentPosition apos | n.argumentOf(_, apos) and apos.isStarArgs(_))
+    or
+    exists(ArgumentPosition apos | n.argumentOf(_, apos) and apos.isDictSplat())
```
hvitved, line 19: We have a similar exclusion in Ruby for implicit hash-splats, but this is only because we haven't yet implemented proper support. Is it the same for Python? I.e., in Ruby we do not currently handle
```rb
def foo(**args)
  args[:p].setField taint
end

foo(p: x)
sink(x.getField)
```

RasmusWL: Yes, the assertion in the following code holds. This was not really something I've thought too much about though, since it's not something I've ever seen in real code -- but I agree that it _could_ happen :+1:

```py
def set_foo(**args):
    args["p"].foo = 42

class MyClass: pass
c = MyClass()
set_foo(p=c)
assert c.foo == 42
```

python/ql/consistency-queries/DataFlowConsistency.ql
RasmusWL: Yes, done in https://github.com/github/codeql/pull/8457/commits/f9d7becd04fde16bbf7cbea6b4bde6c255e68247

RasmusWL: Argh, that made it too strict. For the one case I have debugged so far:

In the code `super(Base, self).foo()` we use `self` as an argument in both the super() call (pos 1) and in the .foo() call (pos self). Will look into fixing that tomorrow

python/ql/consistency-queries/DataFlowConsistency.ql
hvitved: I have documented the missing flow for Ruby here: https://github.com/github/codeql/pull/14859.

RasmusWL: Nice, done for Python here: https://github.com/github/codeql/pull/14936

python/ql/consistency-queries/DataFlowConsistency.ql
```diff
@@ -0,0 +1,122 @@
...
+    // `f = getattr(obj, "foo"); f()` where `obj` is used as (pos self) argument for
+    // `f()` call
+    exists(DataFlowCall getAttrCall, DataFlowCall methodCall, AttrRead attr |
+      call in [getAttrCall, methodCall]
+    |
+      arg = getAttrCall.getArgument(any(ArgumentPosition p | p.isPositional(0))) and
+      arg = methodCall.getArgument(any(ArgumentPosition p | p.isSelf())) and
+      attr.getObject() = arg and
+      attr.(CfgNode).getNode() = getAttrCall.getNode()
+    )
```
yoff, line 102: Could we use `GetAttrCallNode` here for `getAttrCall`? Then we would be a bit more sure that this is the method being called, and `arg = getAttrCall.getObject()` would also cover the `object=` case.

python/ql/consistency-queries/DataFlowConsistency.ql
```diff
@@ -0,0 +1,122 @@
...
+    or
+    // In the code `super(Base, self).foo()` we use `self` as an argument in both the
+    // super() call (pos 1) and in the .foo() call (pos self).
+    exists(DataFlowCall superCall, DataFlowCall methodCall | call in [superCall, methodCall] |
+      exists(superCallTwoArgumentTracker(_, arg)) and
```
yoff, line 107: Is this all our evidence that `superCall` is a call to `super`? Could we have something like `superCall.getNode() = superCallTwoArgumentTracker(_, arg)` instead or does that not eliminate all these inconsistencies?

python/ql/consistency-queries/DataFlowConsistency.ql
```diff
@@ -0,0 +1,122 @@
...
+      arg = methodCall.getArgument(any(ArgumentPosition p | p.isSelf()))
+    )
+    or
+    // in the code `def func(self): super().foo(); super.bar()` we use `self` as the
+    // (pos self) argument in both .foo() and .bar() calls.
+    exists(Function f |
+      exprNode(f.getArg(0)) = arg and
+      call.getNode().getScope() = f and
+      arg = call.getArgument(any(ArgumentPosition p | p.isSelf()))
+    )
```
yoff, line 118: I would think you want an `other` in this logic as well? Otherwise you may be hiding other problems?

python/ql/consistency-queries/DataFlowConsistency.ql
```diff
@@ -0,0 +1,122 @@
...
+private import semmle.python.dataflow.new.internal.DataFlowDispatch
+private import semmle.python.dataflow.new.internal.TaintTrackingImplSpecific
+private import codeql.dataflow.internal.DataFlowImplConsistency
+
+private module Input implements InputSig<PythonDataFlow> {
+  private import Private
+  private import Public
+
+  predicate argHasPostUpdateExclude(ArgumentNode n) {
+    exists(ArgumentPosition apos | n.argumentOf(_, apos) and apos.isStarArgs(_))
```
yoff, line 18: Should we have a comments saying "TODO: make this unnecessary"?

python/ql/consistency-queries/DataFlowConsistency.ql
```diff
@@ -0,0 +1,122 @@
...
+    or
+    exists(ArgumentPosition apos | n.argumentOf(_, apos) and apos.isDictSplat())
```
yoff, line 20: Should we have a comments saying "TODO: make this unnecessary"?

yoff (commented):
Thanks for the great comments spelling out the different situations. I have added some suggestions around aligning the code a bit more with the comments.

python/ql/consistency-queries/DataFlowConsistency.ql
RasmusWL: yes. if you look at the implementation of `superCallTwoArgumentTracker`, it's actually that `arg` is the "object" argument in the super() call.

python/ql/consistency-queries/DataFlowConsistency.ql
RasmusWL: done

python/ql/consistency-queries/DataFlowConsistency.ql
RasmusWL: done

python/ql/consistency-queries/DataFlowConsistency.ql
RasmusWL: good point, done in https://github.com/github/codeql/pull/8457/commits/02f20312395a757435ac9710183fbcfc613ebf07

python/ql/consistency-queries/DataFlowConsistency.ql
yoff: > I also don't see how this would improve things. attr.(CfgNode).getNode() = getAttrCall.getNode() makes sure that the AttrRead and the DataFlowCall shares the underlying CFG node... isn't that enough?

It probably is, but the argument is by casing on the implemented `AttrRead`s and their shape. it might be nicer to ensure directly that `getAttrCall` is a call to the Python function `getattr`.

I will not push hard for this, though, the tests will probably make noise if it becomes a problem..

python/ql/consistency-queries/DataFlowConsistency.ql
yoff: Indeed, and the result is something that the call flows to. That is why I suggested as I did. At the moment, we just know that `arg` is an argument to a call to `super` and that it is also `arg 1` to `superCall` (and `arg self` to `methodCall`). Presumably it could also be an argument to many other things (and the call to `super` could be different from `superCall`), that is what we are testing after all..

yoff (approved):
I still think that we could be a bit more robust, but I am not sure it is terribly important, and getting this in will be great, so I am approving now.

PR #8462, opened 2022-03-16
C#: Fix bad join order in returnNodeAsOutput.

michaelnebel:


hvitved (approved):
Looks plausible.

michaelnebel:
> Looks plausible.

Here are the tuple counts before and after the change, when executed against the same database.
Before:
image

After:
image

PR #8473, opened 2022-03-17
QL: expand redundant-inline-cast, and rename to redundant-cast

erik-krogh:
Inspired by: https://github.com/github/codeql/pull/8254#discussion_r827798028 

Finds nothing new. 
It flags something in a QL-for-QL test, but that shouldn't show up in the code-scanning results.


MathiasVP (approved):
LGTM! The diff provided by the GitHub UI isn't super helpful in this case (since it's a file modification + rename in a single commit), but luckily the change is simple enough to understand!

PR #8474, opened 2022-03-17
Dataflow: Flow-state changing steps should always be in path explanations

hvitved:
Alternative implementation to https://github.com/github/codeql/pull/8381.


MathiasVP:
Thanks for fixing this, Tom! This does indeed look much better 👍.

jketema:
Testing this out on the query I'm rewriting: `localFlowStepPlus` seems significantly slower now (compared to @MathiasVP's proposal).

jketema:
```
[2022-03-17 13:04:23] (135s) Starting to evaluate predicate DataFlowImpl::LocalFlowBigStep::localFlowStepPlus#fffffff/7@i7#e19c7wjr (iteration 7)
[2022-03-17 13:04:33] (144s) Tuple counts for DataFlowImpl::LocalFlowBigStep::localFlowStepPlus#fffffff/7@i7#e19c7wjr after 9.7s:
                      16646     ~5%      {7} r1 = SCAN DataFlowImpl::LocalFlowBigStep::localFlowStepPlus#fffffff#prev_delta OUTPUT In.1, In.5, In.0 'node1', In.2, In.3 'preservesValue', In.4 't', In.6 'cc'
                      313900036 ~0%      {8} r2 = JOIN r1 WITH project#DataFlowImpl::Stage2::revFlow#2#ffffff_120#join_rhs ON FIRST 2 OUTPUT Lhs.3, Rhs.2 'node2', Rhs.1 'config', Rhs.0, Lhs.2 'node1', Lhs.4 'preservesValue', Lhs.5 't', Lhs.6 'cc'
                      24073     ~14%     {7} r3 = JOIN r2 WITH DataFlowImpl::localFlowStepNodeCand1#fff ON FIRST 3 OUTPUT Lhs.1 'node2', Lhs.3 'state', Lhs.4 'node1', Lhs.5 'preservesValue', Lhs.6 't', Lhs.7 'cc', Lhs.2 'config'
                      
                      16646     ~19%     {7} r4 = SCAN DataFlowImpl::LocalFlowBigStep::localFlowStepPlus#fffffff#prev_delta OUTPUT In.2, In.1, In.1, In.5, false, In.0 'node1', In.6 'cc'
                      5202      ~4%      {7} r5 = JOIN r4 WITH DataFlowImpl::LocalFlowBigStep::additionalLocalFlowStepNodeCand2#fffff_01342#join_rhs ON FIRST 4 OUTPUT Rhs.4 'node2', false, Lhs.5 'node1', Lhs.1 'state', Lhs.6 'cc', Lhs.3 'config', Rhs.4 'node2'
                      5116      ~0%      {7} r6 = JOIN r5 WITH DataFlowImpl::NodeEx::getDataFlowType0#ff ON FIRST 1 OUTPUT Lhs.6 'node2', Lhs.3 'state', Lhs.2 'node1', false, Rhs.1 't', Lhs.4 'cc', Lhs.5 'config'
                      
                      29189     ~15%     {7} r7 = r3 UNION r6
                      19928     ~6%      {7} r8 = r7 AND NOT DataFlowImpl::LocalFlowBigStep::localFlowStepPlus#fffffff#prev(Lhs.2 'node1', Lhs.1 'state', Lhs.0 'node2', Lhs.3 'preservesValue', Lhs.4 't', Lhs.6 'config', Lhs.5 'cc')
                      19928     ~3%      {7} r9 = SCAN r8 OUTPUT In.2 'node1', In.1 'state', In.0 'node2', In.3 'preservesValue', In.4 't', In.6 'config', In.5 'cc'
                                         return r9
```

hvitved:
> Testing this out on the query I'm rewriting: `localFlowStepPlus` seems significantly slower now (compared to @MathiasVP's proposal).

Thanks; I'll fix.

hvitved:
@MathiasVP : Does the C++ DCA run look OK to you? There are some (unrelated?) result differences, and a bit of slow-down on two project, but I wonder if that is just noise.

jketema:
@hvitved Which query set did you use for the C++ DCA experiment?

MathiasVP:
> @hvitved Which query set did you use for the C++ DCA experiment?

Looking at https://github.com/github/codeql-dca-main/blob/data/PR-8474-0-cpp__2/outputs/workflow/main-inputs.json it looks like `lgtm-full.qls` was used (which is not a query suite I've ever used for DCA).

All of the result changes are on `abseil-cpp-windows`. The result changes on the queries involving IR can probably be explained by variables-without-types issue. However, the `cpp/constant-comparison`, `cpp/long-switch`, and `loop-variable-changed` query changes don't use IR and thus cannot be explained by that existing issue. But they don't use dataflow, so they must be unrelated to this PR.

hvitved:
> Looking at https://github.com/github/codeql-dca-main/blob/data/PR-8474-0-cpp__2/outputs/workflow/main-inputs.json it looks like `lgtm-full.qls` was used (which is not a query suite I've ever used for DCA).

Yes, that is what I used.

MathiasVP:
> @MathiasVP : Does the C++ DCA run look OK to you? There are some (unrelated?) result differences, and a bit of slow-down on two project, but I wonder if that is just noise.

I think the slow-downs look fine. There are a bunch of slowdowns in queries that aren't related to dataflow at all, so I expect that it's all noise. So if it's just C++ that's showing a slow-down, then I think we're good.

MathiasVP (approved):

MathiasVP:
Since Python has already approved the test changes in https://github.com/github/codeql/pull/8381 I'll just go ahead and merge.

cpp/ql/lib/semmle/code/cpp/dataflow/internal/DataFlowImpl.qll
```diff
@@ -2745,10 +2769,10 @@ private module Stage4 {
 
   bindingset[node, cc, config]
   private LocalCc getLocalCc(NodeEx node, Cc cc, Configuration config) {
-    localFlowEntry(node, _, config) and
```
aschackmull, line 2748: Post-merge drive-by comment: With this removed we no longer need the `config` column. PR: https://github.com/github/codeql/pull/8694
The removed conjunct used to be a needed piece of magic, but the code has changed a lot since it was introduced, so it makes sense that it has become superfluous.

hvitved: Thanks for fixing.

PR #8476, opened 2022-03-17
Python/JS/Ruby: Shared concepts scaffolding

RasmusWL:
This introduces the base scaffolding from https://github.com/github/codeql/pull/8307

As mentioned in that PR, the idea is to do separate PRs for each of the concepts we want to share initially, to align on the concept design and query implementation.


esbena (dismissed):

javascript/ql/lib/semmle/javascript/internal/ConceptsImports.qll
```diff
@@ -0,0 +1,6 @@
+/**
+ * This file contains imports required for the JavaScript version of `ConceptsShared.qll`.
+ * Since they are language-specific, they can't be placed directly in that file, as it is shared between languages.
+ */
+
+import semmle.javascript.dataflow.DataFlow
```
erik-krogh, line 6:
```suggestion
import semmle.javascript.dataflow.DataFlow::DataFlow as DataFlow
```
? 

See https://github.com/github/codeql/pull/8477#issuecomment-1070911561

RasmusWL: Would you perfer it this way, or are you just discussing different solutions? (not 100% clear to me)

javascript/ql/lib/semmle/javascript/internal/ConceptsShared.qll
```diff
@@ -0,0 +1,13 @@
+/**
+ * Provides Concepts which are shared across languages.
+ *
+ * Each language will have a language specific `Concepts.qll` file that can import the
```
hmac, line 4:
```suggestion
 * Each language has a language specific `Concepts.qll` file that can import the
```
I think this is clearer to the reader, who might wonder if it describes the current state or some future plan that hasn't yet been realised.

hmac (approved):
Thanks for splitting this out @RasmusWL! I have one minor comment but 👍 otherwise.

javascript/ql/lib/semmle/javascript/internal/ConceptsImports.qll
RasmusWL: Great, I mentioned this to @hmac, and we agreed to go with this suggestion, since it should JustWork™. If it turns out to cause problems, we can fix it when we need it.

javascript/ql/lib/semmle/javascript/internal/ConceptsShared.qll
RasmusWL: Done in [414764c](https://github.com/github/codeql/pull/8476/commits/414764cceeb77b7072da392b2e5474077b953061)

erik-krogh (approved):

javascript/ql/lib/semmle/javascript/internal/ConceptsShared.qll
esbena: (also, this direct style matches some rule in our doc style guide)

yoff (approved):
LGTM

hmac (approved):

PR #8477, opened 2022-03-17
JS: private import javascript in DataFlow

RasmusWL:
This prevents a circular dependency when we override shared concepts, which leads to ambiguous name resolution errors.

cherry-picked from from https://github.com/github/codeql/pull/8307

/cc @hmac who wrote the change originally

the https://github.com/github/codeql/pull/8307/commits/0d7d98d63d002118b52af503631d1367cfea8faf commit is not included, since the file it improved was deleted in https://github.com/github/codeql/commit/c2743177af8d869cd784ed025896559ffe1c8a33 (already merged into main)


RasmusWL:
fixed with eriks suggestion in https://github.com/github/codeql/pull/8476

PR #8478, opened 2022-03-17
JS: fix context sensitivity bug in store-load matching

asgerf:
Background:
`reachableFromStoreBase` forward-tracks the base object of a store whose RHS is potentially reachable. If the RHS of that store is only reachable with call steps, we do not permit the base object to be tracked out of a return edge, as it would lead to wasted work (i.e. we want to prune infeasible flows early).

However, it turns out we tried to used the same `call` bit for two different purposes:
1. Were call steps used to reach the store RHS? (needed for pruning)
2. Did we use a call step while tracking the base object? (needed for context sensitivity)

This PR adds a new `boolean` parameter to track (1), while the boolean in the `PathSummary` corresponds to (2).

[Evaluation](https://github.com/github/codeql-dca-main/tree/data/asgerf/store-load-flow__default__code-scanning/reports) looks ok. The new results are uninteresting and most likely FPs, although it's hard to triage. I've confirmed that the flow steps enabled by this PR are correct in isolation, but the final path contain some spurious flow caused by imprecise handling of array indices and captured variables.

### DIFF
--- a/javascript/ql/lib/semmle/javascript/dataflow/Configuration.qll
+++ b/javascript/ql/lib/semmle/javascript/dataflow/Configuration.qll
@@ -1365,27 +1365,31 @@ private predicate loadStep(
 
 /**
  * Holds if there is flow to `base.startProp`, and `base.startProp` flows to `nd.endProp` under `cfg/summary`.
+ *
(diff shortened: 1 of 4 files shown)


erik-krogh (approved):
Nice :+1: 

And thanks for the description you wrote as part of this PR. It helped.

PR #8481, opened 2022-03-18
JS: recognize string replacement chains as scheme checks in js/incomplete-url-scheme-check

erik-krogh:
Gets a TP/TN for CVE-2021-43861 

[Evaluation was uneventful](https://github.com/github/codeql-dca-main/tree/data/erik-krogh/pr-8481-235aa9c__default__CustomSuite/reports).


esbena:
LGTM, but lets add a test.

esbena (approved):

PR #8483, opened 2022-03-18
C++: Add additional command line injection tests

jketema:
One additional test involving branching and one additional test involving `strncat` in a called subroutine.

### DIFF
--- a/cpp/ql/test/query-tests/Security/CWE/CWE-078/semmle/ExecTainted/test.cpp
+++ b/cpp/ql/test/query-tests/Security/CWE/CWE-078/semmle/ExecTainted/test.cpp
@@ -168,6 +168,35 @@ void test15(FILE *f) {
   system(command); // GOOD: the user string was converted to an integer and back
 }
 
+void test16(FILE *f, bool use_flags) {
+  // BAD: the user string is injected directly into a command
+  char command[1000] = "mv ", flags[1000] = "-R", filename[1000];
+  fread(filename, 1, 1000, f);
+
+  if (use_flags) {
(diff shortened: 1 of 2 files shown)


cpp/ql/test/query-tests/Security/CWE/CWE-078/semmle/ExecTainted/test.cpp
```diff
@@ -168,6 +168,35 @@ void test15(FILE *f) {
...
+  // BAD: the user string is injected directly into a command
+  char command[1000] = "mv ", flags[1000] = "-R", filename[1000];
+  fread(filename, 1, 1000, f);
+
+  test17_inner(command, flags, filename);
+
+  execl("/bin/sh", "sh", "-c", command);
+}
 
 // TODO: test for call context sensitivity at concatenation site
```
rdmarsh2, line 201: I'd like to see this todo resolved (and ideally we'd have done it as part of the flow state PR) - I think all it needs is a test18 with a call into test17_inner where the filename argument isn't tainted.

rdmarsh2 (approved):
Looks good to me. I'd like to see the TODO resolved but that doesn't need to block a merge

cpp/ql/test/query-tests/Security/CWE/CWE-078/semmle/ExecTainted/test.cpp
jketema: I'll add that test to this PR.

rdmarsh2 (approved):

PR #8486, opened 2022-03-18
Python: switch to shared implementation of IncompleteHostnameRegExp.ql

aibaars:


python/ql/lib/semmle/python/RegexTreeView.qll
```diff
@@ -1009,3 +1028,24 @@ class RegExpBackRef extends RegExpTerm, TRegExpBackRef {
 
 /** Gets the parse tree resulting from parsing `re`, if such has been constructed. */
 RegExpTerm getParsedRegExp(StrConst re) { result.getRegex() = re and result.isRootTerm() }
+
+/**
+ * A node whose value may flow to a position where it is interpreted
```
aibaars, line 1033: @tausbn You asked for this class to me moved to another location. Could you tell me the preferred place?

tausbn: I would just create a file in `lib/semmle/python/dataflow/new` called `Regex.qll`, and import it appropriately. Would that be okay?

yoff (commented):
This looks good to me now, although there are a bunch of QL warnings.
Really nice to see this sharing!

aibaars:
@yoff @tausbn I addressed your comments and fixed the QL warnings.

yoff (approved):
LGTM

PR #8491, opened 2022-03-18
C++: Use flow states in `cpp/command-line-injection`

jketema:
Rework `cpp/command-line-injection` to use flow states. Only thing I'm not sure about is whether the location of a dataflow node uniquely determines the node, which I depend upon:
```codeql
class ExecState extends DataFlow::FlowState {
  DataFlow::Node fst;
  DataFlow::Node snd;

  ExecState() {
    this = "ExecState (" + fst.getLocation() + ", " + snd.getLocation() + ")" and
    interestingConcatenation(fst, snd)
  }
...
```

### DIFF
--- a/cpp/ql/src/Security/CWE/CWE-078/ExecTainted.ql
+++ b/cpp/ql/src/Security/CWE/CWE-078/ExecTainted.ql
@@ -77,7 +77,7 @@ class ExecState extends DataFlow::FlowState {
   ExecState() {
     this =
       "ExecState (" + fst.getLocation() + " | " + fst + ", " + snd.getLocation() + " | " + snd + ")" and
-      interestingConcatenation(fst, snd)
+    interestingConcatenation(fst, snd)
   }
 
   DataFlow::Node getFstNode() { result = fst }
--- a/cpp/ql/src/Security/CWE/CWE-078/ExecTainted.ql
+++ b/cpp/ql/src/Security/CWE/CWE-078/ExecTainted.ql
@@ -75,8 +75,9 @@ class ExecState extends DataFlow::FlowState {
   DataFlow::Node snd;
 
   ExecState() {
-    this = "ExecState (" + fst.getLocation() + ", " + snd.getLocation() + ")" and
-    interestingConcatenation(fst, snd)
+    this =
+      "ExecState (" + fst.getLocation() + " | " + fst + ", " + snd.getLocation() + " | " + snd + ")" and
+      interestingConcatenation(fst, snd)
   }
 
   DataFlow::Node getFstNode() { result = fst }
--- a/cpp/ql/src/Security/CWE/CWE-078/ExecTainted.ql
+++ b/cpp/ql/src/Security/CWE/CWE-078/ExecTainted.ql
@@ -19,9 +19,9 @@ import semmle.code.cpp.security.Security
 import semmle.code.cpp.valuenumbering.GlobalValueNumbering
 import semmle.code.cpp.ir.IR
 import semmle.code.cpp.ir.dataflow.TaintTracking
-import semmle.code.cpp.ir.dataflow.TaintTracking2
 import semmle.code.cpp.security.FlowSources
 import semmle.code.cpp.models.implementations.Strcat
+import DataFlow::PathGraph
 
 Expr sinkAsArgumentIndirection(DataFlow::Node sink) {
   result =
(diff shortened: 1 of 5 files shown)


jketema:
> It doesn't - there's a lot of cases with pointer arguments where we have input and output nodes for the pointed-to values that share a location with the argument.

Good to know. Then the question is, whether it's unique enough for what we're doing here, or whether there's something that better identifies a data flow node.

hvitved:
> > It doesn't - there's a lot of cases with pointer arguments where we have input and output nodes for the pointed-to values that share a location with the argument.
> 
> Good to know. Then the question is, whether it's unique enough for what we're doing here, or whether there's something that better identifies a data flow node.

Note that in this case it will only be a problem if you have two `snd` nodes with the same location that are _also_ in `interestingConcatenation(_, snd)`, which is probably unlikely in practice? If not, appending the `toString()` representation to the location may be enough to disambiguate.

cpp/ql/src/Security/CWE/CWE-078/ExecTainted.ql
```diff
@@ -66,154 +66,69 @@ predicate interestingConcatenation(DataFlow::Node fst, DataFlow::Node snd) {
...
+  conf.hasFlowPath(sourceNode, sinkNode) and
   taintCause = sourceNode.getNode().(FlowSource).getSourceType() and
-  conf1.hasFlowPath(sourceNode, concatSink) and
-  interestingConcatenation(concatSink.getNode(), concatSource.getNode()) and // this loses call context
-  conf2.hasFlowPath(concatSource, sinkNode) and
-  shellCommand(sinkAsArgumentIndirection(sinkNode.getNode()), callChain)
-select sinkAsArgumentIndirection(sinkNode.getNode()), TPathNode1(sourceNode).(MergedPathNode),
-  TPathNode2(sinkNode).(MergedPathNode),
+  shellCommand(sinkAsArgumentIndirection(sinkNode.getNode()), callChain) and
+  concatResult = sinkNode.getState().(ExecState).getSndNode()
```
hvitved, line 130: Since you only care about the `snd` node, there should be no need to record the `fst` node in the flow state as well.

jketema: Indeed. I'm probably unnecessarily conservative there.

jketema:
> If not, appending the toString() representation to the location may be enough to disambiguate.

Indeed, also because that + the location is what the user eventually will see.

MathiasVP:
> > whether the location of a dataflow node uniquely determines the node, which I depend upon
> 
> 
> 
> It doesn't - there's a lot of cases with pointer arguments where we have input and output nodes for the pointed-to values that share a location with the argument.

And more generally, post update nodes piggyback on their pre update nodes for locations as well.

jketema:
DCA Wireshark observations without any of the discussed changes around `toString`:

* `cpp/cleartext-storage-buffer` and `cpp/command-line-injection` get faster. The two DCA experiments show about the same numbers here, although the data from the second experiment if very messy.
* `cpp/tainted-format-string` get slower (goes from 12 to 20 seconds)

cpp/ql/src/Security/CWE/CWE-078/ExecTainted.ql
jketema: I retract my statement. Turns out there is a case where `fst` does matter. See the `test19` I added. When I omit `fst`, 3 paths instead of 2 paths are reported involving l. 220. The same happens when I move `interestingConcatenation` from `ExecState` to `isAdditionalTaintStep`.

jketema:
> Many nodes will definitely have the same toString because we try quite hard to reduce the number of unique strings for performance reasons, but when combined with locations it's probably okay.

Updated accordingly.

> We don't like having query logic depend on the implementation of the toString. We even have a [ql-for-ql query](https://github.com/github/codeql/blob/main/ql/ql/src/queries/style/ToStringInQueryLogic.ql) for this. So it's a bit of a shame that we have to do this.

Is this run on PRs? I didn't explicitly write `toString()` and don't see a warning.

MathiasVP:
> Is this run on PRs? I didn't explicitly write `toString()` and don't see a warning.

It's only at `@precision medium` - so it's not part of the Code Scanning suite.

cpp/ql/src/change-notes/2022-03-21-command-line-injection-with-flow-states.md
```diff
@@ -0,0 +1,4 @@
+---
+category: minorAnalysis
+---
+* The `cpp/command-line-injection` query now takes into account calling contexts across string concatenations. This removes false positives due mismatched calling contexts before and after string concatenations.
```
MathiasVP, line 4:
```suggestion
* The `cpp/command-line-injection` query now takes into account calling contexts across string concatenations. This removes false positives due to mismatched calling contexts before and after string concatenations.
```

MathiasVP (approved):
(I accidentally committed my own suggestion. Sorry about that!)

In any case, this LGTM!

cpp/ql/src/change-notes/2022-03-21-command-line-injection-with-flow-states.md
jketema: Change LGTM 😄

MathiasVP: Thanks! 😂

MathiasVP:
The Code Scanning errors are unrelated. So I'll just go ahead and merge this.

PR #8501, opened 2022-03-19
Java: Delayed unsafe deserialization

artem-smotrakov:
Deserialization can sometimes be implemented in two steps. An untrusted serialized object can be stored in a field but actual deserialization happens only when the object is necessary. CVE-2016-6194 in RabbitMQ is an example of such scenario ([GitHub issue](https://github.com/rabbitmq/rabbitmq-jms-client/issues/3)). Untrusted data that comes from a response is stored in `RMQObjectMessage.buf` field. Then, deserialization happens when `getObject()` method is called. Currently, `java/unsafe-deserialization` query doesn't catch this.

I'd like to propose the following updates to the query:
* Add a flow step to the query that propagates taint from a byte array to a field. I suppose this step can match quite a lot of expressions. I guess enabling this step by default for all configs may be too much, but I think it should be okay for `java/unsafe-deserialization` query because unsafe deserialization is a severe issue.
* Add flow steps for methods in `DataInput` and `ObjectInput` that read byte arrays.
* Add flow sources and steps for RabbitMQ.

With these updates, the query detects CVE-2016-6194. Let me know what you think.

### DIFF
--- a/java/ql/lib/semmle/code/java/frameworks/JavaIo.qll
+++ b/java/ql/lib/semmle/code/java/frameworks/JavaIo.qll
@@ -15,8 +15,8 @@ private class JavaIoSummaryCsv extends SummaryModelCsv {
         "java.io;CharArrayWriter;true;toCharArray;;;Argument[-1];ReturnValue;taint",
         "java.io;ObjectInput;true;read;;;Argument[-1];Argument[0];taint",
         "java.io;DataInput;true;readFully;;;Argument[-1];Argument[0];taint",
-        "java.io;DataInput;true;readLine;();;Argument[-1];Argument[0];taint",
-        "java.io;DataInput;true;readUTF;();;Argument[-1];Argument[0];taint",
+        "java.io;DataInput;true;readLine;();;Argument[-1];ReturnValue;taint",
+        "java.io;DataInput;true;readUTF;();;Argument[-1];ReturnValue;taint",
         "java.nio.channels;ReadableByteChannel;true;read;(ByteBuffer);;Argument[-1];Argument[0];taint",
         "java.nio.channels;Channels;false;newChannel;(InputStream);;Argument[0];ReturnValue;taint"
       ]
--- a/java/ql/lib/semmle/code/java/frameworks/JavaIo.qll
+++ b/java/ql/lib/semmle/code/java/frameworks/JavaIo.qll
@@ -13,6 +13,10 @@ private class JavaIoSummaryCsv extends SummaryModelCsv {
         "java.io;Writer;true;write;;;Argument[0];Argument[-1];taint",
         "java.io;Writer;true;toString;;;Argument[-1];ReturnValue;taint",
         "java.io;CharArrayWriter;true;toCharArray;;;Argument[-1];ReturnValue;taint",
+        "java.io;ObjectInput;true;read;;;Argument[-1];Argument[0];taint",
+        "java.io;DataInput;true;readFully;;;Argument[-1];Argument[0];taint",
+        "java.io;DataInput;true;readLine;();;Argument[-1];Argument[0];taint",
+        "java.io;DataInput;true;readUTF;();;Argument[-1];Argument[0];taint",
         "java.nio.channels;ReadableByteChannel;true;read;(ByteBuffer);;Argument[-1];Argument[0];taint",
         "java.nio.channels;Channels;false;newChannel;(InputStream);;Argument[0];ReturnValue;taint"
       ]
--- a/java/ql/lib/semmle/code/java/security/UnsafeDeserializationQuery.qll
+++ b/java/ql/lib/semmle/code/java/security/UnsafeDeserializationQuery.qll
@@ -256,6 +256,8 @@ class UnsafeDeserializationConfig extends TaintTracking::Configuration {
     createJacksonTreeNodeStep(pred, succ)
     or
     intentFlowsToParcel(pred, succ)
+    or
+    byteArrayFieldFlowStep(pred, succ)
   }
 
   override predicate isSanitizer(DataFlow::Node node) {
@@ -517,5 +519,16 @@ private predicate joddJsonParserConfiguredUnsafely(Expr parserExpr) {
   |
     config.hasFlow(getAnUnsafelyConfiguredParser(), parser) and
     not config.hasFlow(getASafelyConfiguredParser(), parser)
+  )
+}
+
+/**
+ * Holds if `fromNode` to `toNode` is a dataflow step that stores data to a byte array field.
+ */
+private predicate byteArrayFieldFlowStep(DataFlow::Node fromNode, DataFlow::Node toNode) {
+  exists(FieldAccess access, Field field | field = access.getField() |
+    field.getType().hasName("byte[]") and
+    access = fromNode.asExpr() and
+    field.getAnAccess() = toNode.asExpr()
   )
 }
--- a/java/ql/test/query-tests/security/CWE-502/DelayedDeserialization.java
+++ b/java/ql/test/query-tests/security/CWE-502/DelayedDeserialization.java
@@ -1,0 +1,30 @@
+import java.io.ByteArrayInputStream;
+import java.io.DataInput;
+import java.io.DataInputStream;
+import java.io.ObjectInputStream;
+import java.net.Socket;
+
+public class DelayedDeserialization {
+
+    public void test(Socket socket) throws Exception {
+        DataInputStream in = new DataInputStream(socket.getInputStream());
+        ObjectHolder holder = new ObjectHolder();
+        holder.init(in);
+        holder.getObject();
+    }
+}
+
+class ObjectHolder {
+
+    private final byte[] buffer = new byte[1024];
+
+    public void init(DataInput input) throws Exception {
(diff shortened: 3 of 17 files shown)


java/ql/lib/semmle/code/java/frameworks/JavaIo.qll
```diff
@@ -13,6 +13,10 @@ private class JavaIoSummaryCsv extends SummaryModelCsv {
         "java.io;Writer;true;write;;;Argument[0];Argument[-1];taint",
         "java.io;Writer;true;toString;;;Argument[-1];ReturnValue;taint",
         "java.io;CharArrayWriter;true;toCharArray;;;Argument[-1];ReturnValue;taint",
+        "java.io;ObjectInput;true;read;;;Argument[-1];Argument[0];taint",
+        "java.io;DataInput;true;readFully;;;Argument[-1];Argument[0];taint",
+        "java.io;DataInput;true;readLine;();;Argument[-1];Argument[0];taint",
+        "java.io;DataInput;true;readUTF;();;Argument[-1];Argument[0];taint",
```
Marcono1234, line 19: Are those really correct? `readLine()` and `readUTF()` don't have any argument, shouldn't this be:
```suggestion
        "java.io;DataInput;true;readLine;();;Argument[-1];ReturnValue;taint",
        "java.io;DataInput;true;readUTF;();;Argument[-1];ReturnValue;taint",
```

java/ql/lib/semmle/code/java/security/UnsafeDeserializationQuery.qll
```diff
@@ -519,3 +521,14 @@ private predicate joddJsonParserConfiguredUnsafely(Expr parserExpr) {
...
 }
+
+/**
+ * Holds if `fromNode` to `toNode` is a dataflow step that stores data to a byte array field.
+ */
+private predicate byteArrayFieldFlowStep(DataFlow::Node fromNode, DataFlow::Node toNode) {
+  exists(FieldAccess access, Field field | field = access.getField() |
+    field.getType().hasName("byte[]") and
+    access = fromNode.asExpr() and
+    field.getAnAccess() = toNode.asExpr()
```
Marcono1234, line 532: Probably not an issue, but should this cast the access to `FieldRead` to explicitly exclude `FieldWrite`?
```suggestion
    field.getAnAccess().(FieldRead) = toNode.asExpr()
```

java/ql/test/query-tests/security/CWE-502/DelayedDeserialization.java
```diff
@@ -0,0 +1,30 @@
...
+    public void init(DataInput input) throws Exception {
+        input.readFully(buffer, 0, buffer.length);
+    }
+
+    public Object getObject() throws Exception {
+        ByteArrayInputStream bais = new ByteArrayInputStream(buffer);
+        ObjectInputStream ois = new ObjectInputStream(bais);
+        return ois.readObject(); // $unsafeDeserialization
+    }
+}
```
Marcono1234, line 30: Maybe it would be good to also add a test where an assignment to a `byte[]` field transfers taint, e.g.:
(assuming that is the reason why you chose `FieldAccess` instead of the more specific `FieldRead` for your `byteArrayFieldFlowStep` predicate above)
```java
...

class ObjectHolder {
    private byte[] buffer;

    public void init(byte[] tainted) throws Exception {
        buffer = tainted;
    }

    public Object getObject() throws Exception {
        ByteArrayInputStream bais = new ByteArrayInputStream(buffer);
        ObjectInputStream ois = new ObjectInputStream(bais);
        return ois.readObject(); // $unsafeDeserialization
    }
}
```

Marcono1234 (commented):
Hopefully these comments are useful. Feel free to consider them only suggestions since I am not a member of this project.

smowton:
I suggest splitting this PR into two parts: the uncontroversial addition of new sources, sinks and library-specific propagation steps, but then let's consider separately the difficult issue of when it's reasonable to treat a field write as a jump step (one that discards calling context, and supposes that probably at some point a read step takes place)

pwntester:
@artem-smotrakov can you show us in which specific case you need the `byteArrayFieldFlowStep`? It seems like there may be some framework-specific lifecycle (the framework calls user-defined methods as part of an event loop or similar) which we could model ([example](https://github.com/github/codeql/blob/1e62b485a5ac5c0d8c278abe5226a918df70cf53/java/ql/lib/semmle/code/java/frameworks/android/Slice.qll#L18))

java/ql/lib/semmle/code/java/frameworks/JavaIo.qll
artem-smotrakov: Absolutely. Thanks!

java/ql/lib/semmle/code/java/security/UnsafeDeserializationQuery.qll
artem-smotrakov: Maybe. I'm going to move this to a separate PR, let's discuss it there. Thanks for the suggestion!

java/ql/test/query-tests/security/CWE-502/DelayedDeserialization.java
artem-smotrakov: Thanks for the idea!

artem-smotrakov:
Thanks for the review and suggestions @Marcono1234 !

artem-smotrakov:
> I suggest splitting this PR into two parts: the uncontroversial addition of new sources, sinks and library-specific propagation steps, but then let's consider separately the difficult issue of when it's reasonable to treat a field write as a jump step (one that discards calling context, and supposes that probably at some point a read step takes place)

Hi @smowton Sure, let's split it. I'm going to add a model for JMS API as well. If you're not sure about `byteArrayFieldFlowStep`, how about adding an experimental query for this kind of delayed deserialization?

artem-smotrakov:
hey @pwntester 

> can you show us in which specific case you need the byteArrayFieldFlowStep? It seems like there may be some framework-specific lifecycle (the framework calls user-defined methods as part of an event loop or similar) which we could model ([example](https://github.com/github/codeql/blob/1e62b485a5ac5c0d8c278abe5226a918df70cf53/java/ql/lib/semmle/code/java/frameworks/android/Slice.qll#L18))

I got this idea while analysing a [fix](https://github.com/rabbitmq/rabbitmq-jms-client/pull/4/files) for CVE-2016-6194 in RabbitMQ. Here is the details.

`RMQObjectMessage` stores serialized data in `buf` field

https://github.com/rabbitmq/rabbitmq-jms-client/blob/3b907c4baea8ef9eb9f3fd73dd209d12948689fd/src/main/java/com/rabbitmq/jms/client/message/RMQObjectMessage.java#L26

Untrusted data can be written to `buf` by calling `writeBody()` method

https://github.com/rabbitmq/rabbitmq-jms-client/blob/3b907c4baea8ef9eb9f3fd73dd209d12948689fd/src/main/java/com/rabbitmq/jms/client/message/RMQObjectMessage.java#L74

But actuall deserialization of untrusted data happens later when `getObject()` is called

https://github.com/rabbitmq/rabbitmq-jms-client/blob/3b907c4baea8ef9eb9f3fd73dd209d12948689fd/src/main/java/com/rabbitmq/jms/client/message/RMQObjectMessage.java#L58

Honestly, it doesn't look framework-specific to me. Untrusted data flows via byte-arrays and standard Java I/O objects. That is why I thought we can cover this case in the query for unsafe deserialization. In CVE-2016-6194, untrusted data comes from an RabbitMQ message, therfore I think it makes sense to add a source for that.

As I suggested above, we could probably create an experimental query for this kind of delayed deserialization instaead of updating the existing one. What do you think?

artem-smotrakov:
> I suggest splitting this PR into two parts: the uncontroversial addition of new sources, sinks and library-specific propagation steps

Hi @smowton I've moved this part to https://github.com/github/codeql/pull/8765 . Also, I've added a model for JMS as @pwntester suggested.

artem-smotrakov:
> let's consider separately the difficult issue of when it's reasonable to treat a field write as a jump step (one that discards calling context, and supposes that probably at some point a read step takes place)

@smowton I've moved this to a new experimental query so that this wouldn't affect the standard one.

https://github.com/github/codeql/pull/8766

> Hopefully these comments are useful. Feel free to consider them only suggestions since I am not a member of this project.

@Marcono1234 I've applied your suggestions in the new pull request. Please have a look. Your commenta are as usual very useful, I am always happy to see them!

artem-smotrakov:
Hi @pwntester 

> can you show us in which specific case you need the byteArrayFieldFlowStep? It seems like there may be some framework-specific lifecycle

> Honestly, it doesn't look framework-specific to me. Untrusted data flows via byte-arrays and standard Java I/O objects.

As an example, please have a look at the test from the new PR

https://github.com/github/codeql/pull/8766/files#diff-e799d2006a36e76240a753169ed054410e13dbd379365629a410d212bc91d0e2

artem-smotrakov:
This PR has been split to two

https://github.com/github/codeql/pull/8765 Adds new flow sources and steps for JMS API versions 1 and 2, RabbitMQ, and a few for Java I/O
https://github.com/github/codeql/pull/8766 Adds a new experimental query for delayed unsafe deserialization

Closing it.

PR #8507, opened 2022-03-21
C++: Make getUnderlyingType nomagic

geoffw0:
Motivated by some observations of `cpp/system-data-exposure` (by DCA), but the effects of this change will not be limited to this project, and may or may not be positive overall.  So I'll do a DCA run on this PR, then we can discuss...


geoffw0:
DCA:
 - none of the flagged items (build times, one alert change) are likely explained by this change.
   - but sooner or later we're going to get bitten by the unreliability of DCA and how easily we [have to] dismiss results.
 - overall analysis time was quicker but by an insignificant amount.

So DCA suggests this change is safe.  I believe it helps in certain cases.  And @MathiasVP has said he believes it is a good idea.

MathiasVP (approved):
LGTM!

PR #6450, opened 2021-08-09
JS: Migrate to new *Query.qll convention for files only to be imported by queries

asgerf:
Adopts the convention that files defining a data-flow configuration should end with `Query.qll`, and such files should only be imported from `.ql` files. The last part is already enforced by a PR check.

To be consistent with other languages, these library files no longer wrap their contents in a module. So the new way to import the file is:
```codeql
import semmle.javascript.security.dataflow.Foo::Foo // before
import semmle.javascript.security.dataflow.FooQuery // after
```

For backwards compatibility, the original file is still there, with a deprecated module alias pointing to the new file.

PR #8511, opened 2022-03-21
Python: Use `Query.qll` suffix for dataflow configuration definitions

RasmusWL:
Adopting the standard setup of defining the data flow configurations for our path-problem queries in a file called `...Query.qll`, so it's very obvious that this file should only be used by the query. (there is a CI check for this on our own codebase).

Like JS did in https://github.com/github/codeql/pull/6450

When adopting this, I wanted us to do the same as other languages in terms of setup. For simple queries that only require a single data-flow configuration, all langauges define that in the `...Query.qll` file. JS/Ruby always use the name `Configuration` for this configuration, which is also the most straightforward thing for us to do.

- JS [.ql file](https://github.com/github/codeql/blob/f6da0305721c32b86c8acd84d218fee1edf9c4c7/javascript/ql/src/Security/CWE-078/CommandInjection.ql#L18-L19), [Query.qll file](https://github.com/github/codeql/blob/45d1fa7f01af4187afe32969d5bdf443753f9961/javascript/ql/lib/semmle/javascript/security/dataflow/CommandInjectionQuery.qll#L17)
- Ruby [.ql file](https://github.com/github/codeql/blob/976daddd36a63bf46836d141d04172e90bb4b33c/ruby/ql/src/queries/security/cwe-078/CommandInjection.ql#L17), [Query.qll file](https://github.com/github/codeql/blob/976daddd36a63bf46836d141d04172e90bb4b33c/ruby/ql/lib/codeql/ruby/security/CommandInjectionQuery.qll#L19)
- C# [.ql file](https://github.com/github/codeql/blob/4de4753c6735732a283bc8d31c04e214b1fa54b4/csharp/ql/src/Security%20Features/CWE-078/CommandInjection.ql#L17), [Query.qll file](https://github.com/github/codeql/blob/c9f1c9839065d5b63bd6084d64f209ed417cf6eb/csharp/ql/lib/semmle/code/csharp/security/dataflow/CommandInjectionQuery.qll#L28)
- Java
    - for the command injection query that I checked for other languages, they use a setup with a private configuration, and a predicate that exposes the results:
    [.ql file](https://github.com/github/codeql/blob/d5a9f3d87b924cd967c1a417680025832517cbc3/java/ql/src/Security/CWE/CWE-078/ExecTainted.ql#L18), [Query.qll file](https://github.com/github/codeql/blob/8e750f18ada0064a47bdf8b9258cbb4461263e84/java/ql/lib/semmle/code/java/security/CommandLineQuery.qll#L14)
    - but for something like XSLT injection, they do like the rest, exposing the configuration from the Query.qll file, and using that directly in the [.ql file](https://github.com/github/codeql/blob/ad08ccb50b40c6eef2f8b99d06ad317bd071fd03/java/ql/src/Security/CWE/CWE-074/XsltInjection.ql#L14), [Query.qll file](https://github.com/github/codeql/blob/78c12dc505b6732498a7c4cca7d6af141ccd9773/java/ql/lib/semmle/code/java/security/XsltInjectionQuery.qll#L12)

However, it's not quite as obvious how we should handle cases where we need multiple configurations, currently `ServerSideRequestForgeryQuery.qll`, `WeakSensitiveDataHashingQuery.qll`, and `LdapInjectionQuery.qll`. I noted that `LdapInjectionQuery.qll` actually used a very different pattern than the two others. Instead of defining separate modules for each configuration, they just had different names within the same module :thinking:

Honestly, the approach in `LdapInjectionQuery.qll` seems simpler and requires less boiler-plate code, so I will suggest adopting that approach for `ServerSideRequestForgeryQuery.qll`. For `WeakSensitiveDataHashingQuery.qll` I really do think that having the two separate ql-modules makes sense, since they both use different sinks/sources, and could have different sanitizers (exposing them as top-level configurations means that we have to rewrite `source instanceof Source` to `source instanceof WeakSensitiveDataHashingCustomizations::NormalHashFunction::Source` :angry:)

According to `git grep "extends TaintTracking::Configuration" **/*Query.qll | cut -d ':' -f 1 | uniq -c | grep -v "1 "` the only file that has more than one `extends TaintTracking::Configuration` is [csharp/ql/lib/semmle/code/csharp/security/dataflow/UnsafeDeserializationQuery.qll](https://github.com/github/codeql/blob/c9f1c9839065d5b63bd6084d64f209ed417cf6eb/csharp/ql/lib/semmle/code/csharp/security/dataflow/UnsafeDeserializationQuery.qll#L0-L1) -- all 3 used directly in the [query](https://github.com/github/codeql/blob/d1e41689bbff8097d280d572204ce8bf470141e9/csharp/ql/src/Security%20Features/CWE-502/UnsafeDeserializationUntrustedInput.ql#L19). So not too much inspiration to match against.


yoff (approved):
LGTM
It took me a few iterations and jumping between the single-commit-view and the all-commits-view, but I think the commit structure actually made it quite easy to understand the transformation (I just had to check that nothing was missed) 👍

RasmusWL:
> It took me a few iterations and jumping between the single-commit-view and the all-commits-view, but I think the commit structure actually made it quite easy to understand the transformation (I just had to check that nothing was missed)

After doing this I wasn't quite sure if I had made the right choice, so that makes me happy :relaxed:

PR #8515, opened 2022-03-21
C++: generate IR for global variables with initializers

rdmarsh2:


cpp/ql/test/library-tests/ir/ir/PrintAST.expected
```diff
@@ -12842,9 +12842,9 @@ ir.cpp:
 # 1674|       getDeclarationEntry(0): (no string representation)
 # 1674|           Type = [ArrayType] int[2]
 # 1674|         getVariable().getInitializer(): [Initializer] initializer for (unnamed local variable)
-# 1674|           getExpr(): [VariableAccess] xs
+# 1674|           getExpr(): (no string representation)
 # 1674|               Type = [ArrayType] int[2]
-# 1674|               ValueCategory = prvalue(load)
+# 1674|               ValueCategory = prvalue
 # 1674|       getDeclarationEntry(1): [VariableDeclarationEntry] definition of x0
```
jketema, line 12848: These changes imply you used an outdated extractor.

cpp/ql/test/library-tests/ir/ir/raw_ir.ql
```diff
@@ -7,5 +7,7 @@ private import semmle.code.cpp.ir.implementation.raw.PrintIR
 private import PrintConfig
 
 private class PrintConfig extends PrintIRConfiguration {
-  override predicate shouldPrintFunction(Function func) { shouldDumpFunction(func) }
+  override predicate shouldPrintFunction(Declaration decl) {
+    shouldDumpFunction(decl) or decl instanceof GlobalOrNamespaceVariable
```
MathiasVP, line 11: Should we have a `shouldDumpDeclaration` predicate as well? I guess that predicate should exclude global variables with a location that satisfies `locationIsInStandardHeaders` (like we do for `shouldDumpFunction`)

MathiasVP (commented):
Looks great so far! I took a brief pass over it, and I only have a single comment right now.

cpp/ql/test/library-tests/ir/ir/aliased_ssa_consistency.expected
```diff
@@ -21,6 +25,14 @@ useNotDominatedByDefinition
 switchInstructionWithoutDefaultEdge
 notMarkedAsConflated
 wronglyMarkedAsConflated
+| ir.cpp:1038:12:1038:18 | Chi: [...](...){...} | Instruction 'Chi: [...](...){...}' should not be marked as having a conflated result in function '$@'. | ir.cpp:1038:6:1038:8 | (lambda [] type at line 1038, col. 12) lam | (lambda [] type at line 1038, col. 12) lam |
+| ir.cpp:1038:12:1038:18 | Chi: [...](...){...} | Instruction 'Chi: [...](...){...}' should not be marked as having a conflated result in function '$@'. | ir.cpp:1038:6:1038:8 | (lambda [] type at line 1038, col. 12) lam | (lambda [] type at line 1038, col. 12) lam |
+| ir.cpp:1695:16:1695:16 | Chi: 1 | Instruction 'Chi: 1' should not be marked as having a conflated result in function '$@'. | ir.cpp:1695:5:1695:12 | int global_2 | int global_2 |
```
dbartol, line 30: What's causing these failures? The ones in lambdas don't concern me that much, because we know we have issues there, but the ones in initializers should at least be investigated for this PR, even if the outcome is just a separate issue with a repro case.

cpp/ql/test/library-tests/ir/ir/raw_ir.expected
```diff
@@ -107,6 +107,8 @@ bad_asts.cpp:
 #   30|     v30_6(void)         = ExitFunction            : 
 
 clang.cpp:
+#    3| int globalInt
```
dbartol, line 110: I assume that this is a global variable with no initializer? Do we also dump functions that don't have IR as well, like virtual functions with no body? It just seems odd, although not necessarily wrong.

cpp/ql/test/library-tests/ir/ir/raw_ir.expected
```diff
@@ -8842,6 +8857,26 @@ ir.cpp:
 # 1689|     v1689_12(void)                             = AliasedUse                   : ~m?
 # 1689|     v1689_13(void)                             = ExitFunction                 : 
 
+# 1693| int global_1
+
+# 1695| int global_2
+# 1695|   Block 0
+# 1695|     v1695_1(void)       = EnterFunction   : 
+# 1695|     r1695_2(glval<int>) = VariableAddress :
```
dbartol, line 8865: This `VariableAddress` is missing the variable. Could this be the cause of some of those consistency failures? Did you remember to create an `IRVariable` for the global being initialized?

cpp/ql/lib/semmle/code/cpp/ir/implementation/raw/internal/IRConstruction.qll
```diff
@@ -34,6 +34,9 @@ module Raw {
 
   cached
   predicate functionHasIR(Function func) { exists(getTranslatedFunction(func)) }
+  
+  cached
+  predicate varHasIRFunc(GlobalOrNamespaceVariable var) { any() } // TODO: restrict?
```
dbartol, line 39: TODO?

Also, does this cover static data members?

cpp/ql/lib/semmle/code/cpp/ir/implementation/raw/internal/TranslatedElement.qll
```diff
@@ -67,7 +67,8 @@ private predicate ignoreExprAndDescendants(Expr expr) {
   exists(Initializer init, StaticStorageDurationVariable var |
     init = var.getInitializer() and
     not var.hasDynamicInitialization() and
-    expr = init.getExpr().getFullyConverted()
+    expr = init.getExpr().getFullyConverted() and
```
dbartol, line 70: Not important for this PR, but shouldn't we be creating a separate IR function for the initializer of a static local that really is statically initialized?

cpp/ql/lib/semmle/code/cpp/ir/implementation/raw/internal/TranslatedElement.qll
```diff
@@ -949,3 +952,11 @@ abstract class TranslatedElement extends TTranslatedElement {
    */
   final TranslatedElement getParent() { result.getAChild() = this }
 }
+
+abstract class TranslatedInstructionContainer extends TranslatedElement {
```
dbartol, line 956: This could use at least a one-line QLDoc comment

cpp/ql/lib/semmle/code/cpp/ir/implementation/raw/internal/TranslatedInitialization.qll
```diff
@@ -137,7 +137,10 @@ abstract class TranslatedInitialization extends TranslatedElement, TTranslatedIn
 
   final override string toString() { result = "init: " + expr.toString() }
 
-  final override Function getFunction() { result = expr.getEnclosingFunction() }
+  final override Declaration getFunction() {
+    result = expr.getEnclosingFunction() or
```
dbartol, line 141: Seems like we wind up with this exact same disjunction in many places. Maybe factor out a predicate like:
```ql
Declaration getEnclosingDeclaration(Expr e) {
  result = e.getEnclosingFunction() or result = e.getEnclosingVariable().(GlobalOrNamespaceVariable)
}
```
?

cpp/ql/test/library-tests/ir/ir/ir.cpp
```diff
@@ -1690,4 +1690,10 @@ void captured_lambda(int x, int &y, int &&z)
     };
 }
 
+int global_1;
+
+int global_2 = 1;
+
+const int global_3 = 2;
```
dbartol, line 1697: Could use test cases involving constructors.

dbartol, line 1697: We probably shouldn't create IR for anything as trivial as `static const int x = 5;`. Maybe only create it if the variable is every used as a glvalue? Or only if the variable is either non-`const` or has a non-constant initializer?

dbartol (changes requested):
Overall looks good. Several comments, including a couple blocking ones.

cpp/ql/test/library-tests/ir/ir/ir.cpp
jketema, line 1697: My personal hobby horse I guess: a test case for a structured binding would also be welcome (yes these can occur globally outside functions).

cpp/ql/test/library-tests/ir/ir/raw_ir.expected
rdmarsh2: I just added an `IRVariable`, which fixes the missing variable here but doesn't solve the consistency failures.

cpp/ql/test/library-tests/ir/ir/aliased_ssa_consistency.expected
rdmarsh2: Not entirely sure yet - there's Chi instructions being inserted for the global variable, and those are being marked as conflated. I'm not sure the Chis should be there in the first place,

cpp/ql/lib/semmle/code/cpp/ir/implementation/raw/internal/IRConstruction.qll
rdmarsh2: No, it doesn't. I think it could be expanded to `StaticStorageDurationVariable`, but that might make the `getEnclosingFunction` issues worse...

cpp/ql/test/library-tests/ir/ir/aliased_ssa_consistency.expected
rdmarsh2: The Chis are generated and marked as conflated because the global variable is defined to escape, and that's a consistency failure because the chi chain doesn't go back to an `AliasedDefinitionInstruction`.

cpp/ql/lib/semmle/code/cpp/ir/implementation/raw/internal/TranslatedElement.qll
rdmarsh2: Probably. It might be worth addressing now, actually.

cpp/ql/test/library-tests/ir/ir/aliased_ssa_consistency.expected
rdmarsh2: I think the easiest fix here is to consider the variable not to escape in its own initializer, but I'm a little nervous about having no aliased memory instruction in more complicated initializers. What do you think?

cpp/ql/test/library-tests/ir/ir/raw_ir.expected
rdmarsh2: Yes, it is. I think you're right that these can be excluded.

cpp/ql/test/library-tests/ir/ir/ir.cpp
rdmarsh2: This is what the `TODO` in `varHasIRFunc` was about. I was originally planning to include constants in the IR because that means they'll be available as dataflow sources once I add flow through global variables in a followup PR.

cpp/ql/test/library-tests/ir/ir/aliased_ssa_consistency.expected
rdmarsh2: On second thought adding the aliased memory isn't that hard, so I've gone ahead and done it

cpp/ql/lib/semmle/code/cpp/ir/implementation/raw/internal/TranslatedElement.qll
rdmarsh2: Done

cpp/ql/test/library-tests/ir/ir/ir.cpp
rdmarsh2: Added and fixed

dbartol (dismissed):
LGTM now, modulo merge conflicts

MathiasVP (approved):
LGTM!

PR #8517, opened 2022-03-22
Ruby: fix bug with captured variable reads in lambdas

hmac:
In the following example, we currently identify the read of `x` as a method call:
```ruby
x = 1
f = ->(y) { x }
```

This change fixes that, so we instead identify it as a captured variable read.

[Evaluation](https://github.com/github/codeql-dca-main/blob/data/hmac/pr-8517-99b5c58__nightly__lgtm-full/reports/summaries/any.md#analysis-time-per-source) shows no significant performance impact, and otherwise no changes that I can see.

### DIFF
--- a/ruby/ql/lib/codeql/ruby/ast/internal/Variable.qll
+++ b/ruby/ql/lib/codeql/ruby/ast/internal/Variable.qll
@@ -406,7 +406,11 @@ import Cached
 
 /** Holds if this scope inherits `name` from an outer scope `outer`. */
 private predicate inherits(Scope::Range scope, string name, Scope::Range outer) {
-  (scope instanceof Ruby::Block or scope instanceof Ruby::DoBlock) and
+  (
+    scope instanceof Ruby::Block or
+    scope instanceof Ruby::DoBlock or
+    scope instanceof Ruby::Lambda
+  ) and
   not scopeDefinesParameterVariable(scope, name, _) and
   (
     outer = scope.getOuterScope() and
--- a/ruby/ql/test/library-tests/ast/calls/calls.rb
+++ b/ruby/ql/test/library-tests/ast/calls/calls.rb
@@ -345,3 +345,14 @@ foo(x: 42)
 foo(x:, novar:)
 foo(X: 42)
 foo(X:)
+
+# calls inside lambdas
+y = 1
+id = ->(x) { y }
+f = ->(x) { foo x }
+g = ->(x) { unknown_call }
+h = -> (x) do
+  x
+  y
+  unknown_call
+end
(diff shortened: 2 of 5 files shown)


ruby/ql/lib/codeql/ruby/ast/internal/Variable.qll
```diff
@@ -406,7 +406,11 @@ import Cached
 
 /** Holds if this scope inherits `name` from an outer scope `outer`. */
 private predicate inherits(Scope::Range scope, string name, Scope::Range outer) {
-  (scope instanceof Ruby::Block or scope instanceof Ruby::DoBlock) and
+  (
+    scope instanceof Ruby::Block or
+    scope instanceof Ruby::DoBlock or
+    scope instanceof Ruby::Lambda
+  ) and
```
aibaars, line 413: `class DoBlock extends Block`, so we should be able to write:

```suggestion
  (scope instanceof Ruby::Block or scope instanceof Ruby::Lambda) and
```

aibaars (commented):
Looks good to me!

ruby/ql/lib/codeql/ruby/ast/internal/Variable.qll
hvitved: Those are classes from `TreeSitter.qll`, and not `AST.qll`, so I think we still need both cases.

ruby/ql/test/library-tests/ast/calls/calls.rb
```diff
@@ -344,4 +344,15 @@ def foo(a, b, ...)
 foo(x: 42)
 foo(x:, novar:)
 foo(X: 42)
-foo(X:)
\ No newline at end of file
+foo(X:)
+
+# calls inside lambdas
+y = 1
+id = ->(x) { y }
```
hvitved, line 351: nit: I would call it e.g. `one`, `id` makes me expect that it is the identify function.

hvitved (approved):
LGTM, thanks for fixing.

ruby/ql/test/library-tests/ast/calls/calls.rb
hmac: Oh true, haha. I think I initially wrote the identity function and then changed it to be relevant to this fix...

hmac: I'll fix this in a followup PR, so you don't have to re-approve this one.

hmac: https://github.com/github/codeql/pull/8543

PR #8522, opened 2022-03-22
ReDoS refactorization

erik-krogh:
Don't look at this PR, it's just for me right now.  

I'll do the implementation for JS first, and then get Python/Ruby along by running the sync script. 

This PR can only be merged after CodeQL 2.9.0 is released (and that will be a while).  

TODO: 
 - Disable QL-for-QL
 - Depends on https://github.com/github/codeql/pull/8521 
 - Run the sync script
 - Try the bad-tag-filter thing


DawidPlus44:
Yop

PR #8523, opened 2022-03-22
JS: Add dedicated API graph label for the receiver, instead of parameter -1

asgerf:
This changes the API graph label representing the receiver of call from `parameter -1` to `receiver`.

The member predicates of `API::Node` are affected as follows:
- `getParameter(i)` no longer has a result for `i = -1`, previously it returned the receiver.
- `getAParameter()` no longer includes the receiver.
- `getLastParameter()` no longer returns the receiver when the call has zero arguments. (🤦)

In theory this is not a behaviour-preserving change, but in practice it doesn't change much:
- There are 90 uses of `getParameter(i)` outside `ApiGraphs.qll`, all of them unaffected due to how they restrict `i`.
- There are 11 uses of `getAParameter`, some of which are affected, but as far as I can tell, none of them actually intended to get the receiver.
- There are 3 uses of `getLastParameter`, which are affected in theory but unlikely to matter in practice (since the modeled function expects >=1 arguments), and none of them intended to get the receiver.

This was motivated by an upcoming change in MaD, where we want to use `Argument[this]` to represent the receiver (`Argument[self]` in Python/Ruby).

[Evaluation](https://github.com/github/codeql-dca-main/tree/data/asgerf/api-graph-recei__default__code-scanning__2/reports) looks good. A few [spurious taint sources](https://github.com/github/codeql-dca-main/blob/data/asgerf/api-graph-recei__default__code-scanning__2/reports/alert-meta-comparison.md) have been removed, which referred to the receiver argument to a callback.

### DIFF
--- a/javascript/ql/lib/change-notes/2022-03-23-api-graph-receiver-label.md
+++ b/javascript/ql/lib/change-notes/2022-03-23-api-graph-receiver-label.md
@@ -1,0 +1,7 @@
+---
+category: fix
+---
+* The following predicates on `API::Node` have been changed so as not to include the receiver. The receiver should now only be accessed via `getReceiver()`.
+  - `getParameter(int i)` previously included the receiver when `i = -1`
+  - `getAParameter()` previously included the receiver
+  - `getLastParameter()` previously included the receiver for calls with no arguments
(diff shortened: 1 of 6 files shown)


asgerf:
@max-schaefer let me know if this is breaking anything

max-schaefer:
> @max-schaefer let me know if this is breaking anything

I don't think I have any code that relies on this behaviour, so no objection from me!

asgerf:
Rebased in an attempt to fix seemingly the unrelated PR check failure.

PR #8524, opened 2022-03-22
Ruby: Minor change of SSRF concept

RasmusWL:
When I added the Python query in https://github.com/github/codeql/pull/7420, I did some minor changes. This is only part of those changes, I will do another PR for the changes I made to `disablesCertificateValidation` once I have had time to change our query. (this PR had been laying around for some time, so will do it now so I wont forget about it).

I'm of the understanding that we can just make any breaking changes to the API as we want, since Ruby is still in beta, so that's why I haven't done any kind of deprecation :wink: 

Notice that the change-note [would bump the major version of the library pack](https://github.com/github/codeql/blob/main/docs/change-notes.md#library-pack-change-categories). If you don't want this, let me know, and I can change things :blush:

### DIFF
--- a/ruby/ql/lib/change-notes/2022-03-09-http-client-getaurlpart.md
+++ b/ruby/ql/lib/change-notes/2022-03-09-http-client-getaurlpart.md
@@ -1,4 +1,4 @@
 ---
 category: breaking
 ---
-* The `getURL` member-predicate of the `HTTP::Client::Request` and `HTTP::Client::Request::Range` classes from `Concepts.qll` have been renamed to `getAUrlPart`.
+* The `getURL` member-predicates of the `HTTP::Client::Request` and `HTTP::Client::Request::Range` classes from `Concepts.qll` have been renamed to `getAUrlPart`.
--- a/ruby/ql/lib/change-notes/2022-03-09-http-client-getaurlpart.md
+++ b/ruby/ql/lib/change-notes/2022-03-09-http-client-getaurlpart.md
@@ -1,0 +1,4 @@
+---
+category: breaking
+---
+* The `getURL` member-predicate of the `HTTP::Client::Request` and `HTTP::Client::Request::Range` classes from `Concepts.qll` have been renamed to `getAUrlPart`.
--- a/ruby/ql/lib/codeql/ruby/Concepts.qll
+++ b/ruby/ql/lib/codeql/ruby/Concepts.qll
@@ -485,6 +485,14 @@ module HTTP {
       DataFlow::Node getResponseBody() { result = super.getResponseBody() }
 
       /**
+       * DEPRECATED: Use `getAUrlPart` instead.
+       *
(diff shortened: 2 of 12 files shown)


nickrolfe:
We have mostly been attempting to follow the deprecation policy, since we already have query-writers working in other repos. This change looks like a simple rename, so I feel like it would be a good idea to keep the old name with the `deprecated` attribute.

RasmusWL:
I have fixed things up, so uses of the concept follows deprecation policy, but definition of models are required to change their override to the new name. If that's not good enough for "mostly" follow the deprecation policy, I can do further changes :blush:

ruby/ql/lib/change-notes/2022-03-09-http-client-getaurlpart.md
```diff
@@ -0,0 +1,4 @@
+---
+category: breaking
+---
+* The `getURL` member-predicate of the `HTTP::Client::Request` and `HTTP::Client::Request::Range` classes from `Concepts.qll` have been renamed to `getAUrlPart`.
```
nickrolfe, line 4: We need either one of the following, to make the subject and verb agree:
- "The `getURL` member-predicates … have been …"
- "The `getURL` member-predicate … has been …"

nickrolfe (changes requested):
Thanks! Looks good, apart from my grammar nitpick in the changenote.

ruby/ql/lib/change-notes/2022-03-09-http-client-getaurlpart.md
RasmusWL: :ok_hand: 

```suggestion
* The `getURL` member-predicates of the `HTTP::Client::Request` and `HTTP::Client::Request::Range` classes from `Concepts.qll` have been renamed to `getAUrlPart`.
```

nickrolfe (approved):
"Co-authored-by"? Very fancy! 👌

ruby/ql/lib/codeql/ruby/Concepts.qll
```diff
@@ -484,6 +484,14 @@ module HTTP {
       /** Gets a node which returns the body of the response */
       DataFlow::Node getResponseBody() { result = super.getResponseBody() }
 
+      /**
+       * DEPRECATED: Use `getAUrlPart` instead.
```
hmac, line 488: This isn't entirely related to this PR, but I was wondering what you thought about adding `(2022-03-24)` or similar to this comment so it's clear when the predicate was deprecated. I know we can estimate this using git blame, but that can become inaccurate if code gets moved around.

Even if it's a good idea, it might be something better added to all the deprecated predicates as a separate PR. I'm mostly just interested in thoughts.

RasmusWL: I think it's an excellent idea :+1:

Issue #8497, opened 2022-03-19
[CPP][Questions]No effective API to qeury macro used in function parameter declaration


4B5F5F4B:
Hello all,

Recently I'm interested in writing queries to detect common vulnerability pattern specific for Linux kernel codebase. I find that __user marco is used to indicate parameter is user mode pointer, for example

#define __user          //empty macro body

static int sg_scsi_ioctl(struct request_queue *q, fmode_t mode,
		                       struct scsi_ioctl_command __user *sic)    //sic is a user mode pointer

I want to write a query to get all user mode pointer defined in function parameter declaration, I tried to find avaliable API defined in Marco/MacroAccess/MarcoInovation/Function/Function/FunctionDeclarationEntry/ParameterDeclarationEntry, but I can not find one can be used for my purpose.

I have no way but to query macro used in function parameter declaration by combining MarcroAccess and Location, the following code may be work, but it prone to be false positive and ugly:(

class UserParameterDeclarationEntry extends ParameterDeclarationEntry{
    UserParameterDeclarationEntry(){
        exists(MacroAccess m| 
            m.getMacroName() = "__user"
            and m.getFile() = this.getFile()
            and m.getLocation().getEndLine() = this.getLocation().getEndLine()
        )
    }
}

So I hope you guys can be kind enougth to help me find a more effective way to query specific macro in function parameter declaration, thank you:)

MathiasVP:
Hi @4B5F5F4B,

This is indeed an annoying issue. Finding the macro uses is easy (using `MacroInvocation` and the related classes), but associating the macro uses with a parameter is very difficult since macros don't fit nicely into the AST (as it's just text being expanded). So there's no good way to do it other than hacking it via locations. This is how I'd do it, I think:
```ql
import cpp

pragma[inline]
predicate isBefore(Location l1, Location l2) {
  l1.getFile() = l2.getFile() and
  (
    l1.getEndLine() < l2.getStartLine()
    or
    l1.getEndLine() = l2.getStartLine() and
    l1.getEndColumn() < l2.getStartColumn()
  )
}

predicate isUserParameter(Parameter p) {
  exists(MacroInvocation m |
    p =
      min(Parameter q, TypeMention tm |
        q.getType().stripType() = tm.getMentionedType() and
        m.getMacroName() = "__user" and
        isBefore(tm.getLocation(), m.getLocation()) and
        isBefore(m.getLocation(), q.getLocation())
      |
        q order by q.getIndex()
      )
  )
}

[...]

PR #8525, opened 2022-03-22
C++: Consider columns in `Location.isBefore`

MathiasVP:
I was hacking on a solution to https://github.com/github/codeql/issues/8497 and was surprised to learn that `Location.isBefore` is really `Location.lineIsBefore`.

As an added bonus to this PR, this will improve our IR generation for code that uses `goto`s:

Currently, our _only_ use of `Location.isBefore` predicate is to determine whether a `goto` statement creates a back edge in the control-flow graph. So on `main`, the following program:

```cpp
int test() {
  int x = 42;
  goto next; next:

  return x;
}
```
gets translated as:
```cpp
int test()
  Block 0
    v1_1(void)       = EnterFunction      :
    m1_2(unknown)    = AliasedDefinition  :
    m1_3(unknown)    = InitializeNonLocal :
    m1_4(unknown)    = Chi                : total:m1_2, partial:m1_3
    r2_1(glval<int>) = VariableAddress[x] :
    r2_2(int)        = Constant[42]       :
    m2_3(int)        = Store[x]           : &:r2_1, r2_2
    v3_1(void)       = NoOp               :
  Goto (back edge) -> Block 1

  Block 1
    v3_2(void)       = NoOp                     :
    r6_1(glval<int>) = VariableAddress[#return] :
    r6_2(glval<int>) = VariableAddress[x]       :
    r6_3(int)        = Load[x]                  : &:r6_2, m2_3
    m6_4(int)        = Store[#return]           : &:r6_1, r6_3
    r1_5(glval<int>) = VariableAddress[#return] :
    v1_6(void)       = ReturnValue              : &:r1_5, m6_4
    v1_7(void)       = AliasedUse               : m1_3
    v1_8(void)       = ExitFunction             :
```
i.e., we conservatively assume that the goto could be a back edge. However, with this PR the same code is translated as:
```cpp
int test()
  Block 0
    v1_1(void)       = EnterFunction            :
    m1_2(unknown)    = AliasedDefinition        :
    m1_3(unknown)    = InitializeNonLocal       :
    m1_4(unknown)    = Chi                      : total:m1_2, partial:m1_3
    r2_1(glval<int>) = VariableAddress[x]       :
    r2_2(int)        = Constant[42]             :
    m2_3(int)        = Store[x]                 : &:r2_1, r2_2
    v3_1(void)       = NoOp                     :
    v3_2(void)       = NoOp                     :
    r6_1(glval<int>) = VariableAddress[#return] :
    r6_2(glval<int>) = VariableAddress[x]       :
    r6_3(int)        = Load[x]                  : &:r6_2, m2_3
    m6_4(int)        = Store[#return]           : &:r6_1, r6_3
    r1_5(glval<int>) = VariableAddress[#return] :
    v1_6(void)       = ReturnValue              : &:r1_5, m6_4
    v1_7(void)       = AliasedUse               : m1_3
    v1_8(void)       = ExitFunction             :
```
with no back edge!

I admit this is a bit of a niche case. But it's nice to see that it actually does have the intended effect.

### DIFF
--- a/cpp/ql/lib/semmle/code/cpp/Location.qll
+++ b/cpp/ql/lib/semmle/code/cpp/Location.qll
@@ -73,8 +73,24 @@ class Location extends @location {
 
   /** Holds if `this` comes on a line strictly before `l`. */
   pragma[inline]
-  predicate isBefore(Location l) {
-    this.getFile() = l.getFile() and this.getEndLine() < l.getStartLine()
+  predicate isBefore(Location l) { isBefore(l, false) }
+
+  /**
+   * Holds if `this` comes strictly before `l`. The boolean `sameLine` is
+   * true if `l` is on the same line as `this`, but starts at a later column.
+   * Otherwise, `sameLine` is false.
+   */
+  pragma[inline]
+  predicate isBefore(Location l, boolean sameLine) {
+    this.getFile() = l.getFile() and
+    (
+      sameLine = false and
+      this.getEndLine() < l.getStartLine()
+      or
+      sameLine = true and
+      this.getEndLine() = l.getStartLine() and
+      this.getEndColumn() < l.getStartColumn()
+    )
   }
 
   /** Holds if location `l` is completely contained within this one. */
--- a/cpp/ql/lib/semmle/code/cpp/Location.qll
+++ b/cpp/ql/lib/semmle/code/cpp/Location.qll
@@ -73,7 +73,7 @@ class Location extends @location {
 
   /** Holds if `this` comes on a line strictly before `l`. */
   pragma[inline]
-  predicate isBefore(Location l) { isBefore(l, false) }
+  predicate isBefore(Location l) { this.isBefore(l, false) }
 
   /**
    * Holds if `this` comes strictly before `l`. The boolean `sameLine` is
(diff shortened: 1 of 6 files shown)


MathiasVP:
@geoffw0 convinced me that we shouldn't change the behavior of the existing predicate. So I've force-pushed a backward-compatible version.

I've changed IR generation so that it uses the new predicate to get the behavior I described in the PR description.

jketema:
You might want to add the goto example to the IR tests?

MathiasVP:
> You might want to add the goto example to the IR tests?

Good point! I'll do that.

MathiasVP:
Done. I did another force-push to insert the test in the right commit.

cpp/ql/lib/semmle/code/cpp/Location.qll
```diff
@@ -73,8 +73,24 @@ class Location extends @location {
 
   /** Holds if `this` comes on a line strictly before `l`. */
   pragma[inline]
-  predicate isBefore(Location l) {
-    this.getFile() = l.getFile() and this.getEndLine() < l.getStartLine()
+  predicate isBefore(Location l) { isBefore(l, false) }
```
jketema, line 76: It shows a warning about an implicit `this` here. Should that be fixed?

MathiasVP: Yeah, it really should. I'll add the explicit `this`, but I won't force-push the change since that will mess up the DCA run.

jketema (approved):
LGTM.

PR #8530, opened 2022-03-22
InlineExpectationsTest: Allow URLs as values

RasmusWL:
For example,

```py
requests.get("http://example.com") # $  clientRequestUrlPart="http://example.com"
```

would not have been allowed previously, since the `//` part of the value for the URL would have caused the rest of the line to be treated as a normal (non-interpreted) comment.

Requiring a space before the `//` allows this specific example, while still support all the existing uses of `//` to break expectation tests, which is usually to add a comment explaining why the annotation is like it is.


RasmusWL:
@dbartol just wanted to check whether you would be OK with this change? (before pinging all the other language teams)

owen-mc:
Note that `codeql-go` also uses inline expectations. Would it be possible to update it there as well? I spent a while bringing it up to date about 6 months ago, so it should be pretty easy.

RasmusWL:
> Note that `codeql-go` also uses inline expectations. Would it be possible to update it there as well? I spent a while bringing it up to date about 6 months ago, so it should be pretty easy.

Sure thing, I'll do if the overall idea is accepted :+1:

PR #8489, opened 2022-03-18
Ruby: refactor regex libraries

aibaars:
This pull request refactors the regular expression libraries for Ruby to match the Python and JavaScript libraries.

I aligned the contents of the regex parser module (`ruby/ql/lib/codeql/ruby/regexp/internal/ParseRegExp.qll`) and the regex AST module (`ruby/ql/lib/codeql/ruby/regexp/internal/RegExpTreeView.qll`) as much as possible with the respective Python implementations `python/ql/lib/semmle/python/regex.qll` and `python/ql/lib/semmle/python/RegexTreeView.qll`

The additional classes and predicates are moved into `ruby/ql/lib/codeql/ruby/Regexp.qll` .This module should be used by most users and  imports the parser module and re-exports the AST module.

PR #8532, opened 2022-03-22
JS/Ruby/Python: rename RegExpTreeView.qll to ReDoSUtilSpecific.qll

aibaars:
In https://github.com/github/codeql/pull/8489#pullrequestreview-917455000 @nickrolfe  pointed out that 
> having two different RegExpTreeView.qll files seems undesirable. Can they be merged?

There are two files named `RegExpTreeView.qll` in the Python and Ruby libraries. One of them contains language specific implementations of classes/predicates needed the the shared `ReDoSUtil.qll` module. As a result they cannot be merged, however, renaming that module to `ReDoSUtilSpecific.qll` avoids the name clash and also makes it clear that the module is just "implementation details".

### DIFF
--- a/javascript/ql/lib/semmle/javascript/security/performance/ReDoSUtil.qll
+++ b/javascript/ql/lib/semmle/javascript/security/performance/ReDoSUtil.qll
@@ -12,7 +12,7 @@
  * states that will cause backtracking (a rejecting suffix exists).
  */
 
-import RegExpTreeView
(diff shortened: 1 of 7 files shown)


yoff (approved):
LGTM once the tests pass

erik-krogh (approved):
JS :+1:

nickrolfe (approved):
Ruby 👍

PR #8533, opened 2022-03-23
JS/Ruby: Represent non-positional arguments with Argument/Parameter tokens

asgerf:
Previously we had separate tokens for non-positional arguments, like `Receiver` and `BlockArgument`.

Now these are represented as operands to `Argument`/`Parameter` tokens, as discussed with @hvitved:

Ruby:
- `Argument[self]` for the receiver (previously `Receiver`)
- `Argument[block]` for the block argument (previously `BlockArgument`)
- `Argument[foo:]` for a keyword argument named `foo` (previously not implemented)

JS:
- `Argument[this]` for the receiver (previously `Argument[-1]`)

All of the above also apply to the `Parameter` token.

The names `self` and `this` were chosen to coincide with the keyword/convention from the language.

### DIFF
--- a/ruby/ql/lib/codeql/ruby/ApiGraphs.qll
+++ b/ruby/ql/lib/codeql/ruby/ApiGraphs.qll
@@ -473,36 +473,6 @@ module API {
     /** Gets a data flow node that flows to the RHS of a def-node. */
     private DataFlow::LocalSourceNode defCand() { result = defCand(TypeBackTracker::end()) }
 
-    private Label::ApiLabel getLabelFromArgumentPosition(DataFlowDispatch::ArgumentPosition pos) {
-      exists(int n |
-        pos.isPositional(n) and
-        result = Label::parameter(n)
-      )
-      or
-      exists(string name |
-        pos.isKeyword(name) and
-        result = Label::keywordParameter(name)
-      )
-      or
-      pos.isBlock() and
-      result = Label::blockParameter()
-    }
-
-    private Label::ApiLabel getLabelFromParameterPosition(DataFlowDispatch::ParameterPosition pos) {
-      exists(int n |
-        pos.isPositional(n) and
-        result = Label::parameter(n)
-      )
-      or
-      exists(string name |
-        pos.isKeyword(name) and
-        result = Label::keywordParameter(name)
-      )
-      or
-      pos.isBlock() and
-      result = Label::blockParameter()
-    }
-
     /**
      * Holds if there should be a `lbl`-edge from the given call to an argument.
      */
@@ -512,7 +482,7 @@ module API {
     ) {
       exists(DataFlowDispatch::ArgumentPosition argPos |
         argument.sourceArgumentOf(call.asExpr(), argPos) and
-        lbl = getLabelFromArgumentPosition(argPos)
+        lbl = Label::getLabelFromArgumentPosition(argPos)
       )
     }
 
@@ -525,7 +495,7 @@ module API {
     ) {
       exists(DataFlowDispatch::ParameterPosition paramPos |
         paramNode.isSourceParameterOf(callable.asExpr().getExpr(), paramPos) and
-        lbl = getLabelFromParameterPosition(paramPos)
+        lbl = Label::getLabelFromParameterPosition(paramPos)
       )
     }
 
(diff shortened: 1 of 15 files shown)


hmac (approved):
I don't feel qualified to comment closely on the implementation, but it seems straightforward and the test changes are what I'd expect. I think I have a slight preference for `Receiver` over `Argument[self]` (shorter and matches better with `MethodCall.getReceiver`) but having consistency between the languages probably trumps that.

The keyword parameter support is really nice!

erik-krogh (dismissed):
JS :+1:  
(The JS part is simple).

ruby/ql/lib/codeql/ruby/ApiGraphs.qll
```diff
@@ -803,5 +773,37 @@ module API {
...
+        result = Label::parameter(n)
+      )
+      or
+      exists(string name |
+        pos.isKeyword(name) and
+        result = Label::keywordParameter(name)
+      )
+      or
+      pos.isBlock() and
+      result = Label::blockParameter()
```
hvitved, line 790: Should the `toString` of `LabelBlockParameter` be updated to `getParameter(block)` (and similar for `LabelKeywordParameter`? And are we missing labels for the `self` parameter?

ruby/ql/test/library-tests/dataflow/summaries/summaries.rb
```diff
@@ -47,3 +47,20 @@ def userDefinedFunction(x, y)
...
 end
+
+sink(Foo.namedArg(foo: tainted))
+sink(Foo.namedArg(tainted))
+
+Foo.intoNamedCallback(tainted, foo: ->(x) {
+  sink(x)
+})
+Foo.intoNamedParameter(tainted, ->(foo:) {
+  sink(foo)
```
hvitved, line 58: We should document that flow is missing in this case

hvitved (commented):
LGTM; two minor comments.

ruby/ql/lib/codeql/ruby/ApiGraphs.qll
asgerf: The `toString` value of these labels are currently named after predicates on `API::Node` (not MaD tokens). At least that seemed to be the convention, one that I've tried to continue. So it doesn't need updating.

ruby/ql/test/library-tests/dataflow/summaries/summaries.rb
asgerf: Dang, I missed that. I'll try to introduce inline assertions to this test.

ruby/ql/lib/codeql/ruby/ApiGraphs.qll
erik-krogh: I like having the `toString()` be the associated predicate on `API::Node`, so maybe we should update JS to also do that?

ruby/ql/test/library-tests/dataflow/summaries/summaries.rb
asgerf: Updated to use inline test expectations, much better overall, and it's now clear that this flow is missing.

I haven't investigated why the flow is missing, though. I was under the impression that providing the correct `ArgumentPosition` and `ParameterPosition` values would be enough.

erik-krogh (dismissed):

ruby/ql/test/library-tests/dataflow/summaries/summaries.rb
asgerf: @hvitved would you like it addressed in this PR? Please let me know either way

hvitved: No, let's do that follow-up

ruby/ql/test/library-tests/dataflow/summaries/Summaries.ql
```diff
@@ -117,23 +117,18 @@ private class SinkFromModel extends ModelInput::SinkModelCsv {
...
-    )
+    super.isSink(sink)
     or
     sink = ModelOutput::getASinkNode("test-sink").getARhs()
   }
 }
-
-from DataFlow::PathNode source, DataFlow::PathNode sink, Conf conf
-where conf.hasFlowPath(source, sink)
-select sink, source, sink, "$@", source, source.toString()
```
hvitved, line 134: I prefer to actually keep this, in addition to `import PathGraph`, like we do in other flow tests that use `InlineFlowTest`.

hvitved (approved):

erik-krogh (approved):

PR #8534, opened 2022-03-23
C++: Handle bitwise copies in copy constructors

jketema:
Mostly test updates and removal of an IR generation exclusion that is no longer needed.


MathiasVP (dismissed):
LGTM!

jketema:
Does this need a change note? This is probably not actually visible to users unless they start digging into the IR.

jketema:
And I missed that you added the not change note label 😄

MathiasVP:
> And I missed that you added the not change note label 😄

Yeah, sorry. I was going to do a quick merge so I added the label. Only afterward did I see that the internal PR needed a bump. I don't think we need to add a change note. If we want to, we can add a change note once all the bitwise copy stuff has been fixed. I don't think we need to add one that says "we've fixed it in 1 out of 6 places".

MathiasVP (approved):

PR #8535, opened 2022-03-23
Ruby: fix location of setter-call argument

aibaars:
```ruby
x.foo = y
```
is desugared to

```ruby
x.foo=(__synth_0 = y);
__synth_0;
``` 
The expression `__synth_0 = y` and its subexpression `__synth_0` used to  inherit the location of the `x.foo` call. It would be better for those expressions to inherit the location of `y`.


hvitved (approved):
LGTM

PR #8540, opened 2022-03-23
Python: Add call graph meta-query

tausbn:


python/ql/src/meta/analysis-quality/CallGraph.ql
```diff
@@ -0,0 +1,16 @@
+/**
+ * @name Call graph
+ * @description An edge in the points-to call graph.
+ * @kind problem
+ * @problem.severity recommendation
+ * @id py/meta/points-to-call-graph
```
RasmusWL, line 6: I don't see any points-to specific implementation details, so I think we can remove that part from the query-id?

```suggestion
 * @id py/meta/call-graph
```

RasmusWL (changes requested):
The old meta queries in the `analysis-quality/` dir has logic to ignore some files. I would think this is something we want to incorporate here as well? (unless I'm overlooking something)


```codeql
class RelevantCall extends Call {
  RelevantCall() { not this.getLocation().getFile() instanceof IgnoredFile }
}
```

python/ql/src/meta/analysis-quality/CallGraph.ql
RasmusWL: oh, and if we do this, we should probably do it in the description as well

tausbn: `DataFlowCallableValue` is a `DataFlowCallable` explicitly backed by a `CallableValue`. This is why the query is points-to specific. 

I wanted to keep the query-id separate, so as to avoid potential confusion if we ever want to compare results between different meta-queries in the future.

I considered using the `RelevantCall` machinery, but for now I just wanted to get an impression of the complete picture. We can always add it later if needed.

RasmusWL: I misread `DataFlowCallableValue`, so never mind me then.

RasmusWL (approved):

PR #8542, opened 2022-03-23
C++: Use `iterated (post)dominance frontier` algorithm in `IRBlock`

MathiasVP:
I'm not aware of the old implementation causing any performance issues on any databases, but it's an easy win that we might as well do.


jketema (commented):
LGTM, but I'd like to see a non-failing DCA run.

I think this is what https://dl.acm.org/doi/10.1145/75277.75280 describes, right?

MathiasVP:
> LGTM, but I'd like to see a non-failing DCA run.

Good point. I've started a new one now.

> I think this is what https://dl.acm.org/doi/10.1145/75277.75280 describes, right?

That looks like the right source, yes. We use the algorithm in a bunch of places in the codebase. It's also found [here](https://github.com/github/codeql/blob/bbb2847ec1a94608044865ee9de81c094bb6b716/cpp/ql/lib/semmle/code/cpp/controlflow/SSAUtils.qll#L18) and [here](https://github.com/github/codeql/blob/main/csharp/ql/lib/semmle/code/cil/internal/SsaImplCommon.qll#L154).

jketema (approved):
LGTM. I didn't see anything of concern in the last DCA experiment.

PR #8550, opened 2022-03-24
JS: improve the join order of getAClassReference

erik-krogh:
Saves ~10s on VSCode.  

I didn't think such a `noopt` should do a difference, but it did. 

I'm doing an evaluation, I will only merge if that turns out OK.


CryptoGodfatherVA2 (commented):
VAR5

javascript/ql/lib/semmle/javascript/dataflow/Nodes.qll
```diff
@@ -960,7 +962,16 @@ class ClassNode extends DataFlow::SourceNode instanceof ClassNode::Range {
       result.getAstNode().getFile() = this.getAstNode().getFile()
     )
     or
-    exists(DataFlow::TypeTracker t2 | result = this.getAClassReference(t2).track(t2, t))
+    result = getAClassReferenceStep(t)
+  }
+
+  pragma[noopt]
+  private DataFlow::SourceNode getAClassReferenceStep(DataFlow::TypeTracker t) {
```
asgerf, line 969:
```suggestion
  private DataFlow::SourceNode getAClassReferenceRec(DataFlow::TypeTracker t) {
```
We have other cases of a type-tracking "step" predicates being factored out to fix a join ordering, but where some information about the edge itself is matched. This here is just the whole recursive case being factored, so I'd prefer a different name. I was certainly confused for a second, before realising that the `noopt` is reason for factoring it out.

asgerf (commented):
LGTM but let's wait for the evaluation

erik-krogh:
I made a mistake in the inlining, which caused a test to fail, so this PR might have been for nothing.  
I'll try to run an evaluation anyway. If that turns out neutral (which I expect), then I'll close this PR.

javascript/ql/lib/semmle/javascript/dataflow/Nodes.qll
```diff
@@ -970,7 +970,7 @@ class ClassNode extends DataFlow::SourceNode instanceof ClassNode::Range {
     exists(DataFlow::TypeTracker t2, StepSummary summary, DataFlow::SourceNode prev |
       prev = this.getAClassReference(t2) and
       StepSummary::step(prev, result, summary) and
-      t2 = t.append(summary)
+      t = t2.append(summary)
```
esbena, line 973: Smells like a ql4ql query...

erik-krogh:
I ran [an evaluation](https://github.com/github/codeql-dca-main/tree/data/erik-krogh/classJoin__nightly-old__CustomSuite/reports).  
And it shows improved performance :confused:  

No improvement on VSCode (which was the original intent), but other projects see a real performance improvement.  

I think we should merge this, and I should look into improving the join-order of our type-trackers.

asgerf (approved):

PR #8556, opened 2022-03-25
QL: update codeql-action version in QL-for-QL

erik-krogh:
Updating `codeql-action` to [this branch](https://github.com/github/codeql-action/tree/erik-krogh/newQl).  

~~TODO: Wait for [this `codeql-action` PR](https://github.com/github/codeql-action/pull/998) to merge, then cleanup the commit history (both of this PR and the `codeql-action` branch).~~

### DIFF
--- a/.github/workflows/ql-for-ql-build.yml
+++ b/.github/workflows/ql-for-ql-build.yml
@@ -16,7 +16,7 @@ jobs:
       - uses: actions/checkout@v2
       - name: Find codeql
         id: find-codeql
-        uses: github/codeql-action/init@erik-krogh/ql
+        uses: github/codeql-action/init@erik-krogh/newQl
         with:
           languages: javascript # does not matter
       - name: Get CodeQL version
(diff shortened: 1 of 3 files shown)


.github/workflows/ql-for-ql-build.yml
```diff
@@ -16,7 +16,7 @@ jobs:
       - uses: actions/checkout@v2
       - name: Find codeql
         id: find-codeql
-        uses: github/codeql-action/init@erik-krogh/ql
+        uses: github/codeql-action/init@erik-krogh/newQl
```
esbena, line 19: Is anything preventing us from doing a (force) push to erik-krogh/ql instead of creating erik-krogh/newQl? Then we can keep the `ql` branch name here...

erik-krogh: It's to ensure that the branch bump and the workflow change happen at the same time.

erik-krogh: I could force push on the old brach, but that could break existing PRs.
(I've already broken all the PRs once).

esbena: If that is a general concern, then we should change `@erik-krogh/newQl` to the tip:

```suggestion
        uses: github/codeql-action/init@aa93aea877e5fb8841bcb1193f672abf6e9f2980
```

Otherwise, we will have to make `erik-krogh/newQl-2` next time.

esbena: https://github.com/github/codeql-action/tree/erik-krogh/newQl

esbena:
We should pin that commit everywhere we mention `newQl`

esbena (approved):

PR #8560, opened 2022-03-25
PY: move the polynomialbacktracking-test to the test folder

erik-krogh:
It seems to have been misplaced in the `src` folder.


yoff (approved):
Thanks for cleaning this up.
Possibly a temporary state while developing that was never fixed...

PR #8561, opened 2022-03-25
QL: use latest tools in codeql-action/init

erik-krogh:


adityasharad (approved):
LGTM. We can go back to using the cached version once that is recent enough to address the problems you ran into.

PR #8573, opened 2022-03-27
Ruby: Add MissingRegExpAnchor query

hmac:
This is a port of the JS query of the same name. A few adjustments are needed to adapt to our dataflow and regex libraries, so it's not so easy to share the implementation. I've also added a Ruby-specific check for regexes that use `^` and `$`, which match the start/end of a line rather than the whole string. These regexes should probably use `\A` and `\z` instead.


alexrford (approved):
LGTM, I like the addition of checks for `^`/`$` rather than `\A`/`\z`.

There's an unrelated build failure in the DCA run that _may_ be fixed by a rebase.

hmac:
> There's an unrelated build failure in the DCA run that _may_ be fixed by a rebase.

It finally succeeded! It feels like I've been attempting DCA runs on this PR since the dawn of time...

The vast majority of results are in tests, but I've checked a few of the non-test ones and they look legitimate.

PR #8576, opened 2022-03-28
JS: Add decorator edges in API graphs and corresponding MaD tokens

asgerf:
Adds the following `API::Node` members:
- `getADecoratedClass`
- `getADecoratedMember`
- `getADecoratedParameter`

And corresponding MaD tokens:
- `DecoratedClass`
- `DecoratedMember`
- `DecoratedParameter`

For example:
```js
import { D } from "foo";

@D
class C // Matched by: foo;;Member[D].DecoratedClass
```

`DecoratedClass` could alternatively be expressed in terms of `Argument[0]` and `ReturnValue`. The others are not so simple, however, as they involve passing a base object and member name as an argument to the decorator, which typically then performs a reflective lookup. We don't have API edges for reflective lookups as they're tricky to model as unary operators.

So for the time being, no argument/return edges are generated for any of the decorator patterns although it's possible we should do so for class decorators in the future.

Awaiting evaluation when DCA comes up again.

### DIFF
--- a/javascript/ql/lib/semmle/javascript/ApiGraphs.qll
+++ b/javascript/ql/lib/semmle/javascript/ApiGraphs.qll
@@ -847,8 +847,8 @@ module API {
      * Since fields do not have their own data-flow nodes, we generate a node for each read or write.
      * For property writes, the right-hand side becomes a def-node and property reads become use-nodes.
      *
-     * For accessors this predicate computes each use of the accessor. 
-     * The return value inside the accessor is computed by the `decoratorRhsEdge` predicate. 
+     * For accessors this predicate computes each use of the accessor.
+     * The return value inside the accessor is computed by the `decoratorRhsEdge` predicate.
      */
     private predicate decoratorPropEdge(TApiNode base, Label::ApiLabel lbl, DataFlow::PropRef ref) {
       exists(MemberDefinition fieldLike, DataFlow::ClassNode cls |
--- a/javascript/ql/lib/semmle/javascript/ApiGraphs.qll
+++ b/javascript/ql/lib/semmle/javascript/ApiGraphs.qll
@@ -813,12 +813,6 @@ module API {
         lbl = Label::decoratedMember() and
         ref = DataFlow::valueNode(method.getBody())
       )
-      or
-      exists(Parameter param |
-        useNodeFlowsToDecorator(base, param.getADecorator()) and
-        lbl = Label::decoratedParameter() and
-        ref = DataFlow::parameterNode(param)
-      )
     }
 
     /** Holds if `ref` is a use that should have an incoming edge from `base` labelled `lbl`, induced by a decorator. */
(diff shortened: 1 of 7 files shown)


javascript/ql/lib/semmle/javascript/ApiGraphs.qll
```diff
@@ -699,6 +774,98 @@ module API {
...
+        useNodeFlowsToDecorator(base, method.getADecorator()) and
+        not method instanceof AccessorMethodDefinition and
+        lbl = Label::decoratedMember() and
+        ref = DataFlow::valueNode(method.getBody())
+      )
+      or
+      exists(Parameter param |
+        useNodeFlowsToDecorator(base, param.getADecorator()) and
+        lbl = Label::decoratedParameter() and
+        ref = DataFlow::parameterNode(param)
```
erik-krogh, line 820: I'm unsure if there should be a def-node for this parameter-node.   
And the tests pass just fine if I move the `decoratedParameter` edge to the `decoratorUseEdge` predicate.

javascript/ql/lib/semmle/javascript/ApiGraphs.qll
```diff
@@ -699,6 +774,98 @@ module API {
...
+    }
+
+    /**
+     * Holds if `ref` is a reference to a field/accessor that should have en incoming edge from base labelled `lbl`.
+     *
+     * Since fields do not have their own data-flow nodes, we generate a node for each read or write.
+     * For property writes, the right-hand side becomes a def-node and property reads become use-nodes.
+     *
+     * For accessors, we generate nodes both internally in the accessor, and for each use of the accessor. This
+     * predicate only gets the uses.
```
erik-krogh, line 851:
```suggestion
     * For accessors this predicate computes each use of the accessor. 
     * The return value inside the accessor is computed by the `decoratorRhsEdge` predicate. 
```

I found the comment hard to read, I've tried to rephrase it.

asgerf: Agreed this much better 👍

javascript/ql/lib/semmle/javascript/ApiGraphs.qll
asgerf: It's technically possible for the parameter decorator to replace the whole method with one that treats that parameter as a sink.

I have never seen any decorators that would have such behaviour, though. A made-up example might be a decorator that wraps some expensive runtime checks on a value, which could then be a DoS sink.

In any case, it turns out using a parameter node as a sink is actually a terrible idea, at least with the current implementation (see test case in [ca145f2](https://github.com/github/codeql/pull/8576/commits/ca145f21b04534312047c78eecbf19bae82c332f)).

So I've changed it to only generate use-nodes 👍

erik-krogh (approved):
:+1:

erik-krogh (approved):

asgerf:
Thanks for the reviews! I've started an evaluation

asgerf:
[Evaluation](https://github.com/github/codeql-dca-main/tree/data/asgerf/decorated-metho__default__code-scanning/reports) seems ok

PR #8578, opened 2022-03-28
JS: update `toString()` on API-graph labels. 

erik-krogh:
They now correspond to the predicate calls on the API-graph.  
(Similarly to how Python/Ruby does it).  

I like that approach better, and it makes our implementations more consistent across the languages.  

The tests were updated to use the new toString(), and I updated the tests to be more similar to the tests in Python (that themselves were made to look similar to inline expectation tests).  

Drive-by fix of the QLDoc for the same testing library in Python.  

No evaluation planned.


asgerf (commented):
JS changes LGTM

tausbn (dismissed):
So happy to see alignment on this! :heart: 

Python changes LGTM.

asgerf (dismissed):

asgerf (approved):

erik-krogh:
Did a merge to resolve a merge-conflict.  

In the merge I also updated the `toString()` of the new labels (the decorator labels).

tausbn (approved):

PR #8580, opened 2022-03-28
C++: Port PrivateData.qll from C# and use it in cpp/cleartext-transmission

geoffw0:
Port `PrivateData.qll` from C#.
 - the C# version was chosen as a base, rather than the version in experimental as had been originally planned.  They appear to be extremely similar, but the C# version is more complete.
 - the experimental version has been deprecated.
 - additional enhancements / tuning has been done to maximize results with minimal FPs.

Use `PrivateData.qll` in `cpp/cleartext-transmission`.
 - on LGTM we find a small number of new results, generally I think they're good, a few are weak or FP results.
 - `cpp/cleartext-storage-buffer`, `cpp/cleartext-storage-file` and `cpp/cleartext-storage-database` might also benefit, but I need to further investigate whether results are good for these queries.

### DIFF
--- a/cpp/ql/lib/semmle/code/cpp/security/PrivateData.qll
+++ b/cpp/ql/lib/semmle/code/cpp/security/PrivateData.qll
@@ -1,0 +1,66 @@
+/**
+ * Provides classes and predicates for identifying private data and methods for security.
+ *
+ * 'Private' data in general is anything that would compromise user privacy if exposed. This
+ * library tries to guess where private data may either be stored in a variable or produced by a
+ * method.
+ *
+ * This library is not concerned with credentials. See `SensitiveActions` for expressions related
+ * to credentials.
+ */
+
+import csharp
+import semmle.code.csharp.frameworks.system.windows.Forms
+
+/** A string for `match` that identifies strings that look like they represent private data. */
+private string privateNames() {
+  result =
+    [
+      // Inspired by the list on https://cwe.mitre.org/data/definitions/359.html
+      // Government identifiers, such as Social Security Numbers
+      "%social%security%number%",
+      // Contact information, such as home addresses and telephone numbers
+      "%postcode%", "%zipcode%", "%telephone%",
+      // Geographic location - where the user is (or was)
+      "%latitude%", "%longitude%",
+      // Financial data - such as credit card numbers, salary, bank accounts, and debts
+      "%creditcard%", "%salary%", "%bankaccount%",
+      // Communications - e-mail addresses, private e-mail messages, SMS text messages, chat logs, etc.
+      "%email%", "%mobile%", "%employer%",
+      // Health - medical conditions, insurance status, prescription records
+      "%medical%"
+    ]
+}
+
+/** An expression that might contain private data. */
+abstract class PrivateDataExpr extends Expr { }
+
+/** A method call that might produce private data. */
+class PrivateMethodCall extends PrivateDataExpr, MethodCall {
+  PrivateMethodCall() {
+    exists(string s | this.getTarget().getName().toLowerCase() = s | s.matches(privateNames()))
+  }
+}
+
+/** An indexer access that might produce private data. */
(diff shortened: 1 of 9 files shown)


cpp/ql/lib/semmle/code/cpp/security/PrivateData.qll
```diff
@@ -0,0 +1,63 @@
...
+
+/** A string for `match` that identifies strings that look like they represent private data. */
+private string privateNames() {
+  result =
+    [
+      // Inspired by the list on https://cwe.mitre.org/data/definitions/359.html
+      // Government identifiers, such as Social Security Numbers
+      "%social%security%",
+      // Contact information, such as home addresses and telephone numbers
+      "%postcode%", "%zipcode%", "%telephone%",
```
MathiasVP, line 23: Since we already allow arbitrary strings between the word `bank` and `account`, how about we also allow arbitrary strings between `post` and `code`, and `zip` and `code`?

Our tests would improve if we added `%licese%key%` as well (or maybe just `%license%`?). Do we catch more good results if we add those?

(These issues aren't at all blocking. Feel free to postpone them to another PR if you agree that this is something we should do)

geoffw0: > Since we already allow arbitrary strings between the word bank and account, how about we also allow arbitrary strings between post and code, and zip and code?

I did some test runs on LGTM and this turned out to be problematic.  `post`, `code` and `zip` are commonly used in variable and function names referring to technical things, so we get a few too many false positives.

I may try switching from `match` to regexps, and allow exactly 0 or 1 arbitrary character in between words in all of these cases - rather than doing the best we can with the options of any string (often too permissive) or nothing (too restrictive).

> Our tests would improve if we added %licese%key% as well (or maybe just %license%?). Do we catch more good results if we add those?

Planned as a follow-up, as I think that is closer to a password than private personal data so should be a change in `SensitiveExprs.qll` rather than `PrivateData.qll`.

geoffw0:
I've just pushed some changes:
 - updated the approach from using `match` to `regexpMatch`, which allows slightly more accurate expressions and thus slightly more accurate results (at least on tests).  A repeat LGTM run appears to have identical results.  Local testing suggests performance hasn't been affected much.
 - I've added a couple more test cases around the ones for `salary`, I wasn't happy with them as they weren't really exercising the existing code properly.
 - I've added the [CWE-359](https://cwe.mitre.org/data/definitions/359.html) tag to this query, as it now covers many of the types of information described in that CWE.

cpp/ql/lib/semmle/code/cpp/security/PrivateData.qll
geoffw0: I've now addressed your first point, the second will be addressed in a follow-up PR.

geoffw0:
... and I've deleted the experimental `PrivateData.qll` now as well.

MathiasVP (approved):
LGTM!

geoffw0:
Ah, thanks for doing a DCA run.  I was distracted yesterday and didn't get around to it.
 - there's a lost query result; it isn't from a query affected by `PrivateData.qll` or related changes here, so I think its wobble.
 - there are new extraction errors in `microsoft__calculator`; this PR contains no extractor changes and this project commonly wobbles, so again I think no cause for concern.
 - overall analysis time is almost unchanged.

geoffw0:
Merging...

MathiasVP:
> Ah, thanks for doing a DCA run. I was distracted yesterday and didn't get around to it.

No worries. It was really for testing something completely different 😂. But I'm glad you found it useful!

cpp/ql/lib/semmle/code/cpp/security/PrivateData.qll
geoffw0: Follow-up PR here: https://github.com/github/codeql/pull/8623

PR #8582, opened 2022-03-28
Java: Make `JumpStmt` a proper superclass

### LINKED ISSUE #8569: Java: Make `JumpStmt` a proper superclass
Currently CodeQL's `BreakStmt`, `YieldStmt` and `ContinueStmt` have no common superclass, and `JumpStmt` is defined separately. This has the following disadvantages:
- It makes it difficult for users to discover `JumpStmt` because at least in the documentation there is no relation between for example `BreakStmt` and `JumpStmt`
- The predicates for accessing the targeted element (a pretty common use case) are only defined for `JumpStmt`. This requires double casts to access these predicates, multiple of these can be seen in the CodeQL source code, e.g. `exit.(BreakStmt).(JumpStmt).getTarget()`

What do you think about making `JumpStmt` a proper superclass of these classes? This would solve the aforementioned issues.

Marcono1234:
Resolves #8569

### DIFF
--- a/java/ql/lib/change-notes/2022-03-28-JumpStmt-as-superclass.md
+++ b/java/ql/lib/change-notes/2022-03-28-JumpStmt-as-superclass.md
@@ -1,0 +1,4 @@
+---
+category: fix
+---
+* The QL class `JumpStmt` has been made the superclass of `BreakStmt`, `ContinueStmt` and `YieldStmt`. This allows directly using its inherited predicates without having to explicitly cast to `JumpStmt` first.
--- a/java/ql/lib/semmle/code/java/Statement.qll
+++ b/java/ql/lib/semmle/code/java/Statement.qll
@@ -556,8 +556,11 @@ class ThrowStmt extends Stmt, @throwstmt {
   override string getAPrimaryQlClass() { result = "ThrowStmt" }
 }
 
+// Workaround to avoid having to make JumpStmt abstract
+private class JumpStmt_ = @breakstmt or @yieldstmt or @continuestmt;
+
 /** A `break`, `yield` or `continue` statement. */
-abstract class JumpStmt extends Stmt {
+class JumpStmt extends Stmt, JumpStmt_ {
   /**
    * Gets the labeled statement that this `break` or
    * `continue` statement refers to, if any.
--- a/java/ql/lib/semmle/code/java/Statement.qll
+++ b/java/ql/lib/semmle/code/java/Statement.qll
@@ -556,7 +556,6 @@ class ThrowStmt extends Stmt, @throwstmt {
   override string getAPrimaryQlClass() { result = "ThrowStmt" }
 }
 
-// Workaround to avoid having to make JumpStmt abstract
 private class JumpStmt_ = @breakstmt or @yieldstmt or @continuestmt;
 
 /** A `break`, `yield` or `continue` statement. */
--- a/java/ql/lib/semmle/code/java/Statement.qll
+++ b/java/ql/lib/semmle/code/java/Statement.qll
@@ -557,13 +557,7 @@ class ThrowStmt extends Stmt, @throwstmt {
 }
 
 /** A `break`, `yield` or `continue` statement. */
-class JumpStmt extends Stmt {
-  JumpStmt() {
-    this instanceof BreakStmt or
-    this instanceof YieldStmt or
-    this instanceof ContinueStmt
-  }
-
+abstract class JumpStmt extends Stmt {
   /**
    * Gets the labeled statement that this `break` or
    * `continue` statement refers to, if any.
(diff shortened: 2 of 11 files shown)


java/ql/lib/change-notes/2022-03-28-JumpStmt-as-superclass.md
```diff
@@ -0,0 +1,4 @@
+---
+category: fix
```
Marcono1234, line 2: Not really a "fix", but the other categories do not seem to fit much better either.

java/ql/lib/semmle/code/java/Statement.qll
```diff
@@ -626,12 +616,26 @@ class BreakStmt extends Stmt, @breakstmt {
...
+  override SwitchExpr getTarget() {
+    // Get the innermost enclosing SwitchExpr; this works because getParent() is defined for Stmt and
+    // therefore won't proceed after the innermost SwitchExpr (due to it being an Expr)
+    result = this.getParent+()
+  }
+
+  /**
+   * DEPRECATED: `yield` statements cannot have a target label.
+   */
+  deprecated override LabeledStmt getTargetLabel() { none() }
```
Marcono1234, line 637: Have marked this as deprecated to prevent accidental usage. If that is a reasonable approach, it might be necessary to adjust [`config/blame-deprecations.mjs`](https://github.com/github/codeql/blob/main/config/blame-deprecations.mjs) (introduced by #8347) to not flag deprecated predicated which also have an `override` annotation. Otherwise this predicate override might be removed by accident in the future.

java/ql/lib/semmle/code/java/Statement.qll
```diff
@@ -584,19 +578,15 @@ class JumpStmt extends Stmt {
...
-    or
-    not exists(this.getSwitchExprTarget()) and
     result = this.getAPotentialTarget() and
     not exists(Stmt other | other = this.getAPotentialTarget() | other.getEnclosingStmt+() = result)
   }
 
   /**
    * Gets the statement or `switch` expression that this `break`, `yield` or `continue` jumps to.
    */
+  // Note: This implementation only considers `break` and `continue`; YieldStmt overrides this predicate
```
smowton, line 589: Move comment inside function to avoid breaking the qldoc apart from the code

java/ql/lib/semmle/code/java/Statement.qll
```diff
@@ -626,12 +616,26 @@ class BreakStmt extends Stmt, @breakstmt {
...
+    // Get the innermost enclosing SwitchExpr; this works because getParent() is defined for Stmt and
+    // therefore won't proceed after the innermost SwitchExpr (due to it being an Expr)
+    result = this.getParent+()
+  }
+
+  /**
+   * DEPRECATED: `yield` statements cannot have a target label.
+   */
+  deprecated override LabeledStmt getTargetLabel() { none() }
+
```
smowton, line 638:
```suggestion
```

Like your approach in theory but it's a bunch of work for a niche case

java/ql/lib/semmle/code/java/Statement.qll
Marcono1234: Obsolete, see https://github.com/github/codeql/pull/8582#discussion_r837659375

java/ql/lib/semmle/code/java/Statement.qll
```diff
@@ -557,13 +557,7 @@ class ThrowStmt extends Stmt, @throwstmt {
 }
```
aschackmull, line 558: We generally try to avoid exposing public `abstract` classes like this, since subclassing using `extends` can lead to surprising results. It is better to define the union type explicitly:
```suggestion

private class JumpStmt_ = @breakstmt or @yieldstmt or @continuestmt;

```

java/ql/lib/semmle/code/java/Statement.qll
```diff
@@ -557,13 +557,7 @@ class ThrowStmt extends Stmt, @throwstmt {
...
 
 /** A `break`, `yield` or `continue` statement. */
-class JumpStmt extends Stmt {
-  JumpStmt() {
-    this instanceof BreakStmt or
-    this instanceof YieldStmt or
-    this instanceof ContinueStmt
-  }
-
+abstract class JumpStmt extends Stmt {
```
aschackmull, line 560:
```suggestion
class JumpStmt extends Stmt, JumpStmt_ {
```

aschackmull (commented):
Generally looks good to me, but we should avoid making `JumpStmt` an `abstract` class.

java/ql/lib/semmle/code/java/Statement.qll
Marcono1234: Good point, thanks! I have implemented your suggestion manually, and added a small comment explaining the rationale; hopefully that is alright.

A bit unrelated, but I have created a https://github.com/github/codeql/discussions/8713 to discuss making implementing abstract classes less error-prone.

java/ql/lib/semmle/code/java/Statement.qll
```diff
@@ -556,14 +556,11 @@ class ThrowStmt extends Stmt, @throwstmt {
...
 }
 
-/** A `break`, `yield` or `continue` statement. */
-class JumpStmt extends Stmt {
-  JumpStmt() {
-    this instanceof BreakStmt or
-    this instanceof YieldStmt or
-    this instanceof ContinueStmt
-  }
+// Workaround to avoid having to make JumpStmt abstract
```
aschackmull, line 559: This is not really a workaround - I'd say it is the preferred solution.
```suggestion
```

aschackmull (approved):

PR #8589, opened 2022-03-29
ExponentialBackTracking: Speedup `concretise`

hvitved:


erik-krogh (commented):
You've achieved the same thing as I did in https://github.com/github/codeql/pull/8522, but much more concisely :clap: 

Could `SuperlinearBackTracking.qll` get the same treatment?

erik-krogh (dismissed):
:+1: (Assuming the new evaluations work out)

hvitved:
There was some slowdown on `facebookarchive__nuclide`. The latest commit tries to reduce the slowdown.

erik-krogh:
I think you need a merge with `main` to include https://github.com/github/codeql/pull/8532 (I think that's why Ruby is failing).

hvitved:
> I think you need a merge with `main` to include #8532 (I think that's why Ruby is failing).

Turns out that this was already broken on `main`: https://github.com/github/codeql/pull/8624.

hvitved:
Rebased onto latest `main` without changes.

erik-krogh (approved):
:+1: (assuming a new evaluation works out).

hvitved:
There is still a bit of a slowdown on `facebookarchive__nuclide`, but less than before. I would argue that it is an evaluator bug that `strictconcat` is not faster than the old recursive computation (I have made @alexet aware). @erik-krogh is it acceptable to merge?

erik-krogh:
> I would argue that it is an evaluator bug that `strictconcat` is not faster than the old recursive computation.  

I read your comments about that, and I agree. It sounds like there are some unnecessary sorting when `language[monotonicAggregates]` is enabled.  

> @erik-krogh is it acceptable to merge? 

Your solution is faster and more elegant than [the solution I was planning](https://github.com/github/codeql/pull/8522), it massively improves the worst-case performance, and if the evaluator is fixed then the performance should be better in all cases.   
So yeah, big :+1: from me.

PR #8595, opened 2022-03-29
Python : Add query to detect PAM authorization bypass

porcupineyhairs:
Using only a call to `pam_authenticate` to check the validity of a login can lead to authorization bypass vulnerabilities. A `pam_authenticate` only verifies the credentials of a user. It does not check if a user has an appropriate authorization to actually login. This means a user with a expired login or a password can still access the system.

This PR includes a qhelp describing the issue, a query which detects instances where a call to `pam_acc_mgmt` does not follow a call to `pam_authenticate` and it's corresponding tests.

Some of the public CVE's I can find using this query are :
* [CVE-2022-0860](https://nvd.nist.gov/vuln/detail/CVE-2022-0860) found in [cobbler/cobbler](https://www.github.com/cobbler/cobbler)
* [fredhutch/motuz](https://www.huntr.dev/bounties/d46f91ca-b8ef-4b67-a79a-2420c4c6d52b/)


python/ql/src/experimental/Security/CWE-285/PamAuthorization.ql
```diff
@@ -0,0 +1,58 @@
...
+ *       external/cwe/cwe-285
+ */
+
+import python
+import semmle.python.ApiGraphs
+import experimental.semmle.python.Concepts
+import semmle.python.dataflow.new.TaintTracking
+
+private class LibPam extends API::Node {
+  LibPam() {
```
RasmusWL, line 17: Instead of making this a class, please make it a predicate, such as

```codeql
API::Node libPam() {
  exists( ... | ... | result = cdll_call.getReturn() )
}
```

python/ql/src/experimental/Security/CWE-285/PamAuthorization.ql
```diff
@@ -0,0 +1,58 @@
...
+      find_library = API::moduleImport("ctypes.util").getMember("find_library") and
+      cdll_call = cdll.getACall() and
+      find_lib_call = find_library.getACall() and
+      DataFlow::localFlow(DataFlow::exprNode(str), find_lib_call.getArg(0)) and
+      str.getText() = "pam" and
+      cdll_call.getArg(0) = find_lib_call and
+      libpam = cdll_call.getReturn()
+    |
+      libpam = this
+    )
```
RasmusWL, line 32: You can reduce the number of variables here. For example, you can remove the `cdll` variable altogether by doing

```
cdll_call = API::moduleImport("ctypes").getMember("CDLL").getACall()
```

Notice that you should also be using camelCase for CodeQL code, so it should really be `cdllCall`

python/ql/src/experimental/Security/CWE-285/PamAuthorizationBad.py
```diff
@@ -0,0 +1,15 @@
+def authenticate(self, username, password, service='login', encoding='utf-8', resetcreds=True):
+  libpam                    = CDLL(find_library("pam"))
+  pam_authenticate          = libpam.pam_authenticate
+  pam_acct_mgmt          = libpam.pam_acct_mgmt
```
RasmusWL, line 4:
```suggestion
```

python/ql/src/experimental/Security/CWE-285/PamAuthorizationBad.py
```diff
@@ -0,0 +1,15 @@
...
+  pam_authenticate.restype  = c_int
+  pam_authenticate.argtypes = [PamHandle, c_int]
+  pam_acct_mgmt.restype  = c_int
+  pam_acct_mgmt.argtypes = [PamHandle, c_int]
```
RasmusWL, line 8:
```suggestion
```

python/ql/test/experimental/query-tests/Security/CWE-285/good.py
```diff
@@ -0,0 +1,97 @@
...
+        handle = PamHandle()
+        conv   = PamConv(my_conv, 0)
+        retval = pam_start(service, username, byref(conv), byref(handle))
+
+        retval = pam_authenticate(handle, 0)
+        if retval == 0:
+            retval = pam_acct_mgmt(handle, 0)
+        auth_success = retval == 0
+
+        return auth_success
```
RasmusWL, line 97: It would have been much easier to read your test code if you had a _single_ test file, with two methods. `authenticate_good` and `authenticate_bad`.

python/ql/src/experimental/Security/CWE-285/PamAuthorization.ql
```diff
@@ -0,0 +1,58 @@
...
+}
+
+from PamAuthCall p, API::CallNode u, Expr handle
+where
+  u = p.getACall() and
+  handle = u.asExpr().(Call).getArg(0) and
+  not exists(PamActMgt pam |
+    DataFlow::localFlow(DataFlow::exprNode(handle),
+      DataFlow::exprNode(pam.getACall().asExpr().(Call).getArg(0)))
+  )
```
RasmusWL, line 57: This can be simplified as

```suggestion
from PamAuthCall p, API::CallNode u, DataFlow::Node handle
where
  u = p.getACall() and
  handle = u.getArg(0) and
  not exists(PamActMgt pam | DataFlow::localFlow(handle, pam.getACall().getArg(0)))
```

But I think in the end, I would write something like this myself

```codeql
from API::CallNode authenticateCall, DataFlow::Node handle
where
  authenticateCall = libPam().getMember("pam_authenticate").getACall() and
  handle = authenticateCall.getArg(0) and
  not exists(API::CallNode acctMgmtCall | 
    acctMgmtCall = libPam().getMember("pam_acct_mgmt").getACall() and
    DataFlow::localFlow(handle, acctMgmtCall.getArg(0)
  )
select ...
```

eliminating much of the overhead.

python/ql/test/experimental/query-tests/Security/CWE-285/good.py
```diff
@@ -0,0 +1,97 @@
...
+pam_start.argtypes        = [c_char_p, c_char_p, POINTER(PamConv), POINTER(PamHandle)]
+
+pam_setcred               = libpam.pam_setcred
+pam_setcred.restype       = c_int
+pam_setcred.argtypes      = [PamHandle, c_int]
+
+pam_strerror              = libpam.pam_strerror
+pam_strerror.restype      = c_char_p
+pam_strerror.argtypes     = [PamHandle, c_int]
+
```
RasmusWL, line 64: These lines doesn't seem to be needed for this example code.

```suggestion
```

python/ql/src/experimental/Security/CWE-285/PamAuthorization.ql
```diff
@@ -0,0 +1,58 @@
...
+  override string toString() { result = "pam_authenticate" }
+}
+
+class PamActMgt extends API::Node {
+  PamActMgt() { exists(LibPam pam | pam.getMember("pam_acct_mgmt") = this) }
+
+  override string toString() { result = "pam_acct_mgmt" }
+}
+
+from PamAuthCall p, API::CallNode u, Expr handle
```
RasmusWL, line 50: it is unclear what this `u` is from the name itself.

RasmusWL (changes requested):
Besides inline comments, I would like you to merge the test files to a single, and remove all the stuff that is not absolutely required to make the example work (I highlight a bit of it, but I think there are more things that can be removed).

porcupineyhairs:
@RasmusWL Thanks for the quick response. I have made the changes you sought. Any thoughts on how I could handle the `clusterlabs/pcs` case I mentioned above?

RasmusWL:
> I have made the changes you sought. 

Thanks :pray: I made an other minor adjustment, simplifying the code a little bit more. Hope you don't mind.

> Any thoughts on how I could handle the `clusterlabs/pcs` case I mentioned above?

This is sadly a known problem, which is highlighted by [this bit of test code](https://github.com/github/codeql/blob/1d25184b3214078361865d579d3c0a6fd63f61e5/python/ql/test/experimental/dataflow/typetracking/test.py#L33-L36). In the code below, we do currently not track that `also_x` is a result from `foo()` :disappointed: So nothing you can do about it, but thanks for highlighting it :+1:

```
def id(obj):
    return obj

x = foo()
also_x = id(x)
```

RasmusWL (dismissed):
I think this looks ok now :+1:

porcupineyhairs:
@RasmusWL Ping.

RasmusWL:
Thanks. I've just gotten back from vacation. Need to do a little internal testing of this before merging.

RasmusWL (approved):
Had a chance to look at some results, which looks good to me, so added `@precision high`. Will merge this PR once tests turn green.

PR #8596, opened 2022-03-29
C++: IR data flow through global variables

rdmarsh2:


rdmarsh2:
Opened as draft since it builds on #8515

rdmarsh2:
@geoffw0 there's some new false positives in the CWE-611 tests, and I'm not sure whether they're an issue with this PR or just exposing a known issue with the CWE-611 query. Could you take a look at those?

cpp/ql/test/query-tests/Security/CWE/CWE-497/semmle/tests/tests.cpp
```diff
@@ -68,9 +68,9 @@ void test2(bool cond)
 
 	maybe = cond ? global_token : global_other;
 	
-	printf("token = '%s'\n", global_token); // BAD: outputs SECRET_TOKEN environment variable [NOT DETECTED]
+	printf("token = '%s'\n", global_token); // BAD: outputs SECRET_TOKEN environment variable
 	printf("other = '%s'\n", global_other);
-	printf("maybe = '%s'\n", maybe); // BAD: may output SECRET_TOKEN environment variable [NOT DETECTED]
+	printf("maybe = '%s'\n", maybe); // BAD: may output SECRET_TOKEN environment variable
```
geoffw0, line 73: Excellent!

cpp/ql/test/query-tests/Security/CWE/CWE-497/semmle/tests/tests2.cpp
```diff
@@ -79,7 +79,7 @@ void test1()
 
 		send(sock, mysql_get_client_info(), val(), val()); // BAD
 		send(sock, buffer, val(), val()); // BAD
-		send(sock, global1, val(), val()); // BAD [NOT DETECTED]
+		send(sock, global1, val(), val()); // BAD
```
geoffw0, line 82: Great. :+1:

cpp/ql/test/query-tests/Security/CWE/CWE-611/tests3.cpp
```diff
@@ -35,14 +35,14 @@ void test3_2(InputSource &data) {
...
-	p_3_3->parse(data); // BAD (parser not correctly configured) [NOT DETECTED]
+	p_3_3->parse(data); // BAD (parser not correctly configured)
 }
 
 SAX2XMLReader *p_3_4 = XMLReaderFactory::createXMLReader();
 
 void test3_4(InputSource &data) {
 	p_3_4->setFeature(XMLUni::fgXercesDisableDefaultEntityResolution, true);
-	p_3_4->parse(data); // GOOD
+	p_3_4->parse(data); // GOOD [FALSE POSITIVE]
```
geoffw0, line 45: It looks like barriers aren't working for global variables?  I'm not 100% sure whether the problem lies in IR data flow or in the way this query defines its barriers.

I don't think we should tolerate these false positives (because they are likely to be more common than new good results in the wild), but if you can't easily fix them in the IR dataflow I will have a look at the query, and possibly just block results involving global variables for now.  That will undo all of the changes for this query, but keep the (valuable) improvements in the other queries.

Let me know if you think there's an IR dataflow fix first.

geoffw0 (commented):
Reviewed the changes in CPP security query test results only.

cpp/ql/test/query-tests/Security/CWE/CWE-611/tests3.cpp
MathiasVP: I _think_ the issue might be in the way we define transformers using GVN in `DomConfigurationSetParameterTransformer`. It's a totally reasonable way to do it, but since we don't try to do any alias analysis for global variables, the GVNs we get for uses of global variables will pretty much always be unique, which makes GVN for these values quite useless (please correct me if I'm misremembering what the IR actually looks like for uses of global variables @rdmarsh2).

I don't think we have any good way of doing GVN for global variables, so
I think the best solution is to disable detection in the case of global variables in the problematic XML libraries (I guess it's really all of them?).

If we had use-use dataflow (😭) we could use that instead of GVN, and that would support global variables as well.

geoffw0: I've created https://github.com/rdmarsh2/ql/pull/5/files as a workaround, but I'm open to suggestions if we can do better.

rdmarsh2: The way this is implemented now, every read of a global variable has flow from every write to that variable (via the variable node). I think that's what's causing the issue here - the initial object creation *also* flows to the use. I'm not sure I can fix that with a reasonable amount of work, though...

MathiasVP:
The performance on `vim/vim` looks like something we should investigate before we merge this.

jketema (commented):
Two naive questions to start, just to further my understanding.

cpp/ql/lib/semmle/code/cpp/ir/dataflow/internal/DataFlowUtil.qll
```diff
@@ -100,7 +100,7 @@ class Node extends TIRDataFlowNode {
   Declaration getEnclosingCallable() { none() } // overridden in subclasses
 
   /** Gets the function to which this node belongs, if any. */
-  Function getFunction() { none() } // overridden in subclasses
+  Declaration getFunction() { none() } // overridden in subclasses
```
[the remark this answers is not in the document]
rdmarsh2: Because we're no longer talking (exclusively) about functions - we can now have code in global variable initializers that shows up in the dataflow graph, and there's places that assume `getFunction` has a result for those nodes.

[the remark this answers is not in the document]
jketema: Thanks for the explanation. I was aware of the related changes in https://github.com/github/codeql/pull/8515. It does make me wonder whether we should spend some time renaming predicates to better reflect what they're doing (doesn't have to be in this PR).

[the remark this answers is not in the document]
jketema: I also see that in some places `func` was renamed to `decl` and in other places it wasn't. Is that on purpose?

cpp/ql/lib/semmle/code/cpp/ir/implementation/raw/internal/IRConstruction.qll
```diff
@@ -35,29 +36,41 @@ module Raw {
   cached
   predicate functionHasIR(Function func) { exists(getTranslatedFunction(func)) }
 
+  cached
+  predicate varHasIRFunc(GlobalOrNamespaceVariable var) {
```
jketema, line 40: I prefer to have this spelled out, as that's what's done for other predicates.
```suggestion
  predicate varHasIRFunction(GlobalOrNamespaceVariable var) {
```

cpp/ql/lib/semmle/code/cpp/ir/implementation/IRConfiguration.qll
```diff
@@ -25,7 +25,7 @@ class IRConfiguration extends TIRConfiguration {
    * of debug strings for IR that will not be dumped. We still generate the actual IR for these
    * functions, however, to preserve the results of any interprocedural analysis.
    */
-  predicate shouldEvaluateDebugStringsForFunction(Language::Function func) { any() }
+  predicate shouldEvaluateDebugStringsForFunction(Language::Declaration func) { any() }
```
jketema, line 28:
```suggestion
  predicate shouldEvaluateDebugStringsForFunction(Language::Declaration decl) { any() }
```

cpp/ql/lib/semmle/code/cpp/ir/implementation/internal/TIRVariable.qll
```diff
@@ -2,21 +2,21 @@ private import TIRVariableInternal
 private import Imports::TempVariableTag
 
 newtype TIRVariable =
-  TIRUserVariable(Language::Variable var, Language::LanguageType type, Language::Function func) {
+  TIRUserVariable(Language::Variable var, Language::LanguageType type, Language::Declaration func) {
```
jketema, line 5:
```suggestion
  TIRUserVariable(Language::Variable var, Language::LanguageType type, Language::Declaration decl) {
```
Also in a few cases below. I'll stop commenting on this now. I just think this is more consistent. I don't have any problems off-loading this in a separate PR with is just some consistency fixes.

cpp/ql/lib/semmle/code/cpp/ir/implementation/raw/internal/IRConstruction.qll
```diff
@@ -35,29 +36,41 @@ module Raw {
...
+    var.hasInitializer() and
+    (
+      not var.getType().isDeeplyConst()
```
jketema, line 43: This, I assume, is to avoid some of the pref problems you were running into? Maybe worth a comment?

cpp/ql/lib/semmle/code/cpp/ir/implementation/raw/internal/IRConstruction.qll
```diff
@@ -35,29 +36,41 @@ module Raw {
...
+      or
+      var.getInitializer().getExpr() instanceof StringLiteral
```
jketema, line 45: I assume these are constant too? Any reason why we do want to keep these in?

cpp/ql/lib/semmle/code/cpp/ir/implementation/raw/internal/TranslatedElement.qll
```diff
@@ -117,7 +118,8 @@ private predicate ignoreExprOnly(Expr expr) {
   // should not be translated.
   exists(NewOrNewArrayExpr new | expr = new.getAllocatorCall().getArgument(0))
   or
-  not translateFunction(expr.getEnclosingFunction())
+  not translateFunction(expr.getEnclosingFunction()) and
+  not Raw::varHasIRFunc(expr.getEnclosingVariable())
```
jketema, line 122:
```suggestion
  not Raw::varHasIRFunction(expr.getEnclosingVariable())
```

jketema (commented):
Next batch of comments. I'm slowly working my way through this. Feel free to go ahead and fixing things even though I'm not yet done (whatever your preferred way of working is).

cpp/ql/lib/semmle/code/cpp/ir/implementation/raw/internal/TranslatedExpr.qll
```diff
@@ -787,7 +790,7 @@ class TranslatedThisExpr extends TranslatedNonConstantExpr {
 
   override IRVariable getInstructionVariable(InstructionTag tag) {
     tag = ThisAddressTag() and
-    result = this.getEnclosingFunction().getThisVariable()
+    result = this.getEnclosingFunction().(TranslatedFunction).getThisVariable()
```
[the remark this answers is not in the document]
rdmarsh2: That only applies to the translation of a `this` expression, which doesn't exist for global/namespace variables.

cpp/ql/lib/semmle/code/cpp/ir/implementation/raw/internal/IRConstruction.qll
rdmarsh2: Mostly that constant integers will end up getting inlined in IR generation, so we don't need IR for them.

cpp/ql/lib/semmle/code/cpp/ir/implementation/raw/internal/IRConstruction.qll
jketema: > Maybe that applies to all non-numeric types, though...

If that's true. What's the consequence of that at the moment? Just inconsistent IR that will be captured by the consistency check you also introduce in this PR? If so, I think this can be addressed later when we see that this is a problem.

cpp/ql/lib/semmle/code/cpp/ir/implementation/raw/internal/TranslatedGlobalVar.qll
```diff
@@ -0,0 +1,132 @@
...
+      exists(VariableAccess access |
+        access.getTarget() = varUsed and
+        access.getEnclosingVariable() = var
+      )
+      or
+      var = varUsed
+      or
+      varUsed.(LocalScopeVariable).getEnclosingElement*() = var
+      or
+      varUsed.(Parameter).getCatchBlock().getEnclosingElement*() = var
```
jketema, line 124: Can these two cases actually happen?

cpp/ql/lib/semmle/code/cpp/ir/implementation/raw/internal/TranslatedGlobalVar.qll
```diff
@@ -0,0 +1,132 @@
...
+    kind instanceof GotoEdge and
+    (
+      tag = EnterFunctionTag() and
+      result = this.getInstruction(AliasedDefinitionTag())
+      or
+      tag = AliasedDefinitionTag() and
+      result = this.getInstruction(InitializerVariableAddressTag())
+      or
+      tag = InitializerVariableAddressTag() and
+      result = getChild(1).getFirstInstruction()
```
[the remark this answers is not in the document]
jketema: Please fix.

cpp/ql/lib/semmle/code/cpp/ir/implementation/raw/internal/IRConstruction.qll
rdmarsh2: If there were an interesting source in the initializer of a global variable with a struct or pointer type (probably flowing via a field), we wouldn't currently get taint flow from it.

cpp/ql/lib/semmle/code/cpp/ir/implementation/raw/internal/TranslatedGlobalVar.qll
```diff
@@ -0,0 +1,132 @@
...
+  final override GlobalOrNamespaceVariable getAst() { result = var }
+
+  final override Declaration getFunction() { result = var }
+
+  final Location getLocation() { result = var.getLocation() }
+
+  override Instruction getFirstInstruction() { result = this.getInstruction(EnterFunctionTag()) }
+
+  override TranslatedElement getChild(int n) {
+    n = 1 and
```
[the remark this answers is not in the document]
rdmarsh2: Only the order matters for these, not the specific number. `0` would work just as well.

cpp/ql/lib/semmle/code/cpp/ir/implementation/raw/internal/IRConstruction.qll
jketema: Ok, but that's equal to the situation we had before this PR, right?

cpp/ql/lib/semmle/code/cpp/ir/implementation/raw/internal/TranslatedGlobalVar.qll
rdmarsh2: Yes - there's a [GCC extension](https://gcc.gnu.org/onlinedocs/gcc/Statement-Exprs.html) that allows statements in an expression context.

cpp/ql/lib/semmle/code/cpp/ir/implementation/raw/internal/TranslatedGlobalVar.qll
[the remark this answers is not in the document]
jketema: Looking at the rest of the code, we generally seem to start at 0, but this is truly nitpicking, so feel free to ignore this.

cpp/ql/lib/semmle/code/cpp/ir/implementation/raw/internal/TranslatedGlobalVar.qll
jketema: Ok.

jketema:
Looking at the DCA experiment: did you look whether the alert changes made sense?

cpp/ql/lib/semmle/code/cpp/ir/implementation/raw/internal/IRConstruction.qll
rdmarsh2: Right.

jketema:
@rdmarsh2 I think there's still the following to resolve:
* code scanning alerts
* open question regarding DCA
* merge conflicts

jketema:
There's also still: https://github.com/github/codeql/pull/8596#discussion_r920049933 Is this something that needs to be looked at on the extractor side?

cpp/ql/test/library-tests/syntax-zoo/aliased_ssa_consistency.expected
```diff
@@ -98,6 +98,9 @@ thisArgumentIsNonPointer
...
 | pointer_to_member.cpp:23:5:23:54 | Call: call to expression | Call instruction 'Call: call to expression' has a `this` argument operand that is not an address, in function '$@'. | pointer_to_member.cpp:14:5:14:9 | int usePM(int PM::*) | int usePM(int PM::*) |
 | pointer_to_member.cpp:24:5:24:49 | Call: call to expression | Call instruction 'Call: call to expression' has a `this` argument operand that is not an address, in function '$@'. | pointer_to_member.cpp:14:5:14:9 | int usePM(int PM::*) | int usePM(int PM::*) |
+nonUniqueIRVariable
+| misc.c:178:22:178:40 | VariableAddress: __PRETTY_FUNCTION__ | Variable address instruction 'VariableAddress: __PRETTY_FUNCTION__' has no associated variable, in function '$@'. | misc.c:177:6:177:14 | void magicvars() | void magicvars() |
+| misc.c:179:27:179:34 | VariableAddress: __func__ | Variable address instruction 'VariableAddress: __func__' has no associated variable, in function '$@'. | misc.c:177:6:177:14 | void magicvars() | void magicvars() |
```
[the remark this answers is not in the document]
rdmarsh2: This probably is something we need to look at eventually. The consistency check is new, and I think these don't have to do with the rest of the PR.

rdmarsh2:
The DCA alert changes made sense, yes. Responded to that last comment - I'm not sure if they need work on the extractor side or if they need special handling in IR generation.

cpp/ql/test/library-tests/syntax-zoo/aliased_ssa_consistency.expected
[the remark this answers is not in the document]
jketema: Agreed that this doesn't have anything to do with this PR.

jketema:
LGTM. I've kicked off a final DCA experiment, as we changed quite a few things related to (global) variables on the extractor side since this PR was opened. I'll approve once the experiment is done (unless we find some serious problem that is new compared to the previous DCA experiment, which probably isn't very likely).

jketema:
New DCA results are in:
* There's a significant increase in the number of alerts, which makes sense given the extractor related issues we fixed and as we added/enabled systemd and Samate. The alters look ok, but a second pair of eyes would be welcome. 
* The second vim run OOM'ed this might either be related to the OOM problems we were having anyway, or due to a bad join order. Have the join orders been looked at?

MathiasVP:
> There's a significant increase in the number of alerts (compared to the earlier DCA experiment), which makes sense given the extractor related issues we fixed and as we added/enabled systemd and Samate. The alerts look ok to me, but a second pair of eyes would be welcome.

So far:
- All the new results on `cpp/command-line-injection` are new TPs involving global variables 🎉

... I'll update this post as I look through the other new results.

rdmarsh2:
Looks like the OOM is on ExecTainted.ql - all other queries finish before the error. I'll take a deeper look at it.

jketema:
> Looks like the OOM is on ExecTainted.ql - all other queries finish before the error. I'll take a deeper look at it.

You might want to have a brief sync with Mathias about it. If you can't track it down I would suggest repeating the vim part of the DCA experiment with a **lower** `--ram` option than the `14979` currently used. I can explain via our internal channels.

jketema:
I if see correctly, the re-run of DCA on vim was without the `--ram` option, because my directions were slightly wrong. Soory! Nevertheless, the re-run was successful, and the slowdown doesn't seem to have changed compared to the previous run. I propose we go ahead and merge this, and that if we see DCA OOMs again on vim, we move vim to a slightly bigger runner.

jketema (approved):

PR #8600, opened 2022-03-30
C#: Dotnet Runtime models.

michaelnebel:


michaelnebel:
First execution of DCA gave the following *increase* in alerts.
image

michaelnebel:
The recent commit, where flow summaries for simple types and higher order methods have been removed, yielded a lot fewer results. Furthermore, some tests are still failing.

image

michaelnebel:
I have tried to manually inspect the 18 new alerts that are being produced after adding all the CSV rows.
In most cases it is possible to just that flow is now identified as a suitable summary has been added as a "missing link" to identify flow.

michaelnebel:
If we only allow a single read/write (and this generate fewer summaries) we approximately get the same number of alerts as before (number goes from 18 -> 16).

image

michaelnebel:
@hvitved 
Comments to the new alerts
- Most traces are too long to figure out.
- The ZipSlip are more interesting. entry.Fullname now identifies as something that might be tainted. TryAdd and Remove for dictionaries doesn’t have any summaries for Remove and TryAdd. Is that why they are identified as file system operations?
- The generated summaries in many cases contains summaries for Clear, Remove and RemoveRange methods. Should we make a special case in the generator to avoid summary generation of these methods?

michaelnebel:
> @hvitved Comments to the new alerts
> 
> * Most traces are too long to figure out.
> * The ZipSlip are more interesting. entry.Fullname now identifies as something that might be tainted. TryAdd and Remove for dictionaries doesn’t have any summaries for Remove and TryAdd. Is that why they are identified as file system operations?
> * The generated summaries in many cases contains summaries for Clear, Remove and RemoveRange methods. Should we make a special case in the generator to avoid summary generation of these methods?
> * Performance of the DCA run looks acceptable. Total analysis time is only 3%-4% slower.

Running the queries that generates the new alerts locally made it possible to identify summaries used for a step in the path that could explain the new alerts
- User-controlled-bypass: New alerts takes a step using the NameValueCollection string index’er.
- Xml-injection: New alerts takes a step using the NameValueCollection string index’er.
- Zip-slip: New zip-slip alerts can be explained by summary for Uri.LocalPath, Element.Name, Name.LocalName

michaelnebel:
@hvitved 
Latest DCA run shows that
* Performance of executing the queries with the dotnet runtime models is acceptable (see screemshot below).
* The alerts are explainable (see comment above).
image

However, there appears to be and issue with the *mono* project (the execution has failed twice, but the workflow run doesn't yield much information about the failure except that a log is printed on the worker agent).
I will try and re-base the PR and execute and the DCA run again.
If it fails again, we will need to investigate. What would be the best way to do this? Is that downloading the *mono* database and execute the queries locally or is it somehow possible to login in the workflow agents and download the log files?

michaelnebel:
Segmentation fault during DCA execution.
Will try again.

michaelnebel:
DCA confirms that the "bootstrap" issue of including models has now been solved (due to the ranking og summaries and source code):
image

michaelnebel:
@hvitved : Looks like we are close to a merge. Boot strap problem is solved. Analysis of an ordinary DCA is seen below.

The performance of executing all queries seems acceptable (even though dotnet runtime seems a bit slower). The overall slowdown is only reported as being around 3%.
image


We now see many new alerts, but are mostly due to the fact that the analysis of the mono repository is now completed.
The new alerts summary is:
image

michaelnebel:
I looked into a couple of the new alerts for dotnet. As expected they can't be explained by a summary anymore (as source code should take precedence over (generated) summaries).
However, an example of a step in one of the new cs/zipslip query can be seen below. It almost seems like .Clear isn't considered a sanitizer of the data anymore.
Besides adding the dotnet runtime models, we also made some semantical changes on this PR
(1) Ranking of summaries and source code.
(2) Removed the SummarizedCallableDefaultClearsContent class.

The question is now. Should we accept the PR as is, or should we re-factor the semantical change components into seperate PRs that we can execute DCA and get a closer look on the alerts discrepancy (not sure if I am sufficiently cautious.
Any advice is welcome 😄 
image

csharp/ql/lib/semmle/code/csharp/dataflow/internal/DataFlowDispatch.qll
```diff
@@ -83,13 +83,23 @@ newtype TReturnKind =
     )
   }
 
+predicate useFlowSummary(FlowSummary::SummarizedCallable c) {
```
michaelnebel, line 86: Needs QL doc.

hvitved (approved):
🎉

michaelnebel:
It looks like we are finally good to go!
Comments on the latest DCA run
(1) There is a performance regression of around 3%. It is expected that we get some regression when including 10k models - so that is fine.
(2) There is a difference in alerts for the information exposure query on the mono project. The query was re-written, which reduced the multiplicity in results. I have manually run the query against main and against this branch and we get at the exact same query locations. IMO this looks fine as well.

Nothing can possibli go wrong - I mean nothing can possibly go wrong - that is the first thing that has ever gone wrong...
 
@hvitved: Are we ready for launch?

hvitved:
> Nothing can possibli go wrong - I mean nothing can possibly go wrong - that is the first thing that has ever gone wrong...

😆 

> @hvitved: Are we ready for launch?

:shipit:

PR #8601, opened 2022-03-30
Python: Flask: Identify body contents passed via named response parameter in invocations of Response constructor

zbazztian:
As witnessed [here](https://github.com/Netflix/security_monkey/blob/develop/security_monkey/auth/modules.py#L266).

Note that Flask's [Response](https://github.com/pallets/flask/blob/main/src/flask/wrappers.py#L136) class inherits from `werkzeug`'s [Response](https://github.com/pallets/werkzeug/blob/main/src/werkzeug/wrappers/response.py#L172) class.

### DIFF
--- a/python/ql/lib/change-notes/2022-03-30-flask-recognize-body-param.md
+++ b/python/ql/lib/change-notes/2022-03-30-flask-recognize-body-param.md
@@ -1,4 +1,4 @@
 ---
 category: minorAnalysis
 ---
-* The body content of those Flask `Response` objects which were constructed by explicitly referring to the `response` parameter is now detected.
+* Improved modeling of Flask `Response` objects, so passing a response body with the keyword argument `response` is now recognized.
--- a/python/ql/lib/change-notes/2022-03-30-flask-recognize-body-param.md
+++ b/python/ql/lib/change-notes/2022-03-30-flask-recognize-body-param.md
@@ -1,0 +1,4 @@
+---
+category: minorAnalysis
+---
+* The body content of those Flask `Response` objects which were constructed by explicitly referring to the `response` parameter is now detected.
--- a/python/ql/test/library-tests/frameworks/flask/response_test.py
+++ b/python/ql/test/library-tests/frameworks/flask/response_test.py
@@ -37,22 +37,28 @@ def html4():  # $requestHandler
 
 @app.route("/html5")  # $routeSetup="/html5"
 def html5():  # $requestHandler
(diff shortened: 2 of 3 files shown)


python/ql/test/library-tests/frameworks/flask/response_test.py
```diff
@@ -57,6 +57,11 @@ def html7():  # $requestHandler
     resp.set_data("<h1>hello</h1>")  # $ MISSING: responseBody="<h1>hello</h1>"
     return resp  # $ SPURIOUS: HttpResponse mimetype=text/html responseBody=resp
 
+@app.route("/html8")  # $routeSetup="/html8"
+def html8():  # $requestHandler
+    resp = Response(response="<h1>hello</h1>")  # $HttpResponse mimetype=text/html responseBody="<h1>hello</h1>"
+    return resp  # $ SPURIOUS: HttpResponse mimetype=text/html responseBody=resp
+
```
yoff, line 64: I would expect this test case to appear right after `html4`, which, ironically, would entail a bunch of renames...

yoff (changes requested):
I am going to be annoying and ask you to rearrange the code slightly, but otherwise LTGM.
And thanks for fixing this oversight!

python/ql/test/library-tests/frameworks/flask/response_test.py
yoff: (I seem unable to construct a suggestion for this..)

zbazztian: Haha, ok, will do.

zbazztian:
Ok, addressed. I also squashed everything into a single commit.

yoff (dismissed):
Thanks for the fix.

python/ql/lib/change-notes/2022-03-30-flask-recognize-body-param.md
```diff
@@ -0,0 +1,4 @@
+---
+category: minorAnalysis
+---
+* The body content of those Flask `Response` objects which were constructed by explicitly referring to the `response` parameter is now detected.
```
RasmusWL, line 4:
```suggestion
* Improved modeling of Flask `Response` objects, so passing a response body with the keyword argument `response` is now recognized.
```

RasmusWL (dismissed):
Thanks for doing this :+1: I have a minor suggestion for the change-note

zbazztian:
@RasmusWL Thanks, I committed your suggestion.

RasmusWL (approved):

PR #8603, opened 2022-03-30
JavaScript: Improve detection of UMD modules.

max-schaefer:
Specifically, we now detect this pattern:

```js
(function (global, factory) {
    checkForNodeJS() ? module.exports = factory(...) :
    checkForAmd() ? define(..., factory) :
    global.mymodule = factory(...)
}(this, function (...) { ... });
```

We previously missed it because we expected the `define` to appear as an expression statement, but here it is nested a bit more deeply.

I fixed this by using the `inVoidContext` predicate we use in a few other queries, which however needed a bit of extra work. I'm curious to see whether that has any knock-on effects on performance or results, if it does I'm happy to switch to a more targeted fix.

### DIFF
--- a/javascript/ql/lib/Expressions/ExprHasNoEffect.qll
+++ b/javascript/ql/lib/Expressions/ExprHasNoEffect.qll
@@ -37,6 +37,8 @@ predicate inVoidContext(Expr e) {
   )
   or
   exists(LogicalBinaryExpr logical | e = logical.getRightOperand() and inVoidContext(logical))
+  or
+  exists(ConditionalExpr cond | e = cond.getABranch() | inVoidContext(cond))
 }
 
 /**
--- a/javascript/ql/lib/semmle/javascript/AMD.qll
+++ b/javascript/ql/lib/semmle/javascript/AMD.qll
@@ -204,12 +204,21 @@ private class ConstantAmdDependencyPathElement extends PathExpr, ConstantString 
 }
 
 /**
+ * Holds if `nd` is nested inside an AMD module definition.
+ */
+private predicate inAmdModuleDefinition(AstNode nd) {
+  nd.getParent() instanceof AmdModuleDefinition
+  or
+  inAmdModuleDefinition(nd.getParent())
+}
+
+/**
  * Holds if `def` is an AMD module definition in `tl` which is not
  * nested inside another module definition.
  */
 private predicate amdModuleTopLevel(AmdModuleDefinition def, TopLevel tl) {
   def.getTopLevel() = tl and
-  not def.getParent+() instanceof AmdModuleDefinition
+  not inAmdModuleDefinition(def)
 }
 
 /**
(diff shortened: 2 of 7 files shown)


javascript/ql/lib/Expressions/ExprHasNoEffect.qll
```diff
@@ -38,6 +38,8 @@ predicate inVoidContext(Expr e) {
   )
   or
   exists(LogicalBinaryExpr logical | e = logical.getRightOperand() and inVoidContext(logical))
+  or
+  exists(ConditionalExpr cond | e = cond.getABranch())
```
erik-krogh, line 42: Don't you need to add `and inVoidContext(cond)`?

max-schaefer: 🙈

max-schaefer: That would indeed have had a bit of an effect on results! Now fixed.

erik-krogh (approved):
LGTM assuming an evaluation works out.  
(Let me know if you want me to do one for you).

max-schaefer:
> Let me know if you want me to do one for you

That would be fantastic, thanks! (The language tests pass.)

erik-krogh:
> > Let me know if you want me to do one for you
> 
> That would be fantastic, thanks! (The language tests pass.)

I've started one. It should appear as a backref shortly.

max-schaefer:
The evaluation revealed significant performance regressions that need looking into.

max-schaefer:
I took a quick look at the regression, but didn't make much progress. It's easy to reproduce by, for example, running `ClientSideUrlRedirect.ql` on `openlayers/ol2`, which goes from 2s to 105s. We seem to be evaluating a whole load of predicates we previously didn't evaluate (or perhaps they were very quick, I don't know how to read logs anymore these days). I can't justify spending more time on this right now, so I'll just leave this open as a draft in case anyone else wants to pick it up later.

erik-krogh:
I've looked a bit at this, and I think I we could merge it now.  

[New evaluation](https://github.com/github/codeql-dca-main/tree/data/erik-krogh/umd-default-code-scanning__1/reports).  

Some of the performance regression seem to have disappeared from just waiting (core improvements?).    
But there is still a performance regression left.  

However, the cause seems to be that we recognize more flowsteps.   
There are several new results, and all the new results (I sampled a few) seem to be from us having a more complete callgraph, which is caused from recognizing more UMD modules.   
Some of those new results are eventually FPs (e.g. flow through `_.escape(..)` in `adobe/brackets`), but the new edge in the callgraph is good.  

I think the trade-off is worth it.

max-schaefer:
> I think we could merge it now

Ah, self-fixing performance problems! Those are the best.

asgerf (approved):

PR #8604, opened 2022-03-30
JS: refactor most library models away from AST nodes

erik-krogh:
Initially I started with refactoring the HTTP models to use dataflow nodes.  
That was done surprisingly quick, so I continued with more library models.  

I've made deprecated aliases where a renaming made sense, and otherwise I just changed the type. 

The `flowsTo` predicates in the HTTP models were confusing when starting to use `DataFlow::Node`. (I initially made mistakes because of that). So I've deprecated those `flowsTo` predicates. 

## Evaluation

[An evaluation looks OK](https://github.com/github/codeql-dca-main/tree/data/erik-krogh/httpNode__nightly-old__code-scanning__3/reports). But there are some new results and some lost results.  

### New results:  
All appear to be TPs, and they are all from recognizing destructuring assignments as property reads, which allows us to recognize more sources.   
Two of the results appeared from refactoring the HTTP models (the `js/path-injection` results).  
And two other appeared from refactoring the `SensitiveExpr` class (the `js/insufficient-password-hash` resutls).  

### Lost results:  

Both lost results are a consequence of a `DataFlow::PropWrite` having no result for `getRhs()` when the AST looks like:  
```JavaScript
obj.prop += value
```   
One appears to be a TP, the other appears to be an FP.  

I should do some followup work to investigate if we can have a sensible result for `getRhs()` in the above example.  (After this PR has merged).

### DIFF
--- a/javascript/ql/lib/change-notes/2022-04-04-dataflow-models.md
+++ b/javascript/ql/lib/change-notes/2022-04-04-dataflow-models.md
@@ -5,7 +5,7 @@ category: breaking
   The types of some classes have been changed, and these changes may break existing code.
   Other classes and predicates have been renamed, in these cases the old name is still available as a deprecated feature.
 
-* The basetype of the following list of classes has changed from an expression to a dataflow node, and and thus code using these classes might break. 
+* The basetype of the following list of classes has changed from an expression to a dataflow node, and thus code using these classes might break. 
   The fix to these breakages is usually to use `asExpr()` to get an expression from a dataflow node, or to use `.flow()` to get a dataflow node from an expression.  
    - DOM.qll#WebStorageWrite
    - CryptoLibraries.qll#CryptographicOperation
--- a/javascript/ql/lib/change-notes/2022-04-04-dataflow-models.md
+++ b/javascript/ql/lib/change-notes/2022-04-04-dataflow-models.md
@@ -1,0 +1,5 @@
+---
+category: minorAnalysis
+---
+* Many library models have been rewritten to use dataflow nodes instead of the AST.
+  These changes include some renamings where the old name is still available as a deprecated feature.
--- a/javascript/ql/lib/change-notes/2022-04-04-dataflow-models.md
+++ b/javascript/ql/lib/change-notes/2022-04-04-dataflow-models.md
@@ -1,5 +1,6 @@
 ---
-category: minorAnalysis
+category: breaking
 ---
 * Many library models have been rewritten to use dataflow nodes instead of the AST.
-  These changes include some renamings where the old name is still available as a deprecated feature.
+  The types of some classes have been changed, and these changes may break existing code.
+  Other classes and predicates have been renamed, in these cases the old name is still available as a deprecated feature.
--- a/javascript/ql/lib/change-notes/2022-04-04-dataflow-models.md
+++ b/javascript/ql/lib/change-notes/2022-04-04-dataflow-models.md
@@ -4,3 +4,54 @@ category: breaking
 * Many library models have been rewritten to use dataflow nodes instead of the AST.
   The types of some classes have been changed, and these changes may break existing code.
   Other classes and predicates have been renamed, in these cases the old name is still available as a deprecated feature.
+
+* The basetype of the following list of classes has changed from an expression to a dataflow node, and and thus code using these classes might break. 
+  The fix to these breakages is usually to use `asExpr()` to get an expression from a dataflow node, or to use `.flow()` to get a dataflow node from an expression.  
+   - DOM.qll#WebStorageWrite
+   - CryptoLibraries.qll#CryptographicOperation
+   - Express.qll#Express::RequestBodyAccess
+   - HTTP.qll#HTTP::ResponseBody
+   - HTTP.qll#HTTP::CookieDefinition
+   - HTTP.qll#HTTP::ServerDefinition
+   - HTTP.qll#HTTP::RouteSetup
+   - NoSQL.qll#NoSql::Query
+   - SQL.qll#SQL::SqlString
+   - SQL.qll#SQL::SqlSanitizer
+   - HTTP.qll#ResponseBody
+   - HTTP.qll#CookieDefinition
+   - HTTP.qll#ServerDefinition
+   - HTTP.qll#RouteSetup
+   - HTTP.qll#HTTP::RedirectInvocation
+   - HTTP.qll#RedirectInvocation
+   - Express.qll#Express::RouterDefinition
+   - AngularJSCore.qll#LinkFunction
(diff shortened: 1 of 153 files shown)


javascript/ql/lib/semmle/javascript/dataflow/TaintTracking.qll
```diff
@@ -580,16 +576,16 @@ module TaintTracking {
...
   private class ComputedPropWriteTaintStep extends SharedTaintStep {
     override predicate heapStep(DataFlow::Node pred, DataFlow::Node succ) {
-      exists(AssignExpr assgn, IndexExpr idx, DataFlow::SourceNode obj |
-        assgn.getTarget() = idx and
-        obj.flowsToExpr(idx.getBase()) and
-        not exists(idx.getPropertyName()) and
-        pred = DataFlow::valueNode(assgn.getRhs()) and
+      exists(DataFlow::PropWrite assgn, DataFlow::SourceNode obj |
+        not exists(assgn.getPropertyName()) and
+        not assgn.getWriteNode() instanceof Property and // not a write inside an object literal
```
erik-krogh, line 581: This line was added to preserve the behavior from the old predicate (new results appeared otherwise).  
I'm wondering if this line should be deleted?

annarailton:
Not qualified to sign off these changes to the `codeql-ml` code - will need to wait until @henrymercer is back on 6th April

javascript/ql/lib/change-notes/2022-04-04-dataflow-models.md
```diff
@@ -0,0 +1,5 @@
+---
+category: minorAnalysis
+---
+* Many library models have been rewritten to use dataflow nodes instead of the AST.
+  These changes include some renamings where the old name is still available as a deprecated feature.
```
asgerf, line 5: There are _some_ breaking changes, right? I think it's ok for a change like this, where it's practically unavoidable.

But if there are breaking changes, I think we should call out that some breaking changes have occurred, but in most cases, the AST-based predicates and classes are simply deprecated.

asgerf (commented):
What an incredible marathon of a rewrite! 👏 🎉 

I wasn't sure we'd ever get rid of this much legacy, but kudos for powering through here! 💪
I'm very happy to see this merged.

> Both lost results are a consequence of a DataFlow::PropWrite having no result for getRhs() when the AST looks like: `obj.prop += value`

I'm worried about losing `.innerHTML += x` as a sink, even temporarily.

Could you reduce the risk here, by making `DomSink` specifically add this back, perhaps via a helper predicate in `DomPropertyWrite`?

javascript/ql/lib/semmle/javascript/frameworks/Express.qll
```diff
@@ -901,9 +900,9 @@ module Express {
 
     /**
      * DEPRECATED: Use `ref().flowsToExpr()` instead.
-     * Holds if `sink` may refer to this router.
+     * This predicate is now from `DataFlow::SourceNode`.
      */
-    deprecated predicate flowsTo(Expr sink) { this.ref().flowsToExpr(sink) }
+    deprecated override predicate flowsTo(DataFlow::Node n) { super.flowsTo(n) }
```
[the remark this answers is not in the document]
erik-krogh: It can be removed, but if you do that then previous uses of `RouterDefinition::flowsTo` would suddenly point to `SourceNode::flowsTo` without a deprecation warning, which would be a major change in behavior.  

That gotcha happened to me during this rewrite, and it caused weird behavior changes that took me a while to debug.   
So I prefer keeping this deprecation.

javascript/ql/lib/change-notes/2022-04-04-dataflow-models.md
erik-krogh: > There are some breaking changes, right? 

Definitely yes, I hit a decent amount of compile-errors while doing this refactoring.   
I've changed the category to `breaking`, and reworded the change-note a bit.

erik-krogh:
> I'm worried about losing .innerHTML += x as a sink, even temporarily.
> Could you reduce the risk here, by making DomSink specifically add this back, perhaps via a helper predicate in DomPropertyWrite? 

:+1: 
`DomPropertyWrite` already defined it's own `getRhs()` predicate, so I just added another case there.  

---  

I think I've addressed all your review comments.

asgerf:
Thanks. Would you mind running a final evaluation? It's a large change and we should reduce the risk as much as possible.

henrymercer (dismissed):
ATM 👍

erik-krogh:
> Thanks. Would you mind running a final evaluation? It's a large change and we should reduce the risk as much as possible.

[A new evaluation](https://github.com/github/codeql-dca-main/tree/data/erik-krogh/httpNode-eval__default__code-scanning/reports) shows a small performance regression.  
But I haven't been able to nail down anything that's causing the regression.  
The two new results in the evaluation are both from destructuring assignment being seen as property reads.

erik-krogh:
I expanded the change-note to include which classes have changed type.  
I created it using a one-off QL-for-QL query (I had get both versions into the same DB).  

Here is the raw data from the query: https://docs.google.com/spreadsheets/d/1VkOGUofigegJYJ9un-qRczCBiEeNDEcTUsmFim90gRc/edit?usp=sharing

erik-krogh:
I ran the explicit-this patch to fix the alerts reported, and I fixed an instance of `ql/acronyms-should-be-pascal-case` in one of the new classes.   
So now there's no new QL-for-QL warnings from this PR.   

--- 

@asgerf I mentioned the classes with changed base-types in [the change-note](https://github.com/github/codeql/pull/8604/files#diff-e5cfab7a43a447ba2d6917edaff6e0488550c887dc0ddd29baa0d0306662ded1).  
I excluded deprecated classes and classes from Customizations files.  
So you think that works?

erik-krogh:
[New evaluation](https://github.com/github/codeql-dca-main/tree/data/erik-krogh/httpNode__nightly-old__code-scanning__4/reports) looks similar to the previous evaluation.   
There is still a slight performance regression in the same benchmarks, but still no clear cause.   
It might just be from there being more rows to process.   
I still think we can merge it.

calumgrant:
Merge after Universe release.

erik-krogh:
The JavaScript language tests are broken, but that's unrelated.  
I got an internal PR that fixes it (see the backref).

erik-krogh:
[A new evaluation looks good](https://github.com/github/codeql-dca-main/tree/data/erik-krogh/pr-8604-4e14177__nightly-old__code-scanning__1/reports).  

The new results are due to recognizing object destructuring as property reads.

asgerf (approved):
👍

PR #8609, opened 2022-03-30
C#: Operator flow

michaelnebel:
In this PR, we make
* Support for flow models for operators. This requires that we use the operator function name as the qualified name.
* Exclude conversion operators from model generation (there is always flow through conversion operators in library code - see TaintTrackingPrivate.ql).
* Exclude methods, where a namespace is not identified.


hvitved (approved):
LGTM, but we should do a DCA run, since this touches the core `hasQualifiedName` predicate.

michaelnebel:
There doesn't appear to be any query performance degradation during the DCA run (Powershell project is still failing to build, but this is a known issue). Merging now.

Issue #8570, opened 2022-03-26
Java: `SwitchCase.getRuleExpression()` has no result for `switch` statements


Marcono1234:
CodeQL's `SwitchCase.getRuleExpression()` seems to be intended to model `switch` rules (`... -> ...`) where the right side is `Expression;` ([JLS 17 §14.11.1](https://docs.oracle.com/javase/specs/jls/se17/html/jls-14.html#jls-14.11.1)). This works correctly for `switch` expressions, but for `switch` statements this CodeQL predicate does not have a result. Instead there the right side is modelled as `ExprStmt` in this case.

Example query:
```ql
import java

from SwitchStmt s, Top e
where
    e = s.getACase().getRuleExpression()
    or e = s.getACase().getRuleStatement()
select s, e, e.getPrimaryQlClasses()
```
ℹ️ `getRuleExpression()` has no results and instead `getRuleStatement()` has `ExprStmt` entities as result.

This behavior is inconsistent and possibly also incorrect. [JLS 17 §14.11.2](https://docs.oracle.com/javase/specs/jls/se17/html/jls-14.html#jls-14.11.2) explicitly covers this case, saying:
> Every switch rule expression in the switch block is a statement expression ([§14.8](https://docs.oracle.com/javase/specs/jls/se17/html/jls-14.html#jls-14.8))

Notice that it says "statement *expression*". While this sounds similar to "expression *statement*" (as which CodeQL currently models it), they are separate concepts; a statement *expression* is contained within an expression *statement*.

If this behavior of CodeQL is considered intended, it might be good to mention this in the documentation of `getRuleExpression()` to avoid incorrect behavior when users assume `getRuleExpression()` will have results and `getRuleStatement()` will only have block statements or `throw` statements as result.

smowton:
How about https://github.com/github/codeql/pull/8611

I think intuitively the idea that the RHS is either directly an expression or a block containing yield statements is likely to make sense to users, even if the JLS contains text about statement expressions.

PR #8611, opened 2022-03-30
Java: make SwitchCase.getRuleExpression/Statement more consistent

smowton:
Unpack `ExprStmt`s in `getRuleExpression` such that arrow switch rules look the same whether they're used in the context of a switch expression or of a switch statement. Also improve docs.

### DIFF
--- a/java/ql/lib/semmle/code/java/Statement.qll
+++ b/java/ql/lib/semmle/code/java/Statement.qll
@@ -440,10 +440,10 @@ class SwitchCase extends Stmt, @case {
    * Gets the expression on the right-hand side of the arrow, if any.
    *
    * Note, this predicate gets a value when this switch case is of the form
-   * `case e1 -> e2`, where `e2` is neither a block nor a throw statement. 
+   * `case e1 -> e2`, where `e2` is neither a block nor a throw statement.
    * This predicate is mutually exclusive with `getRuleStatement`.
    */
-  Expr getRuleExpression() { 
+  Expr getRuleExpression() {
     result.getParent() = this and result.getIndex() = -1
     or
     exists(ExprStmt es | es.getParent() = this and es.getIndex() = -1 | result = es.getExpr())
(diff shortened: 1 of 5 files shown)


java/ql/lib/semmle/code/java/Statement.qll
```diff
@@ -438,11 +438,17 @@ class SwitchCase extends Stmt, @case {
 
   /**
    * Gets the expression on the right-hand side of the arrow, if any.
+   *
+   * Note this is mutually exclusive with `getRuleStatement`: it gets a value
+   * when this case is of the form `case e1 -> e2`, where `e2` is not a block.
```
intrigus-lgtm, line 443:
```suggestion
   * Note, this predicate gets a value when this switch case is of the form
   * `case e1 -> e2`, where `e2` is not a block. This predicate is mutually
   * exclusive with `getRuleStatement`.
```

java/ql/lib/semmle/code/java/Statement.qll
```diff
@@ -438,11 +438,17 @@ class SwitchCase extends Stmt, @case {
...
    */
   Expr getRuleExpression() { result.getParent() = this and result.getIndex() = -1 }
 
   /**
    * Gets the statement on the right-hand side of the arrow, if any.
+   *
+   * Note this is mutually exclusive with `getRuleExpression`: it gets a value
+   * when this case if of the form `case e1 -> { s1; s2; ... }`.
```
intrigus-lgtm, line 451:
```suggestion
   * when this case is of the form `case e1 -> { s1; s2; ... }`.
```

intrigus-lgtm, line 451:
```suggestion
   * Note, this predicate gets a value when this switch case is of the form
   * `case e1 -> { s1; s2; ... }`. This predicate is mutually exclusive with `getRuleExpression`.
```

intrigus-lgtm (commented):
I find my suggestions slightly easier to understand, but feel to ignore them besides the "if" -> "is" typo suggestion.

smowton:
OK, I had not realised until now you were concerned about the cases

```
// Expression context
int x = switch(y) { case 1 -> f(); default -> g(); }
```
and
```
// Statement context
switch(y) { case 1 -> f(); default -> g(); }
```

getting different representations. This indeed seems unintuitive.

smowton:
@Marcono1234 is this more like what you meant

Marcono1234:
Yes that is what I had in mind, thanks a lot! That seems to resolve #8570, could you please link them so GitHub automatically closes it?
Does this need a change note now, since it changes the behavior of the predicates?

smowton:
I suppose so -- added.

aschackmull (approved):
LGTM.

PR #8615, opened 2022-03-30
Clarify potential uses for `implies`

JLLeitschuh:
Clarify how `if .. then .. else` can be equivalent to `implies` in certain scenarios.

### DIFF
--- a/docs/codeql/ql-language-reference/formulas.rst
+++ b/docs/codeql/ql-language-reference/formulas.rst
@@ -331,7 +331,7 @@ The following query selects files that are not HTML files.
 
 You can use these keywords to write a conditional formula. This is another way to simplify
 notation: ``if A then B else C`` is the same as writing ``(A and B) or ((not A) and C)``.
-``if A then B else any()`` can also be represented as `A implies B`.
+``if A then B else any()`` can also be represented as ``A implies B``.
 
 **Example**
 
--- a/docs/codeql/ql-language-reference/formulas.rst
+++ b/docs/codeql/ql-language-reference/formulas.rst
@@ -331,6 +331,7 @@ The following query selects files that are not HTML files.
 
 You can use these keywords to write a conditional formula. This is another way to simplify
 notation: ``if A then B else C`` is the same as writing ``(A and B) or ((not A) and C)``.
+``if A then B else any()`` can also be represented as `A implies B`.
 
 **Example**
 
@@ -399,7 +400,10 @@ With the following definition, an integer is in the class ``OneTwoThree`` if it 
 ===========
 
 You can use the keyword ``implies`` between two formulas. The resulting formula is called an 
-implication. This is just a simplified notation: ``A implies B`` is the same as writing ``(not A) or B``.
+implication. This is just a simplified notation:
+
+* ``A implies B`` is the same as writing ``(not A) or B``.
+* ``A implies B`` is the same as writing ``if A then B else any()``.
 
 **Example**
 


docs/codeql/ql-language-reference/formulas.rst
```diff
@@ -399,7 +400,10 @@ With the following definition, an integer is in the class ``OneTwoThree`` if it
 ===========
 
 You can use the keyword ``implies`` between two formulas. The resulting formula is called an 
-implication. This is just a simplified notation: ``A implies B`` is the same as writing ``(not A) or B``.
+implication. This is just a simplified notation:
+
+* ``A implies B`` is the same as writing ``(not A) or B``.
```
p0, line 405: This is exactly right -- it is desugared in this way internally. 👍

docs/codeql/ql-language-reference/formulas.rst
```diff
@@ -331,6 +331,7 @@ The following query selects files that are not HTML files.
 
 You can use these keywords to write a conditional formula. This is another way to simplify
 notation: ``if A then B else C`` is the same as writing ``(A and B) or ((not A) and C)``.
+``if A then B else any()`` can also be represented as `A implies B`.
```
owen-mc, line 334:
```suggestion
``if A then B else any()`` can also be represented as ``A implies B``.
```

docs/codeql/ql-language-reference/formulas.rst
```diff
@@ -399,7 +400,10 @@ With the following definition, an integer is in the class ``OneTwoThree`` if it
 ===========
 
 You can use the keyword ``implies`` between two formulas. The resulting formula is called an 
-implication. This is just a simplified notation: ``A implies B`` is the same as writing ``(not A) or B``.
+implication. This is just a simplified notation:
+
+* ``A implies B`` is the same as writing ``(not A) or B``.
+* ``A implies B`` is the same as writing ``if A then B else any()``.
```
p0, line 406: This is not quite right. It means the same, but is not operationally equivalent.

The desugaring for `if A then B else C` is in fact `(A and B) or ((not A) and C)`. Thus, if `C` is `any()`, the if/then/else desugars to `(A and B) or (not A)`. We can distribute the terms to see this is equivalent to `(A or (not A)) and (B or (not A))`, which simplifies to `(not A) or B`, so logically they are the same.

Practically, however, it may be that `(A and B)` is much faster to evaluate than just `B` -- perhaps `A` has very few tuples, while `B` has very many, and so it's faster to just check whether `B` holds of the tuples in `A` rather than enumerating all the tuples in `B`. Or, in other words, writing `if A then B else any()` evaluates the `then` part only for contexts that satisfy the condition, while this constraint does not exist in `A implies B`.

[You might argue that we should therefore desugar `A implies B` to `(not A) or (A and B)`, but that can be worse if `A` in fact does not significantly restrict the values in `B`, and therefore it would be a potentially breaking change. Still, it might be less surprising. 🤔]

```suggestion
* ``A implies B`` means the same as writing ``if A then B else any()``, though the latter is in fact evaluated as ``A implies (A and B)``.
```

docs/codeql/ql-language-reference/formulas.rst
p0, line 334:
```suggestion
``if A then B else any()`` can also be represented as ``A implies (A and B)``, and it means the same as ``A implies B``.
```

(See my other comment for explanation.)

docs/codeql/ql-language-reference/formulas.rst
p0: ... I'm actually not sure whether we want this kind of operational detail in the language reference -- how do others feel?

jbj (commented):
We sometimes rewrite `implies` to the `if` form manually for performance reasons. That trick is probably worth mentioning somewhere, but I'm not sure the QL language reference is the ideal place. If we should mention it here, it should be clearly marked as an optimisation trick that may be manually applied.

JLLeitschuh:
There seems to be a lot of debate on my PR about my update to the explanation around ``` ``implies`` ```.

It doesn't seem clear to me what the "best" way to express `if .. then .. else any()` in codeQL. I'm just looking for an easy way to communicate: if 'predicate holds' then 'this predicate must also hold' else 'give me everything'. I want to communicate this in CodeQL in a  way that is easy to understand and interpret.

Given this, what would you like to see the documentation changed to?

aschackmull:
> It doesn't seem clear to me what the "best" way to express `if .. then .. else any()` in codeQL. I'm just looking for an easy way to communicate: if 'predicate holds' then 'this predicate must also hold' else 'give me everything'. I want to communicate this in CodeQL in a way that is easy to understand and interpret.

The `if .. then .. else any()` matches up directly with your sentence "if 'predicate holds' then 'this predicate must also hold' else 'give me everything'", so if that's what you want to communicate, then that piece of QL seems fine. Anecdotally, I also think this tends to perform a bit better than `implies` in most cases (although this isn't clear-cut).

> Given this, what would you like to see the documentation changed to?

I don't see the need to change the documentation here - both `if .. then .. else ..` and `implies` seem to be adequately explained, and I don't think a discussion of `if A then B else any()` vs. `A implies B` belongs here. It's true that it is an interesting comparison with some subtlety as discussed in the various comments above, but it is also fairly corner-case, and I'm not sure it needs explicit mention outside of, say perhaps, as an example in some highly technical performance debugging deep-dive.

I suggest just closing this PR.

aschackmull:
> Maybe this is better documented under `if .. then .. else ..`?

I'm not sure what else you'd expect in the documentation? `if A then B else C` seems perfectly adequately documented, so I don't know why we'd want to special-case `C = any()`.

owen-mc:
I agree with @JLLeitschuh . `any()` is not something that a relatively new user would be familiar with, and this is one of the first times they might want to use it. Could we add a line to the end of the if-then-else section like:
> Note that when no restriction is required in the else case, the [built-in predicate](https://codeql.github.com/docs/ql-language-reference/ql-language-specification/#non-member-built-ins) `any()` may be used: `if A then B else any()`?

aschackmull:
Right, I see what's missing now, we're not actually documenting the "true" (`any()`) and "false" (`none()`) in the list of logical connectives. In fact, the only mention of `none()` is in an example that's blatantly wrong (it's not valid QL). PR with a suggested fix: https://github.com/github/codeql/pull/8701

aschackmull:
If we still want to mention something like `if A then B else any()` then I think it fits better as an example in the section on `any()`.

PR #8618, opened 2022-03-30
Update existing qhelp comment, if it exists

hmac:
Update the Ruby QHelp preview workflow so that if there's already a comment on the PR with the QHelp preview, it will update that comment instead of creating a new one. The old versions are still visible in the comment history.

This makes use of https://github.com/peter-evans/create-or-update-comment and https://github.com/peter-evans/find-comment, both of which are mentioned in the GH Actions docs so I feel they're relatively legitimate.

I've tested this on [my fork](https://github.com/hmac/codeql/pull/1) and it seems to work correctly.

I've also renamed some things and added some comments to try to explain the split workflow design (as I understand it).

@aibaars I believe you wrote the existing workflow - how do you feel about this change?


.github/workflows/add-pr-qhelp-comment.yml
```diff
@@ -0,0 +1,73 @@
...
+          body="${body//$'\n'/'%0A'}"
+          body="${body//$'\r'/'%0D'}"
+          echo "::set-output name=body::$body"
+
+      - id: get-pr-number
+        run: |
+          pr="$(cat pr.txt)"
+          echo "::set-output name=pr-number::$pr"
+
+      - name: Find existing comment, if it exists
```
aibaars, line 56: Is it possible to run this code as part of the unprivileged workflow?

.github/workflows/add-pr-qhelp-comment.yml
```diff
@@ -0,0 +1,73 @@
...
+          # Escape newlines - sadly required for some reason:
+          # https://github.community/t/set-output-truncates-multiline-strings/16852/3
+          body="${body//'%'/'%25'}"
+          body="${body//$'\n'/'%0A'}"
+          body="${body//$'\r'/'%0D'}"
+          echo "::set-output name=body::$body"
+
+      - id: get-pr-number
+        run: |
+          pr="$(cat pr.txt)"
```
aibaars, line 53: Above we use `PR="$(grep -o '^[0-9]\+$' pr.txt)"` which performs a bit of input validation.

.github/workflows/add-pr-qhelp-comment.yml
```diff
@@ -0,0 +1,73 @@
...
+        uses: peter-evans/find-comment@v2
+        id: find-comment
+        with:
+          issue-number: ${{ steps.get-pr-number.outputs.pr-number }}
+          comment-author: 'github-actions[bot]'
+          body-includes: QHelp previews
+          direction: last
+          token: ${{ github.token }}
+
+      - name: Create or update comment
```
aibaars, line 66: It's probably best to replace this Action with a call to `gh api`.
* https://docs.github.com/en/rest/reference/issues#create-an-issue-comment
* https://docs.github.com/en/rest/reference/issues#update-an-issue-comment

aibaars (changes requested):
We better keep things as simple as possible to reduce possible security issues and make security review easier.

hmac:
I've updated the `on: pull_request` workflow so it uploads the following artifacts:
- `pr_number.txt` - the number of the PR
- `comment_body.txt` - the comment text containing the rendered QHelp
- `comment_id.txt` - the ID of the existing QHelp comment, if it exists. This file is empty otherwise.

The `on: workflow_run` workflow then does the following:
1. Download the artifacts.
2. Check the PR SHA matches the workflow SHA.
3. If `comment_id.txt` contains a comment ID, update that comment with the contents of `comment_body.txt`.
4. Otherwise, create a new comment with the contents of `comment_body.txt`.

I think the result is even a bit simpler than the previous version because we don't have to jump through hoops to get the data into the third-party actions anymore.

.github/workflows/qhelp-pr-add-comment.yml
```diff
@@ -0,0 +1,57 @@
+# This workflow is the second part of the process described in
```
aibaars, line 1: Let's not rename this file. It can actually be used by other (future) jobs that need to post PR comments.

.github/workflows/qhelp-pr-add-comment.yml
```diff
@@ -0,0 +1,57 @@
...
+            COMMENT_ID="$(grep -o '^[0-9]\+$' comment_id.txt)"
+          fi
+
+          if [ -z "$COMMENT_ID" ]
+          then
+            # Create new comment
+            jq --rawfile body comment_body.txt '{"body":$body}' -n | gh api "repos/${{ github.repository }}/issues/${PR_NUMBER}/comments" -X POST --input -
+          else
+            # Update existing comment
+            jq --rawfile body comment_body.txt '{"body":$body}' -n | gh api "repos/${{ github.repository }}/issues/comments/${COMMENT_ID}" -X PATCH --input -
```
aibaars, line 54: It's unlikely that `${{github.repository}}` contains dangerous content like `"` or `$(...)`  that could cause command injection. However, better safe than sorry. Using a normal environment variable should be a little safer .

```suggestion
            jq --rawfile body comment_body.txt '{"body":$body}' -n | gh api "repos/${GITHUB_REPOSITORY}/issues/${PR_NUMBER}/comments" -X POST --input -
          else
            # Update existing comment
            jq --rawfile body comment_body.txt '{"body":$body}' -n | gh api "repos/${GITHUB_REPOSITORY}/issues/comments/${COMMENT_ID}" -X PATCH --input -
```

.github/workflows/qhelp-pr-preview.yml
```diff
@@ -52,12 +71,26 @@ jobs:
...
           name: comment
-          path: comment.txt
+          path: comment_body.txt
+          retention-days: 1
+
+      - name: Save ID of existing QHelp comment (if it exists)
+        run: |
+          # Find the latest comment starting with "QHelp previews"
+          COMMENT_PREFIX="QHelp previews"
+          gh api "repos/${{ github.repository }}/issues/${{ github.event.number }}/comments?per_page=100" --jq "[.[] | select(.body|startswith(\"${COMMENT_PREFIX}\")) | .id] | max" > comment_id.txt
```
aibaars, line 88: Perhaps we should use the `--paginate` flag of `gh api` to fetch all pages of comments so we're sure to get the latest even if there are over 100 comments.

aibaars (commented):
Looks good to me. Could you undo the file rename? I think there is no need to rename the file, and it would make it easier to review the differences.

@adityasharad could you have a look at this too from a security point of view.

.github/workflows/post-pr-comment.yml
```diff
@@ -17,15 +22,36 @@ jobs:
...
+          fi
+
+          if [ -z "$COMMENT_ID" ]
+          then
+            # Create new comment
+            jq --rawfile body comment_body.txt '{"body":$body}' -n | gh api "repos/${GITHUB_REPOSITORY}/issues/${PR_NUMBER}/comments" -X POST --input -
+          else
+            # Update existing comment
+            jq --rawfile body comment_body.txt '{"body":$body}' -n | gh api "repos/${GITHUB_REPOSITORY}/issues/comments/${COMMENT_ID}" -X PATCH --input -
+          fi
```
aibaars, line 55: There is no need for a double `if`. The case where `comment_id.txt` exists but does not contain a number is an error condition that should never happen. I think the workflow wouldn't even fall through to the `-z "${COMMENT_ID"` check because `grep` fails with a non-zero exit code if the pattern isn't matched.

```suggestion
          if [ -s comment_id.txt ]
          then
            COMMENT_ID="$(grep -o '^[0-9]\+$' comment_id.txt)"
            # Update existing comment
            jq --rawfile body comment_body.txt '{"body":$body}' -n | gh api "repos/${GITHUB_REPOSITORY}/issues/comments/${COMMENT_ID}" -X PATCH --input -
          else
            # Create new comment
            jq --rawfile body comment_body.txt '{"body":$body}' -n | gh api "repos/${GITHUB_REPOSITORY}/issues/${PR_NUMBER}/comments" -X POST --input -
          fi
```

.github/workflows/post-pr-comment.yml
```diff
@@ -1,12 +1,17 @@
...
 on:
   workflow_run:
-    workflows: ["Query help preview"]
+    workflows: [Render QHelp changes]
     types:
       - completed
 
 permissions:
   pull-requests: write
+  actions: read
```
aibaars, line 14: The permission looks reasonable, I just wonder why it is needed, we're making the same API call as before, right? Or is this just to be a bit more explicit( which I think is a good thing).?

.github/workflows/qhelp-pr-preview.yml
```diff
@@ -1,7 +1,25 @@
-name: Query help preview
+# This workflow checks for any changes in .qhelp files in pull requests.
+# For any changed files, it renders them to HTML in a file called `comment_body.txt`.
```
aibaars, line 2:
```suggestion
# For any changed files, it renders them to markdown in a file called `comment_body.txt`.
```

.github/workflows/qhelp-pr-preview.yml
```diff
@@ -52,12 +71,26 @@ jobs:
...
+          # Find the latest comment starting with "QHelp previews"
+          COMMENT_PREFIX="QHelp previews"
+          gh api "repos/${{ github.repository }}/issues/${{ github.event.number }}/comments" --paginate --jq "[.[] | select(.body|startswith(\"${COMMENT_PREFIX}\")) | .id] | max" > comment_id.txt
+        env:
+          GITHUB_TOKEN: ${{ github.token }}
+
+      - uses: actions/upload-artifact@v2
+        with:
+          name: comment
+          path: comment_id.txt
```
aibaars, line 95: Shall we add a `if-no-files-found:` field to document that it expected for the `comment_id.txt` to be missing? Perhaps we should add `if-no-files-found: error` to the upload steps for comment_body and pr_number.
```suggestion
          path: comment_id.txt
          if-no-files-found: ignore
```

aibaars (commented):
This looks really good. @adityasharad and @esbena Could you have a quick look at this too to check for potential security issues?

.github/workflows/post-pr-comment.yml
hmac: Unfortunately, the `jq` expression to select the ID of the most recent matching comment (`[ ... ] | max`) will return `null` if there are no matching comments, and `gh` will for some reason render that as an empty line. This means we end up with a `comment_id.txt` containing just a newline, which bash considers a non-empty file (i.e. `-s comment_id.txt == true`).

The approach I've gone with is to catch the grep failure and then use `[ $COMMENT_ID ]` as the test.

.github/workflows/post-pr-comment.yml
hmac: In my testing, I needed this in order for `gh run download` to not throw `HTTP 403: Resource not accessible by integration`. This may be due to a configuration difference between my test repo and `github/codeql`, though. I guess it's good to be explicit regardless.

.github/workflows/qhelp-pr-preview.yml
hmac: Actually due to the issue I mentioned [here](https://github.com/github/codeql/pull/8618#discussion_r842273725), we should always upload a `comment_id.txt` file regardless, so I've made them all `if-no-files-found: error`.

.github/workflows/post-pr-comment.yml
```diff
@@ -17,15 +22,52 @@ jobs:
...
+                                      | select(.body | test(\"^${COMMENT_PREFIX}\"))         \
+                                      | select(.user.login == \"${COMMENT_AUTHOR}\")         \
+                                      | .id"
+            COMMENT_ID=$(gh api "repos/${GITHUB_REPOSITORY}/issues/comments/${COMMENT_ID}" | jq "${FILTER}")
+            if [ $COMMENT_ID ]
+            then
+              # Update existing comment
+              jq --rawfile body comment_body.txt '{"body":$body}' -n | gh api "repos/${GITHUB_REPOSITORY}/issues/comments/${COMMENT_ID}" -X PATCH --input -
+            else
+              echo "Comment ${COMMENT_ID} did not pass validations: not editing."
```
aibaars, line 66: `${COMMENT_ID}` will be empty so printing it is not useful. Let's print the error message to stderr, and exit with a non-zero exit code to make the workflow fail. 
```suggestion
              >&2 echo "error: comment ID did not pass validations: not editing." 
              exit 1
```

.github/workflows/post-pr-comment.yml
```diff
@@ -17,15 +22,53 @@ jobs:
...
+                  | select(.body | startswith($prefix))
+                  | select(.user.login == $author)
+                  | .id'
+            COMMENT_ID=$(gh api "repos/${GITHUB_REPOSITORY}/issues/comments/${COMMENT_ID}" | jq --arg repo "${GITHUB_REPOSITORY}" --arg pr "${PR_NUMBER}" --arg prefix "${COMMENT_PREFIX}" --arg author "${COMMENT_AUTHOR}" "${FILTER}")
+            if [ $COMMENT_ID ]
+            then
+              # Update existing comment
+              jq --rawfile body comment_body.txt '{"body":$body}' -n | gh api "repos/${GITHUB_REPOSITORY}/issues/comments/${COMMENT_ID}" -X PATCH --input -
+            else
+              echo "Comment ${RAW_COMMENT_ID} did not pass validations: not editing." >&2
```
aibaars, line 66: Let's not print the comment ID at all if it is invalid. Printing arbitrary output to the log may be risky. I'm not sure whether Actions picks up the special log-commands from stderr, but better safe than sorry.

hmac: It's guaranteed to be just a sequence of digits though...

aibaars: Ah yes indeed. I somehow thought it was the raw content from the file

aibaars (dismissed):

.github/workflows/post-pr-comment.yml
```diff
@@ -17,15 +22,53 @@ jobs:
...
+          then
+            # Fetch existing comment, and validate:
+            # - comment belongs to the PR with number $PR_NUMBER
+            # - comment starts with the expected prefix ("QHelp previews")
+            # - comment author is github-actions[bot]
+            FILTER='select(.issue_url | endswith($repo+"/issues/"+$pr))
+                  | select(.body | startswith($prefix))
+                  | select(.user.login == $author)
+                  | .id'
+            COMMENT_ID=$(gh api "repos/${GITHUB_REPOSITORY}/issues/comments/${COMMENT_ID}" | jq --arg repo "${GITHUB_REPOSITORY}" --arg pr "${PR_NUMBER}" --arg prefix "${COMMENT_PREFIX}" --arg author "${COMMENT_AUTHOR}" "${FILTER}")
```
esbena, line 60:
```suggestion
            COMMENT_ID=$(gh api "repos/${GITHUB_REPOSITORY}/issues/comments/${RAW_COMMENT_ID}" | jq --arg repo "${GITHUB_REPOSITORY}" --arg pr "${PR_NUMBER}" --arg prefix "${COMMENT_PREFIX}" --arg author "${COMMENT_AUTHOR}" "${FILTER}")
```
typo?

.github/workflows/qhelp-pr-preview.yml
```diff
@@ -52,12 +74,29 @@ jobs:
...
-          path: comment.txt
+          path: comment_body.txt
+          if-no-files-found: error
+          retention-days: 1
+
+      - name: Save ID of existing QHelp comment (if it exists)
+        run: |
+          # Find the latest comment starting with "QHelp previews"
+          COMMENT_PREFIX="QHelp previews"
+          gh api "repos/${GITHUB_REPOSITORY}/issues/${PR_NUMBER}/comments" --paginate --jq "[.[] | select(.body|startswith(\"${COMMENT_PREFIX}\")) | .id] | max" > comment_id.txt
```
esbena, line 92: (nit that we can merge without addressing: `jq ... --arg ...` should be used instead of  environment variable interpolation. )

esbena (approved):
LGTM now. Thanks for the perseverance!

aibaars (approved):

PR #8619, opened 2022-03-31
JS-ML: fix isKnownStepSrc such that it recognizes taint-steps

erik-krogh:
`isKnownStepSrc` was using the deprecated `TaintTracking::AdditionalTaintStep` and `DataFlow::AdditionalFlowStep` classes, but those classes doesn't have any taint-steps these days.    
I updated `isKnownStepSrc` to not use those deprecated classes, and now it recognizes way more taint-steps.  

As a result there are now way less endpoints with features. 
I suppose that is from filtering out endpoints that have an outgoing step. 

/cc: @esbena @kaeluka


henrymercer (approved):
Probably makes most sense for @esbena or @kaeluka to take a look at this too, but LGTM.

esbena:
LGTM too, we definitely want this bugfix. If this was for plain codeql, it should just be merged, but you may want to run it through your usual e2e tests first for a sanity check.

henrymercer:
End-to-end tests are looking positive.  @erik-krogh please feel free to go ahead and merge.

PR #8612, opened 2022-03-30
C++: Fix two performance issues

MathiasVP:
Two unrelated performance fixes:

- The first commit fixes a case of bad magic:
```ql
Tuple counts for m#Overflow::guardedLesser#a1017054#bb/2@5d4bacm0 after 16.8s:
  253512    ~0%     {2} r1 = JOIN Expr::Operation::getAnOperand#dispred#f0820431#ff_10#join_rhs WITH Access::Access::getTarget#dispred#f0820431#ff ON FIRST 1 OUTPUT Rhs.1, Lhs.1 'e'
  136449963 ~1%     {2} r2 = JOIN r1 WITH Overflow::varUse#a1017054#ff ON FIRST 1 OUTPUT Lhs.1 'e', Rhs.1 'use'
  return r2
```

- The second commit fixes a bad join order in an antijoin. Here are the tuple counts on PHP:

Before:
```ql
StackVariableReachability::StackVariableReachability::reaches#f0820431#ffff#antijoin_rhs#1/7@ebd0f4gr after 11s:
    14077    ~3%        {7} r1 = JOIN StackVariableReachability::StackVariableReachability::reaches#f0820431#ffff#shared#2 WITH BasicBlocks::Cached::basic_block_member#8748103a#fff ON FIRST 2 OUTPUT Lhs.5 'arg4', Lhs.2 'arg0', Lhs.3 'arg1', Lhs.1 'arg2', Lhs.4 'arg3', Lhs.0 'arg5', Rhs.2 'arg6'
    14077    ~0%        {9} r2 = JOIN r1 WITH UninitializedLocal::UninitialisedLocalReachability#class#9d706ecd#f ON FIRST 1 OUTPUT Lhs.1 'arg0', Lhs.2 'arg1', Lhs.3 'arg2', Lhs.4 'arg3', Lhs.0 'arg4', Lhs.5 'arg5', Lhs.6 'arg6', (Lhs.4 'arg3' + 1), (Lhs.6 'arg6' - 1)
    69840066 ~0%        {10} r3 = JOIN r2 WITH PRIMITIVE range#bbf ON Lhs.7,Lhs.8
    69840066 ~1%        {8} r4 = SCAN r3 OUTPUT In.2 'arg2', In.9, In.0 'arg0', In.1 'arg1', In.3 'arg3', In.4 'arg4', In.5 'arg5', In.6 'arg6'
[...]

PR #8622, opened 2022-03-31
C++: Fix bad magic and bad join

MathiasVP:
Same as https://github.com/github/codeql/pull/8612. It was causing weird CI failures (maybe due to the PR being retargeted), so I'll see if a fresh PR fixes it 🤞.


jketema (approved):
LGTM

PR #8623, opened 2022-03-31
C++: Some enhancements to SensitiveExprs.qll

geoffw0:
While working on https://github.com/github/codeql/pull/8580, I noticed some extra expressions that `SensitiveExprs.qll` ought to catch.  This PR adds these, and updates the library from using `match` to `regexpMatch` for similar reasons as in that PR.


geoffw0:
Differences in flagged sensitive expressions on some LGTM projects: https://lgtm.com/query/7804715574741333977/ (these will not necessarily lead to query results, depending what is done with the data).

MathiasVP (approved):

PR #8628, opened 2022-03-31
C#: Introduce generated flag as a part of the kind column for flow summaries

michaelnebel:
In this PR, we change the semantics slightly for the "kind" column in flow models.
* A *kind* (eg. `value` or `taint`) can now be prefixed with `generated:`.
* In case a summary kind is prefixed with `generated`, it is only used in case no summary exist that is not generated.


hvitved (approved):
Looks great!

aschackmull:
Thoughts on making this its own column?

michaelnebel:
> Thoughts on making this its own column?

We discussed it, but decided to go with this solutions as it is backwards compatible.
What do you think. Should we reconsider?

aschackmull:
I don't think backwards-compatibility is relevant yet. I think I'm leaning slightly towards making it its own column. @hvitved ?

hvitved:
> I don't think backwards-compatibility is relevant yet. I think I'm leaning slightly towards making it its own column. @hvitved ?

I don't really care; I am OK with adding another column for this.

michaelnebel:
So should we add an extra column after the kind?

PR #8630, opened 2022-03-31
Bazel: add skeleton bazel workspace

redsun82:
This moves in codeql some internal bazel definitions in preparation for
future work.


bazel/README.md
```diff
@@ -0,0 +1,2 @@
+Please notice that any bazel targets and definitions in this repository are currently experimental and for internal use
```
criemen, line 1: Can we move this to the top of the workspace?
I think it would be much more discoverable there.

criemen (commented):
One small comment, otherwise LGTM.

If we do go this direction, we should probably add a CODEOWNERS entry for the relevant parts of the build, though.

bazel/README.md
redsun82: I put it both in `CONTRIBUTING.md` and as a comment to `WORKSPACE.bazel` :+1:

criemen (approved):
LGTM

MathiasVP (approved):

PR #8634, opened 2022-03-31
Python: Promote XXE and XML-bomb queries

RasmusWL:
This was a draft PR to ensure we don't merge it before we have agreed with JS/Ruby about the new `XmlParsing` concept, but they have said good for it, so we're ready to go :+1:

With this PR, I have copied the query setup from JS (Ruby only has XXE query). This means that no query will alert on 'DTD retrieval', which I will postpone for future work (this PR is already quite big).


RasmusWL:
I tried to see whether `API::CallNode` could be used to merge the `xml.etree.ElementTree` modeling, but I couldn't find a nice way to do this, so I left them in this slightly ugly state where they sort of share logic, but doesn't actually share the code.

RasmusWL:
Performance evaluation looks good :+1:

python/ql/lib/change-notes/2022-03-29-add-taint-for-StringIO.md
```diff
@@ -0,0 +1,4 @@
+---
+category: minorAnalysis
+---
+* Added taint propagation for `io.StringIO` and `io.BytesIO`. This addition was originally [submitted as part of an experimental query by @jorgectf](https://github.com/github/codeql/pull/6112).
```
yoff, line 4: Should the lxml parser modelling appear here?

yoff (changes requested):
Good thorough job. Only minor comments. It looks like you deleted a  poc (`this_directory_not_extracted`)?

RasmusWL:
> It looks like you deleted a poc (this_directory_not_extracted)?

Since the old PoC was located with the framework testing code, and the framework testing code was moved to multiple folders, I decided to create a new top-level directory to contain that code: https://github.com/github/codeql/blob/714465bf39d97e31aa6f0a7aa01c57e16f3c3078/python/PoCs/XmlParsing/PoC.py -- I guess I should have mentioned this in a comment :flushed: I'm not sure it's the best solution ever, since this means that the framework testing code is far removed from this PoC, but I also couldn't come up with a better solution :thinking:

python/ql/lib/change-notes/2022-03-29-add-taint-for-StringIO.md
RasmusWL: > In fact there is quite a bit of library modelling added..
True. 

I did write a change-note for the new query itself:
https://github.com/github/codeql/blob/714465bf39d97e31aa6f0a7aa01c57e16f3c3078/python/ql/src/change-notes/2022-04-05-add-xxe-and-xmlbomb.md

My thoughts were that a new query about XML libraries will by default also include modeling of those XML libraries.

yoff: Ah, I think that is fair.

python/ql/lib/semmle/python/frameworks/Stdlib.qll
```diff
@@ -3171,6 +3107,554 @@ private module StdlibPrivate {
...
+              .getMember("ElementTree")
+              .getMember("XMLParser")
+              .getACall()
+        or
+        this =
+          API::moduleImport("xml")
+              .getMember("etree")
+              .getMember("ElementTree")
+              .getMember("XMLPullParser")
+              .getACall()
```
yoff, line 3276:
```suggestion
        this =
          API::moduleImport("xml")
              .getMember("etree")
              .getMember("ElementTree")
              .getMember(["XMLParser", "XMLPullParser"])
              .getACall()
```

yoff (changes requested):
I do not want to drag this out over little refactoring, but since you now did the one for Lxml (that I recall considering), I want to suggest the only other one, I considered. (I think the two branches of `elementTreeInstance` are fine to keep apart, since they have different semantics).

yoff (approved):
LGTM

PR #8637, opened 2022-04-01
JS: fix bad join in XmlParsers.qll

erik-krogh:
Some very large benchmarks failed to terminate due to this.  
(For some reason only when running `js/prototype-pollution-utility`?)


erik-krogh:
I couldn't get anything good out of this.  
I still got some timeouts, but less when I used an `XL` worker.  
It's `js/prototype-pollution-utility` timing out, which is just due to there being a massive number of sources/sinks.

PR #8639, opened 2022-04-01
Java: Improve Spring models

atorralba:
Adds models for the constructor of `MutablePropertyValues` of `spring-beans`. Also fixes a mistake in a couple of classes of `spring-web`.


smowton:
This suggests that

```
  /** Holds if the parameter should not be consider a direct source of taint. */
  predicate isNotDirectlyTaintedInput() {
    this.getType().(RefType).getAnAncestor() instanceof SpringWebRequest or
    this.getType().(RefType).getAnAncestor() instanceof SpringNativeWebRequest or
```
should not have worked before-- add a test?

atorralba:
@smowton good point, thanks. Added in 348b932.

smowton (dismissed):

smowton (approved):

PR #8641, opened 2022-04-01
Data flow: Introduce `ContentSet`

hvitved:
### Background

In Ruby, we are tracking flow through arrays with precise array index information, when available. That is, we are correctly able to handle

```rb
a[0] = tainted;
a[1] = not_tainted;
sink(a[0]); # bad
sink(a[1]); # good
```

Flow is tracked using the data flow library's notion of [contents, reads, and stores](https://github.com/github/codeql/blob/main/docs/ql-libraries/dataflow/dataflow.md#field-flow), and we can model known/unknown array information as follows

```ql
newtype Content =
  KnownArrayElementContent(int i) { i in [0 .. 10] } or
  UnknownArrayElementContent()
```

Then the following statements

```rb
a1[0] = b1
a2[x] = b2
a3[1]
a4[x]
```

correspond to the following read/store steps

```mermaid
graph TD
  B1[b1]-->|"store(KnownArrayElementContent(0))"| A1["[post update] a1"]
  B2[b2]-->|"store(UnknownArrayElementContent())"| A2["[post update] a2"]
  A3-->|"read(UnknownArrayElementContent())"| A3I
  A3[a3]-->|"read(KnownArrayElementContent(1))"| A3I["a3[1]"]
  A4[a4]-->|"read(UnknownArrayElementContent())"| A4I["a4[x]"]
  A4-->|"read(KnownArrayElementContent([0..10]))"| A4I
```

assuming that `x` does not have a known constant value.

### The Problem

In the example above, we can see how `a4[x]` gives rise to 11 read steps, since we need to be able to read out values stored at any known or unknown index. Unfortunately, we need to encode this up-front in the `readStep` relation, which means that all such reads will be replicated 11 times, and it would become even worse if we were to increase the range of `KnownArrayElementContent`.

Another issue is that when the data-flow library [generates reverse store steps](https://github.com/github/codeql/blob/main/docs/ql-libraries/dataflow/dataflow.md#example-2) from read steps

```rb
a[x][1] = taint
```

we will get the following reverse store steps

```mermaid
graph TD
  A["[post update] a[x]"]-->|"store(UnknownArrayElementContent())"| A1["[post update] a"]
  A-->|"store(KnownArrayElementContent([0..10]))"| A1
```

whereas what we really wanted was just a single store step

```mermaid
graph TD
  A["[post update] a[x]"]-->|"store(UnknownArrayElementContent())"| A1["[post update] a"]
```

### A Solution

This PR proposes a solution to both problems above, where we generalize the interface of the data-flow library for `readStep` and `storeStep`, to use a new type `ContentSet` instead of `Content`. `ContentSet` will allow us to represent _sets of_ `Content`s:

```ql
newtype ContentSet =
  SingletonContent(Content c) or
  AnyArrayElementContent()
```

And we will be able to interpret `ContentSet`s differently, depending on whether they occur in reads or in stores

```ql
Content getAStoreContent(ContentSet set) {
  set = SingletonContent(result)
  or
  // for reverse stores
  set = AnyArrayElementContent() and
  result = UnknownArrayElementContent()
}

Content getAReadContent(ContentSet set) {
  set = SingletonContent(result)
  or
  set = AnyArrayElementContent() and
  (result = UnknownArrayElementContent() or result = KnownArrayElementContent(_))
}
```

We will still be using `Content`s in access paths, but we will use `ContentSet`s internally, and the data-flow library will make sure to avoid fan-out duplication via `getAStoreContent/getAReadContent`.

Returning to the original example

```rb
a1[0] = b1
a2[x] = b2
a3[1]
a4[x]
```

we will now have the following read/store steps:
```mermaid
graph TD
  B1[b1]-->|"store(SingletonContent(KnownArrayElementContent(0)))"| A1["[post update] a1"]
  B2[b2]-->|"store(AnyArrayElementContent()))"| A2["[post update] a2"]
  A3-->|"read(SingletonContent(UnknownArrayElementContent()))"| A3I
  A3[a3]-->|"read(SingletonContent(KnownArrayElementContent(1)))"| A3I["a3[1]"]
  A4[a4]-->|"read(AnyArrayElementContent())"| A4I["a4[x]"]
```

And for

```rb
a[x][1] = taint
```

we get just a single reverse store step

```mermaid
graph TD
  A["[post update] a[x]"]-->|"store(AnyArrayElementContent())"| A1["[post update] a"]
```

which translates via `getAStoreContent` into a store into `UnknownArrayElementContent` as desired.

### PR Structure

This PR is structured into a sequence of logically self-contained commits, and I strongly encourage reviewing them in order:
- a5040fd0ce4dbca07f39b55714b9ffb1c94383f5: Adds a Ruby test for the reverse store example above.
- 309fd937c1fbb0d422a457b8c027f2b031e15387: Introduces `ContentSet` into the data-flow library.
- c4fbc618a9ef375c296ba6dabf145301958972ba: Syncs the data flow files.
- 725d76e9344c6444f0b68c8a018bbe1a93ee262e: Implements `ContentSet` for Ruby, as per the description above.
- d99bb65ea93297c7eb21d895b3262a18e9d32d23: Implements `ContentSet` for C++ as a simple bijection into `Content`.
- b91858e7cf98cee89e47028ed2a9df45e1621c88: Implements `ContentSet` for Java as a simple bijection into `Content`.
- 7113c1b29c45ad3463035588ff59386695b6f038: Implements `ContentSet` for C# as a simple bijection into `Content`.
- 57f2a7463619dced767ac5cf9260df3c194976da: Implements `ContentSet` for Python as a simple bijection into `Content`.
- 415a1c21071d5f3116319152827ec84261d1bce9: Updates shared Java/C# code for generating flow summaries.
- cee527e03a762ae39c316ea455d9d095df46758f: Updates the documentation on the shared data flow library.


hvitved:
I have pushed another commit that is not strictly needed yet (it will be for https://github.com/github/codeql/pull/7914), but it belongs logically with this PR.

asgerf:
I think there's a point in the design space worth talking about here. In the literature of (sound) JavaScript analysis the standard solution is to have two "unknown" contents rather than one:
- `StoredWithUnknownKey`: any store with an unknown key will write to here
- `ReadWithUnknownKey`: any read with unknown key will read from here

The rule is then:
- Store to unknown key: writes to `StoredWithUnknownKey` and `ReadFromUnknownKey`
- Store to known key K: writes to `K` and `ReadFromUnknownKey`
- Read from unknown key: reads from `ReadFromUnknownKey`
- Read from known key K: reads from `K` and `StoredWithUnknownKey`

(Another common solution is to collapse a heterogenous array into a homogenous array as soon as any operation acts on it with non-constant index, but that is hard to apply in our setting).

I can't immediately say whether this is better than `ContentSet`, or if it actually maps down to the same thing. I guess you'd still need to inform the core library that the reverse of `ReadFromUnknownKey` is `StoredWithUnknownKey`. And in terms of writing access paths by hand, I can certainly see the convenience in being able to talk about the "unknown element" without having to specify if it's the read or write variant.

Anyway, this isn't meant as an objection, just thought I'd mention it in case it helps somehow.

aschackmull:
As discussed offline: Let's move the store-related `ContentSet`-to-`Content` join into `DataFlowImplCommon` and reduce the diff in `DataFlowImpl` for the store steps.

cpp/ql/lib/semmle/code/cpp/dataflow/internal/DataFlowImpl.qll
```diff
@@ -752,7 +782,7 @@ private module Stage1 {
   private predicate revFlowStore(Content c, NodeEx node, boolean toReturn, Configuration config) {
     exists(NodeEx mid, TypedContent tc |
       revFlow(mid, toReturn, pragma[only_bind_into](config)) and
-      fwdFlowConsCand(c, pragma[only_bind_into](config)) and
+      fwdFlowConsCandSet(_, c, pragma[only_bind_into](config)) and
```
aschackmull, line 785: This can just be `fwdFlowConsCand`, right?
```suggestion
      fwdFlowConsCand(c, pragma[only_bind_into](config)) and
```

cpp/ql/lib/semmle/code/cpp/dataflow/internal/DataFlowImpl.qll
```diff
@@ -860,9 +891,9 @@ private module Stage1 {
 
   pragma[nomagic]
   predicate readStepCand(NodeEx n1, Content c, NodeEx n2, Configuration config) {
-    revFlowIsReadAndStored(c, pragma[only_bind_into](config)) and
-    revFlow(n2, pragma[only_bind_into](config)) and
-    read(n1, c, n2, pragma[only_bind_into](config))
+    revFlowIsReadAndStored(pragma[only_bind_into](c), pragma[only_bind_into](config)) and
+    read(n1, c, n2, pragma[only_bind_into](config)) and
+    revFlow(n2, pragma[only_bind_into](config))
```
aschackmull, line 896: This join-order is not optimal, but presumably good enough. Realising the optimal order would require another helper predicate, so I'm not suggesting any change here, merely noting the fact.

cpp/ql/lib/semmle/code/cpp/dataflow/internal/DataFlowImpl.qll
hvitved: Indeed.

aschackmull (approved):
LGTM.

PR #8659, opened 2022-04-04
C#: Add kind tag to Capture model queries.

michaelnebel:


michaelnebel:
This is just a wild shot in getting the queries to run on DCA, but the changes needs to be merged before we can run DCA properly (as it will otherwise fail on main).

hvitved:
> This is just a wild shot in getting the queries to run on DCA, but the changes needs to be merged before we can run DCA properly (as it will otherwise fail on main).

I don't understand: Why can't we just run the queries against this branch only?

michaelnebel:
> > This is just a wild shot in getting the queries to run on DCA, but the changes needs to be merged before we can run DCA properly (as it will otherwise fail on main).
> 
> I don't understand: Why can't we just run the queries against this branch only?

Oh right - something like this 
image

michaelnebel:
It looks like this is working.
I will try and run at single threaded DCA as well and see what the numbers look like in that case.

michaelnebel:
Looks like the analysis time is now reported correctly via DCA.
image

hvitved (approved):

PR #8660, opened 2022-04-04
Java: Improve `RefType.inherits(Member)`

### LINKED ISSUE #5596: Java: `RefType.inherits(Member)` does not work for member types
The predicate [`RefType.inherits(Member)`](https://codeql.github.com/codeql-standard-libraries/java/semmle/code/java/Type.qll/predicate.Type$RefType$inherits.1.html) does not work for member types (member classes or member interfaces), see also #5591.
If this is the intended, it should at least be documented.

Currently the compiler does not even permit using a `NestedType` as argument for `inherits`.

Marcono1234:
Changes:
- Adds `MemberType` to the results; resolves #5596
- Adds constructors to the result (but only if they are directly declared)
- Fixes fields not checking package access

These changes make the `inherits` predicate functionally more complete (based on the JLS), but I am not sure if this really adds any value and whether these changes are therefore worth it.
I have only executed the newly added tests, but have not checked what effects these changes have to other usage of this predicate.

If you want I can also instead only create a pull request for the missing field package access check.


java/ql/lib/semmle/code/java/Type.qll
```diff
@@ -568,12 +568,30 @@ class RefType extends Type, Annotatable, Modifiable, @reftype {
...
+    or
+    // Constructors are not inherited, but this predicate also considers declared members
+    m.(Constructor).getDeclaringType() = this.getSourceDeclaration()
+    or
+    exists(MemberType memberType | memberType = m |
+      m.(MemberType).getDeclaringType() = this
+      or
+      not memberType.isPrivate() and
+      (memberType.isPackageProtected() implies memberType.getPackage() = this.getPackage()) and
+      // And member type is not hidden by another member type with the same name (regardless of visibility)
```
smowton, line 589: There's also https://docs.oracle.com/javase/tutorial/java/IandI/hidevariables.html which would enable you to treat types and fields the same

smowton (commented):
Seems like a valid improvement that could surprise third-party code. @aschackmull do you think this is worth a deprecation dance that would also give us the opportunity to make the name reflect what it does: deprecate this predicate explaining that it only considers fields and methods, then add `declaresOrInherits(Member m)` that considers all `Member`s?

Marcono1234:
> deprecate this predicate explaining that it only considers fields and methods, then add `declaresOrInherits(Member m)` that considers all `Member`s?

Just to clarify, the other option is also to just update the documentation, fix field inheritance and reject this pull request. I am not sure if the added functionality here adds and value; maybe no one ever needs to check whether member types are inherited.

Regarding the "Remaining" point in the description I am also not sure if it is worth fixing this rare corner case, since it might be closely connected with how CodeQL defines method overrides (have not investigated this extensively yet).

java/ql/lib/semmle/code/java/Type.qll
Marcono1234: That seems to only be about fields though, isn't it? Do you mean fields could hide member types and member types could hide fields? I am not sure if that is possible, at least the JLS does not seem to mention it.

Or do you mean I should try to combine the common checks here for `Field` and `MemberType`?

smowton: I meant fields hiding fields and types hiding types and forgot about the possibility of one hiding the other.

PR #8661, opened 2022-04-04
Java: Fix reflection predicate for `getMethod` having non-public method result

Marcono1234:
[`Class.getMethod`](https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/lang/Class.html#getMethod(java.lang.String,java.lang.Class...)) only has `public` methods as result; the check for this was previously missing.

(For `getField` that check already exists, see a few lines below.)


aschackmull (approved):
Looks correct to me.

PR #8665, opened 2022-04-04
C++: add range analysis diff test

rdmarsh2:


cpp/ql/test/library-tests/ir/modulus-analysis/test.cpp
```diff
@@ -1,4 +1,4 @@
-template<typename T> void mod(T value);
+/*template<typename T> void mod(T value);
```
MathiasVP, line 1: Should this still be commented out?

cpp/ql/lib/experimental/semmle/code/cpp/semantic/SemanticExprSpecific.qll
```diff
@@ -119,27 +119,67 @@ module SemanticExprConfig {
...
 
-    final Location getLocation() { result = super.getLocation() }
+    Location getLocation() { none() }
+
+    IR::Instruction asInstruction() { none() }
+
+    IR::Operand asOperand() { none() }
+  }
+
+  class SsaInstructionVariable extends SsaVariable, TSsaInstruction {
```
MathiasVP, line 136: This class can probably be `private`, no?

cpp/ql/lib/experimental/semmle/code/cpp/semantic/SemanticExprSpecific.qll
```diff
@@ -119,27 +119,67 @@ module SemanticExprConfig {
...
+    SsaInstructionVariable() { this = TSsaInstruction(instr) }
+
+    final override string toString() { result = instr.toString() }
+
+    final override Location getLocation() { result = instr.getLocation() }
+
+    final override IR::Instruction asInstruction() { result = instr }
+  }
+
+  class SsaOperand extends SsaVariable, TSsaOperand {
```
MathiasVP, line 148: This class can be private, right? And could we call it `SsaOperandVariable` for consistency?

MathiasVP (approved):
A couple of non-blocking questions, but otherwise LGTM!

PR #8667, opened 2022-04-05
Field-sensitive flow summary generation

hvitved:
This PR makes use of [ContentDataFlow](https://github.com/github/codeql/pull/9024) to calculate precise flow summaries for C# and Java.


michaelnebel:
This implementation generates around 900k rows for .NET runtime. Whether this is due to errors or better granularity is unknown. In any case with the number of rows in this order of magnitude compilation of the model-generator test queries exceeds one minute.
Furthermore, it is not possible to commit the file and try and run the CI as it exceeds 100MB in size, which is the maximum allowed unless lfs is enabled for the repo.

michaelnebel:
Pushed a naive implementation where we discard  all models where the content accesspath exceeds 1.
This narrows the number of produced models for .NET runtime to around 28k.

There are some issues with the new models.
As an example, all hand written models for tuples and keyvaluepair is not compatible with the generated models. The handwritten models refer to the propety, where as the generated models refers to the models private backing field (which is in terms of models are synthetic fields).

PR #8668, opened 2022-04-05
Python: Rewrite concepts to use `extends ... instanceof ...`

RasmusWL:
This solved performance problems experienced in https://github.com/github/codeql/pull/8634, and this commit+PR is to ensure we get this change in as fast as possible.


yoff (approved):
I honestly thought this was already done by a CodeQL query patch...

PR #8670, opened 2022-04-05
JS: Allow MaD flow through properties

asgerf:
Adds the possibility of stepping through a property read `obj -> obj.foo` using a summary CSV row where:
- The `input` part is the empty string, and
- The `output` part is `Member[foo]`.

@RasmusWL what do you think?

### DIFF
--- a/javascript/ql/lib/semmle/javascript/frameworks/data/ModelsAsData.qll
+++ b/javascript/ql/lib/semmle/javascript/frameworks/data/ModelsAsData.qll
@@ -31,27 +31,16 @@ private class RemoteFlowSourceFromCsv extends RemoteFlowSource {
   override string getSourceType() { result = "Remote flow" }
 }
 
-/**
- * Like `ModelOutput::summaryStep` but with API nodes mapped to data-flow nodes.
- */
-private predicate summaryStepNodes(DataFlow::Node pred, DataFlow::Node succ, string kind) {
-  exists(API::Node predNode, API::Node succNode |
-    Specific::summaryStep(predNode, succNode, kind) and
-    pred = predNode.getARhs() and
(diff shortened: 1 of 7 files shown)


RasmusWL (approved):
Nice, looks good to me :+1:

I assume that using `Member[foo,bar]`, adding flow from `obj` to `obj.foo` and  `obj.bar`, would also work?

asgerf:
> I assume that using Member[foo,bar], adding flow from obj to obj.foo and obj.bar, would also work?

Yes, that should work as well.

But you can't do `Member[foo].Member[bar]` because the simple approach I took here won't work when the `.foo` and `.bar` operations occur in different functions. For that to work we'd have to interpret the initial `.foo` as a call to a function with an output spec of `ReturnValue.Member[bar]`.

asgerf:
After ruminating on this some more, I think the more general solution is to add something like `AccessorCall[x]` which models an access to the `x` attribute as a call (even if it's not considered a call in the language IR).

Then a step like `foo -> foo.bar` could be modelled as
```
package;type;Member[foo].AccessorCall[bar];Argument[self];ReturnValue
```
This avoids the somewhat awkward special case where the output must consist of a single token.

To be clear, with the solution currently in this PR it would look like this:
```
package;type;Member[foo];;Member[bar]
```

@aschackmull do you have any thoughts on this? The context is that we want to be able to model flow through things that aren't calls, but in principle could be modelled as a calls. It looks like the two approaches are:
- Add a special rule for specifying a step through a member accesses. This can be implemented a step, without a flow summary.
- Model member access as a call, even though it's not a call in the language IR. I'm still not familiar enough with the flow summary code to foresee what problems this will incur. (JS and Python don't have flow summaries at the moment, and the question is moot for Ruby since everything is a method call there).

PR #8672, opened 2022-04-05
C++: Add `cpp/unused-local-variable` test case with `switch` initializer

jketema:
This is similar to the test case with the `if` initializer, and we should not forget about it once we support `if` initialization.


geoffw0 (approved):
LGTM

PR #8675, opened 2022-04-06
C#: CaptureModel improvements

michaelnebel:
In this PR, we remove summaries where

- Simple types are used (ie. int, bool, long, etc).
- Remove the previous way of excluding GetHashCode and Equals overrides (and IEqutable.Equals implementations) from summary generations.


csharp/ql/src/utils/model-generator/internal/CaptureModelsSpecific.qll
```diff
@@ -66,7 +56,10 @@ predicate asPartialModel = DataFlowPrivate::Csv::asPartialModel/1;
  * Holds for type `t` for fields that are relevant as an intermediate
  * read or write step in the data flow analysis.
  */
-predicate isRelevantType(CS::Type t) { not t instanceof CS::Enum }
```
hvitved, line 69: I think we should update the QL doc here, as it in fact excludes flow through any data-flow node that does not have a relevant type (and this is why no longer need to exclude `GetHashCode` and `Equals`, since they return a `bool`).

hvitved (approved):

PR #8676, opened 2022-04-06
Make security-related TaintTracking Configuration public

pwntester:
Make security-related taint tracking configuration public so they can be imported for sink identification

- Update Sql Injection queries
- Update CommandLineQuery


atorralba (commented):
`RemoteUserInputToArgumentToExecFlowConfig` and `QueryInjectionFlowConfig` are now missing QLDoc, but otherwise this LGTM.

aschackmull (approved):

PR #8403, opened 2022-03-11
Rename all upper-case variables, and all lower-case modules

erik-krogh:
~~TODO: Wait for https://github.com/github/codeql/pull/8323.~~ 

Adds a QL-for-QL queries that puts every name definitions in two categories:  
- Things that should start with an upper-case letter. 
- Things that should start with a lower-case letter. 

Every violation of the above is fixed as part of this PR, luckily almost no code violated the above.  
A few modules in the public API started with a lower-case letter, I've left a deprecated alias for those.  

/cc @jbj

PR #8679, opened 2022-04-06
Java: rename existing getUrl predicate to getRepositoryUrl

erik-krogh:
I would like to eventually make all our code follow our own style guide.  

That involves changing the location providing predicate `getURL` to `getUrl`.   
(See [providing locations in CodeQL](https://codeql.github.com/docs/writing-codeql-queries/providing-locations-in-codeql-queries/#providing-urls)).  

However, that change would cause issues in `MavenPom.qll` where a `getUrl()` predicate is already used for something else.   
With my change the frontend would read the existing `getUrl` as a predicate that provides a location, and that would cause the `MavenPomDependsOnBintray.ql` query to break. 

I would therefore like to rename it.

### DIFF
--- a/java/ql/lib/change-notes/2022-04-06-getUrl-deprecation.md
+++ b/java/ql/lib/change-notes/2022-04-06-getUrl-deprecation.md
@@ -1,0 +1,4 @@
+---
+category: breaking
+---
+* The `getUrl` predicate of `DeclaredRepository` in `MavenPom.qll` has been renamed to `getRepositoryUrl`. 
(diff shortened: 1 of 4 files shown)


JLLeitschuh (commented):
As the person who originally wrote this query, this LGTM! 👍 

Only downside is that this is an API breaking change,  so you may need to keep the old API but deprecate it

erik-krogh:
> Only downside is that this is an API breaking change, so you may need to keep the old API but deprecate it 

I know that is the usual procedure (and I've done it [a](https://github.com/github/codeql/pull/8604) [lot](https://github.com/github/codeql/pull/8323) [recently](https://github.com/github/codeql/pull/8403)).  
In this case I would prefer not to, as it could break things if I change the CodeQL frontend to understand `getUrl()` as a location providing predicate.

JLLeitschuh:
No complaint from me on the API breaking change, I'm guessing that it isn't going to break anyone. But I don't recommend not including a change note about it. Just my 10 cents.

aschackmull (approved):

PR #8680, opened 2022-04-06
Python: Refactor how we find a `Class` from `API::Node`

RasmusWL:
Based on offline conversation we had today. /cc @erik-krogh


tausbn (approved):
Looks good to me! :+1:

RasmusWL:
Performance didn't change :+1:

PR #8681, opened 2022-04-06
Java: Fix Local Temp File/Dir Incorrect Guard Logic

JLLeitschuh:
Resolves: https://github.com/github/codeql/pull/8032#discussion_r841723906

Thanks @aschackmull for your feedback and review!

### DIFF
--- a/java/ql/src/Security/CWE/CWE-200/TempDirUtils.qll
+++ b/java/ql/src/Security/CWE/CWE-200/TempDirUtils.qll
@@ -58,7 +58,6 @@ private predicate isTaintPropagatingFileTransformation(Expr expSource, Expr expr
  * For example, `taintedFile.getCanonicalFile()` is itself tainted.
  */
 predicate isAdditionalFileTaintStep(DataFlow::Node node1, DataFlow::Node node2) {
-  isFileConstructorArgument(node1.asExpr(), node2.asExpr(), _) or
   isTaintPropagatingFileTransformation(node1.asExpr(), node2.asExpr())
 }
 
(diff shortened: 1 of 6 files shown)


java/ql/src/Security/CWE/CWE-200/TempDirUtils.qll
```diff
@@ -58,7 +58,6 @@ private predicate isTaintPropagatingFileTransformation(Expr expSource, Expr expr
  * For example, `taintedFile.getCanonicalFile()` is itself tainted.
  */
 predicate isAdditionalFileTaintStep(DataFlow::Node node1, DataFlow::Node node2) {
-  isFileConstructorArgument(node1.asExpr(), node2.asExpr(), _) or
```
aschackmull, line 61: Care to comment on this change?

aschackmull (approved):
LGTM

java/ql/src/Security/CWE/CWE-200/TempDirUtils.qll
JLLeitschuh: With this, I was having two paths get generated. When rendered, one of the paths wasn't showing the taint step where `new File([tainted], [other])` was being called, but the other was. I'm not sure if it's because `isAdditionalFileTaintStep` is written wrong, or not. But regardless, this line handles the case I was trying to cover anyways:
https://github.com/github/codeql/blob/879b8a1200fa946d7e8db5dd4ce360a8d37c8d52/java/ql/lib/semmle/code/java/dataflow/ExternalFlow.qll#L331

atorralba:
Note that `ql/java/ql/lib/semmle/code/java/os/OSCheck.qll` needs auto-formatting.

PR #8686, opened 2022-04-07
Java: Timing attack

ahmed-farid-dev:
A constant-time algorithm should be used for checking the value of info. In other words, the comparison time should not depend on the content of the input, Otherwise, an attacker may be able to implement a timing attacks that may reveal the value of sensitive info


ahmed-farid-dev:
Hi, any update here ?

smowton:
Security lab are making their initial assessment. That can take a couple of weeks.

ahmed-farid-dev:
Hi, It's been five months. Any update?

smowton:
We will review the PR here once the conversation at https://github.com/github/securitylab/issues/664 about what the query should do has been resolved

PR #8687, opened 2022-04-07
JS: Fix two missing bits of data flow

asgerf:
Fixes two issues I found while debugging data flow through a codebase.

### DIFF
--- a/javascript/ql/lib/semmle/javascript/DOM.qll
+++ b/javascript/ql/lib/semmle/javascript/DOM.qll
@@ -420,6 +420,13 @@ module DOM {
     t.startInProp("target") and
     result = domEventSource()
     or
+    t.start() and
+    exists(DataFlow::ClassNode cls |
+      cls.getASuperClassNode().getALocalSource() =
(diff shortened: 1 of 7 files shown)


javascript/ql/lib/semmle/javascript/DOM.qll
```diff
@@ -420,6 +420,13 @@ module DOM {
     t.startInProp("target") and
     result = domEventSource()
     or
+    t.start() and
+    exists(DataFlow::ClassNode cls |
+      cls.getASuperClassNode().getALocalSource() =
+        DataFlow::globalVarRef(any(string s | s.matches("HTML%Element"))) and
+      result = cls.getAReceiverNode()
+    )
+    or
```
erik-krogh, line 429: How about using `ClassNode::getAnInstanceReference()`, and placing it in `domValueRef()` below?

erik-krogh (approved):
:+1: (assuming the evaluation works out).

erik-krogh (approved):

Issue #8645, opened 2022-04-02
Java: `ArrayInit.getInit` has indices < 0 for annotations


Marcono1234:
### Version
CodeQL CLI version: 2.8.4 

### Description of the issue

For array inits as part of annotations CodeQL's `ArrayInit.getInit` erroneously has `index` values < 0 as result.

### Example query
```ql
import java

from ArrayInit init, int index, Expr e
where
  init.getInit(index) = e
  and index < 0
select init, index, e
```
[Query Console link](https://lgtm.com/query/6322278065984969890/)

smowton:
Appears to be a deliberate choice, extracting negative-indexed and positive-indexed array members (copies of the same array member, using different ways the compiler exposes information about annotations). There should probably be some reconciliation of the information extracted in QL though, or maybe just hiding whichever version is less preferable to expose to users.

aschackmull:
It appears that only the negatively indexed expressions have associated locations (see https://lgtm.com/query/7342947201087264738/), so the different way of extracting those expressions definitely makes a difference. I agree that the negative indices should not occur at the QL-level and it seems wrong to extract two entities for the same expression, so I think this ought to be fixed in the extractor.

Marcono1234:
The difference between those two seems to be that the one with negative index represents the actual expression used in the source (e.g. a read of Java `String` field `MediaType.APPLICATION_XML`) whereas the expression with positive index is the corresponding constant (e.g. a CodeQL `StringLiteral` without location).

Maybe that is the reason why this is done? Because otherwise for constants from external class files it would not be possible at the moment to obtain the value (see https://github.com/github/codeql/discussions/8650).

Issue #6275, opened 2021-07-13
Java: Add predicates for `AnnotationElement` default value


Marcono1234:
**Description of the issue**
The elements of annotation interfaces can have default values, for example:
```java
@interface MyInterface {
    String value() default "test";
}
```

However, currently `AnnotationElement` does not provide any predicates for detecting such default values. It appears they do not even exist in the database unless an annotation of the declaring annotation type is used somewhere. And even then they only exist as expression without source location.
(Note that the `Annotation` predicates always include the default values in their results.)

It would be good to add the following predicates
- `predicate hasDefaultValue()`
Note that the predicate `isDefault()` does not work for this because an annotation element with default value is not a `default` method, in fact it is an abstract method with the `AnnotationDefault` attribute in bytecode (see [JVM spec 16 §4.7.22](https://docs.oracle.com/javase/specs/jvms/se16/html/jvms-4.html#jvms-4.7.22)).
- `Expr getDefaultValue()`
Note that even though an `AnnotationElement` is effectively a method (see #5399), the default value should probably not be modeled as result of a synthetic `return` in the method body, but be stored directly in the database. As pointed out above, an annotation element with default value is not a regular `default` method, therefore modeling it as such would be irritating (e.g. it would have to be modeled as `abstract` method whose `getBody()` has a result).
Ideally this predicate would also work for annotations declared in third party code, not being part of the source, in case that is possible.

PR #8695, opened 2022-04-07
Java: Add support for Annotation types stub generation

atorralba:
Adds support for Annotation types in our Java stub generator.

Due to default values for Annotation methods not being currently supported, the stubs generated may need manual correction if the original types have default values, e.g. https://github.com/spring-projects/spring-framework/blob/main/spring-web/src/main/java/org/springframework/web/bind/annotation/RequestMapping.java#L85.


java/ql/src/utils/stub-generator/Stubs.qll
```diff
@@ -16,7 +16,12 @@ abstract private class GeneratedType extends ClassOrInterface {
   }
 
   private string stubKeyword() {
-    this instanceof Interface and result = "interface"
+    this instanceof Interface and
+    (
+      this instanceof AnnotationType and result = "@interface"
+      or
+      result = "interface"
```
Marcono1234, line 23: Doesn't this have to use `if ... else` because currently for `AnnotationType` both `@interface` and `interface` would be results?

java/ql/src/utils/stub-generator/Stubs.qll
```diff
@@ -391,6 +402,24 @@ private string stubMember(Member m) {
   )
 }
 
+private string stubAnnotation(Annotation a) {
+  if exists(a.getAValue())
+  then result = a.toString() + "(" + concat(stubAnnotationValue(a.getAValue()), ",") + ")"
+  else result = a.toString()
+}
+
+private string stubAnnotiationSimpleValue(Expr value) {
```
Marcono1234, line 411: There seems to be a typo: <code>stubAnnot<del>i</del>ationSimpleValue</code>

java/ql/src/utils/stub-generator/Stubs.qll
```diff
@@ -391,6 +402,24 @@ private string stubMember(Member m) {
   )
 }
 
+private string stubAnnotation(Annotation a) {
+  if exists(a.getAValue())
+  then result = a.toString() + "(" + concat(stubAnnotationValue(a.getAValue()), ",") + ")"
```
Marcono1234, line 407: If more than one annotation element has a value, it is necessary to include the annotation element names, e.g. `@Annotation(value=1, other=2)`

java/ql/src/utils/stub-generator/Stubs.qll
```diff
@@ -391,6 +402,24 @@ private string stubMember(Member m) {
...
+  else result = a.toString()
+}
+
+private string stubAnnotiationSimpleValue(Expr value) {
+  result = value.(FieldAccess).getField().getQualifiedName() or
+  result = value.(Literal).toString()
+}
```
Marcono1234, line 414: Would it be necessary to cover the following as well?
- annotations containing annotations, e.g. `@Outer(@Inner)`
- `TypeLiteral` (which does not extend CodeQL `Literal`), e.g. `String.class`
- `CompileTimeConstantExpr`, which is neither a `Literal` nor a `FieldRead`, e.g. negative literals or arithmetic operations

java/ql/src/utils/stub-generator/Stubs.qll
```diff
@@ -391,6 +402,24 @@ private string stubMember(Member m) {
...
 
+private string stubAnnotation(Annotation a) {
+  if exists(a.getAValue())
+  then result = a.toString() + "(" + concat(stubAnnotationValue(a.getAValue()), ",") + ")"
+  else result = a.toString()
+}
+
+private string stubAnnotiationSimpleValue(Expr value) {
+  result = value.(FieldAccess).getField().getQualifiedName() or
+  result = value.(Literal).toString()
```
Marcono1234, line 413: Would it make sense to directly use `getLiteral()` instead (to which `toString()` delegates)? Because `toString()` makes no guarantees about its result.

java/ql/src/utils/stub-generator/Stubs.qll
Marcono1234, line 407: Instead of `a.toString()`, would it make sense to explicitly use `a.getType().getName()`? `toString()` does not make any guarantees about its result.

java/ql/src/utils/stub-generator/Stubs.qll
```diff
@@ -391,6 +402,24 @@ private string stubMember(Member m) {
...
+}
+
+private string stubAnnotationValue(Expr value) {
+  result = stubAnnotiationSimpleValue(value)
+  or
+  value instanceof ArrayInit and
+  result = "{" + concat(stubAnnotiationSimpleValue(value.(ArrayInit).getAnInit()), ",") + "}"
```
Marcono1234, line 420: Would it be necessary to order the elements of `concat` by their index in the array? Currently due to usage of `getAnInit()` it appears the values are ordered by their string representation.

Marcono1234 (commented):
Hopefully the following comments are helpful, and this is not too intrusive.
I am not a member of the project so feel free to consider these only as suggestions.

java/ql/src/utils/stub-generator/Stubs.qll
atorralba: I could've sweared this worked just fine, but you're right that it looks it would behave as you say. Changing it, thanks.

```suggestion
    (if this instanceof AnnotationType then result = "@interface" else result = "interface")
```

atorralba: Done in 4f5763f.

java/ql/src/utils/stub-generator/Stubs.qll
atorralba: Fixed in 2fb30d8, thanks!

java/ql/src/utils/stub-generator/Stubs.qll
atorralba: Good point, done in 495ce95.

java/ql/src/utils/stub-generator/Stubs.qll
atorralba: Done in 56d1dde.

java/ql/src/utils/stub-generator/Stubs.qll
atorralba: Done in 495ce95f3856ca9d5416f312abb96624ff306b1e.

java/ql/src/utils/stub-generator/Stubs.qll
atorralba: That's probably better, yes. Done in b18b4cf.

java/ql/src/utils/stub-generator/Stubs.qll
```diff
@@ -404,20 +400,45 @@ private string stubMember(Member m) {
...
 
-private string stubAnnotiationSimpleValue(Expr value) {
-  result = value.(FieldAccess).getField().getQualifiedName() or
-  result = value.(Literal).toString()
+private string stubAnnotationSimpleValue(Expr value) {
+  result = value.(FieldAccess).getField().getQualifiedName()
+  or
+  result = value.(Literal).getLiteral()
+  or
+  not value instanceof Literal and
```
Marcono1234, line 419: Probably also have to make sure that it is not a `FieldAccess` (or rather `FieldRead`) either since `CompileTimeConstantExpr` includes that as well.

java/ql/src/utils/stub-generator/Stubs.qll
```diff
@@ -404,20 +400,45 @@ private string stubMember(Member m) {
...
+  (
+    result = value.(CompileTimeConstantExpr).getStringValue() or
+    result = value.(CompileTimeConstantExpr).getBooleanValue().toString() or
+    result = value.(CompileTimeConstantExpr).getIntValue().toString()
+  )
+  or
+  result = value.(Annotation).getType().getName()
```
Marcono1234, line 426: Is this really correct? Shouldn't it be `result = stubAnnotation(value)`? `@Outer(@Inner)` was only a simple example, the annotations and their values can be nested arbitrarily deep, e.g. `@Outer(@Inner(@Inner2("some value")))`.

This would then require to add the `@` prefix in `subAnnotation` instead of in `stubAnnotations`.

java/ql/src/utils/stub-generator/Stubs.qll
```diff
@@ -404,20 +400,45 @@ private string stubMember(Member m) {
...
-  result = value.(FieldAccess).getField().getQualifiedName() or
-  result = value.(Literal).toString()
+private string stubAnnotationSimpleValue(Expr value) {
+  result = value.(FieldAccess).getField().getQualifiedName()
+  or
+  result = value.(Literal).getLiteral()
+  or
+  not value instanceof Literal and
+  (
+    result = value.(CompileTimeConstantExpr).getStringValue() or
```
Marcono1234, line 421: Would probably be necessary to convert the value to a proper string literal, by adding `"` as prefix and suffix and escaping any `"` and `\` inside the string value.

java/ql/src/utils/stub-generator/Stubs.qll
```diff
@@ -404,20 +400,45 @@ private string stubMember(Member m) {
...
 private string stubAnnotationValue(Expr value) {
-  result = stubAnnotiationSimpleValue(value)
+  result = stubAnnotationSimpleValue(value)
   or
   value instanceof ArrayInit and
-  result = "{" + concat(stubAnnotiationSimpleValue(value.(ArrayInit).getAnInit()), ",") + "}"
+  result =
+    "{" +
+      concat(int i, Expr arrayElement |
+        arrayElement = value.(ArrayInit).getInit(i)
```
Marcono1234, line 438: Might be necessary to verify that this behaves as expected, and does not include duplicate values due to #8645.

Marcono1234 (commented):
Thanks a lot for considering the feedback!
There might be a few remaining issues.

Regarding annotation default values: There might not be any additional work needed because they seem to be included when an annotation is used (at least the last time I checked this), see also #6275.

java/ql/src/utils/stub-generator/Stubs.qll
atorralba: Yes, I thought of that, but using `stubAnnotation` there creates a non-monotonic recursion and CodeQL doesn't like it. I thought that we could at least support the basic case of a nested annotation, and if we end up needing to stub more complex cases we can think about this again.

Marcono1234: Thanks for the clarification, I had not tested the code myself. It probably still needs the `@` prefix then because that won't be included by `getName()`, right?

Do you think it would be useful to add a comment there describing the limitation you mentioned.

atorralba: Sure, did it in 373b64b.

java/ql/src/utils/stub-generator/Stubs.qll
atorralba: On second thought, I think we don't need to handle that much level of detail: these are stubs, so literal values aren't relevant. I changed it to just use the default value we normally use for fields and return values.

java/ql/src/utils/stub-generator/Stubs.qll
atorralba: On second thought, I think we don't need to handle that much level of detail: these are stubs, so literal values aren't relevant. I changed it to just use the default value we normally use for fields and return values.

java/ql/src/utils/stub-generator/Stubs.qll
atorralba: Makes sense. I added `i >= 0`, which should handle this issue just in case, thanks!

java/ql/src/utils/stub-generator/Stubs.qll
Marcono1234: Ok, but you probably have to special-case `String` elements because `stubDefaultValue` would use `null` for them, which is not allowed as annotation element value.

java/ql/src/utils/stub-generator/Stubs.qll
joefarebrother: Using a `language[monotonicAggregates]` annotation on `stubAnnotation` would allow the recursion to be monotinic

atorralba: TIL, thanks 😄 Now it looks better: 1f1e494

tamasvajk (commented):
Should there be a test for this? There's a similar query in C#, we added a not so human readable test in https://github.com/github/codeql/tree/main/csharp/ql/test/query-tests/Stubs/Minimal. It might be worth considering for Java too.

atorralba:
> Should there be a test for this?

Good point @tamasvajk, thanks! I had to create a `.jar` file to test the stub generation, since our usual stubs are source files, which are ignored by the stub generator.

See 7687c8c for details, let me know if something doesn't add up.

java/ql/src/utils/stub-generator/Stubs.qll
```diff
@@ -27,19 +28,24 @@ abstract private class GeneratedType extends ClassOrInterface {
...
-    if this.isStatic() then result = "static " else result = ""
+    if this.(NestedType).isStatic() then result = "static " else result = ""
   }
 
   private string stubAccessibilityModifier() {
     if this.isPublic() then result = "public " else result = ""
   }
 
+  private string stubAnnotations() {
+    result = concat(stubAnnotation(this.(AnnotationType).getAnAnnotation()) + "\n")
```
tamasvajk, line 39: Do we need a stable order for the annotations? I see that we don't order the members in `stubMembers` either, but we do `order by` all `concat`s in C#. Just wondering.

tamasvajk (approved):
Looks reasonable to me. I added a question regarding stable ordering of elements, feel free to ignore it as it's not specific to annotations.

java/ql/src/utils/stub-generator/Stubs.qll
smowton: This would seem wise, ordering by their fully-qualified names for example

java/ql/src/utils/stub-generator/Stubs.qll
```diff
@@ -96,15 +104,14 @@ abstract private class GeneratedType extends ClassOrInterface {
...
-    or
-    result = this.getAGeneratedMember().(NestedType)
+    result = this.getAnInterestingBaseType() or
+    result = this.getAGeneratedMember().(Callable).getReturnType() or
+    result = this.getAGeneratedMember().(Callable).getAParameter().getType() or
+    result = this.getAGeneratedMember().(Field).getType() or
+    result = this.getAGeneratedMember().(NestedType) or
+    result = this.(AnnotationType).getAnAnnotation().getType() or
+    result = this.(AnnotationType).getAnAnnotation().getAValue().getType() or
+    result = this.(AnnotationType).getAnAnnotation().getAValue().(ArrayInit).getAnInit().getType()
```
smowton, line 114: Note that `getAValue` was recently deprecated because it's usually not useful -- use `getAnArrayValue(_)` instead, which unpacks array members

java/ql/src/utils/stub-generator/Stubs.qll
```diff
@@ -391,6 +398,47 @@ private string stubMember(Member m) {
   )
 }
 
+language[monotonicAggregates]
+private string stubAnnotation(Annotation a) {
+  if exists(a.getAValue())
```
smowton, line 403: `getValue(_)`

atorralba:
Thanks for the review @tamasvajk @smowton and sorry for the delay. Comments addressed in 
f860ae8.

smowton (approved):

java/ql/src/utils/stub-generator/Stubs.qll
```diff
@@ -36,7 +36,12 @@ abstract private class GeneratedType extends ClassOrInterface {
   }
 
   private string stubAnnotations() {
-    result = concat(stubAnnotation(this.(AnnotationType).getAnAnnotation()) + "\n")
+    result =
+      concat(Annotation an |
+        this.(AnnotationType).getAnAnnotation() = an
+      |
+        stubAnnotation(an), "\n" order by an.getType().getQualifiedName()
```
Marcono1234, line 43: This is actually a behavior change: Previously it would add a `\n` after each annotation. Now it only adds it between annotations but the last one does not have a trailing `\n` (see also test output changes).

When the last annotation has values (and therefore ends with `)`) that still seems to produce valid Java code, but I assume when an annotation has no values this produces malformed Java code:
```java
@MyAnnotationpublic void m() {
}
```

atorralba: Thanks @Marcono1234, good catch. I addressed this in #10662.

PR #8696, opened 2022-04-07
Python: Improve experimental modeling for `pymongo`

RasmusWL:
Besides just adding new tests, I also fixed up the modeling, and did a small bit of refactoring. We want to promote this query at some point anyway, so could just as well do it now :shrug:


python/ql/src/experimental/semmle/python/frameworks/NoSQL.qll
```diff
@@ -34,40 +38,36 @@ private module NoSql {
...
    */
-  private DataFlow::LocalSourceNode mongoDB(DataFlow::TypeTracker t) {
+  private DataFlow::LocalSourceNode mongoDBInstance(DataFlow::TypeTracker t) {
     t.start() and
     (
       exists(SubscriptNode subscript |
-        subscript.getObject() = mongoInstance().getAUse().asCfgNode() and
+        subscript.getObject() = mongoClientInstance().getAUse().asCfgNode() and
         result.asCfgNode() = subscript
       )
```
tausbn, line 55: This could also be written as 
```ql
result.asCfgNode() = any(SubscriptNode subscript | subscript.getObject() = mongoClientInstance().getAUse().asCfgNode())
```
though it's debatable whether this is an improvement.

Also, we really ought to add support for subscripting in the API graph (as a call to `__getitem__`). It really bothers me that we have to resort to type tracking here. :cry:

tausbn (approved):
This looks good to me! :+1:

Only one minor comment (that does not need to be addressed before merging).

python/ql/src/experimental/semmle/python/frameworks/NoSQL.qll
RasmusWL: > Also, we really ought to add support for subscripting in the API graph (as a call to __getitem__). It really bothers me that we have to resort to type tracking here. cry

Yes, that would be good :+1:

PR #8700, opened 2022-04-08
Swift: first skeleton extractor

redsun82:
This adds a first dummy extractor for swift.

Running `bazel run //swift:install` will create an `extractor_pack`
directory in `swift`. From that moment providing `--search-path=swift`
will pick up the extractor.


AlexDenisov (dismissed):
LGTM, but I'd get some more eyes on it :)

WORKSPACE.bazel
```diff
@@ -1,2 +1,32 @@
 # Please notice that any bazel targets and definitions in this repository are currently experimental
 # and for internal use only.
+
+workspace(name = "ql")
```
redsun82, line 4: @criemen I would have called this workspace `codeql`, but we used that already... should we do some renaming? Or shall we leave it like this?

redsun82: @criemen as you suggested in chat, I'll rename this workspace to `@codeql` and semmle-code to `@semmle-code`, but I think I'll do it in a separate pair of PRs doing only that. Currently this PR does not depend on the internal one (it's the internal one depending on this), while a bazel workspace rename has it actually the other way around (the codeql repo change would require the `codeql` name to be freed up internally before using it). I would prefer to keep PR dependencies one-way :slightly_smiling_face:

redsun82: right!

cpp/BUILD.bazel
```diff
@@ -4,14 +4,14 @@ load("@rules_pkg//:mappings.bzl", "pkg_filegroup")
 
 alias(
     name = "dbscheme",
-    actual = "//cpp/ql/lib:dbscheme",
+    actual = "@ql//cpp/ql/lib:dbscheme",
```
criemen, line 7: Why do we need this? Is this needed for the integration into the main repo?

swift/README.md
```diff
@@ -0,0 +1,9 @@
+## Warning
+
+The Swift codeql package is an experimental and unsupported work in progress.
```
criemen, line 3:
```suggestion
The Swift codeql package is experimental and unsupported work in progress.
```

swift/BUILD.bazel
```diff
@@ -0,0 +1,65 @@
...
+pkg_filegroup(
+    name = "extractor-pack",
+    srcs = [
+        ":extractor-pack-arch",
+        ":extractor-pack-generic",
+    ],
+    visibility = ["//visibility:public"],
+)
+
+pkg_install(
```
criemen, line 58: I wasn't aware of `pkg_install` - it is not documented, and it's replacement, `pkg_tree`, is not ready yet: https://github.com/bazelbuild/rules_pkg/issues/388.

swift/BUILD.bazel
```diff
@@ -0,0 +1,65 @@
...
+    name = "create-extractor-pack",
+    srcs = ["//swift:extractor-pack"],
+    args = [
+        "--destdir",
+        current_source_dir() + "/extractor-pack",
```
criemen, line 63: I wonder if we should move this from the hermeticity-breaking `current_source_dir` rule to the documentation instead.
That would allow us to get rid of the extra repository, and of somebody accidentally using `current_source_dir` anywhere else.
What do you think?

swift/extractor/BUILD.bazel
```diff
@@ -0,0 +1,5 @@
+cc_binary(
+    name = "extractor",
+    srcs = ["main.cpp"],
+    visibility = ["//swift:__pkg__"],
```
criemen, line 4: I suggest adding `target_compatible_with` for linux and macos here.

misc/bazel/workspace.bzl
```diff
@@ -0,0 +1,38 @@
...
+        sha256 = "62eeb544ff1ef41d786e329e1536c1d541bb9bcad27ae984d57f18f314018e66",
+        urls = [
+            "https://mirror.bazel.build/github.com/bazelbuild/rules_pkg/releases/download/0.6.0/rules_pkg-0.6.0.tar.gz",
+            "https://github.com/bazelbuild/rules_pkg/releases/download/0.6.0/rules_pkg-0.6.0.tar.gz",
+        ],
+    )
+
+    maybe(
+        repo_rule = http_archive,
+        name = "platforms",
```
criemen, line 32: Do we need to pull this in?
If we can avoid it, we'd not need to coordinate the bump cross-submodules when we bump it in semmle-code

misc/bazel/BUILD.toolchain.tpl
```diff
@@ -0,0 +1,12 @@
+load("@//misc/bazel:toolchain.bzl", "codeql_cli_toolchain")
```
criemen, line 1: this file is missing.

Also, I think this entire toolchain stuff should be deleted until it's in a shape we can use it somewhere?

misc/bazel/test.bzl
```diff
@@ -0,0 +1,38 @@
...
+_test_script = rule(
+    implementation = _test_script_impl,
+    attrs = {
+        "srcs": attr.label_list(allow_files = True),
+        "_template": attr.label(default = "//misc/bazel:test.template.py", allow_single_file = True),
+    },
+    toolchains = ["//:toolchain_type"],
+)
+
+def codeql_test(*, name, srcs, deps):
```
criemen, line 25: The same here - I don't think this is used anywhere yet, so it shouldn't go in the tree yet.

criemen (commented):
Some comments about the bazel stuff.
I don't know enough about the rest of the extractor pack setup to review for example the manifest and qlpack files.

swift/BUILD.bazel
redsun82: yeah, I was also a bit wary of this, then decided to go with what was done in ruby. We could skip the `--destdir` and point `.clodeqlmanifest` to the corresponding place in the target's runfiles (as we do not need to worry about window here)...

cpp/BUILD.bazel
redsun82: Yeah I think there was an issue not doing this in semmle-code, but I'm not so sure any more. I'll try to revert and report back

misc/bazel/workspace.bzl
redsun82: in semmle-code this is executed after semmle-code itself pulls those in, so the versions specified in semmle-code have priority over these, and semmle-code can change them without needing a submodule bump. Did I just write semmle-code way too many times? :smile:

misc/bazel/BUILD.toolchain.tpl
redsun82: ouch, I was supposed to have git stashed these veeery experimental toolchain stuff, sorry about this

misc/bazel/test.bzl
redsun82: yep, I missed to do a proper git stash here

misc/bazel/workspace.bzl
criemen: Ah, that makes sense. I guess for consistency we should still bump those together, but then it's not too important to do so at the precise same time :)

swift/BUILD.bazel
criemen: Can you explain what you mean with "what was done in ruby"? I have not looked at their extractor build at all.

swift/README.md
redsun82: hmm, "work in progress" is a substantive, not an adjective (although we do tend to use it as if it were one), so to be really nitpickingly correct I think the indeterminate article should be there...

swift/BUILD.bazel
redsun82: they have a `create-extractor-pack.sh` script that when run without any arguments will instantiate the extractor pack in `ruby/extractor-pack`, which in turn is referenced by `ruby/.codeqlmanifest.json`.

Any way, I found a more elegant solution by wrapping the install script setting `DESTDIR` using `BUILD_WORKSPACE_DIRECTORY', so the scary `current_source_dir()` can go away :slightly_smiling_face:

misc/bazel/workspace.bzl
redsun82: also, at some point we will be able to use [bzlmod](https://docs.bazel.build/versions/5.0.0/bzlmod.html) and this will be so, so much better :slightly_smiling_face:

swift/BUILD.bazel
redsun82: (which by the way allows cleaning up the pack directory, which might otherwise pose problems with leftover files)

cpp/BUILD.bazel
redsun82: nope, could not find the reason, removed all `@ql` here

swift/README.md
criemen: I certainly don't feel hard about this one, feel free to discard this :)

swift/BUILD.bazel
criemen: Yeah that's the reason we do the whole delete+unzip in sembuild all the time - having a good solution that's a bit faster would be a good model for something to adopt there, too :-)

swift/extractor/BUILD.bazel
```diff
@@ -1,5 +1,10 @@
 cc_binary(
     name = "extractor",
     srcs = ["main.cpp"],
+    target_compatible_with = select({
```
criemen, line 4: Irritating that we need the select here, instead of just listing linux and macos 😬

criemen (approved):
LGTM, with the expectation of a follow-up PR that does the workspace renaming dance.

swift/extractor/BUILD.bazel
redsun82: yeah, I also got surprised the list was an AND and not an OR, they should have given a way to do general logic expressions here...

AlexDenisov (approved):

PR #8702, opened 2022-04-08
C++: Use `isSanitizerOut(DataFlow::Node node)` in `cpp/command-line-injection`

jketema:
After discussion in https://github.com/github/codeql/pull/8435

I performed a DCA experiment and did a comparative run on all LGTM C/C++ projects, neither showed any regressions.

Still to do: DCA experiment with the `nightly` query set.


aschackmull (approved):

jketema:
DCA shows one slowdown. This brings performance of `cpp/cleartext-storage-buffer` on `wireshark` back to what it was before using flow states in `cpp/command-line-injection`, see also https://github.com/github/codeql-dca-main/issues/3659. So, not blocking.

PR #8710, opened 2022-04-10
JS: drag and drop API Xss sources

bananabr:
The drag and drop JavaScript API is an external input source that can be manipulated in a very similar way to the clipboard API.
This push request includes support for this XSS source.


bananabr:
Hey @erik-krogh,

Thanks for reviewing this so fast. Should I be the one to work on the change notes?

erik-krogh:
> Thanks for reviewing this so fast. Should I be the one to work on the change notes?

I didn't actually review yet, I just triggered the automated tests.   
(And that includes a change-note check).  

But yeah, LGTM :+1:  
Could you add the change-note?   
[This document](https://github.com/github/codeql/blob/main/docs/change-notes.md) describes how we do change-notes now, and you can see some existing examples in the [change-notes folder](https://github.com/github/codeql/tree/main/javascript/ql/lib/change-notes).  
(Check the git history of the folder if you want to see a bit more examples).

bananabr:
Hi @erik-krogh,

Maybe you can help me to improve this contribution further. One event source both this PR and the existing clipboard library are missing, is the one where the onpaste or ondrop properties of a DOM element are assigned explicitly.

```js
let div = document.createElement("div");
div.onpaste = function (e) {
    const { clipboardData } = e.originalEvent;
    if (!clipboardData) return;

    const text = clipboardData.getData('text/plain');
    const html = clipboardData.getData('text/html');
    if (!text && !html) return;

    e.preventDefault();

    const div = document.createElement('div');
    if (html) {
        div.innerHTML = html; // NOT OK
    } else {
        div.textContent = text;
    }
    document.body.append(div);
}
```
I tried including the following _exists_ statement in the **pasteEvent** predicate but it did not work.

```js
  exists(DataFlow::PropWrite pw, DataFlow::FunctionNode fn | pw = DOM::domValueRef().getAPropertyWrite() |
    pw.getPropertyName() = "onpaste" and
    pw.getRhs() = fn and
    result = fn.getParameter(0)
  )
```

erik-krogh:
@bananabr: are you planning on adding more to this PR?  

I want to run another performance evaluation before I merge, but that will take about a week before I can do that (for technical reasons). 
So you got plenty of time if you want to take a look at [InputEvent](https://developer.mozilla.org/en-US/docs/Web/API/InputEvent), or if you want to combine all the DOM-event stuff into a single `DOMEvents.qll` file.

bananabr:
Hi @erik-krogh,

Sorry for the late reply. I will be really busy for the next couple and might not be able to contribute to this PR any time soon.
Feel free to merge the files and include the InputEvent. Although I think it's a more unlikely case, it does indeed have an associated DataTransfer object that can come from both clipboard and drag & drop sources.

kaeluka (approved):
Thanks, @erik-krogh. LGTM!

bananabr:
> Thank you, @bananabr, for a great contribution! As @erik-krogh has added a bunch of code to your PR, I'll be joining as a reviewer.

Glad I could contribute to such an awesome project @kaeluka

PR #8728, opened 2022-04-13
QL: Add facilities for data flow

asgerf:
This adds facilities for reasoning about data flow between expression in CodeQL.

As a rule of thumb, "A flows to B" means the two expressions could be unified given a sufficient number of predicate inlinings, followed by DNF conversion. For example, `A` flows to `B` and `C` here, but `B` does not flow to `C`:
```
x = A and (x = B or x = C)

--> (to DNF)

(x = A and x = B) or (x = A and x = C)
```
Since (a copy of) `A,B` end up in the same clause where they bound by equality, `A` flows to `B` and vice versa. Same for `A,C` but not for `B,C`.

"Flows to" is a symmetric relation, but "flows to with call steps" is not, which often comes up in the implementation. This is why I prefer this terminology over something like "unifiable with".

This PR adds type-tracking, but not `DataFlow::Configuration` or support for path queries. Also it doesn't reason about subclassing, overriding, and dispatch (I had to get back to work at some point).

Like with procedural languages, we reduce flow to a graph problem with call and return edges, where we must find paths that do not violate call/return matching. Any call edge can be converted to a return edge simply by flipping it. Disjunctions are modelled as call sites, treating their operands as individual call targets. To avoid spurious flow between different disjuncts, there is an additional restriction not found in procedural languages: A call edge may not follow immediately after a return edge from the same call-site/disjunction. That is, you can't step back into a call you just came out of. This prevents us from stepping out of a disjunction and immediately back into another operand of that disjunction (which would lead to `B->C` flow in the example above). There is some additional justification for this in the comments.

---
This all started with an attempt to measure framework coverage for models implemented in CodeQL (as an alternative to porting all models to MaD). The PR includes the new `FrameworkCoverage.ql` query which spits out [CSV like this](https://gist.github.com/asgerf/f8a9f1b5e64f257344cecf163465333c). It simply measures the named members we recognize from each library (in principle this should be access paths, but there can be infinitely many access paths).

It finds a lot of interesting flows, though it should be clear that this is not currently an accurate report of our framework coverage:
- The limited handling of subclassing/dispatch leads to incompleteness. E.g. for `express` it does not report anything related to request objects.
- Promise-related libraries look like they have a huge number of features.
- JSON parsers all appear to have a feature called `query` because at some point we look up a property called `query` on a JSON object, but it's not logically a feature of the JSON parser itself.
- (A lot of libraries appear to have no features, but this is usually correct because the library itself is a function, which is all we care about)

Still, I think it's good validation for the fundamentals that are there.

### DIFF
--- a/config/identical-files.json
+++ b/config/identical-files.json
@@ -383,7 +383,8 @@
     "csharp/ql/test/TestUtilities/InlineExpectationsTest.qll",
     "java/ql/test/TestUtilities/InlineExpectationsTest.qll",
     "python/ql/test/TestUtilities/InlineExpectationsTest.qll",
-    "ruby/ql/test/TestUtilities/InlineExpectationsTest.qll"
+    "ruby/ql/test/TestUtilities/InlineExpectationsTest.qll",
+    "ql/ql/test/TestUtilities/InlineExpectationsTest.qll"
   ],
   "C++ ExternalAPIs": [
     "cpp/ql/src/Security/CWE/CWE-020/ExternalAPIs.qll",
--- a/ql/ql/src/codeql_ql/ast/Ast.qll
+++ b/ql/ql/src/codeql_ql/ast/Ast.qll
@@ -79,6 +79,7 @@ class AstNode extends TAstNode {
 
   /** Gets an annotation of this AST node. */
   Annotation getAnAnnotation() {
+    not this instanceof Annotation and // avoid cyclic parent-child relationship
     toQL(this).getParent() = pragma[only_bind_out](toQL(result)).getParent()
   }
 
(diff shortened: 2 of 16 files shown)


erik-krogh:
I think you broke the `ql/deprecated-annotation` query.  
If you check the [Code scanning results](https://github.com/github/codeql/pull/8728/checks?check_run_id=6004057283), there are a bunch of spurious fixes.

erik-krogh (approved):
This is seriously impressive work :rocket: 

I went through all the commits, and I didn't find anything that seemed obviously wrong :+1: 

I remember writing a one-off QL-for-QL query that tried to check for framework coverage.   
In that query I ended up also parsing QL comments in order to find libraries that are not modeled using a `moduleImport` call. (E.g. for libraries that is only used through a global variable).     
But we can look into that if we end up using `FrameworkCoverage.ql` and someone asks why the query is missing something.

asgerf:
Thanks for the review @erik-krogh!

PR #8731, opened 2022-04-13
Python: Delete old dataflow readme

RasmusWL:
Unless there is something in there that you want to keep @yoff?


yoff (approved):
Thanks, I have looked at this a few times and found it out of place.

This file is mostly superseded by the documentation. I have transferred the relevant items from the TODO list to an issue, so this file can go now.

PR #8732, opened 2022-04-13
Python: Don't re-export `python` under `DataFlow::`

RasmusWL:
I have been trying to do this a few times before, but this problems keeps coming back :disappointed: so now I added a test to ensure it doesn't.

I think the actual test might not be needed right now, since the code won't compile if you do this again due to the `Unit` being required from the data-flow implementation.

The previous times we've done this we've agreed that this could potentially be a breaking change, but that in reality it shouldn't be a problem, and should just improve code-completions.

Draft PR to check whether this breaks anything else.


yoff (approved):
LGTM
I think it is wise to keep the test, even if it is not strictly needed at the moment.

PR #8733, opened 2022-04-13
Python: Split `DataFlowPrivate`

RasmusWL:
As part of preparing for type-tracking based call-graph, it seemed nice to move the current points-to based implementation into its own file. I also did this for some other parts of our current data-flow implementation.


yoff (approved):
Thanks for this, I have wanted to do something like this for a while.
I think `DataFlowDispathPointsTo` is well named, it signals that it may be deprecated soon due to use of sketchy technology. The new implementation can just be named `DataFlowDispatch` once it is ready.

PR #8734, opened 2022-04-13
Docs: Fix typo and formatting in "Non-extending subtypes"

henrymercer:
- Fix typo `select any(Foo f) would yield bar` -> `select any(Foo f).foo() would yield bar`
- Fix inline code formatting
- Change `foo_method` to `fooMethod` to follow QL style guide


owen-mc (approved):

annarailton:
Ohhhh so that's where this documentation lives. 

Thank you @henrymercer!

PR #8735, opened 2022-04-13
Swift: dbscheme generator

redsun82:
This patch introduces the basic infrastructure of the code generation
suite and the `dbscheme` generator.

Notice that the checked in `schema.yml` should reflect swift 5.6 but
might need some tweaking.

Closes https://github.com/github/codeql-c-team/issues/979


AlexDenisov:
Let's keep the `answer_to_life_the_universe_and_everything` for now? :)

```
my_first.trap, 1: com.semmle.util.exception.CatastrophicError: The DB scheme does not contain an entry for the relation answer_to_life_the_universe_and_everything
```

AlexDenisov (approved):

AlexDenisov (approved):

swift/codegen/schema.yml
```diff
@@ -0,0 +1,997 @@
...
+  _dir: stmt
+
+RepeatWhileStmt:
+  _extends: LabeledStmt
+  cond: Expr
+  body: Stmt
+
+SwitchStmt:
+  _extends: LabeledStmt
+  subject_expr: Expr
```
MathiasVP, line 851: I'd prefer to call this `expr` (which is what it's called in [C++](https://github.com/github/codeql/blob/main/cpp/ql/lib/semmle/code/cpp/stmts/Stmt.qll#L1488), [C#](https://github.com/github/codeql/blob/main/csharp/ql/lib/semmle/code/csharp/Stmt.qll#L161), [Java](https://github.com/github/codeql/blob/b433f08cef9038d60e2bdf50bfedaf112f35d3f6/java/ql/lib/semmle/code/java/Statement.qll#L397) and [JavaScript](https://github.com/github/codeql/blob/main/javascript/ql/lib/semmle/javascript/Stmt.qll#L617)).

MathiasVP (commented):
A few comments related to the naming of certain columns. Also, keep in mind that I changed some of the names in https://github.com/github/semmle-code/pull/42309 already (which needs to be ported to this PR).

AlexDenisov (approved):

PR #8703, opened 2022-04-08
Dataflow: Revert support for flow-state based in-/out-barriers

aschackmull:
Resolves https://github.com/github/codeql/pull/8435#issuecomment-1087530306

PR #8736, opened 2022-04-13
C++: New query for CWE-611 / XML External Entity Expansion (XXE)

geoffw0:
New query for [CWE-611](https://cwe.mitre.org/data/definitions/611.html) / XML External Entity Expansion (XXE) in C++.  The query is _far_ from complete in terms of libraries supported, the aim of this PR is a working query supporting just one library (`Xerces-C++`), in fact, just one interface (`XercesDOMParser`) within it.  The downside of this approach is that real-world results are currently extremely thin, possibly non-existent in open-source code.  We will need to support more libraries and more popular libraries.

I'd really appreciate a review by @rdmarsh2 (who did some early work on this), and someone who has more XML knowledge than myself (perhaps @smowton or @lcartey ???).  I'm not sure this is checking everything that needs to be checked to make the XML object secure, e.g. I have conflicting sources about whether a `SecurityManager` needs to be attached, or a `setEntityExpansionLimit` set to prevent certain kinds of attack?

TODO:
 - [x] query will probably need repairing for https://github.com/github/codeql/pull/8703
 - [x] check performance; I think it could be formulated without using flow states if they cause problems (though I prefer to use them) --- performance seems good with a warmed up cache, comparable to other IR dataflow queries without (local, kamalio)
 - [x] docs review (note the `.qhelp` is very similar to one for an existing Java query)
 - [ ] run DCA


rdmarsh2:
Looks good to me. I particularly like the flow state transformer pattern.

>  I'm not sure this is checking everything that needs to be checked to make the XML object secure, e.g. I have conflicting sources about whether a SecurityManager needs to be attached, or a setEntityExpansionLimit set to prevent certain kinds of attack?

My understanding is that either a SecurityManager needs to be attached or a setEntityExpansionLimit needs to be set, but it's OK to use only one of them.

MathiasVP:
> * [x]  check performance; I think it could be formulated without using flow states if they cause problems (though I prefer to use them) --- performance seems good with a warmed up cache, comparable to other IR dataflow queries without (local, kamalio)
> *

FWIW, flow states are a very lightweight mechanism that I wouldn't expect to cause any performance problems. Use as many of them as you'd like to!

cpp/ql/src/Security/CWE/CWE-611/XXE.ql
```diff
@@ -0,0 +1,203 @@
...
+import semmle.code.cpp.ir.dataflow.DataFlow
+import DataFlow::PathGraph
+import semmle.code.cpp.ir.IR
+
+/**
+ * A flow state representing a possible configuration of an XML object.
+ */
+abstract class XXEFlowState extends string {
+  bindingset[this]
+  XXEFlowState() { any() } // required constructor
```
MathiasVP, line 25:
```suggestion
  XXEFlowState() { any() } // required characteristic predicate
```

cpp/ql/src/Security/CWE/CWE-611/XXE.ql
```diff
@@ -0,0 +1,203 @@
...
+/**
+ * An `Expr` that changes the configuration of an XML object, transforming the
+ * `XXEFlowState` that flows through it.
+ */
+abstract class XXEFlowStateTranformer extends Expr {
+  /**
+   * Gets the flow state that `flowstate` is transformed into.
+   *
+   * Due to limitations of the implementation the result must always map to
+   * itself, that is, it must be that:
```
MathiasVP, line 37: An alternative formulation of this is that the transformer must be idempotent.

cpp/ql/src/Security/CWE/CWE-611/XXE.ql
```diff
@@ -0,0 +1,203 @@
...
+ */
+class AbstractDOMParserClass extends Class {
+  AbstractDOMParserClass() { this.hasName("AbstractDOMParser") }
+}
+
+/**
+ * The `XercesDOMParser` class.
+ */
+class XercesDOMParserClass extends Class {
+  XercesDOMParserClass() { this.hasName("XercesDOMParser") }
```
MathiasVP, line 56: Should we restrict `XercesDOMParser` to some specific namespace? Or does the class live in the global namespace (so that we should use `hasGlobalName` instead)?

cpp/ql/src/Security/CWE/CWE-611/XXE.ql
```diff
@@ -0,0 +1,203 @@
...
+}
+
+/**
+ * Gets a valid flow state for `XercesDOMParser` flow.
+ *
+ * These flow states take the form `XercesDOM-A-B`, where:
+ *  - A is 1 if `setDisableDefaultEntityResolution` is `true`, 0 otherwise.
+ *  - B is 1 if `setCreateEntityReferenceNodes` is `true`, 0 otherwise.
+ */
+predicate encodeXercesDOMFlowState(string flowstate, int a, int b) {
```
MathiasVP, line 66: Could we call the `a` and `b` parameters something more descriptive? Perhaps `a` could be `defaultEntityResolutionIsDisabled` and `b` could be `createEntityReferenceNodesIsSet`? And should these be booleans instead?

cpp/ql/src/Security/CWE/CWE-611/XXE.ql
```diff
@@ -0,0 +1,203 @@
...
+
+  final override XXEFlowState transform(XXEFlowState flowstate) {
+    exists(int a, int b |
+      encodeXercesDOMFlowState(flowstate, a, b) and
+      (
+        newValue.getValue().toInt() = 1 and // true
+        encodeXercesDOMFlowState(result, 1, b)
+        or
+        not newValue.getValue().toInt() = 1 and // false or unknown
+        encodeXercesDOMFlowState(result, 0, b)
```
MathiasVP, line 109: Two comments about this:
1. Collapsing the `false` and `unknown` cases into one means that we could have FPs if the value is set to a value that we can't deduce, right?

2. Would we catch more good cases by using local dataflow to find all the values that flows into `newValue`? For instance, currently, we would claim that  this is bad, I think:
```cpp
XercesDOMParser* p = ...;
p->setDisableDefaultEntityResolution(complex_condition ? true : false);
```

cpp/ql/src/Security/CWE/CWE-611/XXE.ql
```diff
@@ -0,0 +1,203 @@
...
+  }
+}
+
+/**
+ * The `AbstractDOMParser.parse` method.
+ */
+class ParseFunction extends Function {
+  ParseFunction() {
+    this.getDeclaringType() instanceof AbstractDOMParserClass and
+    this.hasName("parse")
```
MathiasVP, line 153: This sequence of joins has given us performance issues in the past. For this reason, we introduced [Function.getClassAndName()](https://github.com/github/codeql/blob/main/cpp/ql/lib/semmle/code/cpp/Function.qll#L372). Could we use that here?

MathiasVP (commented):
This looks great! I have a few comments that might be worth addressing while we're waiting for DCA to return anyway.

cpp/ql/src/Security/CWE/CWE-611/XXE.ql
rdmarsh2: It lives in a namespace determined by the XERCES_CPP_NAMESPACE macro, which is defined by the user rather than the library.

cpp/ql/src/Security/CWE/CWE-611/XXE.ql
```diff
@@ -0,0 +1,203 @@
...
+
+import cpp
+import semmle.code.cpp.ir.dataflow.DataFlow
+import DataFlow::PathGraph
+import semmle.code.cpp.ir.IR
+
+/**
+ * A flow state representing a possible configuration of an XML object.
+ */
+abstract class XXEFlowState extends string {
```
jketema, line 23: Is there a particular reason you use `extends string` here and not `extends DataFlow::FlowState`?

cpp/ql/src/Security/CWE/CWE-611/XXE.ql
geoffw0: I've updated the QLDoc to use this term.

cpp/ql/src/Security/CWE/CWE-611/XXE.ql
geoffw0: Good idea (done).

cpp/ql/src/Security/CWE/CWE-611/XXE.ql
geoffw0: You're right, `extends DataFlow::FlowState` is better (updated).

cpp/ql/src/Security/CWE/CWE-611/XXE.ql
geoffw0: Done.  The parameter names were references to `a` and `b` in the QLDoc above, motivated by explaining the form of the string simply as `XercesDOM-A-B`. I was worried that renaming the parameters would be confusing in relation to the QLDoc, but having tried it I think its OK.

cpp/ql/src/Security/CWE/CWE-611/XXE.ql
geoffw0: > 1. Collapsing the `false` and `unknown` cases into one means that we could have FPs if the value is set to a value that we can't deduce, right?

Indeed, there's a design choice here of whether we want FPs or FNs resulting from non-constant expressions we can't determine.  Usually we would choose FNs over FPs, but in this case I feel the point of the query is to be _confident_ that the XML object is set up correctly - and that requires rejecting unknown expressions.

I may well be persuaded otherwise when we have more real world results.  We don't yet have many to judge this issue on.

> 2. Would we catch more good cases by using local dataflow to find all the values that flows into `newValue`? For instance, currently, we would claim that this is bad, I think:

Good idea.  I've added it to the TODO list on the XXE query ticket.

geoffw0:
> My understanding is that either a SecurityManager needs to be attached or a setEntityExpansionLimit needs to be set, but it's OK to use only one of them.

Thanks, currently the query checks neither, I will put checking for (either of) these the list of follow-up work.

geoffw0:
@github/docs-content-codeql please could I have a docs review of this PR, in particular the new `XXE.qhelp` file.

MathiasVP:
> @github/docs-content-codeql please could I have a docs review of this PR, in particular the new `XXE.qhelp` file.

FWIW, you need to add the `ready-for-doc-review` label to request a docs review. I've gone ahead and done that for you.

cpp/ql/src/Security/CWE/CWE-611/XXE.ql
```diff
@@ -0,0 +1,212 @@
+/**
+ * @name External Entity Expansion
```
hubwriter, line 2: Possibly add XML here to match the definition of XXE in the qhelp file?

```suggestion
 * @name XML external entity expansion
```

cpp/ql/src/Security/CWE/CWE-611/XXE.ql
```diff
@@ -0,0 +1,212 @@
...
+    // when the flowstate is transformed at a call node, block the original
+    // flowstate value.
+    node.asConvertedExpr().(XXEFlowStateTranformer).transform(flowstate) != flowstate
+  }
+}
+
+from XXEConfiguration conf, DataFlow::PathNode source, DataFlow::PathNode sink
+where conf.hasFlowPath(source, sink)
+select sink, source, sink,
+  "This $@ is not configured to prevent an External Entity Expansion (XXE) attack.", source,
```
hubwriter, line 211: As per previous comment:

```suggestion
  "This $@ is not configured to prevent an XML external entity (XXE) attack.", source,
```

cpp/ql/src/Security/CWE/CWE-611/XXE.qhelp
```diff
@@ -0,0 +1,57 @@
+<!DOCTYPE qhelp PUBLIC "-//Semmle//qhelp//EN" "qhelp.dtd">
+<qhelp>
+
+<overview>
+<p>
+Parsing untrusted XML files with a weakly configured XML parser may lead to an
+XML External Entity (XXE) attack. This type of attack uses external entity references
```
hubwriter, line 7: I'd suggest losing the caps, especially as they don't correspond to the initialism.

```suggestion
XML external entity (XXE) attack. This type of attack uses external entity references
```

hubwriter (dismissed):
Looks good. Just a few very small editorial comments.
👍

cpp/ql/src/Security/CWE/CWE-611/XXE.ql
geoffw0: (I'll make a follow-up commit with the corresponding changes in the `.expected` file for the tests)

hubwriter (approved):

cpp/ql/src/Security/CWE/CWE-611/XXE.ql
```diff
@@ -0,0 +1,212 @@
...
+      f.hasName("setDisableDefaultEntityResolution") and
+      this = call.getQualifier() and
+      newValue = call.getArgument(0)
+    )
+  }
+
+  final override XXEFlowState transform(XXEFlowState flowstate) {
+    exists(int disabledDefaultEntityResolution, int createEntityReferenceNodes |
+      encodeXercesDOMFlowState(flowstate, disabledDefaultEntityResolution,
+        createEntityReferenceNodes) and
```
MathiasVP, line 114: We're not really using the `disabledDefaultEntityResolution` variable for anything in this predicate. Could we just replace it with `_`?

cpp/ql/src/Security/CWE/CWE-611/XXE.ql
```diff
@@ -0,0 +1,212 @@
...
+      f.hasName("setCreateEntityReferenceNodes") and
+      this = call.getQualifier() and
+      newValue = call.getArgument(0)
+    )
+  }
+
+  final override XXEFlowState transform(XXEFlowState flowstate) {
+    exists(int disabledDefaultEntityResolution, int createEntityReferenceNodes |
+      encodeXercesDOMFlowState(flowstate, disabledDefaultEntityResolution,
+        createEntityReferenceNodes) and
```
MathiasVP, line 147: Similarly here: Could we replace `createEntityReferenceNodes` with an underscore?

MathiasVP (dismissed):
The DCA run looks good, and I think this PR is ready to go 🎉. I only have two small stylistic comments that you can choose to ignore if you want.

cpp/ql/src/Security/CWE/CWE-611/XXE.ql
geoffw0: Done.

MathiasVP (approved):
:shipit:

geoffw0:
> The DCA run looks good

Agreed.  Merging...

PR #8737, opened 2022-04-14
Ruby: Model the posix-spawn gem

hmac:
This gem exists primarily to provide methods that spawn subprocesses. We
model these as `SystemCommandExecution`s.


hmac:
https://github.com/github/codeql/pull/8635 has some more complex logic for identifying which arguments are passed to the spawned process - possibly we should try to share some of that to get the same benefits here.

ruby/ql/lib/codeql/ruby/frameworks/PosixSpawn.qll
```diff
@@ -0,0 +1,81 @@
...
+  /**
+   * A call to `POSIX::Spawn.spawn` or a related method.
+   */
+  class SystemCall extends SystemCommandExecution::Range, DataFlow::CallNode {
+    SystemCall() {
+      exists(API::Node spawn | spawn = API::getTopLevelMember("POSIX").getMember("Spawn") |
+        this =
+          posixSpawnModule()
+              .getAMethodCall(["spawn", "fspawn", "popen4", "pspawn", "system", "_pspawn", "`"])
+      )
```
nickrolfe, line 48: You don't need the `spawn` variable:
```suggestion
      this =
        posixSpawnModule()
            .getAMethodCall(["spawn", "fspawn", "popen4", "pspawn", "system", "_pspawn", "`"])
      )
```

ruby/ql/lib/codeql/ruby/frameworks/PosixSpawn.qll
```diff
@@ -0,0 +1,81 @@
...
+    // spawn(env, cmd)
+    // spawn(env, cmd, opts)
+    //
+    // env and opts will be hashes. We over-approximate by assuming the argument
+    // is shell interpreted unless there is another argument with a string
+    // constant value.
+    override predicate isShellInterpreted(DataFlow::Node arg) {
+      exists(int n, int m, DataFlow::Node otherArg |
+        this.argument(arg, n) and
+        this.argument(otherArg, m) and
```
nickrolfe, line 69: Could be simplified to the following, witihout loss of meaning
```suggestion
      exists(DataFlow::Node otherArg |
        this.argument(arg, _) and
        this.argument(otherArg, _) and
```

And you probably want to add `and not arg = otherArg`.

ruby/ql/lib/codeql/ruby/frameworks/PosixSpawn.qll
```diff
@@ -0,0 +1,81 @@
...
+        otherArg.asExpr().getConstantValue().isString(_)
+      )
+    }
+
+    private predicate argument(DataFlow::Node arg, int n) {
```
nickrolfe, line 74: And if you take my suggestion above, then it could be simplified more by removing this `n` param.

nickrolfe (commented):
I have some nitpicks, but otherwise looks great.

nickrolfe (approved):
LGTM 👍

PR #8586, opened 2022-03-29
ATM: Refactors EndpointFeatures.qll and add two new features 

esbena:
This refactors EndpointFeatures.qll and introduces two new features. The new features are not used in production, but the refactoring may introduce performance changes (evaluation pending).

The refactoring moves each endpoint feature to its own class which can be tested independently of active `FeaturizationConfig::getAnEndpointToFeaturize` and other abstract classes that are included conditionally.

The new features are not strictly required for the refactoring part of this PR, but they did inform the refactoring process a lot so I have kept them in since they wont have any runtime effects currently. 
The new features address shortcomings in existing features by being able to produce non-trivial values in many more cases.
A separate discussion of when to enable the features is required.

I'll sleep on this before opening the PR for review.

PR #8738, opened 2022-04-14
JS: ATM: new feature for imports in the file of an endpoint

kaeluka:
This adds a new feature that makes the imports in a file of an endpoint visible to the model.

## Example:

```javascript

import * as FS from 'fs';
//                  ^^^^
import { test } from 'test';
//                   ^^^^^^

const pg = require('pg');
//                 ^^^^


...
```

In this file, any endpoint will have the value `fs pg test` for the `fileImports` feature.

## Depdencies

This depends on https://github.com/github/codeql/pull/8586.

Review starting at the commit "ATM: new feature to list all imports in an endpoint's file".


tiferet:
Small GitHub tip: you can select the base of a PR to be the branch your branch came from (the branch of https://github.com/github/codeql/pull/8586), and then the PR will only show the differences between this branch and that one.

kaeluka:
Thanks for that tip, @tiferet :D I'll actually recreate the PR with this setting!

PR #8748, opened 2022-04-15
Java: Avoid higher-numbered dataflow configs that depend on lower-numbered ones

smowton:
While trying to enumerate sinks as perceived by various queries, https://github.com/github/codeql/discussions/8742 noticed that some of our qll files that define sinks can't be imported concurrently. This turned out to be because some QLLs defy the usual pattern that lower-numbered dataflow configs are defined in terms of higher ones, not the reverse. By flipping the dataflow instance numbers here, it becomes possible to import all of the `java/ql/lib/semmle/code/java/security` directory concurrently.

### DIFF
--- a/config/identical-files.json
+++ b/config/identical-files.json
@@ -51,6 +51,7 @@
     "csharp/ql/lib/semmle/code/csharp/dataflow/internal/tainttracking5/TaintTrackingImpl.qll",
     "java/ql/lib/semmle/code/java/dataflow/internal/tainttracking1/TaintTrackingImpl.qll",
     "java/ql/lib/semmle/code/java/dataflow/internal/tainttracking2/TaintTrackingImpl.qll",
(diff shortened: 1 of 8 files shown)


intrigus-lgtm:
> the usual pattern that lower-numbered dataflow configs are defined in terms of higher ones, not the reverse.

Is this documented somewhere or would it make sense to document this somewhere? (I did not know this pattern before)

aschackmull:
> > the usual pattern that lower-numbered dataflow configs are defined in terms of higher ones, not the reverse.
> 
> Is this documented somewhere or would it make sense to document this somewhere? (I did not know this pattern before)

It's not documented, and we don't intend to, as it will hopefully soon be a thing of the past.

aschackmull (approved):

smowton:
Ignored qldoc complaints since they (a) are shared with existing TaintTracking instances and (b) don't really matter, being requests for a file-level doc block in a file that declares one, documented, entity.

PR #8751, opened 2022-04-15
ML: add .gitkeep to resources dir in which ML models are to be found

jhelie:
fixes https://github.com/github/ml-ql-adaptive-threat-modeling/issues/1847


annarailton:
This just means the `qlpack.yml` reference always points to an existing file, right?

henrymercer (approved):
This makes it slightly easier for contributors to run ML-powered queries against custom ML models, since the folder to drop ML models into will already exist in the checkout.

PR #8760, opened 2022-04-16
Clarify `min`, `max` and `rank` documentation

Marcono1234:
Currently the documentation sounds like the `<expression>` must always be "of numeric type or of type `string`", however that is incorrect, as also described correctly by the [full language specification](https://codeql.github.com/docs/ql-language-reference/ql-language-specification/#aggregations):
> The type of the expression in a `max`, `min` or `rank` aggregation **without an ordering directive expression** must be an orderable type

(emphasis mine)

For example the following is valid:
```codeql
import java

select min(File f | f.getExtension() = "java" | f order by f.getTotalNumberOfLines())
```
```codeql
import java

select rank[5](File f | f.getExtension() = "java" | f order by f.getTotalNumberOfLines())
```


aschackmull (approved):
LGTM, thanks!

PR #8761, opened 2022-04-16
Java: Add `Callable.getErasureStringSignature()`

Marcono1234:
I assume the intention behind `Callable.getStringSignature()` is to provide a concise representation of the signature. However, there are two problems with the representation of parameter types:
- It depends on `Type.toString()` which has no guaranteed format
- For parameterized types, it includes all type arguments
  - This makes usage quite verbose, e.g. `doSomething(Function<? super Integer,? extends String>)`
  - The type arguments are separated only by a comma, but not by a space (unlike the callable parameter types)

This defeats the purpose of `getStringSignature()` a bit. This could be solved by adjusting the `getStringSignature()` implementation, but I chose to instead add a separate predicate called `getErasureStringSignature()`.

Open questions:
- Should the existing `getStringSignature()`, `hasStringSignature(string)` and `paramsString()` be deprecated?
-  With the new `getErasureStringSignature()`, might the erasure for type variables be too error-prone? E.g. you have to know how the erasure is determined and check the bounds for the variable. On the other hand, the question is to what extend the name of a type variable can be considered an implementation detail (which might change between versions).

Any feedback is appreciated!

### DIFF
--- a/java/ql/lib/semmle/code/java/Member.qll
+++ b/java/ql/lib/semmle/code/java/Member.qll
@@ -189,9 +189,16 @@ class Callable extends StmtParent, Member, @callable {
    * The format of the string is `<name><params>`, where `<name>` is the result of
    * the predicate `getName()` and `<params>` is the result of `paramsString()`.
    * For example, the method `void printf(java.lang.String, java.lang.Object...)`
-   * has the string signature `printf(String, Object[])`.
+   * has the string signature `printf(String, Object[])`. For parameters of
+   * parameterized type all their type arguments are included, for example,
+   * `Function<? super String,? extends Integer>` (note that here a comma without
+   * space separates the arguments).
    *
-   * Use `getSignature` to obtain a signature including fully qualified type names.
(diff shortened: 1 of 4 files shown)


java/ql/lib/semmle/code/java/Member.qll
```diff
@@ -201,34 +208,60 @@ class Callable extends StmtParent, Member, @callable {
...
    * For example, the method `void printf(java.lang.String, java.lang.Object...)`
-   * has the params string `(String, Object[])`.
+   * has the params string `(String, Object[])`. For parameters of parameterized type
+   * all their type arguments are included, for example,
+   * `Function<? super String,? extends Integer>` (note that here a comma without space
+   * separates the arguments).
    */
-  pragma[nomagic]
   string paramsString() {
-    exists(int n | n = this.getNumberOfParameters() |
```
Marcono1234, line 208: Have rewritten this to use `concat`. The public Git history did not indicate why this was previously implemented this way. So if there are any internal reasons for this, I can revert the change if you want.

PR #8773, opened 2022-04-19
JS: promote `js/resource-exhaustion` out of experimental

erik-krogh:
I've reintroduced the number sinks. Some of those sinks are only problematic when the tainted value is a number.
However, most of those sinks are not type-overloaded, so we can safely assume that the tainted value is a number if it reaches the sink.
An exception (that is type-overloaded) is new Array(sink), however it seems to be precise enough by using the filters that are in the query.
I removed the new Buffer(sink), because that seemed to be used mostly with non-numbers, and the results for that sink were therefore noisy.

I removed the client-side sources, because I didn't like the client-side results that we found. 

---- 

I added `.length` as a sanitizer-step.  
Most of the time an attacker needs to construct something with a large length in order to be able to taint `.length`.  
And if an attacker can do that in a cheap way, then it's already game over, so there is no point in reporting those.  

---- 

Some example results: https://lgtm.com/query/2868226586233865382/


esbena (approved):
I think this has gone through docreview once before, as we otherwise wouldn't have done the `revert` dance. So if the docs are unchanged, we can merge as is.

erik-krogh:
> I think this has gone through docreview once before, as we otherwise wouldn't have done the `revert` dance. So if the docs are unchanged, we can merge as is.

They are. I copy-pasted the old docstrings from the revert.  
But apparently I accidentally added an extra `<p>` while doing so.

esbena (dismissed):

erik-krogh:
[The evaluation looks good](https://github.com/github/codeql-dca-main/tree/data/erik-krogh/resourceExhaustion/reports).  

Two new results where the source is benign (build-system and internal IPC channel respectively).

esbena (approved):

PR #8775, opened 2022-04-19
CPP: PAM Authorization Bypass

porcupineyhairs:
This PR is similar to my other PRs for
[Python](https://github.com/github/codeql/pull/8595) and
[Golang](https://github.com/github/codeql-go/pull/709).

This PR aims to detect instances were an initiated PAM Transaction invokes the `pam_authenticate` method but does not call the `pam_acct_mgmt` method. This is bad as a call to `pam_authenticate` only verifies the users credentials. It does not check if the user account is still is a valid state.

If only a call to `pam_authenticate` is used to verify the user, a user with an expired account password would still be able to login. This can be prevented by calling the `pam_acct_mgmt` function after a `pam_authenticate` function.


geoffw0 (commented):
This query looks great.  I appreciate the effort you've gone to providing tests, documentation and example code to support it.

A couple of real-world results: https://lgtm.com/query/6091759195696304116/

porcupineyhairs:
@geoffw0 Changes done!

geoffw0 (approved):
LGTM.

geoffw0:
Looks like this should've been merged.

porcupineyhairs:
@geoffw0 Thanks for the quick turnaround.

PR #8777, opened 2022-04-19
Ruby: make PostUpdateNode public

nickrolfe:


ruby/ql/lib/codeql/ruby/dataflow/internal/DataFlowPrivate.qll
```diff
@@ -820,24 +820,15 @@ string ppReprType(DataFlowType t) { result = t.toString() }
...
- * an update to the field.
- *
- * Nodes corresponding to AST elements, for example `ExprNode`, usually refer
- * to the value before the update.
- */
-abstract class PostUpdateNode extends Node {
-  /** Gets the node before the state update. */
-  abstract Node getPreUpdateNode();
+module PostUpdateNode {
+  abstract class Range extends Node {
```
hvitved, line 824: I think we should call it `PostUpdateNodeImpl` instead for consistency.

hvitved (changes requested):
One small suggestion, otherwise LGTM.

hvitved (approved):

PR #8781, opened 2022-04-20
Python dataflow: flow summaries restart

yoff:
This demonstrates that flow summaries can be based on API graphs and all our existing tests can pass 🎉 

I have trimmed down the changes needed to get to this state, and given detailed explanations of the tricky transformations needed to avoid non-monotonic recursion.

For a moment, the commit history was beautiful.

Then main moved on and more changes needed to be made, now it is a bit longer and more hairy.
Also, there is a branch where it is even more hairy, cutting up some of the inheritance hierarchies in order to achieve 
`SourceParameterNode extends LocalSourceNode`, but I suggest we leave that for now. `SourceParameterNode` is no longer something non-internal development should be concerned about.

Names have become a bite nicer and the impact on user code and on `.expected`-files has been minimised.

We might still do more test, showing off the capabilities, especially the use from MaD which is not demonstrated at the moment. However, simply having it compile and the basic tests pass seems like a worthy milestone at this point.


asgerf:
I've only done a superficial glance at this, but I'd like to talk about the issue with `getACall2`. This is a fundamental issue when using API graphs and flow summaries together, and will likely affect all dynamic languages attempting to use flow summaries. I'd like to get a bit more clarity on what the problem is and what solution to use (ideally we'll all end up using the same solution).

The root of the problem is that the following cycle is created between call graph construction and API graphs:
Screenshot 2022-04-20 at 12 36 24

Unless we want API graphs to be mutually recursive with type-tracking steps and call construction, we'll need to break the cycle somehow. As far as I can tell, you've broken the cycle like this:

Screenshot 2022-04-20 at 12 41 11

This means anything not depending on API graphs can contribute to `getACall2`, and things depending on API graphs can use `getACall`. Type tracking steps will only be discovered for `getACall2`, which makes it more suitable for things like standard library modelling.

I believe Ruby's solution so far is to make type-tracking steps explicitly depend on non-library callables only. The downside here is that none of their standard library models are visible to type-tracking. (@hvitved @aibaars do you agree here?)

@yoff Since I'm unfamiliar with the Python libraries I'm not 100% sure my interpretation above is an accurate description of what you've done here. Could you please check if this is the case? If so, I think it's a fine solution provided we can come up with names that aren't confusing.

yoff:
Force pushed because I could see that my amended commit message to the last commit had not been pushed.

yoff:
The diagram looks correct to me. Is the difference with Ruby just that we have not (yet?) written any standard library models using flow summaries? Or is there a more fundamental difference?

python/ql/lib/semmle/python/dataflow/new/internal/DataFlowPrivate.qll
```diff
@@ -886,10 +886,26 @@ predicate nodeIsHidden(Node n) {
...
+  creation.asExpr() = c.(DataFlowLambda).getDefinition()
+  or
+  // normal function
+  kind = kind and
+  exists(Call call, Name f, FunctionDef def |
+    f = call.getAnArg() and
+    def.getDefinedFunction().getName() = f.getId() and
+    // c.getCallableValue() = def.getDefinedFunction().getDefinition() and
+    c.getName() = f.getId()
+  )
```
tausbn, line 939: This may have been fixed in a later commit, but as QL-for-QL helpfully points out, `creation` isn't restricted by this disjunct.

python/ql/lib/semmle/python/dataflow/new/internal/DataFlowDispatchPointsTo.qll
```diff
@@ -421,22 +421,24 @@ class LibraryCallableValue extends DataFlowCallable, TLibraryCallable {
...
 newtype TDataFlowCall =
-  TFunctionCall(CallNode call) { call = any(FunctionValue f).getAFunctionCall() } or
-  /** Bound methods need to make room for the explicit self parameter */
-  TMethodCall(CallNode call) { call = any(FunctionValue f).getAMethodCall() } or
-  TClassCall(CallNode call) { call = any(ClassValue c | not c.isAbsent()).getACall() } or
+  /**
+   * Includes function calls, method calls, class calls and library calls.
+   * All these will be associated with a `CallNode`.
+   */
+  TNonSpecialCall(CallNode call) or
```
tausbn, line 435: Something about `NonSpecial...` rubs me the wrong way. Could we use a positive formulation instead? I'm tempted to suggest `CallNodeCall`, though that is a bit of a mouthful. (However, it does precisely say that these are calls that happen at call nodes.)

Alternatively, since we have `SpecialCall`, perhaps its counterpart should be `NormalCall`?

python/ql/lib/semmle/python/dataflow/new/internal/DataFlowDispatchPointsTo.qll
```diff
@@ -318,24 +319,30 @@ class DataFlowCallable extends TDataFlowCallable {
...
   LibraryCallable asLibraryCallable() { this = TLibraryCallable(result) }
 
   Location getLocation() { none() }
 }
 
+abstract class NonLibraryDataFlowCallable extends DataFlowCallable {
+  /** Gets a callable value for this callable, if one exists. */
+  abstract CallableValue getCallableValue();
+
+  abstract CallNode getACall2();
```
tausbn, line 332: This needs a better name. Even if it is only internal, it certainly doesn't convey what it's actually supposed to do.

python/ql/lib/semmle/python/dataflow/new/internal/DataFlowDispatchPointsTo.qll
```diff
@@ -497,59 +504,69 @@ class NonSpecialCall extends DataFlowSourceCall, TNonSpecialCall {
   override DataFlowCallable getEnclosingCallable() { result.getScope() = call.getNode().getScope() }
 }
 
+abstract class NonLibraryDataFlowSourceCall extends NonSpecialCall {
```
tausbn, line 507: I think this name needs a bit of workshopping. Could we come up with a better description of "non-library"?

python/ql/lib/semmle/python/dataflow/new/internal/DataFlowPublic.qll
```diff
@@ -277,40 +286,58 @@ ExprNode exprNode(DataFlowExpr e) { result.getNode().getNode() = e }
...
+abstract class ParameterNode extends Node {
+  /**
+   * Holds if this node is the parameter of callable `c` at the
+   * (zero-based) index `i`.
+   */
+  abstract predicate isParameterOf(DataFlowCallable c, int i);
+}
+
+class SourceParameterNode extends ParameterNode, CfgNode {
+  //, LocalSourceNode {
```
tausbn, line 298: I'm still a bit ambivalent about these not being local sources, and I would like to understand this a bit more. Is this merely because of negative recursion issues?

tausbn (changes requested):
Thank you for this Herculean effort!

I think in terms of the implementation, this looks pretty solid. I've added a bunch of requests for clarification (and also some requests for renaming stuff). I'll be very happy to discuss this synchronously if that's easier. :slightly_smiling_face:

python/ql/lib/semmle/python/dataflow/new/internal/DataFlowPublic.qll
yoff: Negative recursion is what directly prevents me from inheriting. But I can add that inheritance into the Ruby codebase which basically has the same structure, so I think the solution is to clear up the mutual recursive structure of our predicates, which I fear lacks a bit of discipline at the moment.

python/ql/lib/semmle/python/dataflow/new/internal/DataFlowPrivate.qll
```diff
@@ -282,6 +282,8 @@ module EssaFlow {
       nodeTo = TKwOverflowNode(call, callable) and
       nodeFrom.asCfgNode() = call.getNode().getKwargs().getAFlowNode()
     )
+    or
+    FlowSummaryImpl::Private::Steps::summaryLocalStep(nodeFrom, nodeTo, true)
```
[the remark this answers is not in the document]
yoff: No, there are now two local step relations. One for type trackers and one for the global data flow computation.

python/ql/lib/semmle/python/dataflow/new/internal/DataFlowDispatchPointsTo.qll
```diff
@@ -518,8 +562,34 @@ class SpecialCall extends DataFlowCall, TSpecialCall {
...
+ * A synthesized call inside a callable with a flow summary.
+ *
+ * For example, in
+ * ```python
+ * map(lambda x: x + 1, [1, 2, 3])
+ * ```
+ *
+ * there is a call to the lambda argument inside `map`.
+ */
+class SummaryCall extends DataFlowCall, TSummaryCall {
```
hvitved, line 640: In order for call-backs to work, you need to implement `lambdaCall` and `lambdaCreation` (see https://github.com/github/codeql/blob/main/ruby/ql/lib/codeql/ruby/dataflow/internal/DataFlowPrivate.qll#L890-L931).

python/ql/lib/semmle/python/dataflow/new/internal/DataFlowPrivate.qll
```diff
@@ -2,6 +2,7 @@ private import python
 private import DataFlowPublic
```
hvitved, line 2: I think you are missing `allowParameterReturnInSelf`, see https://github.com/github/codeql/blob/main/ruby/ql/lib/codeql/ruby/dataflow/internal/DataFlowPrivate.qll#L940.

hvitved (commented):
A few drive-by comments. Is the logic actually hooked up yet? I fail to see references to e.g. any of the `FlowSummaryImpl::Private::Steps` predicates. See e.g. `DataFlowPrivate.qll` for ruby for how this is done.

python/ql/lib/semmle/python/dataflow/new/internal/DataFlowDispatchPointsTo.qll
yoff: `lambdaCall` is hiding [here](https://github.com/github/codeql/pull/8781/files#diff-3efed0c245564ccd514237561cd7e9f3a481afee65678c6be4518d98c25eaa75R537-R550). I did not find it at first, because the diff was collapsed, and had [a mild panic](https://github.com/github/codeql/pull/9099) while I thought we were looking at different versions of the code.

`lambdaCreation` is [here](https://github.com/github/codeql/pull/8781/files#diff-80716eaf55e2581b856950d844a39de74e7015eb05fb3fb56ea89662c13e56fdR927-R940).

Also, [the tests](https://github.com/github/codeql/pull/8781/files#diff-76ffbada01b4eb965e68a5348b591926230cf9090e6d1ccdf64935683e60763dR60-R61) show that call-backs do indeed work :-)

yoff:
> A few drive-by comments. Is the logic actually hooked up yet? I fail to see references to e.g. any of the `FlowSummaryImpl::Private::Steps` predicates. See e.g. `DataFlowPrivate.qll` for ruby for how this is done.

They are [mostly in `DataFlowPrivate.qll`](https://github.com/github/codeql/pull/8781/files#diff-80716eaf55e2581b856950d844a39de74e7015eb05fb3fb56ea89662c13e56fdR366-R873) and then [one in `DataFlowUtil.qll`](https://github.com/github/codeql/pull/8781/files#diff-4dec1c684b92d3ebc9760e57f26bb633b8eb34090b32f56736f6273c7a996b36R13-R19). I followed the structure from Ruby closely. The interface does not let you search for these terms inside collapsed files, which is a bit misleading.

python/ql/lib/semmle/python/dataflow/new/internal/DataFlowPrivate.qll
yoff: Thanks! Indeed we just have `none()`.

hvitved:
> They are [mostly in `DataFlowPrivate.qll`](https://github.com/github/codeql/pull/8781/files#diff-80716eaf55e2581b856950d844a39de74e7015eb05fb3fb56ea89662c13e56fdR366-R873) and then [one in `DataFlowUtil.qll`](https://github.com/github/codeql/pull/8781/files#diff-4dec1c684b92d3ebc9760e57f26bb633b8eb34090b32f56736f6273c7a996b36R13-R19). I followed the structure from Ruby closely. The interface does not let you search for these terms inside collapsed files, which is a bit misleading.

Ah, good, sorry for the false alarm. I guess I must have looked at a diff limited to a subset of commits perhaps.

yoff:
> I think in terms of the implementation, this looks pretty solid. I've added a bunch of requests for clarification (and also some requests for renaming stuff). I'll be very happy to discuss this synchronously if that's easier. 🙂

Thank you for the review. I think both renaming and qldocs are warranted. Specifically, when classes are split (in order to only have one participate in a recursion that would otherwise be non-monotonic) the user-facing one should be retaining the old name so as to minimise breaking changes.

python/ql/test/experimental/dataflow/summaries/summaries.py
```diff
@@ -0,0 +1,65 @@
...
+SINK(tainted_mapped[0])  # $ flow="SOURCE, l:-1 -> tainted_mapped[0]"
+
+def explicit_identity(x):
+    return x
+
+tainted_mapped_explicit = map(explicit_identity, [SOURCE])
+SINK(tainted_mapped_explicit[0])  # $ flow="SOURCE, l:-1 -> tainted_mapped_explicit[0]"
+
+tainted_mapped_summary = map(identity, [SOURCE])
+SINK(tainted_mapped_summary[0])  # $ flow="SOURCE, l:-1 -> tainted_mapped_summary[0]"
```
yoff, line 61: @hvitved is this too ambitious? We have to establish that the call-back argument (`identity`) is a summarised callable. I made it work by adding a `getACallback` next to `getACall` to `LibraryCallable` (I first made it work by simply comparing ids).

hvitved: I would expect this to work if you add appropriate logic to `lambdaCreation(ExprNode creation, LambdaCallKind kind, DataFlowCallable c)` where `map` will be the access to `identity` and `c` is the synthesized callable for the flow summary.

python/ql/lib/semmle/python/dataflow/new/internal/DataFlowDispatchPointsTo.qll
```diff
@@ -518,8 +635,125 @@ class SpecialCall extends DataFlowCall, TSpecialCall {
...
+
+/**
+ * A synthesized call inside a callable with a flow summary.
+ *
+ * For example, in
+ * ```python
+ * map(lambda x: x + 1, [1, 2, 3])
+ * ```
+ *
+ * there is a call to the lambda argument inside `map`.
```
RasmusWL, line 658:
```suggestion
 * there is a synthesized call to the lambda argument inside `map`.
```

python/ql/lib/semmle/python/dataflow/new/internal/DataFlowDispatchPointsTo.qll
```diff
@@ -379,90 +434,157 @@ class DataFlowModuleScope extends DataFlowCallable, TModule {
...
+  final override Node getArg(int n) { result = this.getNonLibraryArg(n) }
+
+  abstract DataFlowCallable getNonLibraryCallable();
+
+  final override DataFlowCallable getCallable() { result = this.getNonLibraryCallable() }
 }
 
 /**
- * A call to a function/lambda.
+ * A call to a function.
```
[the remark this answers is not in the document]
RasmusWL: Ahh, I see I need to look more into the comment made by Tom here: https://github.com/github/codeql/pull/8781/files#r868280791

python/ql/lib/semmle/python/dataflow/new/internal/DataFlowDispatchPointsTo.qll
```diff
@@ -518,8 +635,125 @@ class SpecialCall extends DataFlowCall, TSpecialCall {
...
+
 /** Gets a viable run-time target for the call `call`. */
-DataFlowCallable viableCallable(DataFlowCall call) { result = call.getCallable() }
+DataFlowCallable viableCallable(DataFlowSourceCall call) {
+  result = call.getCallable()
+  or
+  exists(LibraryCallable callable |
+    result = TLibraryCallable(callable) and
+    call.getNode().getNode() = callable.getACall()
+  )
```
RasmusWL, line 755: Could this be deleted? Won't it be covered by the first line, and by the implementation of `LibraryCall.getCallable`?

python/ql/lib/semmle/python/dataflow/new/internal/DataFlowDispatchPointsTo.qll
```diff
@@ -533,26 +767,51 @@ class ReturnKind extends TReturnKind {
...
+abstract class OutNode extends Node {
+  /** Gets the underlying call, where this node is a corresponding output of kind `kind`. */
+  abstract DataFlowCall getCall(ReturnKind kind);
+}
+
+private module OutNodes {
+  /**
+   * A data-flow node that reads a value returned directly by a callable.
+   */
+  class ExprOutNode extends OutNode, ExprNode {
```
RasmusWL, line 793: Why not extend CallCfgNode? or at least CfgNode? -- again, I feel like it goes against our practice to drop all the way down to the expression, instead of staying in the control flow layer.

python/ql/lib/semmle/python/dataflow/new/internal/FlowSummaryImplSpecific.qll
```diff
@@ -0,0 +1,191 @@
...
+  )
+}
+
+/**
+ * Gets the summary component for specification component `c`, if any.
+ *
+ * This covers all the Python-specific components of a flow summary.
+ */
+SummaryComponent interpretComponentSpecific(AccessPathToken c) {
+  c = "Argument[_]" and // Ruby has this
```
RasmusWL, line 67: I don't understand the inline comment about Ruby. Should this be deleted? if not, I'd like to understand what it actually means :sweat_smile:

RasmusWL (changes requested):
Overall looks like it could be merged pretty soon :crossed_fingers: :muscle: 

I some things though, most of them minor stuff, but for some of them (especially the lambda stuff), it would be nice to talk things over in-person/zoom.

## General comments:

Can we add ONE production flow-summary? For example for the builtin `reversed` -- just to see where it would go, and how it looks :blush: 

I also see a bit of naming conflict between `NonLibrary` and `Source`, for example in the class hierarchy `NonLibraryNormalCall > NormalCall > DataFlowSourceCall`, so if `Source` is the same as `NonLibrary`, maybe we should stick to just _one_ of the naming schemes?

python/ql/lib/semmle/python/ApiGraphs.qll
```diff
@@ -818,7 +858,7 @@ module API {
     DataFlow::LocalSourceNode trackUseNode(DataFlow::LocalSourceNode src) {
       Stages::TypeTracking::ref() and
       result = trackUseNode(src, DataFlow::TypeTracker::end()) and
-      not result instanceof DataFlow::ModuleVariableNode
+      result instanceof DataFlow::ExprNode
```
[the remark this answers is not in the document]
yoff: Indeed this negation was leading to non-monotonic recursion.

python/ql/lib/semmle/python/ApiGraphs.qll
```diff
@@ -998,7 +1038,7 @@ module API {
     ApiLabel memberFromRef(DataFlow::AttrRef ref) {
       result = member(ref.getAttributeName())
       or
-      not exists(ref.getAttributeName()) and
+      ref.unknownAttribute() and
```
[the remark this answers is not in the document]
yoff: This negation was leading to non-monotonic recursion.

python/ql/lib/semmle/python/dataflow/new/FlowSummary.qll
```diff
@@ -0,0 +1,98 @@
+/** Provides classes and predicates for defining flow summaries. */
+
+import python
+import semmle.python.dataflow.new.DataFlow
+private import internal.FlowSummaryImpl as Impl
+private import internal.DataFlowUtil
+private import internal.DataFlowPrivate
+
+// import all instances below
+private module Summaries {
```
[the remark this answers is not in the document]
yoff: Actually that seems to be [exactly what they do in Ruby](https://github.com/github/codeql/blob/main/ruby/ql/lib/codeql/ruby/dataflow/FlowSummary.qll#L11-L14). I will do that also :-)

python/ql/lib/semmle/python/dataflow/new/TestSummaries.qll
```diff
@@ -0,0 +1,93 @@
+/**
+ * Provides a set of flow summaries to be used for debugging.
+ * Import this file in FlowSummary.qll.
+ */
```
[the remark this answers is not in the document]
yoff: We could delete it. I added it here so I could have summaries available for quick-eval of library code.

python/ql/test/experimental/dataflow/summaries/TestSummaries.qll
```diff
@@ -0,0 +1,88 @@
...
+    preservesValue = true
+    or
+    input = "Argument[0].ReturnValue" and
+    output = "ReturnValue.ListElement" and
+    preservesValue = true
+  }
+}
+
+// Typetracking needs to use a local flow step not including summaries
+// Typetracking needs to use a call graph not including summaries
```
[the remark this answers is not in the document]
yoff: These should no longer be needed here. They should probably go in the document describing our strategy for avoiding non-monotonic recursion, but we have no such thing..

python/ql/lib/semmle/python/dataflow/new/internal/FlowSummaryImplSpecific.qll
```diff
@@ -0,0 +1,191 @@
...
+private import DataFlowPrivate
+private import DataFlowPublic
+private import DataFlowImplCommon
+private import FlowSummaryImpl::Private
+private import FlowSummaryImpl::Public
+private import semmle.python.dataflow.new.FlowSummary as FlowSummary
+
+class SummarizedCallableBase = string;
+
+DataFlowCallable inject(SummarizedCallable c) { result.asLibraryCallable() = c }
```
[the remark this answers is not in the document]
yoff: Ah, sorry, a category theory term seems to have slipped in here. An injection is the opposite of a projection, in this case `asLibraryCallable`. So we are injecting the smaller type `SummarizedCallable` into the bigger type `DataFlowCallable`.
Perhaps `asDataFlowCallable` would be clearer?

python/ql/lib/semmle/python/dataflow/new/internal/DataFlowPrivate.qll
```diff
@@ -326,6 +351,34 @@ predicate runtimeLocalFlowStep(Node nodeFrom, Node nodeTo) {
...
+  // If there is local flow out of a node `node`, we want flow
+  // both out of `node` and any post-update node of `node`.
+  exists(Node node |
+    nodeFrom = update(node) and
+    (
+      importTimeSummaryFlowStep(node, nodeTo) or
+      runtimeSummaryFlowStep(node, nodeTo)
+    )
+  )
+}
```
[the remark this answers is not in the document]
yoff: Nice idea!

python/ql/lib/semmle/python/dataflow/new/internal/DataFlowPublic.qll
```diff
@@ -277,40 +288,56 @@ ExprNode exprNode(DataFlowExpr e) { result.getNode().getNode() = e }
...
+
 /** Gets a node corresponding to parameter `p`. */
-ParameterNode parameterNode(Parameter p) { result.getParameter() = p }
+SourceParameterNode parameterNode(Parameter p) { result.getParameter() = p }
 
 /** A data flow node that represents a call argument. */
-class ArgumentNode extends Node {
-  ArgumentNode() { this = any(DataFlowCall c).getArg(_) }
-
+abstract class ArgumentNode extends Node {
```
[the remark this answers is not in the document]
yoff: I think I simply did not get around to that, and was not forced to by any non-monotonic recursion. But it would likely make for a healthier code base. Is it the same change you will be making or a different one?

python/ql/test/experimental/dataflow/summaries/TestSummaries.qll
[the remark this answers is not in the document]
RasmusWL: a great opportunity to start one :wink:

python/ql/lib/semmle/python/dataflow/new/internal/DataFlowPublic.qll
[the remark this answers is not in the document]
RasmusWL: I looked at what ruby does. Since their definition of ArgumentNode is in `DataFlowPrivate.qll`, it's not as much of a problem. I think it would actually be nice to move more of the internal dataflow stuff to the private .qll file. (but I think it's OK not to change anything in regards to this in this PR)

python/ql/lib/semmle/python/dataflow/new/TestSummaries.qll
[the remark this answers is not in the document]
RasmusWL: > We could delete it.

nice, I think that seems like the right thing to do :+1:

python/ql/lib/semmle/python/dataflow/new/internal/FlowSummaryImplSpecific.qll
yoff: I think I just kept this as an example, I have deleted it now and the tests still pass :-)

python/ql/test/experimental/dataflow/summaries/NormalDataflowTest.ql
```diff
@@ -0,0 +1,3 @@
+import python
```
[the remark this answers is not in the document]
yoff: done

python/ql/lib/semmle/python/dataflow/new/internal/DataFlowDispatchPointsTo.qll
yoff: Yes, you are right! Good spot 👍

python/ql/lib/semmle/python/dataflow/new/internal/DataFlowDispatchPointsTo.qll
yoff: `ExprNode` is a control flow node, but you are right that the charred is defined in terms of expressions. I have amended that.

python/ql/test/experimental/dataflow/summaries/TestSummaries.qll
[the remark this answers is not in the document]
yoff: alright, I have started one..

python/ql/lib/semmle/python/dataflow/new/internal/FlowSummaryImplSpecific.qll
```diff
@@ -0,0 +1,191 @@
...
+
+/** Gets the parameter position of the instance parameter. */
+ArgumentPosition instanceParameterPosition() { none() } // disables implicit summary flow to `this` for callbacks
```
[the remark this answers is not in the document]
yoff: The comment just says that we are not using this feature yet. It is currently used by Java, Go, and Swift (the other languages have the same comment).

yoff:
> ## General comments:
> Can we add ONE production flow-summary? For example for the builtin `reversed` -- just to see where it would go, and how it looks 😊

I put it in `frameworks\Builtins.qll`. It could also go in the `StdLib` folder, I guess. Or perhaps you have a different place that you find natural, feel free to suggest :-)

python/ql/lib/semmle/python/dataflow/new/FlowSummary.qll
```diff
@@ -67,32 +69,44 @@ abstract class SummarizedCallable extends LibraryCallable, Impl::Public::Summari
...
-//       or
-//       kind = "taint" and
-//       preservesValue = false
-//     )
-//   }
-// }
+
+// This gives access to getNodeFromPath, which is not constrained to `CallNode`s
+// as `resolvedSummaryBase` is.
+private import semmle.python.frameworks.data.internal.ApiGraphModels as AGM
```
RasmusWL, line 75: as you say in the commit message, no tests :disappointed: Could we just leave this code commented out until we have time/energy to show that it works as intended? (I might also be persuaded with a promise that you make a followup PR with tests...)

python/ql/lib/semmle/python/dataflow/new/TestSummaries.qll
```diff
@@ -0,0 +1,123 @@
+private import python
```
RasmusWL: image

I've seen this file before... is this not a copy of an existing file? Why do we need it twice? it does not seem to be used.

RasmusWL: ahh, I see this file was deleted again in a later commit.

RasmusWL (changes requested):
I think it would be super nice with a change to the result type of `SummarizedCallable.getACall`. (which seems like a fairly simple change, I can even do it myself if you want me to)

Besides that, I don't think I have requests for any other significant changes.

> I put it in frameworks\Builtins.qll. It could also go in the StdLib folder, I guess. Or perhaps you have a different place that you find natural, feel free to suggest :-)

We have normally put both builtins and standard library modeling in the `Stdlib.qll` file. I guess that is not technically 100% correct. However, on the top of my head, it doesn't seem like we're gaining anything from splitting it into two, so I would suggest you move it into the existing file (but do let me know if I'm overlooking something :shrug:). ((note: if you do want to keep the `Builtins.qll` file, you should add module structure like in all the other files in the frameworks folder)

python/ql/lib/semmle/python/dataflow/new/FlowSummary.qll
yoff: I think leaving it (commented) out is probably the right thing to do. Then a follow-up PR can properly demonstrate what it enables.

python/ql/lib/semmle/python/frameworks/Builtins.qll
```diff
@@ -0,0 +1,24 @@
...
+
+import python
+private import semmle.python.dataflow.new.FlowSummary
+private import semmle.python.ApiGraphs
+
+/** A flow summary for `reversed`. */
+class ReversedSummary extends SummarizedCallable {
+  ReversedSummary() { this = "builtins.reversed" }
+
+  override CallNode getACall() { result = API::builtin("reversed").getACall().getNode() }
```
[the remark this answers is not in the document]
yoff: It is slightly inconsistent for `LibraryCallable` to return a `DatFlow::Node` here, when the other callable return control flow nodes, but I tried it now, see if you like :-)

python/ql/lib/semmle/python/dataflow/new/internal/FlowSummaryImplSpecific.qll
```diff
@@ -1,5 +1,25 @@
 /**
  * Provides Python specific classes and predicates for defining flow summaries.
+ *
+ * Flow summaries are defined for callables that are not extracted.
```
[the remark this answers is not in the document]
yoff: I added something..

python/ql/lib/semmle/python/frameworks/Builtins.qll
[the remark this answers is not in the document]
RasmusWL: yes, I'm happy :+1: Much of it will be changed with the new call-graph anyway :shrug:

python/ql/lib/semmle/python/Frameworks.qll
```diff
@@ -9,6 +9,7 @@ private import semmle.python.frameworks.Aiohttp
 private import semmle.python.frameworks.Aiomysql
 private import semmle.python.frameworks.Aiopg
 private import semmle.python.frameworks.Asyncpg
+private import semmle.python.frameworks.Builtins
```
RasmusWL, line 12:
```suggestion
```

python/ql/lib/semmle/python/dataflow/new/internal/FlowSummaryImplSpecific.qll
```diff
@@ -25,6 +25,10 @@
  *
  * We do not support summaries of special methods via the special methods framework,
  * the summary would have to identify the call.
+ *
+ * We might, while we still extract the standard libraries, want to support flow summaries of
+ * extracted callables, so that we can model part of the standard libraries with flow summaries.
+ * For this to work, we have be careful with the enclosiong callable predicate.
```
RasmusWL, line 31:
```suggestion
 * We might, while we still extract the standard library, want to support flow summaries of
 * extracted callables, so that we can model part of the standard library with flow summaries.
 * For this to work, we have be careful with the enclosing callable predicate.
```

RasmusWL (changes requested):
Minor fixes needed... but almost :sweat_smile:

RasmusWL (dismissed):
:crossed_fingers: for the tests :laughing:

tausbn (commented):
A few minor comments here and there. Otherwise, I think this looks very good!

Regarding the overloading of "source" to mean "in source code", how about "extracted"? This, to me, seems fairly unambiguous...

python/ql/lib/semmle/python/dataflow/new/internal/DataFlowPrivate.qll
```diff
@@ -842,15 +900,38 @@ int accessPathLimit() { result = 5 }
...
+    creation.asExpr() = call.getAnArg() and
+    creation = c.(LibraryCallableValue).getACallback()
+  )
+}
 
 /** Holds if `call` is a lambda call of kind `kind` where `receiver` is the lambda expression. */
-predicate lambdaCall(DataFlowCall call, LambdaCallKind kind, Node receiver) { none() }
+predicate lambdaCall(DataFlowCall call, LambdaCallKind kind, Node receiver) {
+  receiver = call.(SummaryCall).getReceiver() and
+  kind = kind
```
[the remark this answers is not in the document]
yoff: That is correct (and `LambdaCallKind` is currently `Unit`).

python/ql/lib/semmle/python/dataflow/new/internal/DataFlowPublic.qll
yoff: Unfortunately, yes. I do feel we should be able to resolve this, but I had a quick go and it seems there are multiple cycles to untangle, so I would prefer to do this later..

tausbn: Alright. Thanks for checking!

tausbn (approved):
:shipit:

PR #8779, opened 2022-04-20
Revert the stages pattern

erik-krogh:
**WIP:** this PR exists to test and evaluate an experiment.  
It should definitely not be merged yet.

PR #8783, opened 2022-04-20
JS: don't initialize sanitizer-guards in the standard library

erik-krogh:
I did an experiment to [revert the cached stages pattern](https://github.com/github/codeql/pull/8779).  
That gave a whole bunch of QL-for-QL warnings from `ql/abstract-class-import`.  

These warnings are caused by us having sanitizer-guards that were defined in `Xss.qll`.  

I moved things around, including refactoring the old `Xss.qll` where I moved things out into `Query.qll`/`Customizations.qll` files so they match the other queries.  
(Turns out the `js/html-constructed-from-input` query relied on the previous behavior :scream_cat:). 

--- 

The change in results for ATM is caused by: https://github.com/github/codeql/pull/8783/commits/76bf8dec8f8921f57eeec1f770f7ddc728f93916  
The results generally look better (the majority is removal of FPs).    
This PR only affects the `XSS` queries, but the changed ATM results are for `TaintedPath`.   
So the results seem to be a bug caused by importing too many classes.  
So I think we can conclude that this PR is a bug-fix for ATM. 

--- 

[The evaluation looks good](https://github.com/github/codeql-dca-main/tree/data/erik-krogh/jsAbstractBi__nightly-old__code-scanning__1/reports).  
We gained one new result. Which is caused by some barrier-guards no longer falsely limiting the flow.    
Also a slight performance improvement :rocket: 

--- 

I also made some small drive-by fixes of `ql/abstract-class-import` to remove some benign results.


javascript/ql/lib/semmle/javascript/security/dataflow/DomBasedXssCustomizations.qll
```diff
@@ -335,4 +328,16 @@ module DomBasedXss {
 
   /** Gets the flow-label representing tainted values where the prefix is attacker controlled. */
   PrefixString prefixLabel() { any() }
+
+  /**
+   * A sanitizer that blocks the `PrefixString` label when the start of the string is being tested as being of a particular prefix.
+   */
+  abstract class PrefixStringSanitizer extends TaintTracking::SanitizerGuardNode,
+    TaintTracking::LabeledSanitizerGuardNode instanceof StringOps::StartsWith {
```
asgerf, line 339: Is there a need to extend both `SanitizerGuardNode` and `LabeledSanitizerGuardNode`?

javascript/ql/lib/semmle/javascript/security/dataflow/StoredXssCustomizations.qll
```diff
@@ -5,6 +5,11 @@
 
 import javascript
 
+/**
+ * Provides sources, sinks, and sanitizers for reasoning about
+ * cross-site scripting vulnerabilities where the taint-flow passes through a thrown
+ * exception.
+ */
```
asgerf, line 12:
```suggestion
/**
 * Provides default sources, sinks and sanitizers for reasoning about
 * stored cross-site scripting vulnerabilities.
 */
```
Mentions a "thrown exception". I've just suggested the same QLDoc from the file.

javascript/ql/lib/change-notes/2022-04-22-xss-library.md
```diff
@@ -0,0 +1,5 @@
+---
+category: deprecated
+---
+* The `ReflectedXss`, `ReflectedXss`, `StoredXss`, `XssThroughDom`, and `ExceptionXss` modules from `Xss.qll` have been deprecated.
```
asgerf, line 4:
```suggestion
* The `ReflectedXss`, `StoredXss`, `XssThroughDom`, and `ExceptionXss` modules from `Xss.qll` have been deprecated.  
```

asgerf (commented):
Awesome work! 💪

javascript/ql/lib/semmle/javascript/security/dataflow/StoredXssCustomizations.qll
erik-krogh: Right. I accidentally copy-pasted from the wrong place.

javascript/ql/lib/semmle/javascript/security/dataflow/DomBasedXssCustomizations.qll
erik-krogh: Probably not :+1:

erik-krogh: Yep, the tests still pass.

henrymercer (approved):
ML-powered queries 👍

TomBolton:
@henrymercer how do feel if we merge this PR? I would like to use the merge commit on `main` to start preparing a PR to update the QL SHAs in backend (including the worsening commits)

PR #8785, opened 2022-04-20
Ruby: Mention `newtype` constructors in API graph label classes

hvitved:
We may as well give the optimizer these disjointness facts.


hvitved:
> Have you done an evaluation?

Nope, I will start one. I don't expect any changes, but better safe than sorry.

nickrolfe (approved):
I didn't know you could do that! LGTM 👍

PR #8788, opened 2022-04-20
Swift: file extraction

AlexDenisov:
Laying out the foundation for the further PRs.


swift/extractor/Extractor.cpp
```diff
@@ -0,0 +1,101 @@
...
+  std::stringstream ss;
+  ss << "-frontend ";
+  for (auto opt : config.frontendOptions) {
+    ss << std::quoted(opt) << " ";
+  }
+  ss << "\n";
+  trap << "// frontend-options: " << ss.str();
+
+  trap << "#0=*\n";
+  trap << "files(#0, " << std::quoted(srcFilePath.str().str()) << ")\n";
```
AlexDenisov, line 88: This is inaccurate and will move elsewhere once we transfer more code.

swift/extractor/Extractor.cpp
```diff
@@ -0,0 +1,101 @@
...
+    }
+  } else {
+    for (auto s : compiler.getPrimarySourceFiles()) {
+      extractFile(*s);
+    }
+  }
+}
+
+void Extractor::extractFile(swift::SourceFile& file) {
+  if (std::error_code ec = llvm::sys::fs::create_directories(config.trapDir)) {
```
AlexDenisov, line 37: We cannot use stdc++fs, so using to LLVM's fs implementation.

swift/extractor/Extractor.cpp
```diff
@@ -0,0 +1,101 @@
...
+    std::cerr << "Cannot create TRAP directory: " << ec.message() << "\n";
+    return;
+  }
+
+  if (std::error_code ec = llvm::sys::fs::create_directories(config.sourceArchiveDir)) {
+    std::cerr << "Cannot create source archive directory: " << ec.message() << "\n";
+    return;
+  }
+
+  llvm::SmallString<PATH_MAX> srcFilePath(file.getFilename());
```
AlexDenisov, line 47: [llvm::SmallString docs](https://llvm.org/docs/ProgrammersManual.html#llvm-adt-smallstring-h).
Perhaps, we can use something smaller than `PATH_MAX` here, but any other number would be arbitrary. OTOH, QL db filenames can get rather long (especially on macOS).

swift/ql/lib/swift/File.qll
```diff
@@ -0,0 +1,24 @@
+/** TODO: to be replaced */
```
AlexDenisov, line 1: This will be replaced by a combination of generated and hand-written code. Currently, I picked it up to close the "ql test" circle.

swift/extractor/Extractor.cpp
```diff
@@ -0,0 +1,101 @@
...
+#include <llvm/Support/FileSystem.h>
+#include <llvm/Support/Path.h>
+
+using namespace codeql;
+
+Extractor::Extractor(const Configuration& config, swift::CompilerInstance& instance)
+    : config{config}, compiler{instance} {}
+
+void Extractor::extract() {
+  if (compiler.getPrimarySourceFiles().empty()) {
```
jketema, line 21: What's the reason for this `if`? In other words, why would be primary source files be missing?

swift/extractor/Extractor.cpp
```diff
@@ -0,0 +1,101 @@
...
+    swift::ModuleDecl* module = compiler.getMainModule();
+    if (!module->getFiles().empty() &&
+        module->getFiles().front()->getKind() == swift::FileUnitKind::Source) {
+      /// We can only call getMainSourceFile if the first file is of a Source kind
```
jketema, line 25:
```suggestion
      // We can only call getMainSourceFile if the first file is of a Source kind
```

Or is this a special convention I'm not aware of?

swift/extractor/Extractor.cpp
```diff
@@ -0,0 +1,101 @@
...
+      swift::SourceFile& file = module->getMainSourceFile();
+      extractFile(file);
+    }
```
jketema, line 28: Would it make sense to add an else branch that raises an error?

swift/extractor/Extractor.cpp
```diff
@@ -0,0 +1,101 @@
...
+    return;
+  }
+
+  if (std::error_code ec = llvm::sys::fs::copy_file(srcFilePath, dstFilePath)) {
+    std::cerr << "Cannot archive source file '" << srcFilePath.str().str() << "' -> '"
+              << dstFilePath.str().str() << "': " << ec.message() << "\n";
+    return;
+  }
+
+  /// TODO: find a more robust approach to avoid collisions?
```
jketema, line 66: Alternative could be to encode the complete `srcFilePath` by substituting the path separator by some other character.

swift/extractor/Extractor.cpp
jketema: Might be good to intro a typedef, so you don't have to write `llvm::SmallString<PATH_MAX>` all over the place an think every time what the template parameter should be.

swift/extractor/Extractor.cpp
```diff
@@ -0,0 +1,101 @@
...
+  std::string tempTrapName = file.getFilename().str() + '.' + std::to_string(getpid()) + ".trap";
```
jketema, line 67: Why do you first create a temp file and then move it into place? This might be worthy of a comment.

swift/extractor/Extractor.cpp
jketema: Add a TODO so you won't forget?

swift/extractor/main.cpp
```diff
@@ -3,15 +3,58 @@
...
 
 int main(int argc, char** argv) {
+  if (argc == 1) {
+    /// TODO: print usage
+    return 1;
+  }
+  /// The frontend can be called in different modes, we are only interested
+  /// in -frontend mode
+  /// TODO: filter out at the tracer level
+  if ("-frontend"s != argv[1]) {
```
jketema, line 40: Wouldn't it be better to omit this from the command line and insert it yourself (as it's always needed)? That would also make inserting it in `Extractor.cpp` redundant, and also the filtering you do below would then be redundant.

swift/extractor/main.cpp
```diff
@@ -3,15 +3,58 @@
...
+      continue;
     }
+    args.push_back(argv[i]);
   }
-  return 0;
+  configuration.frontendOptions = args;
+  Observer observer(configuration);
+  int frontend_rc = swift::performFrontend(args, "swift-extractor", (void*)main, &observer);
+  llvm::llvm_shutdown();
+  return frontend_rc;
```
jketema, line 59: So, suppose you're unable to write a trap file, and report an error. Do you get a non-zero error code here?

swift/ql/lib/swift.dbscheme
```diff
@@ -8,9 +8,6 @@ sourceLocationPrefix(
   string prefix: string ref
 );
 
-answer_to_life_the_universe_and_everything(
-   int answer: int ref
-)
```
jketema, line 11: Lots of redundant empty lines here now 😄

swift/extractor/Extractor.cpp
AlexDenisov: I guess it's worth adding a comment there.
In general, there are many ways the frontend can be called:
```
swift-frontend -c main.swift # compiles main.swift
swift-frontend -primary-file a.swift b.swift c.swift # compiles a.swift only
swift-frontend <some other command to build a swift module from precompiled files, no source files involved>
```

swift/extractor/Extractor.cpp
AlexDenisov: Just an odd habit of mine 😃

swift/extractor/Extractor.cpp
AlexDenisov: It's not necessary an error, it's just a case when we cannot extract anything useful.

swift/extractor/Extractor.cpp
AlexDenisov: Yeah, taking a hash of the file content is another option. This way we can also "exit early" if such a trap file already exists.

swift/extractor/Extractor.cpp
AlexDenisov: There could be several parallel processes that attempt to extract the same file.
Adding a comment is a great idea!

swift/extractor/main.cpp
AlexDenisov: I hope this will be resolved with the new tracer integration. I don't yet know how much control over CLI args we have in this case.

swift/extractor/Extractor.cpp
jketema: Are all these paths even possible in the context of the extractor?

swift/extractor/main.cpp
AlexDenisov: No, that would still look like a success. For better error propagation we'd need some kind of our own `DiagnosticsEngine` that will record errors and report them here. Though, it's a bit premature at this stage.

swift/extractor/Extractor.cpp
jketema: Yeah, so I find this slightly confusing, as in DoxyGen`///` has a special meaning. I would try to avoid this as much as possible.

swift/extractor/Extractor.cpp
AlexDenisov: The very next step is to replace this code, so I'd reserve TODOs for more "long-term" items :)

swift/extractor/BUILD.bazel
```diff
@@ -8,7 +8,15 @@ alias(
 
 cc_binary(
     name = "extractor",
-    srcs = ["main.cpp"],
+    srcs = [
+        "Extractor.h",
+        "Extractor.cpp",
+        "Configuration.h",
+        "main.cpp",
+    ],
```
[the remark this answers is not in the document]
AlexDenisov: I prefer being explicit and have no issues with being stuck with having to update this every time you add/remove a file 😄
I'm happy to switch to globs, though. But before doing that I'd like to see the "final" structure of the files in place. I.e., if we add more folders, then we'd have to change the glob anyways, right?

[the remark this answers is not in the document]
jketema: True.

swift/extractor/Extractor.cpp
AlexDenisov: Yes, they are. The tracer invokes the extractor without knowing exactly what the swift-frontend does. It's our job later to decide what to do with all the arguments/invocation.

swift/extractor/main.cpp
```diff
@@ -3,15 +3,58 @@
...
+    return 0;
+  }
   PROGRAM_START(argc, argv);
-  if (auto trapDir = getenv("CODEQL_EXTRACTOR_SWIFT_TRAP_DIR")) {
-    std::string file = trapDir;
-    file += "/my_first.trap";
-    if (std::ofstream out{file}) {
-      out << "answer_to_life_the_universe_and_everything(42)\n";
+
+  codeql::Configuration configuration{};
```
sashabu, line 45: Optional: This function is getting quite long. I'd suggest factoring out some of the config setup. Some ideas:
 1. All-in-one: `codeql::Configuration swift_config_from_environment_and_cli_args(int argc, char** argv);`
 1a. As above, but as a static member `codeql::Configuration::from_environment_and_cli_args` for a less clunky name.
 2. Two parts: 
```
// Populates trapDir and sourceArchiveDir
codeql::Configuration swift_configuration_from_environment();
std::vector<std::string> swift_frontend_options_from_cli_args(int argc, char** argv);
```
 3. Just the second part (`swift_frontend_options_from_cli_args`), and keep the rest in line.
 
I think the all-in-one in particularly nice because we can pass the result directly to the `Observer` constructor (if it's going to make a copy). But it does have quite a clunky name unless you can come up with something better.

swift/extractor/main.cpp
```diff
@@ -3,15 +3,58 @@
...
+    if ("-frontend"s == argv[i]) {
+      continue;
     }
+    args.push_back(argv[i]);
   }
-  return 0;
+  configuration.frontendOptions = args;
+  Observer observer(configuration);
+  int frontend_rc = swift::performFrontend(args, "swift-extractor", (void*)main, &observer);
+  llvm::llvm_shutdown();
```
sashabu, line 58: Optional: Might be nice to wrap this in [`wil::scope_exit`](https://github.com/microsoft/wil/wiki/RAII-resource-wrappers#wilscope_exit).

swift/extractor/main.cpp
sashabu: Of course the exit code isn't the only way to track failures: we can also use logs or TRAP contents (and either could be more fine-grained, allowing us to track *which* files we failed to analyse).

swift/ql/lib/swift/File.qll
```diff
@@ -0,0 +1,24 @@
...
+
+/** A file. */
+class File extends @file {
+  /** toString */
```
sashabu, line 5: Nit (mostly just curious): Are the useless comments to silence a linter?

sashabu (changes requested):
I didn't get to reviewing `Extractor.cpp` today, but here are some preliminary comments to get you started.

swift/extractor/Extractor.cpp
andersfugmann: Any reason we cannot use `mkstemps` function, which is robust to collisions?

swift/extractor/Configuration.h
```diff
@@ -0,0 +1,15 @@
+#ifndef SWIFT_EXTRACTOR_CONFIGURATION_H_
+#define SWIFT_EXTRACTOR_CONFIGURATION_H_
+
+#include <string>
+#include <vector>
+
+namespace codeql {
+struct Configuration {
```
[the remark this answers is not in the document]
AlexDenisov: On one hand, we are under the `swift` folder so that should be enough.
OTOH, prefixed names are better for "greppability."
I'm fine with either way, @redsun82 do you have any thoughts on this?

swift/extractor/Extractor.h
```diff
@@ -0,0 +1,24 @@
...
+#define SWIFT_EXTRACTOR_H_
+
+#include "Configuration.h"
+#include <swift/AST/SourceFile.h>
+#include <swift/Frontend/Frontend.h>
+#include <memory>
+
+namespace codeql {
+
+class Extractor {
```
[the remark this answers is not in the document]
AlexDenisov: Simple API would make sense assuming that this is the final shape of the code, but it's a subject to change.
I'd give it more time to stabilize before we can simplify it.

swift/extractor/Extractor.h
```diff
@@ -0,0 +1,24 @@
...
+ public:
+  explicit Extractor(const Configuration& config, swift::CompilerInstance& instance);
+  void extract();
+
+ private:
+  void extractFile(swift::SourceFile& file);
+
+  const Configuration& config;
```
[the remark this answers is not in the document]
AlexDenisov: I'd prefer to stick to references and use "borrowing" semantics here.

swift/extractor/Extractor.h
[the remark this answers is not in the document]
AlexDenisov: I'll add a TODO for this item as the API will change.

swift/extractor/main.cpp
```diff
@@ -3,15 +3,58 @@
 #include <stdlib.h>
 
 #include <swift/Basic/LLVMInitialize.h>
+#include <swift/FrontendTool/FrontendTool.h>
+
+#include "Extractor.h"
+
+using namespace std::string_literals;
+
+class Observer : public swift::FrontendObserver {
```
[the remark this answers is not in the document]
AlexDenisov: Any suggestions what should be included in the doc?

swift/extractor/main.cpp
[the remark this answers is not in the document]
AlexDenisov: These are somewhat orthogonal, but you are right, we should call `INITIALIZE_LLVM()`.

swift/extractor/main.cpp
AlexDenisov: It will grow even more soon, so I'd gather more code here before we can factor out some parts.

swift/extractor/main.cpp
AlexDenisov: FWIW, LLVM has [llvm_shutdown_obj](https://github.com/llvm/llvm-project/blob/09c2b7c35af8c4bad39f03e9f60df8bd07323028/llvm/include/llvm/Support/ManagedStatic.h#L118), though we don't need it as it's shutdown by the `PROGRAM_START`.

swift/ql/lib/swift/File.qll
AlexDenisov: Yes, there is a CI check for this, which we may turn off for the time being.

swift/extractor/main.cpp
```diff
@@ -3,15 +3,58 @@
...
+  void performedSemanticAnalysis(swift::CompilerInstance& instance) override {
+    codeql::Extractor extractor{config, instance};
+    extractor.extract();
+  }
+
+ private:
+  const codeql::Configuration& config;
+};
+
+static std::string getenv_default(const char* envvar, const std::string& def) {
```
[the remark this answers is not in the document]
AlexDenisov: Renamed.

swift/extractor/Extractor.cpp
AlexDenisov: I'll leave it as is for now and will revisit when we have more than one file/function using the SmallString for file/paths related tasks.

swift/extractor/main.cpp
```diff
@@ -3,15 +3,58 @@
...
+    /// TODO: print usage
+    return 1;
+  }
+  /// The frontend can be called in different modes, we are only interested
+  /// in -frontend mode
+  /// TODO: filter out at the tracer level
+  if ("-frontend"s != argv[1]) {
+    return 0;
+  }
   PROGRAM_START(argc, argv);
```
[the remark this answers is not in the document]
AlexDenisov: I just added a comment that this is needed by Swift/LLVM without going much into the details, as those details may change across releases and we'll live with an outdated comment.

swift/extractor/Configuration.h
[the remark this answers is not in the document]
sashabu: > we are under the `swift` folder so that should be enough.

That doesn't affect linkage though. We've spoken about sharing some code between the C++ and Swift extractors in the future. If that happens, there's a clear possibility that the shared CodeQL code have a CodeQL config in a `codeql::Configuration` class. Same goes for if the C++ extractor also went for an "under a `c` folder" argument and named its own config the same thing.

> OTOH, prefixed names are better for "greppability."

Which is also important for "modern", web-based flows such as cs.github.com and PR reviews. And even in an IDE, if I'm switching back and forth, it's nice to know at a glance what I'm looking at.

swift/extractor/Extractor.h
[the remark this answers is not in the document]
sashabu: > I'd give it more time to stabilize before we can simplify it.

Writing complex code with the intention of simplifying it later rarely works out in practice. Not only will the unnecessary complexity slow us down in the meantime, but we're also unlikely to actually go back and refactor it once the API has stabilised. In general, it's better to start with the simplest API that will do the job, and only introduce complexity when it is needed.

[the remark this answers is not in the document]
sashabu: > the API will change

If it'll change in this PR, let's leave this unresolved to track the new API's documentation. If in a later PR, let's document the current API for the benefit of anyone (who might not be you - e.g. if you win the lottery and quit tomorrow) reading this code in the meantime (which may be longer than you expect - e.g. if it turns out this API is good enough, or if the API only changes further down the file and the TODO goes unnoticed at that time).

swift/extractor/Extractor.h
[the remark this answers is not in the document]
sashabu: Could you elaborate on the technical reasons for this preference?

The potential lifetime issues with capturing a temporary or otherwise outliving the captured object are real risks that frequently occur, especially at public API boundaries, and especially in the absence of documentation.

If this were an internal class and the exposed API were a function as per my other comment, this would be a much smaller issue (assuming the lifetime requirement were documented) both due to the much lower "surface area" (caller and callee in the same file, usually with only one call to the constructor), and for this sort of use case due to how it'd be used: the external caller would provide the object (whose lifetime will be at least that of the function call) and we'd just forward it into the object constructed just for that call and destroyed before returning from the function.

swift/extractor/main.cpp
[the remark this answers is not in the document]
sashabu: 1. What it does (triggers extraction) and
2. Why we need it (because that's the API the Swift frontend provides).

Something along the lines of "Wraps the Swift extractor in a visitor for use with Swift's frontend API." - if that's an accurate understanding.

swift/extractor/main.cpp
sashabu: I think that's all the more reason to refactor now, so when we try to add functionality here, understanding the flow doesn't slow us down!

swift/extractor/Extractor.h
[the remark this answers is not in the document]
AlexDenisov: The config object is not owned by any specific class, but shared across (potentially many) of them.
Yes. there is a risk of the lifetime issue, but I think we can take this risk.

swift/extractor/main.cpp
AlexDenisov: This function is ~30 LoC, I don't think it's worth refactoring it right now.

swift/extractor/BUILD.bazel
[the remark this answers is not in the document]
redsun82: personally I prefer globbing, because I think file names are just an implementation detail (and I find keeping up to date the BUILD file just for that to be a bit boring). And having to do the book keeping can lead us not to split big files into smaller ones that might help readability... Regarding separate folders, they should just be separate bazel packages to keep dependencies well separated. So my favourite mode would be to consistently use "flat" globs (`glob(["*.h"])`  for `hdrs`, `glob(["*.cpp"])` for `srcs`), and use separate small bazel packages and `deps`  extensively.

But I'm not adamant about using globbing :)

swift/extractor/Configuration.h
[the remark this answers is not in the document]
redsun82: I'm ok with suffixing this with swift

swift/extractor/BUILD.bazel
```diff
@@ -8,7 +8,15 @@ alias(
...
+    copts = [
+        "-fno-rtti",
```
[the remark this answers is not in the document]
redsun82: as this will also be required by other separate targets, for now I would write this in `.bazel.rc` alongside the setting for `c++17`

swift/extractor/Extractor.cpp
```diff
@@ -0,0 +1,106 @@
...
+
+  trap << "#0=*\n";
+  trap << "files(#0, " << std::quoted(srcFilePath.str().str()) << ")\n";
+
+  // TODO: Pick a better name to avoid collisions
+  std::string trapName = file.getFilename().str() + ".trap";
+  llvm::SmallString<PATH_MAX> trapPath(config.trapDir);
+  llvm::sys::path::append(trapPath, trapName);
+
+  // TODO: The last process wins. Should we do better than that?
```
redsun82, line 100: another strategy would be for the first process to acquire a lock and all subsequent processes trying to wite the same file to just give up and not waste CPU cycles at all

.github/workflows/swift-qltest.yml
```diff
@@ -32,6 +32,9 @@ jobs:
       - name: Build Swift extractor
         run: |
           bazel run //swift:create-extractor-pack
+        env:
+          CC: clang
+          CXX: clang++
```
redsun82, line 37: instead of this you should write in `.bazelrc`
```
build --repo_env=CC=clang  --repo_env=CXX=clang++
```
so that it also auto applies on dev builds

redsun82 (changes requested):

swift/extractor/Extractor.h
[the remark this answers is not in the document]
redsun82: I kind of like @sashabu 's suggestion here. The point of making these small PRs is exactly to reshape parts of the PoC in smaller steps, so we should not refrain from simplifying and refactoring code because more code is to come from the PoC, unless we see already that such a simplification is already down the line in the PoC import (which I don't think is the case here).

swift/codegen/prefix.dbscheme
```diff
@@ -5,6 +5,3 @@ sourceLocationPrefix(
   string prefix: string ref
 );
 
-answer_to_life_the_universe_and_everything(
-   int answer: int ref
-);
```
redsun82, line 10: or should we leave it as an easter egg? :hand_over_mouth: just joking

swift/extractor/Extractor.cpp
redsun82: I don't fully understand the answer here: how would there be a collision with `mstemps`?

.github/workflows/swift-qltest.yml
redsun82: and I'll link [this comment](https://github.com/github/codeql/pull/8788/files#r858429814) here as the PR UI might hide it

swift/extractor/Extractor.cpp
AlexDenisov: Uh, I referred to `mktemp` which does not create the file but only gives a 'unique' file name. I take my words back 😃

swift/extractor/Configuration.h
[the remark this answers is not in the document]
AlexDenisov: Ok, let's prefix it then.

.github/workflows/swift-qltest.yml
AlexDenisov: @redsun82 wouldn't it break cases on Ubuntu with versioned binary names, e.g. `clang-12`?

swift/extractor/BUILD.bazel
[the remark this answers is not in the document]
AlexDenisov: @redsun82 but then it would apply to the whole `ql` repo and not only to the Swift language pack? Is it something we want to do?

swift/extractor/main.cpp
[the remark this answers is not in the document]
AlexDenisov: @sashabu added a comment.

swift/extractor/Extractor.cpp
redsun82: I think we can revisit this when we fix the TODO

.github/workflows/swift-qltest.yml
redsun82: you can always override that creating a `local.bazelrc`, and ubuntu will have a link to a default clang version

swift/extractor/Extractor.h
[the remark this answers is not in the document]
redsun82: for the record, as I have discussed this with Alex on zoom: I would also kind of lean towards safer approaches like copies or `shared_ptr`s, on the other hand one thing that mitigates the risk, is that we are not dealing with a long-lived daemon application, for example creating and destroying services at runtime. We are within an application that does a specific task and then exits, which makes lifetime issues somewhat simpler. In such a scenario, it's not so weird to keep in mind there is for example one single configuration object created at the beginning of `main` that is guarateed to live through the whole run (as long as we avoid objects with static storage duration, which would be a bad idea anyway).

On the other hand, I wouldn't mind switching to `shared_ptr` for peace of mind, as the overhead would probably be very negligible.

swift/extractor/main.cpp
redsun82: does that mean we don't need this line then?

btw, I would rather use [gsl::finally](https://github.com/microsoft/GSL/blob/main/include/gsl/util#L89-L96) than wil::scope_exit (as I'd rather pull in the `gsl` library then `wil`).

PR #8789, opened 2022-04-20
JS/PY: mention newtype constructors in API graph label classes

erik-krogh:
See https://github.com/github/codeql/pull/8785 

Evaluations are underway, should be uneventful.


tausbn (approved):
Python :+1:

erik-krogh:
Both the JS and the Python evaluations hint towards a performance improvement :rocket: (See backrefs).

PR #8790, opened 2022-04-20
Add auto-remediation to InsecureDependencyResolution.qhelp

JLLeitschuh:
Adds an automated remediation suggestion to `InsecureDependencyResolution.qhelp`.

I authored an open rewrite recipe that can be leveraged, free for OSS, to automatically remediate this security vulnerability. The source for this recipe can be found here: https://github.com/openrewrite/rewrite/blob/cfd371c295950ee3f3ee4b15f43cb13c37ed130f/rewrite-maven/src/main/java/org/openrewrite/maven/security/UseHttpsForRepositories.java

### DIFF
--- a/java/ql/src/Security/CWE/CWE-829/InsecureDependencyResolution.qhelp
+++ b/java/ql/src/Security/CWE/CWE-829/InsecureDependencyResolution.qhelp
@@ -21,6 +21,11 @@ against your project's users.
 
 <p>Always use HTTPS or SFTP to download artifacts from artifact servers.</p>
 
+<p>This vulnerability can be automatically remediated with this
+<a href="https://docs.openrewrite.org/reference/recipes/maven/security/usehttpsforrepositories">OpenRewrite Recipe</a>
+or via
+<a href="https://app.moderne.io/recipes/org.openrewrite.maven.security.UseHttpsForRepositories">Moderne</a>.
+
 </recommendation>
 
 <example>


atorralba:
Although this looks good from a security perspective, I've checked with our product team and unfortunately we'll need to reject this. Automated fixes isn't something that we want to suggest we're tackling right now — even if it's only a link to an external project, it could send the wrong message from a product perspective.

atorralba:
Automated fixes is a concept that could make sense to add in one way or another to CodeQL in the future, but it needs to be well-thought. Our concern is that users could get the message that we have decided to tackle automated fixes _right now_ by recommending an external tool to do so. And also, only mentioning it in one query would add to the confusion.

We appreciate the contribution, but it has deeper implications than it would initially seem and it's a can of worms we don't want to open right now.

atorralba:
Reopening the PR since the discussion is active and we can try to find an intermediate solution.

> This balance and this decision seems at-odds with the mission of the GH Security Lab and various proclamations that have been made GitHub regarding the desire to help support securing OSS.

I don't think my previous comments should be interpreted in those terms, and I apologize for any miscommunication on my end. What we are trying to achieve here is that users understand that we're not specifically providing any automated remediation advice, because doing so implies some risk of breaking the users' source code and that liability shouldn't be on CodeQL. I'm going to defer to @turbo on this one if further clarification is needed.

Having said that, we suggest the following:

Since we usually add links to technical blog posts or articles that explain a vulnerability and offer remediation advice (that the reader needs to understand and implement in their codebase appropriately, becoming responsible for it), what would you think of creating a post like that yourself for this vulnerability? In it you could add any remediation advice you find appropriate, including a link to your OpenRewrite recipe. And then there should be no issue in adding a link to that article in the references section of this query (in this PR or in a new one), which is something we've done plenty in the past.

yo-h:
The decision on whether to merge this in its current form or in a modified form falls outside of the engineering team. @turbo, please provide guidance from a product perspective on how to proceed.

PR #8791, opened 2022-04-21
JS: Support static accessor calls and fix some FP accessor-calls

asgerf:
Fixes two problems with our call graph for accessors, and adds support for calls to static accessors declared on a class.

This PR deals with two bugs:
- `ClassNode.getStaticMethod` previously included static accessors. This lead to some FP call edges as a result from confusing accessors and methods.
- A `PropWrite` node exists for accessors, representing the installation of an accessor. We accidentally saw these as potential calls to the setter. 

It is addressed as follows:
- Introduced `ClassNode.getStaticMember`, analogous to `ClassNode.getInstanceMember`, and made `getStaticMethod` only include methods.
- Added accessor-calls to static accessors to the call graph, and made sure static method calls aren't seen as calls to an accessor.
- Added `PropWrite.isAccessorInstallation` to distinguish between ordinary property assignments and accessor installation. Accessor installations are no longer seen as potential calls to a setter.

[Evaluation](https://github.com/github/codeql-dca-main/blob/data/asgerf/static-accessor__default__code-scanning__2/reports/alert-meta-comparison.md) on default slugs:
- 3811 new call edges (calls to static accessors, and method calls whose receiver came from such a call)
- 1687 removed FP call edges (from method calls and accessor installations being seen as accessor calls)
- [Performance](https://github.com/github/codeql-dca-main/tree/data/asgerf/static-accessor__default__code-scanning__2/reports) looks OK, it's not without some cost, but acceptable given the nature of the fix

### DIFF
--- a/javascript/ql/lib/semmle/javascript/dataflow/DataFlow.qll
+++ b/javascript/ql/lib/semmle/javascript/dataflow/DataFlow.qll
@@ -589,6 +589,13 @@ module DataFlow {
      * Gets the node where the property write happens in the control flow graph.
      */
     abstract ControlFlowNode getWriteNode();
+
+    /**
+     * If this installs an accessor on an object, as opposed to a regular property,
+     * gets the body of the accessor. `isSetter` is true if installing a setter, and
+     * false is installing a getter.
+     */
+    DataFlow::FunctionNode getInstalledAccessor(boolean isSetter) { none() }
   }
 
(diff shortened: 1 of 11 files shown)


erik-krogh (approved):
Nice :+1:  
LGTM.  

And the performance evaluation looks to be within the margin of error.

PR #8792, opened 2022-04-21
C#: Exclude model generation queries from all suites

hvitved:
I verified locally that e.g.
```
codeql resolve queries csharp/ql/src/codeql-suites/csharp-code-scanning.qls | grep Model
```
would return the three queries before, but not after.


esbena:
LGTM.
This essentially addresses the same problem that https://github.com/github/codeql/pull/8526 did, perhaps a unified approach is worth discussing elsewhere?

esbena (approved):

MathiasVP:
I guess this doesn't need a change note as it won't affect any customers, right?

michaelnebel (approved):
My bad.
LGTM! :-)

PR #8793, opened 2022-04-21
Exclude internal diagnostics from all selectors

MathiasVP:
Thanks to @esbena for pointing this out.

@esbena I did a sweep over all the selectors and added the exclusion rule in all of them (instead of just `security-extended-selectors.yml`). I hope that's not controversial 🤞.


esbena (approved):
LGTM.

https://github.com/github/codeql-team/issues/919 may change how this is done, but it should be fine to merge for now.

PR #8796, opened 2022-04-21
Remove redundant imports

erik-krogh:
Adds a QL-for-QL query that detects redundant imports.  
This PR also deletes **all** imports that were flagged as redundant by the query. 

An import `A` is redundant if: 
- `A` imports some module/file. 
- There exists an import `B` that also imports the same module/file as `A`. 
- There exists an import `C` that is in the same scope as `A`, and `C` comes before `A`. 
- One of the following holds: 
  - `C` = `B`
     (These are fixed in the [delete duplicate imports](https://github.com/github/codeql/pull/8796/commits/a96489b23dbf93c9ee11ad592ab5ae0b341192e6) commit).  
  - `C` imports a file/module that directly contains `B`.  (Not searching transitive imports, and that's on purpose).  
    (These are fixed in the [delete redundant imports](https://github.com/github/codeql/pull/8796/commits/ff73dbc35c0c4ac556cf55c1534023f7e53817b7) commit).  
  
All imports of the form `import X as Y` are skipped, and `private` imports should be handled correctly to preserve behavior.  

The requirement that `C` must come before `A` should prevent any import order regressions.   

Performance and test outputs appears unaffected, so I'm quite certain that this is completely behavior/performance preserving.  

Sidenote: The query also found some issues with [how Ruby were using the `PathGraph` module](https://github.com/github/codeql/pull/8810). But the fix for that went beyond just "delete all the redundant imports", so I landed that as a separate PR.  

---- 

Evaluations are looking good: [C++](https://github.com/github/codeql-dca-main/tree/data/PR-8796-0-cpp__1/reports), [C#](https://github.com/github/codeql-dca-main/tree/data/PR-8796-1-csharp/reports), [Java](https://github.com/github/codeql-dca-main/tree/data/PR-8796-2-java/reports), [JS](https://github.com/github/codeql-dca-main/tree/data/PR-8796-3-javascript/reports), [Python](https://github.com/github/codeql-dca-main/tree/data/PR-8796-4-python/reports), [Ruby](https://github.com/github/codeql-dca-main/tree/data/PR-8796-5-ruby/reports). 
I got some spurious (I hope) performance regressions in the first CPP evaluation, but it looks fine on a redo.  

---- 

Inspired by some recent review comments: [one](https://github.com/github/codeql/pull/8429#discussion_r842883716), [two](https://github.com/github/codeql/pull/7968#discussion_r838274055), [three](https://github.com/github/codeql/pull/8272#discussion_r830937689).


henrymercer (approved):
ML-powered queries 👍

ql/ql/src/queries/style/RedundantImport.ql
```diff
@@ -0,0 +1,17 @@
+/**
+ * @name Redundant import
+ * @description
+ * @kind problem
+ * @problem.severity warning
+ * @id ql/abstract-class-import
```
joefarebrother, line 6: Is this ID correct?

erik-krogh: Nope.

hvitved (approved):

aschackmull:
I haven't gone through the Java changes in details, but I don't think we'll want to bulk-accept these changes.

ql/ql/src/queries/style/RedundantImport.ql
```diff
@@ -0,0 +1,17 @@
+/**
+ * @name Redundant import
+ * @description An import that is redundant due to an earlier import
```
kaeluka, line 3: FYI, I just noticed that the description ends with a '.' in all or most of javascript queries. This convention is not used in ql-for-ql.

kaeluka (approved):
Javascript 👍

java/ql/lib/semmle/code/java/dataflow/FlowSources.qll
```diff
@@ -27,7 +27,6 @@ import semmle.code.java.frameworks.Guice
 import semmle.code.java.frameworks.struts.StrutsActions
 import semmle.code.java.frameworks.Thrift
 import semmle.code.java.frameworks.javaee.jsf.JSFRenderer
-private import semmle.code.java.dataflow.ExternalFlow
```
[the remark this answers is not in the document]
erik-krogh: > There's likely one or more of the above `semmle.code.java.frameworks.*` imports that also import `ExternalFlow`.  

Yep, there's 3 of them that have a non-private import of `ExternalFlow`: `Jdbc`, `android.Android`, and `android.Intent`. 

---- 

I can revert the deletion of that import.  

Adding a `private` modifier to all the imports of `ExternalFlow` in the `semmle.code.java.frameworks.*` files unfortunately breaks things.  
E.g. the `ImplicitPendingIntents.qll` file depends on `Intent.qll` publicly importing `ExternalFlow`.   

---- 

I can also revert all the Java changes if you're not comfortable with them.  
But they should be behavior preserving.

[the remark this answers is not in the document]
atorralba: I can look into the Android-related errors when making the `ExternalFlow` imports private if it helps.

geoffw0 (approved):
CPP :+1:, though I am a little nervous as to whether your logic is always correct for import cycles?

erik-krogh:
> CPP +1, though I am a little nervous as to whether your logic is always correct for import cycles?

My logic is deliberately only looking one "import-level" down so I don't have to consider import cycles.

geoffw0:
Does import order ever matter for QL correctness, or is it only an issue for caching?

erik-krogh:
> Does import order ever matter for QL correctness, or is it only an issue for caching?

I think it's only an issue for caching. 
The result in the end should always be the same.

geoffw0:
> My logic is deliberately only looking one "import-level" down so I don't have to consider import cycles.

Ah.  I'm suspicious its still possible for it to make a mistake where a pair of files include each other, but that's a specific and probably rare pattern.

> I think it's only an issue for caching.

This makes it difficult to produce a case demonstrating what I mean.

Anyway I don't have any reason to believe this theoretical concern is affecting any of the changes in this PR, so :+1:.

erik-krogh:
> Anyway I don't have any reason to believe this theoretical concern is affecting any of the changes in this PR, so .

Right. I get what you mean now, and yes that could be an issue.  
Example: 
```CodeQL
// A.qll
import B
import C

// B.qll
import A
import C

// C.qll
predicate something() {none()} 
```

Both imports of `C` will be seen as redundant.  
The query could be modified to handle such cases (detect the cycle), but I'll leave that for when it becomes an issue.

aschackmull:
Let's revert the java changes here for now. I think it's a good ql-for-ql query to have, which ought to improve our import-structure going forward, but I don't think these alerts ought to fixed that blindly. I've made a PR that removes the subset that I think is uncontroversial: https://github.com/github/codeql/pull/8875

java/ql/lib/semmle/code/java/dataflow/FlowSources.qll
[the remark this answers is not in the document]
atorralba: Done here https://github.com/atorralba/codeql/commit/4a3ed70a916eba48ddfa12af147504b1c5658c9f. I'll rebase and open a PR when https://github.com/github/codeql/pull/8875 is merged.

erik-krogh:
The test failures also happen on `main`. I'll do a merge when `main` after those are fixed.

java/ql/lib/semmle/code/java/dataflow/FlowSources.qll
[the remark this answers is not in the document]
atorralba: https://github.com/github/codeql/pull/8876

python/ql/src/experimental/Security/CWE-348/ClientSuppliedIpUsedInSecurityCheckLib.qll
```diff
@@ -1,7 +1,6 @@
 private import python
 private import semmle.python.Concepts
 private import semmle.python.ApiGraphs
-private import semmle.python.dataflow.new.DataFlow
```
yoff, line 4: This is the only one that surprises me. It is imported by `ApiGraphs` which should perhaps have done so privately..

erik-krogh: I remember that I tried to make the import of `DataFlow` private inside `ApiGraphs` when I did def-nodes for Python.  
But that caused a surprising amount of compile-errors, so I dropped that change.  

You could look into a better fix in a follow-up PR (like the Java team is doing now).

yoff: Then it is perhaps best to revert just the change to this file. Then we will have an alert to remind us to solve it later.

yoff (approved):
Thanks, LGTM now.

PR #8801, opened 2022-04-21
Ruby: Exclude `SplatExpr` from taint tracking

hvitved:
`SplatExpr`s are modelled using flow summaries, so there is no need to include them explicitly in `defaultAdditionalTaintStep`.


aibaars (approved):
LGTM

Issue #8805, opened 2022-04-21
Missing size for typedef


dbartol:
At least in some situations, the database can wind up with a `Type` with no size, even though that `Type` is for a `typedef` whose underlying type has a known size. This caused bad IR generation, although that was fixable by always using the size of the unspecified type.

To repro, run the following query on the LGTM database for g/nlohmann/json:

```ql
private import cpp

from Type t, Type ut, int size
where ut = t.getUnspecifiedType() and size = ut.getSize() and not exists(t.getSize())
select t, ut, size
```

Expected Result:
No results

Actual Result:
~871 types for which there is no size, despite the unspecified type having a size

I was unable to come up with a small repro in a QL test, so it might be tricky to debug.

PR #8806, opened 2022-04-21
C++: Work around missing size for typedef

dbartol:
See #8805 for the underlying issue. The symptom was bad IR generation on a large number of functions because we never generated an `IRVariable` for a parameter with the affected type.


jketema (approved):
LGTM

PR #8808, opened 2022-04-22
Javascript: Fix PrototypePollutingAssignment example which is incorrect use of express leads to no result when scanning.

vovikhangcdv:


asgerf (approved):
Thanks! LGTM. Will merge after tests pass

PR #8810, opened 2022-04-22
Ruby: dont import the PathGraph module from Query.qll files

erik-krogh:
Most of the time you import the `PathGraph` module from the `.ql` file of a path-query.  
But sometimes you also imported it from the `Query.qll` file (I removed those).  

(I found this from looking at redundant imports, and some of your `PathGraph` imports showed up as redundant).


aibaars:
Looks fine to me. There seems to be a (unrelated) problem with the identical files check.

erik-krogh:
I rebased on `main` to fix the identical files issue.

nickrolfe (approved):

PR #8811, opened 2022-04-22
Java: get tainttracking3/TaintTrackingImpl.qll in sync

erik-krogh:
https://github.com/github/codeql/pull/8748 added a copy of `TaintTrackingImpl.qll`.  

But the shared implementation changed in the meantime.   
So now we got the sync files check failing on main.

### DIFF
--- a/java/ql/lib/semmle/code/java/dataflow/internal/tainttracking3/TaintTrackingImpl.qll
+++ b/java/ql/lib/semmle/code/java/dataflow/internal/tainttracking3/TaintTrackingImpl.qll
@@ -109,32 +109,12 @@ abstract class Configuration extends DataFlow::Configuration {
   /** Holds if taint propagation into `node` is prohibited. */
   predicate isSanitizerIn(DataFlow::Node node) { none() }
 
-  /**
-   * Holds if taint propagation into `node` is prohibited when the flow state is
-   * `state`.
-   */
-  predicate isSanitizerIn(DataFlow::Node node, DataFlow::FlowState state) { none() }
-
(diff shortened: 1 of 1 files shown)


nickrolfe (approved):

aibaars (approved):

aibaars:
The error message from QLDoc seems genuine:
```
A fatal error occurred: Errors during binding:
ERROR: This predicate does not override another predicate (/home/runner/work/codeql/codeql/java/ql/lib/semmle/code/java/dataflow/internal/tainttracking3/TaintTrackingImpl.qll:118,3-120,4)
ERROR: This predicate does not override another predicate (/home/runner/work/codeql/codeql/java/ql/lib/semmle/code/java/dataflow/internal/tainttracking3/TaintTrackingImpl.qll:135,3-137,4)
```

erik-krogh:
> The error message from QLDoc seems genuine:

Yes. I'm confused about which file is actually the latest.

aschackmull (approved):

aschackmull:
> > The error message from QLDoc seems genuine:
> 
> Yes. I'm confused about which file is actually the latest.

Presumably that error is from main, and a side-effect of the files being out-of-sync. The change in this PR looks like the right fix to me.

erik-krogh:
> Presumably that error is from main, and a side-effect of the files being out-of-sync. The change in this PR looks like the right fix to me.

Yes. Parts of the failing CI check is: `HEAD is now at d309e150 Merge pull request #8748 from smowton/smowton/admin/dependent-dataflow-configs`, which suggests that it is trying to compare against the failing `main`.

PR #8813, opened 2022-04-22
C++: Cover variable sized member arrays without a size in `Buffer.qll`

jketema:
Currently the extractor incorrectly emits <strike>-1</strike> 0 for the array `data` below:
```
struct myStruct { // c
   ...
   char data[]; // v
};
```
This will change with the next version of EDG, and  no size will be emitted anymore. This PR makes sure `Buffer.qll` handles arrays without sizes.

This is a separate PR to ensure these these changes do not get lost in a much larger PR and to be able to check that this doesn't cause any regressions already before the EDG upgrade.

### DIFF
--- a/cpp/ql/lib/semmle/code/cpp/commons/Buffer.qll
+++ b/cpp/ql/lib/semmle/code/cpp/commons/Buffer.qll
@@ -10,11 +10,18 @@ import semmle.code.cpp.dataflow.DataFlow
  *   char data[1]; // v
  * };
  * ```
- * This requires that `v` is an array of size 0 or 1.
+ * or
+ * ```
+ * struct myStruct { // c
+ *   int amount;
+ *   char data[]; // v
+ * };
+ * ```
+ * This requires that `v` is an array of size 0 or 1, or that the array has no size.
  */
 predicate memberMayBeVarSize(Class c, MemberVariable v) {
   c = v.getDeclaringType() and
-  v.getUnspecifiedType().(ArrayType).getArraySize() <= 1
+  exists(ArrayType t | t = v.getUnspecifiedType() | not t.getArraySize() > 1)
 }
 
 /**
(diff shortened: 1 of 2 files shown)


cpp/ql/lib/semmle/code/cpp/commons/Buffer.qll
```diff
@@ -40,13 +51,18 @@ int getBufferSize(Expr bufferExpr, Element why) {
...
       // buffer is the parentPtr->bufferVar of a 'variable size struct'
       memberMayBeVarSize(parentClass, bufferVar) and
       why = bufferVar and
       parentPtr = bufferExpr.(VariableAccess).getQualifier() and
       parentPtr.getTarget().getUnspecifiedType().(PointerType).getBaseType() = parentClass and
-      result = getBufferSize(parentPtr, _) + bufferVar.getType().getSize() - parentClass.getSize()
+      (
+        if exists(bufferVar.getType().getSize())
+        then bufferSize = bufferVar.getType().getSize()
+        else bufferSize = 0
```
MathiasVP, line 63: Should this be `-1` to match the current extractor behavior? Setting `bufferSize` to `0` in this case sounds much more reasonable, but reading the PR description it sounds like that's not what we'd get currently?

MathiasVP (commented):
LGTM assuming DCA is happy. Just one minor question.

cpp/ql/lib/semmle/code/cpp/commons/Buffer.qll
jketema: Reading code is difficult 😢  It currently outputs `0`, and after the update it'll not output a value (which internally in the extractor is represented as `-1`). Thanks for questioning me.

I'll update the commit message once the DCA experiment is done.

jketema: So, `0` is correct.

MathiasVP: Ah, I see. Thanks for the explanation!

cpp/ql/lib/semmle/code/cpp/commons/Buffer.qll
```diff
@@ -10,11 +10,22 @@ import semmle.code.cpp.dataflow.DataFlow
...
+ * ```
+ * This requires that `v` is an array of size 0 or 1, or that the array has no size.
  */
 predicate memberMayBeVarSize(Class c, MemberVariable v) {
   c = v.getDeclaringType() and
-  v.getUnspecifiedType().(ArrayType).getArraySize() <= 1
+  exists(ArrayType t | t = v.getUnspecifiedType() |
+    t.getArraySize() <= 1
+    or
+    not exists(t.getArraySize())
```
geoffw0, line 27: I think these two lines could be simplified to:
```
not t.getArraySize() > 1
```

geoffw0 (approved):
Looks good, thanks for maintaining this case!

cpp/ql/lib/semmle/code/cpp/commons/Buffer.qll
jketema: Indeed, thanks for this. I'll apply this change.

jketema: Done.

jketema:
DCA experiment looks good.

MathiasVP (approved):
LGTM!

PR #8817, opened 2022-04-22
Java: Add value-preserving flow steps for Android's SharedPreferences

atorralba:
Adds fluent steps for the `SharedPreferences$Editor` class.


java/ql/lib/semmle/code/java/frameworks/android/SharedPreferences.qll
```diff
@@ -1,6 +1,7 @@
 /** Provides classes related to `android.content.SharedPreferences`. */
 
 import java
+import semmle.code.java.dataflow.ExternalFlow
```
atorralba, line 4:
```suggestion
private import semmle.code.java.dataflow.ExternalFlow
```

aschackmull (approved):

PR #8845, opened 2022-04-24
Ruby: Add partial support for working with RBI (Ruby Interface) files

alexrford:
RBI files are used by Sorbet to glean type information about Ruby code. These are valid Ruby files that can contain additional  information which Sorbet can use to perform both static and runtime type checking. I've focused mostly on the static elements of this in the current PR.

It's possible to define the source code of a program directly in RBI files, but it generally seems more common to have the source of a program in regular `.rb` files and to use separate corresponding RBI files to define type signatures. This approach is useful in defining type signatures for gems, collected into a repository at https://github.com/sorbet/sorbet-typed/tree/master/lib - similar to the https://github.com/DefinitelyTyped/DefinitelyTyped project for TypeScript. External type definitions are the focus of this initial library implementation. One possible use case is to use existing type definitions from `sorbet-typed` to generate a list of type signatures for an API, which could serve as a first step in modelling that API.

As RBI files are "just" Ruby, we just need to tell the extractor to look for `.rbi` files as well in order to extract them. I've not enabled this by default in this PR, partly because I'm concerned that methods with multiple definitions (one real, and one "prototype" for typing purposes) could cause problems with resolving call targets.

This library is marked as experimental for the moment. It includes support for a chunk of RBI, but it's definitely not comprehensive. Some major missing features of the current implementation are:
 - Inheritance and mixins
   - Sorbet has some more advanced features in this area such as sealed classes, final methods, abstract classes and methods, etc.
 - Some generic types like `T::Enumerable`, `T::Set`, `T::Range`
 - Anything relating to runtime typing information, such as `T.reveal_type(x)` to get the type of a variable `x`. These seemed less relevant for just extracting type signatures.
 - Sorbet's typed `T::Struct` and `T::Enum` classes which lets code that uses these concepts be written in a more type-safe way

See https://sorbet.org/docs/rbi for reference


aibaars:
> As RBI files are "just" Ruby, we just need to tell the extractor to look for `.rbi` files as well in order to extract them. I've not enabled this by default in this PR, partly because I'm concerned that methods with multiple definitions (one real, and one "prototype" for typing purposes) could cause problems with resolving call targets.

It may indeed cause problem, although most likely we'd resolve a call to both targets. The prototype likely has no body so its effect on dataflow etc should be minimal. 

In normal Ruby files you can also define the same method multiple times. At runtime the last definition "wins". So the problem with multiple definitions is not new to RBI; it's just more likely to happen with RBI files.

ruby/ql/lib/codeql/ruby/controlflow/CfgNodes.qll
```diff
@@ -751,4 +751,23 @@ module ExprNodes {
...
     PairCfgNode getAKeyValuePair() { result = this.getAnArgument() }
   }
+
+  private class CallableExprChildMapping extends ExprChildMapping, Callable {
+    // TODO: parameters?, callable body?
+    override predicate relevantChild(AstNode n) { none() }
+  }
+
+  /** A control-flow node that wraps a `Callable` AST expression. */
+  class CallableCfgNode extends ExprCfgNode {
```
aibaars, line 761: I think these classes are not very useful. The control flow of a Callable is just a single node representing the creation of the function-like object. There is no flow to the body or its parameters.

ruby/ql/lib/codeql/ruby/experimental/Rbi.qll
```diff
@@ -0,0 +1,412 @@
...
+      RbiTypeAlias() { this = API::getTopLevelMember("T").getAMethodCall("type_alias") }
+
+      // TODO: avoid mapping to AST layer
+      /**
+       * Gets the type aliased by this call.
+       */
+      RbiType getAliasedType() {
+        result.asExpr().getExpr() =
+          this.getBlock().asExpr().(ExprNodes::StmtSequenceCfgNode).getExpr().getLastStmt()
+      }
```
aibaars, line 80:
```suggestion
      RbiType getAliasedType() {
        result.asExpr() =
          this.getBlock().asExpr().(ExprNodes::StmtSequenceCfgNode).getLastStmt()
      }
```

ruby/ql/lib/codeql/ruby/experimental/Rbi.qll
```diff
@@ -0,0 +1,426 @@
...
+  /**
+   * Contains classes representing RBI types.
+   */
+  private module RbiTypes {
+    /**
+     * A node representing a Ruby Interface (RBI) type.
+     */
+    abstract class RbiType extends Expr { }
+
+    class ConstantReadAccessAsRbiType extends RbiType {
```
aibaars, line 27: This class has no qldoc and its name isn't very descriptive. Is it supposed to represent a named class or module type?

aibaars (commented):
This looks fine to me. We might want to have a generalized API that covers RBI and RBS when we add support for RBS files.

ruby/ql/lib/codeql/ruby/experimental/Rbi.qll
alexrford: Forgot to add qldoc for this. It's intended to represent cases like read accesses to the `Integer` and `MyList` classes and the `MyList2` constant in:
```
  MyList2 = T.type_alias(MyList)

  sig { params(l: MyList2).returns(Integer) }
  def len(l); end
```

Generally I think that these types would always represent some class, possibly via some intermediate type aliases, but the class definitions might be not be available in the database (`Integer` and `String` are common examples). I ended up making this class very broad to avoid missing cases like these.

alexrford: Separately, I've noticed that I can simplify this as just `class ConstantReadAccessAsRbiType extends RbiType, ConstantReadAccess { }` after moving this library to the AST layer.

aibaars (approved):

PR #8557, opened 2022-03-25
JS: Add StoredXss and XssThroughDom to ATM QL extraction code

TomBolton:
Add the security queries `StoredXss` and `XssThroughDom` to the ATM CodeQL feature extraction code.

These new queries should not be run in the ATM query suite, so have been explicitly excluded in the `javascript-atm-code-scanning.qls` file.

Will any of the tests need updating @henrymercer?

PR #8857, opened 2022-04-25
JS: Add StoredXss and XssThroughDom to ATM extraction queries

TomBolton:
- Add `StoredXss` and `XssThroughDom` to ATM extraction queries
- Update the expected test output with the two new queries
- Explicitly name each individual security query to include in the ATM query pack
- Update the acronyms in the class names to use camel case to address the code scanning warnings


TomBolton:
Copying Henry's [comment from a previous PR](https://github.com/github/codeql/pull/8557#discussion_r838614441):

> Update these to line up with the existing queries like XssAtm.ql:
> 
> Use an alert message like "(Experimental) This may be a cross-site scripting vulnerability due to $@. Identified using machine learning.", which doesn't include the score.
> Use "experimental" instead of "boosted" in the query name
> Use an @id like js/ml-powered/<RHS of original query id>
> Remove experimental/atm from the tags
> Rename ATM -> Atm

TomBolton:
Closing as the experimentation with renaming `ATM` -> `Atm` did not work as intended.

PR #8859, opened 2022-04-25
Python: Fix a bunch of bad joins

tausbn:
These were all observed on various forks of the `jython/jython` project.

--- 

No change note required, as this is just a performance fix.


yoff (approved):
LGTM

PR #8861, opened 2022-04-25
Swift: Update `schema.yml` and regenerate files

MathiasVP:
This matches the changes I made in the internal repo (but slightly modified to match the newer version of Swift).


AlexDenisov (approved):
🚢

PR #8871, opened 2022-04-26
QL: delete old copy of the identical files scripts

erik-krogh:
The files mentioned in `ql/scripts/identical-files.json` don't exist.  
It's just an old copy that we used when QL-for-QL had it's own repository.


MathiasVP (approved):

cklin:
@erik-krogh
I think `ql/scripts/merge_stats.py` might still be in use:
https://github.com/github/codeql/blob/800e4ea7df38c26a26cbd97d89e6d2282b7183ac/.github/workflows/ql-for-ql-dataset_measure.yml#L80

erik-krogh:
> @erik-krogh
> I think `ql/scripts/merge_stats.py` might still be in use:

You're right. 

Luckily it's identical to the `ql/scripts/merge_stats.py` file, so we can just point the workflow to use that.

PR #8872, opened 2022-04-26
Java: Add Editable.toString flow step

atorralba:
Adds a general flow step to support taint propagation when the contents of a text field in Android are accessed like:

```java
editText.getText().toString()
```

This generalization allows us to remove specific query taint steps that had the same goal.

### DIFF
--- a/java/ql/lib/change-notes/2022-04-26-android-editable-tostring-flow-step.md
+++ b/java/ql/lib/change-notes/2022-04-26-android-editable-tostring-flow-step.md
@@ -1,0 +1,4 @@
+---
+category: minorAnalysis
+---
+Added a flow step for `toString` calls on tainted `android.text.Editable` objects. 
--- a/java/ql/lib/semmle/code/java/frameworks/android/Widget.qll
+++ b/java/ql/lib/semmle/code/java/frameworks/android/Widget.qll
@@ -16,6 +16,18 @@ private class DefaultAndroidWidgetSources extends RemoteFlowSource {
   override string getSourceType() { result = "Android widget source" }
 }
 
+private class EditableToStringStep extends AdditionalTaintStep {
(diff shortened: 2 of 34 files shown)


intrigus-lgtm:
`Editable` inherits the `toString` method from `CharSequence`, so shouldn't this already be covered by this summary:
https://github.com/github/codeql/blob/b433f08cef9038d60e2bdf50bfedaf112f35d3f6/java/ql/lib/semmle/code/java/frameworks/Strings.qll#L59
What am I missing?

atorralba:
Yes, it should. But apparently there's an issue (that we are investigating) by which `CharSequence.toString` isn't extracted, so calls to `Editable.toString` point to the `Object.toString` callable, causing that specific CSV row to have no effect. If we pinpoint the original issue, this PR might very well be discarded. Thanks for pointing this out in any case.

smowton (approved):

aschackmull:
> Yes, it should. But apparently there's an issue (that we are investigating) by which `CharSequence.toString` isn't extracted, so calls to `Editable.toString` point to the `Object.toString` callable, causing that specific CSV row to have no effect. If we pinpoint the original issue, this PR might very well be discarded. Thanks for pointing this out in any case.

What was the result of this investigation?

atorralba:
> What was the result of this investigation?

Sorry for the late response @aschackmull. In the end, I think it has nothing to do with extraction, and everything to do with hierarchy.

The `Editable` interface extends `CharSequence`, which declares `toString`, but there are no concrete implementations of it until you reach `Object.toString` going up the hierarchy.

The concrete implementation of `toString` that makes `Editable.toString` a sensible taint step is in whatever class implements `Editable`, but that isn't seen by `Method::overrides`, which returns `Object.toString` as the only viable override.

That makes the `CharSequence.toString` summary to not apply to `Editable.toString` calls.

PR #8874, opened 2022-04-26
Java insecure cookies query: look through named constants

smowton:


atorralba (approved):
LGTM.

aschackmull (approved):
> LGTM. Would this be worth adding a test?

Nah, this seems simple enough.

PR #8877, opened 2022-04-26
Use flow to collection `Element` in MaD generator

tamasvajk:


csharp/ql/src/utils/model-generator/internal/CaptureModelsSpecific.qll
```diff
@@ -69,10 +69,16 @@ predicate isRelevantType(CS::Type t) {
...
+
 private string parameterAccess(CS::Parameter p) {
-  if Collections::isCollectionType(p.getType())
-  then result = "Argument[" + p.getPosition() + "].Element"
-  else result = "Argument[" + p.getPosition() + "]"
+  result = "Argument[" + p.getPosition() + "]" + getTypeBasedSuffix(p.getType())
+}
+
+private string returnValueAccess(CS::Callable c) {
+  result = "ReturnValue" + getTypeBasedSuffix(c.getReturnType())
```
tamasvajk, line 81: This only makes sense of the generated row is a `value` step and not a `taint` one.

PR #8884, opened 2022-04-26
Add additional `File` taint value flow models

JLLeitschuh:
Adds
 - File::getAbsoluteFile
 - File::getCanonicalFile
 - File::getAbsolutePath
 - File::getCanonicalPath


JLLeitschuh:
These changes are good as-is. It looks like none of the existing tests break. Any issues with these @aschackmull?

atorralba:
The performance of this seems OK. It adds a bunch of results to several projects and queries but I cursorily reviewed them and they make sense.

@JLLeitschuh for some reason your query `java/local-temp-file-or-directory-information-disclosure` loses 1 result (compared to `main`) when run with the changes of this PR on `apache/flink`. Maybe something you'd want to check out — it's a path that goes from `EnvironmentInformation.java:350:16` to `RocksDBKeyedStateBackendBuilder.java:262:21`.

JLLeitschuh:
I've looked at the paths generated from `EnvironmentInformation.java` in the results below (2 paths rendered for both) and they both seem to be exactly the same:

 - Before: https://lgtm.com/query/6811672767670768311/
 - After: https://lgtm.com/query/4726335492173164003/

I'm not really concerned about the loss of a single path here.

atorralba (approved):
LGTM then

java/ql/lib/semmle/code/java/dataflow/ExternalFlow.qll
```diff
@@ -300,6 +300,10 @@ private predicate summaryModelCsv(string row) {
       "java.net;URI;false;toAsciiString;;;Argument[-1];ReturnValue;taint",
       "java.io;File;false;toURI;;;Argument[-1];ReturnValue;taint",
       "java.io;File;false;toPath;;;Argument[-1];ReturnValue;taint",
+      "java.io;File;true;getAbsoluteFile;;;Argument[-1];ReturnValue;taint",
```
Marcono1234, line 303: For consistency with the two lines above should this be changed to not consider subtypes (`false`), though `java.io.File` can be subclassed (even though it is unlikely that many subclasses exist), so maybe the lines above should be changed (`false` → `true`)?

Marcono1234 (commented):
There are some other `File` methods which might also be interesting for taint propagation, but including them is probably out of scope I guess (especially since this pull request is already approved).

java/ql/lib/semmle/code/java/dataflow/ExternalFlow.qll
JLLeitschuh: The class can, and does (rarely) get extended. Solid example of it happening here:

https://github.com/gradle/gradle/blob/4f26967b87f2bdc66927348d8c04bb26c1238967/subprojects/internal-testing/src/main/groovy/org/gradle/test/fixtures/file/TestFile.java#L72

JLLeitschuh: Updated to change them all to `true`

java/ql/src/Security/CWE/CWE-200/TempDirUtils.qll
```diff
@@ -35,32 +35,6 @@ predicate isFileConstructorArgument(Expr expSource, Expr exprDest, int paramCoun
...
-    fileMethodAccess.getQualifier() = expSource and
-    fileMethodAccess = exprDest
-  )
-}
-
-/**
- * Holds if taint should propagate from `node1` to `node2` across some file creation or transformation operation.
- * For example, `taintedFile.getCanonicalFile()` is itself tainted.
- */
-predicate isAdditionalFileTaintStep(DataFlow::Node node1, DataFlow::Node node2) {
```
aschackmull, line 60: Keep this as an empty deprecated predicate for now:
```
deprecated predicate isAdditionalFileTaintStep(DataFlow::Node node1, DataFlow::Node node2) { none() }
```

aschackmull (commented):
Let's not make this a breaking change. Otherwise LGTM.

java/ql/src/Security/CWE/CWE-200/TempDirUtils.qll
intrigus-lgtm: OOI: Why would this be a breaking change?
Isn't this file unimportable by external users because it contains hyphens in it's path?

aschackmull (approved):

PR #8885, opened 2022-04-26
Replace `git.io` link with the actual URL

SukkaW:
All links on git.io will stop redirecting after April 29, 2022: https://github.blog/changelog/2022-04-25-git-io-deprecation/
Replace all git.io links with their actual URLs.


nickrolfe (approved):
Thanks for spotting and fixing these!

PR #8888, opened 2022-04-26
C++: Add support for createLSParser to the CWE-611 XXE query.

geoffw0:
Add support for `createLSParser` to the CWE-611 XXE query.

This particular extension is straightforward and we already have test cases.

### DIFF
--- a/cpp/ql/src/Security/CWE/CWE-611/XXE.ql
+++ b/cpp/ql/src/Security/CWE/CWE-611/XXE.ql
@@ -164,6 +164,13 @@ class ParseFunction extends Function {
 }
 
 /**
+ * The `createLSParser` function that returns a newly created `LSParser` object.
+ */
+class CreateLSParser extends Function {
+  CreateLSParser() { this.hasName("createLSParser") }
+}
+
+/**
  * Configuration for tracking XML objects and their states.
  */
 class XXEConfiguration extends DataFlow::Configuration {
@@ -176,6 +183,13 @@ class XXEConfiguration extends DataFlow::Configuration {
       call.getStaticCallTarget() = any(XercesDOMParserClass c).getAConstructor() and
       node.asInstruction().(WriteSideEffectInstruction).getDestinationAddress() =
         call.getThisArgument() and
+      encodeXercesDOMFlowState(flowstate, 0, 1) // default configuration
+    )
+    or
+    // source is the result of a call to `createLSParser`.
+    exists(Call call |
+      call.getTarget() instanceof CreateLSParser and
+      call = node.asExpr() and
       encodeXercesDOMFlowState(flowstate, 0, 1) // default configuration
     )
   }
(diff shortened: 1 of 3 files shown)


geoffw0:
No change note required - covered by the note in recent PR https://github.com/github/codeql/pull/8736.

cpp/ql/src/Security/CWE/CWE-611/XXE.ql
```diff
@@ -163,6 +163,13 @@ class ParseFunction extends Function {
   ParseFunction() { this.getClassAndName("parse") instanceof AbstractDOMParserClass }
 }
 
+/**
+ * The `createLSParser` function that returns a newly created `LSParser` object.
+ */
+class CreateLSParser extends Function {
+  CreateLSParser() { this.hasName("createLSParser") }
```
MathiasVP, line 170: Is this name too general to just be captured by the function name? If I were to create a function that constructed a class that captured, say, the output of `ls` from the command line I'd call this function `createLSParser` (hm, but maybe with a different casing, though). Should we check either that the return type is of the right type, or that the function lives in some namespace?

MathiasVP (approved):
LGTM other than a tiny comment.

cpp/ql/src/Security/CWE/CWE-611/XXE.ql
geoffw0: Yeah, I should probably figure out the namespace this should be in and check that as well.

geoffw0: It turns out the namespace name varies depending on the Xerces version, and its a class method but I'm not entirely confident the class name is stable.  So I've added matching on the return type, which the docs are clear is `DOMLSParser *`.  That should rule out any false matches.

cpp/ql/src/Security/CWE/CWE-611/XXE.ql
```diff
@@ -164,7 +164,17 @@ class ParseFunction extends Function {
 }
 
 /**
- * Configuration for tracking XML objects and their states.
+ * The `createLSParser` function that returns a newly created `LSParser` object.
+ */
+class CreateLSParser extends Function {
+  CreateLSParser() {
+    this.hasName("createLSParser") and
+    this.getType().(PointerType).getBaseType().getName() = "DOMLSParser" // returns a `DOMLSParser *`.
```
MathiasVP, line 172: This should make it handle typedefs as well.
```suggestion
    this.getUnderlyingType().(PointerType).getBaseType().getName() = "DOMLSParser" // returns a `DOMLSParser *`.
```

cpp/ql/src/Security/CWE/CWE-611/XXE.ql
MathiasVP: Fantastic!

cpp/ql/src/Security/CWE/CWE-611/XXE.ql
MathiasVP: Oops. I think I gave you bad advice. The predicate is called `getUnspecifiedType` (which explains the consistency error reported by ql-for-ql).

MathiasVP (approved):
LGTM!

PR #8889, opened 2022-04-26
Swift: add unit tests to code generation

redsun82:
Tests can be run with
```
bazel test //swift/codegen:tests
```

Coverage can be checked installing `pytest-cov` and running
```
pytest --cov=swift/codegen swift/codegen/test
```


.pre-commit-config.yaml
```diff
@@ -40,3 +42,10 @@ repos:
         language: system
         entry: bazel run //swift/codegen
         pass_filenames: false
+
+      - id: swift-codegen-unit-tests
```
AlexDenisov, line 46: Does it mean the codegen tests will run on every commit? It might convenient, but I'm afraid the overhead is way too high to enable it by default.

AlexDenisov (commented):
In general, LGTM, but I would appreciate clarification on the pre-commit hook.

AlexDenisov (approved):

PR #8893, opened 2022-04-27
Ruby: Simplify flow summary for `fetch`

hvitved:
The outputs into `self` are not needed, since that will be handled by ordinary use-use flow at the call sites.

### DIFF
--- a/ruby/ql/lib/codeql/ruby/frameworks/core/Array.qll
+++ b/ruby/ql/lib/codeql/ruby/frameworks/core/Array.qll
@@ -741,19 +741,14 @@ module Array {
 
     override predicate propagatesFlowExt(string input, string output, boolean preservesValue) {
       (
-        input = "Argument[self].ArrayElement[?]" and
-        output = ["ReturnValue", "Argument[self].ArrayElement[?]"]
-        or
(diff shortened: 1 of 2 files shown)


ruby/ql/lib/codeql/ruby/frameworks/core/Array.qll
```diff
@@ -741,19 +741,14 @@ module Array {
...
       (
-        input = "Argument[self].ArrayElement[?]" and
-        output = ["ReturnValue", "Argument[self].ArrayElement[?]"]
-        or
-        exists(ArrayIndex j | input = "Argument[self].ArrayElement[" + j + "]" |
-          j = i and output = "ReturnValue"
-          or
-          j != i and output = "Argument[self].ArrayElement[" + j + "]"
-        )
+        input = "Argument[self].ArrayElement[?," + i + "]" and
```
nickrolfe, line 744: Cool, I didn't know we supported syntax like `[?,0]`.

nickrolfe (approved):
LGTM

ruby/ql/lib/codeql/ruby/frameworks/core/Array.qll
hvitved: That's also a fairly recent addition thanks to @asgerf .

PR #8896, opened 2022-04-27
Enable Dependabot updates for Actions

henrymercer:
This will automatically create update PRs for workflow files referencing Actions that have since had a major update.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -16,3 +16,11 @@ updates:
     directory: "ruby/autobuilder"
     schedule:
       interval: "daily"
+
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "daily"
+    ignore:
+      - dependency-name: '*'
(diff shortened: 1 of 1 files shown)


.github/dependabot.yml
```diff
@@ -16,3 +16,11 @@ updates:
     directory: "ruby/autobuilder"
     schedule:
       interval: "daily"
+
```
adityasharad, line 19: Rogue space.

edoardopirovano (approved):

adityasharad (approved):

.github/dependabot.yml
henrymercer: If you're referencing the new line, that was intentional to separate this entry for the Actions ecosystem from the previous four entries for the Cargo ecosystem.  Happy to remove it if it's unhelpful though.

adityasharad: Ah that makes sense!

PR #8905, opened 2022-04-27
C#: Add auto generated comment to generated models as data files.

michaelnebel:


tamasvajk (approved):

misc/scripts/models-as-data/generate_flow_model.py
```diff
@@ -133,6 +133,9 @@ def readRows(self, bqrsFile):
 
     def asCsvModel(self, superclass, kind, rows):
         classTemplate = """
+/**
+ * THIS FILE IS AN AUTO-GENERATED MODELS AS DATA FILE. DO NOT EDIT.
+ */
```
tamasvajk, line 138: This should probably go before the `/** Definitions of taint steps in the {self.shortname} framework */` comment.

tamasvajk (changes requested):

misc/scripts/models-as-data/generate_flow_model.py
michaelnebel: Yes, you are right. No reason to have it there for each class.

misc/scripts/models-as-data/generate_flow_model.py
```diff
@@ -167,7 +167,10 @@ def makeContent(self):
             sourceCsv = ""
 
         return f"""
-/** Definitions of taint steps in the {self.shortname} framework */
+/** 
+ * THIS FILE IS AN AUTO-GENERATED MODELS AS DATA FILE. DO NOT EDIT.
+ * Definitions of taint steps in the {self.shortname} framework */
```
michaelnebel, line 172: Need to fix this

tamasvajk (approved):

PR #8909, opened 2022-04-27
Tree sitter update

aibaars:
This pull request updates the tree-sitter-ruby grammar used by our extractor. The pull request implements up- and downgrade scripts that are semantics preserving except in case of `complex` literals (`1i`, `3.14i`, `3ri`,etc).  In the new grammar `complex` literals are no longer a simple token but a node with a float, integer, or rational literal as child. As a result an upgrade script would have to create new nodes, which is not possible. Instead the upgrade script simply converts `complex` literals to `float` literals. This is wrong, but preserves the structure of the AST so there are no gaps in the  control and data flow graphs.


ruby/ql/lib/codeql/ruby/ast/Method.qll
```diff
@@ -252,10 +252,24 @@ class Lambda extends Callable, BodyStmt, TLambda {
...
+   * Get a local variable declared by this block.
+   * For example `local` in `{ | param; local| puts param }`.
+   */
+  LocalVariableWriteAccess getALocalVariable() { result = this.getLocalVariable(_) }
+
+  /**
+   * Gets the `n`th local variable declared by this block.
+   * For example `local` in `{ | param; local| puts param }`.
+   */
+  LocalVariableWriteAccess getLocalVariable(int i) { none() }
```
intrigus-lgtm, line 265:
```suggestion
  LocalVariableWriteAccess getLocalVariable(int n) { none() }
```

ruby/ql/lib/change-notes/2022-04-30-update-grammar.md
```diff
@@ -0,0 +1,4 @@
+---
+category: fix
+---
+The TreeSitter Ruby grammar has been updated; this fixes several issues where Ruby code was parsed incorrectly.
```
intrigus-lgtm, line 4:
```suggestion
The Tree-sitter Ruby grammar has been updated; this fixes several issues where Ruby code was parsed incorrectly.
```

ruby/extractor/Cargo.toml
```diff
@@ -11,7 +11,7 @@ flate2 = "1.0"
 node-types = { path = "../node-types" }
 tree-sitter = "0.19"
 tree-sitter-embedded-template = "0.19"
-tree-sitter-ruby = { git = "https://github.com/tree-sitter/tree-sitter-ruby.git", rev = "1ebfdb288842dae5a9233e2509a135949023dd82" }
+tree-sitter-ruby = { git = "https://github.com/tree-sitter/tree-sitter-ruby.git", rev = "1a3936a3545c0bd9344a0bf983fafc7e17443e39" }
```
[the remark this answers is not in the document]
aibaars: That would work, although I'd like to be explicit. Things won't actually work if you refreshed the lock file because a minor grammar change might require dbscheme, library and upgrade/downgrade scripts.

ruby/ql/lib/codeql/ruby/ast/Method.qll
```diff
@@ -252,10 +252,24 @@ class Lambda extends Callable, BodyStmt, TLambda {
 
 /** A block. */
 class Block extends Callable, StmtSequence, Scope, TBlock {
+  /**
+   * Get a local variable declared by this block.
+   * For example `local` in `{ | param; local| puts param }`.
```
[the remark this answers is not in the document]
aibaars: It's a very old feature, just rarely used.

ruby/ql/lib/codeql/ruby/ast/internal/AST.qll
```diff
@@ -248,9 +252,6 @@ private module Cached {
         casePattern(g)
       )
     } or
-    TScopeResolutionMethodCall(Ruby::ScopeResolution g, Ruby::Identifier i) {
-      isScopeResolutionMethodCall(g, i)
-    } or
```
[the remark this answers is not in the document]
aibaars: The parser now correctly distinguishes between calls using `::` and scope resolutions for constants, so this code is now dead. `expr::identifier` is a call with `::` as the call operator, and `expr::Constant` is a scope resolution.

aibaars:
> On the upgrade/downgrade scripts: I have a sense of what they're doing but don't know how to ensure that they work properly. Do we have any tests that ensure they give the expected result? We have a ["Check DB upgrade/downgrade script"](https://github.com/github/codeql/runs/6229472334?check_suite_focus=true) test but I don't know if it does much more than just run them and make sure nothing throws an error.

The `Check DB upgrade/downgrade script` doesn't check that much. I think it only verifies that the upgrades form a chain from the initial dbscheme to the current and the upgrade/downgrade script can be compiled.

I validated the behaviour of the upgrade scripts by building the `extractor-pack` from the `main` branch and run the QL tests. This causes QL test to build databases with the old extractor but run the test cases with the new library and dbscheme. In this case CodeQL runs the upgrade scripts because the databases are older than the new dbscheme.

These are the steps I used:
```
git checkout upstream/main .
./scripts/create-extractor-pack.sh
git checkout HEAD -f
codeql test run --learn  --search-path extractor-pack/ --search-path ql/lib ql/test
git difftool ql/test
```

For testing the downgrade scripts you can do the opposite, compile a new extractor, but run the QL tests with the old library.
```
./scripts/create-extractor-pack.sh
git checkout upstream/main ql/lib
codeql test run --learn  --search-path extractor-pack/ ql/test
git difftool ql/test
```
Note that there are some differences in the test output, but these should be benign or otherwise expected.

ruby/ql/lib/upgrades/9fdd1d40fd3c3f8f9db8fabf5a353580d14c663a/ruby_scope_resolution_def.ql
```diff
@@ -0,0 +1,16 @@
+class RubyScopeResolution extends @ruby_scope_resolution {
+  RubyScopeResolution() {
+    exists(RubyConstant name | ruby_scope_resolution_def(this, name)) and
+    not ruby_call_def(_, this)
+  }
```
nickrolfe, line 5: I was expecting that we'd simply drop rows where the second column is not a `@ruby_token_constant`, so this charpred seems like it could be deleted.

ruby/ql/lib/upgrades/9fdd1d40fd3c3f8f9db8fabf5a353580d14c663a/ruby_ast_node_info.ql
```diff
@@ -0,0 +1,45 @@
...
+        ruby_ast_node_info(child, oldParent, oldIndex, _) and
+        child != parent.(ScopeResolutionMethodCall).getScopeResolution()
+      |
+        oldParent = parent and x = 1
+        or
+        oldParent = parent.(ScopeResolutionMethodCall).getScopeResolution() and x = 0
+      )
+    |
+      child order by x, oldIndex
+    )
```
nickrolfe, line 44: Can you explain what this is doing, and why?

ruby/ql/lib/codeql/ruby/ast/Method.qll
```diff
@@ -252,10 +252,24 @@ class Lambda extends Callable, BodyStmt, TLambda {
 
 /** A block. */
 class Block extends Callable, StmtSequence, Scope, TBlock {
+  /**
+   * Get a local variable declared by this block.
```
nickrolfe, line 256:
```suggestion
   * Gets a local variable declared by this block.
```

ruby/ql/test/library-tests/ast/TreeSitter.ql
```diff
@@ -0,0 +1,31 @@
...
+
+import codeql.ruby.ast.internal.TreeSitter::Ruby
+
+/**
+ * Holds if `node` belongs to the output tree, and its property `key` has the
+ * given `value`.
+ */
+query predicate nodes(AstNode node, string key, string value) {
+  key = "semmle.label" and
+  value = "[" + node.getPrimaryQlClasses() + "] " + node.toString()
```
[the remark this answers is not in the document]
aibaars: That's a good idea.

ruby/ql/lib/upgrades/9fdd1d40fd3c3f8f9db8fabf5a353580d14c663a/ruby_call_method.ql
```diff
@@ -0,0 +1,52 @@
...
+  NormalRubyCall() { ruby_call_def(this, method) }
+
+  override RubyMethod getMethod() { result = method }
+}
+
+class ImplicitRubyCall extends RubyCall, @ruby_call {
+  ImplicitRubyCall() { ruby_call_def(this, any(@ruby_argument_list a)) }
+
+  override RubyMethod getMethod() { none() }
+}
```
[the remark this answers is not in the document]
aibaars: Yes, this can be deleted.

ruby/ql/lib/upgrades/9fdd1d40fd3c3f8f9db8fabf5a353580d14c663a/ruby_scope_resolution_def.ql
aibaars: The `not ruby_call_def(_, this)` is still important. It says that things like `expr::Foo()` are not scope resolutions. Even though `Foo` looks like a constant name, it is not.

ruby/ql/lib/upgrades/9fdd1d40fd3c3f8f9db8fabf5a353580d14c663a/ruby_ast_node_info.ql
aibaars: A call like `expr::Foo(args) block` used to be represented as `Call{ method: ScopeResolution(expr, Foo), arguments: args, block: block}`. The children of the Call node were:
1. ScopeResolution(expr, Foo)
2. args
3. block
The new parse tree is like `Call{ receiver: expr, method: Foo, arguments: args }` . For this Call the children should be
1. expr
2. Foo
3. args
4. block

To fix the parent child relation for the Call we simply take the children of the `ScopeResolution` first (`x=0`) ordered by `oldIndex` followed by (`x=1`) the other children  of the original Call in the `oldIndex` order. The `rank` is used to ensure the new index is consecutive even if some of the children are missing.

nickrolfe (approved):
Very nice!

ruby/ql/lib/codeql/ruby/ast/internal/AST.qll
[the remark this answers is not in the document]
aibaars: Yes indeed. Both are a `Call` but their `getOperator()` will return a different token.

PR #8912, opened 2022-04-27
C++: Fix IR variable reuse for global var inits

rdmarsh2:
Fixes a performance issue where all `IRVariable`s corresponding to a given global variable were being used in the initializer IRFunction for that global variable.


jketema:
I'm betting the C++ tests are going to fail, because the `expected` files revert some of the changes that were merged for EDG 6.3.

dbartol (dismissed):
LGTM as long as the tests pass.

jketema:
I fixed the test regressions caused by running against the old frontend. However, it turns out, also some syntax zoo tests broke (unrelated to the frontend). I could update those, but can tell if the changes are correct.

Also the IR `VariableAddress`es with names look funny especially for declarations like:
```
int global_6 = global_2;
```
This becomes:
```
# 1769| int global_6
# 1769|   Block 0
# 1769|     v1769_1(void)       = EnterFunction     : 
# 1769|     mu1769_2(unknown)   = AliasedDefinition : 
# 1769|     r1769_3(glval<int>) = VariableAddress   : 
# 1769|     r1769_4(glval<int>) = VariableAddress   : 
# 1769|     r1769_5(int)        = Load[?]           : &:r1769_4, ~m?
# 1769|     mu1769_6(int)       = Store[?]          : &:r1769_3, r1769_5
# 1769|     v1769_7(void)       = ReturnVoid        : 
# 1769|     v1769_8(void)       = AliasedUse        : ~m?
# 1769|     v1769_9(void)       = ExitFunction      : 
```
Which `VariableAddress` is which here?

rdmarsh2:
> Also the IR `VariableAddress`es without names look funny especially for declarations like:

Fixed by d1c6022

jketema:
> Fixed by https://github.com/github/codeql/commit/d1c6022efc7b6948970507ac10fe633c998a6b80

Thanks. Would it be possible to add `int global_6 = global_2;` as a test?

rdmarsh2:
Good call - I've added the test and fixed another case of missing variables

rdmarsh2:
I don't think we do... there hasn't been much else happening on the dataflow side

PR #8930, opened 2022-04-28
QL language reference: variables must be lowerId

jbj:
To prepare for a future QL language change where variable names must start with a lower-case letter, this commit updates the QL language reference (including the language specification) to change the variable name grammar from `simpleId` to `lowerId`.

To find the locations I should change, I grepped for `variable` and `simpleId`. I also looked manually through [the reference section on variables](https://codeql.github.com/docs/ql-language-reference/variables/).

### DIFF
--- a/docs/codeql/ql-language-reference/ql-language-specification.rst
+++ b/docs/codeql/ql-language-reference/ql-language-specification.rst
@@ -364,7 +364,7 @@ A *variable declaration list* provides a sequence of variables and a type for ea
 ::
 
    var_decls ::= (var_decl ("," var_decl)*)?
-   var_decl ::= type simpleId
+   var_decl ::= type lowerId
 
 A valid variable declaration list must not include two declarations with the same variable name. Moreover, if the declaration has a typing environment that applies, it must not use a variable name that is already present in that typing environment.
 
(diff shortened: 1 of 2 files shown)


jf205:
Thanks @jbj. I guess in theory there could be example CodeQL snippets in the documentation that use upper case variable names, but maybe they're not so easy to find. Let's assume there are very few and do ad hoc fixes if we spot any (or they are reported by users). I'll leave the proper review to @alexet.

I can't remember when the variable name change actually comes into effect. Be aware that the docs team updates codeql.github.com/docs for every CLI release, so we shouldn't merge this until we are in the release cycle where this change lands.

cc @felicitymay

jbj:
> I guess in theory there could be example CodeQL snippets in the documentation that use upper case variable names, but maybe they're not so easy to find. Let's assume there are very few and do ad hoc fixes if we spot any (or they are reported by users).

I commented yesterday on the internal issue about how I've looked for upper-case variables in our documentation. In summary, I think I've checked 80-90% of them.

> I can't remember when the variable name change actually comes into effect. Be aware that the docs team updates codeql.github.com/docs for every CLI release, so we shouldn't merge this until we are in the release cycle where this change lands.

If we stick with the timeline in this issue, we can start introducing warnings for upper-case variables on May 15. That means we should introduce those warnings in `main` within a week or so, and then they'll be in the next CLI release. Given that we're so close, I suggest merging this documentation change right away. It means that the compiler will be less strict than the spec for a few weeks, but I don't see a problem with that.

alexet (approved):

PR #8933, opened 2022-04-28
C++: Revert #8515

MathiasVP:
This was causing performance issues on some projects. We already have a fix for the performance [here](https://github.com/github/codeql/pull/8912), but it's causing some new consistency failures in dataflow. So let's just revert the PR for now.

This reverts commit 800e4ea7df38c26a26cbd97d89e6d2282b7183ac, reversing
changes made to 7ce040f33112e07f8e6cdb3ba7dc45f4181b1809.

DCA: https://github.com/github/codeql-dca-main/issues/4212


jketema (approved):
LGTM

PR #8936, opened 2022-04-28
QL: Improve camel case query

jketema:
* Improve displayed message. In the PR interface it was not always clear which name was referred to, especially in the case of classes, which can be many lines long
* Filter out `NewType` like `NewTypeBranch`. This showed up, e.g., here https://github.com/github/codeql/security/code-scanning/29265


erik-krogh (approved):
:+1:

PR #8943, opened 2022-04-28
C++: Remove import order workarounds

jbj:
These workarounds are no longer needed from CodeQL CLI 2.9.0.

So far I've checked that there's no re-evaluation by looking through the output of

```sh
codeql query compile --dump-dil cpp/ql/src/codeql-suites/cpp-security-and-quality.qls > dil.log 2>&1
```

Comparing this output before and after removing the imports, I can see that some auto-generated names change (suffixes like `#2` come and go), but that's not a problem on its own. The important thing is that there's no change in how many distinct copies of each predicate we evaluate. For this, I compared the output of the following command before and after, finding that there were no changes.

```sh
grep '^query ' dil.log |grep -v '^query #select' |sed 's/ =.*//' |sort -u |sed 's/#.*//' |uniq -c |sort -n
```

I'll start a DCA run to double-check. @alexet, can you validate whether my approach is sound? (I'm sure there's a better way to check for re-evaluation than these shell one-liners)

### DIFF
--- a/cpp/ql/lib/semmle/code/cpp/security/Overflow.qll
+++ b/cpp/ql/lib/semmle/code/cpp/security/Overflow.qll
@@ -5,8 +5,6 @@
 
 import cpp
 import semmle.code.cpp.controlflow.Dominance
-// `GlobalValueNumbering` is only imported to prevent IR re-evaluation.
-private import semmle.code.cpp.valuenumbering.GlobalValueNumbering
 import semmle.code.cpp.rangeanalysis.SimpleRangeAnalysis
 import semmle.code.cpp.rangeanalysis.RangeAnalysisUtils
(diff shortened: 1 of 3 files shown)


MathiasVP (approved):
LGTM if DCA is happy. Thanks for cleaning this up!

jbj:
> To what extent is this fixed for good? Can we stop worrying about import order related problems forever?

The main source of import order problems has been fixed. All remaining import order problems are either obscure or unknown.

If you observe problems with unexpected re-evaluation in CLI 2.9.0 and beyond, please report it to my team.

MathiasVP:
The Wireshark failure is unrelated to this PR. The DCA run looks great 🎉.

jketema:
As the DCA looked ok, merging this.

Issue #8846, opened 2022-04-24
False Negatives - Prototype Pollution


dottif:
Hi, here are some prototype pollution false negatives.

### 1 Array / Rest parameters taint analysis
In the [extend function](https://github.com/eggjs/extend2/blob/ddefb57595a369eeec88d001786161389e1ee406/index.js#L64) from the extend2 package, if the parameters are accessed using ```arguments``` (first function) then CodeQL can detect the vulnerability. ([LGTM](https://lgtm.com/projects/g/dottif/prototype-pollution-static-analysis/snapshot/5d946562e24c36607c648e5d2e1540e08268e032/files/case_studies/extend2/index.js?sort=name&dir=ASC&mode=heatmap#x5c3dbd400bc04a82:1))
However if we implement the function using [rest parameters](https://lgtm.com/projects/g/dottif/prototype-pollution-static-analysis/latest/files/case_studies/extend2/index-rest.js?sort=name&dir=ASC&mode=heatmap#L64), or by directly passing an array (third function), the query won’t detect the vulnerability.
```javascript
function extend(){ //ok
  var target = arguments[0];
  target[name] = extend(deep, clone, copy);
}

function extend(...args){ //false negative
  var target = args[0];
  target[name] = extend(deep, clone, copy);
}

function extend(args){ //false negative
  var target = args[0];
  target[name] = extend([deep, clone, copy]);
}
```
### 2 Js-data deepFillIn
This vulnerability is not reported: [LGTM](https://lgtm.com/projects/g/dottif/prototype-pollution-static-analysis/snapshot/5d946562e24c36607c648e5d2e1540e08268e032/files/case_studies/js-data/utilsmod.js?sort=name&dir=ASC&mode=heatmap#L55)
### 3 Mutiny
Also this one is a false negative [LGTM](https://lgtm.com/projects/g/dottif/prototype-pollution-static-analysis/latest/files/case_studies/mutiny/mutiny-original.js)
###### source: https://github.com/BlackFan/client-side-prototype-pollution/blob/master/pp/mutiny.md

kaeluka:
Thank you for that issue, @dottif! You've provided very thorough documentation.

As a first step, I have confirmed that there seems to be a bug. In the three snippets below, we catch the first one, but not the others.

```javascript
function copyUsingForInAndArguments() {
    const dst = arguments[0];
    const src = arguments[1];
    for (let key in src) {
        if (dst[key]) {
[...]

PR #8946, opened 2022-04-28
JS: fix a FN for prototype polluting function query

kaeluka:
Fixes another FN from #8846 (CVE-2021-23574).

The bottom line is, that using iteration like this used to work:

```javascript
function forEachPropNoTempVar(obj, callback) {
    const keys = Object.keys(obj)
    const len = keys.length
    for (let i = 0; i < len; i++) {
        const key = keys[i];
        callback(key, obj[key])
    }
}
```

But iteration like that did not (now fixed):
```javascript
function forEachPropNoTempVar(obj, callback) {
    const keys = Object.keys(obj)
    const len = keys.length
    for (let i = 0; i < len; i++) {
        // no local variable used
        callback(keys[i], obj[keys[i]])
    }
}
```


kaeluka:
The experiments came back; looks all good.

erik-krogh (approved):
:+1:

PR #8952, opened 2022-04-28
Fix syntax errors in QL comments

cklin:
This PR fixes QL comments used to generate QL documentation.  They are mostly about back-quoting angle brackets so that they are property escaped in the HTML output.

### DIFF
--- a/csharp/ql/lib/semmle/code/csharp/exprs/Call.qll
+++ b/csharp/ql/lib/semmle/code/csharp/exprs/Call.qll
@@ -175,8 +175,8 @@ class Call extends DotNet::Call, Expr, @call {
    * - Line 10: The static target is `Type.InvokeMember()`, whereas the run-time targets
    *   are both `A.M()` and `B.M()`.
    *
-   * - Line 16: There is no static target (delegate call) but the delegate `i => { }` (line
-   *   20) is a run-time target.
+   * - Line 16: There is no static target (delegate call) but the delegate `i => { }`
+   *   (line 20) is a run-time target.
    */
   override Callable getARuntimeTarget() {
     exists(DispatchCall dc | dc.getCall() = this | result = dc.getADynamicTarget())
--- a/csharp/ql/src/Documentation/XmldocExtraTypeParam.ql
+++ b/csharp/ql/src/Documentation/XmldocExtraTypeParam.ql
@@ -1,6 +1,6 @@
 /**
  * @name Incorrect type parameter name in documentation
- * @description The type parameter name given in a '<typeparam>' tag does not exist. Rename the parameter or
+ * @description The type parameter name given in a `<typeparam>` tag does not exist. Rename the parameter or
  *              change the name in the documentation to ensure that they are the same.
  * @kind problem
  * @problem.severity recommendation
--- a/javascript/ql/lib/semmle/javascript/Externs.qll
+++ b/javascript/ql/lib/semmle/javascript/Externs.qll
@@ -53,7 +53,7 @@ import javascript
  *
  * /**
  *  * @param {!Object} obj
- *  * @return {!Array<string>}
+ *  * @return {!Array&lt;string&gt;}
  *  *&#47;
  * Object.keys = function(obj) {};
  *
@@ -109,7 +109,7 @@ class ExternalTypedef extends ExternalDecl, VariableDeclarator {
  *
  * /**
  *  * @param {!Object} obj
- *  * @return {!Array<string>}
+ *  * @return {!Array&lt;string&gt;}
  *  *&#47;
  * Object.keys = function(obj) {};
  *
@@ -214,7 +214,7 @@ class ExternalGlobalVarDecl extends ExternalGlobalDecl, VariableDeclarator {
  * <pre>
  * /**
  *  * @param {!Object} obj
- *  * @return {!Array<string>}
+ *  * @return {!Array&lt;string&gt;}
  *  *&#47;
  * Object.keys = function(obj) {};
  *
@@ -273,7 +273,7 @@ class ExternalMemberDecl extends ExternalVarDecl, ExprStmt {
  * <pre>
  * /**
  *  * @param {!Object} obj
- *  * @return {!Array<string>}
+ *  * @return {!Array&lt;string&gt;}
  *  *&#47;
  * Object.keys = function(obj) {};
  *
--- a/cpp/ql/lib/semmle/code/cpp/Function.qll
+++ b/cpp/ql/lib/semmle/code/cpp/Function.qll
@@ -38,8 +38,8 @@ class Function extends Declaration, ControlFlowNode, AccessHolder, @function {
    * int z = min(5, 7);
    * ```
    * The full signature of the function called on the last line would be
-   * "min<int>(int, int) -> int", and the full signature of the uninstantiated
-   * template on the first line would be "min<T>(T, T) -> T".
+   * `min<int>(int, int) -> int`, and the full signature of the uninstantiated
+   * template on the first line would be `min<T>(T, T) -> T`.
    */
   string getFullSignature() {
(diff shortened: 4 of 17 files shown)


javascript/ql/lib/semmle/javascript/Externs.qll
```diff
@@ -53,7 +53,7 @@ import javascript
  *
  * /**
  *  * @param {!Object} obj
- *  * @return {!Array<string>}
+ *  * @return {!Array&lt;string&gt;}
```
aschackmull, line 56: These are wrapped in `<pre>...</pre>`, so are these changes necessary? Changing `<` to `&lt;` hurts readability.

aschackmull:
Most of the changes look fine, so mostly no objections from me. Except for the `<`-to-`&lt;`-rewrite, which I think hurts readability. I also wonder why these changes are even necessary - our qldoc is mostly markdown, so would it instead make sense to simply remove html support in our qldoc? (I know that's likely a much larger change, but still worth a thought.) Then a translation into html could simply do whatever escaping was needed without qldoc writers having to worry about accidental html interpretations of angle brackets.

csharp/ql/lib/semmle/code/csharp/exprs/Call.qll
```diff
@@ -175,8 +175,8 @@ class Call extends DotNet::Call, Expr, @call {
    * - Line 10: The static target is `Type.InvokeMember()`, whereas the run-time targets
    *   are both `A.M()` and `B.M()`.
    *
-   * - Line 16: There is no static target (delegate call) but the delegate `i => { }` (line
-   *   20) is a run-time target.
+   * - Line 16: There is no static target (delegate call) but the delegate `i => { }`
+   *   (line 20) is a run-time target.
```
hvitved, line 179: I don't mind this change, but hopefully it is not needed to generete proper QL doc?

csharp/ql/src/Documentation/XmldocExtraTypeParam.ql
```diff
@@ -1,6 +1,6 @@
 /**
  * @name Incorrect type parameter name in documentation
- * @description The type parameter name given in a '<typeparam>' tag does not exist. Rename the parameter or
+ * @description The type parameter name given in a `<typeparam>` tag does not exist. Rename the parameter or
```
hvitved, line 3: If this is the new way of doing things, we should update the comment `Any code elements included in the description should be enclosed in single quotes` (https://github.com/github/codeql/blob/main/docs/query-metadata-style-guide.md?plain=1#L63), as well as the example.

MathiasVP (approved):
C/C++ 👍

javascript/ql/lib/semmle/javascript/Externs.qll
cklin: Reserved characters such as `<` and `>` still needs to be escaped using their HTML entity.  https://developer.mozilla.org/en-US/docs/Web/HTML/Element/pre

You can see that `@return {!Array<string>}` currently shows up as `@return {!Array}` because the browser treats `<string>` as an unknown element.
https://codeql.github.com/codeql-standard-libraries/javascript/semmle/javascript/Externs.qll/type.Externs$ExternalDecl.html

csharp/ql/lib/semmle/code/csharp/exprs/Call.qll
cklin: It actually is needed.  Here is what the rendered documentation currently looks like, without this fix:

https://codeql.github.com/codeql-standard-libraries/csharp/semmle/code/csharp/exprs/Call.qll/predicate.Call$Call$getARuntimeTarget.0.html
line-20

csharp/ql/src/Documentation/XmldocExtraTypeParam.ql
cklin: I should clarify that this is not a style change but a rendering fix.  Here is what the documentation currently looks like:

https://codeql.github.com/codeql-standard-libraries/csharp/Documentation/XmldocExtraTypeParam.ql/module.XmldocExtraTypeParam.html
typeparam

QL comments can include Markdown and HTML, so the `<` and `>` characters need to be escaped, either by putting them in Markdown code format (the approach I took here) or by replacing them with `&lt;` and `&gt;`.  I can switch it to the second approach if that is more style-compliant.

I agree that using Markdown backquotes for code elements seems sensible, but that should probably be a separate discussion.

aschackmull:
I think it is a problem that we allow html in our qldoc, we should restrict ourselves to markdown and then we wouldn't need such readability-impeding changes like writing `&lt;`. Could we just get rid of html in our qldoc? Then any html rendering could just be plain markdown-to-html.

csharp/ql/src/Documentation/XmldocExtraTypeParam.ql
aschackmull: > this is not a style change but a rendering fix

No, it's _both_ a style change and a rendering fix.

csharp/ql/lib/semmle/code/csharp/exprs/Call.qll
aschackmull: Any explanation why it renders that way? Is it a rendering bug?

javascript/ql/lib/semmle/javascript/Externs.qll
aschackmull: Could we replace the `<pre>` with triple backquotes instead?

cklin: You could, but the `*&#47;` in the next line would stop working.

csharp/ql/lib/semmle/code/csharp/exprs/Call.qll
cklin: I think that is because `20)` is interpreted as a numbered list item starting at number 20.

cklin:
I intended this PR to be a quick low-effort fix for QL documentation that is already broken.  Unfortunately, I really do not have more time to devote to this PR or the issue of broken documentation, so I will just have to leave this PR as-is.

I am sure that you all as the owners of the documentation will be able to resolve the open questions and find the best way to address both the immediate and the long-term issues.  Thank you!  🙏

erik-krogh (approved):
JS :+1:  

The change in `Externs.qll` makes it slightly harder to read the source code.  
But it seems like the best solution, because the rendered documentation should definitely render correctly.

javascript/ql/lib/semmle/javascript/Externs.qll
aschackmull: See - I have no idea what entity number 47 is - I had to look that up, that's pretty poor readability. Luckily the obvious answer is just to replace `&#47;` with `/`.

csharp/ql/lib/semmle/code/csharp/exprs/Call.qll
aschackmull: Ah, that makes a bit more sense then. A markdown gotcha.

aschackmull (changes requested):
As commented above.

javascript/ql/lib/semmle/javascript/Externs.qll
aschackmull: Hmm, right, I see why the entity is used for `/`. But the `<pre>` isn't really a good solution - it doesn't appear to render correctly in VSCode (on predicate mouse-over).

aschackmull (approved):
The change in `javascript/ql/lib/semmle/javascript/Externs.qll` is unfortunate and makes me want to abolish html entirely from our qldoc, but currently I can't think of a way to render `*/` inside a triple-backtick code block. Note that the `<pre>` block doesn't render correctly in VSCode, so going for a markdown solution would be preferable, but I'll revert my position on whether just to merge this PR for now.

Issue #8627, opened 2022-03-31
LGTM.com - false positive


sungam3r:
https://github.com/graphql-dotnet/graphql-dotnet/security/code-scanning/1038

After "fix":

> Expression = (context => argumentDelegate(context)); // CS8917	The delegate type could not be inferred.

hvitved:
Hi.

Are you using C# 10? Because it appears that the cast [is in fact redundant](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/operators/lambda-expressions#natural-type-of-a-lambda-expression) if you do. Our C# extractor is based on C# 10, so our analysis will flag the cast as redundant. I documented this with [a couple of tests](https://github.com/github/codeql/pull/8674), which would have failed extraction if the casts were needed.

sungam3r:
Yes, I use C#10. Documentation you provided says that compiler _may_ infer delegate type (and in some cases it can not do that). `Expression` property (left side of assigment) is of type `LambdaExpression?` in my example. Your test does not cover this case.

sungam3r:
More info. Actually that syntax works:
```c#
Expression = (IResolveFieldContext context) => argumentDelegate(context);
```

And that does not:
```c#
Expression = context => argumentDelegate(context); // CS8917 The delegate type could not be inferred.
```

Compiler needs to know input type here.

hvitved:
Sorry for the late reply.

I hadn't realized that you were using the untyped `LambdaExpression` type. I have been able to reproduce locally, and I have recorded the false positive [here](https://github.com/github/codeql/pull/8955).

sungam3r:
👍

PR #8955, opened 2022-04-29
C#: Add FP test for `cs/useless-cast-to-self`

hvitved:
Documents the FP reported [here](https://github.com/github/codeql/issues/8627).


michaelnebel (approved):
Looks plausible to me.

PR #8956, opened 2022-04-29
Java: Fix Intent Redirection sanitizer

atorralba:
When validating intents to protect against intent redirection, both the package name and the class name must be checked. Otherwise, if only the package name is checked, the intent can be redirected to any of the app's activities (including non-exported ones), whereas if only the class name is checked, the intent can be redirect to an activity with the same name in another app. Both things are dangerous and should be flagged by the intent redirection query.

This PR adds more restrictions to the appropriate intent redirection sanitizer to take the above into account.


smowton (approved):
Looks reasonable to me

PR #8959, opened 2022-04-29
Swift: teach bazel to install python dependencies

AlexDenisov:


redsun82 (approved):
Really great stuff! Wait for it, and without you even noticing you'll become a bazel expert despite everything :stuck_out_tongue_winking_eye:

PR #8971, opened 2022-04-29
Ruby: add safe navigation operator

aibaars:


hvitved (approved):
Looks great! Let's see a successful DCA run before merging.

PR #8975, opened 2022-04-29
Ruby: fix some flow summary join orders

nickrolfe:
The flow summaries that are implemented with an abstract base class restricting the method name, and child classes using that method name, had unfortunate join orders:

```
r1 = JOIN Call::MethodCall::getMethodName#dispred#f0820431#ff WITH Call::MethodCall::getMethodName#dispred#f0820431#ff ON FIRST 1 OUTPUT Lhs.0, (Lhs.1 ++ "_arg"), Rhs.1
```

They weren't causing catastrophic slowdowns, but fixing them should give a small speedup.

### DIFF
--- a/ruby/ql/lib/codeql/ruby/frameworks/core/Array.qll
+++ b/ruby/ql/lib/codeql/ruby/frameworks/core/Array.qll
@@ -199,7 +199,7 @@ module Array {
   /** A call to `[]`, or its alias, `slice`. */
   abstract private class ElementReferenceReadSummary extends SummarizedCallable {
     MethodCall mc;
-    ElementReferenceReadMethodName methodName;
+    ElementReferenceReadMethodName methodName; // adding this as a field helps give a better join order
 
     bindingset[this]
     ElementReferenceReadSummary() { mc.getMethodName() = methodName }
@@ -2018,7 +2018,7 @@ module Enumerable {
 
   abstract private class GrepSummary extends SummarizedCallable {
     MethodCall mc;
-    GrepMethodName methodName;
+    GrepMethodName methodName; // adding this as a field helps give a better join order
 
     bindingset[this]
     GrepSummary() { mc.getMethodName() = methodName }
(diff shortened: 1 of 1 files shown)


nickrolfe:
DCA results don't seem very interesting - probably a small speedup, but not big enough to register as 'interesting'.

ruby/ql/lib/codeql/ruby/frameworks/core/Array.qll
```diff
@@ -266,11 +271,11 @@
~          rl.isInclusive() and end = e
           or
           rl.isExclusive() and end = e - 1
         ) and
-        this = "[](" + start + ".." + end + ")"
+        this = methodName + "(" + start + ".." + end + ")"
```
hvitved, line 275: This means that we are now synthesizing two methods, whereas before there was just one. But that should be OK.

ruby/ql/lib/codeql/ruby/frameworks/core/Array.qll
```diff
@@ -2052,18 +2062,23 @@
...
   }
 
+  private class InjectMethodName extends string {
+    // `reduce` is an alias for `inject`.
+    InjectMethodName() { this = ["inject", "reduce"] }
+  }
+
   abstract private class InjectSummary extends SummarizedCallable {
     MethodCall mc;
+    InjectMethodName methodName; // adding this as a field helps give a better join order
```
hvitved, line 2072: Either remove comment, or add to all fields (e.g. the one in `GrepSummary` doesn't have the comment).

hvitved (commented):
Looks great!

ruby/ql/lib/codeql/ruby/frameworks/core/Array.qll
nickrolfe: Right, I forgot to mention it but I thought I should make the change for consistency.

ruby/ql/lib/codeql/ruby/frameworks/core/Array.qll
nickrolfe: OK, I've added the comment to all the fields.

hvitved (approved):

PR #8990, opened 2022-04-30
JS: Selection API DOM text source

bananabr:
A Selection object represents the range of text selected by the user or the caret's current position.  The Selection API can be used by applications and UI components such as rich-text editors to operate over a user's text selection. The current DOM text sources do not include this API.

### DIFF
--- a/javascript/ql/lib/change-notes/2022-04-30-xss-selection-source.md
+++ b/javascript/ql/lib/change-notes/2022-04-30-xss-selection-source.md
@@ -1,0 +1,4 @@
+---
+category: minorAnalysis
+---
+* Added the `Selection` api as a DOM text source in the `XssThroughDomCustomizations` library.
--- a/javascript/ql/lib/change-notes/2022-04-30-xss-selection-source.md
+++ b/javascript/ql/lib/change-notes/2022-04-30-xss-selection-source.md
@@ -1,4 +1,4 @@
 ---
 category: minorAnalysis
 ---
-* Added the `Selection` api as a DOM text source in the `XssThroughDomCustomizations` library.
+* Added the `Selection` api as a DOM text source in the `js/xss-through-dom` query.
--- a/javascript/ql/lib/semmle/javascript/security/dataflow/XssThroughDomCustomizations.qll
+++ b/javascript/ql/lib/semmle/javascript/security/dataflow/XssThroughDomCustomizations.qll
@@ -219,7 +219,7 @@ module XssThroughDom {
 
 
   /**
-   * A source for text from the DOM from a Selection object toString method call
+   * A call to window.getSelection
    * https://developer.mozilla.org/en-US/docs/Web/API/Selection
    */
   DataFlow::SourceNode getSelectionCall(DataFlow::TypeTracker t) {
@@ -233,6 +233,10 @@ module XssThroughDom {
     exists(DataFlow::TypeTracker t2 | result = getSelectionCall(t2).track(t2, t))
   }
   
+  /**
+   * A source for text from the DOM from a Selection object toString method call
+   * https://developer.mozilla.org/en-US/docs/Web/API/Selection
+   */
   class SelectionSource extends Source {
     SelectionSource() {
       this = getSelectionCall(DataFlow::TypeTracker::end()).getAMethodCall("toString")
--- a/javascript/ql/lib/semmle/javascript/security/dataflow/XssThroughDomCustomizations.qll
+++ b/javascript/ql/lib/semmle/javascript/security/dataflow/XssThroughDomCustomizations.qll
@@ -226,6 +226,8 @@ module XssThroughDom {
     t.start() and
     exists(DataFlow::CallNode call |
       call = DataFlow::globalVarRef("getSelection").getACall()
+      or
+      call = DOM::documentRef().getAMemberCall("getSelection")
     |
       result = call
     )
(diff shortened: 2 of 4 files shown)


javascript/ql/lib/semmle/javascript/security/dataflow/XssThroughDomCustomizations.qll
```diff
@@ -216,4 +216,30 @@ module XssThroughDom {
       }
     }
   }
+
+
+  /**
+   * A call to window.getSelection
+   * https://developer.mozilla.org/en-US/docs/Web/API/Selection
```
erik-krogh, line 223:
```suggestion
   * Gets a reference to a value obtained by calling `window.getSelection()`. 
   * https://developer.mozilla.org/en-US/docs/Web/API/Selection
```

Just a quick comment based on the feedback from the CI.  
Also the autoformatter is failing on `XssThroughDomCustomizations.qll`.

javascript/ql/lib/semmle/javascript/security/dataflow/XssThroughDomCustomizations.qll
```diff
@@ -216,4 +216,30 @@ module XssThroughDom {
...
+
+  /**
+   * Gets a reference to a value obtained by calling `window.getSelection()`. 
+   * https://developer.mozilla.org/en-US/docs/Web/API/Selection
+   */
+  DataFlow::SourceNode getSelectionCall(DataFlow::TypeTracker t) {
+    t.start() and
+    exists(DataFlow::CallNode call |
+      call = DataFlow::globalVarRef("getSelection").getACall()
+    |
```
erik-krogh, line 229:
```suggestion
    exists(DataFlow::CallNode call |
      call = DataFlow::globalVarRef("getSelection").getACall()
      or
      call = DOM::documentRef().getAMemberCall("getSelection")
    |
``` 
Two things: 
 1: The autoformatter was complaining about this line. 
 2: The `getSelection()` method also exists on the `Document` object.  
 
For the last point: I usually check [lib.dom.d.ts](https://github.com/microsoft/TypeScript/blob/main/lib/lib.dom.d.ts) when I need to check which DOM APIs exist.  
(I also checked that file when reviewing your previous PR).  

Could you also add a test for the `document.getSelection()` method?

javascript/ql/lib/change-notes/2022-04-30-xss-selection-source.md
```diff
@@ -0,0 +1,4 @@
+---
+category: minorAnalysis
+---
+* Added the `Selection` api as a DOM text source in the `XssThroughDomCustomizations` library.
```
erik-krogh, line 4:
```suggestion
* Added the `Selection` api as a DOM text source in the `js/xss-through-dom` query.
``` 
I think it's more relevant to mention which query it affects.

javascript/ql/lib/semmle/javascript/security/dataflow/XssThroughDomCustomizations.qll
```diff
@@ -216,4 +216,30 @@ module XssThroughDom {
...
+      result = call
+    )
+    or
+    exists(DataFlow::TypeTracker t2 | result = getSelectionCall(t2).track(t2, t))
+  }
+  
+  /**
+   * A source for text from the DOM from a Selection object toString method call
+   * https://developer.mozilla.org/en-US/docs/Web/API/Selection
+   */
```
erik-krogh, line 239:
```suggestion
  /**
   * A source for text from the DOM from calling `toString()` on a `Selection` object.
   * The `toString()` method returns the currently selected text in the DOM.
   * https://developer.mozilla.org/en-US/docs/Web/API/Selection
   */
```

I was initially confused by the `toString()` call.   
Because usually `toString()` is used as taint-step, and not a taint-source.   
But in this case using `toString()` as a source is correct.   
So I tried to add a thing to the QLDoc to make that more clear.

javascript/ql/lib/semmle/javascript/security/dataflow/XssThroughDomCustomizations.qll
bananabr: Sure can. I will work on tests later today.
I am sorry about the auto-format error. Usually I am a formatting freak, don't know what happened here =)
Thanks for the [lib.dom.d.ts](https://github.com/microsoft/TypeScript/blob/main/lib/lib.dom.d.ts) tip, I had no idea the getSelection method was available through the **document** object as well.

erik-krogh: > Sure can. I will work on tests later today. 

Great. I'll get the CI to run when you've pushed that.

javascript/ql/lib/semmle/javascript/security/dataflow/XssThroughDomCustomizations.qll
```diff
@@ -216,4 +216,33 @@ module XssThroughDom {
       }
     }
   }
+
+
+  /**
```
erik-krogh, line 221:
```suggestion

  /**
```
There is a newline too much here, and I think that's what the autoformatter is complaining about this time.

erik-krogh (approved):
[Evaluation was uneventful](https://github.com/github/codeql-dca-main/tree/data/erik-krogh/pr-8990-bf6663a__nightly-old__CustomSuite/reports) (internal link).  

LGTM :+1:

PR #8994, opened 2022-05-01
C/C++ : Wrong Uint access

HansmannThibaut:
This request find all array access that use the wrong kind of Uint. For instance, 
```
Uint_16t i = 16;
char test[100]; 
test = 100;
```
This query found results on Suricata. It's also part of a project to find bug or bad behaviour in Suricata.

### DIFF
--- a/cpp/ql/src/experimental/Best Practices/WrongUintAccess.qhelp
+++ b/cpp/ql/src/experimental/Best Practices/WrongUintAccess.qhelp
@@ -1,0 +1,18 @@
+<!DOCTYPE qhelp PUBLIC
+  "-//Semmle//qhelp//EN"
+  "qhelp.dtd">
+<qhelp>
+
+<overview>
+<p>Find access to an array with a Uint16 when the array has a size lower than 256.</p>
+</overview>
+
+<recommendation>
+<p>Use a Uint8 instead</p>
+</recommendation>
+
+<example>
+<sample src="WrongUintAcess.cpp" />
+</example>
+
+</qhelp>
--- a/cpp/ql/src/experimental/Best Practices/WrongUintAccess.ql
+++ b/cpp/ql/src/experimental/Best Practices/WrongUintAccess.ql
@@ -1,0 +1,23 @@
+/**
+ * @id cpp/wrong-uint-access
+ * @name Wrong Uint
+ * @descripion Acess an array of size lower than 256 with a uint16.
+ * @kind problem
+ * @problem.severity recommendation
+ * @tags efficiency
+ */
+
+import cpp
+import semmle.code.cpp.controlflow.SSA
+
+from
+  Variable E, ArrayExpr useExpr, ArrayType defExpr, VariableDeclarationEntry def, VariableAccess use
+where
+  def = defExpr.getATypeNameUse() and
+  E = def.getDeclaration() and
+  use = useExpr.getArrayBase() and
+  E = use.getTarget() and
+  useExpr.getArrayOffset().getType() instanceof UInt16_t and
+  defExpr.getArraySize() <= 256
+select useExpr, "Using a UInt16_t to acess the array $@ of size " + defExpr.getArraySize() + ".", E,
+  E.getName()
--- a/cpp/ql/src/experimental/Best Practices/WrongUintAcess.cpp
+++ b/cpp/ql/src/experimental/Best Practices/WrongUintAcess.cpp
@@ -1,0 +1,7 @@
+void test()
+{
+    uint16_t j = 256;
+    char testSubject[122];
+
+    testSubject[j] = 12; // You can use a uint8 here
+}


HansmannThibaut:
Supervised by @catenacyber

catenacyber:
@HansmannThibaut the example you gave is missing the index used for array access ie `test[i]`

cpp/ql/src/experimental/Best Practices/WrongUintAccess.ql
```diff
@@ -0,0 +1,23 @@
...
+import semmle.code.cpp.controlflow.SSA
+
+from
+  Variable E, ArrayExpr useExpr, ArrayType defExpr, VariableDeclarationEntry def, VariableAccess use
+where
+  def = defExpr.getATypeNameUse() and
+  E = def.getDeclaration() and
+  use = useExpr.getArrayBase() and
+  E = use.getTarget() and
+  useExpr.getArrayOffset().getType() instanceof UInt16_t and
```
catenacyber, line 20: Could we have something more generic here ? Like uint16 but also uint32 and uint64 ?

cpp/ql/src/experimental/Best Practices/WrongUintAccess.ql
```diff
@@ -0,0 +1,23 @@
...
+ * @kind problem
+ * @problem.severity recommendation
+ * @tags efficiency
+ */
+
+import cpp
+import semmle.code.cpp.controlflow.SSA
+
+from
+  Variable E, ArrayExpr useExpr, ArrayType defExpr, VariableDeclarationEntry def, VariableAccess use
```
MathiasVP, line 14: Two comments about this line:
- We're _just_ in the process of deprecating uppercase identifiers. See https://github.com/github/codeql/pull/8930. So I'd appreciate it if you'd rename `E` to something lowercase. `v` is always a great name for a `Variable` if you can't think of anything better 😂
- I find it slightly strange that `defExpr` is a _type_ (and not an `Expr` like `useExpr`).

cpp/ql/src/experimental/Best Practices/WrongUintAccess.ql
```diff
@@ -0,0 +1,23 @@
...
+ * @id cpp/wrong-uint-access
+ * @name Wrong Uint
+ * @descripion Acess an array of size lower than 256 with a uint16.
+ * @kind problem
+ * @problem.severity recommendation
+ * @tags efficiency
+ */
+
+import cpp
+import semmle.code.cpp.controlflow.SSA
```
MathiasVP, line 11: I don't think you're actually using the SSA library for anything, are you?

cpp/ql/src/experimental/Best Practices/WrongUintAccess.ql
```diff
@@ -0,0 +1,23 @@
...
+
+from
+  Variable E, ArrayExpr useExpr, ArrayType defExpr, VariableDeclarationEntry def, VariableAccess use
+where
+  def = defExpr.getATypeNameUse() and
+  E = def.getDeclaration() and
```
MathiasVP, line 17: `getATypeNameUse` is really a best-effort predicate that likely won't give you all the cases you're interested in. As far as I can tell, you're really only interested in knowing that `def` is an `ArrayExpr` with some size, right? In that case, I think you could just do: `E.getType() = defExpr`. Instead of using `getType()` you could also use `getUnspecifiedType()` to automatically handle typedefs and `const`/`volatile` specifiers.

MathiasVP (commented):
Thanks for your contribution, @HansmannThibaut. Here are a couple of initial comments.

cpp/ql/src/experimental/Best Practices/WrongUintAccess.ql
HansmannThibaut: Done

cpp/ql/src/experimental/Best Practices/WrongUintAccess.ql
HansmannThibaut: Done
But I'm not sure for the name of defExpr that I changed to defLine. I think that it's more understandable but I'm not sure if it's enough.

cpp/ql/src/experimental/Best Practices/WrongUintAccess.ql
HansmannThibaut: Done

cpp/ql/src/experimental/Best Practices/WrongUintAccess.ql
HansmannThibaut: Done

cpp/ql/src/experimental/Best Practices/WrongUintAccess.ql
```diff
@@ -0,0 +1,27 @@
...
+ * @descripion Acess an array of size lower than 256 with a uint16.
+ * @kind problem
+ * @problem.severity recommendation
+ * @tags efficiency
+ */
+
+import cpp
+
+from
+  Variable var, ArrayExpr useExpr, VariableDeclarationEntry def, ArrayType defLine,
```
MathiasVP, line 13: I don't think you're using `def` anymore now, right?

HansmannThibaut: Yes it's not used again. I forgot to remove it, but it's done.

MathiasVP:
The code looks good now! Only a small formatting issue to fix now:
```
ql/cpp/ql/src/experimental/Best Practices/WrongUintAccess.ql would change by autoformatting.
```
Please see the <b>Formatting</b> section in https://github.com/github/codeql/blob/main/CONTRIBUTING.md#submitting-a-new-experimental-query.

HansmannThibaut:
Everything should be good now.

cpp/ql/src/experimental/Best Practices/WrongUintAccess.ql
MathiasVP: Awesome. Thanks a lot!

cpp/ql/src/experimental/Best Practices/WrongUintAccess.ql
MathiasVP: I would probably have gone with `arrayType` instead of `defLine`. But at this point, we're just splitting hairs :) So this is all fine!

cpp/ql/src/experimental/Best Practices/WrongUintAccess.qhelp
```diff
@@ -0,0 +1,18 @@
...
+<overview>
+<p>Find access to an array with a Uint16 when the array has a size lower than 256.</p>
+</overview>
+
+<recommendation>
+<p>Use a int with a lower bit size instead. For instance in this example use a 8 bit int.</p>
+</recommendation>
+
+<example>
+<sample src="WrongUintAcess.cpp" />
```
MathiasVP, line 15: This should fix the CI failure.
```suggestion
<sample src="WrongUintAccess.cpp" />
```

MathiasVP (approved):
LGTM!

PR #9002, opened 2022-05-02
Java: Add OkHttp and Retrofit models

atorralba:
Adds sinks of kind `open-url` and summaries for the libraries OkHttp and Retrofit.

Also simplifies the query `java/non-https-urls` to also consider sinks that are not method accesses.


java/ql/lib/semmle/code/java/frameworks/OkHttp.qll
```diff
@@ -0,0 +1,55 @@
...
+        "okhttp3;HttpUrl$Builder;false;addPathSegments;;;Argument[-1];ReturnValue;value",
+        "okhttp3;HttpUrl$Builder;false;addQueryParameter;;;Argument[-1];ReturnValue;value",
+        "okhttp3;HttpUrl$Builder;false;build;;;Argument[-1];ReturnValue;taint",
+        "okhttp3;HttpUrl$Builder;false;encodedFragment;;;Argument[-1];ReturnValue;value",
+        "okhttp3;HttpUrl$Builder;false;encodedPassword;;;Argument[-1];ReturnValue;value",
+        "okhttp3;HttpUrl$Builder;false;encodedPath;;;Argument[-1];ReturnValue;value",
+        "okhttp3;HttpUrl$Builder;false;encodedQuery;;;Argument[-1];ReturnValue;value",
+        "okhttp3;HttpUrl$Builder;false;encodedUsername;;;Argument[-1];ReturnValue;value",
+        "okhttp3;HttpUrl$Builder;false;fragment;;;Argument[-1];ReturnValue;value",
+        "okhttp3;HttpUrl$Builder;false;fragment;;;Argument[-1];ReturnValue;value",
```
smowton, line 38: Duplicate rows; should one be adding taint?

java/ql/lib/semmle/code/java/frameworks/OkHttp.qll
```diff
@@ -0,0 +1,55 @@
...
+        "okhttp3;HttpUrl$Builder;false;host;;;Argument[-1];ReturnValue;value",
+        "okhttp3;HttpUrl$Builder;false;password;;;Argument[-1];ReturnValue;value",
+        "okhttp3;HttpUrl$Builder;false;port;;;Argument[-1];ReturnValue;value",
+        "okhttp3;HttpUrl$Builder;false;query;;;Argument[-1];ReturnValue;value",
+        "okhttp3;HttpUrl$Builder;false;removeAllEncodedQueryParameters;;;Argument[-1];ReturnValue;value",
+        "okhttp3;HttpUrl$Builder;false;removeAllQueryParameters;;;Argument[-1];ReturnValue;value",
+        "okhttp3;HttpUrl$Builder;false;removePathSegment;;;Argument[-1];ReturnValue;value",
+        "okhttp3;HttpUrl$Builder;false;scheme;;;Argument[-1];ReturnValue;value",
+        "okhttp3;HttpUrl$Builder;false;scheme;;;Argument[0];Argument[-1];taint",
```
smowton, line 47: Comment on why only `scheme` contributes taint but not the other URL parts?

java/ql/lib/semmle/code/java/frameworks/OkHttp.qll
atorralba: This was a copy-paste mistake, thanks for spotting it. Let's discuss taint on the comment below.

```suggestion
        "okhttp3;HttpUrl$Builder;false;fragment;;;Argument[-1];ReturnValue;value",
```

java/ql/lib/semmle/code/java/frameworks/OkHttp.qll
atorralba: Good point. I added that while figuring out why `java/non-https-urls` wasn't working on an app that used these libraries, but forgot to handle taint more generally. Let me fix that.

atorralba: Done here 2d3b15f.

smowton (approved):
Approving as this looks plausible; one thing to check: I expect e.g. the url-injection query will distinguish calls that may affect the host or path being targeted from those that can only affect the query string or the fragment of the URL, perhaps by not modelling methods like `setFragment`, or perhaps by sanitising when a user-controlled input clearly has to follow a `?` or `#`. Check if we are doing that at the moment for queries that care about URLs, and if so how do we mirror that behaviour for these libraries (by selectively removing models? By adding sanitisers?)

atorralba:
I thought about this while adding these models, and it all boils down to what a URL being (generally) tainted means for us. For instance, there could be cases in which a tainted fragment is still dangerous (e.g. the fragment of a URL being added to HTML output could still cause a XSS), so we can't discard `setFragment` as a general taint-preserving step.

The fact that a query expects only certain kinds of taint is context-specific and I don't think should be handled by removing general purposed taint models. I see two possible solutions:

* Each query adds the appropriate exceptions (i.e. sanitizers) for non-interesting taint types (in this case, the sinks that are not relevant for SSRF).
* We handle this with synthetic fields that would represent the specific kinds of taint (e.g. for URLs, we could have synthetic fields for scheme, host, port, path, query and fragment, and the SSRF query would only allow implicit reads at its sinks of scheme, host and port).

I think I like the first option because it's less convoluted and a bit more obvious. WDYT?

smowton:
These functions generally speaking contributing taint and having sanitisers to exclude cases where only the fragment or query string can be tainted per-query sounds reasonable to me. I don't know whether the Java query suite does this much or at all, though I remember at least the Go queries doing this and likely Java as well, and would encourage you to make a quick survey of the current behaviour of URL-relevant queries before merging a PR that might produce inconsistent behaviour (i.e. some libraries excluding query-string taint and some not) between different ways of constructing a URL.

atorralba:
I looked into this, and it seems that Java queries assume that a tainted `java.net.URL/URI` object means that the hostname part is tainted. There are no specific sanitizers for URL/URI methods that set non-hostname parts (but there aren't general taint steps for these methods either). So it seems the intention has been to only taint URL/URI objects if the taint is relevant for the URL-related queries (i.e. the hostname is tainted).

This is not true however for Spring's `UriBuilder`/`UriComponents`, `JaxWS`'s `UriBuilder`, and Android's `Uri`/`Uri$Builder`, for which taint is propagated generally, regardless of which part of the URI is affected.

So this PR aggravates an already existent inconsistency indeed. The lazy fix is just removing the non-hostname rows from this PR and the library models mentioned above (at the risk of losing flow on specific cases as explained in my previous comment). The involved fix is adding the missing steps to the other models (mainly `java.net.URI` and `java.net.URL`) and adding the appropriate sanitizers for _all_ non-hostname models to the URL-related queries.

I still favor the second, more involved option, but this is starting to look like something that should be handled in a follow-up PR.

atorralba:
Another, maybe better option would be to use Flow States and require flow to go through hostname-altering steps (so an allow list approach, rather than a disallow list).

smowton:
Alright, in that case I don't object to merging this and following up appropriately, probably via a team talk about what the rule ought to be.

PR #8948, opened 2022-04-28
C++: Add support for SAXParser to the CWE-611 XXE query.

geoffw0:
Add test cases + support for `SAXParser` to the CWE-611 XXE query.

Change note covered by the one in recent PR https://github.com/github/codeql/pull/8736.

PR #9018, opened 2022-05-03
C++: Support libxml2 in the XXE query

geoffw0:
Add test cases + support for the `libxml2` library in the CWE-611 XXE query.

It turns out less data flow is required for checking this library than for Xerces, so much of the existing query logic isn't really used here.  However, I feel the results are best presented in a single place, not in two separate queries.

Change note covered by https://github.com/github/codeql/pull/8948.

### DIFF
--- a/cpp/ql/src/Security/CWE/CWE-611/XXE.ql
+++ b/cpp/ql/src/Security/CWE/CWE-611/XXE.ql
@@ -196,9 +196,9 @@ class Libxml2ParseCall extends FunctionCall {
 
   Libxml2ParseCall() {
     exists(string fname | this.getTarget().getName() = fname |
-      fname = ["xmlCtxtUseOptions"] and optionsArg = 1
+      fname = "xmlCtxtUseOptions" and optionsArg = 1
       or
-      fname = ["xmlReadFile"] and optionsArg = 2
+      fname = "xmlReadFile" and optionsArg = 2
       or
       fname = ["xmlCtxtReadFile", "xmlParseInNodeContext", "xmlReadDoc", "xmlReadFd"] and
       optionsArg = 3
@@ -207,7 +207,7 @@ class Libxml2ParseCall extends FunctionCall {
       or
       fname = ["xmlCtxtReadMemory", "xmlReadIO"] and optionsArg = 5
       or
-      fname = ["xmlCtxtReadIO"] and optionsArg = 6
+      fname = "xmlCtxtReadIO" and optionsArg = 6
     )
   }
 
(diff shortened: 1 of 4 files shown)


cpp/ql/src/Security/CWE/CWE-611/XXE.ql
```diff
@@ -188,6 +188,42 @@ class CreateLSParser extends Function {
...
+   * Gets the argument that specifies `xmlParserOption`s.
+   */
+  Expr getOptions() { result = this.getArgument(optionsArg) }
+}
+
+/**
+ * An `xmlParserOption` for `libxml2` that is considered unsafe.
+ */
+class Libxml2BadOption extends EnumConstant {
+  Libxml2BadOption() { this.getName().matches(["XML_PARSE_NOENT", "XML_PARSE_DTDLOAD"]) }
```
MathiasVP, line 224: `_` is interpreted as "any character" by `matches`. I'm guessing that's not what we want here? As far as I can see, we really just want:
```ql
this.getName() = ["XML_PARSE_NOENT", "XML_PARSE_DTDLOAD"]
```
don't we?

cpp/ql/src/Security/CWE/CWE-611/XXE.ql
```diff
@@ -219,6 +255,23 @@ class XXEConfiguration extends DataFlow::Configuration {
         call.getThisArgument() and
       encodeXercesFlowState(flowstate, 0, 1) // default configuration
     )
+    or
+    // source is an `options` argument on a `libxml2` parse call that specifies
```
MathiasVP, line 259: I feel like every new XML library adds a source/sink in the dataflow configuration. It might be a better design to have an abstract class ("XmlLibrary" or whatever) in the query that can provide:
- The transformer
- The source
- The sink
We don't need to do this in this PR, but it might be a good refactoring to do at some point.

cpp/ql/src/Security/CWE/CWE-611/XXE.ql
geoffw0: Yes, well spotted, and I'm not sure why I wrote `matches` there in the first place!

cpp/ql/src/Security/CWE/CWE-611/XXE.ql
geoffw0: I agree.  I've had similar thoughts, and I also want to move the library-specific code out into one or more `.qll` library files at some point to make the query more readable - a change like this will enable that.

Added to the TODO list on the issue.

cpp/ql/src/Security/CWE/CWE-611/XXE.ql
geoffw0: Updated.

PR #9019, opened 2022-05-03
C++: More XXE Tests

geoffw0:
Add test cases for the Xerces `SAX2XMLReader` interface, fix tests for the `LSParser` interface*, and generally improve the XXE query tests.  Query changes will be done as two follow-up PRs (there's enough to look at in this one as it is).

`*` - it turns out that the `DOMLSParser` returned by `createLSParser` is *not* an `AbstractDOMParser` (as advertised *somewhere*) and works a bit differently to how I had previously understood it.  The existing QL for `createLSParser`, that worked on the old tests, is not correct.


MathiasVP (approved):

geoffw0:
Fixed merge issue.

MathiasVP (approved):
Still LGTM!

PR #9020, opened 2022-05-03
Java: Add CWE-377 tag to java/predictable-seed

joefarebrother:


atorralba (dismissed):

atorralba:
I think we should include a `queryMetadata` change note, since this impacts CWE coverage.

atorralba (approved):

PR #9021, opened 2022-05-03
JS: promote `js/actions/injection` out of experimental

erik-krogh:
I did a LGTM run on ~1000 projects to evaluate `js/actions/injection` (see the internal backref).   
I really liked the 3 results it found, so I think we should promote this query basically as is.   

I've expanded the QHelp a bit, such that it's clear you can use this vulnerability to steal tokens.   
(Those tokens have write access :scream:).  

I've put it as a high precision warning query, because I think this is good enough to run by default.  

Commit-by-commit review is recommended. 

[Evaluation was uneventful](https://github.com/github/codeql-dca-main/tree/data/erik-krogh/actions-nightly/reports).


javascript/ql/lib/semmle/javascript/Actions.qll
```diff
@@ -0,0 +1,322 @@
...
+   * ```
+   * and
+   * ```
+   * on:
+   *   pull_request:
+   * ```
+   *
+   * are equivalent.
+   */
+  class MappingOrSequenceOrScalar extends YAMLNode {
```
esbena, line 44: Should this be in the standard library instead of in Actions.qll.
Actions can't be the only place where this liberal parsing is done.

javascript/ql/lib/semmle/javascript/Actions.qll
```diff
@@ -0,0 +1,322 @@
...
+
+  /**
+   * A `uses` field within an Actions job step, which references an action as a reusable unit of code.
+   * See https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#jobsjob_idstepsuses.
+   *
+   * For example:
+   * ```
+   * uses: actions/checkout@v2
+   * ```
+   * TODO: Does not currently handle local repository references, e.g. `.github/actions/action-name`.
```
esbena, line 225: Lets do something about this TODO, or track it elsewhere.

javascript/ql/lib/semmle/javascript/Actions.qll
```diff
@@ -0,0 +1,322 @@
...
+    /** The version reference used when checking out the Action, e.g. `v2` in `actions/checkout@v2`. */
+    string version;
+
+    Uses() {
+      step.lookup("uses") = this and
+      // Simple regular expression to split up an Action reference `owner/repo@version` into its components.
+      exists(string regexp | regexp = "([^/]+)/([^/@]+)@(.+)" |
+        repositoryOwner = this.getValue().regexpCapture(regexp, 1) and
+        repositoryName = this.getValue().regexpCapture(regexp, 2) and
+        version = this.getValue().regexpCapture(regexp, 3)
```
esbena, line 242: Can we bind these in the getters instead of the charpred such that we still have `Uses` instances for the local repository references that we do not yet parse correctly?

javascript/ql/lib/semmle/javascript/Actions.qll
```diff
@@ -0,0 +1,322 @@
...
+    Run() { step.lookup("run") = this }
+
+    /** Gets the step that executes this `run` command. */
+    Step getStep() { result = step }
+
+    /**
+     * Holds if `${{ e }}` is a GitHub Actions expression evaluated within this `run` command.
+     * See https://docs.github.com/en/free-pro-team@latest/actions/reference/context-and-expression-syntax-for-github-actions.
+     */
+    string getAReferencedExpression() {
```
esbena, line 311: Could we avoid the TODO below if we rename the predicate to be more explicit about the limitations, i.e. `getASimpleReferenceExpression`?

In any case, the limitation should also have a positive and a negative example since the regexps are hard to parse mentally.

javascript/ql/src/Security/CWE-094/ExpressionInjection.qhelp
```diff
@@ -12,16 +12,21 @@
 
 		</p>
 
+        <p>
+            Code injection in GitHub actions may allow an attacker to 
+            exfiltrate the temporary GitHub repository authorization token. 
+            The token has write access to the repository, and thus an attacker
+            can use it to modify the repository.
```
esbena, line 19: The GITHUB_TOKEN has write access by default, but it is configurable today: https://docs.github.com/en/actions/security-guides/automatic-token-authentication#modifying-the-permissions-for-the-github_token.

Examples: https://docs.github.com/en/actions/security-guides/automatic-token-authentication#example-1-passing-the-github_token-as-an-input

javascript/ql/lib/semmle/javascript/Actions.qll
erik-krogh: The class is only used in the experimenal `js/actions/pull-request-target` query, and in that query it's only used to check if there is a `actions/checkout` step.   
I think we can just remove the `TODO:` part (and keep the rest of the comment), since it doesn't affect any current query.

javascript/ql/src/Security/CWE-094/ExpressionInjection.qhelp
erik-krogh: I've tried to update the QHelp such that it should be clear that the GITHUB_TOKEN only sometimes has write access.

esbena (dismissed):

stevecat (approved):
Reviewing on behalf of `@github/docs-content` :wave: This looks great, I just made one small suggestion.

stevecat (approved):

esbena (approved):

PR #9022, opened 2022-05-03
Ruby: Fix `isLocalSourceNode` implementation

RasmusWL:
The old code was equivalent with the code below, which seems wrong

```
not n instanceof ExprNode
or
n instanceof ExprNode and
localFlowStepTypeTracker+(..., n)
```

From running on real DB I found that this meant that the following node
types were also included as local source nodes:

- `TReturningNode`
- `TSynthReturnNode`
- `TSummaryNode`
- `TSsaDefinitionNode`

It turned out that `TSynthReturnNode` was required for sure (but I don't fully undestand why), while the others don't immediately seems to be. 

@hvitved I've requested a review from you, since you seem to have been touching this code last. If you think it looks good, once we have filled out proper documentation for the TODO I left in the code, we can take it out of draft :+1:

### DIFF
--- a/ruby/ql/lib/codeql/ruby/dataflow/internal/DataFlowPrivate.qll
+++ b/ruby/ql/lib/codeql/ruby/dataflow/internal/DataFlowPrivate.qll
@@ -302,6 +302,11 @@ private module Cached {
     or
     n instanceof PostUpdateNodes::ExprPostUpdateNode
     or
+    // TODO: Explain why SynthReturnNode is needed!
+    // if we don't include this, we are not able to find this call:
+    // https://github.com/github/codeql/blob/976daddd36a63bf46836d141d04172e90bb4b33c/ruby/ql/test/library-tests/frameworks/http_clients/NetHttp.rb#L24
+    n instanceof SynthReturnNode
+    or
     // Expressions that can't be reached from another entry definition or expression.
     n instanceof ExprNode and
     not localFlowStepTypeTracker+(any(Node n0 |
--- a/ruby/ql/lib/codeql/ruby/dataflow/internal/DataFlowPrivate.qll
+++ b/ruby/ql/lib/codeql/ruby/dataflow/internal/DataFlowPrivate.qll
@@ -302,13 +302,7 @@ private module Cached {
     or
     n instanceof PostUpdateNodes::ExprPostUpdateNode
     or
-    // TODO: Explain why SynthReturnNode is needed!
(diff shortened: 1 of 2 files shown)


ruby/ql/lib/codeql/ruby/dataflow/internal/DataFlowPrivate.qll
```diff
@@ -302,12 +302,18 @@ private module Cached {
     or
     n instanceof PostUpdateNodes::ExprPostUpdateNode
     or
+    // TODO: Explain why SynthReturnNode is needed!
+    // if we don't include this, we are not able to find this call:
+    // https://github.com/github/codeql/blob/976daddd36a63bf46836d141d04172e90bb4b33c/ruby/ql/test/library-tests/frameworks/http_clients/NetHttp.rb#L24
+    n instanceof SynthReturnNode
+    or
     // Expressions that can't be reached from another entry definition or expression.
+    n instanceof ExprNode and
```
hvitved, line 311: What happens if you instead remove this constraint, and the disjunct `n instanceof SynthReturnNode`?

RasmusWL: I can certainly try :+1:

RasmusWL: That seemed to do the trick :ok_hand:

ruby/ql/test/library-tests/dataflow/type-tracker/TypeTracker.expected
```diff
@@ -43,10 +43,6 @@ track
 | type_tracker.rb:12:1:16:3 | self in m | type tracker with call steps | type_tracker.rb:12:1:16:3 | self (m) |
 | type_tracker.rb:12:1:16:3 | self in m | type tracker without call steps | type_tracker.rb:12:1:16:3 | self in m |
 | type_tracker.rb:13:5:13:7 | var | type tracker without call steps | type_tracker.rb:13:5:13:7 | var |
-| type_tracker.rb:13:5:13:23 | ... = ... | type tracker with call steps | type_tracker.rb:2:5:5:7 | self (field=) |
```
hvitved, line 46: For reference: The reason that `... - ...` is (correctly) no longer a `LocalSourceNode` is because it represents the `SsaDefinitionNode` for the assignment, not the `ExprNode`.

hvitved (approved):
Thanks for fixing

PR #9024, opened 2022-05-03
Data flow: Introduce `ContentDataFlow.qll`

hvitved:
This PR introduces a new (internal) data-flow library that, unlike the normal data-flow library, allows for flow from parts of a source to parts of a sink. Example:
```csharp
class C
{
    C FieldA;
    C FieldB;

    void M1()
    {
        var source = new C();
        Sink(source); // direct flow from source to sink
    }

    void M2()
    {
        var source = new C();
        Sink(source.FieldA); // flow from `FieldA` on source to sink
    }

    void M3()
    {
        var source = new C();
        var c = new C() { FieldB = source }
        Sink(c);  // flow from source to `FieldB` on sink
    }

    void M4()
    {
        var source = new C();
        var c = new C() { FieldB = source.FieldA }
        Sink(c); // flow from `FieldA` on source to `FieldB` on sink
    }
}
```

The normal data-flow library will only identify flow in the `M1` case, where as this new library will be able to identify flow in the cases `M2-M4` as well. This requires a new interface, which is identical to the `Configuration` interface for normal data-flow, but where `hasFlow` is instead

```ql
/**
 * Holds if data stored inside `sourceAp` on `source` flows to `sinkAp` inside `sink`
 * for this configuration. `preservesValue` indicates whether any of the additional
 * flow steps defined by `isAdditionalFlowStep` are needed.
 *
 * For the source access path, `sourceAp`, the top of the stack represents the content
 * that was last read from. That is, if `sourceAp` is `Field1.Field2` (with `Field1`
 * being the top of the stack), then there is flow from `source.Field2.Field1`.
 *
 * For the sink access path, `sinkAp`, the top of the stack represents the content
 * that was last stored into. That is, if `sinkAp` is `Field1.Field2` (with `Field1`
 * being the top of the stack), then there is flow into `sink.Field1.Field2`.
 */
Configuration::hasFlow(
  Node source, AccessPath sourceAp, Node sink, AccessPath sinkAp, boolean preservesValue
)
```

One intended use of this library is for precise flow summary generation, but I thought it might be generally applicable enough to factor out into a separate library.

Note that this library does deliberately not expose the underlying `PathGraph`, and there is no `hasFlowPath` predicate. It would make sense to add this at a later point, but it will require synthesis of a new kind of `PathNode`s, such that
```ql
Configuration::hasFlowPath(
  PathNode source, AccessPath sourceAp, PathNode sink, AccessPath sinkAp, boolean preservesValue
)
```
guarantees that all paths from `source` to `sink` will read the relevant contents from the source (`sourceAp`), and write the relevant contents to the sink. It is therefore not enough to simply expose the underlying path graph.

### DIFF
--- a/csharp/ql/lib/semmle/code/csharp/dataflow/internal/ContentDataFlow.qll
+++ b/csharp/ql/lib/semmle/code/csharp/dataflow/internal/ContentDataFlow.qll
@@ -29,9 +29,12 @@ module ContentDataFlow {
    * ```ql
    * class MyAnalysisConfiguration extends ContentDataFlowConfiguration {
    *   MyAnalysisConfiguration() { this = "MyAnalysisConfiguration" }
-   *   // Override `isSource`, `isSink`, and `accessPathLimit`.
+   *   // Override `isSource` and `isSink`.
    *   // Optionally override `isBarrier`.
    *   // Optionally override `isAdditionalFlowStep`.
+   *   // Optionally override `getAFeature`.
+   *   // Optionally override `accessPathLimit`.
+   *   // Optionally override `isRelevantContent`.
    * }
    * ```
    *
--- a/csharp/ql/lib/semmle/code/csharp/dataflow/internal/ContentDataFlow.qll
+++ b/csharp/ql/lib/semmle/code/csharp/dataflow/internal/ContentDataFlow.qll
@@ -1,0 +1,429 @@
+/**
+ * Provides classes for performing global (inter-procedural)
+ * content-sensitive data flow analyses.
+ */
+
+private import DataFlowImplCommon
+private import DataFlowPrivate as DataFlowPrivate
+
+module ContentDataFlow {
+  private import DataFlowImplSpecific::Private
+  private import DataFlowImplForContentDataFlow as DF
+
+  class Node = DF::Node;
+
+  class FlowFeature = DF::FlowFeature;
+
+  class ContentSet = DF::ContentSet;
+
+  predicate stageStats = DF::stageStats/8;
+
+  /**
+   * A configuration of interprocedural data flow analysis. This defines
+   * sources, sinks, and any other configurable aspect of the analysis. Each
+   * use of the global data flow library must define its own unique extension
+   * of this abstract class. To create a configuration, extend this class with
+   * a subclass whose characteristic predicate is a unique singleton string.
+   * For example, write
+   *
+   * ```ql
+   * class MyAnalysisConfiguration extends ContentDataFlowConfiguration {
+   *   MyAnalysisConfiguration() { this = "MyAnalysisConfiguration" }
+   *   // Override `isSource`, `isSink`, and `accessPathLimit`.
+   *   // Optionally override `isBarrier`.
+   *   // Optionally override `isAdditionalFlowStep`.
+   * }
+   * ```
+   *
+   * Unlike `DataFlow::Configuration` (on which this class is based), we allow
+   * for data to be stored (possibly nested) inside contents of sources and sinks.
+   * We track flow paths of the form
+   *
+   * ```
+   * source --value-->* node
+   *        (--read--> node --value-->* node)*
+   *        --(non-value|value)-->* node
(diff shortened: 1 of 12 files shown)


hvitved:
> OOI: Will this get a help article on when to use "normal" data-flow vs. content data-flow and what the difference between the two is?

Thank you for your interest. This PR is still in draft, and for now it will only be an internal library, so users should now have to worry about it for now.

csharp/ql/lib/semmle/code/csharp/dataflow/internal/ContentDataFlow.qll
```diff
@@ -0,0 +1,429 @@
...
+   * sources, sinks, and any other configurable aspect of the analysis. Each
+   * use of the global data flow library must define its own unique extension
+   * of this abstract class. To create a configuration, extend this class with
+   * a subclass whose characteristic predicate is a unique singleton string.
+   * For example, write
+   *
+   * ```ql
+   * class MyAnalysisConfiguration extends ContentDataFlowConfiguration {
+   *   MyAnalysisConfiguration() { this = "MyAnalysisConfiguration" }
+   *   // Override `isSource`, `isSink`, and `accessPathLimit`.
```
michaelnebel, line 32: accessPathLimit is not abstract.

michaelnebel (commented):
Looks plausible to me, but I would like to hear more about it 😄

csharp/ql/lib/semmle/code/csharp/dataflow/internal/ContentDataFlow.qll
```diff
@@ -0,0 +1,432 @@
+/**
+ * Provides classes for performing global (inter-procedural)
+ * content-sensitive data flow analyses.
+ */
+
+private import DataFlowImplCommon
+private import DataFlowPrivate as DataFlowPrivate
```
aschackmull, line 7: Isn't this included in `private import DataFlowImplSpecific::Private` ?

hvitved: Ah yes, it is wrong to rely on the existence of a `DataFlowPrivate.qll` module.

csharp/ql/lib/semmle/code/csharp/dataflow/internal/ContentDataFlow.qll
```diff
@@ -0,0 +1,432 @@
...
+
+  pragma[nomagic]
+  private predicate pathNodeSuccRec(DF::PathNode pred, DF::PathNode succ) {
+    pathNodeSuccEntry(pred) and
+    succ = pred
+    or
+    exists(DF::PathNode mid |
+      pathNodeSuccRec(pred, mid) and
+      succ = mid.getASuccessor() and
+      not DF::PathGraph::subpaths(mid, succ, _, _) and
```
aschackmull, line 332: Don't you need `not DF::PathGraph::subpaths(mid, _, _, succ) and` as well here?

hvitved: Both are actually redundant, as they are implied by `pathNodeSuccExit(mid)`. But it made me realize that there is one case that doesn't work: Flow into a parameter (and not back out) for which there is also a subpath.

csharp/ql/lib/semmle/code/csharp/dataflow/internal/ContentDataFlow.qll
```diff
@@ -0,0 +1,529 @@
...
+        or
+        DF::PathGraph::subpaths(_, node, _, _)
+      )
+    }
+
+    pragma[nomagic]
+    private predicate bigStepExit(DF::PathNode node) {
+      node.getConfiguration() instanceof Configuration and
+      (
+        bigStepEntry(node)
```
aschackmull, line 364: ?

aschackmull: Gotcha.

PR #9025, opened 2022-05-03
C#: Provenance column in Models as Data CSV format.

michaelnebel:
This serves as a replacement for encoding "generated" information in the "kind" column in the CSV format for models.

More specifically, in this PR we
- Introduce a new column in the (source|sink|summary)model CSV format called provenance for C# and Java.
- Allowed values in the provenance column is "manual" and "generated". The former means hand written model and the latter means a model emitted by the model generator tool.
- All existing handwritten and generated models are updated with provenance information
- The Csv validation now also checks the provenance column.
- The model generator is modified to produce modules in the new syntax.


michaelnebel:
DCA looks good. There doesn't appear to be any failures or regressions that can be linked to the content of this PR.

csharp/ql/lib/semmle/code/csharp/dataflow/ExternalFlow.qll
```diff
@@ -69,6 +69,12 @@
  *    sources "remote" indicates a default remote flow source, and for summaries
  *    "taint" indicates a default additional taint step and "value" indicates a
  *    globally applicable value-preserving step.
+ * 9. The `provenance` column is tag to indicate the origin of the summary.
```
tamasvajk, line 72: Minor typo:
```suggestion
 * 9. The `provenance` column is a tag to indicate the origin of the summary.
```

tamasvajk (commented):
Looks reasonable to me. 

I was not involved in the initial discussions on the `provenance` column, so I'm not sure if this was discussed. The `provenance` column is the last one in the row, it could be made optional. We could interpret its absence as `manual`, and then we wouldn't have to append `;manual` to the end of all manually added models.

asgerf (approved):
LGTM

PR #9030, opened 2022-05-04
Ruby: Model various bits of ActiveSupport

hmac:
This PR models some parts of ActiveSupport that are relevant to us:
- Extensions to `String` and `Enumerable`
- Two logger classes: `ActiveSupport::Logger` and `ActiveSupport::TaggedLogging`

### DIFF
--- a/ruby/ql/lib/codeql/ruby/frameworks/stdlib/Logger.qll
+++ b/ruby/ql/lib/codeql/ruby/frameworks/stdlib/Logger.qll
@@ -35,7 +35,7 @@ module Logger {
 
   /**
    * An instantiation of a logger that responds to the std lib logging methods.
-   * This can be extended to recognise additional instances that conform to the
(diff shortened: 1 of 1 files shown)


ruby/ql/lib/codeql/ruby/frameworks/stdlib/Logger.qll
```diff
@@ -33,11 +33,29 @@ module Logger {
     )
   }
 
+  /**
+   * An instantiation of a logger that responds to the std lib logging methods.
+   * This can be extended to recognise additional instances that conform to the
```
nickrolfe, line 38: We should use US spelling in qldoc comments
```suggestion
   * This can be extended to recognize additional instances that conform to the
```

nickrolfe (commented):
I have one pedantic comment, but otherwise LGTM 👍

nickrolfe (approved):

PR #9034, opened 2022-05-04
Swift: add structured C++ generated classes

redsun82:
This adds `cppgen`, creating structured C++ classes mirroring QL classes
out of `schema.yml`.

An example of generated code at the time of this commit can be found
[in this gist][1].

This also introduces the `?*` (repeated optional) type modifier in the schema
specification. On the dbscheme and QL side, this behaves just like `*`, save
for the fact that `getNumberOfX` is not generated. On the C++ side, this 
translates to a `vector<optional<X>>` type.

[1]: https://gist.github.com/redsun82/57304ddb487a8aa40eaa0caa695048fa

Closes https://github.com/github/codeql-c-team/issues/863


redsun82:
this is on top of https://github.com/github/codeql/pull/9028, so you should ignore the first commit for now (the trapgen unit tests)

AlexDenisov (approved):

AlexDenisov (approved):

PR #9044, opened 2022-05-05
C#: Flow summaries test should print, whether a summary is generated or not.

michaelnebel:


csharp/ql/lib/semmle/code/csharp/dataflow/internal/FlowSummaryImpl.qll
```diff
@@ -219,6 +219,11 @@ module Public {
      */
     pragma[nomagic]
     predicate clearsContent(ParameterPosition pos, ContentSet content) { none() }
+
+    /**
+     * Gets whether the summary is auto generated or not.
+     */
+    boolean isAutoGenerated() { result = false }
```
hvitved, line 226: It is more natural for this to simply be a `predicate isAutoGenerated()`.

hvitved (changes requested):

csharp/ql/lib/semmle/code/csharp/dataflow/internal/FlowSummaryImpl.qll
michaelnebel: Check, will fix it :-)

hvitved (approved):

PR #9046, opened 2022-05-05
MaD: Support contributing to concepts

asgerf:
This PR provides a way to populate concepts using the MaD type definition table.

To recap, it is easy enough to populate a concept using type-definitions, the difficulty was in implementing its members
```codeql
private class ModelDatabaseAccess extends DatabaseAccess {
  ModelDatabaseAccess() {
    this = ModelOutput::getATypeNode("codeql", "DatabaseAccess")
  }

  override DataFlow::Node getAQueryArgument() {
    // No way to implement this
  }

  override DataFlow::Node getAResult() {
    // No way to implement this
  }
}
```
(I'm eliding the conversion between data-flow nodes and API nodes in these examples, see the full version in the diff)

This PR solves the above conundrum via a new overload of `ModelOutput::getATypeNode` which provides access to "base" from which a type-definition row was resolved.

```codeql
API::Node getATypeNode(string package, string type) // old
API::Node getATypeNode(string package, string type, API::Node base) // new
```

When resolving a type definition row
```
package1;type1;package2;type2;path
```
the "base" is whatever API node `package2;type2` was resolved to, from which `path` was evaluated.

We can now implement the full class:
```codeql
private class ModelDatabaseAccess extends DatabaseAccess {
  ModelDatabaseAccess() {
    this = ModelOutput::getATypeNode("codeql", "DatabaseAccess")
  }

  override DataFlow::Node getAQueryArgument() {
    result = ModelOutput::getATypeNode("codeql", "DatabaseAccess.getAQueryArgument", this)
  }

  override DataFlow::Node getAResult() {
    result = ModelOutput::getATypeNode("codeql", "DatabaseAccess.getAResult", this)
  }
}
```

The associated CSV rows for populating this from the `@google-cloud/spanner` model then look like this:
```csv
codeql;DatabaseAccess;@google-cloud/spanner;~SqlExecutorDirect;
codeql;DatabaseAccess.getAQueryArgument;@google-cloud/spanner;~SqlExecutorDirect;Argument[0]
codeql;DatabaseAccess.getAQueryArgument;@google-cloud/spanner;~SqlExecutorDirect;Argument[0].Member[sql]
codeql;DatabaseAccess.getAResult;@google-cloud/spanner;~SqlExecutorDirect;Argument[0..].Parameter[1]
codeql;DatabaseAccess.getAResult;@google-cloud/spanner;~SqlExecutorDirect;ReturnValue.Awaited
```

It's worth mentioning that all this does complicate the semantics of type-definition rows a bit. Previously, it was just a grammar. You could factor out any prefix of an access path into its own type definition, and this was guaranteed to preserve behaviour. Such a refactoring is not sound, however, for type-definition rows that contribute to a "concept member" like `codeql.getAQueryPredicate`, because it would change its associated base node.

Currently in draft as I'm waiting for https://github.com/github/codeql/pull/8883 to merge first.


asgerf:
Aaand I pushed the wrong commit. Sorry to anyone who started looking at it.

PR #9047, opened 2022-05-05
C++: Add support for SAX2XMLReader in the CWE-611 XXE query.

geoffw0:
Support for `SAX2XMLReader` in the CWE-611 XXE query.

Change note covered by the one in recent PR https://github.com/github/codeql/pull/8948.

### DIFF
--- a/cpp/ql/src/Security/CWE/CWE-611/XXE.ql
+++ b/cpp/ql/src/Security/CWE/CWE-611/XXE.ql
@@ -185,8 +185,9 @@ class FeatureDisableDefaultEntityResolution extends Variable {
 }
 
 /**
- * A flow state transformer for a call to `SAX2XMLReader.setFeature(XMLUni::fgXercesDisableDefaultEntityResolution, *)`. Transforms the flow
- * state through the qualifier according to this setting.
+ * A flow state transformer for a call to `SAX2XMLReader.setFeature`
+ * specifying the feature `XMLUni::fgXercesDisableDefaultEntityResolution`.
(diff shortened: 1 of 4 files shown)


cpp/ql/src/Security/CWE/CWE-611/XXE.ql
```diff
@@ -168,12 +175,58 @@ class CreateEntityReferenceNodesTranformer extends XXEFlowStateTranformer {
...
+ * specifying the feature `XMLUni::fgXercesDisableDefaultEntityResolution`.
+ * Transforms the flow state through the qualifier according to this setting.
+ */
+class SetFeatureTranformer extends XXEFlowStateTranformer {
+  Expr newValue;
+
+  SetFeatureTranformer() {
+    exists(Call call, Function f |
+      call.getTarget() = f and
+      f.getDeclaringType() instanceof Sax2XmlReader and
```
MathiasVP, line 198: I think this is another case where we can use https://github.com/github/codeql/blob/main/cpp/ql/lib/semmle/code/cpp/Function.qll#L372

MathiasVP (commented):
LGTM other than a small comment. Since this PR involves joining with global value numbers, should we maybe run DCA on it?

cpp/ql/src/Security/CWE/CWE-611/XXE.ql
geoffw0: Done (I found two instances).

geoffw0:
I've started a DCA run.

MathiasVP (approved):
LGTM!

PR #9053, opened 2022-05-07
C/C++ : memory may not be freed on loop

Yonah125:
This query finds memory that might no be freed in a loop.

For exemple : 

```c
for(int i = 0; i < 10; i++){
     char* notfree = malloc(0x100);
     if(i == 5){
         break;
         }
      free(notfree)
 }
 ```
 
A resultat was [netcdf-c](https://github.com/Unidata/netcdf-c) : https://github.com/Unidata/netcdf-c/pull/2339

### DIFF
--- a/cpp/ql/src/experimental/Critical/MemoryMayNotBeFreedOnLoop.qhelp
+++ b/cpp/ql/src/experimental/Critical/MemoryMayNotBeFreedOnLoop.qhelp
@@ -1,0 +1,22 @@
+<!DOCTYPE qhelp PUBLIC
+  "-//Semmle//qhelp//EN"
+  "qhelp.dtd">
+<qhelp>
+
+
+<overview>
+<p>
+This rule looks for loop that allocate memory, but may break/continue without freeing it. This causes the function to leak memory and may eventually lead to software failure.
+</p> 
+</overview>
+<recommendation>
+<p>Ensure that the function frees all dynamically allocated memory it has acquired in all circumstances, unless that memory is returned to the caller.</p>
+
+</recommendation>
+<example>
+<sample src="MemoryMayNotBeFreedOnLoop.cpp" />
+
+<p>In this example, if an exception occurs the memory allocated into <code>notfree</code> might not be freed if the if condition is true.  To fix this memory leak, a free call should be added for <code>notfree</code> in the if block.</p>
+
+</example>
+</qhelp>
--- a/cpp/ql/src/experimental/Critical/MemoryMayNotBeFreedOnLoop.ql
+++ b/cpp/ql/src/experimental/Critical/MemoryMayNotBeFreedOnLoop.ql
@@ -1,0 +1,143 @@
+/**
+ * @name Memory May Not Be Freed On Loop
+ * @description A loo may break/continue before freeing memory that was allocated in the loop. A free/delete call should be added before the break/continue.
+ * @kind problem
+ * @id cpp/memory-may-not-be-freed-on-loop
+ * @problem.severity critical
+ * @tags efficiency
+ *       security
+ */
+
+import cpp
+import semmle.code.cpp.dataflow.DataFlow
+import MemoryFreed
+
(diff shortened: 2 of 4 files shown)

### CODE REVIEW

**Review by geoffw0 — COMMENTED** [review]
This is a promising idea for a query, thank you for your contribution!

I think we should  clean up the QL a bit - I've made some initial suggestions.  In general adding `\\` comments to blocks of code explaining what they do would help make the QL more readable and maintainable.

**Comment by geoffw0** [comment]
I've run the tests on this PR, and it looks like you need to autoformat the QL file.  Let me know if you don't know how to do this.
```
Error: ql/cpp/ql/src/experimental/Critical/MemoryMayNotBeFreedOnLoop.ql would change by autoformatting.
```

I also started an LGTM run to see what kinds of real world results we get from this query: https://lgtm.com/query/7023328987326235286/ .  It looks like there are some good results, but sometimes a very large number of locations are reported.

File: cpp/ql/src/experimental/Critical/MemoryMayNotBeFreedOnLoop.qhelp [hunk 1]
```diff
@@ -0,0 +1,21 @@
...
+<recommendation>
+<p>Ensure that the function frees all dynamically allocated memory it has acquired in all circumstances, unless that memory is returned to the caller.</p>
+
+</recommendation>
+<example>
+<sample src="MemoryMayNotBeFreedOnLoop.cpp" />
+
+<p>In this example, on the iteration with <code>i == 5</code>, the memory allocated into <code>notfree</code> will not be freed.  To fix this memory leak, a free call should be added for <code>notfree</code> in the if block.</p>
+</example>
+</qhelp>
```
geoffw0, line 21: There should be a references section:
```suggestion
<references>
</references>
</qhelp>
```
Ideally references would be given (with`<li>` `</li>`), but even without, having an empty references section at least allows tooling to add certain references automatically.

geoffw0 (commented):
Hi @Yonah125, please could you let us know if you intend to do any more work on this PR?  I'm OK with merging this into `experimental` as it is - but there are some unaddressed comments, in particular the real world results that reference very large numbers of locations.  I think this query would also benefit greatly from a good test (in `cpp/ql/test/experimental`).

catenacyber:
Hi @geoffw0, I think @Yonah125 is done with this after his latest push. How does this look ?

geoffw0:
> I think @Yonah125 is done with this after his latest push. How does this look ?

Yes I think you're right.  Its looks promising but a little unfinished to me.  But for merging into experimental that's OK, it can potentially be built upon in future.

I've started the checks, will merge if they pass.

cpp/ql/src/experimental/Critical/MemoryMayNotBeFreedOnLoop.ql
```diff
@@ -0,0 +1,52 @@
...
+      bs.getChild(y) instanceof ExprStmt and
+      bs.getChild(y).(ExprStmt).getExpr() instanceof AssignExpr and
+      bs.getChild(y).(ExprStmt).getExpr().(Assignment).getRValue*() = v.getAnAccess()
+    ) and
+    exists(int y | y in [i .. bs.getNumStmt() - 1] |
+      (bs.getChild(y) = is or bs.getChild(y).(IfStmt).getAChild*() = is) and
+      sameLoop(bs, is) and
+      (
+        is.getAChild() = b or
+        is.getThen().getAChild() = b or
```
geoffw0, line 38:
```suggestion
        is.getThen().getAChild() = b
```

geoffw0 (commented):
`AllocAndFree.qll` needs autoformatting.

cpp/ql/src/experimental/Critical/MemoryMayNotBeFreedOnLoop.ql
catenacyber: Friendly ping @Yonah125 could you do these changes ?

PR #9082, opened 2022-05-09
QL: add query warning about `count(...) = 0`. 

erik-krogh:
You should use `not exists(..)` instead of `count(...) = 0`. 

With `count(..) = 0` the evaluator will first check how many elements exists in a set, and afterwards check if that count equals 0.   
Using `not exists(..)` instead allows the optimizer to skip the counting.   
(I checked the RA, the optimizer does not optimize `count(..)  = 0` to `not exists(..)`). 

I originally made because I had the idea to replace `count(x)= 1` with `exists(unique(| | x))`, but I didn't like the results I got from that, so I've disabled that part of the query for now.  
(The `count(x) = 1` query got 200+ results). 

---- 

[Evaluation was ueventful](https://github.com/github/codeql-dca-main/tree/data/PR-9082-2-python__1/reports).

### DIFF
--- a/ql/ql/src/queries/style/CountingToZero.ql
+++ b/ql/ql/src/queries/style/CountingToZero.ql
@@ -1,0 +1,21 @@
+/**
+ * @name Counting zero elements
+ * @description Use not exists instead of checking that there is zero elements in a set.
+ * @kind problem
+ * @problem.severity warning
+ * @id ql/counting-to-zero
+ * @precision high
+ */
+
+import ql
+
+from ComparisonFormula comp, int c, string msg
+where
+  c = 0 and // [0, 1] and // skipping the 1 case for now, as it seems too noisy.
+  comp.getOperator() = ["=", "!="] and
+  comp.getAnOperand().(Integer).getValue() = c and
+  comp.getAnOperand().(Aggregate).getKind() = ["count", "strictcount"] and
+  if c = 0
+  then msg = "Use not exists(..) instead of checking that there is zero elements in a set."
+  else msg = "Use unique(.. | |) instead of checking that there is one element in a set."
+select comp, msg
--- a/python/ql/src/Security/CWE-022/TarSlip.ql
+++ b/python/ql/src/Security/CWE-022/TarSlip.ql
@@ -81,11 +81,11 @@ class ExcludeTarFilePy extends Sanitizer {
 
 /* Any call to an extractall method */
 class ExtractAllSink extends TaintSink {
-  CallNode call;
-
   ExtractAllSink() {
-    this = call.getFunction().(AttrNode).getObject("extractall") and
-    not exists(call.getAnArg())
+    exists(CallNode call |
+      this = call.getFunction().(AttrNode).getObject("extractall") and
+      not exists(call.getAnArg())
+    )
   }
 
   override predicate sinks(TaintKind kind) { kind instanceof OpenTarFile }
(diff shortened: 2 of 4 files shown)


ql/ql/src/queries/style/CountingToZero.ql
```diff
@@ -0,0 +1,21 @@
...
+ */
+
+import ql
+
+from ComparisonFormula comp, int c, string msg
+where
+  c = 0 and // [0, 1] and // skipping the 1 case for now, as it seems too noisy.
+  comp.getOperator() = ["=", "!="] and
+  comp.getAnOperand().(Integer).getValue() = c and
+  comp.getAnOperand().(Aggregate).getKind() = ["count", "strictcount"] and
```
MathiasVP, line 17: This is probably not really an issue, but I'm gonna write it up anyway: If a predicate is tagged with `language[monotonicAggregates]` I guess `count/strictcount` and `not exists` have different monotonicity properties. But I guess there's no harm in surfacing the alert so that the author can think about whether it's possible to replace the aggregate with a `not exist`.

java/ql/lib/semmle/code/java/frameworks/javaee/ejb/EJB.qll
```diff
@@ -33,14 +33,14 @@ class SessionEJB extends EJB {
     // Either the EJB does not declare any business interfaces explicitly
     // and implements a single interface candidate,
     // which is then considered to be the business interface...
-    count(this.getAnExplicitBusinessInterface()) = 0 and
-    count(this.getAnImplementedBusinessInterfaceCandidate()) = 1 and
+    not exists(this.getAnExplicitBusinessInterface()) and
+    exists(unique( | | this.getAnImplementedBusinessInterfaceCandidate())) and
```
MathiasVP, line 37: I remember @jbj tried to replace the C++ occurrences of `count(...) = 1` with `exists(unique(...))` and concluded that the generated RA was worse. The conversation about this has disappeared from Slack now, unfortunately. This was also _just_ after the `unique` aggregate was introduced, so the situation might be better now. It's worth checking up on, though!

MathiasVP (dismissed):
The QL and C/C++ changes LGTM 👍 I have a few comments, though!

java/ql/lib/semmle/code/java/frameworks/javaee/ejb/EJB.qll
erik-krogh: The Java evaluation is also not looking all that well.  
I'm waiting for some retries before I conclude anything, but I might end up reverting the Java changes.

erik-krogh: The C++ evaluation also looks weird, so I think I'll just do Python in this PR.

ql/ql/src/queries/style/CountingToZero.ql
erik-krogh: > But I guess there's no harm in surfacing the alert so that the author can think about whether it's possible to replace the aggregate with a not exist. 

I think so too.  
I'm not currently making an automated patch out of this yet. If I did then I would have to consider such cases.

yoff (commented):
This looks good from the Python side. I would be happy to merge these changes in regardless of the query. (And thanks for the extra cleanup also 👍) Is the query noisy for other languages, though? It seems the situation is not as clear cut as the description indicates.

erik-krogh:
> Is the query noisy for other languages, though? It seems the situation is not as clear cut as the description indicates. 

I don't think it's noisy, I would want to fix all the alerts.  

For Java it might have been the `exists(unique( X | | ))` change I included that caused bad performance.   
And the standard deviations in the C++ evaluation are massive, so the regression might be noise.

tausbn (approved):
QL code looks good to me. :+1:

PR #9083, opened 2022-05-09
Actions: Fetch CodeQL CLI using `gh` rather than third-party Action

henrymercer:
We recently got a [Dependabot update](https://github.com/github/codeql/pull/9077) for this Action, but it probably makes more sense to remove it and standardise on `.github/actions/fetch-codeql`.

### DIFF
--- a/.github/workflows/query-list.yml
+++ b/.github/workflows/query-list.yml
@@ -30,20 +30,14 @@ jobs:
       with:
         python-version: 3.8
     - name: Download CodeQL CLI
-      uses: dsaltares/fetch-gh-release-asset@aa37ae5c44d3c9820bc12fe675e8670ecd93bd1c
-      with:
-        repo: "github/codeql-cli-binaries"
-        version: "latest"
-        file: "codeql-linux64.zip"
-        token: ${{ secrets.GITHUB_TOKEN }}
+      uses: ./codeql/.github/actions/fetch-codeql
     - name: Unzip CodeQL CLI
       run: unzip -d codeql-cli codeql-linux64.zip
     - name: Build code scanning query list
       run: |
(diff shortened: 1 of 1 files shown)


.github/workflows/query-list.yml
```diff
@@ -29,21 +29,15 @@
...
       with:
         python-version: 3.8
     - name: Download CodeQL CLI
-      uses: dsaltares/fetch-gh-release-asset@aa37ae5c44d3c9820bc12fe675e8670ecd93bd1c
-      with:
-        repo: "github/codeql-cli-binaries"
-        version: "latest"
-        file: "codeql-linux64.zip"
-        token: ${{ secrets.GITHUB_TOKEN }}
+      uses: ./codeql/.github/actions/fetch-codeql
```
nickrolfe, line 33: If this is a path within the submodule, shouldn't it be...
```suggestion
      uses: ./ql/.github/actions/fetch-codeql
```

henrymercer: We're grabbing it from https://github.com/github/codeql/blob/main/.github/actions/fetch-codeql/action.yml — unusually the `codeql` repo is checked out at `codeql`, since we also want to check out `codeql-go`.

I'll add a comment to explain so others hopefully won't get confused in the future, as this is a little surprising.

nickrolfe: Ah, thanks.

nickrolfe (approved):
In that case, LGTM, assuming you've seen it working.

henrymercer:
Here's an example run from the PR checks: https://github.com/github/codeql/suites/6430300162/artifacts/235114581

I've download the artifact and it makes sense.

PR #9086, opened 2022-05-09
CPP: Add query for CWE-297: Improper Validation of Certificate with Host Mismatch

ihsinme:
this query finds certificate situations without name validation.
I tried to take into account all cases of name processing, so there are old methods.

CVE-2010-1155
CVE-2013-7449
CVE-2016-10937

I had some confusion with the test file, the current one contains only good situations. I don't know how I can proceed in this case, I do a separate simple test without validation, but I'm not sure if it's common practice, two file tests in different folders. tell me how to be.


MathiasVP:
Hi @ihsinme,

Thanks for another contribution. We will review get to review it soon!

MathiasVP:
Hi @ihsinme,

Can you add a comment to this query explaining how it's different from https://github.com/github/codeql/blob/main/cpp/ql/src/Security/CWE/CWE-295/SSLResultNotChecked.ql?

MathiasVP:
> I had some confusion with the test file, the current one contains only good situations. I don't know how I can proceed in this case, I do a separate simple test without validation, but I'm not sure if it's common practice, two file tests in different folders. tell me how to be.

I'm not sure what you mean. Do you want to add the same file to several folders?

ihsinme:
> > I had some confusion with the test file, the current one contains only good situations. I don't know how I can proceed in this case, I do a separate simple test without validation, but I'm not sure if it's common practice, two file tests in different folders. tell me how to be.
> 
> I'm not sure what you mean. Do you want to add the same file to several folders?

the problem is that i have to store the good and bad example in two different files.
Perhaps you have rules for this case.

MathiasVP:
> the problem is that i have to store the good and bad example in two different files.
> Perhaps you have rules for this case.

You can have as many `.c` and `.cpp` test files as you want in the test directory as you want. `codeql test run` should pick them all up.

Does that solve your issue?

cpp/ql/src/experimental/Security/CWE/CWE-297/ValidationCertificateHostMismatch.ql
```diff
@@ -0,0 +1,73 @@
...
+ * @precision medium
+ * @tags correctness
+ *       security
+ *       external/cwe/cwe-297
+ */
+
+import cpp
+import semmle.code.cpp.valuenumbering.GlobalValueNumbering
+
+predicate checkHostnameOlder1() {
```
MathiasVP, line 16: Reading this predicate out loud, this is what I'm reading:

This predicate holds if there exists a function call _anywhere_ in the program that calls `X509_get_ext_d2i`, and either:
- There exists a use of the macro `sk_GENERAL_NAME_num` or a use of the macro `sk_GENERAL_NAME_value`, or
- There exists a call to the function `sk_GENERAL_NAME_value` or a call to the function `sk_GENERAL_NAME_num`.

That doesn't make a lot of sense to me. Can `sk_GENERAL_NAME_num` and `sk_GENERAL_NAME_value` really both be macros and function calls (maybe depending on versions?). Also: Shouldn't this predicate have any parameters? Right now, this is just a property of the _entire_ codebase that either holds of doesn't hold (and I can make it hold by adding a random call to `X509_get_ext_d2i`, `sk_GENERAL_NAME_num` and `sk_GENERAL_NAME_value` anywhere in the program.

What are you really trying to model with this predicate?

cpp/ql/src/experimental/Security/CWE/CWE-297/ValidationCertificateHostMismatch.ql
```diff
@@ -0,0 +1,73 @@
...
+      or
+      exists(FunctionCall fctmp2, FunctionCall fctmp3 |
+        fctmp2.getTarget().hasName("sk_GENERAL_NAME_num") and
+        fctmp3.getTarget().hasName("sk_GENERAL_NAME_value")
+      )
+    )
+  )
+}
+
+predicate checkHostnameOlder2() {
```
MathiasVP, line 33: Same question here: This predicate is a property of the entire codebase that either holds or doesn't hold. And I can make it hold by adding a couple of calls to some functions anywhere in the codebase. Please describe what are you trying to do here, and I'm sure we can fix it up.

cpp/ql/src/experimental/Security/CWE/CWE-297/ValidationCertificateHostMismatch.ql
```diff
@@ -0,0 +1,73 @@
...
+  )
+}
+
+from FunctionCall fc
+where
+  not checkHostnameOlder1() and
+  not checkHostnameOlder2() and
+  not checkHostnameUp110() and
+  not checkHostnameUp102() and
+  fc.getTarget().hasName(["SSL_set_verify", "SSL_get_peer_certificate", "SSL_get_verify_result"])
```
MathiasVP, line 72: Right now you're raising an alert on all uses of the functions whose name is `SSL_set_verify`, `SSL_get_peer_certificate` or `SSL_get_verify_result`. Is that really what you intend?

MathiasVP (commented):
A couple of initial comments on the code.

cpp/ql/src/experimental/Security/CWE/CWE-297/ValidationCertificateHostMismatch.ql
ihsinme: sorry for the delay in replies.

in this predicate I have omitted call connectivity control due to the presence of macros, in the rest of the predicates I use `GVN` for call connectivity control.

cpp/ql/src/experimental/Security/CWE/CWE-297/ValidationCertificateHostMismatch.ql
ihsinme: I have implemented a search for all possible ways to validate the name of the certificate specified in https://wiki.openssl.org/index.php/Hostname_validation.

I'm looking for code where the developer doesn't use any of these name validation methods at all, but uses certificate validation. In this case, it's definitely a threat.
I skip when there is an error in the implementation of one of the methods, or the method is not implemented for all connections.

cpp/ql/src/experimental/Security/CWE/CWE-297/ValidationCertificateHostMismatch.ql
ihsinme: at the moment I want to find a place to work with connection verification, when in the code I did not find a way to check the name of the certificate. every place is a mistake.

ihsinme:
I have added another test file, could you please run the checks

ihsinme:
> To be clear: You don't _have_ to do this, but adding the above extensions will likely improve the final score on your bug bounty submission.

Hi @MathiasVP.

of course I will try to improve this query.
I need some time for this.

Thank you.

fixing the work with the test file, I'll deal with it after the request has been improved.

MathiasVP:
> @MathiasVP, i made additions.

Fantastic!

> but my local environment is not able to work with multiple test files, can you show the results of working with test files?
> 
> Pay attention to the test files, I talked about this at the beginning of the request. if the query logic is looking for something that is missing somewhere in the code, then building the test files is very difficult. the presence of one correct example excludes detection ( Is there any trick for this?

There really isn't any tricks to this. The tests highlights the behavior of the query as you wrote it. I described the problem [here](https://github.com/github/codeql/pull/9086/files#r884989724):
> This predicate is a property of the entire codebase that either holds or doesn't hold. And I can make it hold by adding a couple of calls to some functions anywhere in the codebase. Please describe what are you trying to do here, and I'm sure we can fix it up.

I think the only way to fix the tests is to change the query logic so that the above isn't the case anymore (i.e., so that a single call in _some_ function anywhere in the code doesn't fix all the alerts in the codebase).

ihsinme:
I have established a call chain connection, please have a look.

I would like to make a few points:
1. in the boost, I still couldn't make the correct calls, because I don't know how to define a callback call in the test file. (if you have any thoughts please push me)
2. in file `test.cpp` in function `goodTest3` on line `113` if you change the call to `X509_NAME_get_index_by_NID (X509_get_subject_name (cert), nid, -1);`. then there will be no error, but this call will not be considered `X509_NAME_get_index_by_NID` either. if you could tell me why this is happening, that would be great.
3. when I wrote about the complexity of working with several test files, I thought that there is a mechanism that allows each test file to be considered as a separate project.

MathiasVP:
> in the boost, I still couldn't make the correct calls, because I don't know how to define a callback call in the test file. (if you have any thoughts please push me)

You should be able to do something similar to what they do in [this](https://cpp.hotexamples.com/examples/-/-/set_verify_callback/cpp-set_verify_callback-function-examples.html#0x15e1e63a3965f213c5fbdc5885ca60e7c965298f52512705e57567fe2e4f8b8b-187,,235,) example:

```cpp
class verify_context { /* ... */ };

bool my_verify_callback(
  bool preverified,
  verify_context& ctx
) {
  /* ... */
}

void test(boost::asio::ssl::stream sock){
  // ...
  sock.set_verify_callback(my_verify_callback);
  // ...
}
```
or using a lambda:
```cpp
class verify_context { /* ... */ };

void test(boost::asio::ssl::stream sock){
  // ...
  sock.set_verify_callback([&](bool preverified, verify_context& ctx) {
    // ...
  });
  // ...
}
```
Does that make sense?

ihsinme:
> You should be able to do something similar to what they do in [this](https://cpp.hotexamples.com/examples/-/-/set_verify_callback/cpp-set_verify_callback-function-examples.html#0x15e1e63a3965f213c5fbdc5885ca60e7c965298f52512705e57567fe2e4f8b8b-187,,235,) example:

thanks, added two goodTest functions.

> then there will not be an alert? There shouldn't be an alert in this function anyway, right? (since it's a good test).

I was surprised that I didn't declare a function with three parameters in this test. and expected an error with such a replacement.

> Different .cpp files for a qltest are compiled as different compilation units (like you'd normally do for a C/C++ project), but they're all linked together before running the test.

thanks.

P.S. could you show me the result of working on three file tests?

ihsinme:
good afternoon @MathiasVP.
any news on this PR?

MathiasVP:
> good afternoon @MathiasVP.
> any news on this PR?

Hi! I've just come back from vacation now and I'm catching up with all of the open PRs of yours that I'm assigned to. I promise to have an update on all of those today.

MathiasVP:
CI failure:
```
--- expected
2022-08-03T09:16:17.3406473Z +++ actual
2022-08-03T09:16:17.3407144Z @@ -1,1 +1,5 @@
2022-08-03T09:16:17.3408765Z -| test1.cpp:11:6:11:13 | badTest1 | You may have missed checking the name of the certificate. |
2022-08-03T09:16:17.3410313Z +| test1.cpp:47:5:47:12 | badTest1 | You may have missed checking the name of the certificate. |
2022-08-03T09:16:17.3411826Z +| test1.cpp:53:5:53:12 | badTest2 | You may have missed checking the name of the certificate. |
2022-08-03T09:16:17.3413306Z +| test1.cpp:59:5:59:12 | badTest3 | You may have missed checking the name of the certificate. |
2022-08-03T09:16:17.3414764Z +| test2.cpp:41:6:41:13 | badTest1 | You may have missed checking the name of the certificate. |
2022-08-03T09:16:17.3416365Z +| test.cpp:138:6:138:13 | badTest1 | You may have missed checking the name of the certificate. |
2022-08-03T09:16:17.3425768Z ##[error][6893/7565 comp 7.5s eval 925ms] FAILED(RESULT) ql/cpp/ql/test/experimental/query-tests/Security/CWE/CWE-297/semmle/tests/ValidationCertificateHostMismatch.qlref
```
Looks like the .expected files need to be updated.

ihsinme:
@MathiasVP  please show me what is the error.

ihsinme:
good afternoon @MathiasVP.
maybe I didn’t do something for this PR?

MathiasVP:
Hi @ihsinme,

Apologies for not getting back to this one. There was a technical issue that caused CI to complain. I hope we've fixed it by merging in the latest `main`.

cpp/ql/src/experimental/Security/CWE/CWE-297/ValidationCertificateHostMismatch.ql
```diff
@@ -0,0 +1,141 @@
...
+    globalValueNumber(fctmp1) = globalValueNumber(fctmp2.getArgument(0)) and
+    globalValueNumber(fctmp1) = globalValueNumber(fctmp3.getArgument(0)) and
+    fctmp1.getTarget().hasName("SSL_get0_param") and
+    fctmp2.getTarget().hasName("X509_VERIFY_PARAM_set_hostflags") and
+    fctmp3.getTarget().hasName("X509_VERIFY_PARAM_set1_host") and
+    fctmp4.getTarget().hasName("SSL_set_verify")
+  )
+}
+
+predicate checkHostnameGnuTLS(FunctionCall fc) {
```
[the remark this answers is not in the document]
MathiasVP: No. The Code Scanning alert is saying that QL names must preferably be in PascalCase (for classes and modules) or camelCase (for predicates and variables). So `checkHostnameGnuTLS` should ideally be `checkHostnameGnuTls` to match our [style guide](https://github.com/github/codeql/blob/main/docs/ql-style-guide.md). However, one could argue that, since `TLS` is a common way to write Transport Layer Security, this name is actually fine.

So I don't think you need to change your name to make the Code Scanning alert disappear.

MathiasVP:
Hi @ihsinme,

I'm trying to resolve the CI failure. Not sure why you keep being a victim of this particular issue. Hopefully merging in `main` will resolve it 🤞.

PR #9087, opened 2022-05-09
CPP: Add query for CWE-670: Always-Incorrect Control Flow Implementation when use SSL_shutdown 

ihsinme:
this query looks for improperly closed connections.

CVE-2008-1531

when the situations of calling the second function to close are not processed.

### DIFF
--- a/cpp/ql/src/experimental/Security/CWE/CWE-670/DangerousUseSSL_shutdown.cpp
+++ b/cpp/ql/src/experimental/Security/CWE/CWE-670/DangerousUseSSL_shutdown.cpp
@@ -1,0 +1,11 @@
+...
+SSL_shutdown(ssl); 
+SSL_shutdown(ssl); // BAD
+...
+    switch ((ret = SSL_shutdown(ssl))) {
+    case 1:
+      break;
+    case 0:
+      ERR_clear_error();
+      if (-1 != (ret = SSL_shutdown(ssl))) break; // GOOD
+...
--- a/cpp/ql/src/experimental/Security/CWE/CWE-670/DangerousUseSSL_shutdown.qhelp
+++ b/cpp/ql/src/experimental/Security/CWE/CWE-670/DangerousUseSSL_shutdown.qhelp
@@ -18,6 +18,10 @@
   CERT Coding Standard:
   <a href="https://wiki.sei.cmu.edu/confluence/display/c/EXP12-C.+Do+not+ignore+values+returned+by+functions">EXP12-C. Do not ignore values returned by functions - SEI CERT C Coding Standard - Confluence</a>.
 </li>
+<li>
+  Openssl.org:
+  <a href="https://www.openssl.org/docs/man3.0/man3/SSL_shutdown.html">SSL_shutdown - shut down a TLS/SSL connection</a>.
+</li>
 
 </references>
 </qhelp>
--- a/cpp/ql/src/experimental/Security/CWE/CWE-670/DangerousUseSSL_shutdown.qhelp
+++ b/cpp/ql/src/experimental/Security/CWE/CWE-670/DangerousUseSSL_shutdown.qhelp
@@ -1,0 +1,23 @@
+<!DOCTYPE qhelp PUBLIC
+  "-//Semmle//qhelp//EN"
+  "qhelp.dtd">
+<qhelp>
+<overview>
+<p>Incorrect closing of the connection leads to the creation of different states for the server and client, which can be exploited by an attacker.</p>
+
+</overview>
+
+<example>
+<p>The following example shows the incorrect and correct usage of function SSL_shutdown.</p>
+<sample src="DangerousUseSSL_shutdown.cpp" />
+
+</example>
+<references>
+
+<li>
+  CERT Coding Standard:
+  <a href="https://wiki.sei.cmu.edu/confluence/display/c/EXP12-C.+Do+not+ignore+values+returned+by+functions">EXP12-C. Do not ignore values returned by functions - SEI CERT C Coding Standard - Confluence</a>.
+</li>
+
+</references>
+</qhelp>
--- a/cpp/ql/src/experimental/Security/CWE/CWE-670/DangerousUseSSL_shutdown.ql
+++ b/cpp/ql/src/experimental/Security/CWE/CWE-670/DangerousUseSSL_shutdown.ql
@@ -1,0 +1,33 @@
+/**
+ * @name Dangerous use SSL_shutdown.
+ * @description Incorrect closing of the connection leads to the creation of different states for the server and client, which can be exploited by an attacker.
+ * @kind problem
+ * @id cpp/dangerous-use-of-ssl_shutdown
+ * @problem.severity warning
+ * @precision medium
+ * @tags correctness
+ *       security
+ *       external/cwe/cwe-670
+ */
+
+import cpp
+import semmle.code.cpp.commons.Exclusions
+import semmle.code.cpp.valuenumbering.GlobalValueNumbering
+
+from FunctionCall fc, FunctionCall fc1
+where
+  fc != fc1 and
+  fc.getASuccessor+() = fc1 and
+  fc.getTarget().hasName("SSL_shutdown") and
+  fc1.getTarget().hasName("SSL_shutdown") and
+  fc1 instanceof ExprInVoidContext and
+  (
+    globalValueNumber(fc.getArgument(0)) = globalValueNumber(fc1.getArgument(0)) or
+    fc.getArgument(0).(VariableAccess).getTarget() = fc1.getArgument(0).(VariableAccess).getTarget()
+  ) and
+  not exists(FunctionCall fctmp |
+    fctmp.getTarget().hasName("SSL_free") and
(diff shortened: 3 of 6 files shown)


cpp/ql/src/experimental/Security/CWE/CWE-670/DangerousUseSSL_shutdown.qhelp
```diff
@@ -0,0 +1,23 @@
...
+<example>
+<p>The following example shows the incorrect and correct usage of function SSL_shutdown.</p>
+<sample src="DangerousUseSSL_shutdown.cpp" />
+
+</example>
+<references>
+
+<li>
+  CERT Coding Standard:
+  <a href="https://wiki.sei.cmu.edu/confluence/display/c/EXP12-C.+Do+not+ignore+values+returned+by+functions">EXP12-C. Do not ignore values returned by functions - SEI CERT C Coding Standard - Confluence</a>.
```
geoffw0, line 19: Another potential reference could be the documentation for `SSL_Shutdown`, which explains the meanings of different return values: https://www.openssl.org/docs/man3.0/man3/SSL_shutdown.html

cpp/ql/src/experimental/Security/CWE/CWE-670/DangerousUseSSL_shutdown.cpp
```diff
@@ -0,0 +1,11 @@
+...
+SSL_shutdown(ssl); 
+SSL_shutdown(ssl); // BAD
+...
+    switch ((ret = SSL_shutdown(ssl))) {
```
geoffw0, line 5:
```suggestion
    switch (ret = SSL_shutdown(ssl)) {
```

cpp/ql/src/experimental/Security/CWE/CWE-670/DangerousUseSSL_shutdown.ql
```diff
@@ -0,0 +1,33 @@
...
+import semmle.code.cpp.commons.Exclusions
+import semmle.code.cpp.valuenumbering.GlobalValueNumbering
+
+from FunctionCall fc, FunctionCall fc1
+where
+  fc != fc1 and
+  fc.getASuccessor+() = fc1 and
+  fc.getTarget().hasName("SSL_shutdown") and
+  fc1.getTarget().hasName("SSL_shutdown") and
+  fc1 instanceof ExprInVoidContext and
```
geoffw0, line 23: I understand that we're looking for two calls to `SSL_Shutdown` on the same `SSL` object (without a call to `SSL_Free` in between), where the return value of the second call is not checked.  Why is it OK for the return value of the first call to be unchecked but not the second?

geoffw0 (commented):
Thank you for this contribution.

The code and tests LGTM, though comments in the code would have made it easier to read and understand.

geoffw0:
A run of this query on LGTM: https://lgtm.com/query/1680253232874044114/

cpp/ql/src/experimental/Security/CWE/CWE-670/DangerousUseSSL_shutdown.cpp
MathiasVP: That extra set of parentheses might actually be intentional. It's the standard way of telling people that "yes, I know that this is an assignment in a conditional". We even have a [query](https://github.com/github/codeql/blob/main/cpp/ql/src/Likely%20Bugs/Likely%20Typos/AssignWhereCompareMeant.ql#L55) that uses to weed out false positives 😄.

geoffw0: Ah, I had forgotten about that pattern.  Please disregard my suggestion.

cpp/ql/src/experimental/Security/CWE/CWE-670/DangerousUseSSL_shutdown.ql
ihsinme: You are absolutely right, the first check also needs to be checked.

1. I want to avoid situations with only one call.
2. It seems to me that when the first call check is skipped, and the second one is checked, this is a very rare case.

but if you find grain in it, I will add this detector.

cpp/ql/src/experimental/Security/CWE/CWE-670/DangerousUseSSL_shutdown.qhelp
ihsinme: is such a change possible?

ihsinme:
> A run of this query on LGTM: https://lgtm.com/query/1680253232874044114/

https://lgtm.com/query/4523088424855036098/

ihsinme:
could you show me what kind of errors are generated during checks.

cpp/ql/src/experimental/Security/CWE/CWE-670/DangerousUseSSL_shutdown.qhelp
geoffw0: What you have done looks correct to me.

cpp/ql/src/experimental/Security/CWE/CWE-670/DangerousUseSSL_shutdown.ql
geoffw0: > I want to avoid situations with only one call.

Are they problematic?  Is it sometimes OK for the developer to ignore the return value of `SSL_shutdown` in these simpler cases?

geoffw0:
Great, that's three real world results between our two LGTM runs!

Running the checks now...

cpp/ql/src/experimental/Security/CWE/CWE-670/DangerousUseSSL_shutdown.ql
ihsinme: I am sure that you should always check the return value of this function. however, we are talking about the problem of ambiguous understanding of states on the part of the server and the client, in this case, the presence of one call indicates that the developer does not consider this a problem.

the addition of this detection, in my opinion, will lead to an increase in hits and will correlate with the situation in these PRs 6950 6947.

cpp/ql/src/experimental/Security/CWE/CWE-670/DangerousUseSSL_shutdown.ql
```diff
@@ -0,0 +1,33 @@
+/**
+ * @name Dangerous use SSL_shutdown.
+ * @description Incorrect closing of the connection leads to the creation of different states for the server and client, which can be exploited by an attacker.
+ * @kind problem
+ * @id cpp/dangerous-use-of-ssl_shutdown
```
geoffw0, line 5: It looks like the checks aren't happy with this `@id`.  You could use:
```suggestion
 * @id cpp/dangerous-use-of-ssl-shutdown
```

cpp/ql/src/experimental/Security/CWE/CWE-670/DangerousUseSSL_shutdown.ql
geoffw0: I see, checking for an error on the first call to `SSL_shutdown` suggests that the program needs to handle the error to behave correctly, so an error on the second `SSL_shutdown` should be handled as well.  But its also possible the program doesn't need to handle any errors (perhaps its about to exit anyway), so you're reluctant to catch the case of a single call to `SSL_shutdown` with no error handling.

geoffw0:
I'm happy with this, when the checks pass.

ihsinme:
> I'm happy with this, when the checks pass. And I believe it would have caught the issue described in https://redmine.lighttpd.net/issues/285#comment:18 .

this is a very interesting point.
but it is difficult to determine correctly.
there's a non-null error here, and it's used elsewhere.
to work with this goal, we will need to determine that the client will continue after it is possible to close the connection, and ensure that it uses the shared error store.

I really like this item in the description, but I'm afraid that it will complicate the query very much, and narrow the detection very much.

ihsinme:
if you still think it's reasonable, I am ready to agree with you. let's just talk about what you mean
1. detection of situations when the first or only call is made without checking.
2. detection of a situation when the accumulated error is not reset and can be used in another place, which can lead to a connection break.

geoffw0:
I'm not advising any changes at this time, I'm just trying to better understand why the query works the way it does.  I think I'm happy with the QL and the real world (lgtm.com) results we have seen so far.

geoffw0 (approved):
I think its about time I approve this.  It does what its supposed to do and finds a few good results. :+1:

ihsinme:
> I think its about time I approve this. It does what its supposed to do and finds a few good results. 👍

thanks

ihsinme:
What do I need to do for this PR?

geoffw0:
I believe the bug bounty process concluded that this PR is not eligible for a bounty award.  I'm happy to accept the code into experimental, with that understanding.

ihsinme:
despite the bounty decision, I still find the request useful.
so I will be glad if you accept it and it will be used)

geoffw0:
Great, thanks for your contribution!

PR #9090, opened 2022-05-09
CPP: Add query for CWE-552 Files Accessible to External Parties when using rename

ihsinme:
this query is looking for a mishandling situation rename. when a file is torn off from the program name and overwritten, this situation can lead to various actions, such as using links.

CVE-2012-0787

### DIFF
--- a/cpp/ql/src/experimental/Security/CWE/CWE-552/DangerousUseRename.cpp
+++ b/cpp/ql/src/experimental/Security/CWE/CWE-552/DangerousUseRename.cpp
@@ -2,6 +2,8 @@
 ...
   if (rename(from,to)==0) // BAD
     return;
+  f1 = fopen(from, "r");
+  f2 = fopen(to, "w");
 ...
   if (rename(from,to)==0) // GOOD
     return;
--- a/cpp/ql/src/experimental/Security/CWE/CWE-552/DangerousUseRename.cpp
+++ b/cpp/ql/src/experimental/Security/CWE/CWE-552/DangerousUseRename.cpp
@@ -1,0 +1,13 @@
+
+...
+  if (rename(from,to)==0) // BAD
+    return;
+  f1 = fopen(from, "r");
+  f2 = fopen(to, "w");
+...
+  if (rename(from,to)==0) // GOOD
+    return;
+  f1 = fopen(from, "r");
+  fd = open(to, O_WRONLY|O_CREAT|to_oflags, S_IRUSR|S_IWUSR);
+  f2 = fdopen(fd, "w");
+...
--- a/cpp/ql/src/experimental/Security/CWE/CWE-552/DangerousUseRename.cpp
+++ b/cpp/ql/src/experimental/Security/CWE/CWE-552/DangerousUseRename.cpp
@@ -2,8 +2,6 @@
 ...
   if (rename(from,to)==0) // BAD
     return;
-  f1 = fopen(from, "r");
-  f2 = fopen(to, "w");
 ...
   if (rename(from,to)==0) // GOOD
     return;
--- a/cpp/ql/src/experimental/Security/CWE/CWE-552/DangerousUseRename.qhelp
+++ b/cpp/ql/src/experimental/Security/CWE/CWE-552/DangerousUseRename.qhelp
@@ -3,7 +3,7 @@
   "qhelp.dtd">
 <qhelp>
 <overview>
-<p>If the rename function did not work correctly, simply overwriting the destination file can lead to errors. Use access to the destination file with permission setting.</p>
+<p>If the rename function did not work correctly, attempting to copy the contents of a source file by opening the target file without assessing permissions creates a mechanism for deleting an arbitrary system file.</p>
 
 </overview>
 
--- a/cpp/ql/src/experimental/Security/CWE/CWE-552/DangerousUseRename.qhelp
+++ b/cpp/ql/src/experimental/Security/CWE/CWE-552/DangerousUseRename.qhelp
@@ -1,0 +1,23 @@
+<!DOCTYPE qhelp PUBLIC
+  "-//Semmle//qhelp//EN"
+  "qhelp.dtd">
+<qhelp>
+<overview>
+<p>If the rename function did not work correctly, simply overwriting the destination file can lead to errors. Use access to the destination file with permission setting.</p>
+
+</overview>
+
+<example>
+<p>The following example shows the erroneous and corrected ways to work with the rename function.</p>
+<sample src="DangerousUseRename.cpp" />
+
+</example>
+<references>
+
+<li>
+  CERT Coding Standard:
+  <a href="https://wiki.sei.cmu.edu/confluence/display/c/FIO15-C.+Ensure+that+file+operations+are+performed+in+a+secure+directory">FIO15-C. Ensure that file operations are performed in a secure directory - SEI CERT C Coding Standard - Confluence</a>.
+</li>
+
+</references>
+</qhelp>
--- a/cpp/ql/src/experimental/Security/CWE/CWE-552/DangerousUseRename.ql
+++ b/cpp/ql/src/experimental/Security/CWE/CWE-552/DangerousUseRename.ql
@@ -13,29 +13,20 @@
 import cpp
 import semmle.code.cpp.valuenumbering.GlobalValueNumbering
 
-/** Holds if the function argument is used in opening the file for reading. */
-predicate findFileForRead(FunctionCall fc, FunctionCall fc1) {
-  fc1.getTarget().hasGlobalOrStdOrBslName(["open", "fopen"]) and
+/** Holds if the function argument is used in opening the file for reading or writing. */
+predicate findFileForReadOrWrite(FunctionCall fc, FunctionCall fc1,string mode) {
+  fc1.getTarget().hasName(["open", "fopen"]) and
   globalValueNumber(fc1.getArgument(0)) = globalValueNumber(fc.getArgument(0)) and
-  fc1.getArgument(1).getValue() = "r" and
-  fc.getASuccessor+() = fc1
-}
-
-/** Holds if the function argument is used in opening the file for writing. */
-predicate findFileForWrite(FunctionCall fc, FunctionCall fc1) {
-  fc1.getTarget().hasGlobalOrStdOrBslName(["open", "fopen"]) and
-  globalValueNumber(fc1.getArgument(0)) = globalValueNumber(fc.getArgument(1)) and
-  fc1.getNumberOfArguments() = 2 and
-  fc1.getArgument(1).getValue() = "w" and
+  fc1.getArgument(1).getValue() = mode and
   fc.getASuccessor+() = fc1
 }
 
 from FunctionCall fc
 where
-  fc.getTarget().hasGlobalOrStdOrBslName("rename") and
+  fc.getTarget().hasName("rename") and
   exists(IfStmt ifst, Expr ec, Expr ecd, FunctionCall fc1, FunctionCall fc2 |
-    findFileForRead(fc, fc1) and
-    findFileForWrite(fc, fc2) and
+    findFileForReadOrWrite(fc, fc1, "r") and
+    findFileForReadOrWrite(fc, fc2, "w") and
     ec.getValue() = "0" and
     ecd = ifst.getCondition().getAChild*() and
     (
--- a/cpp/ql/src/experimental/Security/CWE/CWE-552/DangerousUseRename.ql
+++ b/cpp/ql/src/experimental/Security/CWE/CWE-552/DangerousUseRename.ql
@@ -2,7 +2,7 @@
  * @name Dangerous use of rename
  * @description If the rename function did not work correctly, simply overwriting the destination file can lead to errors.
  * @kind problem
- * @id cpp/dangerous-use-rename
+ * @id cpp/dangerous-manual-copy-after-rename
  * @problem.severity warning
  * @precision medium
  * @tags correctness
--- a/cpp/ql/src/experimental/Security/CWE/CWE-552/DangerousUseRename.ql
+++ b/cpp/ql/src/experimental/Security/CWE/CWE-552/DangerousUseRename.ql
@@ -31,7 +31,8 @@ where
     ecd = ifst.getCondition().getAChild*() and
     (
       globalValueNumber(ecd) = globalValueNumber(fc) and
-      not ecd.getParent() instanceof ComparisonOperation
+      not ecd.getParent() instanceof ComparisonOperation and
+      not ecd.getParent() instanceof NotExpr
       or
       exists(Expr evp |
         globalValueNumber(evp) = globalValueNumber(fc) and
--- a/cpp/ql/src/experimental/Security/CWE/CWE-552/DangerousUseRename.ql
+++ b/cpp/ql/src/experimental/Security/CWE/CWE-552/DangerousUseRename.ql
@@ -1,6 +1,6 @@
 /**
  * @name Dangerous use of rename
- * @description If the rename function did not work correctly, simply overwriting the destination file can lead to errors.
+ * @description If the rename function did not work correctly, attempting to copy the contents of a source file by opening the target file without assessing permissions creates a mechanism for deleting an arbitrary system file.
  * @kind problem
  * @id cpp/dangerous-manual-copy-after-rename
  * @problem.severity warning
@@ -55,4 +55,4 @@ where
     )
   )
 select fc,
-  "If the rename function did not work correctly, simply overwriting the destination file can lead to errors."
+  "If the rename function did not work correctly, attempting to copy the contents of a source file by opening the target file without assessing permissions creates a mechanism for deleting an arbitrary system file."
--- a/cpp/ql/src/experimental/Security/CWE/CWE-552/DangerousUseRename.ql
+++ b/cpp/ql/src/experimental/Security/CWE/CWE-552/DangerousUseRename.ql
@@ -1,0 +1,66 @@
+/**
+ * @name Dangerous use rename.
+ * @description If the rename function did not work correctly, simply overwriting the destination file can lead to errors.
+ * @kind problem
+ * @id cpp/dangerous-use-rename
+ * @problem.severity warning
+ * @precision medium
+ * @tags correctness
+ *       security
+ *       external/cwe/cwe-552
+ */
+
+import cpp
+import semmle.code.cpp.valuenumbering.GlobalValueNumbering
+
+/** Holds if the function argument is used in opening the file for reading. */
+predicate findFileForRead(FunctionCall fc, FunctionCall fc1) {
+  fc1.getTarget().hasGlobalOrStdOrBslName(["open", "fopen"]) and
+  globalValueNumber(fc1.getArgument(0)) = globalValueNumber(fc.getArgument(0)) and
+  fc1.getArgument(1).getValue() = "r" and
+  fc.getASuccessor+() = fc1
+}
+
+/** Holds if the function argument is used in opening the file for writing. */
+predicate findFileForWrite(FunctionCall fc, FunctionCall fc1) {
+  fc1.getTarget().hasGlobalOrStdOrBslName(["open", "fopen"]) and
+  globalValueNumber(fc1.getArgument(0)) = globalValueNumber(fc.getArgument(1)) and
+  fc1.getNumberOfArguments() = 2 and
+  fc1.getArgument(1).getValue() = "w" and
+  fc.getASuccessor+() = fc1
+}
+
+from FunctionCall fc
+where
+  fc.getTarget().hasGlobalOrStdOrBslName("rename") and
+  exists(IfStmt ifst, Expr ec, Expr ecd, FunctionCall fc1, FunctionCall fc2 |
+    findFileForRead(fc, fc1) and
+    findFileForWrite(fc, fc2) and
+    ec.getValue() = "0" and
+    ecd = ifst.getCondition().getAChild*() and
+    (
+      globalValueNumber(ecd) = globalValueNumber(fc) and
+      not ecd.getParent() instanceof ComparisonOperation
+      or
+      exists(Expr evp |
+        globalValueNumber(evp) = globalValueNumber(fc) and
+        ecd.(ComparisonOperation).hasOperands(evp, _)
+      )
+    ) and
+    (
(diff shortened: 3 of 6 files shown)


cpp/ql/src/experimental/Security/CWE/CWE-552/DangerousUseRename.cpp
```diff
@@ -0,0 +1,13 @@
...
+    return;
+  f1 = fopen(from, "r");
+  f2 = fopen(to, "w");
+...
+  if (rename(from,to)==0) // GOOD
+    return;
+  f1 = fopen(from, "r");
+  fd = open(to, O_WRONLY|O_CREAT|to_oflags, S_IRUSR|S_IWUSR);
+  f2 = fdopen(fd, "w");
+...
```
jketema, line 13: In both case you seem to check for the proper operation of `rename` before you do anything else. Are the examples correct?

cpp/ql/src/experimental/Security/CWE/CWE-552/DangerousUseRename.ql
```diff
@@ -0,0 +1,66 @@
...
+ *       security
+ *       external/cwe/cwe-552
+ */
+
+import cpp
+import semmle.code.cpp.valuenumbering.GlobalValueNumbering
+
+/** Holds if the function argument is used in opening the file for reading. */
+predicate findFileForRead(FunctionCall fc, FunctionCall fc1) {
+  fc1.getTarget().hasGlobalOrStdOrBslName(["open", "fopen"]) and
```
jketema, line 18: Does the `Bsl` namespace have an `open` or `fopen` function?

cpp/ql/src/experimental/Security/CWE/CWE-552/DangerousUseRename.ql
```diff
@@ -0,0 +1,66 @@
...
+  globalValueNumber(fc1.getArgument(0)) = globalValueNumber(fc.getArgument(0)) and
+  fc1.getArgument(1).getValue() = "r" and
+  fc.getASuccessor+() = fc1
+}
+
+/** Holds if the function argument is used in opening the file for writing. */
+predicate findFileForWrite(FunctionCall fc, FunctionCall fc1) {
```
jketema, line 25: Can this be merged with `findFileForWrite` into a predicate that has an extra `mode` argument?

cpp/ql/src/experimental/Security/CWE/CWE-552/DangerousUseRename.ql
```diff
@@ -0,0 +1,66 @@
...
+  fc1.getTarget().hasGlobalOrStdOrBslName(["open", "fopen"]) and
+  globalValueNumber(fc1.getArgument(0)) = globalValueNumber(fc.getArgument(1)) and
+  fc1.getNumberOfArguments() = 2 and
+  fc1.getArgument(1).getValue() = "w" and
```
jketema, line 29: What about cases where the `mode` argument consists of multiple characters?

cpp/ql/src/experimental/Security/CWE/CWE-552/DangerousUseRename.ql
jketema: Also note that none of this works with Posix `open`, as that takes a numerical `flags` argument. Is that intentional?

cpp/ql/test/experimental/query-tests/Security/CWE/CWE-552/semmle/tests/test.cpp
```diff
@@ -0,0 +1,174 @@
...
+void test1(const char *from,const char *to)
+{
+  FILE *f1 = NULL;
+  FILE *f2 = NULL;
+  char data[1000];
+  int count;
+  f1 = fopen(from, "r");
+  count = fread(data, 1, 1000, f1);
+  fclose(f1);
+  rename(from,to); // GOOD
```
jketema, line 29: Why is this `GOOD`? The return value of `rename` is not checked.

cpp/ql/test/experimental/query-tests/Security/CWE/CWE-552/semmle/tests/test.cpp
```diff
@@ -0,0 +1,174 @@
...
+  fclose(f2);
+}
+void test2(const char *from,const char *to)
+{
+  FILE *f1 = NULL;
+  FILE *f2 = NULL;
+  char data[1000];
+  int count;
+  
+  if (!rename(from,to)) { // BAD
```
jketema, line 40: Why is this `BAD`? The rename seems to have succeeded in this case.

cpp/ql/test/experimental/query-tests/Security/CWE/CWE-552/semmle/tests/test.cpp
```diff
@@ -0,0 +1,174 @@
...
+}
+void test4(const char *from,const char *to)
+{
+  FILE *f1 = NULL;
+  FILE *f2 = NULL;
+  char data[1000];
+  int count;
+  int fd;
+
+  if (rename(from,to)<0) { // GOOD
```
jketema, line 74: Why is this `GOOD`? The rename seems to have failed in this case.

jketema (commented):
Hi @ihsinme,

Thanks for your contribution. A first few questions from my side.

cpp/ql/src/experimental/Security/CWE/CWE-552/DangerousUseRename.cpp
ihsinme: if `rename` did not work, then in the first example we simply use unrestricted access to files, and in the second we restrict it with flags.

cpp/ql/src/experimental/Security/CWE/CWE-552/DangerousUseRename.ql
ihsinme: I'll think about `Posix`. Thanks for the comment.

cpp/ql/src/experimental/Security/CWE/CWE-552/DangerousUseRename.ql
ihsinme: It is important for me to find a job with one flag.
I want to avoid false positives when the files is then manipulated.

cpp/ql/test/experimental/query-tests/Security/CWE/CWE-552/semmle/tests/test.cpp
ihsinme: in the case of opening a file for writing, it occurs with the use of restrictive flags.

cpp/ql/test/experimental/query-tests/Security/CWE/CWE-552/semmle/tests/test.cpp
ihsinme: you are right the example is not successful, from the point of view of the correctness of work with `rename`.
I suggest deleting it.

cpp/ql/test/experimental/query-tests/Security/CWE/CWE-552/semmle/tests/test.cpp
ihsinme: Thank you for your comment.
i fixed this place.

cpp/ql/src/experimental/Security/CWE/CWE-552/DangerousUseRename.cpp
jketema: I see now. This needs to be explained in the `qhelp` file. Also would to be possible to make the examples more focussed? Having `f1 = fopen(from, "r");` in there confuses me, and it's not clear to me why it's there at all.

Given the examples, it also seems that in your query you combine two cases that should really be separate:

1. `rename` fails and it followed by an `fopen` call.
2. the result of `rename` does not matter and an `open` call follows with an appropriate set of `mode` flags.

cpp/ql/test/experimental/query-tests/Security/CWE/CWE-552/semmle/tests/test.cpp
jketema: Please change this to `BAD` and ensure it's capture by your query if possible.

cpp/ql/test/experimental/query-tests/Security/CWE/CWE-552/semmle/tests/test.cpp
jketema: Can you explain this after the `// GOOD`, like you do after `// BAD` in `test3`?

cpp/ql/src/experimental/Security/CWE/CWE-552/DangerousUseRename.ql
jketema: Not sure why you're confused. `bsl` is one of the namespaces of this library: https://github.com/bloomberg/bde, does that help?

ihsinme: I took it as the most general method, including all possible options.
thanks for paying attention.

cpp/ql/src/experimental/Security/CWE/CWE-552/DangerousUseRename.cpp
ihsinme: looks really confusing.

I only work with the first situation.
`rename` failed and an attempt was made to rename the path to an open file without permissions.

how do you look at such a change `qhelp`

```
If the rename function did not work correctly, attempting to overwrite the target file may result in third-party files being affected, for example through the linking mechanism. Use access to the target file with permission setting.
If the rename function did not work correctly, attempting to overwrite the target file may result in third-party files being affected, for example through the linking mechanism. Use access to the target file with permission setting.
```

cpp/ql/src/experimental/Security/CWE/CWE-552/DangerousUseRename.cpp
```diff
@@ -0,0 +1,11 @@
+
+...
+  if (rename(from,to)==0) // BAD
+    return;
+...
```
jketema, line 5: This example is now effectively empty. Is that what you wanted?

cpp/ql/src/experimental/Security/CWE/CWE-552/DangerousUseRename.qhelp
```diff
@@ -0,0 +1,23 @@
+<!DOCTYPE qhelp PUBLIC
+  "-//Semmle//qhelp//EN"
+  "qhelp.dtd">
+<qhelp>
+<overview>
+<p>If the rename function did not work correctly, simply overwriting the destination file can lead to errors. Use access to the destination file with permission setting.</p>
```
jketema, line 6: What kind of errors can occur?

cpp/ql/src/experimental/Security/CWE/CWE-552/DangerousUseRename.qhelp
```diff
@@ -0,0 +1,23 @@
...
+
+</overview>
+
+<example>
+<p>The following example shows the erroneous and corrected ways to work with the rename function.</p>
```
jketema, line 11: Please explain what is erroneous about the erroneous example.

cpp/ql/src/experimental/Security/CWE/CWE-552/DangerousUseRename.ql
```diff
@@ -0,0 +1,58 @@
+/**
+ * @name Dangerous use rename.
```
jketema, line 2:
```suggestion
 * @name Dangerous use of rename
```

cpp/ql/src/experimental/Security/CWE/CWE-552/DangerousUseRename.ql
```diff
@@ -0,0 +1,58 @@
...
+  globalValueNumber(fc1.getArgument(0)) = globalValueNumber(fc.getArgument(na)) and
+  fc1.getArgument(1).getValue() = mode and
+  fc.getASuccessor+() = fc1
+}
+
+from FunctionCall fc
+where
+  fc.getTarget().hasName("rename") and
+  exists(IfStmt ifst, Expr ec, Expr ecd, FunctionCall fc1, FunctionCall fc2 |
+    findFileForReadOrWrite(fc, fc1, "r", 0) and
```
jketema, line 28: Why are you interested in `open(from, "r")`? I see this back in all the examples/tests, but the text output by `select` does not make mention of this and effectively only refers to `open(to, "w")`

cpp/ql/src/experimental/Security/CWE/CWE-552/DangerousUseRename.qhelp
ihsinme: arbitrary files can be deleted on behalf of a running program.

cpp/ql/src/experimental/Security/CWE/CWE-552/DangerousUseRename.cpp
ihsinme: yes I must have been too hasty

cpp/ql/src/experimental/Security/CWE/CWE-552/DangerousUseRename.qhelp
ihsinme: The target file is opened without checking the access rights and its existence, in fact the program overwrites this file.

cpp/ql/src/experimental/Security/CWE/CWE-552/DangerousUseRename.ql
ihsinme: this is necessary to detect exactly the manual renaming of files when a source file and a target file are encountered and the data is copied.

cpp/ql/src/experimental/Security/CWE/CWE-552/DangerousUseRename.qhelp
jketema: This needs to be mentioned here.

cpp/ql/src/experimental/Security/CWE/CWE-552/DangerousUseRename.ql
jketema: Right, so are you interested in situations where, first, the renaming fails, and then the source and target files are opened and the contents is copied manually. If so, *all* the comments and qhelp need to explain this, because at the moment the impression I get from the qhelp is that you're just interested in failing renames followed by opening of the target file.

cpp/ql/src/experimental/Security/CWE/CWE-552/DangerousUseRename.qhelp
ihsinme: I corrected the inscription in qhelp, how do you like it?

cpp/ql/src/experimental/Security/CWE/CWE-552/DangerousUseRename.ql
```diff
@@ -0,0 +1,58 @@
+/**
+ * @name Dangerous use of rename
+ * @description If the rename function did not work correctly, simply overwriting the destination file can lead to errors.
```
jketema, line 3: Can you update these to match the qlhelp file?

cpp/ql/src/experimental/Security/CWE/CWE-552/DangerousUseRename.ql
```diff
@@ -0,0 +1,58 @@
...
+ * @kind problem
+ * @id cpp/dangerous-use-rename
```
jketema, line 5: Maybe:
```suggestion
 * @id cpp/dangerous-manual-copy-after-rename
```

cpp/ql/src/experimental/Security/CWE/CWE-552/DangerousUseRename.ql
```diff
@@ -0,0 +1,58 @@
...
+        or
+        not ecd.(EQExpr).getLeftOperand().getValue() = "0" and
+        not ecd.(EQExpr).getRightOperand().getValue() = "0"
+      ) and
+      ifst.getThen() = fc1.getEnclosingStmt().getParentStmt*() and
+      ifst.getThen() = fc2.getEnclosingStmt().getParentStmt*()
+    )
+  )
+select fc,
+  "If the rename function did not work correctly, simply overwriting the destination file can lead to errors."
```
jketema, line 58: Can you update this too to match the qlhelp?

cpp/ql/src/experimental/Security/CWE/CWE-552/DangerousUseRename.ql
```diff
@@ -0,0 +1,58 @@
...
+  fc1.getTarget().hasName(["open", "fopen"]) and
+  globalValueNumber(fc1.getArgument(0)) = globalValueNumber(fc.getArgument(na)) and
+  fc1.getArgument(1).getValue() = mode and
+  fc.getASuccessor+() = fc1
+}
+
+from FunctionCall fc
+where
+  fc.getTarget().hasName("rename") and
+  exists(IfStmt ifst, Expr ec, Expr ecd, FunctionCall fc1, FunctionCall fc2 |
```
jketema, line 27:
```suggestion
  exists(IfStmt ifst, Expr ec, Expr ecd, FunctionCall readCall, FunctionCall writeCall |
```
with matching changes below of course.

cpp/ql/src/experimental/Security/CWE/CWE-552/DangerousUseRename.ql
```diff
@@ -0,0 +1,58 @@
...
+
+/** Holds if the function argument is used in opening the file for reading or writing. */
+predicate findFileForReadOrWrite(FunctionCall fc, FunctionCall fc1, string mode, int na) {
+  fc1.getTarget().hasName(["open", "fopen"]) and
+  globalValueNumber(fc1.getArgument(0)) = globalValueNumber(fc.getArgument(na)) and
+  fc1.getArgument(1).getValue() = mode and
+  fc.getASuccessor+() = fc1
+}
+
+from FunctionCall fc
```
jketema, line 24:
```suggestion
from FunctionCall renameCall
```
with matching changes below of course.

cpp/ql/src/experimental/Security/CWE/CWE-552/DangerousUseRename.ql
```diff
@@ -0,0 +1,58 @@
...
+where
+  fc.getTarget().hasName("rename") and
+  exists(IfStmt ifst, Expr ec, Expr ecd, FunctionCall fc1, FunctionCall fc2 |
+    findFileForReadOrWrite(fc, fc1, "r", 0) and
+    findFileForReadOrWrite(fc, fc2, "w", 1) and
+    ec.getValue() = "0" and
```
jketema, line 30: Could you move this down to where it's actually used? That is, to just above `ecd.(EQExpr).hasOperands(_, ec)`

cpp/ql/src/experimental/Security/CWE/CWE-552/DangerousUseRename.ql
```diff
@@ -0,0 +1,58 @@
...
+      exists(Expr evp |
+        globalValueNumber(evp) = globalValueNumber(renameCall) and
+        ecd.(ComparisonOperation).hasOperands(evp, _)
+      )
+    ) and
+    (
+      ec.getValue() = "0" and
+      ecd.(EQExpr).hasOperands(_, ec) and
+      forall(Expr st | st = ifst.getThen().getASuccessor*() | st != readCall) and
+      forall(Expr st | st = ifst.getThen().getASuccessor*() | st != writeCall)
```
jketema, line 45: Which of your test cases covers this?

jketema:
Two general comments, which also apply to most of your other queries:

* I would help reviewers if you pick informative names for variables. For example, looking at this PR, from a reviewer's perspective it's hard to understand what variables like `ec`, `ecd` and `ecv` are supposed to represent. I'm not even sure if this is totally consistent throughout the query, we leads me to my other point.
* I would help if things that logically belong together in your queries occur textually close and a separate from cases that are actually different. For example, in your current query, when I read [this first part](https://github.com/github/codeql/pull/9090/files#diff-c13d8f6a2f12a75ee83d7b9415c47c2b4ebb62ac0ee28c58c58dee994e4b4c2bR31-R40), I expect [this following part](https://github.com/github/codeql/pull/9090/files#diff-c13d8f6a2f12a75ee83d7b9415c47c2b4ebb62ac0ee28c58c58dee994e4b4c2bR42-R45) to apply to both cases described in that first part. However, that is clearly not the case, as [this](c13d8f6a2f12a75ee83d7b9415c47c2b4ebb62ac0ee28c58c58dee994e4b4c2bR32) rules out that `ecd` is a `ComparisonOperator`, while that is actually needed [here](https://github.com/github/codeql/pull/9090/files#diff-c13d8f6a2f12a75ee83d7b9415c47c2b4ebb62ac0ee28c58c58dee994e4b4c2bR42-R43).

jketema:
To add to the above: this would allow us to give faster and better feedback, which is important, especially when there's a large number of PRs open.

ihsinme:
You are absolutely right, I will try to follow the rules in the future.
Thanks.

cpp/ql/src/experimental/Security/CWE/CWE-552/DangerousUseRename.ql
ihsinme: `test5` `test6` `test7` `test8`

jketema: I don't see this. If we look at `test5` for example:
```
  if (rename(from,to)==0) // BAD
    return;
  f1 = fopen(from, "r");
  count = fread(data, 1, 1000, f1);
  fclose(f1);
  remove(to);
  f2 = fopen(to, "w");
  fwrite(data, count, 1, f2);
  fclose(f2);
```
Then, `ifst.getThen().getASuccessor*()` gives you all the statements and expressions textually lower than `if (rename(from,to)==0)`, including `fopen(from, "r")` and `fopen(to, "w")`, so the `forall` cannot hold.

jketema:
Note by the way that `cpp/world-writable-file-creation` exists. Which should capture most, if not all cases here. That query just checks whether a file is created (using `fopen`, e.g.) using very lax permissions.

ihsinme:
I apologize for the wasted time.

I am very serious about de-duplicates in my PRs, but in this case my review system failed and I didn't notice this request.

Sorry again, I'll try to be more careful in the future.

PR #9091, opened 2022-05-09
C#: Port the existing compiler-tracing.spec files to Lua.

criemen:
Right now, this is an unmodified port, without the additional functionality for `dotnet build` that we envision.

This means that we can switch back and forth between Lua and legacy tracing (ideally) without user-impacting differences.

### DIFF
--- a/csharp/tools/tracing-config.lua
+++ b/csharp/tools/tracing-config.lua
@@ -1,10 +1,7 @@
 function RegisterExtractorPack()
     local extractor = GetPlatformToolsDirectory() ..
                           'Semmle.Extraction.CSharp.Driver'
-    if OperatingSystem == 'windows' then
-        extractor = GetPlatformToolsDirectory() ..
-                        'Semmle.Extraction.CSharp.Driver.exe'
-    end
+    if OperatingSystem == 'windows' then extractor = extractor .. '.exe' end
     local windowsMatchers = {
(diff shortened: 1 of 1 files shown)


cklin (approved):
I think the titles of the commit and the PR should both mention csharp specifically to reduce ambiguity.  Other than that 👍

criemen:
@cklin good point, thanks. Fixed now.

cklin (approved):
👍 but I don't have write access so you will have to get approval from someone else too.

csharp/tools/tracing-config.lua
```diff
@@ -0,0 +1,55 @@
+function RegisterExtractorPack()
+    local extractor = GetPlatformToolsDirectory() ..
+                          'Semmle.Extraction.CSharp.Driver'
+    if OperatingSystem == 'windows' then
+        extractor = GetPlatformToolsDirectory() ..
+                        'Semmle.Extraction.CSharp.Driver.exe'
```
hvitved, line 6: I don't know lua, but could you simply append `.exe` to `extractor` in this case?

hvitved (commented):
Looks plausible to me.

csharp/tools/tracing-config.lua
criemen: Good point, yes that's possible

hvitved (approved):

PR #9094, opened 2022-05-09
Swift codegen: add predicate properties

redsun82:
Properties marked with `predicate` in the schema are now accepted.

* in the dbscheme, they will translate to a table with a single `id`
  column (and the table name will not be pluralized)
* in C++ classes, they will translate to `bool` fields
* in QL classes, they will translate to predicates

Closes https://github.com/github/codeql-c-team/issues/1016


MathiasVP (approved):
LGTM!

PR #9099, opened 2022-05-10
Python dataflow/flow summaries from scratch git discrepancy

yoff:
I observed to my horror that the code being reviewed was not the code I had written, which is a terrible waste of reviewer's time. Specifically, I locally have the same list of commits (with the same shas) as listed online, but my local code differs from what is shown under `Files changed`. So now I created a new branch from my local situation to see how that looks online.


yoff:
Aha, it may have been an artefact of not expanding a large diff...

PR #9100, opened 2022-05-10
Swift: changes required for TBD node rework

redsun82:
These changes are required to allow a new type-safe approach to TBD
nodes, that will come in a separate commit.

This introduces:
* the possibility to add properties to the root `Element`
* a functor taking tags to the corresponding binding trap entry
* `hasProp()` methods for optional properties in QL
* `getPrimaryQlClass()` method

### DIFF
--- a/swift/codegen/templates/ql_class.mustache
+++ b/swift/codegen/templates/ql_class.mustache
@@ -8,6 +8,8 @@ class {{name}}Base extends {{db_id}}{{#bases}}, {{.}}{{/bases}} {
   {{#root}}
   string toString() { none() } // overridden by subclasses
 
+  string getPrimaryQlClass() { none() } // overridden by subclasses
+
   {{name}}Base getResolveStep() { none() } // overridden by subclasses
 
   {{name}}Base resolve() {
@@ -17,7 +19,7 @@ class {{name}}Base extends {{db_id}}{{#bases}}, {{.}}{{/bases}} {
   }
   {{/root}}
   {{#final}}
-  override string toString() { result = "{{name}}" }
+  override string getPrimaryQlClass() { result = "{{name}}" }
   {{/final}}
   {{#properties}}
 
@@ -32,6 +34,12 @@ class {{name}}Base extends {{db_id}}{{#bases}}, {{.}}{{/bases}} {
     {{tablename}}({{#tableparams}}{{^first}}, {{/first}}{{param}}{{/tableparams}})
     {{/type_is_class}}
   }
+  {{#is_optional}}
+
+  predicate has{{singular}}({{#is_repeated}}int index{{/is_repeated}}) {
+    exists({{getter}}({{#is_repeated}}index{{/is_repeated}}))
+  }
+  {{/is_optional}}
   {{#is_repeated}}
 
   {{type}} {{indefinite_getter}}() {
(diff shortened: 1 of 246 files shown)


swift/codegen/templates/ql_class.mustache
```diff
@@ -8,6 +8,8 @@ class {{name}}Base extends {{db_id}}{{#bases}}, {{.}}{{/bases}} {
   {{#root}}
   string toString() { none() } // overridden by subclasses
 
+  string getPrimaryQlClass() { none() } // overridden by subclasses
```
MathiasVP, line 11: The predicate should actually be called `getAPrimaryQlClass`, and we should have a corresponding `getPrimaryQlClasses` defined as:
```ql
final string getPrimaryQlClasses() { result = concat(this.getAPrimaryQlClass(), ",") }
```
Currently, for Swift, we only return a single result. But that's not true for C/C++ today, and it's not necessarily the case in the future.
(See https://github.com/github/codeql/blob/main/cpp/ql/lib/semmle/code/cpp/Element.qll#L53 for how it's done for C/C++.)

AlexDenisov (dismissed):

swift/ql/lib/codeql/swift/generated/expr/Argument.qll
```diff
@@ -3,7 +3,7 @@ import codeql.swift.elements.Element
 import codeql.swift.elements.expr.Expr
 
 class ArgumentBase extends @argument, Element {
-  override string toString() { result = "Argument" }
+  override string getPrimaryQlClass() { result = "Argument" }
```
MathiasVP, line 6: So where is `toString` now actually overridden for these classes?

redsun82: this is in the companion PR I'm writing, which implements it in the hand-written `Element` override

swift/codegen/templates/ql_class.mustache
redsun82: I see!

swift/ql/lib/codeql/swift/generated/expr/Argument.qll
MathiasVP: Awesome. No complaints from me here then :)

MathiasVP (dismissed):
LGTM!

swift/ql/lib/codeql/swift/generated/File.qll
```diff
@@ -2,7 +2,7 @@
 import codeql.swift.elements.Element
 
 class FileBase extends @file, Element {
-  override string toString() { result = "File" }
+  override string getAPrimaryQlClass() { result = "File" }
```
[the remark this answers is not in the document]
redsun82: oh come on

MathiasVP:
The Code Scanning alerts are all correct (the implementation of `getAPrimaryQlClass` in a class `P` should really be `result = "P"`), but Swift is a special-case due to the whole autogenerated base class setup.

MathiasVP:
I'd prefer not including 9707cd68c4208210384fc1c4bdd0da671ceed28a. Instead, we should fixup the ql-for-ql query (later) to exclude things inside autogenerated folders.

redsun82:
> I'd prefer not including [9707cd6](https://github.com/github/codeql/commit/9707cd68c4208210384fc1c4bdd0da671ceed28a). Instead, we should fixup the ql-for-ql query (later) to exclude things inside autogenerated folders.

ok, I reverted it. I was under the false impression that was blocking the merge.

MathiasVP:
> ok, I reverted it. I was under the false impression that was blocking the merge.

Ah, sorry. Not at all.

MathiasVP (approved):

PR #9107, opened 2022-05-10
Swift: transfer `TrapArena`

redsun82:


swift/extractor/trap/TrapArena.h
```diff
@@ -0,0 +1,70 @@
+//
+// Created by redsun82 on 18.01.22.
```
AlexDenisov, line 2: These are autogenerated by an IDE, right? Do we want to keep them?

redsun82: uh, right, forgot to remove that! yeah, it was CLion before I removed the template

swift/extractor/trap/TrapArena.h
```diff
@@ -0,0 +1,70 @@
...
+
+#include "swift/extractor/trap/TrapLabel.h"
+
+namespace codeql {
+
+// TrapArena has the responsibilities to
+// * allocate distinct trap #-labels outputting their assignment to '*' or a @-keys to a trap output
+// * forwarding trap entries to said output
+// TODO: split or move these responsibilities into separate classes
+class TrapArena {
```
jketema, line 19: I find the name very uninformative. Maybe just address the above todo, and split this into parts that do have informative names?

swift/extractor/trap/TrapArena.h
```diff
@@ -0,0 +1,70 @@
...
+ public:
+  explicit TrapArena(std::ostream& out) : out_{out} {}
+
+  // get a new label for *
+  template <typename Tag>
+  TrapLabel<Tag> getLabel() {
+    auto ret = allocateLabel<Tag>();
+    print(ret, "=*");
```
jketema, line 27: If you write "get", I expect it to do just that, but it's also printing (same for the other `getLabel`s)?

swift/extractor/trap/TrapLabel.h
```diff
@@ -34,6 +34,9 @@ class TrapLabel : public UntypedTrapLabel {
 
   using UntypedTrapLabel::UntypedTrapLabel;
 
+  // we want only TrapArena to create non-default labels
+  friend class TrapArena;
```
jketema, line 38: Why is this necessary? I also do not understand the comment here.

redsun82: we want just one place creating labels, to avoid conflicting labels

swift/extractor/trap/TrapArena.h
redsun82: good point

swift/extractor/trap/TrapArena.h
redsun82: yeah, I'm thinking about putting the printing responsibility in a `TrapFile` class, and moving the allocation to either the dispatcher, or the upcoming label store

swift/extractor/trap/TrapLabel.h
jketema: That doesn't explain why this needs to be a friend. Also with what you just wrote, it more seems like you want some kind of singleton?

redsun82: no, I hate singletons :laughing:

The "one place" was meant in a "one place in code" to make reasoning about it easier when reading the code. We also don't want anyone to create Labels from integers by mistake.

We could have labels as a nested class of whoever needs to create them, and then something like
```
using TrapLabel = WhoeverBuildsLabels::Label;
```

jketema: Err. Ok, I don't see how you enforce uniqueness of labels then, as `uint64_t id_{0};` in the `TrapArena` is not static.

redsun82: We know there will be one TrapArena (or anything taking its place) per trap file, which is then enough. `TrapLabel` are handled everywhere, so the true risk of conflicting labels is more writing `TrapLabel label{0}` (maybe by someone new reading the public interface of `TrapLabel` and seeing that constructor), rather than multiple `TrapArena`s existing for the same file.

Would you do something else here?

sashabu: Another option is to keep the unsafe "constructor" public (because it might be good to use in tests too) but "name" it something scary: that is, make the constructor itself private, but add a public static function that calls it. Something along the lines of:

```
 private:
  using UntypedTrapLabel::UntypedTrapLabel;
 public:
  // The caller is responsible for ensuring ID uniqueness.
  static TrapLabel UnsafeCreateFromExplicitId(uint64_t id) { return {id}; }
```

Compared to `friend`, this has the benefit of being more restrictive (i.e. it only exposes this specific constructor, but not any other privates), but the downside (or also a benefit?) of losing control over who can call it.

I think `friend` would be fine here, but might suggest moving the `TrapArena` into the same header as `TrapLabel` if they're part of the same API (the user needs to construct the label and then use it - the spelling of those steps is an implementation detail).

jketema: Right, so that warrants a comment here: https://github.com/github/codeql/pull/9107/files#diff-c25ca6b4acbbb5f8f0d20adf6aea2ceaf6c94f5e7ede65cca4546543669b8e5fR79, not at its current location.

I also still do not understand you this needs to be a friend class.

swift/extractor/trap/TrapArena.h
redsun82: I've moved most of the `TrapArena` guts (the outputting part) to `TrapOutput`, also pulling in the trap temp file creation and moving into place.

With this, `TrapArena` is holding very little code, that will be moved elsewhere in upcoming PRs

swift/extractor/trap/TrapLabel.h
redsun82: I like the `unsafe` wording, and the option it opens up to create distinct labels in tests

AlexDenisov:
This separation looks much better indeed!

Though, I strongly disagree with the move of the `ostream` initialization to the `TrapOutput` constructor - if the opening fails, then we get somewhat half-broken object. The same applies to the move semantics.
Do we actually need to make it movable?

redsun82:
> This separation looks much better indeed!
> 
> Though, I strongly disagree with the move of the `ostream` initialization to the `TrapOutput` constructor - if the opening fails, then we get somewhat half-broken object. The same applies to the move semantics. Do we actually need to make it movable?

On the move semantics, I think it makes the use easier, clearer and safer. We can have it then moved inside the dispatcher with a clear ownership transfer... but I see I'm getting dangerously near the raw references discussion :hand_over_mouth: 

About the opening failure (or anything leading to it like directory creation): it's a valid concern, but I think it's one that should be tackled with a proper error handling strategy that we do not have yet (and this is precisely the background to [my question on slack](https://github.slack.com/archives/C02TU8JET9B/p1652273990921369)). This is not much different from how `ofstream` itself works, where failure to open leaves us with a half-broken object as well: what do we do? Well, without exceptions one should check the object is valid, definitely after creation, ideally after every operation, and this is enabled as well by this.

Notice that the behaviour by `TrapOutput` on creation errors is practically the same as the default behaviour of `ofstream`: it converts to false, and any further operation on it is silently ignore. We can put some better error handling though, and specifically do a better job at deciding if and when to abort moving the object in place at destruction time.

sashabu:
> > This separation looks much better indeed!
> > Though, I strongly disagree with the move of the `ostream` initialization to the `TrapOutput` constructor - if the opening fails, then we get somewhat half-broken object. The same applies to the move semantics. Do we actually need to make it movable?
> 
> On the move semantics, I think it makes the use easier, clearer and safer. We can have it then moved inside the dispatcher with a clear ownership transfer... but I see I'm getting dangerously near the raw references discussion 🤭 and about moving and half-brokenness, do you mean the moved from instance? Because the moved-from object is left in a perfectly valid state, but one with which you can't do much, pretty much as the standard dictates for move.
> 
> About the opening failure (or anything leading to it like directory creation): it's a valid concern, but I think it's one that should be tackled with a proper error handling strategy that we do not have yet (and this is precisely the background to [my question on slack](https://github.slack.com/archives/C02TU8JET9B/p1652273990921369)). This is not much different from how `ofstream` itself works, where failure to open leaves us with a half-broken object as well: what do we do? Well, without exceptions one should check the object is valid, definitely after creation, ideally after every operation, and this is enabled as well by `TrapOutput` via the bool operator (we can add a more explicit `valid()` method or something similar if you prefer).
> 
> Notice that the behaviour by `TrapOutput` on creation errors is practically the same as the default behaviour of `ofstream`: it converts to false, and any further operation on it is silently ignored. We can put some better error handling though, and specifically do a better job at deciding if and when to abort moving the object in place at destruction time. This code does exactly the same as the current state of `main`.

Another option is making the constructor private and adding a factory function that returns an `std::optional<TrapOutput>` so the caller has to handle the empty case.

redsun82:
> > > This separation looks much better indeed!
> > > Though, I strongly disagree with the move of the `ostream` initialization to the `TrapOutput` constructor - if the opening fails, then we get somewhat half-broken object. The same applies to the move semantics. Do we actually need to make it movable?
> > 
> > 
> > On the move semantics, I think it makes the use easier, clearer and safer. We can have it then moved inside the dispatcher with a clear ownership transfer... but I see I'm getting dangerously near the raw references discussion hand_over_mouth and about moving and half-brokenness, do you mean the moved from instance? Because the moved-from object is left in a perfectly valid state, but one with which you can't do much, pretty much as the standard dictates for move.
> > About the opening failure (or anything leading to it like directory creation): it's a valid concern, but I think it's one that should be tackled with a proper error handling strategy that we do not have yet (and this is precisely the background to [my question on slack](https://github.slack.com/archives/C02TU8JET9B/p1652273990921369)). This is not much different from how `ofstream` itself works, where failure to open leaves us with a half-broken object as well: what do we do? Well, without exceptions one should check the object is valid, definitely after creation, ideally after every operation, and this is enabled as well by `TrapOutput` via the bool operator (we can add a more explicit `valid()` method or something similar if you prefer).
> > Notice that the behaviour by `TrapOutput` on creation errors is practically the same as the default behaviour of `ofstream`: it converts to false, and any further operation on it is silently ignored. We can put some better error handling though, and specifically do a better job at deciding if and when to abort moving the object in place at destruction time. This code does exactly the same as the current state of `main`.
> 
> Another option is making the constructor private and adding a factory function that returns an `std::optional<TrapOutput>` so the caller has to handle the empty case.

ah exactly, I wanted to write that and then forgot. Similarly, the factory function can return `std::unique_ptr` and then one can scratch move semantics as it gets borrowed from there (at the cost of using heap). But all in all when possible I quite like `ofstream`'s approach, baking the optionality into the object itself via `explicit operator bool`, but that gets hard when a `null` state is less easy to define within the class invariants.

redsun82:
@AlexDenisov @sashabu how would you feel if FS related errors led to `std::abort` in `TrapOutput`, then a `TrapOutput` instance would always be a well defined object (and we could scratch `explicit operator bool` and the corresponding error checking). In any case stopping extraction seems like the most sensible thing to do in this case

AlexDenisov:
> @AlexDenisov @sashabu how would you feel if FS related errors led to `std::abort` in `TrapOutput`, then a `TrapOutput` instance would always be a well defined object (and we could scratch `explicit operator bool` and the corresponding error checking). In any case stopping extraction seems like the most sensible thing to do in this case

We can also rollback to the previous implementation and only create the object when we can do so with a "valid" stream.
Otherwise, it's up to you 😄

swift/extractor/trap/TrapLabel.h
```diff
@@ -56,8 +59,10 @@ inline auto trapQuoted(const std::string& s) {
   return std::quoted(s, '"', '"');
 }
 
-template <typename Tag>
+template <typename TagParam>
 struct Binding {
+  using Tag = TagParam;
```
jketema, line 64: What's the reason for changing this?

swift/extractor/trap/TrapOutput.h
```diff
@@ -0,0 +1,119 @@
...
+  TrapOutput(TrapOutput&& other) : out_{std::move(other.out_)}, target_{std::move(other.target_)} {
+    assert(!other.out_.is_open());
+  }
+
+  TrapOutput& operator=(TrapOutput&& other) {
+    close();
+    out_ = std::move(other.out_);
+    target_ = std::move(other.target_);
+    return *this;
+  }
```
jketema, line 65: Excuse my ignorance, but why is this needed (doesn't seem to be quite covered by the ongoing discussion)?

redsun82: it's gone now :laughing:

swift/extractor/trap/TrapLabel.h
redsun82: this allows to get a tag relevant for an `Entry` with `Entry::Tag`, which is needed in some functions using `Entry` as a template parameter. But to be honest, I don't remember any more where I used that (or if it's still used in the current state of the code). Still, accessing template parameters with nested type aliases is a common enough requirement when dealing with templates.

jketema: Ah, I had not immediately realized that `Tag` is public. I'd personally wait with introducing this until I really need it.

redsun82: done

redsun82:
for the record, after a live chat with @AlexDenisov I conceded that moving the file opening logic to `TrapOutput` might be premature, and should be done with more context later.

sashabu (approved):

PR #9109, opened 2022-05-10
Initial Kotlin support

igfoo:

PR #9110, opened 2022-05-10
QL: add implicit -all to a query pack to match the CodeQL pack resolution

erik-krogh:
Resolves the QL-for-QL errors seen here: https://github.com/github/codeql/pull/9109


tausbn (approved):
Changes LGTM. :+1:

PR #9113, opened 2022-05-11
Claim Go 1.18 support

smowton:
As of https://github.com/github/codeql-go/pull/686 landing we support extracting generics, dataflow analysis in programs that use generics, etc. Note this hasn't  gone out in a release yet but I would expect it to be in 2.9.2.


atorralba (approved):

MathiasVP:
> As of [github/codeql-go#686](https://github.com/github/codeql-go/pull/686) landing we support extracting generics, dataflow analysis in programs that use generics, etc. Note this hasn't gone out in a release yet but I would expect it to be in 2.9.2.

It should indeed be in 2.9.2, yes. It's certainly part of the [change log](https://github.com/github/codeql-go/pull/736) at least. 😂.

PR #9118, opened 2022-05-11
C#: Tag telemetry queries with `telemetry`

henrymercer:
This will exclude the results of these queries from the summary tables produced by `codeql database analyze` in a future version of the CodeQL CLI.


michaelnebel (approved):
LGTM!

PR #9119, opened 2022-05-11
Fix non-US spellings and the corresponding query

nickrolfe:
The query was missing multi-line comments containing 'analyse', because `.` does not match newline characters in `regexpMatch`.

### DIFF
--- a/ql/ql/src/codeql_ql/style/docs/NonUSSpellingQuery.qll
+++ b/ql/ql/src/codeql_ql/style/docs/NonUSSpellingQuery.qll
@@ -18,7 +18,7 @@ predicate contains_non_us_spelling(string s, string wrong, string right) {
     wrong != "analyse"
     or
     // analyses (as a noun) is fine
-    s.regexpMatch(".*analyse[^s].*") and
+    s.regexpReplaceAll("[\\r\\n]", " ").regexpMatch(".*analyse[^s].*") and
     wrong = "analyse"
   )
 }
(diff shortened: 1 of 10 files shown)


jketema (commented):
Small comment, otherwise C++ 👍

erik-krogh (approved):

jketema (dismissed):

tausbn (approved):

ql/ql/src/codeql_ql/style/docs/NonUSSpellingQuery.qll
```diff
@@ -18,7 +18,7 @@ predicate contains_non_us_spelling(string s, string wrong, string right) {
     wrong != "analyse"
     or
     // analyses (as a noun) is fine
-    s.regexpMatch(".*analyse[^s].*") and
+    s.regexpReplaceAll("[\\r\\n]", " ").regexpMatch(".*analyse[^s].*") and
```
tausbn, line 21: I wonder if this is faster or slower than using `[\\s\\S]` (IIRC)to match any character.

atorralba (approved):
Java 👍

nickrolfe:
I just pushed a commit to fix the changenote formatting. Looks like I got approvals from all languages before I did that, so I'll merge as soon as I get one new approval (and CI checks pass).

erik-krogh (approved):

nickrolfe:
Finally, the tests pass!

ql/ql/src/codeql_ql/style/docs/NonUSSpellingQuery.qll
nickrolfe: I'm not sure, but I doubt it would be noticeable either way given the number of strings we're dealing with.

I'm about to gut this query to generalise it, and will likely replace this code, anyway.

haaami01:
Thanks!  all 31 tests were successful✅👌

PR #9122, opened 2022-05-11
Kotlin: Apply changes since https://github.com/github/codeql/pull/9109 branched away from kotlin-main

smowton:


java/ql/lib/semmle/code/java/frameworks/KotlinStdLib.qll
```diff
@@ -0,0 +1,13 @@
...
+
+import java
+private import semmle.code.java.dataflow.ExternalFlow
+
+private class KotlinStdLibSummaryCsv extends SummaryModelCsv {
+  override predicate row(string row) {
+    row =
+      [
+        "kotlin.jvm.internal;ArrayIteratorKt;false;iterator;(Object[]);;Argument[0].ArrayElement;ReturnValue.Element;value"
+      ]
```
atorralba, line 11:
```suggestion
    row = "kotlin.jvm.internal;ArrayIteratorKt;false;iterator;(Object[]);;Argument[0].ArrayElement;ReturnValue.Element;value"
```

atorralba:
The change in the dbscheme is causing an error. Do we need an upgrade script?

```
Files target/temp/db_upgrade/java/testdb/semmlecode.dbscheme and ql/java/ql/lib/config/semmlecode.dbscheme differ
```

atorralba (approved):
Looks plausible to me.

PR #9123, opened 2022-05-11
Kotlin: fix cases where type variables were used out of scope

smowton:
* Substitute type parameters when creating a SAM class instance that satisfies a generic instance
* In the process, fix source-locations of SAM class instances to indicate the instantiation site, not the interface being satisfied
* Fix generic arg extraction when Function23+ get rewritten to FunctionN in the JVM lowering
* Fix the type-parameter-out-of-scope test to detect these situations in future

~This is temporarily based on https://github.com/github/codeql/pull/9122 -- the real diff is https://github.com/smowton/codeql/compare/smowton/admin/update-kotlin...smowton:smowton/fix/type-variable-in-scope-consistency~

### DIFF
--- a/java/kotlin-extractor/src/main/kotlin/KotlinUsesExtractor.kt
+++ b/java/kotlin-extractor/src/main/kotlin/KotlinUsesExtractor.kt
@@ -215,6 +215,17 @@ open class KotlinUsesExtractor(
     fun makeTypeGenericSubstitutionMap(c: IrClass, argsIncludingOuterClasses: List<IrTypeArgument>) =
         getTypeParametersInScope(c).map({ it.symbol }).zip(argsIncludingOuterClasses.map { it.withQuestionMark(true) }).toMap()
 
+    fun makeGenericSubstitutionFunction(c: IrClass, argsIncludingOuterClasses: List<IrTypeArgument>) =
+        makeTypeGenericSubstitutionMap(c, argsIncludingOuterClasses).let {
+            { x: IrType, useContext: TypeContext, pluginContext: IrPluginContext ->
+                x.substituteTypeAndArguments(
+                    it,
+                    useContext,
+                    pluginContext
+                )
+            }
+        }
+
     // The Kotlin compiler internal representation of Outer<A, B>.Inner<C, D>.InnerInner<E, F>.someFunction<G, H>.LocalClass<I, J> is LocalClass<I, J, G, H, E, F, C, D, A, B>. This function returns [A, B, C, D, E, F, G, H, I, J].
     fun orderTypeArgsLeftToRight(c: IrClass, argsIncludingOuterClasses: List<IrTypeArgument>?): List<IrTypeArgument>? {
         if(argsIncludingOuterClasses.isNullOrEmpty())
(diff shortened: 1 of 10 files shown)


java/kotlin-extractor/src/main/kotlin/KotlinUsesExtractor.kt
```diff
@@ -212,10 +253,12 @@ open class KotlinUsesExtractor(
         val extractClass = substituteClass ?: c
 
         // `KFunction1<T1,T2>` is substituted by `KFunction<T>`. The last type argument is the return type.
+        // Similarly Function23 and above get replaced by kotlin.jvm.functions.FunctionN with only one type arg, the result type.
         // References to SomeGeneric<T1, T2, ...> where SomeGeneric is declared SomeGeneric<T1, T2, ...> are extracted
         // as if they were references to the unbound type SomeGeneric.
         val extractedTypeArgs = when {
-            c.symbol.isKFunction() && typeArgs != null && typeArgs.isNotEmpty() -> listOf(typeArgs.last())
+            extractClass.symbol.isKFunction() && typeArgs != null && typeArgs.isNotEmpty() -> listOf(typeArgs.last())
+            extractClass.fqNameWhenAvailable == FqName("kotlin.jvm.functions.FunctionN") && typeArgs != null && typeArgs.isNotEmpty() -> listOf(typeArgs.last())
```
tamasvajk, line 261: Is there a test case for this? How come we didn't see this issue before? We have tests that cover big-arity functions.

tamasvajk:
Are the kotlin CI tests running on this PR? If not, I think it would make sense to open the PR still in codeql-kotlin.

java/kotlin-extractor/src/main/kotlin/KotlinUsesExtractor.kt
smowton: I can add one -- I think we simply didn't notice or care that the type extracted was `FunctionN<FirstArgumentType>` not `FunctionN<ReturnType>`. It's probably not a critical bug but rather a weird surprise waiting for someone down the line.

smowton:
@tamasvajk it appears yes they are-- in recent runs of the semmle-code job `Java Language Tests Linux` we see lines like

```
[2/8] [991/1179 comp 4.3s eval 206ms] PASSED /home/runner/work/semmle-code/semmle-code/ql/java/ql/test/kotlin/library-tests/type_equivalences/type_equivalences.ql
```

I believe our integration tests are not running however: I'll make a PR to fix this later today.

smowton:
Ah no I take it back, the job https://github.com/github/semmle-code/runs/6414176878 for example runs the Kotlin integration tests. To check: do we have the right file filters selecting when the integration tests run, and do we need a single-language tests job for Kotlin?

smowton:
@tamasvajk added a test -- if you revert the commit "Fix extracted type arguments of kotlin.jvm.functions.FunctionN" then you'll see the FunctionN type arguments change improperly:

```
-| test.kt:4:5:4:138 | f1 | FunctionN<String> | String |
-| test.kt:5:5:5:134 | f2 | FunctionN<T1> | T1 |
-| test.kt:6:5:6:110 | f3 | FunctionN<T3> | T3 |
+| test.kt:4:5:4:138 | f1 | FunctionN<String> | Integer |
+| test.kt:5:5:5:134 | f2 | FunctionN<T1> | Integer |
+| test.kt:6:5:6:110 | f3 | FunctionN<T3> | T2 |
```

smowton:
Requires semmle-code PR https://github.com/github/semmle-code/pull/42569

igfoo (approved):

smowton:
One test failure is expected (taken care of by the corresponding internal PR); one is an unrelated failure also affecting `main`.

PR #9126, opened 2022-05-12
Python: Fully disallow `API::moduleImport` of module with dots

RasmusWL:
Inspired by discussion about this for MaD in https://github.com/github/codeql/pull/8883#discussion_r865858084

(and I actually thought this was already the case :grimacing:)

### DIFF
--- a/python/ql/lib/change-notes/2022-05-12-moduleimport-disallow-dots.md
+++ b/python/ql/lib/change-notes/2022-05-12-moduleimport-disallow-dots.md
@@ -1,0 +1,4 @@
+---
+category: breaking
+---
+`API::moduleImport` no longer has any results for dotted names, such as `API::moduleImport("foo.bar")`. Using `API::moduleImport("foo.bar").getMember("baz").getACall()` previously worked if the Python code was `from foo.bar import baz; baz()`, but not if the code was `import foo.bar; foo.bar.baz()` -- we are making this change to ensure the approach that can handle all cases is always used.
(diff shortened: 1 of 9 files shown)


tausbn (approved):

python/ql/lib/change-notes/2022-05-12-moduleimport-disallow-dots.md
```diff
@@ -0,0 +1,4 @@
+---
+category: fix
+---
+`API::moduleImport` no longer has any results for dotted names, such as `API::moduleImport("foo.bar")`. This previously did not yield usable results, so should not break any existing code.
```
tausbn, line 4: Thought it in fact _did_ yield usable results, which is why we're now disallowing it. (In addition to already stating it in the QLDoc.)

asgerf: I'm also a bit confused by this formulation. You mean it had results, but those API nodes were not connected to the graph in the right way?

RasmusWL: > I'm also a bit confused by this formulation. You mean it had results, but those API nodes were not connected to the graph in the right way?

Yes. That was my point, it had results, but you couldn't really use them for anything. Happy to accept a suggestion to the wording if either of you can make it more clear.

tausbn: The failing tests would seem to disagree with them not being usable for anything. :wink: 

Perhaps we should instead emphasise the real reason why we don't allow `moduleImport("foo.bar.baz")`, that being that it's _less general_ and may miss out on some instances. E.g. it will match

```python
import foo.bar.baz
```

but not 

```python
from foo import bar
bar.baz
```

RasmusWL: yes, I was actually just working on this (without commenting here :flushed:).

tausbn (approved):
Looks good to me! Thanks for making the necessary changes. :slightly_smiling_face:

PR #9127, opened 2022-05-12
Java: Sensitive Info Log query improvements

atorralba:
Adds two changes to the query `java/sensitive-log`:
* Stop considering usernames sensitive info.
* Require variables to be `static` (in addition to `final`) to be considered constants.

### DIFF
--- a/java/ql/lib/semmle/code/java/security/SensitiveLoggingQuery.qll
+++ b/java/ql/lib/semmle/code/java/security/SensitiveLoggingQuery.qll
@@ -6,12 +6,12 @@ import semmle.code.java.dataflow.TaintTracking
 import semmle.code.java.security.SensitiveActions
 import DataFlow
 
-/** A variable that may hold sensitive information, judging by its name. * */
+/** A variable that may hold sensitive information, judging by its name. */
 class CredentialExpr extends Expr {
   CredentialExpr() {
     exists(Variable v | this = v.getAnAccess() |
-      v.getName().regexpMatch([getCommonSensitiveInfoRegex(), "(?i).*(username).*"]) and
-      not v.isFinal()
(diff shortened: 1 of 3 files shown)


java/ql/lib/semmle/code/java/security/SensitiveLoggingQuery.qll
```diff
@@ -4,14 +4,15 @@ import java
...
 
-/** A variable that may hold sensitive information, judging by its name. * */
+/** A variable that may hold sensitive information, judging by its name. */
 class CredentialExpr extends Expr {
   CredentialExpr() {
     exists(Variable v | this = v.getAnAccess() |
-      v.getName().regexpMatch([getCommonSensitiveInfoRegex(), "(?i).*(username).*"]) and
-      not v.isFinal()
+      v.getName().regexpMatch(getCommonSensitiveInfoRegex()) and
+      not (v.isFinal() and v.isStatic())
```
smowton, line 15: Would `... and not this instanceof CompileTimeConstantExpr` get what you want? That already includes clauses

```
      // Access to a final variable initialized by a compile-time constant.
      exists(Variable v | this = v.getAnAccess() |
        v.isFinal() and
        v.getInitializer().isCompileTimeConstant()
      )
      or
      this instanceof LiveLiteral
```

atorralba: Oh, good catch! I think we still need the `LiveLiteral` sanitizer though, because there are other elements with awful names being generated by the Compose compiler plugin that don't fit in the `LiveLiteral` class but flow through it.

But this is still cleaner for the "not a constant" check, so I'll add it, thanks.

smowton (approved):

PR #9129, opened 2022-05-12
Java: Add Expr::getUnderlyingExpr predicate

atorralba:
Adds a predicate to look through casts and not-null expressions to obtain the underlying expression. Uses this new predicate in a couple of queries that need it to work properly in Kotlin codebases.


smowton:
Add a test case for the android changes?

atorralba:
Done, although we need the extractor to expand the classpath from `options` files.

aschackmull (approved):

PR #9130, opened 2022-05-12
C++: Handle C++17 if and switch initializers

jketema:
Relatively straight-forward, but requires DB scheme update. The downgrade script leaves behind the initializer, although it no longer has a parent. I don't know whether this is sufficient, or whether an attempt should be made to remove the whole initializer.


cpp/ql/lib/semmle/code/cpp/stmts/Stmt.qll
```diff
@@ -213,6 +213,26 @@ class ConditionalStmt extends ControlStructure, TConditionalStmt { }
 class IfStmt extends ConditionalStmt, @stmt_if {
   override string getAPrimaryQlClass() { result = "IfStmt" }
 
+  /**
+   * Gets the initialization statement of this 'if' statement.
+   *
+   * For example, for
+   * ```
+   * if (x = y; b) { f(); }
```
MathiasVP, line 221: Should we also have an example with a declaration like this?
```cpp
if(int x = y; b) { f(); }
```

cpp/ql/lib/semmle/code/cpp/stmts/Stmt.qll
```diff
@@ -213,6 +213,26 @@ class ConditionalStmt extends ControlStructure, TConditionalStmt { }
 class IfStmt extends ConditionalStmt, @stmt_if {
   override string getAPrimaryQlClass() { result = "IfStmt" }
 
+  /**
+   * Gets the initialization statement of this 'if' statement.
```
MathiasVP, line 217: This makes it clear that there might not be any initialization statement. (I know you've written it further down in the qldoc, but this phrasing matches our style guide.)
```suggestion
   * Gets the initialization statement of this 'if' statement, if any.
```

cpp/ql/lib/semmle/code/cpp/stmts/Stmt.qll
```diff
@@ -299,6 +319,28 @@ class IfStmt extends ConditionalStmt, @stmt_if {
 class ConstexprIfStmt extends ConditionalStmt, @stmt_constexpr_if {
   override string getAPrimaryQlClass() { result = "ConstexprIfStmt" }
 
+  /**
+   * Gets the initialization statement of this 'constexpr if' statement.
```
MathiasVP, line 323: Same here.
```suggestion
   * Gets the initialization statement of this 'constexpr if' statement, if any.
```

cpp/ql/lib/semmle/code/cpp/stmts/Stmt.qll
```diff
@@ -299,6 +319,28 @@ class IfStmt extends ConditionalStmt, @stmt_if {
...
+   *
+   * For example, for
+   * ```
+   * if constexpr (x = y; b) { f(); }
```
MathiasVP, line 327: Same here: Maybe we should have an example with `if constexpr (int x = y; b) { f(); }` or something like that.

cpp/ql/lib/semmle/code/cpp/stmts/Stmt.qll
```diff
@@ -1470,6 +1512,28 @@ class DefaultCase extends SwitchCase {
 class SwitchStmt extends ConditionalStmt, @stmt_switch {
   override string getAPrimaryQlClass() { result = "SwitchStmt" }
 
+  /**
+   * Gets the initialization statement of this 'switch' statement.
```
MathiasVP, line 1516:
```suggestion
   * Gets the initialization statement of this 'switch' statement, if any.
```

MathiasVP (commented):
The code LGTM! I have a couple of not-super-important QLDoc comments.

> The downgrade script leaves behind the initializer, although it no longer has a parent. I don't know whether this is sufficient, or whether an attempt should be made to remove the whole initializer.

I don't think it's a problem to have a parent-less element like that. That probably happens quite often due to extraction problems anyway, so I think our code is well battle-tested against these kinds of malformed databases already.

cpp/ql/lib/semmle/code/cpp/stmts/Stmt.qll
jketema: Tempted to make that the only example, as that's the common use-case, while `if (x = y; b)` is a bit esoteric. What do you think?

MathiasVP: Yeah, I think that would be fine as well.

jketema: Done.

PR #9132, opened 2022-05-12
QL for QL: generalise non-US spelling query

nickrolfe:
1. Catch common misspellings as well.
2. Also check names of classes, predicates, etc.

The old query matched against the entire QLDoc comment string, but now I break each comment string up into individual words (and do the same for node names). I was a bit worried about string pool pollution and performance, but they seem OK. We're talking about tens of thousands of strings for a db containing all of `github/codeql`, which is not that big a number.


nickrolfe:
If @github/codeql-ql-for-ql-reviewers think this is ok, I'll add followup commits to fix some/many/all of the typos it finds.

erik-krogh (commented):
> If https://github.com/orgs/github/teams/codeql-ql-for-ql-reviewers think this is ok, I'll add followup commits to fix some/many/all of the typos it finds. 

This looks good :+1:  

And there are plenty of typos to fix :see_no_evil:

tausbn:
_Finally_ I won't be able to misspell "primordial" in my code. :muscle:

nickrolfe:
I've pushed fixes to all the spelling errors in QLDoc comments, and a couple of small refinements to the QL-for-QL query. I'm inclined to leave the remaining misspelled predicate and class names as they are for now.

jketema (approved):
C++ 👍 

@geoffw0 You might want to have a look at the XXE transformer misspellings flagged up here. I think those are coming from the query you're currently working on?

tausbn (approved):
Python changes LGTM. :+1:

nickrolfe:
@geoffw0 - the updated query also flags a few class names in [XXE.ql](https://github.com/github/codeql/blob/main/cpp/ql/src/Security/CWE/CWE-611/XXE.ql) that misspell 'transformer' as 'tranformer'.

geoffw0:
> @geoffw0 - the updated query also flags a few class names in [XXE.ql](https://github.com/github/codeql/blob/main/cpp/ql/src/Security/CWE/CWE-611/XXE.ql) that misspell 'transformer' as 'tranformer'.

Oh I see, there are some issues flagged but not fixed here.  I can fix them in an upcoming PR?

Also: how did I not spot these myself...

nickrolfe:
> Oh I see, there are some issues flagged but not fixed here. I can fix them in an upcoming PR?

Yeah, correcting spelling in comments is low-risk, so I just fixed them, but changing class and predicate names is a bit more involved, as I have to figure out whether I need to deprecate the old name and write a change-note.

So I've left that as a followup exercise, and I'd be more than happy for you to deal with those particular misspellings.

geoffw0:
> I'd be more than happy for you to deal with those particular misspellings.

Will do.

hvitved (approved):
C# / Ruby 👍

PR #9133, opened 2022-05-12
Devcontainer: Install test dependencies

tausbn:
These _should_ get installed automatically if missing, but in my
experience this can be a bit flaky. Installing them here should make
this a bit more robust.


shati-patel (approved):
Yep, it can be easy to miss the installation pop-up 🙈 

Thanks for adding this, LGTM!

PR #9137, opened 2022-05-12
Ruby: Add getAPrimaryQlClass to CfgNodes classes

hmac:
I've excluded the `ChildMapping` classes because
- they're not often directly used
- they cause some test fixtures to bloat
- they can subclass AST classes which have a `final` definition of `getAPrimaryQlClass`


nickrolfe:
> I've excluded the `ChildMapping` classes

Is that also why you haven't added the method to the base `CfgNode` class?

nickrolfe (approved):

PR #9138, opened 2022-05-12
Ruby: Make StringArrayInclusion more sensitive

hmac:
We now recognise the following pattern as a barrier guard for `x`:

```ruby
values = ["foo", "bar"]

if values.include? x
  sink x
end
```

I've used SSA to track this, which I think ensures that we have universal flow from the RHS of `values = ...` to the read in `values.include?`.

### DIFF
--- a/ruby/ql/test/library-tests/ast/Ast.expected
+++ b/ruby/ql/test/library-tests/ast/Ast.expected
@@ -1556,6 +1556,35 @@ constants/constants.rb:
 #   73|       getAnOperand/getLeftOperand: [ClassVariableAccess] @@fourty_six
 #   73|       getAnOperand/getRightOperand: [ConstantReadAccess] FOURTY_SIX
 #   73|         getScopeExpr: [ConstantReadAccess] Mod3
+#   78|   getStmt: [AssignExpr] ... = ...
+#   78|     getAnOperand/getLeftOperand: [LocalVariableAccess] a
+#   78|     getAnOperand/getRightOperand: [ArrayLiteral] [...]
+#   78|       getElement: [IntegerLiteral] 1
+#   78|       getElement: [IntegerLiteral] 2
+#   78|       getElement: [IntegerLiteral] 3
+#   79|   getStmt: [AssignExpr] ... = ...
+#   79|     getAnOperand/getLeftOperand: [ConstantAssignment] A
+#   79|     getAnOperand/getRightOperand: [ArrayLiteral] [...]
+#   79|       getElement: [IntegerLiteral] 1
+#   79|       getElement: [IntegerLiteral] 2
+#   79|       getElement: [IntegerLiteral] 3
+#   80|   getStmt: [AssignExpr] ... = ...
+#   80|     getAnOperand/getLeftOperand: [ConstantAssignment] B
+#   80|     getAnOperand/getRightOperand: [LocalVariableAccess] a
+#   81|   getStmt: [AssignExpr] ... = ...
+#   81|     getAnOperand/getLeftOperand: [ConstantAssignment] C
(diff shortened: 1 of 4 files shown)


ruby/ql/lib/codeql/ruby/dataflow/BarrierGuards.qll
```diff
@@ -56,19 +58,29 @@ class StringConstCompare extends DataFlow::BarrierGuard,
...
 class StringConstArrayInclusionCall extends DataFlow::BarrierGuard,
   CfgNodes::ExprNodes::MethodCallCfgNode {
   private CfgNode checkedNode;
 
   StringConstArrayInclusionCall() {
-    exists(ArrayLiteral aLit |
-      this.getExpr().getMethodName() = "include?" and
-      [this.getExpr().getReceiver(), this.getExpr().getReceiver().(ConstantReadAccess).getValue()] =
-        aLit
+    this.getMethodName() = "include?" and
```
aibaars, line 66: I think we could use the InclusionTests library here, that also covers inclusion tests writing like `A.indexOf(B) != nil`

hvitved (commented):
Looks good. Alternatively, we could add something like this to `internal/Constants.qll`, and then reference that:
```ql
  predicate isArray(ExprCfgNode e, ArrayLiteralCfgNode a) {
    e = a
    or
    isArrayExpr(e.getExpr().(ConstantReadAccess).getValue(), a)
    or
    isArray(getSource(e), a)
  }

  predicate isArrayExpr(Expr e, ArrayLiteralCfgNode a) {
    isArrayExpr(e.(ConstantReadAccess).getValue(), a)
    or
    forex(ExprCfgNode n | n = e.getAControlFlowNode() | isArray(n, a))
  }
```

ruby/ql/lib/codeql/ruby/ast/internal/Constant.qll
```diff
@@ -509,3 +509,23 @@ private module Cached {
...
+/**
+ * Holds if `e` is an array constructed from an array literal.
+ * Example:
+ * ```rb
+ * [1, 2, 3]
+ * C = [1, 2, 3]; C
+ * x = [1, 2, 3]; x
+ * ```
+ */
+predicate isArrayConstant(ExprCfgNode e, ArrayLiteralCfgNode arr) {
```
hvitved, line 522: This is a bit less general than my proposal, which would handle (probably not likely in practice) cases like
```
x = [1, 2, 3]; C = x; y = C; y
```

aschackmull:
I'm about to deprecate the `BarrierGuard` class, so I would appreciate if you could hold this PR until after https://github.com/github/codeql/pull/9618 has been merged.

ruby/ql/lib/codeql/ruby/ast/internal/Constant.qll
nickrolfe: Tom's example doesn't make sense to me either. I *think* he meant to write this:

```codeql
predicate isArray(ExprCfgNode e, ArrayLiteralCfgNode a) {
  e = a
  or
  isArrayExpr(e.getExpr(), a)
  or
  isArray(getSource(e), a)
}

predicate isArrayExpr(Expr e, ArrayLiteralCfgNode a) {
  isArrayExpr(e.(ConstantReadAccess).getValue(), a)
  or
  forex(ExprCfgNode n | n = e.getAControlFlowNode() | isArray(n, a))
}
```

If you're not feeling confident with that change, I still think it's worth merging your version of `isArrayConstant`, but perhaps with a TODO comment including the example Tom gave with the series of assignments your version won't handle.

nickrolfe (approved):
Looks great, assuming the DCA run doesn't find anything.

PR #9139, opened 2022-05-12
Kotlin: Add failing test for missing field flow

atorralba:
The `StringLiteral` initializer of a `KtInitializerAssignExpr` seems to not be flowing to the field it's initializing nor its accesses.


atorralba:
Superseded by https://github.com/github/codeql/pull/9155.

PR #9145, opened 2022-05-13
Kotlin: Respect `override` modifier in useless parameter query

tamasvajk:


atorralba (approved):
LGTM. OOC, why isn't this covered by `not this.(Method).overridesOrInstantiates(_)` in the charpred?

tamasvajk:
> LGTM. OOC, why isn't this covered by `not this.(Method).overridesOrInstantiates(_)` in the charpred?

Good question, I haven't investigated this now, I just added the modifier check next to the java equivalent: `this.getAnAnnotation() instanceof OverrideAnnotation`. Also, I added two commits to show that the change does fix the issue. I vaguely remember that the `overridesOrInstantiates` check the signature of candidate methods, and for generics it doesn't always match.

smowton:
Does the signature mismatching inhibit dataflow from spotting a possible callee of an overridden method? If so we should fix that

tamasvajk:
@smowton I've added an internal issue for this.

atorralba (approved):
Approving under the assumption that we'll investigate and, if necessary, handle the dataflow issue in a separate PR.

atorralba (approved):

PR #9147, opened 2022-05-13
Swift: publish C++ generated code as artifacts

redsun82:


AlexDenisov (commented):
I don't think it provides much value as I'd prefer seeing the actual code locally when I change things, but I'm not opposed to this change if it is helpful to anyone.

redsun82:
> I don't think it provides much value as I'd prefer seeing the actual code locally when I change things, but I'm not opposed to this change if it is helpful to anyone.

as stated on the other PR, this is not intedended for active developers on a particular change, but for people reviewing a PR, or browsing code through github, or even just needing to post a link to a particular generation of the file on slack

AlexDenisov (dismissed):

AlexDenisov (approved):

PR #9149, opened 2022-05-13
Kotlin: Exclude operands of `NotNullExpr` from NullMaybe query

tamasvajk:


java/ql/src/Likely Bugs/Nullness/NullMaybe.ql
```diff
@@ -21,6 +21,8 @@ from VarAccess access, SsaSourceVariable var, string msg, Expr reason
 where
   nullDeref(var, access, msg, reason) and
   // Exclude definite nulls here, as these are covered by `NullAlways.ql`.
-  not alwaysNullDeref(var, access)
+  not alwaysNullDeref(var, access) and
+  // Exclude operands of `!!` in Kotlin, as `!!` means explicit opt out by the user.
+  not access.getParent() instanceof NotNullExpr
```
tamasvajk, line 26: Another option would be to exclude all kotlin source code from this check.

smowton: That seems sensible to me, since the language is enforcing proper checks for us.

atorralba (approved):

atorralba (approved):

PR #9152, opened 2022-05-13
Kotlin: Fix extraction of reflective call generated by Parcelize

tamasvajk:
This PR fixes the extraction of the `java` property call in `C::class.java` generated by the Parcelize plugin. Generally, these calls are properly associated to `kotlin.jvm.JvmClassMappingKt::java`, and therefore need no special handling. However, with the Parcelize plugin this is not the case.

I'm planning on merging this fix without integration test. It feels way more effort to add the test than the benefit.


tamasvajk:
> Looks reasonable -- can we test it?

We could, but 
> I'm planning on merging this fix without integration test. It feels way more effort to add the test than the benefit.

PR #9153, opened 2022-05-13
Kotlin: Unify loop `break`/`continue` statement handling between java and kotlin

tamasvajk:
This PR removes the kotlin relations for loops, break and continue statements and instead uses the java equivalents.

The PR removes DB structures, so it requires new stats generation and upgrade folder.


java/ql/lib/semmle/code/java/Statement.qll
```diff
@@ -887,27 +888,3 @@ class SuperConstructorInvocationStmt extends Stmt, ConstructorCall, @superconstr
...
-
-  /** Gets the target loop statement of this `break`. */
-  KtLoopStmt getLoopStmt() { result = loop }
-}
-
-/** A Kotlin `break` statement. */
-class KtBreakStmt extends BreakStmt, KtBreakContinueStmt { }
-
-/** A Kotlin `continue` statement. */
-class KtContinueStmt extends ContinueStmt, KtBreakContinueStmt { }
```
igfoo, line 913: The changelog documented these being added, so I think we should have a changenote entry about them being removed too.

java/ql/lib/upgrades/b9225587bc0a643ae484ec215b9a6f19d17d0fc2/upgrade.properties
```diff
@@ -0,0 +1,3 @@
+description: Remove unneeded break/continue structures for Kotlin
+compatibility: backwards
```
igfoo, line 2: There should be a `delete` line for the removed relation (see https://github.com/github/codeql/blob/main/docs/prepare-db-upgrade.md)

aschackmull: > I also changed the compatibility to `partial`

Why? `backwards` sound right to me.

tamasvajk: You're probably right. Removing the relation and the corresponding QL classes would not produce worse results than before. The results were already wrong as CFG didn't take into account any of these break/continue targets.

java/ql/lib/semmle/code/java/Statement.qll
```diff
@@ -29,8 +29,7 @@ class Stmt extends StmtParent, ExprParent, @stmt {
    */
   Stmt getEnclosingStmt() {
     result = this.getParent() or
-    result = this.getParent().(SwitchExpr).getEnclosingStmt() or
-    result = this.getParent().(WhenExpr).getEnclosingStmt()
```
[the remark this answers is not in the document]
igfoo: Yes, I agree we need that change, but having a separate PR for it would be cleaner.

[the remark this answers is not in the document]
tamasvajk: Fixed in https://github.com/github/codeql/pull/9647

igfoo (approved):

PR #9157, opened 2022-05-13
Ruby/Python: Add a `BlockMode` concept for `CryptographicOperations`

alexrford:
Cryptographic operations that use a block cipher such as AES often require a block mode of operation. A common case for this is in encrypting data of arbitrary length using a block cipher. These block modes can have inherent weaknesses that make them less suited to cryptographic purposes - in particular the electronic codebook (ECB) scheme can reveal patterns in data even if the underlying block cipher is secure.

The existing `rb/weak-cryptographic-algorithm` query checks for uses of ECB, but the message reported to the user was inaccurate, as in cases such as `AES-ECB`, it would report `AES` as being insecure rather than the `ECB` block mode. This is corrected in this PR by adding a new `BlockMode` concept with an `isWeak` member predicate and adding a `BlockMode getBlockMode()` predicate to `CryptographicOperation` alongside the existing `CryptographicAlgorithm getAlgorithm()` predicate.

I've made a similar change to the python version of this query. This mostly changes the `cryptodome` and `cryptography` libraries. I'm happy to remove the Python changes, or to move them out to a separate PR if that's more convenient than having everything in this PR.

Whilst implementing this, I noticed that the Ruby and Python `CryptographicOperation`s are conceptually different. Ruby uses it to mean an instantiation of a cryptographic cipher or a use of that cipher (e.g. in encrypting some data), whereas it seems like Python uses it to mean only a use of such a cipher. I think that the Python version of this makes more sense, as the Ruby version seems like it could be unnecessarily noisy.

On another note, `CryptographicOperation` encompasses quite a lot of different operations, including many where a block mode is not relevant. This means that many implementations are just `none()`. This might suggest that `CryptographicOperation` should be subclassed for different cases (e.g. encrypting a stream of data, cryptographic hashing, etc).


erik-krogh (approved):

ruby/ql/test/library-tests/concepts/CryptographicOperation.ql
```diff
@@ -0,0 +1,34 @@
...
+
+  override string getARelevantTag() {
+    result in [
+    "CryptographicOperation", "CryptographicOperationInput", "CryptographicOperationAlgorithm",
+    "CryptographicOperationBlockMode"
+    ]
+  }
+
+  override predicate hasActualResult(Location location, string element, string tag, string value) {
+    exists(location.getFile().getRelativePath()) and
```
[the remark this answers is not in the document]
alexrford: Copied this from another test but it's not relevant here. Deleted.

alexrford:
@hmac thanks for the review. I noticed a regression in the DCA results where I was making an incorrect assumption that ciphers instantiated with `OpenSSL::Cipher.new` always used `CBC`, which was not the case when the argument to `new` included an explicit block mode (e.g. `'bf-ecb'`) or when the argument specified a stream cipher. I've fixed this in https://github.com/github/codeql/pull/9157/commits/f92782d4e783d4f0acc9097c15eb00b1e9fe0d09

> The Ruby changes look good to me! I wonder, could we directly share the `CryptographicOperation` concept between Python and Ruby by moving it to `ConceptsShared.qll`?

Yeah, I think this makes sense. I held off doing this mostly because Javascript implements `weak-cryptographic-algorithm` differently to Python and Ruby (the Ruby version of this query was based on the Python version), so the `CryptographicOperation` concept doesn't currently exist in JS. Naively, I could imagine this concept being useful for JS in some circumstances as well, though I'm not sure if we want to add concepts that don't have any concrete implementations.

RasmusWL:
> > The Ruby changes look good to me! I wonder, could we directly share the `CryptographicOperation` concept between Python and Ruby by moving it to `ConceptsShared.qll`?
> 
> Yeah, I think this makes sense. I held off doing this mostly because Javascript implements `weak-cryptographic-algorithm` differently to Python and Ruby (the Ruby version of this query was based on the Python version), so the `CryptographicOperation` concept doesn't currently exist in JS. Naively, I could imagine this concept being useful for JS in some circumstances as well, though I'm not sure if we want to add concepts that don't have any concrete implementations.

:wave: just following a bit along on the sidelines, since I should be reviewing this PR soon. It's fine to share the concept between Ruby and Python initially, and let JS adopt it later on.

alexrford:
Cool, I will have a go at moving this into `ConceptsShared.qll` then.

python/ql/src/change-notes/2022-05-16-broken-crypto-block-mode.md
```diff
@@ -0,0 +1,4 @@
+---
+category: minorAnalysis
+---
+The query "Use of a broken or weak cryptographic algorithm" (`py/weak-cryptographic-algorithm`) now report if a cryptographic operation is potentially insecure due to use of a weak block mode.
```
RasmusWL, line 4: These should always start with `*` -- for example see [this fixup](https://github.com/github/codeql/pull/9141/commits/43265c41334f6af08dfcaa50d2062c0ba3b9356c)

```suggestion
* The query "Use of a broken or weak cryptographic algorithm" (`py/weak-cryptographic-algorithm`) now report if a cryptographic operation is potentially insecure due to use of a weak block mode.
```

python/ql/lib/semmle/python/frameworks/Cryptography.qll
```diff
@@ -182,7 +193,12 @@ private module CryptographyModel {
               .getACall() and
         algorithmClassRef(algorithmName).getReturn().getAUse() in [
             call.getArg(0), call.getArgByName("algorithm")
-          ]
+          ] and
+        exists(DataFlow::Node modeArg | modeArg in [call.getArg(1), call.getArgByName("mode")] |
+          modeArg = modeClassRef(modeName).getReturn().getAUse()
+          or
+          modeArg.asExpr() instanceof None and modeName = "<none>"
```
RasmusWL, line 200: In general I don't find QL very good at tracking multiple properties of instances, so I understand why you wrote the code as you did. However, if we for some reason don't manage to match either part of the disjunct for the `modeName`, we also loose everything about the algorithmName :disappointed: 

I see that some of the [official examples](https://cryptography.io/en/latest/hazmat/primitives/symmetric-encryption/#cryptography.hazmat.primitives.ciphers.algorithms.ChaCha20) use `mode=None`, so good catch on handling that as well :+1:

I would feel better with a defensive version like one below:

```suggestion
          if modeArg = modeClassRef(modeName).getReturn().getAUse()
          then any()
          else modeName = "<None or unknown>"
```

it does feel slightly hacky to have this `<None or unknown>` instead of just not having a result... I guess this could be fixed by having two sets cipher instance tracking (one for algorithm name, one for mode) -- but then you might end up with cross-talk between unrelated algorithm and mode pairs :thinking:

_note: I haven't actually tested this code, but I hope it "should just work" :crossed_fingers:_

RasmusWL (changes requested):
Nice :tada: Thanks for porting this to Python as well :muscle: A few minor things, otherwise looks good to me :+1:

ruby/ql/src/change-notes/2022-05-16-broken-crypto-message.md
```diff
@@ -0,0 +1,4 @@
+---
+category: minorAnalysis
+---
+The query "Use of a broken or weak cryptographic algorithm" (`rb/weak-cryptographic-algorithm`) now report if a cryptographic operation is potentially insecure due to use of a weak block mode.
```
RasmusWL, line 4:
```suggestion
* The query "Use of a broken or weak cryptographic algorithm" (`rb/weak-cryptographic-algorithm`) now report if a cryptographic operation is potentially insecure due to use of a weak block mode.
```

python/ql/lib/semmle/python/frameworks/Cryptography.qll
alexrford: Thanks, that's a good point. I think I prefer this fallback case over trying to track the two things separately for the ease of keeping the algorithm and mode bundled together. The slight hackiness seems acceptable as the `modeName` itself shouldn't be exposed to users of the `CryptographyGenericCipherOperation` class as this is exposed via the `Cryptography::BlockMode getBlockMode()` predicate, which is restricted to only known block modes.

alexrford:
I've run into a somewhat interesting issue in sharing the `CryptographicOperation` concept. In the Ruby version of this, we currently have an `isWeak()` member predicate. This was previously used in the broken crypto query, but has been replaced by (effectively) `op.getAlgorithm().isWeak() or op.getBlockMode().isWeak()`, which is the same approach used in the Python version of the query in this PR.

This predicate is now unused in Ruby. I think it should probably be deprecated and eventually removed - whilst an operation may be insecure in other ways (e.g. insufficient key size), I think we'd still usually want to expose to a user _why_ the operation is weak - something that a simple predicate on the operation alone can't communicate.

I've approached this in the current PR by:
 - Adding the `CryptographicOperation#isWeak()` predicate to the shared concept
 - Adding a default implementation to the `abstract class CryptographicOperation::Range`, which lets us avoid implementing this predicate for Python
 - Immediately deprecating this predicate

It doesn't feel great to add a predicate which is then immediately deprecated, but it seemed like the least bad solution which complied with the deprecation policy.

RasmusWL (approved):
I just double checked my suggestion in https://github.com/github/codeql/pull/9157#discussion_r874951397, and it was not working as intended, so I just fixed that up. (and happy to have done that before the PR was merged :relaxed:)

But overall looks good to me now :+1: (not sure whether this is blocked on anything at the moment?)

alexrford:
> I just double checked my suggestion in [#9157 (comment)](https://github.com/github/codeql/pull/9157#discussion_r874951397), and it was not working as intended, so I just fixed that up. (and happy to have done that before the PR was merged ☺️)
> 
> But overall looks good to me now 👍 (not sure whether this is blocked on anything at the moment?)

Thanks for the fix, I see the problem now after reading through the commit message. It's a shame that the `any()` version has some spurious tuples as I find myself wanting to express "either something that satisfies the given predicate, or otherwise a default value" fairly often and this looked like a neat way to do it without any repetition.

PR #9172, opened 2022-05-16
Swift: replace `getCanonicalPointer` with `std::variant`

redsun82:
This turned out easier than expected previously. `llvm::PointerUnion`
was also considered, which would have less memory footprint, but it
would require more effort as it is lacking the same implicit conversions
and operators that `std::variant` provides.

Also renamed `ToTag<E>` to `TrapTagOf<E>` and introduced a derived
convenience functor `TrapLabelOf<E>`.

This is a follow-up to https://github.com/github/codeql/pull/9112

### DIFF
--- a/swift/extractor/SwiftDispatcher.h
+++ b/swift/extractor/SwiftDispatcher.h
@@ -46,7 +46,7 @@ std::string inline getKindName<swift::TypeRepr, swift::TypeReprKind>(swift::Type
 
 }  // namespace detail
 
-// The main reponsibilities of the SwiftDispatcher are as follows:
+// The main responsibilities of the SwiftDispatcher are as follows:
 // * redirect specific AST node emission to a corresponding visitor (statements, expressions, etc.)
 // * storing TRAP labels for emitted AST nodes (in the TrapLabelStore) to avoid re-emission
 // Since SwiftDispatcher sees all the AST nodes, it also attaches a location to every 'locatable'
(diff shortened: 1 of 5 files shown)


swift/extractor/SwiftDispatcher.h
```diff
@@ -72,19 +72,28 @@ class SwiftDispatcher {
...
   template <typename E>
-  TrapLabel<ToTag<E>> fetchLabel(E* e) {
+  TrapLabelOf<E> fetchLabel(E* e) {
     // this is required so we avoid any recursive loop: a `fetchLabel` during the visit of `e` might
     // end up calling `fetchLabel` on `e` itself, so we want the visit of `e` to call `fetchLabel`
-    // only after having called `assignNewLabel` on `e`
-    assert(!waitingForNewLabel && "fetchLabel called before assignNewLabel");
+    // only after having called `assignNewLabel` on `e`. `index() == 0` indicates the monostate of
+    // `Store::Handle`, i.e. its default construction
+    assert(waitingForNewLabel.index() == 0 && "fetchLabel called before assignNewLabel");
```
sashabu, line 92: Optional: I think `holds_alternative<std::monostate>` would be clearer (and more resilient) than relying on the index of `monostate`?

I generally prefer the type-based accessors to the index-based accessors because they make the behaviour obvious without having to see the specific variant type (and in some cases save a fair bit of counting!) - and a side benefit of not caring about the order is that if the order changes, the user code doesn't need updating.

swift/extractor/SwiftDispatcher.h
```diff
@@ -100,12 +109,12 @@ class SwiftDispatcher {
   // it actually gets emitted to handle recursive cases such as recursive calls, or recursive type
   // declarations
   template <typename E>
-  TrapLabel<ToTag<E>> assignNewLabel(E* e) {
-    assert(waitingForNewLabel == getCanonicalPointer(e) && "assignNewLabel called on wrong entity");
-    auto label = getLabel<ToTag<E>>();
+  TrapLabelOf<E> assignNewLabel(E* e) {
+    assert(waitingForNewLabel == Store::Handle{e} && "assignNewLabel called on wrong entity");
```
sashabu, line 113: Nit: Do we need to construct a `Store::Handle` or can we just compare to `e` and rely on an implicit conversion?

swift/extractor/SwiftDispatcher.h
```diff
@@ -100,12 +109,12 @@ class SwiftDispatcher {
...
+    auto label = getLabel<TrapTagOf<E>>();
     trap.assignStar(label);
     store.insert(e, label);
-    waitingForNewLabel = nullptr;
+    waitingForNewLabel = {};
```
sashabu, line 117: Optional: As above, I'd prefer this to spell out `monostate` rather than relying on it being the first member.

sashabu (changes requested):

swift/extractor/SwiftDispatcher.h
redsun82: no, it did not work. If I get it correctly, the compiler is not able to apply the two implicit casts (say, from `VarDecl*` to `Decl*` and from `Decl*` to `Store::Handle`), so it needs a bit of a hand

swift/extractor/SwiftDispatcher.h
redsun82: on the other hand, this is the documented main use case of `std::monospace` https://en.cppreference.com/w/cpp/utility/variant/monostate but I'm ok with being more explicit.

swift/extractor/SwiftDispatcher.h
redsun82: done!

sashabu (approved):

PR #9175, opened 2022-05-16
Ruby: stop considering post-update nodes to be local source nodes

nickrolfe:


nickrolfe:
Hmm, I may have got this wrong. On the one hand, dataflow queries on `rapid7/metasploit-framework` now complete in a reasonable time, but on the other hand, this doesn't look great:
```
Tuple counts for DataFlowPrivate::Cached::isLocalSourceNode#462ff392#f#antijoin_rhs@46f66bui:
        1463003991      ~0%    {2} r1 = JOIN boundedFastTC(DataFlowPrivate::Cached::localFlowStepTypeTracker#462ff392#ff_10#higher_order_body,DataFlowPrivate::Cached::TNode#462ff392#f) WITH DataFlowPrivate::Cached::TNode#462ff392#f ON FIRST 1 OUTPUT Lhs.1, Rhs.0
                           
            369909      ~3%    {1} r2 = JOIN r1 WITH DataFlowPrivate::Cached::entrySsaDefinition#462ff392#f ON FIRST 1 OUTPUT Lhs.1
                           
         731612184  ~88927%    {1} r3 = JOIN r1 WITH DataFlowPrivate::Cached::TExprNode#462ff392#ff_1#join_rhs ON FIRST 1 OUTPUT Lhs.1
                           
         731982093  ~81105%    {1} r4 = r2 UNION r3
                               return r4
```

ruby/ql/lib/codeql/ruby/dataflow/internal/DataFlowPrivate.qll
```diff
@@ -300,14 +300,13 @@ private module Cached {
...
-    n instanceof PostUpdateNodes::ExprPostUpdateNode
-    or
     // Nodes that can't be reached from another entry definition or expression.
     not localFlowStepTypeTracker+(any(Node n0 |
         n0 instanceof ExprNode
         or
         entrySsaDefinition(n0)
-      ), n)
+      ), n) and
+    not n instanceof PostUpdateNode
```
hvitved, line 309: Actually, I would expect that this should be `n instanceof ExprNode`.

ruby/ql/src/queries/security/cwe-116/IncompleteSanitization.ql
```diff
@@ -136,7 +136,7 @@ DataFlow::CallNode getAMethodCall(StringSubstitutionCall call) {
       // post-update receiver to the receiver of another method call?
       call.isDestructive() and
       exists(DataFlow::PostUpdateNode post | post.getPreUpdateNode() = call.getReceiver() |
-        post.(DataFlow::LocalSourceNode).flowsTo(receiver)
+        DataFlow::localFlow(post, receiver)
```
hvitved, line 139: May as well just be `DataFlow::localFlow(call.getReceiver(), receiver)`.

nickrolfe: I'm not convinced that's right – we want flow from the post-update node, not the pre-update node. The test fails with that change.

ruby/ql/lib/codeql/ruby/dataflow/internal/DataFlowPrivate.qll
nickrolfe: Aha, thanks.

nickrolfe:
I still see the massive tuple count in the antijoin I posted above.

ruby/ql/src/queries/security/cwe-116/IncompleteSanitization.ql
hvitved: Local flow should really not be concerned with post-update nodes, as that is a feature of the global data flow library used only in connection with field-flow. I am surprised that the test does not pass.

ruby/ql/src/queries/security/cwe-116/IncompleteSanitization.ql
```diff
@@ -125,21 +124,16 @@ predicate removesFirstOccurence(StringSubstitutionCall sub, string str) {
...
+  // for a non-destructive string substitution, is there flow from it to the
+  // receiver of another method call?
+  not call.isDestructive() and
+  DataFlow::localFlow(call, result.getReceiver())
+  or
+  // for a destructive string substitution, is there flow from its
+  // post-update receiver to the receiver of another method call?
+  call.isDestructive() and
+  DataFlow::localFlow(call.getReceiver(), result.getReceiver()) and
+  not call = result
```
hvitved, line 136:
```suggestion
  DataFlow::localFlowStep+(call.getReceiver(), result.getReceiver())
```

hvitved:
> I still see the massive tuple count in the antijoin I posted above.

I have pushed a fix for the antijoin.

nickrolfe:
@hmac the test changes I committed aren't the only ones – the `n instanceof ExprNode` change suggested by Tom introduced more test failure, and I haven't managed to look into them yet, but my gut instinct was that something got broken.

PR #9176, opened 2022-05-16
C++: Clean up the XXE query QL.

geoffw0:
This is the promised follow-up PR to clean up the somewhat messy XXE query code, with a unifying `XmlLibrary` class defining sources and sinks.  Then split the QL for the query into four files:
 - `XXE.ql`, containing the query itself, now a straightforward dataflow query.
 - `XXE.qll`, containing the abstract classes / models used by the query.
 - `Xerces.qll`, containing the implementation details for all four supported Xerces interfaces.
 - `libxml2.qll`, containing the implementation details for the libxml2 library.

This makes the code considerably easier to read, navigate, and potentially extend.  There are a few limitations as things stand though:
 - I have not attempted to further split `Xerces.qll` into the individual interfaces.  Its easily the largest of the four files and it would be logical to break it up, but there's a fair bit of crossover between the interfaces that might make it tricky to do so.
 - I have not merged the existing `XXEFlowStateTransformer` class into `XmlLibrary`.  Again it would be logical, but the crossover between the Xerces interfaces that might make this more difficult than it seems.
 - I have not attempted to make the `XML.qll` library general purpose.  It resides in the query directory.  I can imagine a future where we have an `XML` library in `models`, and such a thing could grow from what we have here, but right now what it offers is quite specialized to the needs of the XXE query in particular.

There should be no changes to query results from this PR.

I'll check performance locally and with another DCA run.


geoffw0:
Local performance is about the same as before.

geoffw0:
DCA looks fine as well.

cpp/ql/src/Security/CWE/CWE-611/XXE.ql
```diff
@@ -13,344 +13,10 @@
  */
 
 import cpp
-import semmle.code.cpp.ir.dataflow.DataFlow
+import XML
+import Xerces
+import Libxml2
```
MathiasVP, line 18: Maybe the two libraries should be imported right after the definition of the `XxeFlowStateTransformer` class in `XML.ql`? In that way, we're sure that no one accidentally imports `XML` without any of the implementation classes?

MathiasVP (approved):
LGTM! I've got one single comment that you can choose to address or disagree with :)

cpp/ql/src/Security/CWE/CWE-611/XXE.ql
geoffw0: I've pushed that solution, but let me know if you don't like it.

MathiasVP: Yeah, it should be fine to import them at the top 👍.

MathiasVP (approved):

PR #9181, opened 2022-05-16
Java: Rename `FloatingPointLiteral` to `FloatLiteral`

### LINKED ISSUE #9170: Java: Rename `FloatingPointLiteral` to `FloatLiteral`
Currently the CodeQL class modeling a Java `float` literal such as `0.1f` is called `FloatingPointLiteral`. This is rather confusing because "floating point" usually refers to the concept as a whole, and not to a specific data type (`double` or `float`). This is also how the [JLS uses this term](https://docs.oracle.com/javase/specs/jls/se18/html/jls-3.html#jls-3.10.2):

> A floating-point literal is of type `float` if it is suffixed with an ASCII letter `F` or `f`; otherwise its type is `double`

("floating-point literal" here covers both `float` and `double`)

Would it therefore make sense to rename `FloatingPointLiteral` to `FloatLiteral`, and to keep `FloatingPointLiteral` as deprecated alias for now?

Marcono1234:
Resolves #9170

### DIFF
--- a/java/ql/src/Likely Bugs/Comparison/NoComparisonOnFloats.ql
+++ b/java/ql/src/Likely Bugs/Comparison/NoComparisonOnFloats.ql
@@ -12,16 +12,6 @@
 import semmle.code.java.Type
 import semmle.code.java.Expr
 
-// Either `float`, `double`, `Float`, or `Double`.
-class Floating extends Type {
-  Floating() {
-    exists(string s | s = this.getName().toLowerCase() |
-      s = "float" or
-      s = "double"
-    )
-  }
-}
-
 predicate trivialLiteral(Literal e) {
   e.getValue() = "0.0" or
   e.getValue() = "0" or
@@ -61,7 +51,7 @@ predicate similarVarComparison(EqualityTest e) {
 
 from EqualityTest ee
 where
-  ee.getAnOperand().getType() instanceof Floating and
+  ee.getAnOperand().getType() instanceof FloatingPointType and
   not ee.getAnOperand() instanceof NullLiteral and
   not trivialLiteral(ee.getAnOperand()) and
   not definedConstant(ee.getAnOperand()) and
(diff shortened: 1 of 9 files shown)


java/ql/src/Likely Bugs/Comparison/NoComparisonOnFloats.ql
```diff
@@ -61,7 +51,7 @@ predicate similarVarComparison(EqualityTest e) {
 
 from EqualityTest ee
 where
-  ee.getAnOperand().getType() instanceof Floating and
+  ee.getAnOperand().getType() instanceof FloatingPointType and
```
Marcono1234, line 54: This change it not actually directly related, but I noticed it while searching for usages of the old expression name.
If you want I can revert this change.

java/ql/test/kotlin/library-tests/literals/literals.expected
```diff
@@ -15,9 +15,9 @@
 | literals.kt:17:30:17:33 | 15 | LongLiteral |
 | literals.kt:18:30:18:39 | 11 | LongLiteral |
 | literals.kt:19:30:19:38 | 1234567 | LongLiteral |
-| literals.kt:20:32:20:35 | 0.0 | FloatingPointLiteral |
-| literals.kt:21:32:21:37 | 123.4 | FloatingPointLiteral |
-| literals.kt:22:32:22:38 | -123.4 | FloatingPointLiteral |
+| literals.kt:20:32:20:35 | 0.0 | FloatLiteral |
+| literals.kt:21:32:21:37 | 123.4 | FloatLiteral |
+| literals.kt:22:32:22:38 | -123.4 | FloatLiteral |
```
Marcono1234, line 20: Could not verify whether this test still succeeds because test execution from within VS Code failed:
```
extraction error: #redacted#\codeql\java\ql\test\kotlin\library-tests\literals\literals.ql
IOException while executing process with args: [#redacted#\AppData\Roaming\Code\User\globalStorage\github.vscode-codeql\distribution23\codeql\java\tools\kotlin-extractor, --java-home, #redacted#\AppData\Roaming\Code\User\globalStorage\github.vscode-codeql\distribution23\codeql\tools\win64\java, --, -J-Xmx2G, literals.kt]
(eventual cause: IOException "CreateProcess error=193, %1 is not a valid Win32 application")
```

But I assume this is "expected" while Kotlin support is still work in progress.

aschackmull (approved):
LGTM, thank you for the PR.

PR #9144, opened 2022-05-13
Swift: Introduce visitors

AlexDenisov:

PR #9185, opened 2022-05-17
Swift: move TBD code to ql

redsun82:
This allows to avoid bypassing label type correcness in the extractor,
and allows to independently resolve TBD extractions, as with this
approach TBD nodes do have the correctly typed trap label. The TBD
status is now a predicate on the QL side.

The code taking care of making unknwon entities be printed as `TBD(Class)`
is in the hand written `Element.qll` file.

This requires:
* a default visit using the correct type, which is achieved via macro
  metaprogramming in `VisitorBase.h`, following the way
  `swift::ASTVisitor` is programmed
* a mapping from labels to corresponding binding trap entries. The
  functor is defined in `TrapTagTraits.h` and instantiated in generated
  `TrapEntries.h`
* Binding trap entries for TBD unknown entities must not have any other
  field than the `id` (after all, we are supposed to not extract them
  yet). This is why all unextracted fields in `schema.yml` have been
  commented out, and will be uncommentend when visitors are added

closes https://github.com/github/codeql-c-team/issues/1024


swift/extractor/visitors/StmtVisitor.h
```diff
@@ -1,39 +1,12 @@
...
  public:
-#define ABSTRACT_STMT(CLASS, PARENT)                           \
-  void visit##CLASS##Stmt(swift::CLASS##Stmt* E) {             \
-    return static_cast<CrtpSubclass*>(this)->visit##PARENT(E); \
-  }
-#define STMT(CLASS, PARENT) ABSTRACT_STMT(CLASS, PARENT)
-#include "swift/AST/StmtNodes.def"
-};
-
-}  // namespace detail
```
redsun82, line 23: this code is made obsolete by the default unknown emission in `VisitorBase.h`

MathiasVP:
> There are quite a lot of code scanning complaints. What is the deal with those?

These are because ql-for-ql doesn't know about our autogeneration scheme for Swift. I've written about it [here](https://github.com/github/codeql-ql/issues/145). Since that's a private repo I'll copy/paste the relevant part:

<i>Since https://github.com/github/codeql/pull/9100 we added getAPrimaryQlClass for Swift the [ql/primary-ql-class-consistency](https://github.com/github/codeql/blob/main/ql/ql/src/queries/style/GetAPrimaryQlClassConsistency.ql) query currently (rightfully) reports a lot of inconsistencies.

This is because we generate two classes for each ... thing ... in Swift. For instance, we have ApplyExpr and ApplyExprBase for a function call. The ApplyExprBase class is an autogenerated and private class that contains (among other things) the implementation of getAPrimaryQlClass, whereas ApplyExpr contains additional manually written predicates.

This means that ApplyExprBase has a getAPrimaryQlClass predicate that returns "ApplyExpr" (instead of the expected "ApplyExprBase").</i>

It should be straightforward to fix the ql-for-ql query to recognize this. The issue I wrote has two solutions, and Erik has suggested a third solution (and they should all be one-line fixes in ql-for-ql).

redsun82:
> There are quite a lot of code scanning complaints. What is the deal with those?

@jketema This is a known issue, that I kindly asked @MathiasVP to put in writing, so you'll find details here: https://github.com/github/codeql-ql/issues/145 :)

redsun82:
@sashabu or @jketema , can you have a look at the C++ part? If you start looking, can you immediately comment, so the other does not need to? And @MathiasVP, you might want to have a look at `elements/Element.qll`, which is the only non-generated QL code required for this to work.

MathiasVP:
> And @MathiasVP, you might want to have a look at elements/Element.qll, which is the only non-generated QL code required for this to work.

Thanks for the heads up. This LGTM 👍.

sashabu:
> @sashabu or @jketema , can you have a look at the C++ part? If you start looking, can you immediately comment, so the other does not need to?

I'll step back since Jeroen's already started commenting.

jketema:
> I'll step back since Jeroen's already started commenting.

Just to be clear: I started commenting on the QL part, which I am much more interested in.

swift/extractor/visitors/VisitorBase.h
```diff
@@ -0,0 +1,63 @@
...
+
+#include <swift/AST/ASTVisitor.h>
+#include <swift/AST/TypeVisitor.h>
+
+#include "swift/extractor/SwiftDispatcher.h"
+
+namespace codeql {
+
+namespace detail {
+class DispatcherWrapper {
```
jketema, line 11: First it seemed to me that this just created an extra level of indirection around `SwiftDispatcher` and didn't add any functionality. However, this just seems to be the shared base class for `AstVisitorBase` and `TypeVisitorBase`. Wouldn't `VisitorBase` be a more appropriate name. Also If not other `VisitorBase`s will be added in the future and if no methods are expected to be added to this class, it's probably perfectly fine to just introduce a dispatcher field in both `VisitorBase`s?

swift/extractor/SwiftDispatcher.h
```diff
@@ -57,18 +21,21 @@ class SwiftDispatcher {
...
+  void emit(const Entry& entry) {
+    trap.emit(entry);
+  }
+
   // This is a helper method to emit TRAP entries for AST nodes that we don't fully support yet.
-  template <typename Parent, typename Child>
-  void TBD(Child* entity, const std::string& suffix) {
-    using namespace std::string_literals;
+  template <typename E>
+  void emitUnknown(E* entity) {
```
jketema, line 31: I'm not sure if this would make sense, and maybe this is already done somewhere else, but it might be appropriate to log each call of this function? So we can go back and see if there are any remaining entity types that we missed.

swift/codegen/schema.yml
```diff
@@ -28,9 +28,12 @@ Location:
   end_line: int
   end_column: int
 
-Type:
-  diagnostics_name: string
-  canonical_type: Type
+Type: {}
+#  diagnostics_name: string
```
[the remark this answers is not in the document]
jketema: I think this might be related to the `static_assert` in `SwiftDispatcher`?

swift/extractor/visitors/VisitorBase.h
```diff
@@ -0,0 +1,63 @@
...
+}  // namespace detail
+
+// we want to override the default swift visitor behaviour of chaining calls to immediate
+// superclasses by default and instead provide our own TBD default (using the exact type)
+#define DEFAULT(KIND, CLASS, PARENT) \
+  void visit##CLASS##KIND(swift::CLASS##KIND* e) { dispatcher_.emitUnknown(e); }
+
+// base class for our AST visitors, getting a SwiftDispatcher member and default emission for
+// unknown/TBD entities
+template <typename CrtpSubclass>
```
jketema, line 29: What does `crtp` stand for? Is this for me somewhat cryptic name needed for the macro programming going on below?

swift/ql/lib/codeql/swift/elements/Element.qll
```diff
@@ -1,4 +1,11 @@
-// generated by codegen/codegen.py, remove this comment if you wish to edit this file
 private import codeql.swift.generated.Element
 
-class Element extends ElementBase { }
+class Element extends ElementBase {
```
jketema, line 3: Something that still eludes me. Why is there both an `Element` and an `ElementBase`? Why isn't there just one class?

swift/codegen/templates/trap_traps.mustache
```diff
@@ -29,5 +29,14 @@ inline std::ostream &operator<<(std::ostream &out, const {{name}}{{trap_affix}}
...
 }
+{{#id}}
+
+namespace detail {
+template<>
+struct ToBindingTrapFunctor<{{type}}> {
+  using type = {{name}}{{trap_affix}};
+};
+}
+{{/id}}
```
jketema, line 40: Am I correct that the generated QL code is included in the repo (for obvious reasons), but the generated C++ code is not (it's just generated during the build when needed)?

swift/extractor/SwiftDispatcher.h
```diff
@@ -57,18 +21,21 @@ class SwiftDispatcher {
...
     auto label = assignNewLabel(entity);
-    auto kind = detail::getKindName<Parent>(static_cast<const Parent*>(entity)->getKind());
-    auto name = "TBD ("s + kind + suffix + ")";
-    if constexpr (std::is_same_v<Parent, swift::TypeBase>) {
-      trap.emit(UnknownTypesTrap{label, name});
-    } else {
-      trap.emit(UnknownAstNodesTrap{label, name});
-    }
+    using Trap = BindingTrapOf<E>;
+    static_assert(sizeof(Trap) == sizeof(label),
```
jketema, line 34: `ToBindingTrapFunctor` seems pretty empty to me, so doesn't this always hold? I'm probably just missing details here.

jketema (commented):
Generally looks ok, but I think there are some details I'm missing. See questions below.

swift/extractor/SwiftDispatcher.h
AlexDenisov: As soon as we have `PrintAst.ql` ready, we will have a lot of `SomeThing (TBD)` nodes visible in the AST.
So they are going to be "logged" and visible. In fact, we relied on this feature a lot during the PoC development, i.e. "pick and extract the next TBD node".

swift/extractor/visitors/VisitorBase.h
AlexDenisov: That's [Curiously recurring template pattern](https://en.wikipedia.org/wiki/Curiously_recurring_template_pattern), as proposed by @sashabu [here](https://github.com/github/codeql/pull/9144#discussion_r872473570).

swift/codegen/schema.yml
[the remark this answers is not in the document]
redsun82: TL;DR is that we only need to comment out class properties of final classes, but it was more systematic (and easier with regexp edits) to comment out all properties, and bring them back in the upcoming PRs.

The more detailed explanation follows.

In order to emit a correctly typed `@` trap ids (say for example `@concrete_var_decl`), we need to emit the corresponding "binding" trap entry (`concrete_var_decls`). Now if this had a `@` ref column, we would need to fill that with a valid label. So for example if we have
```
my_entries(unique int id: @my_entry, int prop: @other ref);
```
which generates trap entry
```cpp
strut MyEntriesTrap { TrapLabel<MyEntryTag> id{}; TrapLabel<OtherTag> prop{}; }
```
and we emit `MyEntriesTrap{label}`, this will result in a trap emission like
```
my_entries(#1a, #ffffffff)
```
and the trap importer will silently ignore all entries using undefined labels (took me a while to diagnose). It would be ok to have fields in non-binding traps (for example those related to abstract classes) because we just do not emit those for unknown entities, and it would also be ok for non-class properties to be present (because default initialization would not pose problems for the trap importer). But I just took an easier restriction to check, and removed all properties from schema as it was easier.

swift/ql/lib/codeql/swift/elements/Element.qll
MathiasVP: There are a couple of autogenerated predicates whose rootdef is in `ElementBase` (e.g., the `resolve` predicate that enables us to hide implicit conversions from the "main" AST ). We _could_ put those predicates in `Element` instead, but then we'd have a weird situation where `Element` should contain both autogenerated predicates and manually written predicates.

swift/codegen/schema.yml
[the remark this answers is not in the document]
redsun82: By the way, in an upcoming PR I added an `assert` checking that we never emit an undefined label, which will catch those unfortunate default initialization of labels. We can rework the `assert` later with a proper diagnostics module. Another mitigation against the risk of using default-initialized labels could be to delete the default constructor of trap entries and only have the constructors with the full number of fields initialized.

swift/codegen/BUILD.bazel
```diff
@@ -27,7 +27,10 @@ py_binary(
 py_binary(
     name = "trapgen",
     srcs = ["trapgen.py"],
```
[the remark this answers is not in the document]
redsun82: will do!

swift/ql/lib/codeql/swift/elements/Element.qll
jketema: So the general pattern will be: each `Base` class only contains auto generated stuff, and each class extending it will contain handwritten functionality?

swift/extractor/visitors/VisitorBase.h
redsun82: For the first point I'm perfectly fine with renaming this `VisitorBase`. For the second, we already share the constructor this way, and I already have some draft PRs adding shared methods here, so I would leave the shared base class.

swift/extractor/SwiftDispatcher.h
redsun82: this makes perfect sense, however I will just add a TODO for now, as we still don't have a proper logging infrastructure yet.

swift/extractor/visitors/VisitorBase.h
redsun82: this was suggested by @sashabu and stands for the [curiously recurring template pattern](https://en.wikipedia.org/wiki/Curiously_recurring_template_pattern). It's probably better to put a comment about it, as also for me that have used this pattern a lot and am very familiar with its full nomenclautre, the acronym takes a bit to ring a bell.

swift/codegen/BUILD.bazel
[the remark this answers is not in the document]
jketema: Feel free to do this in a separate PR.

swift/extractor/visitors/VisitorBase.h
jketema: If methods are going to be added, then definitely having a class here is useful.

jketema: So just rename.

swift/ql/lib/codeql/swift/elements/Element.qll
redsun82: You might find more details about this choice in [this section of the EDR document](https://docs.google.com/document/d/1vFVZKEvocpQE-9SA1EjYfhKAIyTNGCxxdloIAZWdkvY/edit#heading=h.sinlay1iaclz).

In a nutshell, this is the separation we have between fully generated methods and hand written ones. `XXXBase` is fully generated and kept up to date with `schema.yml`. `XXX` is initially a generated empty stub, but can be customized inheriting from the generated one. 

The alternatives could be:
* one generated file, that stops being generated once you edit it -> this would mean you would lose new generated methods and would need to import them by hand
* like now with two files, but without using inheritance -> you would probably need a lot of boilerplate to "import" the generated implementations in the non-generated file (I think this is similar to what python has)
* some kind of specifications of what needs to be added to the generated file (a diff file? some other "include" specification for the codegen suite? probably something triggered from `schema.yml` or some kind of annotation file thereof): seemed a less clean solution requiring more effort

jketema: Current approach is fine with me. I was just trying to understand what is going on.

swift/codegen/templates/trap_traps.mustache
redsun82: yes, you are right. However in a PR you can consult the generated headers by going to the summary of the `Swift: Check code generation / codegen` action run ([here](https://github.com/github/codeql/actions/runs/2356916003) for this PR at the moment)

swift/extractor/SwiftDispatcher.h
redsun82: `ToBindingTrapFunctor` is implemented by `TrapEntries.h` in the [generated headers](https://github.com/github/codeql/suites/6588171517/artifacts/247105982). This particular assertion fires whenever a final class in `schema.yml` has a single property (i.e. non-optional, non-repeated). For example uncommenting line 995 in `schema.yml` (field `introducer_int` in class `ConcreteVarDecl`) will result in 
```
./swift/extractor/SwiftDispatcher.h:34:5: error: static_assert failed due to requirement 'sizeof(codeql::ConcreteVarDeclsTrap) == sizeof (label)' "Binding traps of unknown entities must only have the `id` field (the class should be empty in schema.yml)"
    static_assert(sizeof(Trap) == sizeof(label),
    ^             ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

swift/codegen/templates/trap_traps.mustache
redsun82: (FYI it does require 2 clicks from the PR discussion, first on the run `details`, and then on the `Summary` at the left, which might not be super intuitive)

swift/codegen/BUILD.bazel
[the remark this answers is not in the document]
redsun82: ok I've [opened an internal ticket for this](https://github.com/github/codeql-c-team/issues/1036)

swift/extractor/SwiftDispatcher.h
jketema: I still don't get this, because:
```
template<>
struct ToBindingTrapFunctor<TrapLabel<FileTag>> {
  using type = FilesTrap;
};
```
pretty much looks like an empty class to me.

redsun82: ah, sorry, I did not get that that was your doubt.

It's the `type` nested alias that does the trick. From `TrapTagTraits`:
```cpp
template <typename T>
using BindingTrapOf = typename detail::ToBindingTrapFunctor<TrapLabelOf<T>>::type;
```
so for example `BindingTrapOf<swift::VarDecl>` will resolbe to `typename detail::ToBindingTrapFunctor<TrapLabel<ConcreteVarDeclTag>>::type` which in the generated file is defined as `ConcreteVarDeclsTrap`.

redsun82: This is the same pattern used for transformation type traits like [`std::underlying_type`](https://en.cppreference.com/w/cpp/types/underlying_type) and similar transformations in `type_traits`. Although I see I am using mixed naming conventions using lowercase `type`. The `std` convention would have two separate definitions like in
```cpp
template <typename T>
using binding_trap_of = detail::to_binding_trap_functor<...>;

template <typename T>
using binding_trap_of_t = typename binding_trap_of<T>::type;
```

jketema: Sure but that's because there's actual work needed to implement `underlying_type` it's not just an alias like we have here.

redsun82: > Got it. But then why do this at all and not just:
> 
> ```
> static_assert(sizeof(E) == sizeof(label),
> ```

`E` is the swift frontend tool entity (like `swift::VarDecl`) which is distinct from the trap entry (for example `ConcreteVarDeclsTrap`). We also need to compute that trap entry class to actually emit it on [line 37 below](https://github.com/github/codeql/blob/redsun82/swift-tbd-rework/swift/extractor/SwiftDispatcher.h#L37)

jketema: Yes, that should probably just be `TrapLabel<E>{label}`.?

redsun82: > Sure but that's because there's actual work needed to implement `underlying_type` it's not just an alias like we have here.

I think that alias is there to hide away an implementation detail, which does the actual work in terms of template specialization. 

Would you find it clearer if we wrote something like
```cpp
  template <typename Label>
  using BindingTrapOf = typename detail::ToBindingTrap<Label>::type;
```
and then in the code using it did:
```cpp
  using Trap = BindingTrapOf<decltype(label)>;
```
or more explicitly
```
  using Label = TrapLabelOf<E>;
  using Trap = BindingTrapOf<Label>;
  Label label = assignNewLabel(e);
  ...
```
?

jketema: I understand less and less of what this is code is doing. Why is all this indirect mess needed at all? This is going to cause a lot of trouble for anyone who is going to maintain this in the future.

jketema: I want something that's just 2 lines of code instead of 12.

redsun82: > Yes, that should probably just be `TrapLabel<E>{label}`.?

Hmm, no, labels cannot be emitted in place of trap entries. `TrapLabelOf<E>` which resolves to `TrapLabel<TrapTagOf<E>>` (for example, `TrapLabel<ConcreteVarDeclTag>`) denotes a typed trap label, so something that is a `#` label in the trap file. We need to emit an actual trap (like `concrete_var_decls(#1e)`) which in C++ is done printing a `ConcreteVarDeclsTrap` class.

jketema (approved):

PR #9186, opened 2022-05-17
Kotlin: Add support for InlineExpectationsTest

atorralba:
Gives `InlineExpectationsTest` the ability to recognize Kotlin comments, allowing its use in `.kt` tests.

### DIFF
--- a/java/ql/lib/semmle/code/java/Javadoc.qll
+++ b/java/ql/lib/semmle/code/java/Javadoc.qll
@@ -153,6 +153,15 @@ class KtComment extends Top, @ktcomment {
   /** Gets the full text of this comment. */
   string getText() { ktComments(this, _, result) }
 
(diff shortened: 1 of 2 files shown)


smowton:
Looks like you're trying to accept `/* c-style comments */` for expectations, but that would be inconsistent with Java, which only allows EOL comments for expectations -- suggest instead exposing the `type` column of `ktComments` via `class KtComment`, which is an integer (1 = `// EOL comment`, 2 = `/* block comment */`, 3 = `/** doc comment /`), then we can write `comment.isEolComment()` and simply strip `//` and any leading whitespace from the comment text.

atorralba:
Good suggestion, thanks @smowton. Is dbf249b what you were thinking of?

smowton (approved):

PR #9188, opened 2022-05-17
QL: Allow class + `Base` in `ql/primary-ql-class-consistency`

MathiasVP:
This should hopefully mean fewer false positives for `ql/primary-ql-class-consistency` in Swift.

### DIFF
--- a/ql/ql/src/queries/style/GetAPrimaryQlClassConsistency.ql
+++ b/ql/ql/src/queries/style/GetAPrimaryQlClassConsistency.ql
@@ -10,18 +10,31 @@
 
 import ql
 
-from ClassPredicate pred, String constant
+/**
+ * Gets a string that is allowed to be returned from the `getAPrimaryQlClass` implementation
+ * in class `c`.
+ *
+ * For most languages this is just the name of `c`, but QL for Swift implements
+ * the `getAPrimaryQlClass` predicate for a class `Foo` in another class `FooBase`.
+ */
+string getAnAllowedResultForClass(Class c) {
+  result = c.getName()
+  or
+  c.getName() = result + "Base"
+}
+
+from ClassPredicate pred, String constant, Class c
 where
-  exists(string className, string constantName |
-    pred.getParent().getName() = className and
-    pred.getName() = "getAPrimaryQlClass" and
+  c = pred.getParent() and
+  pred.getName() = "getAPrimaryQlClass" and
+  exists(string constantName |
     constant = pred.getBody().(ComparisonFormula).getRightOperand() and
     constant.getValue() = constantName and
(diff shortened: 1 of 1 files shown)


ql/ql/src/queries/style/GetAPrimaryQlClassConsistency.ql
```diff
@@ -10,18 +10,29 @@
...
-from ClassPredicate pred, String constant
+/**
+ * Gets a string that is allowed to be returned from the `getAPrimaryQlClass` implementation
+ * in class `c`.
+ *
+ * For most languages this is just the name of `c`, but QL for Swift implements
+ * the `getAPrimaryQlClass` predicate for a class `Foo` in another class `FooBase`.
+ */
+string getAnAllowedResultForClass(Class c) {
+  exists(string className | c.getName() = className | result = className + ["", "Base"])
```
erik-krogh, line 21:
```suggestion
  result = c.getName() + ["", "Base"]
```
?

MathiasVP: Indeed! That was me copy-pasting without thinking 😂. Thanks!

erik-krogh (dismissed):

MathiasVP:
@erik-krogh do you mind re-approving? I had accidentally allowed `FooBaseBase` as a result of `getAPrimaryQlClass` in class `FooBase` instead of allowing `Foo` as a result. That's fixed now! And this is correctly reflected in Code Scanning: https://github.com/github/codeql/pull/9188/checks?check_run_id=6474343910

ql/ql/src/queries/style/GetAPrimaryQlClassConsistency.ql
```diff
@@ -10,18 +10,31 @@
...
-from ClassPredicate pred, String constant
+/**
+ * Gets a string that is allowed to be returned from the `getAPrimaryQlClass` implementation
+ * in class `c`.
+ *
+ * For most languages this is just the name of `c`, but QL for Swift implements
+ * the `getAPrimaryQlClass` predicate for a class `Foo` in another class `FooBase`.
+ */
+string getAnAllowedResultForClass(Class c) {
+  result = c.getName()
```
intrigus-lgtm, line 21: I guess it's not possible to restrict this to only apply to `Swift` dbs?

erik-krogh (approved):

ql/ql/src/queries/style/GetAPrimaryQlClassConsistency.ql
MathiasVP: We could restrict it to only apply to files that live in the Swift directory, but we decided this solution was sufficient.

PR #9189, opened 2022-05-17
Swift: declaration visitor

redsun82:
This transfers the current state of `DeclVisitor` from the
proof-of-concept.

TODO: make the `declarations` tests in `extractor-tests` more
comprehensive.


MathiasVP:
Did the visitors transfer somewhat cleanly from Swift 5.5 to Swift 5.6?

redsun82:
> Did the visitors transfer somewhat cleanly from Swift 5.5 to Swift 5.6?

For declarations yes, without any hiccup. The TBD PR had some hiccups related to new entities, but nothing serious. And in the upcoming statement visitor I had to dig up some of my changes from the 5.6 branch of the PoC, and I expect the same for the expression visitor

AlexDenisov:
> TODO: make the declarations tests in extractor-tests more comprehensive.

@redsun82 could you expand on what is missing?

AlexDenisov (approved):

MathiasVP:
The schema and (autogenerated) QL look good to me!

redsun82:
> > TODO: make the declarations tests in extractor-tests more comprehensive.
> 
> @redsun82 could you expand on what is missing?

Most of what we actually extract is missing. If you look at what I added for `FuncDecl`, we should be doing something similar for each and every kind of declaration, testing their properties and not just their existence.

Which is super boring, so I'm thinking that these kind of QL tests should really be generated as well. I will open an issue for that, but for now I would avoid going so much in depth with the tests by hand.

sashabu:
I've removed both @jketema and myself from reviewers per https://github.slack.com/archives/CP75PA8Q1/p1652790443962019?thread_ts=1652784504.197839&cid=CP75PA8Q1.

PR #9191, opened 2022-05-17
Ruby: Force cached taint tracking predicates to be evaluated in data flow stage

hvitved:
Avoids recomputation of some flow-summary related predicates.


aibaars (approved):
LGTM

PR #9193, opened 2022-05-17
JS: Add individual per-security-query counting queries

TomBolton:
The ATM project uses an AML pipeline to select which CodeQL databases to use as the evaluation databases for each security query being boosted. 

Previously, evaluation sets for _all_ security queries would be produced at once, using the `CountAlertsAndEndpoints.ql` query. However, this query does not scale with adding more security queries, and it is not efficient to create evaluation sets for all queries during every pipeline run (when adding a new query, you simply want to create an evaluation set for that one query, and not the others).

Therefore, after a discussion, it was deemed better to specify one security query that you would like to create an evaluation set for. For this, we would need a way to count alerts and endpoints for a single query. The most simple solution is to add per-query counting queries, `Count{QUERY_NAME}.ql`, which can be used by the selection pipeline.

The selection pipeline has been tested with the `CountCodeInjection.ql` query.

### DIFF
--- a/javascript/ql/experimental/adaptivethreatmodeling/modelbuilding/counting/CountAlertsAndSinks.qll
+++ b/javascript/ql/experimental/adaptivethreatmodeling/modelbuilding/counting/CountAlertsAndSinks.qll
@@ -1,0 +1,20 @@
+/*
+ * For internal use only.
+ *
+ *
+ * Count the number of sinks and alerts for a particular dataflow config.
+ */
+
+import javascript
+import evaluation.EndToEndEvaluation
+
+query predicate countAlertsAndSinks(int numAlerts, int numSinks) {
+  numAlerts =
+    count(DataFlow::Configuration cfg, DataFlow::Node source, DataFlow::Node sink |
+      cfg.hasFlow(source, sink) and not isFlowExcluded(source, sink)
+    ) and
+  numSinks =
+    count(DataFlow::Node sink |
+      exists(DataFlow::Configuration cfg | cfg.isSink(sink) or cfg.isSink(sink, _))
+    )
+}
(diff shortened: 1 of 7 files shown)


esbena:
This introduces a lot of duplicate codeql code that can be avoided using imports and query predicates. 

I think each query could look something like this:

```ql
import semmle.javascript.security.dataflow.CodeInjectionQuery
import CountThings
```

Where CountThings.ql looks something like:

```ql
import javascript
import evaluation.EndToEndEvaluation

query predicate countThings(int numAlerts, int numSinks) { 
  numAlerts =
    count(DataFlow::Node source, DataFlow::Node sink |
      cfg.hasFlow(source, sink) and not isFlowExcluded(source, sink)
    ) and
  numSinks = count(DataFlow::Node sink |
    exists(DataFlow::Configuration cfg | cfg.isSink(sink) or cfg.isSink(sink, _))
  )
```

Could you consider if such a refactoring is possible / worth it?

TomBolton:
Yes, good point @esbena and I definitely think it's worth it.

Basic QL question: what would the individual queries look like? I.e. how would they actually use the `CountThings` import? I've checked the documentation and it's not clear how to import and use a predicate

TomBolton:
I've just attempted the suggestion refactor @esbena - would you mind giving it another look when you have time?

TomBolton:
@github/codeql-ml-powered-queries-reviewers if anyone has a spare moment, could someone provide a quick review?

esbena (approved):
the `as CodeInjection/TaintedPath/...` namings can be dropped now.
Otherwise LGTM.

TomBolton:
Thanks for the review @esbena - I've now removed the `as` in the imports. 

I accidentally dismissed your approval, would you mind approving again sorry?

esbena (approved):
(Your edit caused the approval dismissal)

PR #9195, opened 2022-05-17
Java: Performance fixes for local flow relation

aschackmull:
The local flow relation was accidentally made mutually recursive with `AdditionalValueStep`. It makes sense to allow the use for local flow to define `AdditionalValueStep::step`, so removing the latter from the former allows this without recursion.

A non-monotonic dummy reference is added to prevent re-introduction of recursion here, and this highlighted the need for a minor refactor in `FlowSummaryImpl` (since `ReturnNodeExt` depends on local flow).

The local flow relation is also fixed to force the use of the fastTC hop, as this otherwise was observed to result in a very slow materialisation.

### DIFF
--- a/csharp/ql/lib/semmle/code/csharp/dataflow/internal/FlowSummaryImpl.qll
+++ b/csharp/ql/lib/semmle/code/csharp/dataflow/internal/FlowSummaryImpl.qll
@@ -813,9 +813,7 @@ module Private {
      * be useful to include in the exposed local data-flow/taint-tracking relations.
      */
     predicate summaryThroughStepTaint(ArgNode arg, Node out) {
-      exists(ReturnNodeExt ret |
-        summaryLocalStep(summaryArgParam(arg, ret, out), ret, false)
-      )
+      exists(ReturnNodeExt ret | summaryLocalStep(summaryArgParam(arg, ret, out), ret, false))
     }
 
     /**
--- a/csharp/ql/lib/semmle/code/csharp/dataflow/internal/FlowSummaryImpl.qll
+++ b/csharp/ql/lib/semmle/code/csharp/dataflow/internal/FlowSummaryImpl.qll
@@ -791,15 +791,30 @@ module Private {
     }
 
     /**
-     * Holds if `arg` flows to `out` using a simple flow summary, that is, a flow
-     * summary without reads and stores.
+     * Holds if `arg` flows to `out` using a simple value-preserving flow
+     * summary, that is, a flow summary without reads and stores.
      *
      * NOTE: This step should not be used in global data-flow/taint-tracking, but may
      * be useful to include in the exposed local data-flow/taint-tracking relations.
      */
-    predicate summaryThroughStep(ArgNode arg, Node out, boolean preservesValue) {
+    predicate summaryThroughStepValue(ArgNode arg, Node out) {
+      exists(ReturnKind rk, ReturnNode ret, DataFlowCall call |
+        summaryLocalStep(summaryArgParam0(call, arg), ret, true) and
+        ret.getKind() = rk and
+        out = getAnOutNode(call, rk)
+      )
+    }
+
+    /**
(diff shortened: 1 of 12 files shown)


java/ql/lib/semmle/code/java/dataflow/internal/FlowSummaryImpl.qll
```diff
@@ -790,15 +790,31 @@ module Private {
...
+     * Holds if `arg` flows to `out` using a simple value-preserving flow
+     * summary, that is, a flow summary without reads and stores.
      *
      * NOTE: This step should not be used in global data-flow/taint-tracking, but may
      * be useful to include in the exposed local data-flow/taint-tracking relations.
      */
-    predicate summaryThroughStep(ArgNode arg, Node out, boolean preservesValue) {
+    predicate summaryThroughStepValue(ArgNode arg, Node out) {
+      exists(ReturnKind rk, ReturnNode ret, DataFlowCall call |
+        // summaryLocalStep(summaryArgParam(arg, rk, out), ret, true) and
```
atorralba, line 801: Should we remove this?

atorralba (commented):
Looks plausible to me, just one minor comment.

There are a couple of tricks (namely `fastTC` and `(any() or strictcount(Node n1, Node n2, AdditionalValueStep a | a.step(n1, n2)) < 0)`) that I would like to learn more about at some point, though :)

java/ql/lib/semmle/code/java/dataflow/internal/FlowSummaryImpl.qll
aschackmull: done

aschackmull:
> There are a couple of tricks (namely `fastTC` and `(any() or strictcount(Node n1, Node n2, AdditionalValueStep a | a.step(n1, n2)) < 0)`) that I would like to learn more about at some point, though :)

`fastTC` just means the same as the `+` suffix on a predicate, but forces the compiler to use the specialised code for this (using `+` leaves the choice to the optimiser, which will often also result in using `fastTC`, but didn't in this case (and that happened to be bad)).
The other ugly line of obscure QL is a hack. It doesn't really matter much what it contains as `and (any() or ...)` optimises away to nothing at all, but a predicate reference inside a `strictcount` can't be recursive, so we ensure that we don't introduce accidental recursion in the future through this hack.

csharp/ql/lib/semmle/code/csharp/dataflow/internal/FlowSummaryImpl.qll
```diff
@@ -791,18 +791,31 @@ module Private {
...
      * be useful to include in the exposed local data-flow/taint-tracking relations.
      */
-    predicate summaryThroughStep(ArgNode arg, Node out, boolean preservesValue) {
-      exists(ReturnNodeExt ret |
-        summaryLocalStep(summaryArgParam(arg, ret, out), ret, preservesValue)
+    predicate summaryThroughStepValue(ArgNode arg, Node out) {
+      exists(ReturnKind rk, ReturnNode ret, DataFlowCall call |
+        summaryLocalStep(summaryArgParam0(call, arg), ret, true) and
+        ret.getKind() = rk and
+        out = getAnOutNode(call, rk)
```
hvitved, line 804: Shouldn't we use the same pragmas here as in `summaryArgParam`?

java/ql/lib/semmle/code/java/dataflow/internal/DataFlowUtil.qll
```diff
@@ -118,6 +121,16 @@ predicate localFlowStep(Node node1, Node node2) {
  */
 cached
 predicate simpleLocalFlowStep(Node node1, Node node2) {
```
hvitved, line 123: Perhaps wrap `localFlowStep` and `simpleLocalFlowStep` in a cached internal module to make sure they are evaluated together.

aschackmull:
DCA looks good now.

hvitved (approved):

PR #9200, opened 2022-05-17
Python: Modernise weak file permissions query

tausbn:
Using API graphs instead of points-to.

Unfortunately, some results will be lost because of this, due to the fact that points-to tracks bitwise operations on small numbers (i.e. flags), whereas API graphs does no such thing. This means using something like `stat.S_IWUSR | stat.S_IWGRP` will not work.

A custom type tracker (like the one used for `re` flags) could be used to recapture this behaviour, but I think that's best left as future work, as it's not clear to me that this query is actually worth the effort it would take to implement this.

---

As this is just modernising a somewhat minor query, I don't think it needs a change note (but I'll happily add one if necessary).

### DIFF
--- a/python/ql/src/Security/CWE-732/WeakFilePermissions.ql
+++ b/python/ql/src/Security/CWE-732/WeakFilePermissions.ql
@@ -12,6 +12,7 @@
  */
 
 import python
+import semmle.python.ApiGraphs
 
 bindingset[p]
 int world_permission(int p) { result = p % 8 }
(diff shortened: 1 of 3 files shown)


yoff (commented):
I am generally happy with rewriting `pointsTo` to `getAValueReachingRhs`. We do need to look into the version-dependence, though...

yoff (approved):
Bit crude, but fine in this case

PR #9207, opened 2022-05-18
Java: Add sources for Android external storage

joefarebrother:
Adds as sources the contents of files that come from an external storage directory.

### DIFF
--- a/java/ql/lib/semmle/code/java/frameworks/android/ExternalStorage.qll
+++ b/java/ql/lib/semmle/code/java/frameworks/android/ExternalStorage.qll
@@ -1,0 +1,46 @@
+/** Provides definitions for working with uses of Android external storage */
+
+import java
+import semmle.code.java.dataflow.DataFlow
+private import semmle.code.java.dataflow.ExternalFlow
+
+private class ExternalStorageDirSourceModel extends SourceModelCsv {
+  override predicate row(string row) {
+    row =
+      [
+        //"package;type;overrides;name;signature;ext;spec;kind"
+        "android.content;Context;true;getExternalFilesDir;(String);;ReturnValue;android-external-storage-dir",
+        "android.content;Context;true;getExternalFilesDirs;(String);;ReturnValue.ArrayElement;android-external-storage-dir",
+        "android.content;Context;true;getExternalCachesDir;(String);;ReturnValue;android-external-storage-dir",
+        "android.content;Context;true;getExternalCachesDirs;(String);;ReturnValue.ArrayElement;android-external-storage-dir",
+        "android.os;Environment;false;getExternalStorageDirectory;(String);;ReturnValue.ArrayElement;android-external-storage-dir",
+        "android.os;Environment;false;getExternalStoragePublicDirectory;(String);;ReturnValue.ArrayElement;android-external-storage-dir",
+      ]
+  }
+}
+
+private predicate externalStorageFlowStep(DataFlow::Node node1, DataFlow::Node node2) {
+  DataFlow::localFlowStep(node1, node2)
+  or
+  exists(ConstructorCall c | c.getConstructedType() instanceof TypeFile |
+    node1.asExpr() = c.getArgument(1) and
+    node2.asExpr() = c
+  )
+}
+
+private predicate externalStorageFlow(DataFlow::Node node1, DataFlow::Node node2) {
+  externalStorageFlowStep*(node1, node2)
+}
+
+/**
+ * Holds if `n` is a node that reads the contents of an external file in Android.
+ * This may be controlable by third-party applications, so is treated as a remote flow source.
+ */
+predicate androidExternalStorageSource(DataFlow::Node n) {
+  exists(ConstructorCall fInp, DataFlow::Node externalDir |
+    fInp.getConstructedType().hasQualifiedName("java.io", "FileInputStream") and
+    n.asExpr() = fInp and
+    sourceNode(externalDir, "android-external-storage-dir") and
+    externalStorageFlow(externalDir, DataFlow::exprNode(fInp.getArgument(0)))
+  )
+}
--- a/java/ql/lib/semmle/code/java/frameworks/android/ExternalStorage.qll
+++ b/java/ql/lib/semmle/code/java/frameworks/android/ExternalStorage.qll
@@ -10,12 +10,12 @@ private class ExternalStorageDirSourceModel extends SourceModelCsv {
     row =
       [
         //"package;type;overrides;name;signature;ext;spec;kind"
-        "android.content;Context;true;getExternalFilesDir;(String);;ReturnValue;android-external-storage-dir",
-        "android.content;Context;true;getExternalFilesDirs;(String);;ReturnValue;android-external-storage-dir",
-        "android.content;Context;true;getExternalCacheDir;();;ReturnValue;android-external-storage-dir",
-        "android.content;Context;true;getExternalCacheDirs;();;ReturnValue;android-external-storage-dir",
-        "android.os;Environment;false;getExternalStorageDirectory;();;ReturnValue;android-external-storage-dir",
-        "android.os;Environment;false;getExternalStoragePublicDirectory;(String);;ReturnValue;android-external-storage-dir",
+        "android.content;Context;true;getExternalFilesDir;(String);;ReturnValue;android-external-storage-dir;manual",
+        "android.content;Context;true;getExternalFilesDirs;(String);;ReturnValue;android-external-storage-dir;manual",
+        "android.content;Context;true;getExternalCacheDir;();;ReturnValue;android-external-storage-dir;manual",
+        "android.content;Context;true;getExternalCacheDirs;();;ReturnValue;android-external-storage-dir;manual",
+        "android.os;Environment;false;getExternalStorageDirectory;();;ReturnValue;android-external-storage-dir;manual",
+        "android.os;Environment;false;getExternalStoragePublicDirectory;(String);;ReturnValue;android-external-storage-dir;manual",
       ]
   }
 }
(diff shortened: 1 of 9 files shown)


java/ql/lib/semmle/code/java/frameworks/android/ExternalStorage.qll
```diff
@@ -0,0 +1,50 @@
...
+ * This is controlable by third-party applications, so is treated as a remote flow source.
+ */
+predicate androidExternalStorageSource(DataFlow::Node n) {
+  exists(ConstructorCall fInp, DataFlow::Node externalDir |
+    fInp.getConstructedType().hasQualifiedName("java.io", "FileInputStream") and
+    n.asExpr() = fInp and
+    sourceNode(externalDir, "android-external-storage-dir") and
+    externalStorageFlow(externalDir, DataFlow::exprNode(fInp.getArgument(0)))
+  )
+}
```
atorralba, line 50: I think all this should be handled by the dataflow library itself, rather than creating a source that depends on dataflow, which could create recursion.

If I'm not mistaken, the only thing that this is adding that doesn't already exist as a dataflow step or taintflow step is the step from `FileInputStream`'s constructor argument to its return value. I tried adding a row to `JavaIoSummaryCsv` like the following:

```
"java.io;FileInputStream;true;FileInputStream;;;Argument[0];Argument[-1];taint",
```

And then I could remove everything in `ExternalStorage.qll` except the source CSV rows without breaking any tests.

Is there any other reason why you added this `externalStorageFlow` logic?

joefarebrother: The issue with this approach is that it could cause false positives in other queries. Any flow to a `File` object, which may be a result of the `PathInjection` query but not if it's sanitized for that query, that is read by a `FileInputStream` would be tainted.

atorralba: Isn't this something that queries themselves should handle though? If it's a query interested in paths, it should include path sanitizers in its configuration, which will prune the `src` -> `sanitizer` -> `File` -> `FileInputStream` flow. If it doesn't care about paths, but about otherwise tainted `File` objects (even if I can't come up with an example of that right now), it won't have the sanitizer in the configuration and the flow will go through as desired.

Does that make sense?

atorralba: Dumb example: a filename getting sanitized to remove `..` characters is used to read a file and print out its contents in an HTML page. In the XSS query we don't care against path sanitizers, but rather XSS sanitizers, so in that case we want the flow to go through from the external storage to the XSS sink even if `..`s are removed from the filename.

joefarebrother: However a query that doesn't care about tainted `File`s (and doesn't implement related sanitizers) can still be affected by the `File` -> `FileInputStream` step; as currently there are steps that imply that a `File` is tainted when it's path is tainted. 

For example, in somthing like the following:
```java
String inp = source();
File f = new File("/some/trusted/path/" + inp);
if (!isValid(f)) 
  throw new Exception();
InputStream is = FileInputStream(f);
byte[] data = new byte[1024];
is.read(data);
sink(data);
```
some query unrelated to `File`s will get new results.

joefarebrother: An approach I've just thought of which could work is to have these steps go trough a synthetic field `FileContents`; as to not intefere with other tainted `Flie`s. However I believe that source (and sink) flow models don't currently support the synthetic steps from access paths that steps do.

joefarebrother: Hmm; is it really the case that the external storage methods should return `File`s have tainted filenames? I'd assume their filenames are controlled by the system.

atorralba: 🤦‍♂️ You're right, my example was confusing. The directory returned by `getExternaFilesDir` et al is not externally-controllable. What I was thinking was that any file in that directory _can_ have an arbitrary name and contents, but that's not quite what we're discussing, so generally tainting the sources doesn't work.

Ok, I think I'm convinced that using local flow is the only way to do it, since we'd need to be able to partially taint sources depending on the access path to implement the precise flow solution through synthetic fields.

Thanks for walking me through this and sorry for the back-and-forth.

java/ql/lib/semmle/code/java/frameworks/android/ExternalStorage.qll
```diff
@@ -0,0 +1,50 @@
...
+ * Holds if `n` is a node that reads the contents of an external file in Android.
+ * This is controlable by third-party applications, so is treated as a remote flow source.
+ */
+predicate androidExternalStorageSource(DataFlow::Node n) {
+  exists(ConstructorCall fInp, DataFlow::Node externalDir |
+    fInp.getConstructedType().hasQualifiedName("java.io", "FileInputStream") and
+    n.asExpr() = fInp and
+    sourceNode(externalDir, "android-external-storage-dir") and
+    externalStorageFlow(externalDir, DataFlow::exprNode(fInp.getArgument(0)))
+  )
```
atorralba, line 49: I'd suggest using `FileReadExpr` from `FileReadWrite.qll` instead of only the `FileInputStream` constructor to catch more ways of reading files.

```suggestion
predicate androidExternalStorageSource(DataFlow::Node n) {
  exists(DataFlow::Node externalDir |
    sourceNode(externalDir, "android-external-storage-dir") and
    externalStorageFlow(externalDir, DataFlow::exprNode(n.asExpr().(FileReadExpr).getFileVarAccess()))
  )
```

Although I think we'd need to change the `FileReadExpr::getFileVarAccess` predicate to be something like `getFileExpr`, to not restrict ourselves to `VarAccess`es. That impacts the `ReadingFromWorldWritableFile.ql` query, so we can handle that in a follow-up PR if you prefer.

joefarebrother: It looks like `FileReadExpr` includes nested calls, which don't exactly correspond to dataflow and thus could cause FPs. 
For example, in the following: 
```java
String foo(File f) {
  this.is = FileInputStream(f);
  return "something else";
}

void bar(Contex ctx) {
  String s = foo(new File(ctx.getExternalFilesDir(null), "file.txt"));
  sink(s);
}
```
then the call to `foo` is considered a `FileReadExpr` but should not be tainted.

Instead perhaps we could define a variant of `FileReadExpr` that doesn't look through nested calls.

atorralba: Good catch. Yes, my main concern was not repeating a list of methods that read from files if possible.

Maybe the current `FileReadExpr` (which I think needs a name change to indicate its recursive nature) could build upon the new version, which wouldn't look through nested calls and could be reused by other libraries interested in file-reading methods.

java/ql/lib/semmle/code/java/security/FileReadWrite.qll
```diff
@@ -1,9 +1,9 @@
 import java
 
 /**
- * Holds if `fileAccess` is used in the `fileReadingExpr` to read the represented file.
+ * Holds if `fileAccess` is directly used in the `fileReadingExpr` to read the represented file.
  */
-private predicate fileRead(VarAccess fileAccess, Expr fileReadingExpr) {
+predicate directFileRead(Expr fileAccess, Expr fileReadingExpr) {
```
atorralba, line 6: This can remain `private` right?

java/ql/lib/semmle/code/java/frameworks/android/ExternalStorage.qll
```diff
@@ -0,0 +1,50 @@
...
+  node2.asExpr().(FieldRead).getField().getInitializer() = node1.asExpr()
+}
+
+private predicate externalStorageFlow(DataFlow::Node node1, DataFlow::Node node2) {
+  externalStorageFlowStep*(node1, node2)
+}
+
+/**
+ * Holds if `n` is a node that reads the contents of an external file in Android.
+ * This is controlable by third-party applications, so is treated as a remote flow source.
```
atorralba, line 42:
```suggestion
 * This is controllable by third-party applications, so is treated as a remote flow source.
```

java/ql/lib/semmle/code/java/security/FileReadWrite.qll
```diff
@@ -49,3 +56,15 @@ class FileReadExpr extends Expr {
...
 }
+
+/**
+ * An expression that directly reads from a file and returns its contents.
+ */
+class DirectFileReadExpr extends Expr {
+  DirectFileReadExpr() { directFileRead(_, this) }
+
+  /**
+   * Gets the `Expr` representing the file that is read
```
atorralba, line 67:
```suggestion
   * Gets the `Expr` representing the file that is read.
```

atorralba (commented):
A few more comments, but otherwise LGTM. I'd like to see a DCA run on all projects (not just Android) so that we see if this has any impact on performance.

java/ql/lib/semmle/code/java/frameworks/android/ExternalStorage.qll
```diff
@@ -0,0 +1,50 @@
...
+private import semmle.code.java.security.FileReadWrite
+private import semmle.code.java.dataflow.DataFlow
+private import semmle.code.java.dataflow.ExternalFlow
+
+private class ExternalStorageDirSourceModel extends SourceModelCsv {
+  override predicate row(string row) {
+    row =
+      [
+        //"package;type;overrides;name;signature;ext;spec;kind"
+        "android.content;Context;true;getExternalFilesDir;(String);;ReturnValue;android-external-storage-dir",
```
michaelnebel, line 13: Please note that the content of this PR: https://github.com/github/codeql/pull/9025
changes the format of the modles. You need to postfix fix ";manual"

atorralba (approved):

PR #9213, opened 2022-05-19
Kotlin: fix implementation of SAM classes that inherit their abstract method

smowton:
Based on / blocked by https://github.com/github/codeql/pull/9123


java/kotlin-extractor/src/main/kotlin/KotlinFileExtractor.kt
```diff
@@ -654,9 +654,13 @@ open class KotlinFileExtractor(
...
-        if (isFake(f)) return null
+    fun extractFunction(f: IrFunction, parentId: Label<out DbReftype>, extractBody: Boolean, extractMethodAndParameterTypeAccesses: Boolean, typeSubstitution: TypeSubstitution?, classTypeArgsIncludingOuterClasses: List<IrTypeArgument>?, idOverride: Label<DbMethod>? = null, locOverride: Label<DbLocation>? = null) =
+        if (isFake(f))
+            null
+        else
+            extractNonFakeFunction(f, parentId, extractBody, extractMethodAndParameterTypeAccesses, typeSubstitution, classTypeArgsIncludingOuterClasses, idOverride, locOverride)
 
+    fun extractNonFakeFunction(f: IrFunction, parentId: Label<out DbReftype>, extractBody: Boolean, extractMethodAndParameterTypeAccesses: Boolean, typeSubstitution: TypeSubstitution?, classTypeArgsIncludingOuterClasses: List<IrTypeArgument>?, idOverride: Label<DbMethod>? = null, locOverride: Label<DbLocation>? = null): Label<out DbCallable>  {
```
tamasvajk, line 663: `extractNonFakeFunction` is actually extracting fake and non-fake functions too, so its name is misleading. I think `forceExtractFunction` is somewhat better. Another option would be adding a Boolean parameter to `extractFunction` to skip the `isFake` test.

tamasvajk (approved):

PR #9214, opened 2022-05-19
Data flow: Do not discard call context when computing reverse lambda flow through jumps

hvitved:
Since we are computing flow backwards, and we are recording the *first* call seen during reverse flow, it is valid to keep the call context even when we go through a jump step.


aschackmull (approved):
LGTM. Nice fix.

PR #9215, opened 2022-05-19
Ruby: Fixes for `Argument[any,any-named]` in MaD

RasmusWL:
This PR does 2 things:

1. Removes `self` from `Argument[any]` for flow-summaries, inspired by the internal discussion that made us do the same for Python in https://github.com/github/codeql/pull/8883 -- this is used for the input/output componenets of the summary table in MaD
2. Adds support for `Argument[any/any-named]` and `Parameter[any/any-named]` in the `path` component in MaD

I recommend viewing this PR commit by commit.

I noted that `getLabelFromArgumentPosition`/`getLabelFromParameterPosition` in ApiGraphs.qll did not support if the argument/parameter position represented `self`. For this PR, I did not go into adding that support. I'm not in a position to figure out whether that's important, and I'm already fairly far away from what I had hoped to be doing today :smile:

_I also realize now that I did not add any tests of `Parameter[any/any-named]`_ :grimacing:

### DIFF
--- a/ruby/ql/lib/codeql/ruby/ApiGraphs.qll
+++ b/ruby/ql/lib/codeql/ruby/ApiGraphs.qll
@@ -793,13 +793,11 @@ module API {
       or
       pos.isAnyNamed() and
       result = Label::keywordParameter(_)
+      //
       // Note: there is currently no API graph label for `self`.
       // It was omitted since in practice it means going back to where you came from.
       // For example, `base.getMethod("foo").getSelf()` would just be `base`.
       // However, it's possible we'll need it later, for identifying `self` parameters or post-update nodes.
-      // or
-      // pos.isSelf() and
-      // ...
     }
 
     /** Gets the API graph label corresponding to the given parameter position. */
@@ -829,10 +827,11 @@ module API {
       or
       pos.isAnyNamed() and
       result = Label::keywordParameter(_)
-      // TODO: needs handling of `self` ParameterPosition
-      // or
-      // pos.isSelf() and
-      // ...
+      //
+      // Note: there is currently no API graph label for `self`.
+      // It was omitted since in practice it means going back to where you came from.
+      // For example, `base.getMethod("foo").getSelf()` would just be `base`.
+      // However, it's possible we'll need it later, for identifying `self` parameters or post-update nodes.
     }
   }
 }


ruby/ql/lib/codeql/ruby/ApiGraphs.qll
```diff
@@ -780,6 +780,20 @@ module API {
...
+      pos.isAny() and
+      (
+        result = Label::parameter(_)
+        or
+        result = Label::keywordParameter(_)
+        or
+        result = Label::blockParameter()
+        // NOTE: `self` should NOT be included, as described in the QLDoc for `isAny()`
+      )
+      // TODO: needs handling of `self` ArgumentPosition
```
asgerf, line 793: FWIW it was deliberate that I didn't include `self` here. I didn't see a use-case for it in Ruby API graphs, since in practice it means going back to where you came from. E.g. `base.getMethod("foo").getSelf()` would just be `base`. It's possible we'll need it for marking the post-update node of `self` as a taint source, I'm not sure.

(JS is different since any value can be passed as `this`, and some libraries take advantage of this)

ruby/ql/test/library-tests/dataflow/summaries/summaries.rb
```diff
@@ -82,4 +85,22 @@ def userDefinedFunction(x, y)
...
 sink(a[0])
 sink(a[1])
-sink(a[2]) # $ hasValueFlow=elem2
\ No newline at end of file
+sink(a[2]) # $ hasValueFlow=elem2
+
+x = Foo.new
+y = []
+z = []
+x.flowToAnyArg(tainted, y, key: z)
```
asgerf, line 93: Could you add similar calls to `sink()` before the call to `flowToAnyArg`?

asgerf (commented):
Thanks! I don't immediately understand why the argument/parameter positions needed to be flipped, though maybe @hvitved can tell.

ruby/ql/test/library-tests/dataflow/summaries/summaries.rb
RasmusWL: Done :+1:

ruby/ql/lib/codeql/ruby/dataflow/internal/FlowSummaryImplSpecific.qll
```diff
@@ -257,6 +257,12 @@ ArgumentPosition parseParamBody(string s) {
   or
   s = "block" and
   result.isBlock()
+  or
+  s = "any" and
+  result.isAny()
+  or
+  s = "any-named" and
+  result.isKeyword(_)
```
hvitved, line 265: We need a new `TAnyKeywordArgumentPosition` for this instead.

ruby/ql/lib/codeql/ruby/dataflow/internal/FlowSummaryImplSpecific.qll
```diff
@@ -281,4 +287,10 @@ ParameterPosition parseArgBody(string s) {
...
```
hvitved, line 295: We need a new `TAnyKeywordParameterPosition` for this instead.

RasmusWL:
> Thanks! I don't immediately understand why the argument/parameter positions needed to be flipped

Previously for `Argument[...]` it used `ParameterPosition`, and for `Parameter[...]` it used `ArgumentPosition`. So if you wrote `Argument[key:]` to try and match a keyword argument with name `key`, such as `method(key: 42)`, that would only be possible if there existed a `TKeywordParameterPosition` with name `key`, and in turn that would only exist if there was an actual `KeywordParameter` called `key` :grimacing: 

The old relevant code is

https://github.com/github/codeql/blob/d18e03cf9a5fb6b84959ceb2248ae026bafd46de/ruby/ql/lib/codeql/ruby/frameworks/data/internal/ApiGraphModelsSpecific.qll#L130-L133

https://github.com/github/codeql/blob/0879b6ae12a48c28ee27270fdd32c2ff0cf0ec30/ruby/ql/lib/codeql/ruby/dataflow/internal/DataFlowDispatch.qll#L277-L281

ruby/ql/lib/codeql/ruby/dataflow/internal/FlowSummaryImplSpecific.qll
RasmusWL: I can certainly add that :+1: Can you elaborate on why?

The current solution certainly does cause number of results for `parseParamBody` to grow needlessly, since `"any-named"` will have results for all keyword arguments. However, that is 2x at most, if all arguments are positional, which I wouldn't assume to have a major impact on performance (although it certainly doesn't help).

I guess I'm just curious whether this is to ensure we have a clean codebase, or whether I should maybe have worried more about the performance aspect.

RasmusWL:
Performance not looking very good :grimacing: A nearly 5x for the worst project so far :flushed:

ruby/ql/lib/codeql/ruby/dataflow/internal/FlowSummaryImplSpecific.qll
hvitved: It is for performance reasons: Otherwise a MaD row containing `Argument[any-named]` will expand to a row for each possible named argument, which we don't want. Instead, we push that fan-out into the inlined predicate `parameterMatch`, and let the data-flow library make sure to only apply the fan-out when needed.

cpp/ql/lib/semmle/code/cpp/dataflow/internal/DataFlowImplCommon.qll
```diff
@@ -80,8 +80,8 @@ predicate accessPathCostLimits(int apLimit, int tupleLimit) {
 pragma[noinline]
 predicate argumentPositionMatch(DataFlowCall call, ArgNode arg, ParameterPosition ppos) {
   exists(ArgumentPosition apos |
-    arg.argumentOf(call, apos) and
-    parameterMatch(ppos, apos)
+    arg.argumentOf(call, pragma[only_bind_into](apos)) and
```
hvitved, line 83: I would prefer to revert this change, and instead add `pragma[nomagic]` to the `parameterMatch` predicate.

RasmusWL: done. let's talk over later on why this is better :blush:

RasmusWL:
Performance for `pragma[nomagic]` also looks good :+1:

ruby/ql/lib/codeql/ruby/dataflow/internal/DataFlowDispatch.qll
```diff
@@ -259,7 +259,9 @@ private module Cached {
       exists(any(Call c).getKeywordArgument(name))
       or
       FlowSummaryImplSpecific::ParsePositions::isParsedKeywordParameterPosition(_, name)
-    }
+    } or
+    TAnyArgumentPosition() or
+    TAnyKeywordArgumentPosition()
```
hvitved, line 264: I am not sure why we need this, actually?

RasmusWL: In https://github.com/github/codeql/pull/9215/commits/0879b6ae12a48c28ee27270fdd32c2ff0cf0ec30#diff-3a6f42ed2c69a87189b2f6c85dae0a2162031d5b4371d02b9d0770b446c25423 i make it such that Argument positions are used for `Argument[...]` parsing for path-components in MaD, and Parameter positions are used fro `Parameter[...]` parsing.

Since we want to allow you to write both `Argument[any/any-named]` and `Parameter[any/any-named]`, we also need both any/any-named positions for both parameters and arguments

RasmusWL:
Fixed the merge conflict, in regards to hash splat handling. Performance was good before this, so I assume that it still is.

nickrolfe:
Is this ready for a re-review? Calum asked me to take a look, but it looks like there are a couple of unresolved comments from Asger and Tom.

ruby/ql/lib/codeql/ruby/ApiGraphs.qll
RasmusWL:
```suggestion
      // Note: it is deliberate that `self` ArgumentPosition is not included here.
      // I (@asgerf) didn't see a use-case for it in Ruby API graphs, since in practice
      // it means going back to where you came from. E.g. base.getMethod("foo").getSelf() 
      // would just be base. It's possible we'll need it for marking the post-update node 
      // of self as a taint source, I'm not sure.
```

RasmusWL:
Thanks @nickrolfe, a review from you would be good. We didn't manage it before @hvitved went OoO, but we did agree (in private DM) that it could be merged as-is, and the fixup requested by him could be done later.

I've made a suggestion for the concerns raised by `@asgerf`, hope you're in a position where you're comfortable with accepting it :crossed_fingers: 

(newest performance test still looks good)

ruby/ql/lib/codeql/ruby/ApiGraphs.qll
asgerf: Sorry, there's no need to change this, I just wanted to explain why `self` isn't handled. The above isn't a style of comment I'd recommend inserting in the codebase.

nickrolfe (dismissed):
I don't really feel qualified to review this properly, but, Tom said he was happy, so I'm happy, and the latest comment change looks pretty reasonable.

ruby/ql/lib/codeql/ruby/ApiGraphs.qll
RasmusWL: Why not? I think it's useful context for the next person looking at this code, and certainly more informative than the TODO I have left right now.

asgerf: OK then, I've tried to reformulate it to something I'd be willing to commit:
```suggestion
      // Note: there is currently no API graph label for `self`.
      // It was omitted since in practice it means going back to where you came from.
      // For example, `base.getMethod("foo").getSelf()` would just be `base`.
      // However, it's possible we'll need it later, for identifying `self` parameters or post-update nodes.
```

RasmusWL:
With the comment changes merged, I think this is good to go :rocket:

nickrolfe (approved):
Thanks!

PR #9217, opened 2022-05-19
Kotlin: Tweak logging

igfoo:
Makes it easier to filter out the peak memory info


smowton (approved):
We have truly achieved... peak memory.

PR #9220, opened 2022-05-19
Associate certain companion object fields with the parent class

smowton:
Specifically `const`, `lateinit` and `@JvmField` properties get a static field which belongs to the containing class not the companion object, such that Java can address them via the containing class name rather than have to navigate a companion object pointer.

### DIFF
--- a/java/kotlin-extractor/src/main/kotlin/KotlinFileExtractor.kt
+++ b/java/kotlin-extractor/src/main/kotlin/KotlinFileExtractor.kt
@@ -135,7 +135,7 @@ open class KotlinFileExtractor(
                     Unit
                 }
                 is IrField -> {
-                    val parentId = useDeclarationParent(declaration.parent, false)?.cast<DbReftype>()
+                    val parentId = useDeclarationParent(getFieldParent(declaration), false)?.cast<DbReftype>()
                     if (parentId != null) {
                         extractField(declaration, parentId)
                     }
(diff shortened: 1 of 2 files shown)


tamasvajk (commented):
Looks plausible to me, could you add tests that cover the new cases?

java/kotlin-extractor/src/main/kotlin/KotlinFileExtractor.kt
```diff
@@ -826,10 +826,13 @@ open class KotlinFileExtractor(
                 }
 
                 if (bf != null && extractBackingField) {
-                    val fieldId = extractField(bf, parentId)
-                    tw.writeKtPropertyBackingFields(id, fieldId)
-                    if (p.isDelegated) {
-                        tw.writeKtPropertyDelegates(id, fieldId)
+                    val fieldParentId = useDeclarationParent(getFieldParent(bf), false)
+                    if (fieldParentId != null) {
```
igfoo, line 830: Should we give an error in the null case?

java/kotlin-extractor/src/main/kotlin/KotlinUsesExtractor.kt
```diff
@@ -1255,9 +1256,22 @@ open class KotlinUsesExtractor(
...
+        f.parentClassOrNull?.let {
+            if (it.isCompanion && isDirectlyExposedCompanionObjectField(f))
+                it.parent
+            else
+                null
+        } ?: f.parent
 
     fun getFieldLabel(f: IrField): String {
-        val parentId = useDeclarationParent(f.parent, false)
+        val parentId = useDeclarationParent(getFieldParent(f), false)
```
igfoo, line 1274: Shouldn't we be checking that `parentId` isn't null?

java/kotlin-extractor/src/main/kotlin/KotlinFileExtractor.kt
smowton: We already get one from `useDeclarationParent`

java/kotlin-extractor/src/main/kotlin/KotlinUsesExtractor.kt
igfoo: Let's leave it for a bigger cleanup, then

igfoo (approved):

PR #9229, opened 2022-05-20
Kotlin: extract non-private members of class supertypes

tamasvajk:


java/kotlin-extractor/src/main/kotlin/KotlinFileExtractor.kt
```diff
@@ -404,8 +404,7 @@ open class KotlinFileExtractor(
                 }
 
                 extractClassModifiers(c, id)
-                val forceExtractSupertypeMembers = !isExternalDeclaration(c)
-                extractClassSupertypes(c, id, inReceiverContext = forceExtractSupertypeMembers)
+                extractClassSupertypes(c, id, inReceiverContext = true)
```
igfoo, line 407: Would it be worth adding a comment giving the rationale for `inReceiverContext = true` here?

igfoo (approved):

PR #9230, opened 2022-05-20
Swift: use structured generated C++ classes in `DeclVisitor`

redsun82:
This PR introduces using structured generated C++ classes in `DeclVisitor`.

The coding style moves to translation methods taking in a swift entity and filling a trap class instance, a bit in the style of gRPC method implementations. Shortcuts are provided to fill optional and repeated properties (`fetchOptionalLabel` and `fetchRepeatedLabels`). No explicit mention of trap entries is needed any more, but they could still be used alongside this higher-level interface.

Also, the `Binding` template class has been removed and aggregate initialization of generated C++ classes disallowed.

`Binding`  was meant to allow aggregate initialization of generated C++ entries having the label `id` as first argument, before fields from superclasses.

As aggregate initialization turned out to be undesirable (names of fields are not explicit, and `{}` must be inserted for empty
superclasses), `Binding` is removed it and aggregate initialization are disallowed by defining empty constructors for generated
classes.


swift/extractor/visitors/DeclVisitor.h
```diff
@@ -9,247 +9,177 @@
...
-// TODO: maybe make the above chain of calls automatic with a bit of macro metaprogramming?
 class DeclVisitor : public AstVisitorBase<DeclVisitor> {
  public:
-  using AstVisitorBase<DeclVisitor>::AstVisitorBase;
-
-  void visitFuncDecl(swift::FuncDecl* decl) {
-    auto label = dispatcher_.assignNewLabel(decl);
-    dispatcher_.emit(ConcreteFuncDeclsTrap{label});
-    emitAbstractFunctionDecl(decl, label);
+  void translateFuncDecl(swift::FuncDecl* decl, ConcreteFuncDecl* entry) {
```
AlexDenisov, line 14: As a nit, would it make sense to add namespace to the second arguments as well? This way it could be easier to see that we are translating from `swift::Something` to `codeql::Something`.

AlexDenisov (commented):
Love it! ❤️

swift/extractor/visitors/DeclVisitor.h
redsun82: good point!

I was also thinking about the possible name collisions with using the name as is for entries, and thinking about using a more specific prefix like `ConcreteFuncDeclTrap` (which however might be confused with plain single entry traps) or `ConcreteFuncDeclTrapClass`. Makes for longer names, but could save us from possible collisions, and might look clearer in code. Yet another possibility would be to not add any suffix but use two separate namespaces for our application code written by hand (`codeql`) and the generated one (`codeqlgen` or `codeql_gen`).

redsun82:
I made the PR a bit smaller than before and added some comments to the new functions.

With respect to the previous version of this PR, I undid the `is_unknown` move. It's still kind of an issue, as there will be repeated `is_unknown` fields in the classes layouts, but it does not prevent the current PR from working. There can be future changes to address that (of which I prefer the possibility to optionally skip C++ generation for some optional or repeated fields).

I also moved `fetchOptionalLabel` and `fetchRepeatedLabels` to the dispatcher, as it was unclear yet what belonged to the dispatcher and what to `VisitorBase`. Maybe we can reorganize the classes later to move all methods relevant to the visitors to `VisitorBase`.

swift/extractor/visitors/VisitorBase.h
```diff
@@ -21,14 +21,34 @@ class VisitorBase {
...
 // superclasses by default and instead provide our own TBD default (using the exact type)
-#define DEFAULT(KIND, CLASS, PARENT) \
-  void visit##CLASS##KIND(swift::CLASS##KIND* e) { dispatcher_.emitUnknown(e); }
+// if we the implementation class has translate##CLASS##KIND (that uses generated C++ classes), we
+// want to use that. We detect that by comparing the member function pointer to the one for a
+// private undefined member function of the same name and signature in {Ast,Type}VisitorBase, which
+// will be shadowed (and thus different) in case the implementation class has it.
+#define DEFAULT(KIND, CLASS, PARENT)                                              \
+ public:                                                                          \
+  void visit##CLASS##KIND(swift::CLASS##KIND* e) {                                \
```
sashabu, line 30: Even with the comment (which is very helpful!), this makes me a bit uneasy: can we actually compare pointers to members of different types? even if they have different access? will they definitely never compare equal, or does the base one not being defined give the implementation some leeway that we're not expecting?

I can live with it, but would suggest thinking about alternatives:
1. Instead of `translate##CLASS##KIND` taking an output parameter (see also my comment on the declaration), can it return the TRAP class instead? If yes, the base implementation can return `void` (or some other sentinel - even a runtime sentinel like `nullptr` or `nullopt`), and we can check for that here. That also has the potentially useful property of subclasses being able to return void when they want to do something but not emit TRAP.
2. Can we add a base implementation that does the "right" thing (i.e. populates a TrapClassOf that causes `emit` to act as `emitUnknown`)?
3. If we keep the check, it might be more obviously correct with SFINAE or `std::is_invocable`?

swift/extractor/visitors/VisitorBase.h
```diff
@@ -21,14 +21,34 @@ class VisitorBase {
...
+      TrapClassOf<swift::CLASS##KIND> entry{};                                    \
+      static_cast<CrtpSubclass*>(this)->translate##CLASS##KIND(e, &entry);        \
+      dispatcher_.emit(entry);                                                    \
+    } else {                                                                      \
+      dispatcher_.emitUnknown(e);                                                 \
+    }                                                                             \
+  }                                                                               \
+                                                                                  \
+ private:                                                                         \
+  void translate##CLASS##KIND(swift::CLASS##KIND*, TrapClassOf<swift::CLASS##KIND>*);
```
sashabu, line 43: Instead of taking an output parameter, can we use the return value for the TRAP class instead? See https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rf-out.

If not (presumably because it's also an input), we should document a contract: can it be null? any other requirements on it on entry? what happens to the input contents on success? on failure? etc. Should it be a reference instead of a pointer? Implementations seem to assume it's not null.

sashabu (commented):
I've reviewed the "interesting" parts of this PR, but:
 1. I haven't done a thorough review of `DeclVisitor.h` since it looks fairly mechanical and
 2. I'm not familiar with Mustache syntax, so I only commented on what I think the resulting C++ will look like

since Alex is the main reviewer. Let me know if you'd like a more thorough review from me.

swift/codegen/templates/cpp_classes.mustache
```diff
@@ -13,7 +13,12 @@
 namespace {{namespace}} {
 {{#classes}}
 
-struct {{name}}{{#final}} : Binding<{{name}}Tag>{{#bases}}, {{ref.name}}{{/bases}}{{/final}}{{^final}}{{#has_bases}}: {{#bases}}{{^first}}, {{/first}}{{ref.name}}{{/bases}}{{/has_bases}}{{/final}} {
+struct {{name}}{{#has_bases}} : {{#bases}}{{^first}}, {{/first}}{{ref.name}}{{/bases}}{{/has_bases}} {
+  {{name}}() {}
```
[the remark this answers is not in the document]
redsun82: makes sense!

swift/extractor/SwiftDispatcher.h
```diff
@@ -120,6 +120,27 @@ class SwiftDispatcher {
...
+      -> std::optional<decltype(fetchLabel(std::forward<Arg>(arg)))> {
+    if (arg) {
+      return fetchLabel(std::forward<Arg>(arg));
+    }
+    return std::nullopt;
+  }
+
+  // map `fetchLabel` on the iterable `arg`, returning a vector of all labels
+  template <typename Iterable>
+  auto fetchRepeatedLabels(Iterable&& arg) -> std::vector<decltype(fetchLabel(*arg.begin()))> {
```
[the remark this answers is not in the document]
redsun82: good point!

swift/extractor/SwiftDispatcher.h
```diff
@@ -120,6 +120,27 @@ class SwiftDispatcher {
                    locatableLabel);
   }
 
+  // return `std::optional(fetchLabel(arg))` if arg converts to true, otherwise std::nullopt
+  template <typename Arg>
+  auto fetchOptionalLabel(Arg&& arg)
+      -> std::optional<decltype(fetchLabel(std::forward<Arg>(arg)))> {
+    if (arg) {
+      return fetchLabel(std::forward<Arg>(arg));
```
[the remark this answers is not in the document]
redsun82: I'll put perfect forwarding in `fetchRepeatedLabels` as well. The use of universal references there has to deal more with catching both temporary objects and non-const references (we have both cases) than with forwarding, but it's true that it's better to be consistent

swift/extractor/visitors/VisitorBase.h
redsun82: good point. The pointer was mainly for consistency with the pointer to the swift entity, but thinking about it, as a new "visiting protocol" is provided here, we don't need that either. I've changed the translation signatures to a more semantically correct `codeql::XXXDecl translateXXX(const swift::XXXDecl& decl)`

swift/extractor/SwiftDispatcher.h
```diff
@@ -83,6 +83,11 @@ class SwiftDispatcher {
     return label;
   }
 
+  template <typename E, typename = std::enable_if_t<!std::is_pointer_v<E>>>
```
sashabu, line 86: Nit: We can still call this by specifying the SFINAE type argument:
```
int* foo = nullptr;
assignNewLabel<int*, void>(foo)
```

See https://gcc.godbolt.org/z/borzbbxfG for more examples.

One way to get around that is to use the `enable_if_t` in a non-type template parameter, so the caller only has control of the value, not the type (which then fails to substitute regardless of the user-specified value):

```
template <typename E, std::enable_if_t<!std::is_pointer_v<E>>* = nullptr>
```

sashabu (commented):
LGTM with the same caveats that:
1. I haven't done a thorough review of DeclVisitor.h since it looks fairly mechanical and
2. I'm not familiar with Mustache syntax, so I only commented on what I think the resulting C++ will look like

swift/extractor/SwiftDispatcher.h
sashabu: I don't like that as much because it makes the return type less obvious. The alternative of putting it in the template parameter list makes it less obvious what the "real" template parameters are, but I think that's less important? Happy to concede that it's just a matter of taste.

Really what we need is concepts, but given that even our C++17 support is incomplete, that might be a stretch...

redsun82: yeah, I wish we had concepts as well, we could just write `template<not_a_pointer E>` with the appropriate concept definition, and it would be readable and provide meaningful error messages when needed.

I'll go with the non-type parameter trick :+1:

redsun82:
@AlexDenisov I've merged your deduplication branch into this and selected it as a target. Once we merge that, I can rebase this and put it back to review.

AlexDenisov (approved):

PR #8606, opened 2022-03-30
JS/Python/Ruby: Document how API graphs should be interpreted

asgerf:
I've written what I find to be a useful interpretation of API graphs.

I'm hoping we can use this PR to discuss and align on what is the "correct" way to interpret API graphs. Much of the text is language-agnostic so I'm hoping to port it to Python and Ruby as well, after we've iterated on it a bit.

I should mentioned that the example with the getter at the end doesn't work in the current JS implementation. Support for getters is in flight [here](https://github.com/github/codeql/pull/8602) -- should probably be merged first, so at not to make incorrect clams in the documentation.

I haven't renamed anything in this PR. Let's settle on the text before going through with renaming.

PR #9234, opened 2022-05-20
JS: API graph support for accessors (and classes members)

asgerf:
Generates API graph edges in two cases:
- When an object escapes, generates `getMember` edges to the return-value of getters.
- When an instance of a class escapes, generates `getMember` edges to the instance members of the class (including getters, as above).
  - Previously, such edges were generated only when the class itself escapes, not when an _instance_ of the class escapes.

Makes the example in https://github.com/github/codeql/pull/8606 actually work in JS.

[Evaluation](https://github.com/github/codeql-dca-main/tree/data/asgerf/api-graph-acces__default__code-scanning__9/reports) looks reasonable.

### DIFF
--- a/javascript/ql/test/ApiGraphs/accessors/index.js
+++ b/javascript/ql/test/ApiGraphs/accessors/index.js
@@ -1,28 +1,28 @@
 const foo = require('foo');
 
 foo({
-    myMethod(x) { /* use (parameter 0 (member myMethod (parameter 0 (member exports (module foo))))) */
+    myMethod(x) { /* use=moduleImport("foo").getMember("exports").getParameter(0).getMember("myMethod").getParameter(0) */
         console.log(x);
     }
 });
 
 foo({
     get myMethod() {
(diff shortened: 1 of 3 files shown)


javascript/ql/test/ApiGraphs/accessors/index.js
```diff
@@ -0,0 +1,65 @@
+const foo = require('foo');
+
+foo({
+    myMethod(x) { /* use (parameter 0 (member myMethod (parameter 0 (member exports (module foo))))) */
```
erik-krogh, line 4:
```suggestion
    myMethod(x) { /* use=moduleImport("foo").getMember("exports").getParameter(0).getMember("myMethod").getParameter(0) */
```

[I recently changed the testing syntax](https://github.com/github/codeql/pull/8578).  
All your tests need to be rewritten to the new syntax, because currently they don't test anything.

erik-krogh (approved):

PR #9235, opened 2022-05-20
JS: Update the extractor to use TypeScript 4.7

kaeluka:
_Erik here:_

A plain update to TypeScript 4.7 without surprises.  

There is not much special happening, but TypeScript now supports `.mts` and `.cts` files, so we had to add those to the extractor.  

Tests have been added that we can handle all the features mentioned in [the change-notes](https://devblogs.microsoft.com/typescript/announcing-typescript-4-7/).  

I can't test it locally, but the tests passed on the release candidate, so I highly expect them to pass on this stable release.   
The tests will fail here until the internal PR has been merged.


javascript/extractor/lib/typescript/package.json
```diff
@@ -2,7 +2,7 @@
     "name": "typescript-parser-wrapper",
     "private": true,
     "dependencies": {
-        "typescript": "4.6.2"
+        "typescript": "4.7.1-rc"
```
kaeluka, line 5: this will be updated to 4.7.2 or something like that

javascript/extractor/lib/typescript/yarn.lock
```diff
@@ -6,7 +6,7 @@
   version "12.7.11"
   resolved node-12.7.11.tgz#be879b52031cfb5d295b047f5462d8ef1a716446
 
-typescript@4.6.2:
-  version "4.6.2"
-  resolved "https://registry.yarnpkg.com/typescript/-/typescript-4.6.2.tgz#fe12d2727b708f4eef40f51598b3398baa9611d4"
-  integrity sha512-HM/hFigTBHZhLXshn9sN37H085+hQGeJHJ/X7LpBWLID/fbc2acUMfU+lGD98X81sKP+pFa9f0DZmCwB9GnbAg==
+typescript@4.7.1-rc:
+  version "4.7.1-rc"
```
kaeluka, line 10: will be updated as well

erik-krogh (commented):
I don't see a test/implementation of `moduleSuffixes`.  
It can wait until after this PR has been merged, but we should consider it.  
https://devblogs.microsoft.com/typescript/announcing-typescript-4-7-rc/#resolution-customization-with-modulesuffixes 

It probably requires some changes in `NodeModuleResolutionImpl.qll` around the `tryExtensions` predicate.  

---- 

Otherwise LGTM, with two small comments from a quick look.

kaeluka:
I've now added a test for the moduleSuffixes feature as well!

erik-krogh:
> I've now added a test for the moduleSuffixes feature as well!

You've tested that TypeScript can figure out the types, but not that our own module resolution can figure out the imports.  
(See `NodeModuleResolutionImpl.qll`).  
I can wait until after this PR has been merged.

kaeluka:
Yes, that's what I did :-) i know there's still work left after merging :+1:

erik-krogh:
I've bumped the version to the stable release that just happened.  
It would be nice to get someone else to review it.

javascript/ql/test/library-tests/TypeScript/Types/tst.ts
```diff
@@ -293,4 +293,83 @@ module TS46 {
...
+): void {};
+
+f({
+  produce: n => n, // <- (n: string) => string
+  consume: x => x.toLowerCase()
+});
+
+///////////
+
+const ErrorMap = Map<string, Error>;
```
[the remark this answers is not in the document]
erik-krogh: The `Map<string,Error>` is an `ExpressionWithTypeArguments` from `TypeScript.qll`.  

But we did not show that in the printAst output, so I fixed that.

javascript/ql/test/library-tests/TypeScript/Types/tst.ts
```diff
@@ -293,4 +293,83 @@ module TS46 {
...
+
+const errorMap = new ErrorMap(); // <- Map<string, Error>
+
+////////////
+
+type FirstString<T> =
+  T extends [infer S extends string, ...unknown[]]
```
asgerf, line 332: Do we extract the `extends string` part?

Usually we've tried to extract enough of the AST to prevent spurious unused variable/import alerts. Since the `extends string` part could instead contain a reference to an imported name, we'd want to extract it.

The unused variable alert isn't that important anymore, but I still think we should try to maintain it.

erik-krogh: Yes we extract it, and I've added a test that it works with our unused variable query.

erik-krogh:
> Thanks for addressing my questions @erik-krogh!
> 
> Everything LGTM, but could you please run an evaluation before merging?

[Evaluation was uneventful](https://github.com/github/codeql-dca-main/tree/data/erik-krogh/pr-9235-62fd3fd__nightly-old__code-scanning/reports).  
We extract one more LOC in Babel. I suppose that comes from a TS test-case.

asgerf (approved):
Great! LGTM

PR #9257, opened 2022-05-22
Java: Update commons-io SHA for model regeneration and update models.

michaelnebel:
In this PR we 
- Update the SHA for java/commons-io in the regenerate models workflow as a database couldn't be created anymore based on the old SHA.
- Exclude certain sinks from being generated. More specifically regex-use and write-file have been excluded. For technical reasons we need to exclude the latter to avoid derived models of Writer.[write|append]. For some reason, we can't (for java) just define models for the relevant subtypes of Writer for methods that are directly inherited from Writer. This will need further investigation, but not as a part of this PR.
- Regenerate and patch the generated models for commons-io.
- Extend the handwritten models to avoid conflict and/or incorrect re-generation of models.

@smowton @joefarebrother @atorralba : We probably need to go with the review mindset that we are aiming at not making anything (significantly) worse than it was before. The idea with the PR is simply to get the workflow running again and refresh the models with the existing logic. I have made some minor changes to cater for some of the review comments as they are relevant compared to the previously generated models.
Is that approach acceptable?


michaelnebel:
The failing integration tests are not related to the content in this PR.

java/ql/lib/semmle/code/java/frameworks/apache/IOGenerated.qll
```diff
@@ -1,684 +1,828 @@
...
+        "org.apache.commons.io;IOUtils;true;toString;(URL,String);;Argument[0];generated:open-url",
+        "org.apache.commons.io;IOUtils;true;write;(CharSequence,Writer);;Argument[0];generated:write-file",
+        "org.apache.commons.io;IOUtils;true;write;(String,Writer);;Argument[0];generated:write-file",
+        "org.apache.commons.io;IOUtils;true;write;(StringBuffer,Writer);;Argument[0];generated:write-file",
+        "org.apache.commons.io;IOUtils;true;write;(byte[],Writer);;Argument[0];generated:write-file",
+        "org.apache.commons.io;IOUtils;true;write;(byte[],Writer,Charset);;Argument[0];generated:write-file",
+        "org.apache.commons.io;IOUtils;true;write;(byte[],Writer,String);;Argument[0];generated:write-file",
+        "org.apache.commons.io;IOUtils;true;write;(char[],Writer);;Argument[0];generated:write-file",
+        "org.apache.commons.io;IOUtils;true;writeChunked;(char[],Writer);;Argument[0];generated:write-file",
+        "org.apache.commons.io;IOUtils;true;writeLines;(Collection,String,Writer);;Argument[1];generated:write-file",
```
atorralba, line 198: Strictly speaking, if we consider this a `write-file` sink, we should also consider `Argument[0].Element`, but this can't be done as of now, so I think the model generator did its job correctly :)

Something to think about for queries using `write-file` sinks though — we might need to cover this in the traditional CodeQL way (not in this PR).

java/ql/lib/semmle/code/java/frameworks/apache/IOGenerated.qll
```diff
@@ -1,684 +1,828 @@
...
 }
 
-private class IOSourcesCsv extends SourceModelCsv {
+private class IOGeneratedSourcesCsv extends SourceModelCsv {
   override predicate row(string row) {
     row =
       [
-        "org.apache.commons.io;IOUtils;true;resourceToByteArray;(String,ClassLoader);;ReturnValue;remote",
-        "org.apache.commons.io;IOUtils;true;toByteArray;(URI);;ReturnValue;remote"
+        "org.apache.commons.io.function;IOConsumer;true;forEachIndexed;(Stream,IOConsumer);;Argument[0];generated:remote",
```
atorralba, line 212: I'm not convinced these should be considered remote sources — at most they seem local (they're reading local files/resources). Are these generated this way because they return a remote source?

java/ql/lib/semmle/code/java/frameworks/apache/IOGenerated.qll
```diff
@@ -1,684 +1,828 @@
...
+        "org.apache.commons.io;IOUtils;true;read;(Reader,char[],int,int);;Argument[1];generated:write-file",
+        "org.apache.commons.io;IOUtils;true;readFully;(Reader,char[]);;Argument[0];generated:write-file",
+        "org.apache.commons.io;IOUtils;true;readFully;(Reader,char[]);;Argument[1];generated:write-file",
+        "org.apache.commons.io;IOUtils;true;readFully;(Reader,char[],int,int);;Argument[0];generated:write-file",
+        "org.apache.commons.io;IOUtils;true;readFully;(Reader,char[],int,int);;Argument[1];generated:write-file",
+        "org.apache.commons.io;IOUtils;true;skip;(Reader,long);;Argument[0];generated:write-file",
+        "org.apache.commons.io;IOUtils;true;skipFully;(Reader,long);;Argument[0];generated:write-file",
+        "org.apache.commons.io;IOUtils;true;toByteArray;(Reader);;Argument[0];generated:write-file",
+        "org.apache.commons.io;IOUtils;true;toByteArray;(Reader,Charset);;Argument[0];generated:write-file",
+        "org.apache.commons.io;IOUtils;true;toByteArray;(Reader,String);;Argument[0];generated:write-file",
```
atorralba, line 173: These don't look like `write-file` sinks, do they? Apparently they only read from a file, not write into it.

atorralba (commented):
Disclaimer: I only reviewed how the new generated models compare to the old manual ones.

Looking good! I added some comments that we might need to address before merging, specially the one regarding sources.

java/ql/lib/semmle/code/java/frameworks/apache/IOGenerated.qll
```diff
@@ -1,684 +1,828 @@
...
-        "org.apache.commons.io;IOUtils;true;toString;(URI);;Argument[0];open-url",
-        "org.apache.commons.io;IOUtils;true;toString;(URI,Charset);;Argument[0];open-url",
-        "org.apache.commons.io;IOUtils;true;toString;(URI,String);;Argument[0];open-url",
-        "org.apache.commons.io;IOUtils;true;toString;(URL);;Argument[0];open-url",
-        "org.apache.commons.io;IOUtils;true;toString;(URL,Charset);;Argument[0];open-url",
-        "org.apache.commons.io;IOUtils;true;toString;(URL,String);;Argument[0];open-url",
-        "org.apache.commons.io;RandomAccessFileMode;false;create;(File);;Argument[0];create-file",
-        "org.apache.commons.io;RandomAccessFileMode;false;create;(Path);;Argument[0];create-file",
-        "org.apache.commons.io;RandomAccessFileMode;false;create;(String);;Argument[0];create-file"
+        "org.apache.commons.io.file;PathFilter;true;accept;(Path,BasicFileAttributes);;Argument[-1];generated:regex-use[0]",
```
joefarebrother, line 13: The `regex-use` sinks that include a number are meant to convey some additional information about the predicate they're used on, in particular that the specified argument is the string that the regex is matched against. As such they should probably not be included in generated models. 

Additionally, the queries that use them only consider value flow and not taint flow; so generated models of `regex-use` sinks that included taint flow may not be correct.

java/ql/lib/semmle/code/java/frameworks/apache/IOGenerated.qll
michaelnebel: @atorralba : Thank you for all the input!!! This is really valuable!

@aschackmull : For Java model generation we explicitly declare char[] and byte[] as "bulk" types, which means that we don't generate the summary as Argument[0].Element (but instead generate Argument[0]). This is a java specialty that I left alone in the java specific part of the model generator implementation, when everything was refactored. Should we remove this Java specific part of the implementation?

atorralba: To be clear: I'm referring to `writeLines;(Collection,String,Writer);;Argument[1]`, which should also have a model for `writeLines;(Collection,String,Writer);;Argument[0].Element`,  but using access paths in sink specifications is not currently supported (afaik).

michaelnebel: Ah ok. That has not been generated. I will only focus on trying to fix issues related to incorrect models as a part of the update (basically my intention was just to refresh the models as the old SHA didn't work anymore). At least now we get one of the models 😄

java/ql/lib/semmle/code/java/frameworks/apache/IOGenerated.qll
michaelnebel: Lets just avoid creating new sinks based on regex-use then. Thank you @joefarebrother !

java/ql/lib/semmle/code/java/frameworks/apache/IOGenerated.qll
michaelnebel: I tried to make a path problem version of the create source model queries and ran it against commons-io.
It turns out that a path can be found, where the IOConsumer writes from a remote source to the Stream argument of forEachIndexed.

atorralba: What's the source? The method `IOStream.forEachIndexed` (which is called from `IOConsumer.forEachIndexed`) doesn't seem to contain any: https://github.com/apache/commons-io/blob/bcd5c9ba0cd1607b286ef0c116f119983e3f8a15/src/main/java/org/apache/commons/io/function/IOStreams.java#L39

Or is it that it exists one remote source somewhere in `commons-io` that gets passed to the first argument of `forEachIndexed`? If that's the case, it doesn't mean that the first argument of `forEachIndexed` is always a source, does it?

java/ql/lib/semmle/code/java/frameworks/apache/IOGenerated.qll
```diff
@@ -1,684 +1,828 @@
...
+        "org.apache.commons.io.file;PathFilter;true;accept;(Path,BasicFileAttributes);;Argument[0];generated:create-file",
```
smowton, line 14: This doesn't by itself create anything

java/ql/lib/semmle/code/java/frameworks/apache/IOGenerated.qll
```diff
@@ -1,684 +1,828 @@
...
+        "org.apache.commons.io.file;PathUtils;false;copyFile;(URL,Path,CopyOption[]);;Argument[0];generated:open-url",
+        "org.apache.commons.io.file;PathUtils;false;copyFile;(URL,Path,CopyOption[]);;Argument[1];generated:create-file",
+        "org.apache.commons.io.file;PathUtils;false;copyFileToDirectory;(URL,Path,CopyOption[]);;Argument[0];generated:open-url",
+        "org.apache.commons.io.file;PathUtils;false;newOutputStream;(Path,boolean);;Argument[0];generated:create-file",
+        "org.apache.commons.io.file;PathUtils;false;writeString;(Path,CharSequence,Charset,OpenOption[]);;Argument[0];generated:create-file",
+        "org.apache.commons.io.filefilter;FileFilterUtils;true;filter;(IOFileFilter,File[]);;Argument[1];generated:create-file",
+        "org.apache.commons.io.filefilter;FileFilterUtils;true;filterList;(IOFileFilter,File[]);;Argument[1];generated:create-file",
+        "org.apache.commons.io.filefilter;FileFilterUtils;true;filterSet;(IOFileFilter,File[]);;Argument[1];generated:create-file",
```
smowton, line 22: None of these by themselves create anything; if they did they should be looking at array indices of Argument[1]

java/ql/lib/semmle/code/java/frameworks/apache/IOGenerated.qll
```diff
@@ -1,684 +1,828 @@
...
+        "org.apache.commons.io.filefilter;RegexFileFilter;true;RegexFileFilter;(String);;Argument[0];generated:regex-use[]",
+        "org.apache.commons.io.filefilter;RegexFileFilter;true;RegexFileFilter;(String,IOCase);;Argument[0];generated:regex-use[]",
+        "org.apache.commons.io.filefilter;RegexFileFilter;true;RegexFileFilter;(String,int);;Argument[0];generated:regex-use[]",
+        "org.apache.commons.io.input;Tailer$Tailable;true;getRandomAccess;(String);;Argument[-1];generated:create-file",
+        "org.apache.commons.io.input;XmlStreamReader;true;XmlStreamReader;(URL);;Argument[0];generated:open-url",
+        "org.apache.commons.io.output;DeferredFileOutputStream;true;writeTo;(OutputStream);;Argument[0];generated:create-file",
```
smowton, line 28: create-file on OutputStream doesn't make much sense -- this should have been caught when the stream was created.

java/ql/lib/semmle/code/java/frameworks/apache/IOGenerated.qll
```diff
@@ -1,684 +1,828 @@
...
+        "org.apache.commons.io.output;LockableFileWriter;true;LockableFileWriter;(String);;Argument[0];generated:create-file",
+        "org.apache.commons.io.output;LockableFileWriter;true;LockableFileWriter;(String,boolean);;Argument[0];generated:create-file",
+        "org.apache.commons.io.output;LockableFileWriter;true;LockableFileWriter;(String,boolean,String);;Argument[0];generated:create-file",
+        "org.apache.commons.io.output;XmlStreamWriter;true;XmlStreamWriter;(File);;Argument[0];generated:create-file",
+        "org.apache.commons.io.output;XmlStreamWriter;true;XmlStreamWriter;(File,String);;Argument[0];generated:create-file",
+        "org.apache.commons.io.serialization;ClassNameMatcher;true;matches;(String);;Argument[-1];generated:regex-use[0]",
+        "org.apache.commons.io.serialization;ClassNameMatcher;true;matches;(String);;Argument[0];generated:regex-use[f]",
+        "org.apache.commons.io;CopyUtils;true;copy;(InputStream,Writer);;Argument[0];generated:write-file",
+        "org.apache.commons.io;CopyUtils;true;copy;(InputStream,Writer,String);;Argument[0];generated:write-file",
```
smowton, line 56: InputStreams aren't writeable

java/ql/lib/semmle/code/java/frameworks/apache/IOGenerated.qll
```diff
@@ -1,684 +1,828 @@
...
+        "org.apache.commons.io;CopyUtils;true;copy;(Reader,OutputStream);;Argument[0];generated:write-file",
+        "org.apache.commons.io;CopyUtils;true;copy;(Reader,OutputStream,String);;Argument[0];generated:write-file",
+        "org.apache.commons.io;CopyUtils;true;copy;(Reader,Writer);;Argument[0];generated:write-file",
+        "org.apache.commons.io;CopyUtils;true;copy;(String,OutputStream);;Argument[0];generated:write-file",
+        "org.apache.commons.io;CopyUtils;true;copy;(String,OutputStream,String);;Argument[0];generated:write-file",
+        "org.apache.commons.io;CopyUtils;true;copy;(String,Writer);;Argument[0];generated:write-file",
+        "org.apache.commons.io;CopyUtils;true;copy;(byte[],Writer);;Argument[0];generated:write-file",
+        "org.apache.commons.io;CopyUtils;true;copy;(byte[],Writer,String);;Argument[0];generated:write-file",
```
smowton, line 64: These all look like they should be taint-prop steps arg0 -> arg1, plus the written arg is arg1 not arg0

java/ql/lib/semmle/code/java/frameworks/apache/IOGenerated.qll
```diff
@@ -1,684 +1,828 @@
...
+        "org.apache.commons.io;FileSystemUtils;true;freeSpace;(String);;Argument[0];generated:regex-use[f]",
+        "org.apache.commons.io;FileSystemUtils;true;freeSpaceKb;(String);;Argument[0];generated:regex-use[f]",
+        "org.apache.commons.io;FileSystemUtils;true;freeSpaceKb;(String,long);;Argument[0];generated:regex-use[f]",
```
smowton, line 67: Not regexes AFAICT

java/ql/lib/semmle/code/java/frameworks/apache/IOGenerated.qll
```diff
@@ -1,684 +1,828 @@
...
+        "org.apache.commons.io;FileUtils;true;copyDirectory;(File,File);;Argument[1];generated:create-file",
+        "org.apache.commons.io;FileUtils;true;copyDirectory;(File,File,FileFilter);;Argument[1];generated:create-file",
+        "org.apache.commons.io;FileUtils;true;copyDirectory;(File,File,FileFilter,boolean);;Argument[1];generated:create-file",
+        "org.apache.commons.io;FileUtils;true;copyDirectory;(File,File,FileFilter,boolean,CopyOption[]);;Argument[1];generated:create-file",
+        "org.apache.commons.io;FileUtils;true;copyDirectory;(File,File,boolean);;Argument[1];generated:create-file",
+        "org.apache.commons.io;FileUtils;true;copyDirectoryToDirectory;(File,File);;Argument[1];generated:create-file",
```
smowton, line 73: Above in `copyFileToDirectory` we didn't treat 'copy to directory' (i.e. copy from `arg0` to `arg1/basename(arg0)`) as if they were creating arg1 -- we should treat them alike, probably including them even though `arg1` isn't the exact path created.

java/ql/lib/semmle/code/java/frameworks/apache/IOGenerated.qll
```diff
@@ -1,684 +1,828 @@
...
+        "org.apache.commons.io;FilenameUtils;true;getFullPath;(String);;Argument[0];generated:regex-use[f]",
+        "org.apache.commons.io;FilenameUtils;true;getFullPathNoEndSeparator;(String);;Argument[0];generated:regex-use[f]",
+        "org.apache.commons.io;FilenameUtils;true;getPath;(String);;Argument[0];generated:regex-use[f]",
+        "org.apache.commons.io;FilenameUtils;true;getPathNoEndSeparator;(String);;Argument[0];generated:regex-use[f]",
+        "org.apache.commons.io;FilenameUtils;true;getPrefix;(String);;Argument[0];generated:regex-use[f]",
+        "org.apache.commons.io;FilenameUtils;true;getPrefixLength;(String);;Argument[0];generated:regex-use[f]",
+        "org.apache.commons.io;FilenameUtils;true;normalize;(String);;Argument[0];generated:regex-use[f]",
+        "org.apache.commons.io;FilenameUtils;true;normalize;(String,boolean);;Argument[0];generated:regex-use[f]",
+        "org.apache.commons.io;FilenameUtils;true;normalizeNoEndSeparator;(String);;Argument[0];generated:regex-use[f]",
+        "org.apache.commons.io;FilenameUtils;true;normalizeNoEndSeparator;(String,boolean);;Argument[0];generated:regex-use[f]",
```
smowton, line 139: None of these are regexes AFAICT

java/ql/lib/semmle/code/java/frameworks/apache/IOGenerated.qll
```diff
@@ -1,684 +1,828 @@
...
+        "org.apache.commons.io;IOUtils;true;contentEquals;(Reader,Reader);;Argument[0];generated:write-file",
+        "org.apache.commons.io;IOUtils;true;contentEquals;(Reader,Reader);;Argument[1];generated:write-file",
```
smowton, line 141: These are read-only ops

java/ql/lib/semmle/code/java/frameworks/apache/IOGenerated.qll
```diff
@@ -1,684 +1,828 @@
...
+        "org.apache.commons.io;IOUtils;true;copy;(InputStream,Writer);;Argument[0];generated:write-file",
+        "org.apache.commons.io;IOUtils;true;copy;(InputStream,Writer,Charset);;Argument[0];generated:write-file",
+        "org.apache.commons.io;IOUtils;true;copy;(InputStream,Writer,String);;Argument[0];generated:write-file",
+        "org.apache.commons.io;IOUtils;true;copy;(Reader,Appendable);;Argument[0];generated:write-file",
+        "org.apache.commons.io;IOUtils;true;copy;(Reader,Appendable,CharBuffer);;Argument[0];generated:write-file",
+        "org.apache.commons.io;IOUtils;true;copy;(Reader,Appendable,CharBuffer);;Argument[2];generated:write-file",
+        "org.apache.commons.io;IOUtils;true;copy;(Reader,OutputStream);;Argument[0];generated:write-file",
+        "org.apache.commons.io;IOUtils;true;copy;(Reader,OutputStream,Charset);;Argument[0];generated:write-file",
+        "org.apache.commons.io;IOUtils;true;copy;(Reader,OutputStream,String);;Argument[0];generated:write-file",
+        "org.apache.commons.io;IOUtils;true;copy;(Reader,Writer);;Argument[0];generated:write-file",
```
smowton, line 151: More writes against read-only types

java/ql/lib/semmle/code/java/frameworks/apache/IOGenerated.qll
```diff
@@ -1,684 +1,828 @@
...
+        "org.apache.commons.io;IOUtils;true;copy;(URL,File);;Argument[0];generated:open-url",
+        "org.apache.commons.io;IOUtils;true;copy;(URL,File);;Argument[1];generated:create-file",
+        "org.apache.commons.io;IOUtils;true;copy;(URL,OutputStream);;Argument[0];generated:open-url",
+        "org.apache.commons.io;IOUtils;true;copyLarge;(Reader,Writer);;Argument[0];generated:write-file",
+        "org.apache.commons.io;IOUtils;true;copyLarge;(Reader,Writer,char[]);;Argument[0];generated:write-file",
+        "org.apache.commons.io;IOUtils;true;copyLarge;(Reader,Writer,char[]);;Argument[2];generated:write-file",
+        "org.apache.commons.io;IOUtils;true;copyLarge;(Reader,Writer,long,long);;Argument[0];generated:write-file",
+        "org.apache.commons.io;IOUtils;true;copyLarge;(Reader,Writer,long,long,char[]);;Argument[0];generated:write-file",
+        "org.apache.commons.io;IOUtils;true;copyLarge;(Reader,Writer,long,long,char[]);;Argument[4];generated:write-file",
```
smowton, line 160: write-file against Reader, and against `char[]` which is the written data

smowton (commented):
I've only reviewed as far as the root package but hopefully this gives an idea of the kind of erroneous model it appears we're generating

java/ql/lib/semmle/code/java/frameworks/apache/IOGenerated.qll
michaelnebel: Here is a screenshot of the path. The interesting bit is the jump from step 3-4. Not sure, how that happens. Basically, the entire method is considered a new remote due to this jump as inputstream is created from the URL Connection.

image

java/ql/lib/semmle/code/java/frameworks/apache/IOGenerated.qll
michaelnebel: It turned out that these summaries were derived from the following hand written summaries.
"java.io;Writer;true;append;;;Argument[0];write-file"
"java.io;Writer;true;write;;;Argument[0];write-file"

Writer is an abstract class for writing character streams - so they probably shouldn't be write-file sinks.

java/ql/lib/semmle/code/java/frameworks/apache/IOGenerated.qll
michaelnebel: Also existed before re-generation

java/ql/lib/semmle/code/java/frameworks/apache/IOGenerated.qll
michaelnebel: also existed before re-generation

java/ql/lib/semmle/code/java/frameworks/apache/IOGenerated.qll
michaelnebel: I don't fully understand, but in any case these models also existed before.

java/ql/lib/semmle/code/java/frameworks/apache/IOGenerated.qll
michaelnebel: The model generator will no longer derive "regex-use" sinks.

java/ql/lib/semmle/code/java/frameworks/apache/IOGenerated.qll
michaelnebel: These were derived due to java.io.Writer handwritten write-file sink. I have removed this sink.

java/ql/lib/semmle/code/java/frameworks/apache/IOGenerated.qll
michaelnebel: These were derived due to java.io.Writer handwritten write-file sink. I have removed this sink.

java/ql/lib/semmle/code/java/frameworks/apache/IOGenerated.qll
michaelnebel: These were derived due to java.io.Writer handwritten write-file sink. I have removed this sink.

java/ql/lib/semmle/code/java/frameworks/apache/IOGenerated.qll
michaelnebel: These were derived due to java.io.Writer handwritten write-file sink. I have removed this sink.

java/ql/lib/semmle/code/java/frameworks/apache/IOGenerated.qll
michaelnebel: These were derived due to java.io.Writer handwritten write-file sink. I have removed this sink.

java/ql/lib/semmle/code/java/frameworks/apache/IOGenerated.qll
michaelnebel: Fixed

yo-h:
For ease of review, it might be useful to split the changes to `IOGenerated.qll` into two commits: one that simply adds the `generated:` prefix to the `kind` column, and another one for the actual changes to the models.

michaelnebel:
> For ease of review, it might be useful to split the changes to `IOGenerated.qll` into two commits: one that simply adds the `generated:` prefix to the `kind` column, and another one for the actual changes to the models.

Yes, that would be ideal, I can add the "generated:" prefix manually to the existing generated models (as we can't use the generator for this - the "old" SHA is not working anymore and there might have been several changes to the model generator since the Java models have been re-generated last time around).

java/ql/lib/semmle/code/java/frameworks/apache/IOGenerated.qll
michaelnebel: Should we consider to just delete this row from the generated file?

atorralba: This looks like virtual dispatch creating more data flow edges than it should. I think we should remove the whole `IOGeneratedSourcesCsv` class for the time being.

atorralba:
> We probably need to go with the review mindset that we are aiming at not making anything (significantly) worse than it was before. The idea with the PR is simply to get the workflow running again and refresh the models with the existing logic. I have made some minor changes to cater for some of the review comments as they are relevant compared to the previously generated models.
Is that approach acceptable?

Sounds good to me, specially after having already handled all the comments above. I think a DCA run would make sense to validate we aren't significantly worsening our FP ratio, though.

michaelnebel:
> > We probably need to go with the review mindset that we are aiming at not making anything (significantly) worse than it was before. The idea with the PR is simply to get the workflow running again and refresh the models with the existing logic. I have made some minor changes to cater for some of the review comments as they are relevant compared to the previously generated models.
> > Is that approach acceptable?
> 
> Sounds good to me, specially after having already handled all the comments above. I think a DCA run would make sense to validate we aren't significantly worsening our FP ratio, though.

Yes, sure - I will trigger a DCA execution if the PR turns green with the latest changes!

atorralba (approved):
> DCA has not completed and there has not been identified any performance regressions or any new alerts. A single project has failed during DCA execution.

Yes, don't worry about the failing project, that happened on `main` too, so it has nothing to do with this PR.

Without having extensively reviewed the generated models after the latest changes, I think this fits the criteria of getting the ball rolling again and keeping results mostly as they were. So LGTM! 👍

PR #7029, opened 2021-11-02
JS: add js/session-fixation query

erik-krogh:
Inspired by [this csharp query](https://github.com/github/codeql/blob/e2cb53c65f804bd90bdce9313165d328cec46e70/csharp/ql/src/Security%20Features/CWE-384/AbandonSession.ql).  (Which has no results on all of lgtm.com). 

[An evaluation looks ok-ish](https://github.com/github/codeql-dca-main/tree/data/erik-krogh/cwe384__default__CustomSuite/reports).  
The new results appear to be TPs, but I'm not sure if they are worth fixing.  
So I set the severity to `warning` and the precision to `medium`, so that the alerts are not shown by default.

PR #9261, opened 2022-05-23
JS: remove support for passport in the session-fixation query

erik-krogh:
Passport added protection against session-fixation, so we should no longer detect it using our query.  
See this announcement: https://medium.com/passportjs/fixing-session-fixation-b2b68619c51d, and this helpful comment from the maintainer: https://github.com/github/codeql/pull/7029#issuecomment-1132940870

[Evaluation was uneventful](https://github.com/github/codeql-dca-main/tree/data/erik-krogh/pr-9261-1a8cf23__nightly-old__CustomSuite/reports).  
We loose some results, but that was the point.


asgerf (dismissed):
Ultra-minor nit, otherwise LGTM

asgerf (approved):

PR #9263, opened 2022-05-23
Kotlin: Add support for versions 1.5.0, 1.5.10, and 1.5.21

tamasvajk:


tamasvajk:
I've tested this locally by extracting gradle builds that required different versions of kotlin.

igfoo (approved):

PR #9264, opened 2022-05-23
Swift: fix extractor built with `NDEBUG`

redsun82:
There was a call with side effects in an `assert`, that was therefore
not being called with `NDEBUG` turned on, changing extractor results.

### DIFF
--- a/swift/extractor/SwiftDispatcher.h
+++ b/swift/extractor/SwiftDispatcher.h
@@ -153,7 +153,11 @@ class SwiftDispatcher {
   template <typename Tag, typename... Ts>
   TrapLabel<Tag> fetchLabelFromUnion(const llvm::PointerUnion<Ts...> u) {
     TrapLabel<Tag> ret{};
-    assert((... || fetchLabelFromUnionCase<Tag, Ts>(u, ret)) && "llvm::PointerUnion not set");
+    // with logical op short-circuiting, this will stop trying on the first successful fetch
+    // don't feel tempted to replace the variable with the expression inside the `assert`, or
+    // building with `NDEBUG` will not trigger the fetching
+    bool unionCaseFound = (... || fetchLabelFromUnionCase<Tag, Ts>(u, ret));
+    assert(unionCaseFound && "llvm::PointerUnion not set to a known case");
     return ret;
   }
 


swift/extractor/SwiftDispatcher.h
```diff
@@ -153,7 +153,9 @@ class SwiftDispatcher {
   template <typename Tag, typename... Ts>
   TrapLabel<Tag> fetchLabelFromUnion(const llvm::PointerUnion<Ts...> u) {
     TrapLabel<Tag> ret{};
-    assert((... || fetchLabelFromUnionCase<Tag, Ts>(u, ret)) && "llvm::PointerUnion not set");
+    // with logical op short-circuiting, this will stop trying on the first successful fetch
+    bool unionCaseFound = (... || fetchLabelFromUnionCase<Tag, Ts>(u, ret));
```
AlexDenisov, line 157: We should be marking these with `__used` or similar, otherwise there are lots of "unused variable" warnings when built without assertions.

AlexDenisov (approved):

AlexDenisov (approved):

swift/extractor/SwiftDispatcher.h
redsun82: yeah, I saw a lot of that as well, however maybe I would not do that before we roll out the shiny `Diagnostic` module that will save us all, as those warnings will probably go away anyway with that... We could maybe wrap `assert`s into a function maybe to avoid the warnings for now?

AlexDenisov: I think it's fine as it is right now, we can just wait for the diagnostics

sashabu: This is one of the motivating use cases for `[[maybe_unused]]` (and indeed the example at https://en.cppreference.com/w/cpp/language/attributes/maybe_unused).

And of course for more complex cases, another option is checking `NDEBUG` ourselves:
```
#ifdef NDEBUG
  bool unionCaseFound = 
#endif
      (... || fetchLabelFromUnionCase<Tag, Ts>(u, ret));
  assert(unionCaseFound && "llvm::PointerUnion not set to a known case");
```

Here, I think `[[maybe_unused]]` is the right "hammer": it's not very noisy, and does silence the warning (at the risk of leaving the unused variable if the assert goes away - but that's no worse a risk than learning to ignore warnings!). But checking `NDEBUG` can be useful when we *don't* want to run something in a release mode, but it doesn't comfortably fit inside an `assert` condition (e.g. if we'd written this as a loop instead of a fold).

PR #9267, opened 2022-05-23
Ruby: Add Improper Memoization query

hmac:
This query finds cases where a method memoizes its result but fails to
include one or more of its parameters in the memoization key (or doesn't
use memoization keys at all). This can lead to the method returning
incorrect results when subsequently called with different arguments.

I've added it to `experimental` initally, as it can flag methods which aren't security vulnerabilities.

### DIFF
--- a/ruby/ql/lib/codeql/ruby/security/ImproperMemoizationQuery.qll
+++ b/ruby/ql/lib/codeql/ruby/security/ImproperMemoizationQuery.qll
@@ -41,7 +41,7 @@ private class HashMemoStmt extends MemoStmt {
  * A method that may be performing memoization.
  */
 private class MemoCandidate extends Method {
-  MemoCandidate() { exists(MemoStmt m | m.getEnclosingMethod() = this) }
+  MemoCandidate() { this = any(MemoStmt m).getEnclosingMethod() }
 }
 
 /**
@@ -49,7 +49,7 @@ private class MemoCandidate extends Method {
  */
 private predicate parameterUsedInMemoValue(Method m, Parameter p, MemoStmt a) {
   p = m.getAParameter() and
-  a.getParent+() = m and
+  a.getEnclosingMethod() = m and
   p.getAVariable().getAnAccess().getParent*() = a.getRightOperand()
 }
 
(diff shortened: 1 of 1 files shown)


ruby/ql/lib/codeql/ruby/security/ImproperMemoizationQuery.qll
```diff
@@ -0,0 +1,87 @@
...
+      e = this.getLeftOperand() and this.getVariableAccess().getParent+() = e
+    )
+  }
+}
+
+/**
+ * A method that may be performing memoization.
+ */
+private class MemoCandidate extends Method {
+  MemoCandidate() { exists(MemoStmt m | m.getEnclosingMethod() = this) }
```
nickrolfe, line 44: Nitpick – I think we generally prefer to write it this way. At least, that's what Tom tells me :)
```suggestion
  MemoCandidate() { this = any(MemoStmt m).getEnclosingMethod() }
```

ruby/ql/lib/codeql/ruby/security/ImproperMemoizationQuery.qll
```diff
@@ -0,0 +1,87 @@
...
+}
+
+/**
+ * Holds if parameter `p` of `m` is read in the right hand side of `assign`.
+ */
+private predicate parameterUsedInMemoValue(Method m, Parameter p, MemoStmt a) {
+  p = m.getAParameter() and
+  a.getParent+() = m and
+  p.getAVariable().getAnAccess().getParent*() = a.getRightOperand()
```
nickrolfe, line 53: I'm always a little suspicious of any use of `getParent()` in queries – ideally there should be some more specific way of navigating the AST like that. Would writing line 52 as `a.getEnclosingMethod() = m` mean the same thing?

ruby/ql/lib/codeql/ruby/security/ImproperMemoizationQuery.qll
```diff
@@ -0,0 +1,87 @@
...
+}
+
+/**
+ * Holds if parameter `p` of `m` is read in the left hand side of `assign`.
+ */
+private predicate parameterUsedInMemoKey(Method m, Parameter p, HashMemoStmt a) {
+  p = m.getAParameter() and
+  a.getParent+() = m and
+  p.getAVariable().getAnAccess().getParent+() = a.getLeftOperand()
```
nickrolfe, line 62: Same question here.

nickrolfe (commented):
Looks like a nice query! I have just a few questions and nitpicks.

nickrolfe (approved):
👍🚀

PR #9270, opened 2022-05-23
C#: Summarized callable

michaelnebel:
In this PR, we
- Fix issue with the isAutoGenerated predicate (contains to much).
- Fix issue with dataflow library such that only relevant summarized callables will be used.
- Update tests and dependencies to other languages.


csharp/ql/lib/semmle/code/csharp/dataflow/internal/DataFlowPrivate.qll
```diff
@@ -743,11 +744,11 @@ private module Cached {
       FlowSummaryImpl::Public::SummarizedCallable c,
       FlowSummaryImpl::Private::SummaryNodeState state
     ) {
-      useFlowSummary(c) and
+      c instanceof DataFlowSummarizedCallable and
```
hvitved, line 747: You may as well change the type of `c` to be `DataFlowSummarizedCallable`.

csharp/ql/lib/semmle/code/csharp/dataflow/internal/DataFlowPrivate.qll
```diff
@@ -743,11 +744,11 @@ private module Cached {
...
       FlowSummaryImpl::Private::summaryNodeRange(c, state)
     } or
     TSummaryParameterNode(FlowSummaryImpl::Public::SummarizedCallable c, ParameterPosition pos) {
-      useFlowSummary(c) and
+      c instanceof DataFlowSummarizedCallable and
```
hvitved, line 751: Same as above.

csharp/ql/lib/semmle/code/csharp/dataflow/internal/DataFlowPrivate.qll
michaelnebel: Oh right!

michaelnebel:
None of the DCA runs shows any performance regressions or changes in alerts. This should be good to go.

hvitved (approved):

PR #9275, opened 2022-05-23
Swift: Add shared dataflow library

MathiasVP:
The first two commits are 100% bitwise copies from the internal repo. We can't yet add any tests since we don't extract Swift comments yet, and therefore cannot yet use the inline expectations test framework.


MathiasVP:
The Code Scanning alerts are all either:
- In shared files (which I don't want to change in this PR), or
- Caused by the fact that this is currently dead code.

rdmarsh2 (approved):

PR #9280, opened 2022-05-24
Kotlin: Fix missing kotlin to java property mapping

tamasvajk:


java/ql/test/kotlin/library-tests/java-map-methods/test.kt
```diff
@@ -1 +1,3 @@
 fun test(m: Map<Int, Int>) = m.getOrDefault(1, 2)
+
+fun test2(s: String) = s.length
```
smowton, line 3: Rename the test, since it no longer tests specifically map methods

tamasvajk: It is specific to mapping to java methods. :-)

smowton: Hah ok, that's now how I intended the name but I'm happy for it to start meaning that :)

smowton (approved):

PR #9284, opened 2022-05-24
Swift: CFG for `TypeExpr`, `MemberRefExpr`, `DefaultArgumentExpr` and `ForceValueExpr`

MathiasVP:
Turns out we already extract those, but we forgot to include them in the CFG.

### DIFF
--- a/swift/ql/test/library-tests/controlflow/graph/Cfg.expected
+++ b/swift/ql/test/library-tests/controlflow/graph/Cfg.expected
@@ -31,11 +31,14 @@ cfg.swift:
 #-----|  -> ReturnStmt
 
 #   15| DeclRefExpr
-#-----|  -> DotSyntaxCallExpr
+#-----|  -> TypeExpr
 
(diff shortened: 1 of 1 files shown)


MathiasVP:
> OOI, have you enabled the [CFG consistency queries](https://github.com/github/codeql/blob/3e1a6a777e15bb4fdf3fa99bab8bf04fb63d08ac/swift/ql/lib/codeql/swift/controlflow/internal/ControlFlowGraphImplShared.qll#L926) for Swift, or would that be too noisy at this stage?

We have not done that yet, no. That's probably a good thing to do next now that we have the entire CFG library in the CodeQL repo. Thanks for the reminder! I will add that once this PR has been merged.

MathiasVP:
@rdmarsh2 I've marked you as a reviewer since I think you had actually started on the CFG for some of these before moving to dataflow.

rdmarsh2 (approved):
Looks good to me

PR #9287, opened 2022-05-24
Ruby: flow through getters/setters

aibaars:
This pull request adds support for data flow through getter/setter methods. It assumes that setter methods like `field=` write to an instance variable named `@field`. This is to handle cases where the body of the setter method is not visible to the analysis. While this is an approximation, I'd expect very few cases of false flow. If my assumption turns out to be wrong we should make the analysis more precise.

### DIFF
--- a/ruby/ql/lib/codeql/ruby/dataflow/internal/DataFlowPrivate.qll
+++ b/ruby/ql/lib/codeql/ruby/dataflow/internal/DataFlowPrivate.qll
@@ -349,7 +349,14 @@ private module Cached {
     TUnknownElementContent() or
     TKnownPairValueContent(ConstantValue cv) or
     TUnknownPairValueContent() or
-    TFieldContent(string name) { name = any(InstanceVariable v).getName() }
+    TFieldContent(string name) {
+      name = any(InstanceVariable v).getName()
+      or
+      exists(SetterMethodCall c, string methodName |
+        methodName = c.getMethodName() and
+        name = "@" + methodName.prefix(methodName.length() - 1)
+      )
+    }
 
   /**
    * Holds if `e` is an `ExprNode` that may be returned by a call to `c`.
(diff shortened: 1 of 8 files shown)


ruby/ql/lib/codeql/ruby/dataflow/internal/DataFlowPrivate.qll
```diff
@@ -342,7 +342,14 @@ private module Cached {
       not cv.isInt(_) or
       cv.getInt() in [0 .. 10]
     } or
-    TFieldContent(string name) { name = any(InstanceVariable v).getName() } or
+    TFieldContent(string name) {
+      name = any(InstanceVariable v).getName()
+      or
+      exists(SetterMethodCall c, string methodName |
+        methodName = c.getMethodName() and
+        name = "@" + methodName.prefix(methodName.length() - 1)
```
hvitved, line 350: Perhaps we should have a
```ql
string SetterMethodCall::getTargetName() {
  exists(string methodName |
    methodName = c.getMethodName() and
    result = methodName.prefix(methodName.length() - 1)
  )
}
```
and then call that here and further down.
}
```

File: ruby/ql/lib/codeql/ruby/dataflow/internal/DataFlowPrivate.qll [hunk 2]
```diff
@@ -833,6 +854,19 @@ predicate readStep(Node node1, ContentSet c, Node node2) {
         ))
     )
   or
+  // Attribute read, `receiver.field`. Note that we do not check whether
+  // the `field` method is really an attribute reader. This is probably fine
+  // because the read step has only effect if there exists a matching store step
+  // (instance variable assignmentor setter method call).
```
hvitved, line 860: `assignment or`

hvitved (changes requested):
LGTM. Let's see what DCA says.

hvitved (approved):
LGTM. Let's wait for DCA before merging.

PR #9288, opened 2022-05-24
JS: Remove Buffer.from sink from js/resource-exhaustion

asgerf:
I came across what I believe to be a FP from this query, since the 3-argument version of `Buffer.from` does not allocate a new underlying buffer.

The three-argument version exposes a slice of an existing `ArrayBuffer`, and throws an exception if the slice is out of bounds.

https://nodejs.org/dist/latest-v10.x/docs/api/buffer.html#buffer_class_method_buffer_from_arraybuffer_byteoffset_length

[Evaluation](https://github.com/github/codeql-dca-main/blob/data/asgerf/resource-exhaus__default__CustomSuite/reports/alert-comparison.md) shows the fixed FP


erik-krogh (approved):
Nice catch :+1:

PR #9289, opened 2022-05-24
JS: Support the remaining of the finished ES2022 proposals

erik-krogh:
I took a look at all the finished 2022 proposals from this list: https://github.com/tc39/proposals/blob/main/finished-proposals.md  

We already had support for most of the features, so there wasn't that much more to add.  

I also looked at [RegExp Match Indices](https://github.com/tc39/proposal-regexp-match-indices), but I concluded that we have nothing to support there, because all the values in the new `indices` array are numbers representing various indexes, and those numbers are not tainted.  

[Evaluation was unevenful](https://github.com/github/codeql-dca-main/tree/data/erik-krogh/pr-9289-82c6c22__nightly-old__code-scanning/reports).


asgerf (approved):
👍

PR #9291, opened 2022-05-24
Swift: CFG for property reads and writes

MathiasVP:
This PR implements control flow for property reads/writes, subscript read/writes, and property observers.

Sadly, it turns out to require quite a bit of setup before this is possible. Therefore, I recommend a commit-by-commit review. Here's the overall description of each commit:

- 6f6d3b53025a82855a52f4225b73f528c6836319 creates a new IPA type whose branches provide the elements in the CFG. As of this PR this is the AST nodes (like on `main`) along with 3 new branches:
  - One for property reads
  - One for property writes
  - One for property observers
- 5b58e6f5c4256293f80f18dbc5f8951d47616392 is a sad copy/paste of some of the CFG library's internal code. This is a solution @hvitved and I came up with to fix a problem that the shared CFG library imposes on us. The commit message has a detailed description of the problem and the fix. (@hvitved I'd be curious to know if you think parameterized modules will remove the need for this copy/pasting. I imagine we could also do something with `sync-identical-files` otherwise.)
- 0047421fe98ec9fedf7f2351af8ff9dfb262f580 IPAs the CFG construction code.
- 9b764b6df5efff270ddd98cefe41d0ac59088693 _finally_ creates the CFG for reads/writes, subscript read/writes, and property observers.

This PR is currently a draft because it's on top of https://github.com/github/codeql/pull/9284. Once that is merged I'll rebase `main` onto this PR and pull it out of draft.


swift/ql/lib/codeql/swift/controlflow/internal/AstControlFlowTrees.qll
```diff
@@ -0,0 +1,264 @@
...
+  abstract class AstLeafTree extends AstPreOrderTree, AstPostOrderTree {
+    override predicate propagatesAbnormal(ControlFlowElement child) { none() }
+
+    override predicate succ(ControlFlowElement pred, ControlFlowElement succ, Completion c) {
+      none()
+    }
+  }
+
+  private class LeafTreeAdaptor extends LeafTree {
+    AstLeafTree tree;
```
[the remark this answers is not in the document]
MathiasVP: While the alert is genuine, I think it should be ignored: This is a special case of a pattern all throughout the file (i.e., bind the wrapped tree in the char pred and leave it as a class field).

swift/ql/lib/codeql/swift/controlflow/internal/AstControlFlowTrees.qll
```diff
@@ -0,0 +1,264 @@
...
+ * Holds if `first` is the first element executed within control flow
+ * element `cft`.
+ */
+predicate astFirst(AstControlFlowTree cft, ControlFlowElement first) { cft.first(first) }
+
+/**
+ * Holds if `last` with completion `c` is a potential last element executed
+ * within control flow element `cft`.
+ */
+predicate astLast(AstControlFlowTree cft, ControlFlowElement last, Completion c) {
```
hvitved, line 61: Couldn't this be `last(any(ControlFlowTree tree | tree.asAstNode()), last, c`?

swift/ql/lib/codeql/swift/controlflow/internal/AstControlFlowTrees.qll
```diff
@@ -0,0 +1,264 @@
...
+
+    /**
+     * Holds if `succ` is a control flow successor for `pred`, given that `pred`
+     * finishes with completion `c`.
+     */
+    pragma[nomagic]
+    abstract predicate succ(ControlFlowElement pred, ControlFlowElement succ, Completion c);
+  }
+
+  private class ControlFlowTreeAdaptor extends ControlFlowTree {
```
hvitved, line 30: I believe it is `Adapter`

swift/ql/lib/codeql/swift/controlflow/internal/AstControlFlowTrees.qll
```diff
@@ -0,0 +1,264 @@
...
+  )
+}
+
+module PreOrderTrees {
+  /** An element that is executed in pre-order. */
+  abstract class AstPreOrderTree extends AstControlFlowTree {
+    final override predicate first(ControlFlowElement first) { first.asAstNode() = this }
+  }
+
+  private class PreOrderTreeAdaptor extends PreOrderTree {
```
hvitved, line 77: I would think we should only need one adapter.

swift/ql/lib/codeql/swift/controlflow/internal/AstControlFlowTrees.qll
```diff
@@ -0,0 +1,264 @@
...
+  }
+}
+
+import ControlFlowTrees
+
+/**
+ * Holds if `first` is the first element executed within control flow
+ * element `cft`.
+ */
+predicate astFirst(AstControlFlowTree cft, ControlFlowElement first) { cft.first(first) }
```
hvitved, line 55: It means the same (now), but I prefer `first(any(ControlFlowTree tree | tree.asAstNode()), first)`.

swift/ql/lib/codeql/swift/controlflow/internal/ControlFlowElements.qll
```diff
@@ -0,0 +1,162 @@
+private import swift
+
+newtype TControlFlowElement =
```
hvitved, line 3: Cached?

hvitved:
An alternative solution, which avoids duplication, is to define
```ql
  abstract class AstControlFlowTree extends ControlFlowTree {
    AstNode n;

    AstControlFlowTree() { n = this.asAstNode() }
  }

  abstract class AstPreOrderTree extends AstControlFlowTree, PreOrderTree {}
```

and then do dispatch by overriding the field, e.g.

```ql
  private class IfStmtTree extends AstPreOrderTree {
    override IfStmt n;

    final override predicate propagatesAbnormal(ControlFlowElement child) {
      child.asAstNode() = n.getCondition().getFullyUnresolved() or
      child.asAstNode() = n.getThen() or
      child.asAstNode() = n.getElse()
    }

    final override predicate last(ControlFlowElement last, Completion c) {
      // Condition exits with a false completion and there is no `else` branch
      astLast(n.getCondition().getFullyUnresolved(), last, c) and
      c instanceof FalseCompletion and
      not exists(n.getElse())
      or
      // Then/Else branch exits with any completion
      astLast(n.getBranch(_), last, c)
    }

    final override predicate succ(ControlFlowElement pred, ControlFlowElement succ, Completion c) {
      // Pre-order: flow from statement itself to first element of condition
      pred.asAstNode() = n and
      astFirst(n.getCondition().getFullyUnresolved(), succ) and
      c instanceof SimpleCompletion
      or
      astLast(n.getCondition().getFullyUnresolved(), pred, c) and
      (
        // Flow from last element of condition to first element of then branch
        c instanceof TrueCompletion and astFirst(n.getThen(), succ)
        or
        // Flow from last element of condition to first element of else branch
        c instanceof FalseCompletion and astFirst(n.getElse(), succ)
      )
    }
  }
```

The only down-side to that approach is that it may result in higher BDD node usage (right, @alexet ?).

MathiasVP:
Thanks a lot for that suggestion, Tom! I think I'll update the PR to your proposed design, and then we can always roll it back to this original PR if BDD node usage becomes an issue in the future.

swift/ql/lib/codeql/swift/controlflow/internal/ControlFlowElements.qll
MathiasVP: Yeah, this should definitely be cached. Thanks!

swift/ql/lib/codeql/swift/controlflow/internal/AstControlFlowTrees.qll
MathiasVP: Indeed, that's much better. I've pushed that change now: https://github.com/github/codeql/pull/9291/commits/1f4924f978bc5e3fff7a32c13cc9e307085a611a#diff-a3a428364dc34dde5d544b5ba69082c665a60fdf4dee850acb1dddd46e901b5bR18

To keep the commit history clean for the reviewers on the C team I've forced pushed your proposed change. That makes it slightly more difficult for you, but I think it'll make the PR a lot easier to review for everybody else.

swift/ql/lib/codeql/swift/controlflow/internal/AstControlFlowTrees.qll
MathiasVP: Indeed. 1f4924f978bc5e3fff7a32c13cc9e307085a611a (vacuously) fixes this.

swift/ql/lib/codeql/swift/controlflow/internal/ControlFlowGraphImpl.qll
```diff
@@ -300,532 +331,681 @@ module Stmts {
...
+      astLast(ast.getSubExpr().getFullyConverted(), pred, c) and
       c instanceof NormalCompletion and
-      succ = this
+      succ.asAstNode() = ast
     }
   }
 
-  class DoOrDoCatchStmt = @do_stmt or @do_catch_stmt;
+  abstract class DoOrDoCatchTree extends AstPreOrderTree {
+    DoOrDoCatchTree() {
```
hvitved, line 629: Same comment about charpred

swift/ql/lib/codeql/swift/controlflow/internal/ControlFlowGraphImpl.qll
```diff
@@ -837,298 +1017,536 @@ module Exprs {
...
+    }
   }
 
   module Conversions {
     class ConversionOrIdentity = @identity_expr or @explicit_cast_expr or @implicit_conversion_expr;
 
-    abstract class ConversionOrIdentityTree extends StandardPostOrderTree instanceof ConversionOrIdentity {
+    abstract class ConversionOrIdentityTree extends AstStandardPostOrderTree {
+      ConversionOrIdentityTree() {
+        ast instanceof IdentityExpr or
```
hvitved, line 1520: Same

swift/ql/lib/codeql/swift/controlflow/internal/ControlFlowGraphImpl.qll
```diff
@@ -284,14 +308,21 @@ module Stmts {
 
     class LoopStmt = @for_each_stmt or ConditionalLoop;
 
-    abstract class LoopTree extends PreOrderTree instanceof ConditionalLoop {
-      abstract AstNode getCondition();
+    abstract class LoopTree extends AstPreOrderTree {
+      LoopTree() {
+        ast instanceof WhileStmt or
+        ast instanceof RepeatWhileStmt
+      }
```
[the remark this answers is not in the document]
MathiasVP: Because that doesn't compile :( I'll send you the link to my exact same question on Slack.

MathiasVP:
This PR should be good to go, @rdmarsh2. The actual technical content of this PR isn't very deep, but do let me know if you want to go over it over Zoom due to the _number_ of edits (since the PR basically changes every tree in the CFG construction).

swift/ql/lib/codeql/swift/controlflow/internal/ControlFlowGraphImpl.qll
```diff
@@ -284,14 +308,21 @@ module Stmts {
 
     class LoopStmt = @for_each_stmt or ConditionalLoop;
 
-    abstract class LoopTree extends PreOrderTree instanceof ConditionalLoop {
-      abstract AstNode getCondition();
+    abstract class LoopTree extends AstPreOrderTree {
+      LoopTree() {
+        ast instanceof WhileStmt or
```
rdmarsh2, line 313: why not `ast instanceof ConditionalLoop`?

swift/ql/lib/codeql/swift/controlflow/internal/ControlFlowElements.qll
```diff
@@ -0,0 +1,163 @@
...
+  e.(SubscriptExpr).hasOrdinarySemantics()
+  or
+  e.(DeclRefExpr).hasOrdinarySemantics()
+}
+
+private predicate hasDirectToImplementationOrOrdinarySemantics(Expr e) {
+  hasDirectToImplementationSemantics(e) or hasOrdinarySemantics(e)
+}
+
+predicate isPropertySetterElement(AccessorDecl accessor, AssignExpr assign) {
```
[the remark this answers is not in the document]
MathiasVP: I'm actually using the two-parameter forms a couple of times in the impl file, but I might as well replace those with the 3-parameter version and add a `_`. So I'll do this in a follow-up. Thanks!

swift/ql/lib/codeql/swift/controlflow/internal/ControlFlowGraphImpl.qll
MathiasVP: Good point. I'll fix that in a follow-up in a second.

MathiasVP: PR with a fix: https://github.com/github/codeql/pull/9329

swift/ql/lib/codeql/swift/controlflow/internal/ControlFlowElements.qll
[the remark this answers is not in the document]
MathiasVP: PR with a fix: https://github.com/github/codeql/pull/9329

PR #9294, opened 2022-05-24
Java: Add support for BarrierGuards as parameterised modules.

aschackmull:
This adds a parameterised-module alternative to the existing BarrierGuard class.

### DIFF
--- a/java/ql/lib/semmle/code/java/dataflow/internal/DataFlowUtil.qll
+++ b/java/ql/lib/semmle/code/java/dataflow/internal/DataFlowUtil.qll
@@ -305,7 +305,7 @@ class ContentSet instanceof Content {
 }
 
 /**
- * Holds if `g` validates the `e` upon evaluating to `branch`.
+ * Holds if the guard `g` validates the expression `e` upon evaluating to `branch`.
  *
  * The expression `e` is expected to be a syntactic part of the guard `g`.
  * For example, the guard `g` might be a call `isSafe(x)` and the expression `e`
--- a/java/ql/lib/semmle/code/java/dataflow/internal/DataFlowUtil.qll
+++ b/java/ql/lib/semmle/code/java/dataflow/internal/DataFlowUtil.qll
@@ -313,6 +313,12 @@ class ContentSet instanceof Content {
  */
 signature predicate guardChecksSig(Guard g, Expr e, boolean branch);
 
+/**
+ * Provides a set of barrier nodes for a guard that validates an expression.
+ *
+ * This is expected to be used in `isBarrier`/`isSanitizer` definitions
+ * in data flow and taint tracking.
+ */
 module BarrierGuard<guardChecksSig/3 guardChecks> {
   /** Gets a node that is safely guarded by the given guard check. */
   Node getABarrierNode() {
--- a/java/ql/lib/semmle/code/java/dataflow/internal/DataFlowUtil.qll
+++ b/java/ql/lib/semmle/code/java/dataflow/internal/DataFlowUtil.qll
@@ -314,7 +314,7 @@ class ContentSet instanceof Content {
 signature predicate guardChecksSig(Guard g, Expr e, boolean branch);
 
 module BarrierGuard<guardChecksSig/3 guardChecks> {
-  /** Gets a node that is safely guarded by the given guard. */
+  /** Gets a node that is safely guarded by the given guard check. */
   Node getABarrierNode() {
     exists(Guard g, SsaVariable v, boolean branch, RValue use |
       guardChecks(g, v.getAUse(), branch) and
--- a/java/ql/lib/semmle/code/java/dataflow/internal/DataFlowUtil.qll
+++ b/java/ql/lib/semmle/code/java/dataflow/internal/DataFlowUtil.qll
@@ -305,6 +305,27 @@ class ContentSet instanceof Content {
 }
 
 /**
+ * Holds if `g` validates the `e` upon evaluating to `branch`.
+ *
+ * The expression `e` is expected to be a syntactic part of the guard `g`.
+ * For example, the guard `g` might be a call `isSafe(x)` and the expression `e`
+ * the argument `x`.
+ */
+signature predicate guardChecksSig(Guard g, Expr e, boolean branch);
+
+module BarrierGuard<guardChecksSig/3 guardChecks> {
+  /** Gets a node that is safely guarded by the given guard. */
+  Node getABarrierNode() {
+    exists(Guard g, SsaVariable v, boolean branch, RValue use |
+      guardChecks(g, v.getAUse(), branch) and
+      use = v.getAUse() and
+      g.controls(use.getBasicBlock(), branch) and
+      result.asExpr() = use
+    )
+  }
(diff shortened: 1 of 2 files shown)


ginsbach:
The use of parameterised modules here looks unproblematic to me, so I'd say it's good to go from that perspective.

smowton:
This is pleasingly legible for someone seeing parameterized modules for the first time!

java/ql/lib/semmle/code/java/dataflow/internal/DataFlowUtil.qll
```diff
@@ -304,6 +304,27 @@ class ContentSet instanceof Content {
...
+/**
+ * Holds if `g` validates the `e` upon evaluating to `branch`.
+ *
+ * The expression `e` is expected to be a syntactic part of the guard `g`.
+ * For example, the guard `g` might be a call `isSafe(x)` and the expression `e`
+ * the argument `x`.
+ */
+signature predicate guardChecksSig(Guard g, Expr e, boolean branch);
+
+module BarrierGuard<guardChecksSig/3 guardChecks> {
```
RasmusWL, line 316: I think a QLDoc here explaining _how to use_ this parameterized module would be great! From _just_ reading the change in this file, I did not know whether I was expected to use this with `isBarrier`/`isBarrierIn`/`isBarrierOut`.

I think it could also still make sense to follow the [QLDoc style guide for modules](https://github.com/github/codeql/blob/main/docs/qldoc-style-guide.md#qldoc-for-modules), such that it reads

```codeql
/**
 * Provides logic for barrier guards, where an expression is validated to be safe
 * in a specific branch.
 *
 * Expected to be used in dataflow/taint-tracking configurations with `isBarrier`/`isSanitizer`
 */
...
```

:thinking: initially I wrote the first line as `Provides logic for barrier guards` which feels like an autogenerated JavaDoc that doesn't bring much value...

java/ql/lib/semmle/code/java/dataflow/internal/DataFlowUtil.qll
```diff
@@ -304,6 +304,27 @@ class ContentSet instanceof Content {
   }
 }
 
+/**
+ * Holds if `g` validates the `e` upon evaluating to `branch`.
```
RasmusWL, line 308: nitpick, either of these would be fine for me -- but I would suspect the 1 or 3 to be the preferred one. I think I actually prefer the third one.

```suggestion
 * Holds if `g` validates `e` upon evaluating to `branch`.
```

```suggestion
 * Holds if `g` validates the expression `e` upon evaluating to `branch`.
```

```suggestion
 * Holds if the guard `g` validates the expression `e` upon evaluating to `branch`.
```

java/ql/lib/semmle/code/java/dataflow/internal/DataFlowUtil.qll
```diff
@@ -304,6 +304,27 @@ class ContentSet instanceof Content {
...
+ *
+ * The expression `e` is expected to be a syntactic part of the guard `g`.
+ * For example, the guard `g` might be a call `isSafe(x)` and the expression `e`
+ * the argument `x`.
+ */
+signature predicate guardChecksSig(Guard g, Expr e, boolean branch);
+
+module BarrierGuard<guardChecksSig/3 guardChecks> {
+  /** Gets a node that is safely guarded by the given guard. */
```
RasmusWL, line 317:
```suggestion
  /** Gets a node that is safely guarded by the given guard check. */
```

java/ql/lib/semmle/code/java/dataflow/internal/DataFlowUtil.qll
```diff
@@ -304,6 +304,27 @@ class ContentSet instanceof Content {
...
+  Node getABarrierNode() {
+    exists(Guard g, SsaVariable v, boolean branch, RValue use |
+      guardChecks(g, v.getAUse(), branch) and
+      use = v.getAUse() and
```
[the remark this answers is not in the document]
aschackmull: This is a good point, but something that I think makes sense to possibly tackle in a separate PR. For this PR I want to keep the current semantics of BarrierGuards.

java/ql/lib/semmle/code/java/dataflow/internal/DataFlowUtil.qll
aschackmull: fixed

aschackmull:
Delaying change note to a follow-up PR, which is going to deprecate the existing BarrierGuard abstract class.

java/ql/lib/semmle/code/java/dataflow/internal/DataFlowUtil.qll
```diff
@@ -304,6 +304,27 @@ class ContentSet instanceof Content {
...
+ *
+ * The expression `e` is expected to be a syntactic part of the guard `g`.
+ * For example, the guard `g` might be a call `isSafe(x)` and the expression `e`
+ * the argument `x`.
+ */
+signature predicate guardChecksSig(Guard g, Expr e, boolean branch);
+
+module BarrierGuard<guardChecksSig/3 guardChecks> {
+  /** Gets a node that is safely guarded by the given guard. */
+  Node getABarrierNode() {
```
michaelnebel, line 324: Maybe rename to getAGuardedNode for consistency (looks like this is the higher order version of the getAGuardedNode in the class BarrierGuard.

java/ql/src/Security/CWE/CWE-022/TaintedPath.ql
```diff
@@ -41,10 +39,8 @@ class TaintedPathConfig extends TaintTracking::Configuration {
 
   override predicate isSanitizer(DataFlow::Node node) {
     exists(Type t | t = node.getType() | t instanceof BoxedType or t instanceof PrimitiveType)
-  }
-
-  override predicate isSanitizerGuard(DataFlow::BarrierGuard guard) {
-    guard instanceof ContainsDotDotSanitizer
+    or
+    node = DataFlow::BarrierGuard<containsDotDotSanitizer/3>::getABarrierNode()
```
michaelnebel, line 43: Can you elaborate, why you can basically turn the isSanitizer and isSanitizerGuard predicate into a disjunkt in isSanitizer?

michaelnebel (approved):
Looks plausible to me from a semantic equivalence perspective.
Have added some questions out of curiosity.

java/ql/src/Security/CWE/CWE-022/TaintedPath.ql
aschackmull: That's exactly how BarrierGuards are currently implemented by the library.

java/ql/lib/semmle/code/java/dataflow/internal/DataFlowUtil.qll
aschackmull: I like `getABarrierNode()` better, since it's supposed to plug directly into `isBarrier(Node)`, so I think this will be more intuitive. Also, consistency will be trivially achieved once the BarrierGuard class is deleted.

PR #9314, opened 2022-05-24
Convert `.codeqlmanifest.json` file to `codeql-workspace.yml`

aeisenberg:
Also add an explanation about why `codeql/javascript-experimental-atm-model` isn't in the workspace file.


aeisenberg:
Looks like `Check synchronized files` is broken on main right now. When that gets fixed, I'll re-run the job here and it should pass. The failure is unrelated to my change.

henrymercer (commented):
This was a deliberate design choice, albeit one that we could have documented better.

Most users will want to run `codeql pack install` to install the model pack from the registry.  Users will only want to resolve the local model pack if they are running ML-powered queries with a custom model.  Therefore by default we want the model pack to be resolved from the package cache.

Does `.codeqlmanifest.json` support JSON with comments so we can document this inline?

aeisenberg:
@henrymercer I changed this PR to convert to the `codeql-workspace.yml` file, and I added an appropriate comment.

codeql-workspace.yml
```diff
@@ -0,0 +1,29 @@
...
+  - "*/ql/lib/qlpack.yml"
+  - "*/ql/test/qlpack.yml"
+  - "*/ql/examples/qlpack.yml"
+  - "*/ql/consistency-queries/qlpack.yml"
+  - "cpp/ql/test/query-tests/Security/CWE/CWE-190/semmle/tainted/qlpack.yml"
+  - "go/ql/config/legacy-support/qlpack.yml"
+  - "go/build/codeql-extractor-go/codeql-extractor.yml"
+  - "javascript/ql/experimental/adaptivethreatmodeling/lib/qlpack.yml"
+  # This pack is explicitly excluded from the workspace since we
+  # require consumers to use a version of this pack from the package cache.
```
henrymercer, line 12:
```suggestion
  # This pack is explicitly excluded from the workspace since most users
  # will want to use a version of this pack from the package cache. Internal
  # users can uncomment the following line and place a custom ML model
  # in the corresponding pack to test a custom ML model within their local
  # checkout.
```

henrymercer (dismissed):
LGTM, suggest we also update the PR description.

henrymercer (approved):

PR #9315, opened 2022-05-25
Swift: Sync changes to DataFlowImplCommon from PR #9024.

michaelnebel:


MathiasVP:
Thanks a lot for fixing this!

PR #9319, opened 2022-05-25
Kotlin: Add taint step for String.valueOf(Editable)

atorralba:
Kotlin inlines `expr.toString()` as `String.valueOf(expr)` when `expr` is nullable.

This PR extends #8872 by adding a taint step to handle the above when `expr` is an `Editable`.

### DIFF
--- a/java/ql/test/library-tests/frameworks/android/widget/TestWidgetKt.kt
+++ b/java/ql/test/library-tests/frameworks/android/widget/TestWidgetKt.kt
@@ -1,0 +1,16 @@
+ import android.text.Editable
+
+ class TestWidget {
+
+    fun source() : Editable? { return null }
+    fun sink(sink : String) {}
+
+    fun test() {
+        val t = source()
+        sink(t.toString()); // $ hasTaintFlow
+
+        val t2 : Any? = source()
+        sink(t2.toString()); // $ MISSING: hasTaintFlow
+    }
+}
+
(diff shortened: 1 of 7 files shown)


java/ql/test/library-tests/frameworks/android/widget/TestWidgetKt.kt
```diff
@@ -0,0 +1,16 @@
...
+
+    fun source() : Editable? { return null }
+    fun sink(sink : String) {}
+
+    fun test() {
+        val t = source()
+        sink(t.toString()); // $ hasTaintFlow
+
+        val t2 : Any? = source()
+        sink(t2.toString()); // $ MISSING: hasTaintFlow
```
smowton, line 13: Why does it still fail?

smowton (approved):
Changes look reasonable; curious to know why the result is still tagged `MISSING`

java/ql/test/library-tests/frameworks/android/widget/TestWidgetKt.kt
atorralba: `t2`'s type is `Object` because of the `Any?` declaration. So the taint step works for `Editable?` but not for `Any?`.

This is also causing other issues, e.g. `toString`'s declaring type is `Any` in both cases, which interferes with our `Method::override` logic.

smowton: I believe that's consistent with the Java extractor -- if the receiver type doesn't supply an override then MethodAccess.getMethod() points to the inherited method.

atorralba:
Tests are now green after https://github.com/github/codeql/pull/9405.

PR #9320, opened 2022-05-25
Ruby: Flow through hash-splat parameters

hvitved:
This PR implements flow into hash-splat parameters, now that we have support for [flow through hashes](https://github.com/github/codeql/pull/8942). For a call with keyword arguments

```rb
m(foo: 1, bar: 2)
```

we synthesize a new hash-splat argument at the call-site, which contains `{:foo => 1, :bar => 2}`, and map that to the hash-splat parameter of the callee using newly introduced [argument/parameter positions](https://github.com/github/codeql/pull/7260). However, we also need to make sure that the callee filters away keyword arguments that occur as parameters, that is, in

```rb
def m(foo:, **rest)
```

we make sure that `**rest` only contains `{:bar => 2}`.

### DIFF
--- a/ruby/ql/test/library-tests/dataflow/params/params_flow.rb
+++ b/ruby/ql/test/library-tests/dataflow/params/params_flow.rb
@@ -23,11 +23,13 @@ keyword(p2: taint(5), p1: taint(6))
 keyword(:p2 => taint(7), :p1 => taint(8))
 
 def kwargs(p1:, **kwargs)
-    sink p1 # $ hasValueFlow=9
+    sink p1 # $ hasValueFlow=9 $ hasValueFlow=13
     sink (kwargs[:p1])
     sink (kwargs[:p2]) # $ hasValueFlow=10
-    sink (kwargs[:p3]) # $ hasValueFlow=11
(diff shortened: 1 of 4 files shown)


ruby/ql/test/library-tests/dataflow/params/params_flow.rb
```diff
@@ -21,3 +21,13 @@ def keyword(p1:, p2:)
...
+
+def kwargs(p1:, **kwargs)
+    sink p1 # $ hasValueFlow=9
+    sink (kwargs[:p1])
+    sink (kwargs[:p2]) # $ hasValueFlow=10
+    sink (kwargs[:p3]) # $ hasValueFlow=11
+    sink (kwargs[:p4])
+end
+
+kwargs(p1: taint(9), p2: taint(10), p3: taint(11), p4: "")
```
aibaars, line 33: Shall we also add a test case like
```
args = { p3: taint(11), p4: "" }
kwargs2(p1: taint(9), p2: taint(10), **args)
```

aibaars (commented):
Looks good to me.

aibaars (approved):

PR #9321, opened 2022-05-25
Swift: do not duplicate 'external' declarations

AlexDenisov:


swift/extractor/SwiftDispatcher.h
```diff
@@ -120,6 +132,33 @@ class SwiftDispatcher {
...
+  bool shouldEmitDeclBody(swift::Decl* decl) {
+    switch (extractionMode) {
+      case SwiftExtractionMode::Module: {
+        return currentModule == decl->getModuleContext();
+      } break;
+      case SwiftExtractionMode::PrimaryFile: {
+        swift::SourceLoc location = decl->getStartLoc();
+        if (!location.isValid()) {
+          return false;
+        }
```
redsun82, line 152: now that I think about it, this might exclude some compiler synthesized declarations, which we currently explicitly exclude from tests (which is probably not good).

This is probably not something we can really do...

We might need maybe to climb the `ASTContext` tree until we have a valid location and then use that.

and we need to have a test showing synthesized stuff, maybe tweaking the test inclusion predicate

swift/extractor/trap/TrapOutput.h
```diff
@@ -39,6 +39,12 @@ class TrapOutput {
     print(e);
   }
 
+  template <typename... Args>
```
redsun82, line 42: maybe we should leave a TODO here to make this conditional on a flag or environment variable?

redsun82 (commented):
Thanks for this! I have a hunch that something is not completely right with the `shouldEmitDeclBody` test, but I need some time to investigate that to give more substance.

swift/extractor/trap/TrapOutput.h
AlexDenisov: So far it's only a way to emit debug info while working locally, not something we call unconditionally (at least I did not intend to actually use it in this PR).

swift/extractor/SwiftDispatcher.h
AlexDenisov: That's a good point! What about checking for filename only if the location is present, otherwise emitting it under condition that the module is the same? 🤔

swift/extractor/visitors/DeclVisitor.h
```diff
@@ -309,6 +354,9 @@ class DeclVisitor : public AstVisitorBase<DeclVisitor> {
     }
     emitValueDecl(decl, label);
   }
+
+ private:
+  swift::Mangle::ASTMangler mangler;
```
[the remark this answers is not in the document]
AlexDenisov: Either option works for me. My personal preference is to use consistent naming regardless of whether the variable belongs to a class, or it's a local variable. But I don't really have any strong opinions on this one.

swift/extractor/SwiftDispatcher.h
```diff
@@ -18,8 +20,18 @@ namespace codeql {
 class SwiftDispatcher {
  public:
   // sourceManager, arena, and trap are supposed to outlive the SwiftDispatcher
```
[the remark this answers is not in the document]
redsun82: you could also update it to a catch-all
```suggestion
  // all references passed as parameters to this constructor are supposed to outlive the SwiftDispatcher
```
so that it does not need to be updated (unless we start passing a reference that is needed only during construction) :slightly_smiling_face:

swift/extractor/SwiftDispatcher.h
redsun82: (by another file I mean another file of the same module we are currently building)

swift/extractor/trap/TrapOutput.h
redsun82: ok I see! we probably should be having some more permanent trap debugging tools, but we can do it at a later stage

swift/extractor/visitors/DeclVisitor.h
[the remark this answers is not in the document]
redsun82: I also have no strong opinion, maybe we can let slack decide for us? :slightly_smiling_face:

swift/extractor/SwiftDispatcher.h
AlexDenisov: Yes, you are right!
I'll add a TODO here, and then when we have an example we can extend the functionality. WDYT?

redsun82: ah, whoops, now the enum test actually shows this problem :hand_over_mouth: If you want to solve it at a later stage, you can disable the enum test by renaming `enums.swift` to `enum.swift.TODO` and running codegen again.

redsun82 (approved):

PR #9322, opened 2022-05-25
QL/RB: fix the QL-for-QL and ruby autobuilders

erik-krogh:
See the post-merge discussion here for context:  https://github.com/github/codeql/pull/7444  

The CodeQL CLI now has an `--also-match` option for `codeql database index-files`, which we can use to make the autobuilder work as it should.   

This will unbreak the [`paths:` option in codeql-action](https://docs.github.com/en/code-security/code-scanning/automatically-scanning-your-code-for-vulnerabilities-and-errors/configuring-code-scanning#specifying-directories-to-scan) for Ruby.  
(Currently the `paths:` option does nothing, the Autobuilder just includes all the files even if you tell it to only include a specific folder).  

The test that it works is that QL-for-QL doesn't blow up.  
(QL-for-QL uses the `paths:` and `paths-ignore:` feature in `codeql-action`). 

I had to remove `tools:latest`, because that was stuck on `2.8.3` for some reason.   
(`2.8.3` doesn't have the `--also-match` option).

### DIFF
--- a/ql/autobuilder/src/main.rs
+++ b/ql/autobuilder/src/main.rs
@@ -29,18 +29,11 @@ fn main() -> std::io::Result<()> {
         .split('\n')
     {
         if let Some(stripped) = line.strip_prefix("include:") {
-            cmd.arg("--also-match").arg(absolutelyfy(stripped));
+            cmd.arg("--also-match=".to_owned() + stripped);
         } else if let Some(stripped) = line.strip_prefix("exclude:") {
-            cmd.arg("--exclude").arg(stripped);
+            cmd.arg("--exclude=".to_owned() + stripped);
         }
     }
     let exit = &cmd.spawn()?.wait()?;
     std::process::exit(exit.code().unwrap_or(1))
-}
-
-// converts the relative path `stripped` to an absolute path by prepending the working directory
-fn absolutelyfy(stripped: &str) -> String {
-    let pwd = env::current_dir().unwrap();
-
-    pwd.join(stripped).into_os_string().into_string().unwrap()
 }
--- a/ruby/autobuilder/src/main.rs
+++ b/ruby/autobuilder/src/main.rs
@@ -29,18 +29,11 @@ fn main() -> std::io::Result<()> {
         .split('\n')
     {
         if let Some(stripped) = line.strip_prefix("include:") {
-            cmd.arg("--also-match").arg(absolutelyfy(stripped));
+            cmd.arg("--also-match=".to_owned() + stripped);
         } else if let Some(stripped) = line.strip_prefix("exclude:") {
-            cmd.arg("--exclude").arg(stripped);
+            cmd.arg("--exclude=".to_owned() + stripped);
         }
     }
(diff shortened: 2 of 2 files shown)


nickrolfe (commented):
Looks reasonable to me, but I would like to hear that @aibaars is happy.

From the code, it looks like
```
include:foo/bar
exclude:baz/qux
```
would be translated to
```
--also-match <pwd>/foo/bar/** --also_match !<pwd>/baz/qux/**
```

aibaars:
This looks good to me. I think it would be good to add some test cases to validate the behaviour.

erik-krogh:
> This looks good to me. I think it would be good to add some test cases to validate the behavior.

I'm not sure where or how to add those tests.   

The QL-for-QL autobuilder is tested by it being used in the CI checks.  
That workflow uses both the `paths:` and `paths-ignore:` features, so both includes and excludes are tested that way.

aibaars:
> > This looks good to me. I think it would be good to add some test cases to validate the behavior.
> 
> I'm not sure where or how to add those tests.
> 
> The QL-for-QL autobuilder is tested by it being used in the CI checks. That workflow uses both the `paths:` and `paths-ignore:` features, so both includes and excludes are tested that way.

You're right, there is currently not an obvious place for such a test, except perhaps the "LGTM autobuild tests for Ruby".

erik-krogh:
> > 
> 
> You're right, there is currently not an obvious place for such a test, except perhaps the "LGTM autobuild tests for Ruby".

I've added some tests (see the backref).   
That required changing some things to get the tests to work, so good thing you pointed me to where to add tests.  

Now I'm hoping QL-for-QL still works nicely :crossed_fingers: 

For some reason the `--exclude` flag only works with relative paths, but `--also-match` only works with absolute paths :confused:.

aibaars:
> For some reason the `--exclude` flag only works with relative paths, but `--also-match` only works with absolute paths

@cklin ^ is that expected?

aibaars (approved):

erik-krogh:
> > For some reason the `--exclude` flag only works with relative paths, but `--also-match` only works with absolute paths
> 
> @cklin ^ is that expected?

Here is a good example for reproduction of the error:  
This command finds the `*.rb` files in the current working dir:  `codeql resolve files . --include "*.rb"` 
This finds nothing: `codeql resolve files . --include "*.rb" --also-match "*.rb"`

cklin:
> > For some reason the `--exclude` flag only works with relative paths, but `--also-match` only works with absolute paths
> 
> @cklin ^ is that expected?

I don't think that is expected.  I will look into that later in the day.  Thank you for bringing it up!

erik-krogh:
The `--also-match` fix is not part of a stable release yet, so we have to wait for 2.9.4 before we can continue here.

erik-krogh:
The workflows are now run using CodeQL 2.9.4, but the QL-for-QL autobuilder is still failing, and I'm not sure why.  

I'll look into it later.

nickrolfe (approved):
Thanks for doing this.

PR #9327, opened 2022-05-25
C#: Re-create summary models and include source and sink models as well.

michaelnebel:
In this PR we
- Include the generation of new sources and sinks (very limited outcome as we have only very few sources and sinks in CSV).
- Recreate the summaries. The most recent commit of the summaries was not created based on the newest versions of the model generator. At least some of the removed summaries can be explained by the newly introduced ranking (ie. we don't follow up code path if a handwritten summary exist). This means that eg. summary for IEnumerable.GetEnumerator will cover many existing implementations and the source code will not be used.


michaelnebel:
DCA looks fine. No changes in performance or alerts.

csharp/ql/test/library-tests/dataflow/library/FlowSummariesFiltered.expected
```diff
@@ -275,9 +274,6 @@
 | System.Collections.Generic;CollectionExtensions;false;GetValueOrDefault<,>;(System.Collections.Generic.IReadOnlyDictionary<TKey,TValue>,TKey,TValue);;Argument[1];ReturnValue;taint;generated |
 | System.Collections.Generic;CollectionExtensions;false;GetValueOrDefault<,>;(System.Collections.Generic.IReadOnlyDictionary<TKey,TValue>,TKey,TValue);;Argument[2];ReturnValue;taint;generated |
 | System.Collections.Generic;CollectionExtensions;false;Remove<,>;(System.Collections.Generic.IDictionary<TKey,TValue>,TKey,TValue);;Argument[0].Element;ReturnValue;taint;generated |
-| System.Collections.Generic;CollectionExtensions;false;TryAdd<,>;(System.Collections.Generic.IDictionary<TKey,TValue>,TKey,TValue);;Argument[0].Element;Argument[2];taint;generated |
-| System.Collections.Generic;CollectionExtensions;false;TryAdd<,>;(System.Collections.Generic.IDictionary<TKey,TValue>,TKey,TValue);;Argument[1];Argument[0].Element;taint;generated |
-| System.Collections.Generic;CollectionExtensions;false;TryAdd<,>;(System.Collections.Generic.IDictionary<TKey,TValue>,TKey,TValue);;Argument[2];Argument[0].Element;taint;generated |
```
tamasvajk, line 280: These two rows seem to be correct flow summaries.

csharp/ql/test/library-tests/dataflow/library/FlowSummariesFiltered.expected
```diff
@@ -327,8 +323,6 @@
 | System.Collections.Generic;IList<>;true;get_Item;(System.Int32);;Argument[Qualifier].Element;ReturnValue;value;manual |
 | System.Collections.Generic;IList<>;true;set_Item;(System.Int32,T);;Argument[1];Argument[Qualifier].Element;value;manual |
 | System.Collections.Generic;ISet<>;true;Add;(T);;Argument[0];Argument[Qualifier].Element;value;manual |
-| System.Collections.Generic;KeyValuePair;false;Create<,>;(TKey,TValue);;Argument[0];ReturnValue;taint;generated |
-| System.Collections.Generic;KeyValuePair;false;Create<,>;(TKey,TValue);;Argument[1];ReturnValue;taint;generated |
```
tamasvajk, line 331: These also seem to be correct flows. Why are these removed? Without checking, I expect the `Create<,>` method to be quite simple, so we should be able to identify flow in there.

csharp/ql/test/library-tests/dataflow/library/FlowSummariesFiltered.expected
tamasvajk, line 280: These two rows seem to be correct flow summaries.

csharp/ql/lib/semmle/code/csharp/frameworks/generated/dotnet/Runtime.qll
```diff
@@ -9945,27 +9736,6 @@ private class RuntimeSummaryCsv extends SummaryModelCsv {
...
-        "System;TupleExtensions;false;ToTuple<,,,,,,,>;(System.ValueTuple<T1,T2,T3,T4,T5,T6,T7,System.ValueTuple<T8>>);;Argument[0];ReturnValue;taint;generated",
-        "System;TupleExtensions;false;ToTuple<,,,,,,>;(System.ValueTuple<T1,T2,T3,T4,T5,T6,T7>);;Argument[0];ReturnValue;taint;generated",
-        "System;TupleExtensions;false;ToTuple<,,,,,>;(System.ValueTuple<T1,T2,T3,T4,T5,T6>);;Argument[0];ReturnValue;taint;generated",
-        "System;TupleExtensions;false;ToTuple<,,,,>;(System.ValueTuple<T1,T2,T3,T4,T5>);;Argument[0];ReturnValue;taint;generated",
-        "System;TupleExtensions;false;ToTuple<,,,>;(System.ValueTuple<T1,T2,T3,T4>);;Argument[0];ReturnValue;taint;generated",
-        "System;TupleExtensions;false;ToTuple<,,>;(System.ValueTuple<T1,T2,T3>);;Argument[0];ReturnValue;taint;generated",
-        "System;TupleExtensions;false;ToTuple<,>;(System.ValueTuple<T1,T2>);;Argument[0];ReturnValue;taint;generated",
-        "System;TupleExtensions;false;ToTuple<>;(System.ValueTuple<T1>);;Argument[0];ReturnValue;taint;generated",
```
tamasvajk, line 9968: I also find these removals unexpected. Is it somehow an interplay of `manual` summaries that use `value` steps, and autogenerated flows that use `taint`?

tamasvajk (commented):
There are quite a few unexpected summary deletions in this PR. Should these be investigated? Or should we just update these models, and then investigate and possibly improve the generator in a separate PR?

michaelnebel:
> There are quite a few unexpected summary deletions in this PR. Should these be investigated? Or should we just update these models, and then investigate and possibly improve the generator in a separate PR?

I will try and investigate. It looks like all the removals are related to the heuristic of not using source code, when a summary exist. Will try and elaborate, if I can find any good answers.

csharp/ql/test/library-tests/dataflow/library/FlowSummariesFiltered.expected
michaelnebel: It appears that these rows have disappered since we no longer use the source code for the KeyValuePair constructor as there are CSV rows it.
The rows look like
```
"System.Collections.Generic;KeyValuePair<,>;false;KeyValuePair;(TKey,TValue);;Argument[0];ReturnValue.Property[System.Collections.Generic.KeyValuePair<,>.Key];value;manual",
"System.Collections.Generic;KeyValuePair<,>;false;KeyValuePair;(TKey,TValue);;Argument[1];ReturnValue.Property[System.Collections.Generic.KeyValuePair<,>.Value];value;manual"
```

That is, based on these rows the model generator no longer yields any results for
```csharp
    public static KeyValuePair<TKey, TValue> Create<TKey, TValue>(TKey key, TValue value) =>
        new KeyValuePair<TKey, TValue>(key, value);
```
but a summary will be derived for 
```csharp
    public static TKey Create2<TKey, TValue>(TKey key, TValue value) =>
        new KeyValuePair<TKey, TValue>(key, value).Key;
```

csharp/ql/lib/semmle/code/csharp/frameworks/generated/dotnet/Runtime.qll
michaelnebel: I suspect the issue is the same as above. We have a Field specific summary for the method that is being called and this doesn't propagate the taint. The first example has been added to the Models as Data issue as something we can consider to investigate. Maybe we should do this after making the field based implementation of the summary generator.

michaelnebel:
@tamasvajk : Thank you much for the review. There is identified at least one interesting aspect of the model generator in its current state, which we probably should look into, when the model generator has been made field based.
Thank you for looking into this with care. I think we should merge as is, as this reflects the current state of the generator with the precision that we have.

hvitved (approved):

PR #9333, opened 2022-05-25
Swift: local dataflow

rdmarsh2:


swift/ql/lib/codeql/swift/dataflow/internal/DataFlowPrivate.qll
```diff
@@ -173,7 +221,9 @@ class Unit extends TUnit {
  */
 predicate isUnreachableInCall(Node n, DataFlowCall call) { none() }
 
-newtype LambdaCallKind = TODO_TLambdaCallKind()
+newtype LambdaCallKind =
+  TYieldCallKind() or
+  TLambdaCallKind()
```
MathiasVP, line 226: I forgot to remove these when I ripped out the guts of the Ruby files. Do you mind replacing this with `TODO_TLambdaCallKind()` or similar again?

MathiasVP (approved):
LGTM! Only a tiny not-at-all-blocking comment.

swift/ql/lib/codeql/swift/dataflow/internal/DataFlowPrivate.qll
rdmarsh2: I'll do that as a separate PR

rdmarsh2: Actually just saw there's an autoformat failure so I'll fix it in this

MathiasVP (approved):

PR #9334, opened 2022-05-25
Ruby: Data flow for hash-splat expressions in hash literals

hvitved:
Support for
```rb
hash1 = { :foo => taint }
hash = { :bar => no_taint, **hash1 }
sink(hash[:foo]} # bad
sink(hash[:bar]} # good
```

### DIFF
--- a/ruby/ql/test/library-tests/dataflow/hash-flow/hash_flow.rb
+++ b/ruby/ql/test/library-tests/dataflow/hash-flow/hash_flow.rb
@@ -723,3 +723,27 @@ def m44(x)
 end
 
 m44(:c)
+
+def m45()
+    hash1 = {
+        :a => taint(45.1),
+        :b => 1,
+        :c => taint(45.2)
+    }
+    hash2 = {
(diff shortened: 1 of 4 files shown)


aibaars (approved):

ruby/ql/test/library-tests/dataflow/hash-flow/hash_flow.rb
```diff
@@ -723,3 +723,27 @@ def m44(x)
...
+        :a => taint(45.1),
+        :b => 1,
+        :c => taint(45.2)
+    }
+    hash2 = {
+        :d => taint(45.3),
+        :e => 2,
+        :f => taint(45.4)
+    }
+    hash = { **hash1, :g => taint(45.5), **hash2, :h => 3 }
```
aibaars, line 738: We should probably add some more tests to cover cases where some of the keys of `hash1` and `hash2` or the keyword arguments overlap. This can be done in a follow-up though.

PR #9336, opened 2022-05-26
Swift: CFG for local declarations

MathiasVP:
We were missing some trivial CFG trees to handle local function or type declarations.


redsun82:
@MathiasVP I'm ready to approve it after you resolve the conflicts :slightly_smiling_face:

MathiasVP:
Thanks for reminding me! I think we should still merge it. I'll fix the merge conflicts and see if it's still relevant (which I'm pretty sure it is).

PR #8800, opened 2022-04-21
C++: Add a test for experimental query NoCheckBeforeUnsafePutUser.ql.

geoffw0:
I'm struggling to help with https://github.com/github/codeql/pull/8739/ and think it will be easier to work on the query (and understand each other) with some tests in place.

@4B5F5F4B please could you review these and make sure they are close enough to reality for the purposes of the query.  Note that the code is intentionally a simplification and *not* a copy of any particular real-world code.

PR #9340, opened 2022-05-26
C++: Improve cpp/linux-kernel-no-check-before-unsafe-put-user

geoffw0:
Improve `cpp/linux-kernel-no-check-before-unsafe-put-user` in `experimental`.  Inspired by https://github.com/github/codeql/pull/8739 (which was closed) and the tests I developed for that in https://github.com/github/codeql/pull/8800 .    Further improvements (in particular to the way dataflow is used) are definitely possible, but the goal of this PR is just to get specific knowledge out of my head and into the query.

In the first commit I define the sources properly, rather than coming from any parameter.  In the second commit I adjust the query and test to be closer to reality.


MathiasVP (approved):
Thanks for picking this up!

PR #9341, opened 2022-05-26
Ruby: Make `ActiveRecordInstance` public and fix some misidentifications

alexrford:
The first commit here fixes some cases where we misidentified `self` in a `SingletonMethod` belonging to an active record model class as referring to an instance of that class, rather than referring to the class itself. This bug had also masked cases where we were missing instances of `ActiveRecordModelClass`es, for example, in:
```
class User < ApplicationRecord
  def self.get_some_user
    find(:first)
  end
end
```
We have a call to `find` with an implicit `self` receiver that refers to the `User` class - this was not something that was accounted for in `ActiveRecordModelFinderCall` before, but now is.

The second commit makes `ActiveRecordInstance` publicly available, as it's been requested as a potentially useful class for use in queries.


nickrolfe (approved):
LGTM

PR #9342, opened 2022-05-26
Swift: initial interprocedural data flow implementation

rdmarsh2:
Most of this PR is fairly straightforward pairings of input and output nodes, plus the addition of flow steps for Phi nodes. The main subtlety is the way functions are modeled in the call graph - there's now a non-AST control flow node for each function with the parameter declarations and body as children. The reason these aren't `AstControlFlowNode`s is so that nested functions don't appear in the CFG of their containing function and result in duplicated nodes.


swift/ql/lib/codeql/swift/dataflow/internal/DataFlowDispatch.qll
```diff
@@ -42,47 +43,35 @@ class DataFlowCallable extends TDataFlowCallable {
...
-  DataFlowCallable viableCallable(DataFlowCall call) { none() }
+  DataFlowCallable viableCallable(DataFlowCall call) { 
+    result = TDataFlowFunc(call.asExpr().(CallExpr).getStaticTarget())
+  }
 
   cached
-  newtype TArgumentPosition = TODO_TArgumentPosition()
+  newtype TArgumentPosition =
+    TThisArgument() or
+    TPositionalArgument(int n) { n in [0 .. 100] } // we rely on default exprs generated in the caller for ordering. TODO: compute range properly. TODO: varargs?
```
MathiasVP, line 69: It looks like the compiler is actually giving us a lot of help here. The following example:
```swift
func sum1(_ n : Int, _ numbers: Int...) {}

sum1(1, 2, 3, 4)
```

has the following AST:
```
(call_expr
  (declref_expr ...) // the function
  (tuple_expr // the arguments
    (integer_literal_expr ... 1 ... ) // the first argument
    (vararg_expansion_expr // the varargs
      (array_expr ...
        (integer_literal_expr ... 2 ... )
        (integer_literal_expr ... 3 ... )
        (integer_literal_expr ... 4 ... )
      )
    )
  )
)
```
So if we just replace `100` with `max(ApplyExpr apply | apply.getNumberOfArguments())` I guess we'll get the correct behavior even for varargs.

An even better solution would be to add an `getIndex()` predicate to the `Argument` class and specify the body of this branch as `n = any(Argument arg).getIndex()`.

swift/ql/lib/codeql/swift/dataflow/internal/DataFlowDispatch.qll
```diff
@@ -42,47 +43,35 @@ class DataFlowCallable extends TDataFlowCallable {
...
 
   cached
-  newtype TParameterPosition = TODO_TParameterPosition()
+  newtype TParameterPosition = 
+  TThisParameter() or
+  TPositionalParameter(int n) { n in [0 .. 100] } // TODO: compute range properly
```
MathiasVP, line 74: See my comment for the `ArgumentPosition` case above.

swift/ql/lib/codeql/swift/dataflow/internal/DataFlowPrivate.qll
```diff
@@ -60,6 +74,9 @@ private module Cached {
       or
       // use-use flow
       def.adjacentReadPair(nodeFrom.getCfgNode(), nodeTo.getCfgNode())
+      // step from previous read to Phi node
+      or
```
MathiasVP, line 78:
```suggestion
      or
      // step from previous read to Phi node
```

MathiasVP (dismissed):
Excellent! A couple of non-blocking comments, but otherwise this LGTM.

swift/ql/lib/codeql/swift/dataflow/internal/DataFlowDispatch.qll
MathiasVP: I pushed a commit with this last suggestion in 2106d487857b8a0e4607e0f4b440a77a0f57cefd.

swift/ql/lib/codeql/swift/dataflow/internal/DataFlowDispatch.qll
MathiasVP: I pushed a commit with this in 2106d487857b8a0e4607e0f4b440a77a0f57cefd.

MathiasVP (approved):

swift/ql/lib/codeql/swift/dataflow/internal/DataFlowPrivate.qll
```diff
@@ -31,12 +33,25 @@ private class ExprNodeImpl extends ExprNode, NodeImpl {
...
 
   override string toStringImpl() { result = def.toString() }
+
+  override DataFlowCallable getEnclosingCallable() {
+    result = TDataFlowFunc(def.getBasicBlock().getScope())
+  }
+}
+
+private predicate localFlowSsaInput(Node nodeFrom, Ssa::Definition def, Ssa::Definition next) {
+  exists(BasicBlock bb, int i | SsaImpl::lastRefRedef(def, bb, i, next) |
```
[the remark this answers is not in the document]
hvitved: Or maybe that is no longer the case after "stages are abolished", I am not yet up-to-date on the consequences of that.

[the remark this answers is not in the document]
MathiasVP: Oh, that's a good point. @rdmarsh, I think we should have a `Cached` module https://github.com/github/codeql/blob/main/ruby/ql/lib/codeql/ruby/dataflow/internal/SsaImpl.qll similar to how it's done in Ruby.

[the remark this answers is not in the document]
hvitved: Yes. Just make sure that this new cached module belongs to the same stage as `SsaImplCommon`, possibly by adding an artificial reverse dependency.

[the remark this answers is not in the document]
MathiasVP: Fixed in https://github.com/github/codeql/pull/9961.

PR #9343, opened 2022-05-26
Kotlin: extract methods defined on collections types with their Java signatures

smowton:
Collection, List and Map all define various methods which are either made more generic in Kotlin (e.g. `remove(Object) -> remove(E)`, `containsAll(Collection<?>) -> containsAll(Collection<E>)`), or are made invariant (e.g. `addAll(Collection<? extends E>) -> addAll(Collection<E>)`). This substitutes the types back to their Java signatures,
thereby avoiding differing trap labels and duplicated methods for these types and their descendents.

### DIFF
--- a/java/kotlin-extractor/src/main/kotlin/KotlinUsesExtractor.kt
+++ b/java/kotlin-extractor/src/main/kotlin/KotlinUsesExtractor.kt
@@ -5,7 +5,6 @@ import com.github.codeql.utils.versions.isRawType
 import com.semmle.extractor.java.OdasaOutput
 import org.jetbrains.kotlin.backend.common.extensions.IrPluginContext
 import org.jetbrains.kotlin.backend.common.ir.allOverridden
-import org.jetbrains.kotlin.backend.common.ir.isOverridableOrOverrides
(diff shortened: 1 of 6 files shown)


java/kotlin-extractor/src/main/kotlin/KotlinUsesExtractor.kt
```diff
@@ -9,12 +9,14 @@ import org.jetbrains.kotlin.backend.common.lower.parentsWithSelf
 import org.jetbrains.kotlin.backend.jvm.ir.propertyIfAccessor
 import org.jetbrains.kotlin.builtins.StandardNames
 import org.jetbrains.kotlin.descriptors.*
+import org.jetbrains.kotlin.ir.ObsoleteDescriptorBasedAPI
```
igfoo, line 12: I assume that you can't see an easy way to do this without the obsolete API?

smowton: I'm afraid not; their map of methods that get special generics treatment was keyed by a descriptor. If the obsolete API goes away we can always just copy-paste the thing in in a slightly different form since it's unlikely to change over time.

PR #9349, opened 2022-05-27
Swift: Fix two CFG inconsistencies

MathiasVP:
These two commits reduce the number of [deadEnd](https://github.com/github/codeql/blob/main/swift/ql/lib/codeql/swift/controlflow/internal/ControlFlowGraphImplShared.qll#L1002) inconsistency results on the Signal database from ~3000 to ~200.

- 73c189957ddb7ceb2d01d7b5b5727775894c72ee fixes an issue related to the `StmtCondition` and `ConditionElement` classes: A `StmtCondition` can be either:
  - A boolean check (i.e., the `e` in `if e then { ... } else { ... }`) or a pattern with an initializer (i..e, in `if let x = foo() in { ... } else { ... }`). The important distinction is that the first case returns a `BooleanCompletion`, and the second case returns a `MatchingCompletion`. This first commit ensures that conditional statements such as `if`, `guard`, `while`, etc. all handle both cases.
- 7d36bc1c73468da5d374e378083eaed029097ab3 is a very simple fix to a simple issue where I had used `SimpleCompletion` instead of `NormalCompletion` in the CFG construction for `foreach`.


swift/ql/lib/codeql/swift/controlflow/internal/ControlFlowGraphImpl.qll
```diff
@@ -207,29 +207,70 @@ module Stmts {
     override FailStmt ast;
   }
 
-  private class StmtConditionTree extends AstPostOrderTree {
```
rdmarsh2, line 210: Why the switch from PostOrder to PreOrder here?

MathiasVP: Good question! When it's a PreOrder the final `Completion` will just be whatever `Completion` of the children (i.e., a `BooleanCompletion` if the `StmtCondition` is a boolean expression, and a `MatchingCompletion` if the `StmtCondition` is a declaration with a pattern match). 

In contrast, if we keep it as a `PostOrder` we have to include all the logic about whether the child generates a `BooleanCompletion` or `MatchingCompletion` in the  [`isValidForSpecific`](https://github.com/github/codeql/blob/main/swift/ql/lib/codeql/swift/controlflow/internal/Completion.qll#L45) predicate.

rdmarsh2 (approved):

PR #9352, opened 2022-05-27
Revert "Convert `.codeqlmanifest.json` file to `codeql-workspace.yml`"

henrymercer:
Reverts github/codeql#9314

This is causing some issues with the language tests, so reverting while we resolve them.


smowton (approved):
Please run the integration tests too as these were also failing for qlpack-resolution reasons

PR #9353, opened 2022-05-27
Swift: Extract `isThrowing` and `isAsync`

MathiasVP:
This fills in the `any_function_type_is_throwing` table which is needed to improve CFG precision (I'll open another PR with those changes after this is merged).

I also added an `any_function_type_is_async` table since that was basically a copy-paste of the code required for the `throwing` version. I'm sure we'll need this at some point.


AlexDenisov (approved):
:shipit:

PR #9359, opened 2022-05-27
Migrate `.codeqlmanifest.json` to `codeql-workspace.yml`

henrymercer:
This is the new format for defining a workspace and also lets us add comments e.g. to explain why the ATM model pack isn't included in the workspace by default.


swift/README.md
```diff
@@ -13,7 +13,7 @@ bazel run //swift:create-extractor-pack
 which will install `swift/extractor-pack`.
 
 Using `--search-path=swift/extractor-pack` will then pick up the Swift extractor. You can also use
-`--search-path=.`, as the extractor pack is mentioned in the root `.codeqlmanifest.json`.
+`--search-path=.`, as the extractor pack is mentioned in the root `.codeql-workspace.json`.
```
intrigus-lgtm, line 16:
```suggestion
`--search-path=.`, as the extractor pack is mentioned in the root `.codeql-workspace.yml`.
```

This should be `.yml`, right?

henrymercer: That's correct, thanks for the spot!

aeisenberg (approved):
Looks good. 

`.github/workflows/ruby-build.yml` still generates a codeqlmanifest and that could be converted to a codeql-workspace, but that's really unrelated to this PR and can happen later.

henrymercer:
> `.github/workflows/ruby-build.yml` still generates a codeqlmanifest and that could be converted to a codeql-workspace, but that's really unrelated to this PR and can happen later.

Yep, that was my thinking too — no need to unnecessarily expand the scope here!

PR #9362, opened 2022-05-30
C++: Introduce relation for tracking braced initializers

jketema:
Per the title. This also exposes the relation via `Initializer.isBraced`.

Note that it is safe to merge this without merging the internal PR at the same time.

For DCA experiment see the internal PR.


cpp/ql/lib/change-notes/2022-05-30-braced-initializers.md
```diff
@@ -0,0 +1,4 @@
+---
+category: feature
+---
+* An `isBraced` predicate was added to the `Initializer`class which holds when a C++ braced initializer was used in the initialization.
```
MathiasVP, line 4:
```suggestion
* An `isBraced` predicate was added to the `Initializer` class which holds when a C++ braced initializer was used in the initialization.
```

MathiasVP (approved):
Just one tiny comment. Otherwise this LGTM!

cpp/ql/lib/change-notes/2022-05-30-braced-initializers.md
jketema: Fixed. Needs re-approve now.

MathiasVP (approved):

PR #9364, opened 2022-05-30
Ruby: API graph renaming an documentation

asgerf:
The Ruby version of https://github.com/github/codeql/pull/8606


nickrolfe (approved):
Thanks for doing this.

PR #9368, opened 2022-05-30
Python: Port test model to API graphs

RasmusWL:
I found an old evaluation on LGTM for comparing the difference. Overall it looks good :+1:

One improvement is that the old modeling would also flag custom test runners (subclasses of `unittest.TextTestRunner`) and test loaders (subclasses of `unittest.TestLoader`). Although these are related to testing, they are certainly not a unit test class ;)

The new modeling captures quite a lot of new real testcases, from subclasses such as

- `from flask_testing import TestCase, LiveServerTestCase`
- `from nose.plugins import doctests as npd; npd.DocTestCase`
- `tornado.testing.AsyncHTTPTestCase`
- `from django.test import TestCase`

Overall I think this looks like an improvement in behavior for our test-detection, although it isn't used for as much with Code Scanning as it was for LGTM. It also moves us away from points-to. Seems like a win-win to me :blush:


python/ql/test/library-tests/filters/tests/Filter.expected
```diff
@@ -1,6 +1,13 @@
-| test.py:4:1:4:23 | Class MyTest |
-| test.py:6:5:6:21 | Function test_1 |
-| test.py:9:5:9:21 | Function test_2 |
```
tausbn, line 3: Is the fact that these are no longer matched intended? If so, then maybe we should just get rid of `test.py` entirely?

tausbn (changes requested):
One question, otherwise this looks good to me. :+1:

python/ql/test/library-tests/filters/tests/Filter.expected
RasmusWL: yeah I noted in https://github.com/github/codeql/pull/9368/commits/a8b4b6a374f2528400111f5e7529957d5c068fd1: 

> Notice that although we loose the contrived examples in `test.py`, we do
gain support for real-world test-case construction, which seems worth
the tradeoff.

I think I'll delete that file :+1:

tausbn: Ah! I didn't see that commit message. Thanks!

tausbn (approved):

PR #9369, opened 2022-05-30
Python: API graph renaming and documentation

asgerf:
The Python version of https://github.com/github/codeql/pull/8606


python/ql/lib/semmle/python/ApiGraphs.qll
```diff
@@ -12,76 +12,159 @@ import semmle.python.dataflow.new.DataFlow
...
+   *
+   * There are two kinds of nodes in the API graphs, distinguished by who is "holding" the value:
+   * - **Use-nodes** represent values held by the current codebase, which came from an external library.
+   *   (The current codebase is "using" a value that came from the library).
+   * - **Def-nodes** represent values held by the external library, which came from this codebase.
+   *   (The current codebase "defines" the value seen by the library).
+   *
+   * API graph nodes are associated with data-flow nodes in the current codebase.
+   * (Since external libraries are not part of the database, there is no way to associate with concrete
+   * data-flow nodes from the external library).
```
RasmusWL, line 52: This part is technically not true for Python right now, since we're still installing and extracting external dependencies. I feel like it would be worth pointing this out -- here is my attempt at describing how things work right now, and our goals for the future.

```suggestion
   * (The API graph is designed for usecase where external libraries are not part of the
   * database, meaning there is no way to associate API graph nodes with concrete
   * data-flow nodes from the external library. Note that currently we do include
   * external libraries in Python, but it is our goal to stop doing this, and use API
   * graphs as a means to overcome this problem).
```

python/ql/lib/semmle/python/ApiGraphs.qll
```diff
@@ -113,39 +104,48 @@ module API {
     }
 
     /**
-     * Gets a data-flow node corresponding to the right-hand side of a definition of the API
-     * component represented by this node.
+     * Gets a data-flow node where this value leaves the current codebase and flows into an
+     * external library (or in general, any external codebase).
```
RasmusWL, line 108: I'm curious about your statement in the parenthesis. What would be an example of an interaction with an external codebase, where the external codebase could not be labeled as an external library?

python/ql/lib/semmle/python/ApiGraphs.qll
```diff
@@ -12,76 +12,159 @@ import semmle.python.dataflow.new.DataFlow
...
+     *
+     * # API::moduleImport("re").getMember("escape").asSource()
+     * re.escape
      *
-     * Unlike `getAUse()`, this predicate only gets the immediate references, not the indirect uses
-     * found via data flow. This means that in `x = re.escape` only `re.escape` is a reference
-     * to the `escape` member of `re`, neither `x` nor any node that `x` flows to is a reference to
-     * this API component.
+     * # API::moduleImport("re").getMember("escape").getReturn().asSource()
+     * re.escape()
```
RasmusWL, line 147:
```suggestion
     * # API::moduleImport("re").getMember("escape").getReturn().asSource()
     * x = re.escape()
     *
     * # below, the use of 'x' is not matched by
     * # API::moduleImport("re").getMember("escape").getReturn().asSource()
     * # but matched by 
     * # API::moduleImport("re").getMember("escape").getReturn().getAValueReachableFromSource()
     * print(x)
```

python/ql/lib/semmle/python/ApiGraphs.qll
```diff
@@ -12,76 +12,159 @@ import semmle.python.dataflow.new.DataFlow
...
+     *
+     * This is similar to `asSink()` but additionally includes nodes that transitively reach a sink by data flow.
+     * See `asSink()` for examples.
      */
-    DataFlow::Node getAValueReachingRhs() { result = Impl::trackDefNode(this.getARhs()) }
+    DataFlow::Node getAValueReachingSink() { result = Impl::trackDefNode(this.asSink()) }
 
     /**
-     * Gets an immediate use of the API component represented by this node.
+     * Gets a data-flow node where this value enters the current codebase.
```
RasmusWL, line 136: I quite like that the old QLDocs were clear in the difference between `getAUse` and   `getAnImmediateUse`. So I've come up with suggestions to also make that clear in the new QLDocs.

```suggestion
     * Gets a data-flow node where this value enters the current codebase.
     *
     * Note: Does not include nodes that are transitively reachable by data flow,
     * for that functionality, see `getAValueReachableFromSource`.
```

RasmusWL (changes requested):
A few minor suggestions for QLDocs. Otherwise looks OK to me. 

Typing `getAValueReachableFromSource` instead of `getAUse` will probably take some time to get used to :sweat_smile: 

I see that you added the `no-change-note-required` label, but I don't agree. As an external query-writer, I would be very happy to see this mentioned :blush: Could you add one?

python/ql/lib/semmle/python/ApiGraphs.qll
```diff
@@ -12,76 +12,159 @@ import semmle.python.dataflow.new.DataFlow
...
     /**
-     * Gets a data-flow node corresponding to the right-hand side of a definition of the API
-     * component represented by this node.
+     * Gets a data-flow node where this value leaves the current codebase and flows into an
+     * external library (or in general, any external codebase).
      *
-     * For example, in the property write `foo.bar = x`, variable `x` is the the right-hand side
-     * of a write to the `bar` property of `foo`.
+     * Concretely, this is either an argument passed to a call to external code,
+     * or the right-hand side of a property write on an object flowing into such a call.
```
tausbn, line 111:
```suggestion
     * or the right-hand side of an attribute write on an object flowing into such a call.
```

tausbn (dismissed):
Two minor suggestions, but otherwise this looks good to me! :+1:

python/ql/lib/semmle/python/ApiGraphs.qll
asgerf: Ah yes that's a good point. I've rephrased it slightly differently, PTAL

python/ql/lib/semmle/python/ApiGraphs.qll
asgerf: External libraries are the only use-case today. But in the original use case it could also be the other way around, where the current codebase is a library and the external code-base is a user of that library (this was important when extracting summaries of a library).

There were also some other use cases that haven't been explored much, like modelling database contents with API nodes, so a project that writes to the database will generate a summary that is usable by projects that read from the database, even if the two projects are written in different languages.

python/ql/lib/semmle/python/ApiGraphs.qll
asgerf: Hm, I found this change to be distracting from the main point of the example. At the point where we add `getReturn()` to the access path we suddenly also add `x = ` in front of the example, and then it's easier to overlook the important bit, which is that we're now calling `escape()`.

Instead I added a mention of `getAValueReachableFromSource` after the example, PTAL.

python/ql/lib/semmle/python/ApiGraphs.qll
asgerf: Added something similar as part of the previous comment

python/ql/lib/semmle/python/ApiGraphs.qll
RasmusWL: oh, that's a very good point :+1: Thanks for clarifying

PR #9376, opened 2022-05-30
Clean merge to preserve submodule invariant of internal repo

henrymercer:
This is a clean merge to ensure that `013b4c8768ae41c9ad04b8a4b7d381c0a4ab1e7e` appears on `main`, preserving the variant on the internal repo that all submodule pointer commits in the git history are reachable from `main`.


smowton (approved):
Looks right to me

PR #9378, opened 2022-05-30
Golang : Add query to detect JWT signing vulnerabilities

porcupineyhairs:
Supersedes github/codeql-go#705


porcupineyhairs:
@smowton @owen-mc This PR is now ready for review. I have added the changes you requested in the earlier PR as well as some new sanitizers to reduce FP's.

go/ql/src/experimental/CWE-321/HardcodedKeys.qhelp
```diff
@@ -0,0 +1,50 @@
...
+          <p>
+               A JWT token is safe for enforcing authentication and access control as long as it can't be forged by a malicious actor. However, when a project exposes this secret publicly, these seemingly unforgeable tokens can now be easily forged.
+               Since the authentication as well as access control is typically enforced through these JWT tokens, an attacker armed with the secret can create a valid authentication token for any user and may even gain access to other privileged parts of the application.
+          </p>
+
+     </overview>
+     <recommendation>
+
+          <p>
+               Generating a crytograhically secure secret key during application initialization and using this generated key for future JWT signing requests can prevent this vulnerability.
```
smowton, line 21:
```suggestion
               Generating a cryptograhically secure secret key during application initialization and using this generated key for future JWT signing requests can prevent this vulnerability.
```

go/ql/src/experimental/CWE-321/HardcodedKeysLib.qll
```diff
@@ -0,0 +1,271 @@
...
+            // Model the `Sign` method of the `SigningMethod` interface
+            // type SigningMethod interface {
+            //   Verify(signingString, signature string, key interface{}) error
+            //   Sign(signingString string, key interface{}) (string, error)
+            //   Alg() string
+            // }
+            m.getTarget().hasQualifiedName(pkg, "SigningMethod", "Sign")
+          |
+            this = m.getArgument(1)
+          )
```
smowton, line 87:
```suggestion
          exists(DataFlow::MethodCallNode m |
            // Models the `SignedString` method
            // `func (t *Token) SignedString(key interface{}) (string, error)`
            m.getTarget().hasQualifiedName(pkg, "Token", "SignedString") and
            this = m.getArgument(0)
            or
            // Model the `Sign` method of the `SigningMethod` interface
            // type SigningMethod interface {
            //   Verify(signingString, signature string, key interface{}) error
            //   Sign(signingString string, key interface{}) (string, error)
            //   Alg() string
            // }
            m.getTarget().hasQualifiedName(pkg, "SigningMethod", "Sign") and
            this = m.getArgument(1)
          )
```

go/ql/src/experimental/CWE-321/HardcodedKeysLib.qll
```diff
@@ -0,0 +1,271 @@
...
+      //   }
+      //   return string(ret), nil
+      // }
+      exists(
+        DataFlow::CallNode randint, DataFlow::MethodCallNode bigint, DataFlow::ElementReadNode r
+      |
+        randint.getTarget().hasQualifiedName("crypto/rand", "Int") and
+        bigint.getTarget().hasQualifiedName("math/big", "Int", "Int64") and
+        bigint.getReceiver() = randint.getResult(0).getASuccessor*() and
+        r.reads(this, bigint.getAResult().getASuccessor*())
```
smowton, line 229: This is similar enough to the previous case, you could replace `getASuccessor*` with an augmented get-a-successor that includes the taint step across `big.Int.Int64` (or always include that as a taint step, and use localTaint like the case below)

go/ql/src/experimental/CWE-321/HardcodedKeysLib.qll
```diff
@@ -0,0 +1,271 @@
...
+      //   var bytes = make([]byte, size)
+      //   rand.Read(bytes)
+      //   for i, x := range bytes {
+      //     bytes[i] = characters[x%byte(len(characters))]
+      //   }
+      //   return string(bytes)
+      // }
+      exists(DataFlow::CallNode randread, DataFlow::Node rand, DataFlow::ElementReadNode r |
+        randread.getTarget().hasQualifiedName("crypto/rand", "Read") and
+        TaintTracking::localTaint(randread.getArgument(0).getAPredecessor*().getASuccessor*(), rand) and
```
smowton, line 243:
```suggestion
        TaintTracking::localTaint(any(DataFlow::PostUpdateNode pun | pun.getPreUpdateNode() = randread.getArgument(0)), rand) and
```

go/ql/src/experimental/CWE-321/HardcodedKeysLib.qll
```diff
@@ -0,0 +1,271 @@
...
+        (
+          exists(ModExpr e | e.getAnOperand() = rand.asExpr() |
+            r.reads(this, e.getGlobalValueNumber().getANode())
+          )
+          or
+          r.reads(this.getAPredecessor*(), rand)
```
smowton, line 249: What is this case doing?

go/ql/test/experimental/CWE-321/HardcodedKeysGood.go
```diff
@@ -0,0 +1,38 @@
...
+		num, err := crand.Int(crand.Reader, big.NewInt(int64(len(chars))))
+		if err != nil {
+			return "", err
+		}
+		ret[i] = chars[num.Int64()]
+	}
+	return string(ret), nil
+}
+
+func good() (interface{}, error) {
```
smowton, line 25: Test your various sanitizers too

go/ql/src/experimental/CWE-321/HardcodedKeysLib.qll
porcupineyhairs: I am not sure I follow. Do you mean to say I should create a new taint tracking config and add this step there?

go/ql/test/experimental/CWE-321/HardcodedKeysGood.go
porcupineyhairs: done!

go/ql/src/experimental/CWE-321/HardcodedKeysLib.qll
smowton: No, I mean instead of using getASuccessor* aka localFlow to track flow from rand.Int() etc to Int.Int64() and then using it again to track from Int64() to an array-read, you could extend `AdditionalTaintStep` or `FunctionModel` to make `Int64()` taint-conserving and then use a single localTaint call to track flow from `rand.Int()` through to an array-read that you want to sanitize. Similarly you could make `%` modulus operations taint-conserving so they wouldn't have to be sanitized.

One other related question: it looks like in your sanitizers you sanitize the input to `constant[random]`, whereas logically you should probably sanitize the result instead.

The ideal version would look something like

using source = post-update-node of rand.Read(_) or result of rand.Int() etc, and additional taint steps = big.Int functions and modulus operations, sanitize track taint forwards to any index operand and sanitize that array-read's result.

porcupineyhairs: @smowton I think I partially understand what you wish to say. I have made some changes did you mean something on these lines?

go/ql/src/experimental/CWE-321/HardcodedKeysLib.qll
```diff
@@ -0,0 +1,291 @@
...
+      //   }
+      //   return string(ret), nil
+      // }
+      exists(
+        DataFlow::CallNode randint, string name, DataFlow::ElementReadNode r, DataFlow::Node index
+      |
+        (
+          randint.getTarget().hasQualifiedName("math/rand", name) or
+          randint.getTarget().(Method).hasQualifiedName("math/rand", "Rand", name) or
+          randint.getTarget().hasQualifiedName("crypto/rand", "Int")
```
smowton, line 210: Nitpick: since `name` isn't used in this disjunct, it's better to pull out the list of names as `getARandIntFunctionName()` and use it in the two disjuncts that need it. This may stop the evaluator from pointlessly forming the cartesian product of `randint.getTarget().hasQualifiedName("crypto/rand", "Int")` and `name = [ ... ]`, though in this particular instance it wouldn't hurt too much since it would only mean 12 duplicates of the set of random int function calls.

porcupineyhairs:
@smowton Changes done! PTAL.

go/ql/src/experimental/CWE-321/HardcodedKeysLib.qll
```diff
@@ -0,0 +1,297 @@
...
+        randomValue =
+          any(DataFlow::PostUpdateNode pun |
+            pun.getPreUpdateNode() = getARandReadCall().getArgument(0)
+          )
+      |
+        TaintTracking::localTaint(randomValue, index) and
+        (
+          this.(DataFlow::ElementReadNode).reads(_, randomValue) or
+          any(DataFlow::ElementReadNode r).reads(this, index)
+        )
```
smowton, line 243:
```suggestion
        this.(DataFlow::ElementReadNode).reads(_, index)
```

go/ql/src/experimental/CWE-321/HardcodedKeysLib.qll
```diff
@@ -0,0 +1,297 @@
...
+   * Mark any comparision expression where any operand is tainted as a
+   * sanitizer for all instances of the taint
+   */
+  private class CompareExprSanitizer extends Sanitizer {
+    CompareExprSanitizer() {
+      exists(BinaryExpr c |
+        c.getAnOperand().getGlobalValueNumber() = this.asExpr().getGlobalValueNumber()
+      )
+    }
+  }
```
smowton, line 166: I tried removing this sanitizer, and it doesn't seem to be useful:
* It handles a few cases where an empty string is returned alongside an error value, but this would be better handled by looking for the pattern of an empty string `""` guarded by a test `if x != nil`, where `x` is of error type.
* The other two cases were truly unreachable code: `x := "hello"; if x != "hello" { ... }`, which doesn't seem like a useful test (but perhaps you could change the test to better represent a real-world situation?)

go/ql/src/experimental/CWE-321/HardcodedKeysLib.qll
```diff
@@ -0,0 +1,297 @@
...
+  private class EmptyErrorSanitizer extends Sanitizer {
+    EmptyErrorSanitizer() {
+      exists(ReturnStmt r, DataFlow::CallNode c |
+        c.getTarget().hasQualifiedName("errors", "New") and
+        r.getNumChild() > 1 and
+        r.getAChild() = c.getAResult().getASuccessor*().asExpr() and
+        r.getAChild() = this.asExpr()
+      )
+    }
+  }
```
smowton, line 178: Removing this has no effect on the test results, so it either doesn't work or isn't tested

go/ql/src/experimental/CWE-321/HardcodedKeysLib.qll
smowton: It does not; I've re-verified that this works and will push it myself shortly along with some other last cleanups.

It's possible you're thinking of `this.(DataFlow::ElementReadNode).reads(_, randomValue)` which indeed does not work because we didn't follow the taint trail to `index`.

go/ql/src/experimental/CWE-321/HardcodedKeysLib.qll
```diff
@@ -177,6 +188,24 @@ module HardcodedKeys {
...
   }
 
+  /**
+   * Marks anything returned alongside an error-value that is known
+   * to be non-nil by virtue of a guarding check as harmless.
+   *
+   * For example, `if err != nil { return "", err }` is unlikely to be
+   * contributing a dangerous hardcoded key.
+   */
+  private class ReturnedAlongsideErrorSanitizerGuard extends Sanitizer {
```
porcupineyhairs, line 198: @smowton Is this `guard` not pretty much the same as the `Sanitizer` above?

smowton (approved):

porcupineyhairs:
Thanks for the quick TAT @smowton

PR #9380, opened 2022-05-31
Swift: generate `getParent` implementation

redsun82:
By explicitly marking children in the `schema.yml` file, an internal
`getAnImmediateChild` predicate is implemented, that is in turn used in `AstNode`
to implement `getImmediateParent`. "Immediate" means the closest child or
parent is picked up regardless of node resolution/expr conversion. That is,
the child/parent relation is the one in the so-called "hidden" AST.

This is yet to be used in the control flow library to replace the
hand-rolled implementation.

A further, more complex step is to use the same information to fully
generate the core implementation of `PrintAst` (including the
accessor string). This will be done later.

The `parent` tests use the same swift code as the extractor tests, and
this is currently enforced by `sync-files.py`. Notice that `qltest.sh`
had to be modified to deal with multiple files, which was not working
yet.


MathiasVP:
A couple of notes. I didn't add it as review comments because GitHub renders the diff view incredibly slow for me 😂:

1. I don't think we should expose the `getAChild` predicate on the AST classes. It's very easy to misuse them and generate a giant transitive closure that won't terminate on sufficiently large databases.
2. I'm surprised by definitions like:
```swift
class AbstractTypeParamDeclBase extends @abstract_type_param_decl, TypeDecl {
  cached
  override Element getAChild() {
    none() or
    result = TypeDecl.super.getAChild()
  }
}
```

The `getAChild` predicate is already available on instances of `AbstractTypeParamDeclBase` because `AbstractTypeParamDeclBase` inherits from `TypeDecl`. So why do we need to override it with an implementation that directly forwards to the super class' implementation? I imagine we avoid a tonne of overrides if we don't generate these unnecessary predicates.
3. We shouldn't cache every `getAChild` implementation on the AST classes (for performance reasons). Rather, we should cache some predicate that's defined as a giant disjunction of every `getAChild` result. Something like:
```ql
cached
Element getAChild(Element e) {
  result = e.getAChild() // <--- This predicate shouldn't be cached
}
```

Combining this with 1.: I was hoping that we'd generate a giant predicate like:
```ql
cached Element getAChild(Element e) {
  result = e.(AbstractFuncDecl).getBody()
  or
  result = e.(AbstractFuncDecl).getParam(_)
  or
  // ... etc
}
```
to have 1 cached predicate which is not exposed on any AST classes.

Does that make sense?

swift/ql/lib/codeql/swift/elements/AstNode.qll
```diff
@@ -1,4 +1,6 @@
-// generated by codegen/codegen.py, remove this comment if you wish to edit this file
 private import codeql.swift.generated.AstNode
+private import codeql.swift.generated.Children
 
-class AstNode extends AstNodeBase { }
+class AstNode extends AstNodeBase {
+  AstNode getParent() { result = unique(AstNode x | this = getAChild(x) | x) }
```
MathiasVP, line 5: Like the `getAChild` predicate, I don't think this should be exposed on the AST classes.

In my mind there are two use-cases for `unique`:
- If something is functionally determined by its input, we can apply `unique` to tell the optimizer that there is in fact only one result. This helps the optimizer during join ordering (and probably a bunch of other things I don't know anything about)
- If something is in fact not functionally determined by its input and can have multiple results, but we only want it to have a result if there is one such result. For instance, if we'd like to find the unique `return` statement in a function or something like that.

Does the `unique` fit one of these cases? If so, I think we should add a comment here saying which case.

swift/ql/lib/codeql/swift/generated/Children.qll
```diff
@@ -0,0 +1,210 @@
+// generated by codegen/codegen.py
+import codeql.swift.elements
+
+cached
+Element getAChild(Element e) {
+  not exists(e.getResolveStep()) and
+  (
+    none()
+    or
```
MathiasVP, line 9: Is this `none()` or disjunction necessary?

swift/ql/lib/codeql/swift/generated/Children.qll
```diff
@@ -0,0 +1,210 @@
+// generated by codegen/codegen.py
+import codeql.swift.elements
+
+cached
+Element getAChild(Element e) {
+  not exists(e.getResolveStep()) and
```
MathiasVP, line 6: Do we need this condition? To use this in CFG construction we'd need to get rid of this condition since CFG construction uses the fully converted expression everywhere (i.e., it doesn't skip "hidden" AST nodes).

MathiasVP (commented):
This looks a lot better! I have a couple more comments that I think we should address before we can merge this, I think.

swift/ql/lib/codeql/swift/elements/AstNode.qll
MathiasVP: I think it makes sense to move it to `Children.qll` 👍.

swift/ql/lib/codeql/swift/generated/Children.qll
redsun82: it makes it easier on the generation, as admittedly mustache is not very good at joining with a separator. 
In order to get rid of that I would need to specifically mark the first (or last) child property. Doable, but it would require more code (and tests). This way I can instead blindly append `or result = ...` for each child property, including the first.

swift/ql/lib/codeql/swift/elements/AstNode.qll
MathiasVP: > This comes from from a discussion on slack ([internal link](https://github.slack.com/archives/C02TU8JET9B/p1648643847232739?thread_ts=1648629027.395099&cid=C02TU8JET9B)), and as far as I can tell falls in the first case. I can add a comment.

Ah, right. It does indeed look like this is for the first case.

swift/ql/lib/codeql/swift/generated/Children.qll
redsun82: For the record, what we just discussed online: it seems we need `getAChild/getParent` to ignore hiding of AST nodes. I will rename them to `getAnImmediateChild` and `getImmediateParent` and ignore resolution by using db predicates

swift/codegen/schema.yml
```diff
@@ -514,11 +549,13 @@ BoolPattern:
...
+  _children:
+    sub_pattern: Pattern?
 
 ExprPattern:
   _extends: Pattern
-  sub_expr: Expr
+  _children:
+    sub_expr: Expr
 
 IsPattern:
```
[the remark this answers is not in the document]
redsun82: whoops :slightly_smiling_face: done now

MathiasVP (approved):
LGTM!

PR #9381, opened 2022-05-31
ReDoS: Improve performance in ExponentialBackTracking.qll.

aschackmull:
Two perf fixes:
1. Magic the shortestDistances HOP.
2. Fix the join-order in `isReachableFromFork`.


erik-krogh (approved):
:+1: 

Assuming the tests pass and the evaluations work out (did you start any?).

aschackmull:
> Assuming the tests pass and the evaluations work out (did you start any?).

You should see them now.

erik-krogh:
> You should see them now.

They all look OK.  
There is a performance regression in the Java evaluation, but I expect that to be a fluke.  
(DB caching in DCA :crossed_fingers:)

PR #9384, opened 2022-05-31
Swift: Add qlpacks

MathiasVP:
This is needed for DCA (along with a non-empty `src/queries/` directory which we'll add in a later PR).

### DIFF
--- a/swift/ql/src/qlpack.yml
+++ b/swift/ql/src/qlpack.yml
@@ -1,0 +1,10 @@
+name: codeql/swift-queries
+version: 0.0.0
+groups:
+  - swift
+  - queries
+suites: codeql-suites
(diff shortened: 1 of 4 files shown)


swift/ql/src/qlpack.yml
```diff
@@ -0,0 +1,10 @@
+name: codeql/swift-queries
+version: 0.0.0
+groups:
+  - swift
+  - queries
+suites: codeql-suites
```
geoffw0, line 6: I have this line as well:
```suggestion
suites: codeql-suites
extractor: swift
```

geoffw0 (commented):
LGTM.  One suggestion below, one addition in a PR onto this one.

PR #9393, opened 2022-06-01
C#: ASP.NET Core like members are tainted

michaelnebel:
Consider all properties of ASP.NET Core like objects to also be sources of tainted data.

I have taken the liberty to collapse a couple of remote flow source types into one type (everything related to ASP.Net Core) and then only perceive members of these to be tainted as well instead of doing it generally for all remote flow sources.

### DIFF
--- a/csharp/ql/src/change-notes/2022-06-02-aspnetcoretaintedmembers.md
+++ b/csharp/ql/src/change-notes/2022-06-02-aspnetcoretaintedmembers.md
@@ -1,0 +1,4 @@
+---
+category: minorAnalysis
+---
+* All auto implemented public properties with public getters and setters on ASP.NET Core remote flow sources are now also considered to be tainted.


michaelnebel:
DCA run looks fine.

tamasvajk (commented):
I think we should add a change note. It has an impact on a wide audience.

michaelnebel:
> I think we should add a change note. It has an impact on a wide audience.

Makes sense. I will add a change note.

csharp/ql/test/library-tests/dataflow/flowsources/aspremote/AspRemoteFlowSource.cs
```diff
@@ -0,0 +1,20 @@
...
+        public object Query;
+    }
+
+    public class TestController : Controller
+    {
+        public object MyAction(ViewModel viewModel)
+        {
+            throw null;
+        }
+    }
```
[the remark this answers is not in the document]
michaelnebel: Yes, that is correct. In `Remote.qll` RemoteFlowSource are only extended on types used as parameter types on methods in Controller like classes.

csharp/ql/test/library-tests/dataflow/flowsources/aspremote/AspRemoteFlowSource.cs
```diff
@@ -0,0 +1,20 @@
...
+
+namespace Testing
+{
+
+    public class ViewModel
+    {
+        public string RequestId { get; set; }
+
+        public object Query;
+    }
```
tamasvajk, line 11: Nit picking, but what happens with `get`-only properties and `private` members?

michaelnebel: Hmm, they are also included. There is no further restrictions on the Assignable members. Let's change this such that it is only public and non static assignable members.
Good catch!

michaelnebel: Also - this is not nit picking! Really good that you ask these questions!

michaelnebel: We can also fine-grain it even more. Do you have any special opinion on, which members we should included/exclude?

michaelnebel: @tamasvajk : Thank you!
Ah yes, I think you are right. I read some of this (I know it is old): https://docs.microsoft.com/en-us/archive/msdn-magazine/2012/february/asp-net-mvc-the-features-and-foibles-of-asp-net-mvc-model-binding

It doesn't explicitly state that properties should be get/set, but it seems like it is implicitly assumed and this could serve is a conservative extension of the existing logic (cover more with lower risk of false positives).
My suggestion would be that we take all public properties with public getters and setters and all public non-readonly fields. No statics allowed of course. Then we should be on the safe side and also cover most cases!

tamasvajk: I think we can ignore fields altogether. There are some commonly used attributes to specify from where data should be bound. For example if data is taken from the request body, then `FromBody` is used: https://docs.microsoft.com/en-us/dotnet/api/microsoft.aspnetcore.mvc.frombodyattribute?view=aspnetcore-6.0. These attributes have no `Field` target, so I think we can safely ignore fields.

michaelnebel: Alright! ASP.NET is complicated :-) It looks like the canonical examples all uses public (non static) auto implemented properties with public getters and setters. Maybe that is sufficiently conservative as a starting point.

felickz:
Confirmed - Fixed in 2.10.0 CLI bundle and 0.2.0 codeql/csharp-queries ([changelog](https://github.com/github/codeql/blob/codeql-cli/v2.10.0/csharp/ql/src/CHANGELOG.md#020))
All auto implemented public properties with public getters and setters on [ASP.NET](http://asp.net/) Core remote flow sources are now also considered to be tainted.
Forcing Action to 2.10.0 lights up a bunch of alerts that were missing for me ( [new alerts on complex types](https://github.com/vulna-felickz/DotNetCoreWebApp/pull/4/checks?check_run_id=7140710486))

csharp/ql/lib/semmle/code/csharp/security/dataflow/flowsources/Remote.qll
```diff
@@ -171,6 +171,23 @@ class ActionMethodParameter extends RemoteFlowSource, DataFlow::ParameterNode {
 /** A data flow source of remote user input (ASP.NET Core). */
 abstract class AspNetCoreRemoteFlowSource extends RemoteFlowSource { }
 
+/**
+ * Data flow for AST.NET Core.
```
hvitved, line 175: Typo

csharp/ql/lib/semmle/code/csharp/security/dataflow/flowsources/Remote.qll
```diff
@@ -171,6 +171,23 @@ class ActionMethodParameter extends RemoteFlowSource, DataFlow::ParameterNode {
...
+private class AspNetCoreRemoteFlowSourceMember extends TaintTracking::TaintedMember {
+  AspNetCoreRemoteFlowSourceMember() {
+    this.getDeclaringType() = any(AspNetCoreRemoteFlowSource source).getType() and
+    this.isPublic() and
+    not this.isStatic() and
+    exists(Property p | p = this |
+      p.isAutoImplemented() and p.getGetter().isPublic() and p.getSetter().isPublic()
+    )
+  }
+}
```
hvitved, line 189: Can be simplified a bit:
```ql
private class AspNetCoreRemoteFlowSourceMember extends TaintTracking::TaintedMember, Property {
  AspNetCoreRemoteFlowSourceMember() {
    this.getDeclaringType() = any(AspNetCoreRemoteFlowSource source).getType() and
    this.isPublic() and
    not this.isStatic() and
    this.isAutoImplemented() and
    this.getGetter().isPublic() and
    this.getSetter().isPublic()
  }
}
```

PR #9396, opened 2022-06-01
CI: Add Kotlin and Go to labeler.yml

igfoo:


.github/labeler.yml
```diff
@@ -6,22 +6,32 @@
...
 
 JS:
   - any: [ 'javascript/**/*', '!javascript/ql/experimental/adaptivethreatmodeling/**/*' ]
   - change-notes/**/*javascript*
 
+Kotlin:
+  - java/kotlin-extractor/**/*
+  - java/kotlin-explorer/**/*
+  - java/ql/test/kotlin/**/*
+  - change-notes/**/*java.*
```
aibaars, line 25: We should probably just drop this line. I think the `change-notes` folder is no longer used.

.github/labeler.yml
```diff
@@ -6,22 +6,32 @@
   - csharp/**/*
   - change-notes/**/*csharp*
 
+Go:
+  - go/**/*
+  - change-notes/**/*go.*
```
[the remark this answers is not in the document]
aibaars: Perhaps we should just drop this pattern. The legacy Go changenotes are in `go/old-change-notes` .

.github/labeler.yml
igfoo: I've removed it from Kotlin.

I'll leave looking at whether the changenotes patterns should generally be updated for a future PR.

PR #9397, opened 2022-06-01
Swift: Use the autogenerated `getParent`

MathiasVP:


swift/ql/lib/codeql/swift/controlflow/internal/Scope.qll
```diff
@@ -27,224 +28,13 @@ class Scope extends AstNode instanceof Scope::Range {
...
-    or
-    result = ast.(DeferStmt).getBody()
-    or
-    result = ast.(DoStmt).getBody()
-    or
-    result = ast.(ReturnStmt).getResult()
-    or
-    result = ast.(ThrowStmt).getSubExpr()
+  private AstNode getParentOfAstStep(Element e) {
+    not e instanceof AstNode and
```
redsun82, line 32: this predicate here is because there may be edges in the parent/children tree traversing non-AstNodes? Which ones?

redsun82: anyway it might be good to comment why we need this :slightly_smiling_face:

MathiasVP: Oops, you're right. The return type should indeed be `Element`.

MathiasVP: > There's also ConditionElement I think. We could promote them both to AstNode though. It sounds kinda weird that the child of an AstNode is not an AstNode, doesn't it?

Indeed, it does. I think they should both be `AstNode`s, yeah.

redsun82: anyway, I'm ok with both leaving this as it is or expanding `AstNode` if you think it's the way to go.

PR #9398, opened 2022-06-01
Swift: simplify `GetImmediateParent.qll`

redsun82:
This is a quick follow up to https://github.com/github/codeql/pull/9380, getting rid of the intermediate class implementation there (by leveraging the fact that we can replace `this` with `e` using the fact that it is always the first parameter in the db predicate, and that repeated properties have always the same shape of the parameters to the db predicate).


swift/ql/lib/codeql/swift/generated/GetImmediateParent.qll
```diff
@@ -1,240 +1,230 @@
...
-  }
-}
+import codeql.swift.elements.Element
 
 /**
  * Gets any of the "immediate" children of `e`. "Immediate" means not taking into account node resolution: for example
  * if the AST child is the first of a series of conversions that would normally be hidden away, this will select the
  * next conversion down the hidden AST tree instead of the corresponding fully uncoverted node at the bottom.
- * This predicate is mainly intended to be used to test uniqueness of parents.
+ * Outside this module this file is mainly intended to be used to test uniqueness of parents.
```
MathiasVP, line 8: I think we should be more aggressive with the wording. For some reason, I can't get `Add a suggestion` to work 🤔. I'd replace this line with an initial comment saying "INTERNAL: Do not use". You'll see this [a lot](https://github.com/github/codeql/search?l=CodeQL&q=INTERNAL%3A+Do+not+use) in QL code.

MathiasVP (approved):
LGTM with one tiny comment.

swift/ql/lib/codeql/swift/generated/GetImmediateParent.qll
redsun82: Would you place that on both the predicates, or only on `getAnImmediateChild`?

MathiasVP: I'd place it only on `getAnImmediateChild` as a replacement to the `Outside this module this file is mainly intended to be used to test uniqueness of parents.` part of the comment.

PR #9400, opened 2022-06-01
Kotlin: Fix fake raw type symbols used by the Parcelize plugin

tamasvajk:
The Parcelize plugin generates IR that we would otherwise not see. Specifically, there are method invocations on raw types, such as `java.util.ArrayList.add(Object)`. This PR changes these raw type accesses to their generic counterparts, parameterized with `Object`. So the above becomes `java.util.ArrayList<Object>.add(Object)`.

### DIFF
--- a/java/kotlin-extractor/src/main/kotlin/KotlinUsesExtractor.kt
+++ b/java/kotlin-extractor/src/main/kotlin/KotlinUsesExtractor.kt
@@ -419,31 +419,35 @@ open class KotlinUsesExtractor(
 
     // `typeArgs` can be null to describe a raw generic type.
     // For non-generic types it will be zero-length list.
-    fun addClassLabel(cBeforeReplacement: IrClass, argsIncludingOuterClasses: List<IrTypeArgument>?, inReceiverContext: Boolean = false): TypeResult<DbClassorinterface> {
+    private fun addClassLabel(cBeforeReplacement: IrClass, argsIncludingOuterClasses: List<IrTypeArgument>?, inReceiverContext: Boolean = false): TypeResult<DbClassorinterface> {
         val c = tryReplaceAndroidSyntheticClass(cBeforeReplacement)
-        val classLabelResult = getClassLabel(c, argsIncludingOuterClasses)
+        val p = tryReplaceParcelizeRawType(c)
+        val replacedClass = p?.first ?: c
+        val replacedArgsIncludingOuterClasses = p?.second ?: argsIncludingOuterClasses
+
+        val classLabelResult = getClassLabel(replacedClass, replacedArgsIncludingOuterClasses)
 
         var instanceSeenBefore = true
 
-        val classLabel : Label<out DbClassorinterface> = tw.getLabelFor(classLabelResult.classLabel, {
+        val classLabel : Label<out DbClassorinterface> = tw.getLabelFor(classLabelResult.classLabel) {
             instanceSeenBefore = false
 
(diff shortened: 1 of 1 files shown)


java/kotlin-extractor/src/main/kotlin/KotlinUsesExtractor.kt
```diff
@@ -414,31 +414,35 @@ open class KotlinUsesExtractor(
 
     // `typeArgs` can be null to describe a raw generic type.
     // For non-generic types it will be zero-length list.
-    fun addClassLabel(cBeforeReplacement: IrClass, argsIncludingOuterClasses: List<IrTypeArgument>?, inReceiverContext: Boolean = false): TypeResult<DbClassorinterface> {
+    private fun addClassLabel(cBeforeReplacement: IrClass, argsIncludingOuterClasses: List<IrTypeArgument>?, inReceiverContext: Boolean = false): TypeResult<DbClassorinterface> {
```
smowton, line 417: Recommend renaming the direct parameter `argsIncludingOuterClasses` to `argsBeforeReplacement`, pulling out `tryReplaceAndroidSyntheticClass` and `tryReplaceParcelizeRawType` into a function, and generally avoiding having `c` and `argsIncludingOuterClasses` sitting around in scope but inappropriate to use -- it's better if the 'wrong' symbols have names like `BeforeReplacement` to avoid someone accidentally using them instead of the replaced version.

java/kotlin-extractor/src/main/kotlin/KotlinUsesExtractor.kt
```diff
@@ -448,6 +452,35 @@ open class KotlinUsesExtractor(
             classLabelResult.shortName)
     }
 
+    private fun tryReplaceParcelizeRawType(c: IrClass): Pair<IrClass, List<IrTypeArgument>?>? {
+        if (c.superTypes.isNotEmpty() ||
+            c.origin != IrDeclarationOrigin.DEFINED ||
+            c.hasEqualFqName(FqName("java.lang.Object"))) {
```
smowton, line 458: Why check this when it'll get rejected by the `when` test anyhow?

java/kotlin-extractor/src/main/kotlin/KotlinUsesExtractor.kt
```diff
@@ -716,7 +749,15 @@ open class KotlinUsesExtractor(
                 } else {
                     extractFileClass(dp)
                 }
-            is IrClass -> if (classTypeArguments != null && !dp.isAnonymousObject) useClassInstance(dp, classTypeArguments, inReceiverContext).typeResult.id else useClassSource(dp)
+            is IrClass ->
+                if (classTypeArguments != null && !dp.isAnonymousObject) {
+                    useClassInstance(dp, classTypeArguments, inReceiverContext).typeResult.id
+                } else {
+                    // `inReceiverContext == false` is used unless we have identified that we're dealing with a raw type
+                    // produced by the Parcelize plugin. In that case we're using the original `inReceiverContext`. Note
```
smowton, line 757: If the class should really be the generic specialisation `List<Object>` or so, why aren't we going down the `useClassInstance` route?

tamasvajk: Good point. We could. I'll change this.

smowton (approved):
Looks good!

PR #9405, opened 2022-06-01
Kotlin: Introduce / restore implied wildcard types

smowton:
For Kotlin code, introduce wildcards where the compiler would introduce them when lowering to bytecode (when a type parameter has declaration-site variance, by default only in parameter context, special-casing types that can't be extended / with no supertypes, and obeying the `@JvmWildcard` and `@JvmSuppressWildcards` annotations.

For Java code, use the original Java type to suppress wildcard introduction if the Java type didn't use a wildcard even though it would have been appropriate (for example, Java can ask for exactly `Comparable<String>` even though it would lose nothing to permit `Comparable<? super String>`)


tamasvajk (approved):
Looks plausible to me.

PR #9410, opened 2022-06-02
Swift: extract only primary files

AlexDenisov:
While the (removed) comment is correct and the frontend can be called in different modes, both `swift build` and `xcodebuild` always use `-primary-file` when compiling Swift projects.
The other mode was present only within our test runner (`qltest.sh`), so removing it and doing what the official build systems do simplifies our code base.

Additionally, file archival is now a separate function/operation.


swift/extractor/SwiftExtractor.cpp
```diff
@@ -18,9 +18,7 @@
 
 using namespace codeql;
 
-static void extractFile(const SwiftExtractorConfiguration& config,
-                        swift::CompilerInstance& compiler,
-                        swift::SourceFile& file) {
+static void archiveFile(const SwiftExtractorConfiguration& config, swift::SourceFile* file) {
```
redsun82, line 21: nit: while the swift frontend library does use pointers, in our codebase I'd start using pointers only if we expect them to be nullable (I know, we don't do that yet in the dispatcher). I'd keep `swift::SourceFile` a reference (also for consistency with `config` and `compiler`).

swift/extractor/SwiftExtractor.cpp
```diff
@@ -49,12 +47,16 @@ static void extractFile(const SwiftExtractorConfiguration& config,
               << dstFilePath.str().str() << "': " << ec.message() << "\n";
     return;
   }
+}
 
+static void extractFile(const SwiftExtractorConfiguration& config,
+                        swift::CompilerInstance& compiler,
+                        swift::SourceFile* file) {
```
redsun82, line 54: same here, I would rather use `swift::SourceFile&`

swift/extractor/SwiftExtractor.cpp
AlexDenisov: Just for the record: are you suggesting to use references? 😆 

Ok, will switch back to them.

redsun82 (approved):
LGTM! :rocket:

PR #9411, opened 2022-06-02
Swift: extract system and builtin modules separately

AlexDenisov:


swift/extractor/SwiftExtractor.cpp
```diff
@@ -112,8 +114,21 @@ static void extractFile(const SwiftExtractorConfiguration& config,
...
-    extractFile(config, compiler, s);
+  for (auto& [_, module] : compiler.getASTContext().getLoadedModules()) {
+    // We only extract system and builtin modules here as the other "user" modules can be built
+    // during the build process and then re-used at a later stage. In this case, we extract the
+    // user code twice: once during the module build in a form of a source file, and then as
+    // a pre-built module during building of the dependent source files.
+    if (module->isSystemModule() || module->isBuiltinModule()) {
+      llvm::SmallVector<swift::Decl*> decls;
+      module->getTopLevelDecls(decls);
+      extractDeclarations(config, compiler, module->getModuleFilename(), decls);
```
redsun82, line 125: if we add extraction of `ModuleDecl`, we could fill `decls` with just that (and we would have the module declaration itself also filled up). We can do that later, but then I'd put a TODO here maybe

swift/ql/test/library-tests/controlflow/graph/Cfg.expected
```diff
@@ -4940,8 +4940,8 @@ cfg.swift:
 #-----|  -> exit deinit
 
 #  378| deinit
-#-----|  -> call to ...
 #-----|  -> TBD (OtherConstructorDeclRefExpr)
+#-----|  -> call to ...
```
redsun82, line 4944: uh, why is the order changing here?

AlexDenisov: That's a good question to which I don't have an answer right now 😞

AlexDenisov: That's because now we extract more: I just removed the module extraction and diff is gone.

swift/extractor/SwiftExtractor.cpp
AlexDenisov: Added TODO.

swift/ql/test/library-tests/controlflow/graph/Cfg.expected
AlexDenisov: well, it's flaky now 😞

redsun82 (approved):
:shipit:

PR #9413, opened 2022-06-02
Misc: query-list should run on `codeql-cli/*` tags

RasmusWL:
Since go doesn't need to be cloned with the same tag, that makes this change much easier to implement in practice :muscle:


adityasharad (approved):
Great idea.

PR #9415, opened 2022-06-02
Add TCP and UDP socket client taint sources

JarLob:
In order to support scenarios as:

```cs
            using (var tcpConn = new TcpClient("host.example.org", 39544))
            {
                /* read input from socket */
                using (StreamReader sr = new StreamReader(tcpConn.GetStream()))
                {
                    Sink(sr.ReadLine());
                }
            }
```

Also added StreamReader summary until MaD generator is fixed


JarLob:
Question, since `GetStream` is a source of taint by itself are any changes to `"System.Net.Sockets;TcpClient;false;GetStream;();;Argument[Qualifier];ReturnValue;generated:taint",` in runtime.qll needed?

michaelnebel:
> Question, since `GetStream` is a source of taint by itself are any changes to `"System.Net.Sockets;TcpClient;false;GetStream;();;Argument[Qualifier];ReturnValue;generated:taint",` in runtime.qll needed?

I don't think so.
Maybe also create small testcase for this to make sure that source models and taint steps are applied correctly during taint tracking.

michaelnebel:
The new models looks plausible to me.
Unit test failures needs to be inspected.

csharp/ql/lib/semmle/code/csharp/frameworks/system/io/StreamReader.qll
```diff
@@ -0,0 +1,31 @@
...
+class SystemIOStreamReader extends SystemIOClass {
+  SystemIOStreamReader() { this.hasName("StreamReader") }
+}
+
+/** Data flow for `System.IO.StreamReader`. */
+private class SystemIOStreamReaderFlowModelCsv extends SummaryModelCsv {
+  override predicate row(string row) {
+    row =
+      [
+        "System.IO;StreamReader;false;StreamReader;(System.IO.Stream);;Argument[0];Argument[Qualifier];taint",
```
michaelnebel, line 17: Please note that the content of this [PR](https://github.com/github/codeql/pull/9025) changes the CSV format. All the entered rows here needs to be postfixed with ";manual"

JarLob: Thanks, for the heads up. I'll rebase. I expect to get back to this in a week.

JarLob:
Since I have added only source and summary it is quite different form the usual add new query test. There are no tests for Compression.qll for an instance. I'm looking for an example of source and summary tests I could add. Do you have and idea?

michaelnebel:
> Since I have added only source and summary it is quite different form the usual add new query test. There are no tests for Compression.qll for an instance. I'm looking for an example of source and summary tests I could add. Do you have and idea?

If there doesn't already exist a test then maybe it is fine to run the query against a database with a known vulnerability that should now be discovered by the addition of these models (@hvitved : Or is there another policy on what to do?).

The existing failing unit tests should be updated to reflect the new model changes (let me know, if you need any help with that.)

csharp/ql/lib/semmle/code/csharp/frameworks/system/io/StreamReader.qll
```diff
@@ -0,0 +1,31 @@
...
+        "System.IO;StreamReader;false;StreamReader;(System.String,System.Text.Encoding,System.Boolean,System.IO.FileStreamOptions);;Argument[0];Argument[Qualifier];taint;manual",
+        "System.IO;StreamReader;false;StreamReader;(System.String,System.Text.Encoding,System.Boolean,System.Int32);;Argument[0];Argument[Qualifier];taint;manual",
+        "System.IO;StreamReader;false;StreamReader;(System.IO.Stream,System.Text.Encoding,System.Boolean,System.Int32);;Argument[0];Argument[Qualifier];taint;manual",
+        "System.IO;StreamReader;false;StreamReader;(System.IO.Stream,System.Text.Encoding,System.Boolean);;Argument[0];Argument[Qualifier];taint;manual",
+        "System.IO;StreamReader;false;StreamReader;(System.String,System.Text.Encoding,System.Boolean);;Argument[0];Argument[Qualifier];taint;manual",
+        "System.IO;StreamReader;false;StreamReader;(System.String,System.IO.FileStreamOptions);;Argument[0];Argument[Qualifier];taint;manual",
+        "System.IO;StreamReader;false;StreamReader;(System.String,System.Boolean);;Argument[0];Argument[Qualifier];taint;manual",
```
hvitved, line 24: `Argument[Qualifier]` needs to be `Argument[this]` instead.

michaelnebel:
> @michaelnebel I was getting `Wrong number of columns in source model row, expected 9, got 8 in System.Net.Sockets;TcpClient;false;GetStream;;;ReturnValue;remote. |` so I have added `;manual` in attempt to fix that, now it still complains `| Invalid kind "remote" in source model. |`. Any advise what is wrong?

Yes, there is a validation of the allowed *kinds* used in all types of models. The only allowed kinds for source models are *local* and *file*, but you are welcome to introduce a new *kind* for source models. 
If you want to allow new source kinds, you need to change the validation here: https://github.com/github/codeql/blob/28b7ab7fbeab2a5852b0743b180302956c47a6f9/csharp/ql/lib/semmle/code/csharp/dataflow/ExternalFlow.qll#L373

hvitved:
> Finally, everything is green.

I'll start a performance/results comparison job.

hvitved (dismissed):

michaelnebel:
> It fails... @michaelnebel could you please suggest what is wrong?

Yes, we have migrated all models as data models to be defined in data extensions instead of inlined in the code.
That is, they need to written in YAML format instead and put in the `lib/ext` folder.
There is a a file pr. namespace (named `<namespace>.model.yml`.

csharp/ql/lib/ext/System.IO.StreamReader.model.yml
```diff
@@ -0,0 +1,18 @@
+extensions:
+  - addsTo:
+      pack: codeql/csharp-all
+      extensible: extSummaryModel
+    data:
+      - ["System.IO", "StreamReader", False, "StreamReader", "",                                                                                "", "Argument[0]", "Argument[this]", "taint", "manual"]
```
michaelnebel, line 6: These models should be included in `System.IO.model.yml`.
The convention is that it is the namespace that decides which file the models should be added to (corresponding to the first component of the model).
Otherwise everything looks good!

csharp/ql/lib/semmle/code/csharp/security/dataflow/flowsources/Remote.qll
```diff
@@ -225,3 +226,31 @@ class AspNetCoreActionMethodParameter extends AspNetCoreRemoteFlowSource, DataFl
...
+
+  override string getSourceType() { result = "external" }
+}
+
+/** TcpClient `GetStream` call. */
+private class TcpClientSourceModelCsv extends SourceModelCsv {
+  override predicate row(string row) {
+    row =
+      [
+        "System.Net.Sockets;TcpClient;false;GetStream;;;ReturnValue;remote"
```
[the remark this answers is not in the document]
michaelnebel: That was an outdated dangling comment. Not sure, what happened there.
Nothing further needs to be moved - the model you see for `GetStream` in `dotnet_runtime.model.yml` is an autogenerated summary model.
I will just rebase the PR and delete a couple of things.

hvitved (approved):

michaelnebel:
DCA shows
(1) No performance regressions.
(2) A single new alert in mono/mono where the newly added `GetStream` is a remote source. Assuming we trust the query the alerts seems plausible.
Merging now.

PR #9421, opened 2022-06-02
Inline Expectation Tests: Allow `tag[foo bar]`

RasmusWL:
This is partly motivated by the MaD tests which looks much better now in my opinion.

I also wanted this for testing argument passing. In Python we're adopting the same argument positions as Ruby has [here](https://github.com/github/codeql/blob/4f3751dfea06c84b97685301739f5303b9678a18/ruby/ql/lib/codeql/ruby/dataflow/internal/DataFlowDispatch.qll#L508-L540)

So it would be nice if `arg[keyword foo]=...` was allowed, without having to transform the `toString()` result of an argument position into something without a space.

~Draft PR until https://github.com/github/codeql/pull/9420 is merged~

### DIFF
--- a/csharp/ql/test/TestUtilities/InlineExpectationsTest.qll
+++ b/csharp/ql/test/TestUtilities/InlineExpectationsTest.qll
@@ -242,17 +242,21 @@ private string getColumnString(TColumn column) {
  * more comma-separated tags optionally followed by `=` and the expected value.
  *
  * Tags must be only letters, digits, `-` and `_` (note that the first character
- * must not be a digit), but can contain anything enclosed in a square brackets.
+ * must not be a digit), but can contain anything enclosed in a single set of
+ * square brackets.
  *
  * Examples:
  * - `tag`
  * - `tag=value`
  * - `tag,tag2=value`
  * - `tag[foo bar]=value`
+ *
+ * Not allowed:
+ * - `tag[[[foo bar]`
  */
 private string expectationPattern() {
   exists(string tag, string tags, string value |
-    tag = "[A-Za-z-_](?:[A-Za-z-_0-9]|\\[[^\\]]*\\])*" and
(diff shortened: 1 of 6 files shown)


smowton (approved):

alexrford (approved):
Ruby 👍

MathiasVP (dismissed):
C/C++ 👍

(These tags are starting to become very unwieldy, but I guess someone will write a proper parser in QL once we find a bug in the regex 😂)

csharp/ql/test/TestUtilities/InlineExpectationsTest.qll
```diff
@@ -239,12 +239,20 @@ private string getColumnString(TColumn column) {
...
+ * Examples:
+ * - `tag`
+ * - `tag=value`
+ * - `tag,tag2=value`
+ * - `tag[foo bar]=value`
  */
 private string expectationPattern() {
   exists(string tag, string tags, string value |
-    tag = "[A-Za-z-_][A-Za-z-_0-9]*" and
+    tag = "[A-Za-z-_](?:[A-Za-z-_0-9]|\\[[^\\]]*\\])*" and
```
michaelnebel, line 255:
```java
\\[[^\\]]*\\])*
```
If I managed to parse this correctly, it means that you match anything enclosed in [X], but where X can't contain "]", but the regexp matches [[[foo bar]. Maybe consider avoiding the use of "[" in X as well?

michaelnebel (dismissed):
C# LGTM

csharp/ql/test/TestUtilities/InlineExpectationsTest.qll
RasmusWL: I think that's a good suggestion, so I implemented this :+1:

tausbn (approved):

michaelnebel (approved):
C# LGTM!

PR #9430, opened 2022-06-03
Swift: add `UnknownLocation`

redsun82:
`getLocation()` will now exists for all entities. When there is no
valid location, the location will still not be emitted in the DB, but
on the QL side we will then assign a special `UnknownLocation` with
empty filename and 0 for line/column start/end.

This unknown location is currently emitted (with a unique `@` key) at
the start of every extraction, but we can move it elsewhere (and
possibly in a unique global trap file) at a later stage, possibly after
or when we rework the trap file strategy.

This should solve flakiness that was observed on the control flow tests,
which is probably caused by the `nodes` predicate in the `TestOutput`
class in `ControlFlowGraphImplShared.qll` not able to assign a proper
rank when the node does not have a location.


MathiasVP (approved):
LGTM!

PR #9432, opened 2022-06-03
RB: Adding decompression-api to experimental ruleset

thiggy1342:
This PR aims to add a query to catch cases where user-controlled input flows into the first argument of common Ruby decompression libraries. The aim is to catch cases where a user may be able to upload a highly compression object and have it extracted server-side, resulting in a denial of service.

The goal was to catch the uses of `Zlib` and Ruby `Zip` libraries calling their respective inflate or extract functions. I also hope to include some common sanitizer guards that will reduce false positives if the user input is compared against strings or an array of strings before flowing into the sink. I've also added some tests to help maintain robustness of the query over time.

I did run into a few areas where I could use some guidance:
1. Tests are currently not passing as expected for uses of `Zip::Entry::Extract`, so I think my query isn't quite working for these use cases.
2. ~The sanitizer guard tests also don't seem to be passing as expected. Is there a step that I'm missing in configuring those in my query?~ Update: I think I fixed my test for these?
3. I tried to get my metadata to line up with the contribution guidelines, but I'm a little confused about the security-severity score. Because I've tagged this with a CWE, will a score be automatically assigned or should I be able to calculate that number? Is the same true of the precision tag? I've marked it as medium now, but not sure if there's a more concrete way to set that.

Thank you in advance for any guidance with this!


ruby/ql/src/experimental/decompression-api/DecompressionAPI.ql
```diff
@@ -0,0 +1,57 @@
...
+import codeql.ruby.dataflow.BarrierGuards
+import codeql.ruby.TaintTracking
+import DataFlow::PathGraph
+
+class DecompressionAPIUse extends DataFlow::Node {
+
+    // this should find the first argument of Zlib::Inflate.inflate or Zip::File.extract
+    DecompressionAPIUse() {
+        this = API::getTopLevelMember("Zlib").getMember("Inflate").getAMethodCall("inflate").getArgument(0) or
+        this = API::getTopLevelMember("Zip").getMember("File").getAMethodCall("open").getArgument(0) or
```
thiggy1342, line 25: After looking at this again, I think that this constraint is out of scope of this specific query targeting CWE-409. This would be better suited in a different query targeting path traversal vulnerabilities in these two libraries.

ruby/ql/test/query-tests/security/decompression-api/decompression_api.rb
```diff
@@ -0,0 +1,42 @@
...
+        end
+    end
+
+    def safe_zlib_unzip
+        Zip::Entry.extract("testfile.gz")
+    end
+
+    def sanitized_zlib_unzip
+        fname = params[:fname]
+        if fname == "safe_file.gz"
```
thiggy1342, line 30: I think I was able to get this test fixed where it wouldn't result in a hit from my query, but I'm still a little confused why it wasn't working before. Looking at the implementation for `StringConstCompare`, it looks like it should be able to see the comparison regardless of which side of the assignment operator it's on, but that didn't seem to be the case. Is it because the tainted value must first be assigned to `fname` in order for a DataFlow node to exist?

aibaars (commented):
Looking at the CWE and description of this query I understand that the objective is to find cases where user-provided contents is decompressed. 

It looks like you confused "user supplied content" and "user supplied file path". A "user supplied file path" is probably not very risky in this context as it requires the malicious zip file to already be present on the file system of the server. Allowing a user to open arbitrary files is of course pretty bad, but not the subject of this query. That's what the path traversal queries are for.

User supplied compressed data would typically be the body of a POST/PUT request in `application/zip` format or similar. This is indeed risky as a user might supply a "zip bomb" to such an end-point and cause a high load on the server. Having API end-points accept compressed contents is fairly common, and therefore the query could be a little "noisy". 

It's not clear to me how we can do better though. For the related "XML bomb" vulnerability, the issue can be mitigated by configuring the XML parser to disallow certain types of entity expansion, and it is fairly easy for a query to detect such measures. I don't think the extraction libraries have any built-in protection mechanisms. I guess wrapping the stream in a sort of "limited reader" could prevent problems by aborting the decompression when the decompressed size hits the limit. For zip files we could looks for checks on the `entry.size` value. See also: https://github.com/rubyzip/rubyzip/issues/193

thiggy1342:
@aibaars Yep, you're totally right that this query is making the wrong assumptions about what flows into these sinks 🤦🏻 . This PR I've also opened is better for targeting the path traversal, and so I'll need to rethink this one a bit: https://github.com/github/codeql/pull/9477

The original query that was the inspiration for this was a pretty noisy regex check that just looked for any call to `Zlib::Inflate`, but I think you're right that we probably don't want this to be that noisy. I'll noodle on this a bit further and see if there's a way this can be improved. If I can't come up with something, then perhaps this query is better off as a member of our custom query set and not a great candidate for the experimental suite here.

aibaars:
> If I can't come up with something, then perhaps this query is better off as a member of our custom query set and not a great candidate for the experimental suite here.

I think it is fine as an experimental query with a `medium` or `low` precision. We can always refine it further to improve precision and detect more mitigations.

ruby/ql/src/experimental/decompression-api/DecompressionApi.ql
```diff
@@ -0,0 +1,56 @@
...
+import codeql.ruby.dataflow.BarrierGuards
+import codeql.ruby.TaintTracking
+import DataFlow::PathGraph
+
+class DecompressionApiUse extends DataFlow::Node {
+
+    // this should find the first argument of Zlib::Inflate.inflate or Zip::File.extract
+    DecompressionApiUse() {
+        this = API::getTopLevelMember("Zlib").getMember("Inflate").getAMethodCall("inflate").getArgument(0) or
+        this = API::getTopLevelMember("Zip").getMember("Entry").getAMethodCall("extract").getArgument(0)
```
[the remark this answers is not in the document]
thiggy1342: Yep, you're right. I'm struggling to think of a good way to query for this is terms of tainted data flowing to `.extract` if we're not constraining to an argument. From what I can find, it looks like [this method](https://www.rubydoc.info/gems/rubyzip/1.2.0/Zip/CentralDirectory#read_from_stream-instance_method) is called under the hood in `Zip::File.open`, so would it make sense to query for tainted data that flows to that or `Zip::File.open`? Or does that seem like I'm going about this the wrong way and should just alert for any usage of `Zip::Entry.extract` since this query is shaping up to be pretty noisy anyway?

ruby/ql/src/experimental/decompression-api/DecompressionApi.ql
```diff
@@ -0,0 +1,56 @@
...
+
+    // our Decompression APIs defined above will the the sinks we use for this query
+    override predicate isSink(DataFlow::Node sink) {
+        sink instanceof DecompressionApiUse
+    }
+
+    // I think it would also be helpful to reduce false positives by adding a simple sanitizer config in the event
+    // that the code first checks the file name against a string literal or array of string literals before calling
+    // the decompression API
+    override predicate isSanitizerGuard(DataFlow::BarrierGuard guard) {
```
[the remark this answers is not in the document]
thiggy1342: Yep, this was my misunderstanding.

ruby/ql/src/experimental/decompression-api/DecompressionApi.ql
```diff
@@ -0,0 +1,56 @@
...
+        guard instanceof StringConstCompare or
+        guard instanceof StringConstArrayInclusionCall
+    }
+  }
+  
+
+from Configuration config, DataFlow::PathNode source, DataFlow::PathNode sink
+where
+  config.hasFlowPath(source, sink)
+select sink.getNode(), source, sink, "This call to $@ is unsafe because user-controlled data is used to set the object being decompressed, which could lead to a denial of service attack or malicious code extracted from an unknown source.", sink.getNode().asExpr().getExpr().getParent().toString(), sink.getNode().asExpr().getExpr().getParent().toString()
```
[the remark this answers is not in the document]
thiggy1342: Ah nice, this is a good tip!

ruby/ql/src/experimental/decompression-api/DecompressionApi.ql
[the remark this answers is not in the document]
aibaars: treating `Zip::File.open` and `Zip::File.new` as a sinks makes sense. We can add some barrier guard to exclude flow of string values (file paths). Those should be caught the path traversal query. In addition you probably want to add `Zip::File.open_buffer`; for this sink there should be no barrier guard, because a string value is treated as contents not as file path.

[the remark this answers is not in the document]
aibaars: Tracking `Zip::Entry.extract` instead would avoid flagging up cases where a zip file is opened but no data is extracted. However, opening a zip file and not extracting anything from it is likely very rare in practice. In addition there are many different ways data can be extracted from a `Zip::Entry`, so tracking `Zip::File.open` should be simpler. 

We might want to augment the query later on with some additional checks to reduce false positives. For example if the code has tests that `Zip::Entry.size` is within reasonable bounds or the compression rate is "sane" then that mitigates the effect of a zip bomb.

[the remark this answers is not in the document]
thiggy1342: Ok, this makes sense. I'm going to wait until https://github.com/github/codeql/pull/9477 is finished and merged and then I'll model `Zip::File.open_buffer` in Archives.qll and add it to this query. I've also dropped the barrier guards here since they don't make sense in this context.

ruby/ql/src/experimental/decompression-api/DecompressionApi.ql
```diff
@@ -0,0 +1,43 @@
...
+import ruby
+import codeql.ruby.ApiGraphs
+import codeql.ruby.DataFlow
+import codeql.ruby.dataflow.RemoteFlowSources
+import codeql.ruby.dataflow.BarrierGuards
+import codeql.ruby.TaintTracking
+import DataFlow::PathGraph
+
+class DecompressionApiUse extends DataFlow::Node {
+  // this should find the first argument of Zlib::Inflate.inflate or Zip::File.extract
```
aibaars, line 21: The comment needs updating
```suggestion
  // this should find the first argument of Zlib::Inflate.inflate
```

ruby/ql/src/experimental/decompression-api/DecompressionApi.ql
```diff
@@ -0,0 +1,56 @@
...
+    // I think it would also be helpful to reduce false positives by adding a simple sanitizer config in the event
+    // that the code first checks the file name against a string literal or array of string literals before calling
+    // the decompression API
+    override predicate isSanitizerGuard(DataFlow::BarrierGuard guard) {
+        guard instanceof StringConstCompare or
+        guard instanceof StringConstArrayInclusionCall
+    }
+  }
+  
+
```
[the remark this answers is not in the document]
aibaars: Let's get rid of the rather cryptic `sink.getNode().asExpr().getExpr().getParent()` expressions. Assuming we use the first of the definitions of `DecompressionApiUse` above it would be something like `sink.getNode().(DecompressionApiUse),getCall()` and `sink.getNode().(DecompressionApiUse),getCall().getMethodname()`

aibaars (commented):
This looks good to me . A few suggestions for improvements. 

I also recommend adding a `ruby/ql/src/experimental/decompression-api/DecompressionApi.qhelp` file documenting the query. This is optional for an experimental query, but having additional documentation improves the user experience . You can "compile" qhelp files locally using `codeql generate query-help --format=markdown` . There is also a CI job that adds a preview of the rendered qhelp file to the pull request.

aibaars (approved):
Looks good to me. Just needs a fix to the qhelp file.

ruby/ql/src/experimental/decompression-api/DecompressionApi.qhelp
```diff
@@ -0,0 +1,28 @@
...
+  </recommendation>
+  <example>
+    <p>
+      In this example, the size of the input buffer chunks and total size are checked before each chunk is written to the output.
+    </p>
+    <sample src="examples/decompress.rb" />
+  </example>
+  
+  <references>
+  <a href="https://cwe.mitre.org/data/definitions/409.html">https://cwe.mitre.org/data/definitions/409.html</a>
```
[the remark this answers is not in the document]
thiggy1342: ah ok thanks!

ruby/ql/src/experimental/decompression-api/DecompressionApi.ql
```diff
@@ -0,0 +1,51 @@
...
+import codeql.ruby.ApiGraphs
+import codeql.ruby.DataFlow
+import codeql.ruby.dataflow.RemoteFlowSources
+import codeql.ruby.TaintTracking
+import DataFlow::PathGraph
+
+class DecompressionApiUse extends DataFlow::Node {
+  private DataFlow::CallNode call;
+
+  // this should find the first argument of Zlib::Inflate.inflate
```
aibaars, line 22: Shall we update or remove the comment ?

ruby/ql/src/experimental/decompression-api/DecompressionApi.ql
```diff
@@ -0,0 +1,51 @@
...
+    )
+  }
+
+  DataFlow::CallNode getCall() { result = call }
+}
+
+class Configuration extends TaintTracking::Configuration {
+  Configuration() { this = "DecompressionApiUse" }
+
+  // this predicate will be used to contstrain our query to find instances where only remote user-controlled data flows to the sink
```
aibaars, line 37:
```suggestion
  // this predicate will be used to constrain our query to find instances where only remote user-controlled data flows to the sink
```

ruby/ql/src/experimental/decompression-api/DecompressionApi.ql
```diff
@@ -0,0 +1,51 @@
...
+  override predicate isSource(DataFlow::Node source) { source instanceof RemoteFlowSource }
+
+  // our Decompression APIs defined above will the the sinks we use for this query
+  override predicate isSink(DataFlow::Node sink) {
+    sink.(DataFlow::CallNode) instanceof DecompressionApiUse
```
aibaars, line 42: Let's not restrict the sink to be a call node. Otherwise we can only find things like `inflate(some_call())`
```suggestion
    sink instanceof DecompressionApiUse
```

ruby/ql/test/query-tests/security/decompression-api/decompression_api.rb
```diff
@@ -0,0 +1,21 @@
+class TestController < ActionController::Base
+    # this should get picked up
+    def unsafe_zlib_unzip
+        Zlib::Inflate.inflate(params[:file])
+    end
```
aibaars, line 5: Let's write the test case like to to make sure we also find indirect uses of `params[..]`
```suggestion
    def unsafe_zlib_unzip
        path = params[:file]
        Zlib::Inflate.inflate(path)
    end
```

ruby/ql/src/experimental/decompression-api/DecompressionApi.ql
thiggy1342: Updated the comments to reflect the new state of the query

ruby/ql/src/experimental/decompression-api/DecompressionApi.ql
```diff
@@ -0,0 +1,52 @@
...
+class Configuration extends TaintTracking::Configuration {
+  Configuration() { this = "DecompressionApiUse" }
+
+  // this predicate will be used to constrain our query to find instances where only remote user-controlled data flows to the sink
+  override predicate isSource(DataFlow::Node source) { source instanceof RemoteFlowSource }
+
+  // our Decompression APIs defined above will be the sinks we use for this query
+  override predicate isSink(DataFlow::Node sink) {
+    sink instanceof DecompressionApiUse
+  }
```
aibaars, line 44:
```suggestion
  override predicate isSink(DataFlow::Node sink) { sink instanceof DecompressionApiUse }
```

aibaars (dismissed):
Let's update the test output and format the query.

aibaars:
@thiggy1342 I guess we need to update the `edges` and `nodes` section in the test output . See also: https://github.com/github/codeql/runs/6970132383?check_suite_focus=true#step:5:222

aibaars (approved):

PR #9455, opened 2022-06-07
Fix `codeql/suite-helpers` dependency for Go

dbartol:
Now that Go is part of the `github/codeql` repo, we always pick up its dependency on `codeql/suite-helpers` from within the repo. Rather than specify a specific version dependency, we just use `"*"` like we do for all other in-repo dependencies.


igfoo:
I think this needs a rebase.

PR #9457, opened 2022-06-08
JS: Some more improvements to d.ts file analysis

asgerf:
Some more improvements needed for model generation, mainly by improving analysis of `.d.ts` files.

[Evaluation](https://github.com/github/codeql-dca-main/tree/data/asgerf/madman-prep2__default__code-scanning__3/reports) looks OK

### DIFF
--- a/javascript/extractor/src/com/semmle/ts/extractor/TypeScriptASTConverter.java
+++ b/javascript/extractor/src/com/semmle/ts/extractor/TypeScriptASTConverter.java
@@ -591,7 +591,7 @@ public class TypeScriptASTConverter {
         return convertTryStatement(node, loc);
       case "TupleType":
         return convertTupleType(node, loc);
-      case "NamedTupleMember": 
+      case "NamedTupleMember":
         return convertNamedTupleMember(node, loc);
       case "TypeAliasDeclaration":
         return convertTypeAliasDeclaration(node, loc);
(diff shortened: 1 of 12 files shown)


javascript/extractor/src/com/semmle/ts/extractor/TypeScriptASTConverter.java
```diff
@@ -1710,7 +1710,10 @@ private Node convertNamespaceDeclaration(JsonObject node, SourceLocation loc) th
     }
     if (nameNode instanceof Literal) {
       // Declaration of form: declare module "X" {...}
-      return new ExternalModuleDeclaration(loc, (Literal) nameNode, body);
+      ExternalModuleDeclaration decl = new ExternalModuleDeclaration(loc, (Literal) nameNode, body);
+      attachSymbolInformation(decl, node);
+      System.out.println("ExternalModuleDeclaration symbol = " + decl.getSymbol());
```
erik-krogh, line 1715: Debug print? Remove?

javascript/ql/lib/semmle/javascript/NPM.qll
```diff
@@ -180,6 +180,35 @@ class PackageJson extends JsonObject {
   Module getMainModule() {
     result = min(Module m, int prio | m.getFile() = resolveMainModule(this, prio) | m order by prio)
   }
+
+  /**
+   * Gets the `types` or `typings` field of this package.
+   */
+  string getTypings() { result = this.getPropStringValue(["types", "typings"]) }
```
erik-krogh, line 187: There is also the `"types`"  field under `"exports"`: https://devblogs.microsoft.com/typescript/announcing-typescript-4-7-rc/#package-json-exports-imports-and-self-referencing  

(Might not be that relevant, so you can probably skip it).

erik-krogh (commented):
very quick first look. 
Seems reasonable, but two comments.

javascript/ql/lib/semmle/javascript/NPM.qll
asgerf: The new `exports`/`imports` fields are too complex to handle in this PR

erik-krogh (approved):

PR #9469, opened 2022-06-09
Swift: Dataflow through inout parameters

rdmarsh2:


swift/ql/lib/codeql/swift/dataflow/internal/DataFlowPrivate.qll
```diff
@@ -60,7 +61,25 @@ private module Cached {
   cached
   newtype TNode =
     TExprNode(ExprCfgNode e) or
-    TSsaDefinitionNode(Ssa::Definition def)
+    TSsaDefinitionNode(Ssa::Definition def) or
+    TInoutReturnNode(ParamDecl param, ControlFlowNode exit) {
```
MathiasVP, line 65: So if a function has `n` inout parameters and `m` return statements then we'll have `n * m` `TInoutReturnNode`s? That sounds a bit scary. Do we really need the `exit` column here? I would imagine we could just flow from the last ref of the `inout` parameter (using the `lastRef` maybe?) to a `InoutReturnNode` node that's indexed by the `param` column only.

swift/ql/lib/codeql/swift/dataflow/internal/SsaImplSpecific.qll
```diff
@@ -42,4 +50,12 @@ predicate variableRead(BasicBlock bb, int i, SourceVariable v, boolean certain)
     v = ref.getDecl() and
     certain = true
   )
+  or
+  exists(ReturnStmt return, AbstractFunctionDecl func |
```
MathiasVP, line 54: So at every `ReturnStmt` we implicitly read all of the inout parameters, right? What happens if there's no `ReturnStmt`, but there are still inout parameters that are modified by the function? Does the compiler insert dummy returns for us?

swift/ql/lib/codeql/swift/dataflow/internal/DataFlowPrivate.qll
```diff
@@ -172,6 +211,27 @@ private module ReturnNodes {
...
+
+    override DataFlowCallable getEnclosingCallable() {
+      result = TDataFlowFunc(param.getDeclaringFunction())
+    }
+
+    ParamDecl getParameter() { result = param }
+
+    override Location getLocationImpl() { result = exit.getLocation() }
+
+    override string toStringImpl() { result = param.toString() + "[return]" }
```
MathiasVP, line 233: I think it looks better with a space. Feel free to disagree!

```suggestion
    override string toStringImpl() { result = param.toString() + " [return]" }
```

swift/ql/lib/codeql/swift/dataflow/internal/DataFlowPrivate.qll
rdmarsh2: > Do we really need the exit column here?

I don't think so, but I believe we'd need extra CFG nodes to make that work.

swift/ql/lib/codeql/swift/dataflow/internal/SsaImplSpecific.qll
MathiasVP: We should already have a single exit point for a given function, I think. The `ExitNode` class should represent the `ControlFlowNode` that is the last node of a given callable.

rdmarsh2: Ok, then we can just reuse that.

swift/ql/lib/codeql/swift/dataflow/internal/DataFlowPrivate.qll
```diff
@@ -60,7 +61,21 @@ private module Cached {
   cached
   newtype TNode =
     TExprNode(ExprCfgNode e) or
-    TSsaDefinitionNode(Ssa::Definition def)
+    TSsaDefinitionNode(Ssa::Definition def) or
+    TInoutReturnNode(ParamDecl param) { param.isInout() } or
+    TInOutUpdateNode(ParamDecl param, CallExpr call) {
+      param.isInout() and
+      call.getStaticTarget() = param.getDeclaringFunction()
+    }
```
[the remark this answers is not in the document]
rdmarsh2: No - the OutNodes are per-call on the caller side, the return nodes are per-function on the callee side.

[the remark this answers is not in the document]
MathiasVP: Oh, right.

PR #9477, opened 2022-06-11
RB: Adding experimental query for detecting path traversal in Archive libraries

thiggy1342:
This query aims to find instances where user-controlled data can flow into filepath arguments of two common archive libraries used in Ruby. These two uses are TarReader and RubyZip. I also included string comparison sanitizers in order to reduce false positives.

One question I did have is if both of these instances would be covered by the existing Path Traversal query. I wanted to include both just in case, but I feel like the use of `File.open` under the hood may make the TarReader query redundant. Would there be any use to including it just for additional messaging, or do you think the additional alert the end user sees would be confusing if there were two alerts for the same line of code?


aibaars (commented):
This query seems to be a special case of https://github.com/github/codeql/blob/main/ruby/ql/src/queries/security/cwe-022/PathInjection.ql

I think it would be better to simply add additional "sinks" to https://github.com/github/codeql/blob/main/ruby/ql/lib/codeql/ruby/security/PathInjectionCustomizations.qll .

thiggy1342:
@aibaars Thanks for reviewing this, I had a suspicious it might be redundant. Just to make sure I'm following along, would the preferred method be to add an `or` clause to `FileSystemAccess::Range` similar to this: https://github.com/github/codeql/blob/main/ruby/ql/lib/codeql/ruby/frameworks/Files.qll#L37-L44 that targets the `TarReader` instantiation, or should I just keep it simple by adding the sink to the existing `PathInjectionCustomizations` class?

aibaars:
> @aibaars Thanks for reviewing this, I had a suspicious it might be redundant. Just to make sure I'm following along, would the preferred method be to add an `or` clause to `FileSystemAccess::Range` similar to this: https://github.com/github/codeql/blob/main/ruby/ql/lib/codeql/ruby/frameworks/Files.qll#L37-L44 that targets the `TarReader` instantiation, or should I just keep it simple by adding the sink to the existing `PathInjectionCustomizations` class?

Wouldn't the TarReader instantiation already be covered? The existing query should already flag up the `File.open` call in for example `TarReader.new(File.open(fname)`. It might be worth adding the `Zip.open` call though. It's probably best to put that in a new file so we have a good place to add additional "models" for the rubyzip gem.

ruby/ql/lib/codeql/ruby/frameworks/Archive.qll
```diff
@@ -0,0 +1,22 @@
...
+
+private import ruby
+private import codeql.ruby.Concepts
+private import codeql.ruby.DataFlow
+private import codeql.ruby.ApiGraphs
+
+/**
+ * Classes and predicates for modelling the RubyZip library
+ */
+module RubyZip {
```
thiggy1342, line 13: @aibaars I think I follow, but is this what you meant by creating a new file for Archive libraries where we can store models for them together?

ruby/ql/test/query-tests/security/cwe-022/ArchiveApiPathTraversal.rb
```diff
@@ -0,0 +1,69 @@
+class TestContoller < ActionController::Base
```
thiggy1342, line 1: @aibaars I moved the tests in their entirety to the test directory for the existing Path Traversal query to prove that 1. it would pick up the use of `File.open()` within `TarReader` as expected (it does) and 2. that it would find uses of `Zip::File.open()` now that it's been added as a model extending `FileSystemAccess`. I'm happy to get rid of this file and fold them into the existing test file if that works better.

ruby/ql/test/library-tests/frameworks/archive/Archive.ql
```diff
@@ -0,0 +1,4 @@
+private import ruby
+private import codeql.ruby.frameworks.Archive
+
+query predicate rubyZipFileOpens(RubyZip::RubyZipFileOpen f) { any() }
```
thiggy1342, line 4: I think I got the lib test right?

ruby/ql/lib/codeql/ruby/frameworks/Archive.qll
aibaars: Yes, this looks good. Let's also add `API::getTopLevelMember("Zip").getMember("File").getAnInstantiation()` to cover `Zip::File.new`.

aibaars (commented):
This looks good to me! Let's also add `Zip::File.new` as a file system access.

@hmac Do you agree naming the new file `frameworks/Archive.qll` or should be rename it?

aibaars (approved):
Thanks for improving our path injection query! This looks good to me.

PR #9511, opened 2022-06-13
Swift: Fix completion for "naked" patterns

MathiasVP:
(By "naked" pattern I mean a pattern that is not a child of a `CaseLabelItem`).

Previously we were seeing both a `no-match` and a `match` completion in the CFG when flowing from `42` to `x` for things like:
```swift
func foo() {
  var x = 42
}
```
because the CFG library assumed that a pattern was always a child of a `CaseLabelItem`. This PR fixes this issue.


redsun82 (approved):
LGTM!

PR #9516, opened 2022-06-13
Swift: Reintroduce control-flow for key paths

MathiasVP:
This PR reverts https://github.com/github/codeql/pull/9419 (which is a revert of https://github.com/github/codeql/pull/9373).

The original PR surfaced some issues in the CFG for certain pattern matching operations, and due to missing locations (which was fixed in https://github.com/github/codeql/pull/9430) these CFG issues led to some strange and flaky test output. The actual CFG fixes landed in https://github.com/github/codeql/pull/9511 and this PR now only shows the intended effects in the graph output diff.


rdmarsh2 (approved):
LGTM. Should we do an extra test run or two to make sure the wobble is gone?

MathiasVP:
No wobbles from rerunning the tests. So I'll hit the merge button 🤞.

PR #9521, opened 2022-06-13
Swift: generated extractor tests

redsun82:
This introduces generation of test queries.

The strategy is to create test queries for each final class in a separate directory. Test generation is however turned off with a warning for a specific class if no swift source is provided, so we can add those sources progressively. As an example for this PR, this was done for `File` and the first declaration classes.

This also means that optimizations of CI time can be left for the future.


docs/codeql/ql-training/query-examples/java/query-injection-java-1.ql
```diff
@@ -4,4 +4,4 @@ from Method m, MethodAccess ma
 where
   m.getName().matches("sparql%Query") and
   ma.getMethod() = m
-select ma, m
\ No newline at end of file
+select ma, m
```
MathiasVP, line 7: This was probably unintentional? (I assume this is also why the documentation label was added)

swift/ql/lib/codeql/swift/generated/GetImmediateParent.qll
```diff
@@ -24,6 +24,18 @@ Element getAnImmediateChild(Element e) {
       or
       abstract_storage_decl_accessor_decls(e, _, x)
       or
+      enum_case_decl_elements(e, _, x)
```
MathiasVP, line 27: I assume some new kind of sorted was done here?

swift/ql/test/extractor-tests/generated/File/File.ql
```diff
@@ -0,0 +1,10 @@
+// generated by codegen/codegen.py
+import codeql.swift.elements
+import TestUtils
+
+from File x, string isUnknown, string getName
+where
+  toBeTested(x) and
+  (if x.isUnknown() then isUnknown = "yes" else isUnknown = "no") and
+  getName = x.getName()
+select x, "isUnknown:", isUnknown, "getName:", getName
```
MathiasVP, line 10: Should we maybe leave out the `"isUnknown:"` column if it's always emitted? It would reduce the noise a tiny bit in the expected output. It's probably good to have a name for the other columns (like `getName`), though.

docs/codeql/ql-training/query-examples/java/query-injection-java-1.ql
redsun82: I'm perplexed :smile:

swift/ql/lib/codeql/swift/generated/GetImmediateParent.qll
redsun82: ah, yes, classes are sorted by (dir, name) which bunches together same kinds (decls, expr, etc), previously they were sorted only by name

swift/ql/test/extractor-tests/generated/File/File.ql
redsun82: I don't know, we could accidentally remove extraction of an entity (maybe some stateless token without fields) and we would not see that. But now that I think about that, we can maybe add `not x.isUnknown()` to the `where` clauses and we would detect that as well

MathiasVP: Yeah, I think adding `not x.isUnknown()` would be a good idea.

redsun82: I've just committed that

PR #9526, opened 2022-06-13
QL: Add parser support for parameterised modules

tausbn:
See https://github.com/tausbn/tree-sitter-ql/compare/725395405e65814f10095a451404b0ced5dc6289...tausbn:various-fixes for the actual changes to the grammar (and tests thereof).

---

A brief summary of the various nodes is perhaps in order.

- `ModuleApplication` represents the "instantiation" of a parameterised module (e.g. `Foo<...>`. Such applications may have children, each of which will be a...
- ... `SignatureExpr`, which corresponds to a predicate `bar::foo/1` or a class `Cls`. Both the `::` prefix and the arity are optional (and if the latter is absent, it's a class).
- `ModuleParam`, which represents a parameter of a parameterised module declaration. Each of these consists of two parts, a `SignatureExpr` and an identifier (representing the "type" and "variable" respectively).

### DIFF
--- a/ql/Cargo.lock
+++ b/ql/Cargo.lock
@@ -514,7 +514,7 @@ dependencies = [
 [[package]]
 name = "tree-sitter-ql"
 version = "0.19.0"
-source = "git+https://github.com/tausbn/tree-sitter-ql.git?rev=a8e519e676ff6e7aa9c1d0a8a329edefcaac69ae#a8e519e676ff6e7aa9c1d0a8a329edefcaac69ae"
+source = "git+https://github.com/tausbn/tree-sitter-ql.git?rev=91cdab850c2b6d97d1211da0423adb53c67fef3e#91cdab850c2b6d97d1211da0423adb53c67fef3e"
 dependencies = [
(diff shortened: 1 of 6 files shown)


erik-krogh:
> ModuleApplication represents the "instantiation" of a parameterised module (e.g. Foo<...>. Such applications may have children, each of which will be a... 

How about naming it `ModuleInstantiation`?

erik-krogh:
How about something like this defintion for `signatureExpr`?    

```JavaScript
signatureExpr: $ => choice(
  field("typeexpr", $.typeExpr), 
  field("predicate", $.predicateExpr)
),
```

(I'm not sure if my use of fields make sense).  
I'm basically trying to avoid some duplication, which should also make the later implementation easier.   
It might be a bad suggestion, I'm not sure.  

----  

Are you still planning on splitting up the grammar in the style of Ruby, or is that on hold?

erik-krogh (approved):
:+1:  

I'm looking forward to the "total rewrite" :rocket:

PR #7010, opened 2021-10-31
JS: make array taint-step better

japroc:
Hello! 👋

I have met a case when taint tacking was broken on expression like: `filter((x) => !!x)`. Actually it's pretty straightforward taint-step as  `filter((x) => x)`. So i made a PR.

There are also 2 moments to discuss:
1. I didnt found test case for current filter taint-step
2. Why you dont consider cases like `filter((x) => x.name)` as correct taint-step? Because of potential false positives? A have some code:
```
call.(DataFlow::MethodCallNode).getMethodName() = "filter" and
    pred = call.getReceiver() and
    succ = call and
    exists(DataFlow::FunctionNode callback, Expr retexpr, DataFlow::Node retnode |
      callback = call.getArgument(0).getAFunctionValue() and
      retexpr =
        [
          callback.getAReturn().asExpr(),
          callback.getAReturn().asExpr().(LogNotExpr).getOperand().(LogNotExpr).getOperand()
        ] and
      DataFlow::exprNode(retexpr) = AccessPath::getAReferenceTo(retnode, _) and
      retnode = callback.getParameter(0).getALocalUse() and
      (retexpr instanceof DotExpr or retexpr instanceof VarRef)
    )
```
It would be interesting to know what you think about this.

I saw in blame that you removed wide taint-step for any filter. https://github.com/github/codeql/commit/4ac21e9f3ffaacd50392fee6bd665c5d2f15989f. And of course, filter like `(x>5)` or `x.Includes(allowlist)` are incorrect taint steps. But i think that cases like `x => x.name`, `x => !!x`, `x => !!x.name` are points for discussion.

Waiting for your thoughts 😊

PR #9535, opened 2022-06-13
Tests for rebased 7010

esbena:
See #7010


esbena:
The internal experiment looks fine, although there are three surprising changes to the cache sizes. I do not think they should hold this PR back.

@erik-krogh, since we discussed the cache changes offline, can you approve?
Then we will merge this PR, and close the conflictfull #7010 with a delayed thank you.

erik-krogh (approved):
👍

PR #9542, opened 2022-06-14
Swift: extract extension declarations

AlexDenisov:


swift/extractor/visitors/DeclVisitor.h
```diff
@@ -193,6 +193,13 @@ class DeclVisitor : public AstVisitorBase<DeclVisitor> {
     emitAbstractStorageDecl(decl, label);
   }
 
+  void visitExtensionDecl(swift::ExtensionDecl* decl) {
+    auto label = dispatcher_.assignNewLabel(decl);
+    dispatcher_.emit(ExtensionDeclsTrap{label});
```
MathiasVP, line 198: Should we not also extract what type we're extending? I think we can get this from https://github.com/apple/swift/blob/main/include/swift/AST/Decl.h#L1404.

MathiasVP: Thanks!

MathiasVP (approved):
LGTM!

PR #9545, opened 2022-06-14
Java: Fix RefType.getAStrictAncestor() in the presence of type hierarchy cycles

igfoo:


smowton:
Do you know why Map (for example) has a cyclic type hierarchy?

aschackmull:
> Do you know why Map (for example) has a cyclic type hierarchy?

That's due to our interpretation of https://docs.oracle.com/javase/specs/jls/se17/html/jls-4.html#jls-4.5.1. It's possible that we ought to revise that interpretation slightly to avoid this. Clearly `? extends Object <= ?`, but whether we should have `? <= ? extends Object` as well is less clear.

aschackmull (approved):
LGTM. This is a minor semantic change, though, so it could potentially affect query results.

smowton (approved):

PR #9547, opened 2022-06-14
Kotlin: Fix wildcard introduction vs. constructor parameters

smowton:
Previously we handled the case of *methods* with potentially-wildcarded types that Java nontheless constrains to be invariant, but missed out the constructor case.


igfoo:
LGTM. Is it possible to add a test though?

igfoo (approved):
LGTM

PR #9549, opened 2022-06-14
Swift: remove toString from UnknownElement

AlexDenisov:
Another attempt to remove flakiness

### DIFF
--- a/swift/ql/lib/codeql/swift/elements/Element.qll
+++ b/swift/ql/lib/codeql/swift/elements/Element.qll
@@ -17,6 +17,4 @@ class Element extends ElementBase {
 
 class UnknownElement extends Element {
   UnknownElement() { isUnknown() }
-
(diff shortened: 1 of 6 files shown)


MathiasVP (approved):
LGTM if this unblocks https://github.com/github/codeql/pull/9411. Here's an explanation of why I believe this fix works:

- When an entity is marked as `unknown` by the extractor (i.e., the `isUnknown()` predicate holds for the entity) the entity will also be an element of the `UnknownElement` class.
- The `UnknownElement` class defines its own `toString` (which is the one you removed in this PR).
- However, the entity will also be a member of some different class (i.e., the class it actually belongs to), and that class also has a `toString`.

So in the end we get two results of `toString` on entities that are `UnknownElements`: the one they're supposed to have (that I implemented in https://github.com/github/codeql/pull/9328), and the `TBD (ClassName)` result. With this PR, they just get the `toString` that they're supposed to have.

PR #9550, opened 2022-06-14
C++: Add test for Dereferenced.qll.

geoffw0:
Add test for `Dereferenced.qll`.  I was investigating an issue this morning that turned out to be nothing (I think I was reading a result incorrectly).  But it seems good to have a permanent test for this.


MathiasVP (commented):
LGTM other than this one comment!

geoffw0:
Accepted your suggestion.

MathiasVP (approved):

PR #9551, opened 2022-06-14
python: Straight port of tarslip

yoff:
A straight port, which does not manually handle nested `not` (since I think that is not the job of a query).
I want to change the heuristic that any function with a name that contains `path` is a sanitizer..

Flow states were not necessary.

I wonder if I should make it similar to the path injection query instead of how it is now. It depends on which sanitizers we would like to support.


python/ql/lib/semmle/python/security/dataflow/TarSlipCustomizations.qll
```diff
@@ -0,0 +1,141 @@
...
+      // If argument refers to a string object, then it's a hardcoded path and
+      // this tarfile is safe.
+      not this.(DataFlow::CallCfgNode).getArg(0).getALocalSource().asExpr() instanceof StrConst and
+      /* Ignore opens within the tarfile module itself */
+      not this.getLocation().getFile().getBaseName() = "tarfile.py"
+    }
+  }
+
+  /**
+   * A sanitizer based on file name. This beacuse we extract the standard library.
```
yoff, line 54:
```suggestion
   * A sanitizer based on file name. This because we extract the standard library.
```

python/ql/lib/semmle/python/security/dataflow/TarSlipCustomizations.qll
```diff
@@ -0,0 +1,141 @@
...
+        this in [call.getArg(0), call.getArgByName("members")]
+      )
+    }
+  }
+
+  /**
+   * A sanitizer guard heuristic.
+   *
+   * For a "check-like function-name" (matching `"%path"`), `checkPath`,
+   * and a call `checkPath(info.name)`, the variable `info` is considered checked.
```
RasmusWL, line 116:
```suggestion
   * The test `if <check_path>(info.name)` should clear taint for `info`,
   * where `<check_path>` is any function matching `"%path"`.
   * `info` is assumed to be a `TarInfo` instance.
```

python/ql/lib/semmle/python/security/dataflow/TarSlipCustomizations.qll
```diff
@@ -0,0 +1,141 @@
...
+        // Assume that any test with "path" in it is a sanitizer
+        call.getAChild*().(AttrNode).getName().matches("%path")
+        or
+        call.getAChild*().(NameNode).getId().matches("%path")
+      )
+    }
+
+    override predicate checks(ControlFlowNode checked, boolean branch) {
+      checked = tarInfo and
+      branch in [true, false]
```
RasmusWL, line 138: the old code only clears taint on the false edge. I don't see any mention of why this was changed, but the fact that our tests results did not change because of this, hints that our tests could have been better :smile:

python/ql/lib/semmle/python/security/dataflow/TarSlipCustomizations.qll
```diff
@@ -0,0 +1,141 @@
...
+
+  /**
+   * A sanitizer for "tar slip" vulnerabilities.
+   */
+  abstract class Sanitizer extends DataFlow::Node { }
+
+  /**
+   * A sanitizer guard for "tar slip" vulnerabilities.
+   */
+  abstract class SanitizerGuard extends DataFlow::BarrierGuard { }
```
RasmusWL, line 37: I guess with https://github.com/github/codeql/pull/9590 merged, this will now give a deprecation warning :disappointed:

RasmusWL (changes requested):
A few minor things, otherwise looks fine :+1:

python/ql/lib/semmle/python/security/dataflow/TarSlipCustomizations.qll
yoff: True, I never got to write this down. I felt that the heuristic "contains `path`" did not warrant an assumption about whether the function tested for good or bad paths...

python/ql/lib/semmle/python/security/dataflow/TarSlipCustomizations.qll
```diff
@@ -0,0 +1,141 @@
...
+
+    TarFileInfoSanitizer() {
+      exists(CallNode call, AttrNode attr |
+        this = call and
+        // We must test the name of the tar info object.
+        attr = call.getAnArg() and
+        attr.getName() = "name" and
+        attr.getObject() = tarInfo
+      |
+        // Assume that any test with "path" in it is a sanitizer
```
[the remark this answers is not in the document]
yoff: I think it would match something like `utilities.pathlib.filters.relative()`.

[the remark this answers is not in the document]
RasmusWL: nope, try for yourself with

```codeql
from string s
where
  s in ["ends_with_path", "path_foo"] and
  s.matches("%path")
select s
```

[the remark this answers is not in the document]
yoff: Ah, right. My example should have been `utilities.path.filters.relative()`.

python/ql/lib/semmle/python/security/dataflow/TarSlipCustomizations.qll
```diff
@@ -0,0 +1,149 @@
...
+   */
+  abstract class Sanitizer extends DataFlow::Node { }
+
+  /**
+   * DEPRECATED: Use `Sanitizer` instead.
+   *
+   * A sanitizer guard for "tar slip" vulnerabilities.
+   */
+  abstract deprecated class SanitizerGuard extends DataFlow::BarrierGuard { }
+
```
RasmusWL, line 40: Since this is new code, we might as well just not introduce the deprecated class at all.

```suggestion
```

python/ql/lib/semmle/python/security/dataflow/TarSlipQuery.qll
```diff
@@ -0,0 +1,29 @@
...
+
+  override predicate isSource(DataFlow::Node source) { source instanceof Source }
+
+  override predicate isSink(DataFlow::Node sink) { sink instanceof Sink }
+
+  override predicate isSanitizer(DataFlow::Node node) { node instanceof Sanitizer }
+
+  deprecated override predicate isSanitizerGuard(DataFlow::BarrierGuard guard) {
+    guard instanceof SanitizerGuard
+  }
```
RasmusWL, line 28:
```suggestion
```

python/ql/lib/semmle/python/security/dataflow/TarSlipCustomizations.qll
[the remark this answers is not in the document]
RasmusWL: Right, I'm still not 100% happy about the comment. I have two suggestions:

1. change the comment entirely, so it reads something like `// the assumption that any test that matches `%path` is a sanitizer might be too broad`
2. make the comment more precise

After having written out (1), I'm very much in favor of that version, since I guess that is what highlights the intent behind the initial comment: that the sanitizer might be too broad.

RasmusWL (changes requested):
Besides a few minor things, looks good :+1:

python/ql/lib/semmle/python/security/dataflow/TarSlipCustomizations.qll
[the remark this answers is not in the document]
yoff: Right, that does seem more helpful.

RasmusWL (approved):
LGTM (as long as newest performance test ends up looking good)

yoff:
Newest performance test is basically "nothing to report".

PR #9552, opened 2022-06-14
C++: Introduce relation for tracking braced initializers

jketema:


edoardopirovano (approved):
As release manager I'm happy with this being in `rc/3.6` and it is a clean cherry-pick of https://github.com/github/codeql/pull/9552. @MathiasVP since you reviewed the original PR, are you also happy with this change going into 2.9.4 and therefore GHES 3.6?

edoardopirovano:
> @edoardopirovano Does anything special need to be done with the included change note file?

Nope, this should be okay. The release automation will do the right thing by including it in the changelog then deleting it, and the deletion should propagate from `main` when we do the mergeback so that it doesn't get re-included in the next release.

MathiasVP:
> As release manager I'm happy with this being in `rc/3.6` and it is a clean cherry-pick of #9552. @MathiasVP since you reviewed the original PR, are you also happy with this change going into 2.9.4 and therefore GHES 3.6?

No objects from me 🚢!

jketema:
@edoardopirovano There's a failing integration test, that seems a Kotlin failure. Is that expected?

edoardopirovano:
> @edoardopirovano There's a failing integration test, that seems a Kotlin failure. Is that expected?

Yes. Unfortunately our CI isn't too clever and tries to run tests by default always against the `main` branch of `semmle-code` which often causes failures when the PR is targetting an RC branch due to other languages changing their extractors. When I sent the `/qlucie ql --test-pr 9552 --code-ref rc/3.6` message on Slack, it triggered a new run that actually used the `rc/3.6` branch in `semmle-code`. This run [passed](https://github.com/github/semmle-code/runs/6881837622), hence why the PR is green, but the GitHub UI is still showing the original failing run as well. You're all good to ignore this and merge.

PR #9554, opened 2022-06-14
Kotlin: Add more Kotlin <-> Java special method name mappings

smowton:
* `removeAt` is an overload of `remove` in the JVM world
* `Enum`'s `ordinal` and `name` methods are properties from Kotlin's point of view
* `Number`'s `toInt` etc are named `intValue` etc on `java.lang.Number`


tamasvajk (approved):
LGTM

PR #9555, opened 2022-06-14
Swift: Add MaD skeleton

MathiasVP:
This PR adds a skeleton for MaD for Swift. It's currently completely dead code, and a bunch of predicates are implemented as `none()` with a `// TODO` added next to it. Those predicates will need to be filled out before we can start to use the library. I prefer doing that in a different commit, though.

@rdmarsh2 I'd appreciate it if you could take a look at the changes in `DataFlowDispatch.qll` and `DataFlowPrivate.qll`. Other than that, I don't think a thorough review is required on this PR.


swift/ql/lib/codeql/swift/dataflow/internal/DataFlowDispatch.qll
```diff
@@ -57,17 +61,78 @@ class DataFlowCallable extends TDataFlowCallable {
 
   /** Gets the location of this callable. */
   Location getLocation() { result = func.getLocation() }
+
+  AbstractFunctionDecl getUnderlyingCallable() { result = func }
```
rdmarsh2, line 65: Should this return something compatible with closures?

MathiasVP: Yeah, it probably should. How about we add `ClosureExpr` to https://github.com/github/codeql/blob/main/swift/ql/lib/codeql/swift/controlflow/internal/Scope.qll#L5 and use `Callable::TypeRange` instead?

MathiasVP: https://github.com/github/codeql/pull/9557 makes the required changes for this to work. Once that PR is merged I'll fix this comment.

MathiasVP: Fixed in eff046e2f75f6ab8ef48029ddbae97f95bad9504.

rdmarsh2 (approved):

PR #9566, opened 2022-06-15
Ruby: recognize ActiveRecord `find_by_x` methods

alexrford:
This lets us recognize the result of a `find_by_<field_name>` dynamic finder call as a potential active record model object.

See https://guides.rubyonrails.org/active_record_querying.html#dynamic-finders


alexrford:
Sorry for the noise in the test output diffs - the test cases were in a file originally copied from the SQLi tests that I wanted to tidy up a bit. The changes should make more sense commit-by-commit.

PR #9567, opened 2022-06-15
Python: allow class decorators in `.getASubclass()`

RasmusWL:
Instead of just adding the test example, I thought I would try to fix it. I'm totally up for discussing whether this is the right approach, but personally I think it aligns with our goals.

We don't currently allow decorators for functions, so there will be a little bit of inconsistency...


python/ql/lib/semmle/python/ApiGraphs.qll
```diff
@@ -591,8 +591,18 @@ module API {
...
+          // Potentially a class decorator could do anything, but we assume they are
+          // "benign" and let subclasses edges flow through anyway.
+          // see example in https://github.com/django/django/blob/c2250cfb80e27cdf8d098428824da2800a18cadf/tests/auth_tests/test_views.py#L40-L46
+          (
+            not exists(clsExpr.getADecorator()) and
+            ref.asExpr() = clsExpr
+            or
+            ref.asExpr() = clsExpr.getADecoratorCall() and
+            not exists(PY::Call otherDecorator | otherDecorator.getArg(0) = ref.asExpr())
+          )
```
yoff, line 605: I am not saying we should do this, but it seems (from your test) that the code could be simplified as follows
```suggestion
          (
            ref.asExpr() = clsExpr
            or
            ref.asExpr() = clsExpr.getADecoratorCall()
          )
```
but possibly the code in the PR is more performant in that it creates fewer tuples?

yoff (commented):
I agree with the heuristic in this PR; I would expect decorators to preserve the sub-class relationship often enough to make this over-approximation useful. Are the evaluation results as you expect?

The suggestion is mostly me being curious, the code might well be better as is.

python/ql/lib/semmle/python/ApiGraphs.qll
tausbn: Yeah, I was going to point this out as well. In some sense, it might be more correct to treat the result of each decorator as a subclass (if, say, you want to use API graphs to figure out which subclasses are decorated with a particular decorator).

In terms of performance I wouldn't expect it to make much of a difference. Yes, it's more API graph nodes, but all of the "inner" decorators don't flow anywhere except into the decorator immediately preceding them.

RasmusWL: I did this to optimize for performance yes. I can see the argument that it's nice to be able to also be able to track this into the arguments of the decorators, so I'll accept the suggestion :+1:

RasmusWL:
Accepting this change also fixed the tests :+1: :open_mouth: 

The problem was this bit of modeling code (in Flask.qll)

```codeql
  class FlaskMethodViewClass extends FlaskViewClass {
    FlaskMethodViewClass() {
      api_node = Views::MethodView::subclassRef() and
      this.getParent() = api_node.getAnImmediateUse().asExpr()
    }
```

would no longer recognize the `Nested` class itself as being a `MethodView` subclass :disappointed: (since the `ClassExpr` was not treated as a subclass)

```py
    @flask_admin.expose_plugview("/flask-class") # $ routeSetup="/flask-class"
    @flask_admin.expose_plugview(url="/flask-class/<arg>") # $ routeSetup="/flask-class/<arg>"
    class Nested(MethodView):
        def get(self, cls, arg="default"): # $ requestHandler routedParameter=arg
            assert isinstance(cls, ExampleClass)
            ensure_tainted(arg) # $ tainted
            ensure_not_tainted(cls)
            return "GET: " + arg # $ HttpResponse
```

tausbn:
I wonder if it would be even more correct to say that (for the purposes of the API graph), the results of applying the decorators are not just subclasses of the same superclass, but in fact the very same class. In some sense, decorating a class should not really change the identity of that class (though it may e.g. wrap it in some other object).

This might also be needed in order to get tracking of instances through decorators to work correctly. Consider

```python
@foo
class Bar:
    ...
    def baz(self):
        ...

Bar().baz()
```
we want the call above to resolve `Bar.foo`, but if we just follow the flow, we'll find that `Bar` comes from a desugared assignment `Bar = foo(Bar)`, which doesn't obviously have a `baz` method (at least, not syntactically). If we let classes "flow through" decorators, then there will be flow from the inner `Bar` to the call-site.

yoff (approved):
LGTM

PR #9568, opened 2022-06-15
QL: Allow module applications to the right of `::`

tausbn:
At this point, `getName` is not really a good name for this member predicate, but I'll leave that bit of cleanup for later.


erik-krogh (approved):
image
:tada:  

(Some new alerts were introduced, but we don't look at that).

PR #9574, opened 2022-06-16
Ruby: More Rails modeling

hmac:
- Recognise `ActionCable::Connection::TaggedLoggerProxy` as a logger
- Recognise `File.atomic_write` (a core extension from ActiveSupport) as a file write access
- Recognise `redirect_back` and `redirect_back_or_to` as URL redirects
- Recognise various methods in `Rails::Generators::Actions` as command executions


nickrolfe (approved):
Looks great, thanks!

PR #9575, opened 2022-06-16
QL: support for parameterized modules

erik-krogh:
Supports all the features currently used in `main` (empty consistency predicates and no parse errors :tada:). 

~~But doesn't support the features that will (hopefully) be used soon.   
I'm going to work on that next.~~  
All current features should be supported!   

Commit-by-commit review partially recommended.   
The commit history is the raw history of how I implemented this (e.g. the initial parser changes were wrong).  

image

### DIFF
--- a/ql/ql/src/codeql_ql/ast/Ast.qll
+++ b/ql/ql/src/codeql_ql/ast/Ast.qll
@@ -726,11 +726,9 @@ class Module extends TModule, ModuleDeclaration {
   /** Holds if the `i`th parameter of this module has `name` and type `sig`. */
   predicate hasParameter(int i, string name, SignatureExpr sig) {
     exists(QL::ModuleParam param |
-      param = mod.getParameter(i) and name = param.getParameter().getValue()
-    |
-      toQL(sig) = param.getSignature().getPredicate()
-      or
-      toQL(sig) = param.getSignature().getTypeExpr()
+      param = mod.getParameter(i) and
+      name = param.getParameter().getValue() and
+      sig.toQL() = param.getSignature()
     )
   }
 }
@@ -2240,15 +2238,24 @@ class ModuleExpr extends TModuleExpr, ModuleRef {
    */
   SignatureExpr getArgument(int i) {
     exists(QL::ModuleInstantiation instantiation | instantiation.getParent() = me |
-      toQL(result) = instantiation.getChild(i).getPredicate()
-      or
-      toQL(result) = instantiation.getChild(i).getTypeExpr()
+      result.toQL() = instantiation.getChild(i)
     )
   }
 }
 
 /** A signature expression, either a `PredicateExpr` or a `TypeExpr`. */
-class SignatureExpr extends TSignatureExpr, AstNode { }
+class SignatureExpr extends TSignatureExpr, AstNode {
+  QL::SignatureExpr sig;
+
+  SignatureExpr() {
+    toQL(this) = sig.getPredicate()
+    or
+    toQL(this) = sig.getTypeExpr()
+  }
+
+  /** Gets the generated AST node that contains this signature expression. */
+  QL::SignatureExpr toQL() { result = sig }
+}
 
 /** An argument to an annotation. */
 private class AnnotationArg extends TAnnotationArg, AstNode {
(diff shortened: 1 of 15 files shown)


ql/ql/src/codeql_ql/ast/internal/Module.qll
```diff
@@ -288,14 +313,6 @@ module ModConsistency {
         .regexpMatch(".*/(test|examples|ql-training|recorded-call-graph-metrics)/.*")
   }
 
-  query predicate noResolveModuleExpr(ModuleExpr me) {
```
tausbn, line 291: I'm slightly puzzled as to why this was just removed. Does it not provide any useful information now?

ql/ql/src/codeql_ql/ast/internal/Predicate.qll
```diff
@@ -162,17 +176,20 @@ module PredConsistency {
     c > 1 and
     resolvePredicateExpr(pe, p)
   }
+  // This can happen with parametarized modules
```
tausbn, line 179:
```suggestion
  // This can happen with parameterized modules
```

I wonder, could we resurrect this test (and others) if we filter out stuff related to parameterised modules?

ql/ql/test/callgraph/ParamModules.qll
```diff
@@ -0,0 +1,63 @@
+module PredicateSig {
+  signature predicate fooSig(int i);
+
+  module UsesFoo<fooSig/1 fooImpl> {
+    predicate bar(int i) { fooImpl(i + 1) }
+  }
+
+  predicate myFoo(int i) { i = 42 }
+
+  predicate use(int i) { UsesFoo<myFoo/1>::bar(i) }
```
tausbn, line 10: I may be mistaken, but I don't see anything that `UsesFoo` resolves to here. Is this intended? Would it perhaps make sense to resolve it to `module UsesFoo` above?

tausbn (approved):
Overall this looks okay to me, and I'm happy to merge it now and fix up bugs later.

I note that there's no special handling of the `signature` annotation (or at least if there is, I couldn't find it). This would perhaps make sense to add.

ql/ql/src/codeql_ql/ast/internal/Module.qll
erik-krogh: It is because a `ModuleExpr` is now just a special `TypeRef`.   
And we test that all `TypeRef`s resolve to a type in `Type.qll`.

ql/ql/src/codeql_ql/ast/internal/Predicate.qll
erik-krogh: We could.... 
We can check if one of the targets is a `signature predicate`, and then allow multiple if that's the case.

ql/ql/test/callgraph/ParamModules.qll
erik-krogh: I don't get your comment, the `UsesFoo` here is resolved to the `module UsesFoo` above.

ql/ql/src/codeql_ql/ast/Ast.qll
```diff
@@ -2190,6 +2216,10 @@ class ModuleExpr extends TModuleExpr, ModuleRef {
     result = me.getName().(QL::SimpleId).getValue()
     or
     not exists(me.getName()) and result = me.getChild().(QL::SimpleId).getValue()
+    or
+    exists(QL::ModuleInstantiation instantiation | instantiation.getParent() = me |
+      result = instantiation.getName().getChild().getValue()
+    )
```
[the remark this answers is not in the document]
erik-krogh: Yes it could. I'll throw my patching tool at QL-for-QL in the followup PR.

ql/ql/test/callgraph/ParamModules.qll
tausbn: As far as I can tell from the expected output, we resolve `bar` to the `bar` on line 5, and `myFoo` to the `myFoo` on line eight, but I would also expect `UsesFoo` to resolve to the `UsesFoo` on line 4. Where is this evident in the `.expected` output?

(Actually, I guess I have my answer: this is only a test of the call graph, and `UsesFoo` isn't being called. I'll assume, then, that it resolves as you say.)

PR #9576, opened 2022-06-16
Swift: add empty implementation of `defaultImplicitTaintRead`

erik-krogh:
[This call](https://github.com/github/codeql/blob/7c5a83833b974ee40be5d76faed3cdd1b5066708/swift/ql/lib/codeql/swift/dataflow/internal/tainttracking1/TaintTrackingImpl.qll#L166) had no targets, because there was no implementation of `defaultImplicitTaintRead` in the Swift implementation.    
So I added an empty implementation.  

I found it because it was the only consistency error reported by QL-for-QL (after #9575).


MathiasVP (approved):
Thanks!

PR #9577, opened 2022-06-16
Go autobuilder: don't attempt a go mod tidy when there's a vendor directory present

smowton:
This is likely to spuriously remove dependencies leading to a later build failure due to missing requirements.

### DIFF
--- a/go/extractor/cli/go-autobuilder/go-autobuilder.go
+++ b/go/extractor/cli/go-autobuilder/go-autobuilder.go
@@ -291,7 +291,7 @@ func main() {
 	}
 
 	// Go 1.16 and later won't automatically attempt to update go.mod / go.sum during package loading, so try to update them here:
-	if depMode == GoGetWithModules && semver.Compare(getEnvGoSemVer(), "1.16") >= 0 {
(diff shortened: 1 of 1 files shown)


smowton:
LGTM dist build results are good (no new failures)

go/extractor/cli/go-autobuilder/go-autobuilder.go
```diff
@@ -291,7 +291,7 @@ func main() {
 	}
 
 	// Go 1.16 and later won't automatically attempt to update go.mod / go.sum during package loading, so try to update them here:
-	if depMode == GoGetWithModules && semver.Compare(getEnvGoSemVer(), "1.16") >= 0 {
+	if modMode != ModVendor && depMode == GoGetWithModules && semver.Compare(getEnvGoSemVer(), "1.16") >= 0 {
```
igfoo, line 294: Later (https://github.com/github/codeql/blob/main/go/extractor/cli/go-autobuilder/go-autobuilder.go#L498) `modMode` might be changed. Does checking it at this point have the desired behaviour?

smowton: Yes I think this is ok; the change is applied post-build and we need to make a decision pre-build as the build may expect go.mod-updating behaviour no longer present in go >= 1.16.

igfoo (approved):

PR #9579, opened 2022-06-16
Python: Improve `BarrierGuard`

yoff:
Expressions inside a `not` expression have so far been excluded from being `BarrierGuard`s. This PR fixes that.

This PR focuses on `not` specifically. An alternative strategy is to look at all children of the last statement of a `ConditionBlock`.


aschackmull:
Please hold this PR until https://github.com/github/codeql/pull/9590 is merged.

python/ql/lib/semmle/python/dataflow/new/internal/DataFlowPublic.qll
```diff
@@ -527,16 +527,32 @@ class StarPatternElementNode extends Node, TStarPatternElementNode {
   override Location getLocation() { result = consumer.getLocation() }
 }
 
+ControlFlowNode guardNode(ConditionBlock conditionBlock, boolean flipped) {
```
RasmusWL, line 530: I wish there was a bit more QLDoc for this change.

I don't fully understand how it works now :thinking: 

We have a `ConditionBlock`, which is _A basic block which terminates in a condition, splitting the subsequent control flow_.

Previously we said that the **last** node in such a block would control which basic block to evaluate next.

From doing a bit of testing, the last node in the top-most basicblock is as expected, so for `foo` it is `is_safe(s)` and for `bar` it is `not is_safe(s)` (a UnaryExprNode).

```py
def foo():
    if is_safe(s):
        ...

def bar():
    if not is_safe(s):
        ...
```

Previously a QL writer would have to taget the `UnaryExprNode` with the `BarrierGuard` subclass, and handle the flipping of branches themselves.

What you've done in this PR, is to say that for the first back block in `bar`, BOTH `not is_safe(s)` and `is_safe(s)` are GuardNodes, but the latter has `flipped=true`. This means that suddenly the old `BarrierGuard` subclass  whose char-pred has `this = <is_safe>.getACall()` is valid even for the `is_safe(s)` inside the `not is_safe(s)`.

I'm not sure what the QLDoc should say exactly, but something to explain why this predicate must exist, and be like it is, would probably aid the next person trying to understand it :blush:

RasmusWL (changes requested):
Very cool :muscle: love how we're suddenly able to handle all the `not is_safe(s)` cases with just this small change :1st_place_medal:

python/ql/lib/semmle/python/dataflow/new/internal/DataFlowPublic.qll
RasmusWL: Hmm, it seems like much of this explanation was inside the commit message (which I did not read until now :sweat_smile:) -- I still think it would be nice to have an explanatory comment inside the actual code :blush:

yoff: I keep writing explanations in the commit messages and forgetting that it is the code that actually persists. Sorry for that, I will write some documentation in the code.

yoff:
Evaluation looks fine.

PR #9589, opened 2022-06-17
Update tree-sitter-embeded-template

aibaars:


nickrolfe (approved):
Looks great, assuming you've done the usual test of the upgrade scripts by running all the tests with an old extractor.

PR #9593, opened 2022-06-17
QL: followup fixes to parameterized modules

erik-krogh:
This is a followup PR to https://github.com/github/codeql/pull/9575 

The `only resolve module types ...` mistake was caught by a test-case in my patching utility.  

Lots of changes.  
Now every call has one, and only one, target!  (Except for some cases laid out in the consistency test).  
There were a whole bunch of small bugs that caused us to spuriously resolve too many predicates.


tausbn (approved):
Nice work! All of the changes look good to me. :+1:

PR #9605, opened 2022-06-18
Experimental manually check request verb query

thiggy1342:
The aim of this query is to find all calls to `.get?`, `.post?`, `.patch?`, `.put?`, `.delete?`, and `.head?`, as well as element references to `env["REQUEST_METHOD"]` that are used in control flow expressions. The general idea is that if controller action methods are manually checking http methods as part of control flow, they might be unintentionally bypassing security measures that are intended to run before the controller method runs. For example, this could include authorization checks, CSRF token checks, etc.

There are a few areas where I'd like to refine this query:
- If possible, I'd like to limit findings to just those found in subclasses of `ActionController`. The idea here is that while some of the underlying scaffolding of a Rails app may need to use these checks in legitimate ways, it's probably not a good idea to do this in your standard controller classes. One thought I had was maybe checking the filepath of the file in question, but that seems a bit naive even though standard Rails convention would have the controllers in the `app/controllers` directory.
- Another feature I'd like to add is to check instances where `env["REQUEST_METHOD"]` references isn't directly used in a conditional statement or passed into `include?`, but is instead assigned to a variable who's value then flows into this check. I've tried to add this as part of `CheckRequestMethodFromEnv` but it doesn't seem to be working as expected.
- While my query does seem to be returning results from an existing application database, I don't seem to be getting any hits for my test classes. I need to figure out what I'm missing here.

### DIFF
--- a/ruby/ql/src/experimental/manually-check-http-verb/ManuallyCheckHttpVerb.ql
+++ b/ruby/ql/src/experimental/manually-check-http-verb/ManuallyCheckHttpVerb.ql
@@ -1,6 +1,6 @@
 /**
  * @name Manually checking http verb instead of using built in rails routes and protections
- * @description Manually checking HTTP verbs is an indication that multiple requests are routed to the same controller action. This could lead to bypassing necessary authorization methods and other protections, like CSRF protection. Prefer using different controller actions for each HTTP method and relying Rails routing to handle mappting resources and verbs to specific methods. 
+ * @description Manually checking HTTP verbs is an indication that multiple requests are routed to the same controller action. This could lead to bypassing necessary authorization methods and other protections, like CSRF protection. Prefer using different controller actions for each HTTP method and relying Rails routing to handle mappting resources and verbs to specific methods.
  * @kind problem
  * @problem.severity error
  * @security-severity 7.0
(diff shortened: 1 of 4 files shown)


thiggy1342:
@hmac Thanks for taking a look and that is a great question! This query aims to find examples where a Rails apps and controller routing is implemented in a non-standard way and folks might be getting themselves into trouble. Specifically, more than one uri path + HTTP verb combo would need to be routed to the same controller method. Something like this in `routes.rb`:
``` ruby
get "/example/resource", to "example#resource_method"
post "/example/resource", to "example#resource_method"
```
Then, inside of `Example.resource_method`, logic would branch based on the results from `request.get?` or `request.post` or `env['REQUEST_METHOD'] == 'GET'`, etc.

I did some further digging, and as you mention, GET and HEAD requests are the only ones that don't get CSRF protection by default, and the detection happens at the [request level](https://github.com/rails/rails/blob/01f58d62c2f31f42d0184e0add2b6aa710513695/actionpack/lib/action_controller/metal/request_forgery_protection.rb#L432). I'm not sure why I thought it was done as a component of routing (i.e. if the method name is "create", CSRF is checked but not for methods named "index.", so I don't think we'll find many cases where CSRF is missing.

There is still the possibility of a single method handling two HTTP verbs (again, in an app that where the default Rails routing is misused). For example if a single method handles PATCH and DELETE requests in a controller, and an "editor" user account is allowed to update a resource but not delete it, an unknowing developer may change `before_action :ensure_admin` to `before_action :ensure_editor` check to add functionality for PATCH requests routed to the method.

I don't have a great example of this in the wild, but the CSRF check has bit us before in the past. These may be too opinionated for the experimental ruleset, and perhaps a custom ruleset for internal use is the better path for this query.

thiggy1342:
@hmac I've narrowed down the query scope to look specifically for control flow checks involving calls to `request.get?`, as it looks like `HEAD` requests are generally treated like `GET` requests, and having branching logic that checks just `.get?` could result in `HEAD` requests hitting codepaths that should be protected by CSRF. I have a few specific questions on parts of my query as well.

ruby/ql/src/experimental/manually-check-http-verb/ManuallyCheckHttpVerb.ql
```diff
@@ -0,0 +1,58 @@
...
+    (
+      // and is this node a param of a call to `.include?`
+      exists(MethodCall call | call.getAnArgument() = this | call.getMethodName() = "include?")
+      or
+      // check if env["REQUEST_METHOD"] is compared to GET
+      exists(EqualityOperation eq | eq.getAChild() = this |
+        eq.getAChild().(StringLiteral).toString() = "GET"
+      )
+    )
+  }
```
thiggy1342, line 35: The idea here is that I would also like to find instances where the value of `env["REQUEST_METHOD"] is compared to a string literal as part of a control flow check. I'm having a bit of trouble wrapping my head around how I'd constrain things to be part of a ConditionalExpr from here, or am I thinking about this the wrong way? Should I just add that part directly to my query?

ruby/ql/src/experimental/manually-check-http-verb/ManuallyCheckHttpVerb.ql
```diff
@@ -0,0 +1,53 @@
...
+class CheckGetFromEnv extends AstNode {
+  CheckGetFromEnv() {
+    // is this node an instance of `env["REQUEST_METHOD"]
+    this instanceof GetRequestMethodFromEnv and
+    // check if env["REQUEST_METHOD"] is compared to GET
+    exists(EqualityOperation eq | eq.getAChild() = this |
+      eq.getAChild().(StringLiteral).toString() = "GET"
+    )
+  }
+}
```
thiggy1342, line 36: I was also thinking of capturing comparisons between env["REQUEST_METHOD"] and "GET" used in a control flow statement, but I'm having a bit of trouble wrapping my head around how to get there from here. Am I going about this the wrong way? Should it be another class or baked into the query itself?

ruby/ql/src/experimental/manually-check-http-verb/ManuallyCheckHttpVerb.ql
thiggy1342: Also, does this seem useful or should I just stick to singling out calls to `request.get?`?

thiggy1342: Ah awesome! I've been leaning on `.toString()` far too much, so this is great!

ruby/ql/src/experimental/manually-check-http-verb/ManuallyCheckHttpVerb.ql
```diff
@@ -0,0 +1,53 @@
...
+  CheckGetRequest() { this.getMethodName() = "get?" }
+}
+
+class ControllerClass extends ModuleBase {
+  ControllerClass() { this.getModule().getSuperClass+().toString() = "ApplicationController" }
+}
+
+class CheckGetFromEnv extends AstNode {
+  CheckGetFromEnv() {
+    // is this node an instance of `env["REQUEST_METHOD"]
```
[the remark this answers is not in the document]
thiggy1342: Thanks for taking a look! I've pushed up some changes where I think I've followed this pattern, but the quick evals aren't quite working like I'd expect. While I'm not sure we'd see the full query find any results, the quick eval that I'm running on the `Request` class for example isn't returning any results. I'm wondering if there's anything I might have missed when trying to follow this pattern. I'm almost certain that I should be able to easily find any uses of `request` in just about any rails app.

thiggy1342:
@hmac After making the recommended changes, it looks like I can capture instances of `reqeust.get?` but nothing else. One thing I did notice is that when calling `request.get?` there isn't a comparison operation, just testing if it returns `true`. I'm wondering if I'm missing something for instances where the data from our source needs to flow through a comparison operations i.e. `request.method == 'GET'` and our current query include those?

thiggy1342:
@hmac That did the trick! I also went ahead and expanded the sink to catch cases where the result of on of our method checks flows into a `case` statement as well, since that could also be used for control flow. I've updated the tests and added  a qhelp file, so if you think things look good, this might be about ready to ship.

thiggy1342:
perhaps not, I need to sort out this path problem formatting.

PR #9616, opened 2022-06-20
JS: Add withoutPropStep and model raw 'await' step with it

asgerf:
Doing `await x` on non-promise value just returns `x` itself, but our type-tracking steps didn't take this into account, leading to missed flow.

This PR adds support for this by adding `PropertySet` and `withoutPropStep(Node pred, Node succ, PropertySet props)`. This steps allows flow from `pred -> succ` but without the properties in `props` (see qldoc in PR).

This new step is deliberately similar to the `WithoutElement` step that @hvitved added in https://github.com/github/codeql/pull/8938. (The `WithElement` step is called a `loadStoreStep` in JS)

[Evaluation](https://github.com/github/codeql-dca-main/tree/data/asgerf/without-prop-st__default__code-scanning/reports) was quiet


erik-krogh (approved):
:+1:

PR #9617, opened 2022-06-20
Go: Properly escape dash in tracing-config.lua

criemen:
Previously, the pattern didn't match what it was intended to match.


edoardopirovano (approved):
Good find!

AlexDenisov:
Given the amount of time I spent debugging exactly the same issue, would it make sense to add a check to the tracer to issue a warning for such cases?

cklin (approved):

PR #9619, opened 2022-06-20
Swift: Generate better declaration names

MathiasVP:
Prior to this PR, the name we generate for the declarations:
```swift
func f(x first: Int, y second: Int, _ third: Int) { }
func f(z first: Int, w second: Int, _ third: Int) { }
```
were both just `f`. This makes it inconvenient to mark the first function overload as a taint source purely based on the name. In addition, the Swift compiler refers to each overload as `f(x:y:_:)` and `f(z:w:_:)` respectively.

With this PR, we generate the expected names (i.e., `f(x:y:_:)` and `f(z:w:_:)`) for the function overloads.

Note that as a special case of this, the name of a function `f` with no parameters:
```swift
func f() {  }
```
is actually `f()` with parentheses! That's a bit confusing to me since it looks like a function call (even though it's actually a reference to the function). This matches the output of `swiftc`, though.

This special case required a slight modification to `swift/ql/test/library-tests/dataflow/dataflow/DataFlow.ql` (which is included in c049988116f02de7fba0a698fcd82ad92013459e).


AlexDenisov (approved):
This looks so much better now! ❤️

geoffw0 (approved):

PR #9620, opened 2022-06-20
Swift: add devcontainer setup

redsun82:
This adds a specific devcontainer setup to spin up a codespace ready for use for swift development.

The configuration is selectable from the `Configure and create codespace` subcommand (selectable by clicking the down arrow next to the default `Create codespace on main` button), and on codespace creation it will:

* install required dependencies to build the swift extractor
* install bazelisk and make it callable via `bazel`
* install the latest version of codeql (can be updated with `update-codeql` at a later stage)
* create a swift extractor pack with the current version of the checked-out code
* create a codeql config file setting up `--search-path` so that the extractor pack is picked up
* set up `pre-commit`

One still needs to update the codeql CLI via the vscode command. After that, swift QL tests will be runnable immediately without any further steps or command line flags.

The initial devcontainer setup was taken from the default C++ one. I'm not entirely sure whether we need the `runArgs` that were taken from there.

### DIFF
--- a/.devcontainer/swift/Dockerfile
+++ b/.devcontainer/swift/Dockerfile
@@ -1,0 +1,11 @@
+# See here for image contents: https://github.com/microsoft/vscode-dev-containers/tree/v0.236.0/containers/cpp/.devcontainer/base.Dockerfile
+
+# [Choice] Debian / Ubuntu version (use Debian 11, Ubuntu 18.04/22.04 on local arm64/Apple Silicon): debian-11, debian-10, ubuntu-22.04, ubuntu-20.04, ubuntu-18.04
+ARG VARIANT="bullseye"
+FROM mcr.microsoft.com/vscode/devcontainers/cpp:0-${VARIANT}
+
+ARG BAZELISK_VERSION=v1.10.1
+ARG BAZELISK_DOWNLOAD_SHA=dev-mode
+USER root
+ADD root.sh /tmp/root.sh
+RUN bash /tmp/root.sh && rm /tmp/root.sh
--- a/.devcontainer/swift/Dockerfile
+++ b/.devcontainer/swift/Dockerfile
@@ -5,4 +5,5 @@ FROM mcr.microsoft.com/vscode/devcontainers/cpp:0-ubuntu-22.04
 
 USER root
 ADD root.sh /tmp/root.sh
+ADD update-codeql.sh /usr/local/bin/update-codeql
 RUN bash /tmp/root.sh && rm /tmp/root.sh
--- a/.devcontainer/swift/Dockerfile
+++ b/.devcontainer/swift/Dockerfile
@@ -1,11 +1,8 @@
 # See here for image contents: https://github.com/microsoft/vscode-dev-containers/tree/v0.236.0/containers/cpp/.devcontainer/base.Dockerfile
 
 # [Choice] Debian / Ubuntu version (use Debian 11, Ubuntu 18.04/22.04 on local arm64/Apple Silicon): debian-11, debian-10, ubuntu-22.04, ubuntu-20.04, ubuntu-18.04
-ARG VARIANT="bullseye"
-FROM mcr.microsoft.com/vscode/devcontainers/cpp:0-${VARIANT}
+FROM mcr.microsoft.com/vscode/devcontainers/cpp:0-ubuntu-22.04
 
-ARG BAZELISK_VERSION=v1.10.1
-ARG BAZELISK_DOWNLOAD_SHA=dev-mode
 USER root
 ADD root.sh /tmp/root.sh
 RUN bash /tmp/root.sh && rm /tmp/root.sh
--- a/.devcontainer/swift/update-codeql.sh
+++ b/.devcontainer/swift/update-codeql.sh
@@ -4,9 +4,15 @@ URL=https://github.com/github/codeql-cli-binaries/releases
 LATEST_VERSION=$(curl -L -s -H 'Accept: application/json' $URL/latest | sed -e 's/.*"tag_name":"\([^"]*\)".*/\1/')
 CURRENT_VERSION=v$(codeql version 2>/dev/null | sed -ne 's/.*release \([0-9.]*\)\./\1/p')
 if [[ $CURRENT_VERSION != $LATEST_VERSION ]]; then
-  curl -fSqL -o /tmp/codeql.zip $URL/download/$LATEST_VERSION/codeql-linux64.zip
-  unzip /tmp/codeql.zip -qd /opt
-  rm /tmp/codeql.zip
+  if [[ $UID != 0 ]]; then
+    echo "update required, please run this script with sudo:"
+    echo "  sudo $0"
+    exit 1
+  fi
+  ZIP=$(mktemp codeql.XXXX.zip)
+  curl -fSqL -o $ZIP $URL/download/$LATEST_VERSION/codeql-linux64.zip
+  unzip -q $ZIP -d /opt
+  rm $ZIP
   ln -sf /opt/codeql/codeql /usr/local/bin/codeql
   echo installed version $LATEST_VERSION
 else
--- a/.devcontainer/swift/update-codeql.sh
+++ b/.devcontainer/swift/update-codeql.sh
@@ -1,0 +1,14 @@
+#!/bin/bash -e
+
+URL=https://github.com/github/codeql-cli-binaries/releases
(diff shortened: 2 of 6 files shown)


.devcontainer/swift/user.sh
```diff
@@ -0,0 +1,22 @@
...
+bazel run swift/create-extractor-pack
+
+#install and set up pre-commit
+python3 -m pip install pre-commit --no-warn-script-location
+$HOME/.local/bin/pre-commit install
+
+cat >> $HOME/.bashrc << EOF
+
+# have the codeql binary installed by vscode automatically in PATH
+bin=\$(ls \$HOME/.vscode-remote/data/User/globalStorage/github.vscode-codeql/*/codeql/codeql -t 2>/dev/null | head -1)
```
criemen, line 18: Does this work? I.e. is the codeql vscode extension initialized before this script is run?

.devcontainer/swift/Dockerfile
```diff
@@ -0,0 +1,11 @@
+# See here for image contents: https://github.com/microsoft/vscode-dev-containers/tree/v0.236.0/containers/cpp/.devcontainer/base.Dockerfile
+
+# [Choice] Debian / Ubuntu version (use Debian 11, Ubuntu 18.04/22.04 on local arm64/Apple Silicon): debian-11, debian-10, ubuntu-22.04, ubuntu-20.04, ubuntu-18.04
+ARG VARIANT="bullseye"
```
criemen, line 4: I'd argue that we should rather use a recent Ubuntu build `ubuntu-20.04` or even `ubuntu-22.04` for consistency with what we use elsewhere instead of debian, but I'm not feeling strongly about it.

.devcontainer/swift/Dockerfile
```diff
@@ -0,0 +1,11 @@
...
+FROM mcr.microsoft.com/vscode/devcontainers/cpp:0-${VARIANT}
+
+ARG BAZELISK_VERSION=v1.10.1
+ARG BAZELISK_DOWNLOAD_SHA=dev-mode
```
criemen, line 8: Is this a leftover?

criemen (commented):
some minor nits, otherwise LGTM.

.devcontainer/swift/user.sh
redsun82: no, and in any case user intervention is required any way to download the codeql CLI. So this works only after doing that and eventually restarting the terminal (not the codespace connection). Do you think there is a more user-friendly solution? Should we bypass the codeql CLI installation doing it in the codespace setup?

.devcontainer/swift/Dockerfile
redsun82: these first lines are copied from one of the default dockerfiles for codespaces, and is actually overridden in `devcontainer.json` to be `ubuntu-20.04` (for consistency with the semmle-code version, but I would not mind picking `ubuntu-22.04`. Still, there's probably no need to keep the genericity of the default dockerfile as I have modified it in any case, so I'll change this

.devcontainer/swift/Dockerfile
redsun82: I moved it to the `root.sh` script with the appropriate sha (and bumped up the version while I was at it)

.devcontainer/swift/user.sh
criemen: I missed the part with it being added to bashrc.
That means that zsh users (like me) don't get this at all.

I'd rather just download the latest CLI release and put it on the PATH - vscode can be configured to use it as well with a setting you can set from the devcontainer.

criemen: (arguably, then zsh users still need to set up their PATH manually, but it'd be a bit easier)

redsun82: I've ended up putting the auto-installation in `root.sh`, with a symbolic link in `/usr/local/bin` that will then be picked up automatically by both the vscode plugin and zsh :slightly_smiling_face: 

And a handy `update-codeql` command is also thrown in in the bucket.

criemen: perfect!

criemen (approved):

.devcontainer/swift/Dockerfile
```diff
@@ -0,0 +1,9 @@
+# See here for image contents: https://github.com/microsoft/vscode-dev-containers/tree/v0.236.0/containers/cpp/.devcontainer/base.Dockerfile
+
+# [Choice] Debian / Ubuntu version (use Debian 11, Ubuntu 18.04/22.04 on local arm64/Apple Silicon): debian-11, debian-10, ubuntu-22.04, ubuntu-20.04, ubuntu-18.04
+FROM mcr.microsoft.com/vscode/devcontainers/cpp:0-ubuntu-22.04
+
+USER root
```
KevinHuerta26, line 6: e0784e06801b22b4ef92155ebb4ab9701b2610fe

KevinHuerta26, line 6: https://github.com/github/codeql/blob/main/docs%2Fcodeql%2Freusables%2Fgo-further-reading.rst

.devcontainer/swift/Dockerfile
```diff
@@ -0,0 +1,9 @@
...
+ADD root.sh /tmp/root.sh
+ADD update-codeql.sh /usr/local/bin/update-codeql
```
KevinHuerta26, line 8: https://github.com/github/codeql/tree/main

.devcontainer/swift/Dockerfile
```diff
@@ -0,0 +1,9 @@
+# See here for image contents: https://github.com/microsoft/vscode-dev-containers/tree/v0.236.0/containers/cpp/.devcontainer/base.Dockerfile
+
+# [Choice] Debian / Ubuntu version (use Debian 11, Ubuntu 18.04/22.04 on local arm64/Apple Silicon): debian-11, debian-10, ubuntu-22.04, ubuntu-20.04, ubuntu-18.04
+FROM mcr.microsoft.com/vscode/devcontainers/cpp:0-ubuntu-22.04
```
KevinHuerta26, line 4: 400d7afc94ebe3f87616dd8611cc7d915c15f64c

.devcontainer/swift/Dockerfile
KevinHuerta26, line 6: e0784e06801b22b4ef92155ebb4ab9701b2610fe

.devcontainer/swift/Dockerfile
KevinHuerta26, line 4: api.md

.devcontainer/swift/Dockerfile
KevinHuerta26, line 6: approach.md

.devcontainer/swift/update-codeql.sh
```diff
@@ -0,0 +1,20 @@
...
+    echo "  sudo $0"
+    exit 1
+  fi
+  ZIP=$(mktemp codeql.XXXX.zip)
+  curl -fSqL -o $ZIP $URL/download/$LATEST_VERSION/codeql-linux64.zip
+  unzip -q $ZIP -d /opt
+  rm $ZIP
+  ln -sf /opt/codeql/codeql /usr/local/bin/codeql
+  echo installed version $LATEST_VERSION
+else
```
KevinHuerta26, line 18: - `[]()_****_`

.devcontainer/swift/user.sh
```diff
@@ -0,0 +1,13 @@
+set -xe
+
+# add the workspace to the codeql search path
+mkdir -p /home/vscode/.config/codeql
+echo "--search-path /workspaces/codeql" > /home/vscode/.config/codeql/config
```
KevinHuerta26, line 5:
```suggestion
echo "--search-path /workspaces/codeql" > /home/vscode/.config/codeql/config
```
_****_

.devcontainer/swift/Dockerfile
KevinHuerta26, line 4:
```suggestion
FROM mcr.microsoft.com/vscode/devcontainers/cpp:0-ubuntu-22.04
```

.devcontainer/swift/Dockerfile
```diff
@@ -0,0 +1,9 @@
+# See here for image contents: https://github.com/microsoft/vscode-dev-containers/tree/v0.236.0/containers/cpp/.devcontainer/base.Dockerfile
+
+# [Choice] Debian / Ubuntu version (use Debian 11, Ubuntu 18.04/22.04 on local arm64/Apple Silicon): debian-11, debian-10, ubuntu-22.04, ubuntu-20.04, ubuntu-18.04
```
KevinHuerta26, line 3: # - ~~~~

PR #9623, opened 2022-06-20
Swift: Interpret MaD strings

MathiasVP:
This PR is a follow-up to https://github.com/github/codeql/pull/9555. It fixes the remaining TODOs in `ExternalFlow` and `FlowSummaryImplSpecific` so that we can actually mark stuff as flow sources.

Commit-by-commit review recommended. @rdmarsh2 I'd, in particular, like your comments on f9c8926fcc2fd8da0aba0c9cce64c796240b919e and.


swift/ql/lib/codeql/swift/dataflow/ExternalFlow.qll
```diff
@@ -344,11 +344,109 @@ private predicate elementSpec(
...
+}
+
+/**
+ * Gets a parenthesized string containing all parameter types of this callable, separated by a comma.
+ *
+ * Returns the empty string if the callable has no parameters.
+ * Parameter types are represented by their type erasure.
+ */
+cached
+string paramsString(AbstractFunctionDecl c) {
```
MathiasVP, line 366: This part is taken from Java: https://github.com/github/codeql/blob/9cc10e4511992c0c9282cae1187a140c7e0ec5d5/java/ql/lib/semmle/code/java/dataflow/ExternalFlow.qll#L657

PR #9624, opened 2022-06-20
Implement Kotlin 1.7 support

smowton:
- Implement getIrStubFromDescriptor for Kotlin 1.7
- Stop using ClassSymbol.signature, which is now only populated for classes built from Kotlin, and noteworthily is null for primitive and other internally-synthesised types.


smowton:
Merged via https://github.com/github/codeql/pull/9625

PR #9625, opened 2022-06-20
Switch to using Kotlin 1.7 by default

smowton:
We may not want to merge this just yet, but this will at least test https://github.com/github/codeql/pull/9624 in a Kotlin 1.7 environment.


java/kotlin-extractor/src/main/kotlin/KotlinFileExtractor.kt
```diff
@@ -94,6 +95,9 @@ open class KotlinFileExtractor(
         if (d.isFakeOverride) {
             return true
         }
+        if ((d as? IrFunction)?.descriptor?.isHiddenToOvercomeSignatureClash == true) {
+            return true
+        }
```
tamasvajk, line 100: Can we add an example to such a hidden function in the tests?

java/kotlin-extractor/src/main/kotlin/ExternalDeclExtractor.kt
```diff
@@ -6,28 +6,30 @@ import com.semmle.extractor.java.OdasaOutput
...
     val fieldSignature = ";field"
 
-    fun extractLater(d: IrDeclaration, signature: String): Boolean {
+    fun extractLater(d: IrDeclarationWithName, signature: String): Boolean {
         if (d !is IrClass && !isExternalFileClassMember(d)) {
             logger.errorElement("External declaration is neither a class, nor a top-level declaration", d)
             return false
         }
-        val ret = externalDeclsDone.add(d)
+        val ret = externalDeclsDone.add(Pair(d.fqNameWhenAvailable!!, signature))
```
tamasvajk: This blows up if `d.fqNameWhenAvailable` is `null`. Does `IrDeclarationWithName` guarantee that it's not null?

java/kotlin-extractor/src/main/kotlin/KotlinFileExtractor.kt
smowton: I actually don't know how to deliberately manufacture one yet, but if you revert this change you should see a wide variety of tests fail with Kotlin 1.7.0 because of `CharBuffer.get`

java/kotlin-extractor/src/main/kotlin/ExternalDeclExtractor.kt
smowton: I replaced this with using binary names because they're (a) always available and (b) distinguish overloads that clash in Java but are given unique `JvmName`s.

tamasvajk (approved):

PR #9626, opened 2022-06-20
Kotlin: don't emit synthetic parameter names

smowton:
The QL library already expects these to be missing in some cases and generates its own names when they are absent. Writing synthetic names to the database can produce inconsistencies if the true name is seen later.

### DIFF
--- a/java/kotlin-extractor/src/main/kotlin/KotlinFileExtractor.kt
+++ b/java/kotlin-extractor/src/main/kotlin/KotlinFileExtractor.kt
@@ -507,6 +507,9 @@ open class KotlinFileExtractor(
         return FieldResult(instanceId, instanceName)
     }
 
+    @OptIn(ObsoleteDescriptorBasedAPI::class)
+    private fun hasSynthesizedParameterNames(f: IrFunction) = f.descriptor.hasSynthesizedParameterNames()
+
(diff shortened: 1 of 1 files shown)


java/kotlin-extractor/src/main/kotlin/KotlinFileExtractor.kt
```diff
@@ -523,16 +526,19 @@ open class KotlinFileExtractor(
...
                 extractTypeAccessRecursive(substitutedType, location, id, -1)
             }
-            return extractValueParameter(id, substitutedType, vp.name.asString(), location, parent, idx, useValueParameter(vp, parentSourceDeclaration), vp.isVararg)
+            val syntheticParameterNames = (vp.parent as? IrFunction)?.let { hasSynthesizedParameterNames(it) } ?: true
+            return extractValueParameter(id, substitutedType, vp.name.asString(), location, parent, idx, useValueParameter(vp, parentSourceDeclaration), vp.isVararg, syntheticParameterNames)
         }
     }
 
-    private fun extractValueParameter(id: Label<out DbParam>, t: IrType, name: String, locId: Label<DbLocation>, parent: Label<out DbCallable>, idx: Int, paramSourceDeclaration: Label<out DbParam>, isVararg: Boolean): TypeResults {
+    private fun extractValueParameter(id: Label<out DbParam>, t: IrType, name: String, locId: Label<DbLocation>, parent: Label<out DbCallable>, idx: Int, paramSourceDeclaration: Label<out DbParam>, isVararg: Boolean, syntheticParameterNames: Boolean): TypeResults {
```
atorralba, line 534: NIT: The name of the new parameter `syntheticParameterNames` doesn't read as a boolean — intuitively it sounds like it contains the actual names. It might be redundant, but `hasSynthesizedParameterNames` or similar could be clearer.

atorralba (approved):
Looks plausible to me. Added a minor nitpick that can very well be ignored if you don't think it's an issue.

Also, would it make sense to add a unit test for this? Or is this behavior hard to replicate?

PR #9629, opened 2022-06-21
Swift: ValueParametrizedNode implemented in QL

redsun82:


swift/ql/lib/codeql/swift/elements/decl/ParamDecl.qll
```diff
@@ -1,10 +1,18 @@
 private import codeql.swift.generated.decl.ParamDecl
-private import codeql.swift.elements.decl.AbstractFunctionDecl
+private import codeql.swift.elements.AstNode
+
+abstract class ValueParametrizedNode extends AstNode {
```
MathiasVP, line 4: Having abstract AST classes introduces a horrible footgun (in fact, removing abstract AST classes was one of my good-first-issues for C++ 😂). To see why, consider some user writing a query, and wanting to create a class that encapsulates some function they care about. So they write:
```ql
class MyFunction extends FuncDecl, ValueParametrizedNode /* <-- They really shouldn't have extended `ValueParametrizedNode` here. */ {
  MyFunction() { this.getName() = "foo(x:)" } // Or whatever
}
```

And they _think_ they've restricted `MyFunction` to be those functions whose name is `foo(x:)`, but due to the semantics of abstract classes they've actually extended the set of functions! This means all caching that depends on AST classes (read: everything) must be recomputed. Suddenly their query is super slow because it has to redo both CFG construction and dataflow (and potentially more things as we add more libraries).

This is one of the reasons why `instanceof` was introduced: it fixes the issue that, when looking at a class declaration like:
```ql
class X extends Base { /* ... */ }
```
it's not clear whether we're _restricting_ the set of entities belonging to `Base`, or whether we're _extending_ the set of entities belonging to `Base`.

If we want an "abstract-like" class in the AST layer we have to do it without the `abstract` keyword. Something like:

```ql
class ValueParametrizedNode extends AstNode {
  ValueParametrizedNode() { this instanceof AbstractFunctionDecl or this instanceof ClosureExpr }

  ParamDecl getParam(int index) { none() }

  ParamDecl getAParam() { none() }

  int getNumberOfParams() { none() }
}
```
With this implementation users don't accidentally invalidate the cache if they extend `ValueParametrizedNode`.

MathiasVP (commented):
In general, I'm a big fan of moving as much of this into QL as we can, but I think we should find an approach that doesn't have this giant footgun that I've described below. I've added a suggested alternative, and I'll let Robert comment on what he'd prefer as well.

swift/ql/lib/codeql/swift/elements/decl/ParamDecl.qll
MathiasVP: Also a tiny bikeshedding note: `ValueParametrizedNode` should probably be renamed to something like `Callable`.

PR #9633, opened 2022-06-21
Merge `rc/3.6` into `main`

edoardopirovano:
Merges back the RC branch into `main` to prepare for the next release. Also bumps the minor version of every package that we routinely release, so that if we need to do another release of query changes from the RC branch we have version numbers to do this.


dbartol (approved):
LGTM, including the version number bumps

PR #9634, opened 2022-06-21
JS: add all jquery plugin parameters as source to js/html-constructed-from-input

erik-krogh:
Gets a TP for CVE-2022-30241 

[Evaluation was uneventful](https://github.com/github/codeql-dca-main/tree/data/erik-krogh/jqueryParam__default__CustomSuite/reports).


esbena:
LGTM.

I had to convince myself that this is the right choice for *this* query when it is not the right choice for the plugin query. But the required string concatenation makes a huge difference. Now I wonder if there are additional DOM plugin systems that should be treated in the same way...

esbena (approved):

PR #9638, opened 2022-06-21
Swift: String length conflation tests (for CVE-2022-23625)

geoffw0:
Tests and a bit of wiring for the new string length conflation (CVE-2022-23625) query.  The query in this branch is little more than a TODO, but I'm keen to make this PR anyway so we can discuss the tests / my beginner Swift code.  Also there's a fix to the test `qlpack.yml` that allows `.qlref` files to function correctly.


swift/ql/test/query-tests/Security/CWE-135/StringLengthConflation.swift
```diff
@@ -0,0 +1,125 @@
+
+// --- stubs ---
+
+func print(_ items: Any...) {}
+
+typealias unichar = UInt16
+
+class NSString
```
MathiasVP, line 8: We might want to declare a mockup of [NSObject](https://developer.apple.com/documentation/objectivec/nsobject) and make `NSString` extend that class to keep the types realistic.

swift/ql/test/query-tests/Security/CWE-135/StringLengthConflation.swift
```diff
@@ -0,0 +1,125 @@
...
+{
+    init(string: String) { length = string.count }
+
+    func character(at: Int) -> unichar { return 0 }
+    func substring(from: Int) -> String { return "" }
+    func substring(to: Int) -> String { return "" }
+
+    var length: Int
```
MathiasVP, line 16: Fields like these are normally declared like:
```swift
var length: Int {
  get {
    // code that computes the length
  }
}
```
which makes it a read-only "computed" property (as opposed to a "stored property") since there is no corresponding `get` implementation. Obviously, we can't implement `length` as a computed property, but we can still model the fact that length cannot be modified from outside the class by writing:
```swift
private(set) var length: Int
```

MathiasVP (approved):
Generally LGTM! A few Swift-specific comments, but feel free to ignore them for now if you want.

swift/ql/test/query-tests/Security/CWE-135/StringLengthConflation.swift
```diff
@@ -0,0 +1,125 @@
+
+// --- stubs ---
+
+func print(_ items: Any...) {}
```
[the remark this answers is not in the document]
geoffw0: Yes, I am specifically _not_ copy-pasting from the Apple docs into our test code.  I wrote this myself and saw no need to support `separator` or `terminator` in the test.

swift/ql/test/query-tests/Security/CWE-135/StringLengthConflation.swift
geoffw0: Will do.

swift/ql/test/query-tests/Security/CWE-135/StringLengthConflation.swift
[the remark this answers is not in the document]
MathiasVP: That's fair 👍.

swift/ql/test/query-tests/Security/CWE-135/StringLengthConflation.swift
geoffw0: Done.

swift/ql/test/query-tests/Security/CWE-135/StringLengthConflation.swift
geoffw0: Done.

MathiasVP (approved):
LGTM!

PR #9639, opened 2022-06-21
Swift: some expression extractions

redsun82:


AlexDenisov (approved):

swift/codegen/schema.yml
```diff
@@ -445,7 +446,7 @@ LookupExpr:
   _extends: Expr
   _children:
     base_expr: Expr
-  member: Decl
+  member: Decl?
```
redsun82, line 449: In case you wonder, I had an assertion firing on this :slightly_smiling_face:

AlexDenisov (approved):

swift/codegen/schema.yml
MathiasVP: OOI which lookup expr didnt have a decl? 🤔

PR #9641, opened 2022-06-21
Make ActiveRecordInstanceMethodCall Public

boveus:
👋  Hi folks.  I am writing a CodeQL query and it would be helpful to make use of the `ActiveRecordInstanceMethodCall` class.  I currently have it copied in my query, but it seems like it would be easier to maintain if I could call the class directly.  Other folks may find this class useful in the future as well.


ruby/ql/lib/codeql/ruby/frameworks/ActiveRecord.qll
```diff
@@ -331,7 +331,7 @@ class ActiveRecordInstance extends DataFlow::Node {
 }
 
 // A call whose receiver may be an active record model object
-private class ActiveRecordInstanceMethodCall extends DataFlow::CallNode {
+class ActiveRecordInstanceMethodCall extends DataFlow::CallNode {
```
alexrford, line 334:
```suggestion
/**
 * A call whose receiver may be an `ActiveRecordInstance`.
 */
class ActiveRecordInstanceMethodCall extends DataFlow::CallNode {
```

alexrford (changes requested):
Yep, I'm fine with making this a public class.

boveus:
@hmac could you help me out with these failing CI checks?  Will this change require a change note?

It also looks like the qlformat check is failing, but it isn't clear why.

boveus:
@hmac - I just pushed up the formatting change, but unfortunately I don't have write access to the this repo.  Could you please add the `no-change-note-required` label for me?  Thanks.

boveus:
@alexrford does this look good to you?

alexrford:
Sorry, I didn't realize that you couldn't merge this - merging now.

PR #9644, opened 2022-06-22
C++: Ensure we can round trip between (forward) class declarations

jketema:
This was already possible when the forward class declaration and the class definition occurred in the same scope. However, there is a common C++ usage pattern in which this is not the case (when only a pointer to the class is needed). In this latter scenario we could not round trip between the (forward) `DeclarationEntry` and the `Declaration`.
    
Effectively this changes the code to:
```
if exists(TypeDeclarationEntry e | e.getType() = this)
then result.getType() = this
else ...
```
We use `type_decls` instead to stay close to the original code.

### DIFF
--- a/cpp/ql/lib/change-notes/2022-06-22-class-declaration-entry-fix.md
+++ b/cpp/ql/lib/change-notes/2022-06-22-class-declaration-entry-fix.md
@@ -1,0 +1,4 @@
+---
+category: fix
+---
+* `UserType.getADeclarationEntry()` now yields all forward declarations when the user type is a `class`, `struct`, or `union`.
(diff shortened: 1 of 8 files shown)


MathiasVP (dismissed):
Nice find! LGTM is DCA is happy 🤞.

jketema:
Yeah, this confused me quite a lot yesterday. Indeed let's see what DCA has to say.

jketema:
@MathiasVP DCA completed. Would you mind having a look at the results? I ran with tuple counting enabled, and I'm not sure if there's anything concerning int there.

MathiasVP:
> @MathiasVP DCA completed. Would you mind having a look at the results? I ran with tuple counting enabled, and I'm not sure if there's anything concerning in there.

I think it all looks good 👍. It's a bit difficult to read the tuple-count tables since this PR changes a fundamental AST class that most higher-level predicates depend on (which means all dependent predicates get a different hash).

MathiasVP (approved):

PR #9659, opened 2022-06-22
Java: Report log-injection at the source rather than the sink

smowton:
This should remove the problem of excessive grouping of different alerts that share a sink location, often due to wrapper functions that form the ultimate sink of all logging calls in a given codebase.


aschackmull (approved):

aibaars:
@smowton @aschackmull While I think this change is good (I suggested it myself elsewhere) there is a drawback. The old alerts reported on the sinks will be closed and instead (possibly many more) new alerts will be created on the source positions. This could confuse users and may cause an increase in support questions in the short term.

PR #9667, opened 2022-06-22
JS: create downgrades pack

nickrolfe:


nickrolfe:
N.B. I haven't tested this at all. For now, somebody could test it by following [these instructions](https://github.com/github/codeql/blob/main/docs/prepare-db-upgrade.md#doing-the-downgrade-manually) (search for the paragraph starting "To test the downgrade script...").

cklin:
> N.B. I haven't tested this at all. For now, somebody could test it by following [these instructions](https://github.com/github/codeql/blob/main/docs/prepare-db-upgrade.md#doing-the-downgrade-manually) (search for the paragraph starting "To test the downgrade script...").

I tested it by adding `codeql_check_db_downgrades('javascript')` to the `build` file and running `target/test/check-db-downgrades-javascript`.  The test passed, so at least the downgrade is well-formed and applies successfully to an empty database.

Once this PR is merged and the `semmle-code/ql` submodule bumped, I can put up a PR to add the downgrade script check for JS and hook that up to `checks-javascript-language-tests-linux`.

cklin (approved):
LGTM in terms of the general structure of the downgrade pack and the downgrade script.

PR #9668, opened 2022-06-22
Swift: Make sure property setters and getters also have `ExprNodes`

MathiasVP:
This should fix a problem @geoffw0 has been having with `MemberRefExpr`s not having an `ExprNode`. The problem is that the CFG library maps these to `PropertyGetterElement`s to model them as calls to the (sometimes compiler-generated) functions. The same applies to `PropertySetterElement`.

There might be an issue where we now have several `ExprNodes` mapping to the same `Expr`, but I think we can fix that later by restricting which nodes will have an `AstNode` in the CFG library (in particular, the nodes for which there exists a `PropertyGetterElement` or `PropertySetterElement` shouldn't have an `AstNode`).

What do you think about this solution, @rdmarsh2?

### DIFF
--- a/swift/ql/lib/codeql/swift/controlflow/BasicBlocks.qll
+++ b/swift/ql/lib/codeql/swift/controlflow/BasicBlocks.qll
@@ -214,15 +214,15 @@ private module JoinBlockPredecessors {
   }
 
   int getId(JoinBlockPredecessor jbp) {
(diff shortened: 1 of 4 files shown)


geoffw0:
Yep, this solves my problem, thanks!

I'd like to wait for @rdmarsh2 s review before merging.

PR #9673, opened 2022-06-22
Swift: String length conflation query

geoffw0:
String length conflation query, i.e. query for [CVE-2022-23625](https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2022-23625) ([patch](https://github.com/wireapp/wire-ios-transport/commit/02e90aa45edaf7eb2d8b97fa2377cd8104274170)) / [CWE-135](https://cwe.mitre.org/data/definitions/135.html).

~This is a draft PR because there are a bunch of things I need to clean up that I'm looking for advice on - see the `TODO` comments in the code.  Many of these will involve small additions in the libraries and I'm seeking discussion of the design of these improvements.~

~The query is also missing qldoc, qhelp, Swift example code, query metadata, a proper violation message, change note ... and results in the `NSString` -> `Swift.String` direction.  I don't anticipate any problems in these areas.~  UPDATE: the rest of this will be addressed in a follow-up PR.

### DIFF
--- a/swift/ql/src/queries/Security/CWE-135/StringLengthConflation.qhelp
+++ b/swift/ql/src/queries/Security/CWE-135/StringLengthConflation.qhelp
@@ -1,0 +1,34 @@
+<!DOCTYPE qhelp PUBLIC
+  "-//Semmle//qhelp//EN"
+  "qhelp.dtd">
+<qhelp>
+<overview>
+<p>Using a length value from an <code>NSString</code> in a <code>String</code>, or a count from a <code>String</code> in an <code>NSString</code>, may cause unexpected behavior. This is because certain unicode sequences are represented as one character in a <code>String</code> but as a sequence of multiple characters in an <code>NSString</code>. For example, a 'thumbs up' emoji with a skin tone modifier (&#x1F44D;&#x1F3FF;) is represented as U+1F44D (&#x1F44D;) then the modifier U+1F3FF.</p>
+
+</overview>
+<recommendation>
+
+<p>Use <code>String.count</code> when working with a <code>String</code>. Use <code>NSString.length</code> when working with an <code>NSString</code>. Do not mix values for lengths and offsets between the two types as they are not compatible measures.</p>
+
+<p>If you need to convert between <code>Range</code> and <code>NSRange</code>, do so directly using the appropriate constructor. Do not attempt to use incompatible length and offset values to accomplish conversion.</p>
+
+</recommendation>
+<example>
+
+<p>In the following example, a <code>String</code> is converted to <code>NSString</code>, but a range is created from the <code>String</code> to do some processing on it.</p>
+
+<sample src="StringLengthConflationBad.swift" />
+
+<p>This is dangerous because, if the input contains certain characters, the range computed on the <code>String</code> will be wrong for the <code>NSString</code>.  This will lead to incorrect behaviour in the string processing that follows. To fix the problem, we can use <code>NSString.length</code> to create the <code>NSRange</code> instead, as follows:</p>
+
+<sample src="StringLengthConflationGood.swift" />
+
+</example>
+<references>
+
+<li>
+  <a href="https://talk.objc.io/episodes/S01E80-swift-string-vs-nsstring">Swift String vs. NSString</a>
+</li>
+
+</references>
+</qhelp>
--- a/swift/ql/src/queries/Security/CWE-135/StringLengthConflation.qhelp
+++ b/swift/ql/src/queries/Security/CWE-135/StringLengthConflation.qhelp
@@ -3,7 +3,7 @@
   "qhelp.dtd">
 <qhelp>
 <overview>
-<p>Using a length value from an <code>NSString</code> in a <code>String</code>, or a count from a <code>String</code> in an <code>NSString</code>, may cause unexpected behavior. This is because certain unicode sequences are represented as one character in a <code>String</code> but as a sequence of multiple characters in an <code>NSString</code>. For example, a 'thumbs up' emoji with a skin tone modifier (&#x1F44D;&#x1F3FF;) is represented as U+1F44D (&#x1F44D;) then the modifier U+1F3FF.</p>
+<p>Using a length value from an <code>NSString</code> in a <code>String</code>, or a count from a <code>String</code> in an <code>NSString</code>, may cause unexpected behavior including (in some cases) buffer overwrites. This is because certain unicode sequences are represented as one character in a <code>String</code> but as a sequence of multiple characters in an <code>NSString</code>. For example, a 'thumbs up' emoji with a skin tone modifier (&#x1F44D;&#x1F3FF;) is represented as U+1F44D (&#x1F44D;) then the modifier U+1F3FF.</p>
 
 </overview>
 <recommendation>
--- a/swift/ql/src/queries/Security/CWE-135/StringLengthConflation.ql
+++ b/swift/ql/src/queries/Security/CWE-135/StringLengthConflation.ql
@@ -20,7 +20,7 @@ class StringLengthConflationConfiguration extends DataFlow::Configuration {
   override predicate isSource(DataFlow::Node node, string flowstate) {
     // result of a call to to `String.count`
     exists(MemberRefExpr member |
-      member.getBaseExpr().getType().toString() = "String" and // TODO: use of toString
+      member.getBaseExpr().getType().getName() = "String" and
       member.getMember().toString() = "count" and // TODO: use of toString
       node.asExpr() = member and
       flowstate = "String"
--- a/swift/ql/src/queries/Security/CWE-135/StringLengthConflation.ql
+++ b/swift/ql/src/queries/Security/CWE-135/StringLengthConflation.ql
@@ -58,14 +58,7 @@ class StringLengthConflationConfiguration extends DataFlow::Configuration {
       c.getAMember() = f and // TODO: will this even work if its defined in a parent class?
       call.getFunction().(ApplyExpr).getFunction().(DeclRefExpr).getDecl() = f and
       call.getFunction().(ApplyExpr).getFunction().toString() = methodName and // TODO: use of toString
-      call.getFunction()
-          .(ApplyExpr)
-          .getFunction()
-          .(DeclRefExpr)
-          .getDecl()
-          .(AbstractFunctionDecl)
-          .getParam(arg)
-          .getName() = argName and
+      f.getParam(arg).getName() = argName and
       call.getArgument(arg).getExpr() = node.asExpr() and
       flowstate = "String" // `String` length flowing into `NSString`
     )
@@ -73,9 +66,9 @@ class StringLengthConflationConfiguration extends DataFlow::Configuration {
     // arguments to function calls...
     exists(string funcName, string argName, CallExpr call, int arg |
       // `NSMakeRange`
-      funcName = "NSMakeRange" and
+      funcName = "NSMakeRange(_:_:)" and
       argName = ["loc", "len"] and
-      call.getStaticTarget().getName().matches(funcName + "%") and
+      call.getStaticTarget().getName() = funcName and
       call.getStaticTarget().getParam(arg).getName() = argName and
       call.getArgument(arg).getExpr() = node.asExpr() and
       flowstate = "String" // `String` length flowing into `NSString`
@@ -85,4 +78,4 @@ class StringLengthConflationConfiguration extends DataFlow::Configuration {
 
 from StringLengthConflationConfiguration config, DataFlow::PathNode source, DataFlow::PathNode sink
 where config.hasFlowPath(source, sink)
-select sink, source, sink, "RESULT"
+select sink.getNode(), source, sink, "RESULT"
--- a/swift/ql/src/queries/Security/CWE-135/StringLengthConflation.ql
+++ b/swift/ql/src/queries/Security/CWE-135/StringLengthConflation.ql
@@ -21,7 +21,7 @@ class StringLengthConflationConfiguration extends DataFlow::Configuration {
     // result of a call to to `String.count`
     exists(MemberRefExpr member |
       member.getBaseExpr().getType().getName() = "String" and
-      member.getMember().toString() = "count" and // TODO: use of toString
+      member.getMember().(VarDecl).getName() = "count" and
       node.asExpr() = member and
       flowstate = "String"
     )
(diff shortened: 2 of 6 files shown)


swift/ql/src/queries/Security/CWE-135/StringLengthConflation.ql
```diff
@@ -11,5 +11,78 @@
...
+        methodName = "substring" and
+        argName = ["from", "to"]
+        or
+        // `NSMutableString.insert`
+        className = "NSMutableString" and
+        methodName = "insert" and
+        argName = "at"
+      ) and
+      c.toString() = className and // TODO: use of toString
+      c.getAMember() = f and // TODO: will this even work if its defined in a parent class?
```
rdmarsh2, line 58: No, it won't. I don't think we've implemented anything with the inheritance tree yet.

swift/ql/src/queries/Security/CWE-135/StringLengthConflation.ql
```diff
@@ -11,5 +11,78 @@
...
+      call.getFunction().(ApplyExpr).getFunction().(DeclRefExpr).getDecl() = f and
```
MathiasVP, line 59: If you make sure that this branch includes https://github.com/github/codeql/pull/9623/commits/e2478d1106bb464f3171b33a0c1375e17eabd48d then I think this should be equivalent
```suggestion
      call.getStaticTarget() = f and
```

swift/ql/src/queries/Security/CWE-135/StringLengthConflation.ql
```diff
@@ -11,5 +11,78 @@
...
+      call.getFunction().(ApplyExpr).getFunction().toString() = methodName and // TODO: use of toString
```
MathiasVP, line 60: This should be the same, right? Note that `getName` is slightly different from the `toString` in some cases (which is something we should fix eventually). For instance, I expect the `getName` of https://developer.apple.com/documentation/foundation/nsstring/1414645-character to be `character(at:)`, so your `isSink` should probable be updated.
```suggestion
      f.getName() = methodName and
```

swift/ql/src/queries/Security/CWE-135/StringLengthConflation.ql
```diff
@@ -11,5 +11,78 @@
...
+      call.getFunction()
+          .(ApplyExpr)
+          .getFunction()
+          .(DeclRefExpr)
+          .getDecl()
+          .(AbstractFunctionDecl)
+          .getParam(arg)
+          .getName() = argName and
```
MathiasVP, line 68: I hope this is equivalent
```suggestion
      f.getParam(arg).getName() = argName and
```

swift/ql/src/queries/Security/CWE-135/StringLengthConflation.ql
```diff
@@ -11,5 +11,78 @@
...
+      call.getArgument(arg).getExpr() = node.asExpr() and
+      flowstate = "String" // `String` length flowing into `NSString`
+    )
+    or
+    // arguments to function calls...
+    exists(string funcName, string argName, CallExpr call, int arg |
+      // `NSMakeRange`
+      funcName = "NSMakeRange" and
+      argName = ["loc", "len"] and
+      call.getStaticTarget().getName().matches(funcName + "%") and
```
MathiasVP, line 78: If you instead write `funcName = NSMakeRange(_:_:)` then I hope you can just do `call.getStaticTarget().getName() = funcName` and avoid the `matches`. The name of Swift functions by convention includes the name of its parameters (each one terminated by a colon).

swift/ql/src/queries/Security/CWE-135/StringLengthConflation.ql
```diff
@@ -11,5 +11,78 @@
...
+      call.getStaticTarget().getParam(arg).getName() = argName and
```
MathiasVP, line 79: I'm surprised this works 🤔. I thought the names of arguments would both be `_` given the definition of `NSMakeRange` in https://github.com/github/codeql/pull/9638/files#diff-b282853b05abe71aceb437d459bce12406e24ec056e6004c512ee8575b001992R35

swift/ql/src/queries/Security/CWE-135/StringLengthConflation.ql
```diff
@@ -11,5 +11,78 @@
...
+      call.getArgument(arg).getExpr() = node.asExpr() and
+      flowstate = "String" // `String` length flowing into `NSString`
+    )
+  }
+}
+
+from StringLengthConflationConfiguration config, DataFlow::PathNode source, DataFlow::PathNode sink
+where config.hasFlowPath(source, sink)
+select sink, source, sink, "RESULT"
```
MathiasVP, line 88: This is the usual pattern we use for `path-problem`s. As for `"RESULT"`, how about something like: `"NSRange" object constructed from string count value.` (with a suitable replacement for the string `"NSRange"` that matches what the user had actually written)?
```suggestion
select sink.getNode(), source, sink, "RESULT"
```

MathiasVP (commented):
This looks great, Geoffrey! Here are my initial comments.

swift/ql/src/queries/Security/CWE-135/StringLengthConflation.ql
```diff
@@ -11,5 +11,78 @@
...
+class StringLengthConflationConfiguration extends DataFlow::Configuration {
+  StringLengthConflationConfiguration() { this = "StringLengthConflationConfiguration" }
+
+  override predicate isSource(DataFlow::Node node, string flowstate) {
+    // result of a call to to `String.count`
+    exists(MemberRefExpr member |
+      member.getBaseExpr().getType().toString() = "String" and // TODO: use of toString
+      member.getMember().toString() = "count" and // TODO: use of toString
+      node.asExpr() = member and
+      flowstate = "String"
```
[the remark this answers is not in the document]
geoffw0: It will be needed to track results in the `NSString` -> `Swift.String` direction separately from results in the `Swift.String` -> `NSString` direction.  The tests have plenty of examples in both directions, but as yet the query only supports the latter.

swift/ql/src/queries/Security/CWE-135/StringLengthConflation.ql
MathiasVP: `_ loc: Int` means "when you call the function you can't specify an argument name, but _inside_ the function body the parameter is named `loc`).

MathiasVP: Ah, right. This all makes sense now. What you're calling `argName` is really the name of a parameter. In Swift these are two different things: A function `func foo(x y : Int) { .. }` is called like `foo(x : 42)`, but inside ` ... ` you refer to the parameter as `y`.

swift/ql/src/queries/Security/CWE-135/StringLengthConflation.ql
geoffw0: Doesn't work unfortunately, but it looks like what I should be doing.  I'll look into it.

swift/ql/src/queries/Security/CWE-135/StringLengthConflation.ql
geoffw0: Well spotted, thanks.  Done.

swift/ql/src/queries/Security/CWE-135/StringLengthConflation.ql
geoffw0: Done.  But more importantly I need to implement `getName` in the appropriate places that we can move away from `toString`.

swift/ql/src/queries/Security/CWE-135/StringLengthConflation.ql
geoffw0: Done the `.getNode()` change.

I'll have a think about the message, thanks for the suggestion.

geoffw0:
Some comments addressed, more in progress.  Sorry I haven't 'accepted' the suggestions, I want to do everything locally and test before pushing.

swift/ql/src/queries/Security/CWE-135/StringLengthConflation.ql
geoffw0: Also doesn't work right now, I'll check I have that library fix that made this work on constructors.

MathiasVP: Please do! It certainly appeared to do that when looking at the [diff](https://github.com/github/codeql/commit/e2478d1106bb464f3171b33a0c1375e17eabd48d#diff-2e7af3b9876580e16b61a25d4f8e824927a2e2be4ac855964b0d27cad4101df7R47) 🤔.

MathiasVP: Oh, wait. Yeah, I see why my change didn't work: you're writing is equivalent to:
```swift
call.getFunction().(ApplyExpr).getStaticTarget() = f
```
Now, why you need that second layer of indirection isn't clear to me.

geoffw0:
I've got rid of all the uses of `toString` now.  I'd like to draw attention to my addition of `Type.getName` - is my implementation correct / sufficiently stable?

swift/ql/src/queries/Security/CWE-135/StringLengthConflation.ql
geoffw0: Done.

swift/ql/src/queries/Security/CWE-135/StringLengthConflation.ql
geoffw0: Yep, that works (change made).  I assume the second level of indirection is because `s.count` itself is a call to an implicit `ApplyExpr` on `s` that returns the member function `count`???  I'm not sure at all though.

MathiasVP:
> I've got rid of all the uses of toString now. I'd like to draw attention to my addition of Type.getName - is my implementation correct / sufficiently stable?

I think this is fine. We should definitely have a `getName` predicate on `Type`, and forwarding to `getDiagnosticsName` is probably the most sensible thing to do right now 👍. So this LGTM!

swift/ql/src/queries/Security/CWE-135/StringLengthConflation.ql
geoffw0: I think maybe a chunk of this logic should be made to look something like this:
```
...
c.getMember(methodName) = f and
call.get???Target() = f and
...
```

swift/ql/src/queries/Security/CWE-135/StringLengthConflation.ql
geoffw0: Works now (with other changes).  :+1:

swift/ql/src/queries/Security/CWE-135/StringLengthConflation.ql
MathiasVP: Ahhh... Yeah, now I see what's going on. Gosh, this is a bit annoying:

For this piece of code:
```swift
let ns = NSString(string: s)
```
There are two function calls. This is because it's represented as `(NSString.init(this))(string: s)`. So the inner-most function supplies the `this` argument, and returns the function to be invokes, and the outer-most function actually does the call.

We should figure out what the best way of navigating this in QL is. For now, it looks like you've worked around the issue quite well here 👍.

swift/ql/src/queries/Security/CWE-135/StringLengthConflation.qhelp
```diff
@@ -0,0 +1,34 @@
...
+<qhelp>
+<overview>
+<p>Using a length value from an <code>NSString</code> in a <code>String</code>, or a count from a <code>String</code> in an <code>NSString</code>, may cause unexpected behavior. This is because certain unicode sequences are represented as one character in a <code>String</code> but as a sequence of multiple characters in an <code>NSString</code>. For example, a 'thumbs up' emoji with a skin tone modifier (&#x1F44D;&#x1F3FF;) is represented as U+1F44D (&#x1F44D;) then the modifier U+1F3FF.</p>
+
+</overview>
+<recommendation>
+
+<p>Use <code>String.count</code> when working with a <code>String</code>. Use <code>NSString.length</code> when working with an <code>NSString</code>. Do not mix values for lengths and offsets between the two types as they are not compatible measures.</p>
+
+<p>If you need to convert between <code>Range</code> and <code>NSRange</code>, do so directly using the appropriate constructor. Do not attempt to use incompatible length and offset values to accomplish conversion.</p>
```
MathiasVP, line 13: Should we also mention using string.utf16.count like how it was done in the patch?

swift/ql/src/queries/Security/CWE-135/StringLengthConflation.ql
```diff
@@ -11,5 +11,76 @@
...
+        or
+        // `NSMutableString.insert`
+        className = "NSMutableString" and
+        methodName = "insert(_:at:)" and
+        paramName = "at"
+      ) and
+      c.getName() = className and
+      c.getAMember() = f and // TODO: will this even work if its defined in a parent class?
+      call.getFunction().(ApplyExpr).getStaticTarget() = f and
+      f.(AbstractFunctionDecl).getName() = methodName and
```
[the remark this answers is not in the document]
geoffw0: This was fixed.

swift/ql/src/queries/Security/CWE-135/StringLengthConflation.qhelp
geoffw0: Yeah, I think I'm going to do that (including any doc changes) as follow-up though, because its a fairly big extension and I haven't yet figured it all out.  I *think* `string.utf16.count` *is* equivalent of `NSString.length`, for example.

geoffw0:
Docs team - please would you review the new `qhelp` file.  This is for our first query for Swift, it isn't going to be used by customers for a while yet, but I'd like to set a good example for future queries to follow.

Also, I'd like to apologize for putting emojis in the doc.  They are on topic, I promise.

swift/ql/src/queries/Security/CWE-135/StringLengthConflation.ql
geoffw0: There's a proper message as well now.

swift/ql/src/queries/Security/CWE-135/StringLengthConflation.ql
geoffw0: I'm now wondering if it matters.  Is something like the following a situation we have to be concerned for in Swift?
```
class NSStringBase : NSObject
{
    // some of the implementation of NSString here
}

class NSString : NSStringBase
{
    // rest of the implementation of NSString
}
```
... or can we safely assume the implementations are in the places the docs say they are?

swift/ql/src/queries/Security/CWE-135/StringLengthConflation.qhelp
```diff
@@ -0,0 +1,34 @@
+<!DOCTYPE qhelp PUBLIC
+  "-//Semmle//qhelp//EN"
+  "qhelp.dtd">
+<qhelp>
+<overview>
+<p>Using a length value from an <code>NSString</code> in a <code>String</code>, or a count from a <code>String</code> in an <code>NSString</code>, may cause unexpected behavior. This is because certain unicode sequences are represented as one character in a <code>String</code> but as a sequence of multiple characters in an <code>NSString</code>. For example, a 'thumbs up' emoji with a skin tone modifier (&#x1F44D;&#x1F3FF;) is represented as U+1F44D (&#x1F44D;) then the modifier U+1F3FF.</p>
```
geoffw0, line 6: On reflection I wonder if I should use a stronger phrase than "unexpected behavior"?  It turns out that it is possible to directly cause a crash / buffer overrun using an invalid `NSRange` to operate on an `NSString` (at least in an online Swift playground), which I had not expected.

rdmarsh2: NSString is actually an Objective C class, so I'm not terribly surprised. I'd definitely use a stronger phrase.

swift/ql/src/queries/Security/CWE-135/StringLengthConflation.ql
MathiasVP: I think that's something time will have to tell 😂. For now I think it's fine to assume the docs are precise about it

rdmarsh2: NSString is actually an Objective-C class, so it's probably something we need to worry about.

MathiasVP: That's true. Looking at the ClassDecl for NSString in an image I have of Swift it doesn't look like it has any members at all :( this might be because it's defined in Obj-C code, or it might be due to some other extractor weirdness. In any case, the query right now might not work as well on a real snapshot as it does on the tests

swift/ql/src/queries/Security/CWE-135/StringLengthConflation.qhelp
MathiasVP: Agreed

swift/ql/src/queries/Security/CWE-135/StringLengthConflation.ql
geoffw0: Thanks, I'll keep that in mind when I test on real code, which is high on my priorities right now.

swift/ql/src/queries/Security/CWE-135/StringLengthConflation.qhelp
geoffw0: Done.  I've gone with:
```
may cause unexpected behavior including (in some cases) buffer overwrites.
```

geoffw0:
Yep, any comments from the doc team here will be picked up and addressed in follow-up.

geoffw0:
@github/docs-content-codeql we're still waiting for the docs review on this.

geoffw0:
Thanks!

PR #9679, opened 2022-06-23
JS: Downgrade ast_node_symbol relation

asgerf:
Avoids introducing a DB inconsistency when downgrading


nickrolfe (approved):
The downgrade makes sense to me - it drops `ast_node_symbol` tuples where the node is an `@external_module_declaration.

I ran a few hundred of the JS qltests on an old checkout of `github/codeql`, from before the dbscheme change, using a dist that includes the recent extractor and dbchanges, as well as this downgrade script. They passed, and I observed this downgrade script being applied in the output of `codeql test run -vv ...`.

PR #9681, opened 2022-06-23
Kotlin: reintroduce obinit when we have multiple secondary constructors and no primary

smowton:
This avoids DB inconsistencies because complex initialisers are extracted to more than one function.

### DIFF
--- a/java/ql/consistency-queries/calls.ql
+++ b/java/ql/consistency-queries/calls.ql
@@ -3,5 +3,8 @@ import java
 from MethodAccess ma
 // Generally Kotlin calls will always use an explicit qualifier, except for calls
 // to the synthetic instance initializer <obinit>, which use an implicit `this`.
-where not exists(ma.getQualifier()) and ma.getFile().isKotlinSourceFile() and not ma.getCallee() instanceof InstanceInitializer
+where
+  not exists(ma.getQualifier()) and
+  ma.getFile().isKotlinSourceFile() and
+  not ma.getCallee() instanceof InstanceInitializer
 select ma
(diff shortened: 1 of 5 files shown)


java/ql/consistency-queries/calls.ql
```diff
@@ -1,5 +1,7 @@
 import java
 
 from MethodAccess ma
-where not exists(ma.getQualifier()) and ma.getFile().isKotlinSourceFile()
+// Generally Kotlin calls will always use an explicit qualifier, except for calls
+// to the synthetic instance initializer <obinit>, which use an implicit `this`.
+where not exists(ma.getQualifier()) and ma.getFile().isKotlinSourceFile() and not ma.getCallee() instanceof InstanceInitializer
```
tamasvajk, line 6: Should we add an explicit `this` qualifier to the extraction of `IrInstanceInitializerCall`?

tamasvajk (approved):
Looks plausible to me.

java/ql/consistency-queries/calls.ql
smowton: We could try; not sure how to make the right expression though so figured it was better to just rely on the QL lib's existing support for implicit `this`.

tamasvajk: I thought it would be simply like this: https://github.com/github/codeql/blob/eb1b3f801d080ec2f02e5c3bbbd2a697de0470f0/java/kotlin-extractor/src/main/kotlin/KotlinFileExtractor.kt#L2559-L2566

PR #9685, opened 2022-06-23
Swift: split up generated C++ code

redsun82:
To aid incremental builds, generated C++ is now split up between header and source, as well as in separate directories. This way for example the `DeclVisitor` can include `swift/extractor/trap/generated/decl/TrapClasses.h` without including all the rest of the definitions.

Notice that in order to achieve this in trap entries, the information needed to be added to the dbscheme file in form of special `//dir=foo` comments.


redsun82:
you can have a look at [the generated files](https://github.com/github/codeql/suites/7060670724/artifacts/278407418) to see what it all does.

AlexDenisov (approved):

PR #9677, opened 2022-06-23
DataFlow: Improve join order in `additionalJumpStep`

jketema:
Before:

```
[2022-05-14 19:07:41] Evaluated non-recursive predicate DataFlowImpl3::additionalJumpStep#c0e9fca6#fff#shared@91a642tl in 4035ms (size: 44500020).
Tuple counts for DataFlowImpl3::additionalJumpStep#c0e9fca6#fff#shared@91a642tl:
        44724248  ~6%    {3} r1 = SCAN num#DataFlowImpl3::TNodeNormal#c0e9fca6#ff OUTPUT "AdjustedConfiguration", In.0, In.1
        44500020  ~6%    {3} r2 = r1 AND NOT DataFlowImpl3::fullBarrier#c0e9fca6#ff(Lhs.2, "AdjustedConfiguration")
                         return r2

[2022-05-14 19:07:53] Evaluated non-recursive predicate DataFlowImpl3::additionalJumpStep#c0e9fca6#fff#shared#1@84e9d2av in 9555ms (size: 10806062).
Tuple counts for DataFlowImpl3::additionalJumpStep#c0e9fca6#fff#shared#1@84e9d2av:
        44500020  ~0%    {4} r1 = SCAN DataFlowImpl3::additionalJumpStep#c0e9fca6#fff#shared OUTPUT In.1, "AdjustedConfiguration", In.1, In.2
        44500018  ~1%    {4} r2 = JOIN r1 WITH DataFlowImplCommon::Cached::nodeEnclosingCallable#59e7a193#ff ON FIRST 1 OUTPUT Rhs.1, "AdjustedConfiguration", Lhs.2, Lhs.3
        44500018  ~0%    {4} r3 = JOIN r2 WITH @declaration#f ON FIRST 1 OUTPUT Lhs.2, "AdjustedConfiguration", Lhs.3, Lhs.0
        11191390  ~2%    {4} r4 = JOIN r3 WITH DataFlowImpl3::Configuration::isAdditionalFlowStep#dispred#f0820431#fff#cpe#23 ON FIRST 1 OUTPUT Rhs.1, "AdjustedConfiguration", Lhs.2, Lhs.3
        11191390  ~7%    {5} r5 = JOIN r4 WITH num#DataFlowImpl3::TNodeNormal#c0e9fca6#ff ON FIRST 1 OUTPUT "AdjustedConfiguration", Lhs.2, Lhs.3, Lhs.0, Rhs.1
[...]

PR #9686, opened 2022-06-23
Dataflow performance: Avoid node scans

aschackmull:
When constructing the config-specific step relations, put the base step relations first to avoid scans of `DataFlow::Node`. This mostly only matters for very large databases where we've seen it pop up from time to time as a source of bad performance. Let's see what DCA says.


MathiasVP:
Looks like this PR subsumes https://github.com/github/codeql/pull/9677?

jketema:
> Looks like this PR subsumes https://github.com/github/codeql/pull/9677?

Yup.

jketema:
It seems that all the DCA experiments failed for reasons outside our control 😔

jketema:
I re-ran all the DCA experiments. These suggest that there's a minor, but noticeable performance improvement for all languages.

jketema (approved):

PR #9691, opened 2022-06-23
Swift: remove check for inout params in SSA gen

rdmarsh2:
This check is unnecessary since it's enforced by the compiler, and is causing a bad join order.


MathiasVP (approved):
LGTM! And I guess it's also nice that we now don't assume that it's a call to something we can resolve statically 👍.

PR #9700, opened 2022-06-24
C++: Ensure only one `Variable` exists for every global variable

jketema:
Depending on the extraction order, before this change there might be multiple `GlobalVariable`s per declared global variable. See the tests in `cpp/ql/test/library-tests/variables/global`. This change ensures that only one of those `GlobalVariable`s is visible to the user if we can locate a unique definition. If not, the old situation persists.

Note that an exception needs to be made for templated variables. Here, the definition refers to the non-instantiated template, while a declaration that is not a definition refers to an instantiation. In case the instantiation refers to a template parameter, the mangled names of the template and the instantiation will be identical. This happens for example in the following case:
```
template <typename T>
T x = T(42);           // Uninstantiated templated variable

template <typename T>
class C {
  T y = x<T>;          // Instantiation using a template parameter
};
```
Since the uninstantiated template and the instantiation are two different entities, we do not unify them as described above.

To note:

* `ResolveGlobalVariable.qll` was derived from `ResolveClass.qll`, so it's fairly aggressive with `pragma[noinline]`. I have not checked what happens performance-wise when I remove them. Let me know if it's worth checking this.
* Depends on two internal PRs, so CI will currently fail on the updated test.
* The above template case is covered by `cpp/ql/test/library-tets/templates/variables`

### DIFF
--- a/cpp/ql/lib/semmle/code/cpp/internal/ResolveGlobalVariable.qll
+++ b/cpp/ql/lib/semmle/code/cpp/internal/ResolveGlobalVariable.qll
@@ -1,0 +1,60 @@
+private predicate hasDefinition(@globalvariable g) {
+  exists(@var_decl vd | var_decls(vd, g, _, _, _) | var_def(vd))
+}
+
+pragma[noinline]
+private predicate onlyOneCompleteGlobalVariableExistsWithMangledName(@mangledname name) {
+  strictcount(@globalvariable g | hasDefinition(g) and mangled_name(g, name)) = 1
+}
+
+/** Holds if `g` is a unique global variable with a definition named `name`. */
+pragma[noinline]
+private predicate isGlobalWithMangledNameAndWithDefinition(@mangledname name, @globalvariable g) {
+  hasDefinition(g) and
+  mangled_name(g, name) and
+  onlyOneCompleteGlobalVariableExistsWithMangledName(name)
+}
+
+/** Holds if `g` is a global variable without a definition named `name`. */
+pragma[noinline]
+private predicate isGlobalWithMangledNameAndWithoutDefinition(@mangledname name, @globalvariable g) {
+  not hasDefinition(g) and
+  mangled_name(g, name)
+}
+
+/**
+ * Holds if `incomplete` is a global variable without a definition, and there exists
+ * a unique global variable `complete` with the same name that does have a definition.
+ */
+private predicate hasTwinWithDefinition(@globalvariable incomplete, @globalvariable complete) {
+  exists(@mangledname name |
+    not variable_instantiation(incomplete, complete) and
+    isGlobalWithMangledNameAndWithoutDefinition(name, incomplete) and
+    isGlobalWithMangledNameAndWithDefinition(name, complete)
+  )
+}
+
+import Cached
+
+cached
+private module Cached {
+  /**
+   * If `v` is a global variable without a definition, and there exists a unique
+   * global variable with the same name that does have a definition, then the
+   * result is that unique global variable. Otherwise, the result is `v`.
+   */
+  cached
+  @variable resolveGlobalVariable(@variable v) {
+    hasTwinWithDefinition(v, result)
+    or
+    not hasTwinWithDefinition(v, _) and
+    result = v
+  }
+
+  cached
+  predicate isVariable(@variable v) {
+    not v instanceof @globalvariable
+    or
+    v = resolveGlobalVariable(_)
+  }
+}
--- a/cpp/ql/lib/change-notes/2022-06-24-unique-variable.md
+++ b/cpp/ql/lib/change-notes/2022-06-24-unique-variable.md
@@ -1,0 +1,4 @@
+---
+category: fix
+---
+* Under certain circumstances a variable declaration that is not also a definition could be associated with a `Variable` that did not have the definition as a `VariableDeclarationEntry`. This is now fixed, and a unique `Variable` will exist that has both the declaration and the definition as a `VariableDeclarationEntry`.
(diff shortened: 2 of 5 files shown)


sashabu:
> Note that this also maps the following declarations of `a` onto a single variable of type `int`
> 
> ```
> // a.cpp 
> int a;
> ```
> 
> ```
> // b.cpp
> extern long a;
> ```
> 
> This is consistent with linker behaviour, but might not be desirable from a CodeQl perspective? Note that the there will still be two `VariableDeclarationEntries`: one of type `int` and one of type `long`.

Assuming these are linked together, that's an ODR violation. I don't think we need to go to great lengths to analyse code with ODR violations considering it's UB and we don't even have enough information to *guess* at the symptoms (e.g. if `int` is 16 bits and `long` is 32 bits, we have no way of guessing what other object a write to `extern long a;` might clobber in practice, or indeed whether optimisations accidentally "unclobber" an overlapping object).

If we do want to "handle" such cases, I think the best thing to do would be to detect ODR violations in a dedicated query and assume they're not present in all other queries. Since as you say there are two `VariableDeclarationEntry` values, we have enough information for this.

However what happens if we tweak the example to be valid C++ by giving a.cpp's `a` internal linkage (e.g. `const int a;` or `static int a;`)? I'm hoping name mangling takes care of giving the two objects different names, but it might be good to add a test for this.

MathiasVP:
I think I agree with @sashabu that we shouldn't try to mitigate ODR issues like this on the extractor side. However, I do have a comment with regards to this:

> If we do want to "handle" such cases, I think the best thing to do would be to detect ODR violations in a dedicated query and assume they're not present in all other queries.

ODR violations have given us plenty of performance issues in the past because people didn't know their code contained ODR violations, and it turned some analyses into awful exponential-time algorithms.

Luckily, they're normally quite simple fixes on the QL side (see for example the [old GVN](https://github.com/github/codeql/blob/main/cpp/ql/lib/semmle/code/cpp/valuenumbering/GlobalValueNumberingImpl.qll) library and the [new GVN](https://github.com/github/codeql/blob/main/cpp/ql/lib/semmle/code/cpp/ir/implementation/aliased_ssa/gvn/internal/ValueNumberingInternal.qll) library for some of the mitigation stuff we've done to guard ourselves against stuff like "a variable with multiple types", or "a field lookup returning multiple fields").

So we can't quite ignore ODR issues in the analyses because we've alerted about it in some other query, since the presence of an ODR violation can prevent the suite from completing.

sashabu:
> So we can't quite ignore ODR issues in the analyses because we've alerted about it in some other query, since the presence of an ODR violation can prevent the suite from completing.

@MathiasVP - I think having the two `VariableDeclarationEntry` values is sufficient for this?

You're quite right that my comment was overly general. More precisely, what I was trying to say is that it's good to have a way to _detect_ ODR violations, but we don't necessarily need to try and model the semantics of a C++-like language where ODR violations are well-defined (e.g. by defining differently-typed "overloads" of the variable). I think we're in agreement on that?

MathiasVP:
> You're quite right that my comment was overly general. More precisely, what I was trying to say is that it's good to have a way to detect ODR violations, but we don't necessarily need to try and model the semantics of a C++-like language where ODR violations are well-defined (e.g. by defining differently-typed "overloads" of the variable). I think we're in agreement on that?

Yes, totally in agreement on that 😄.

jketema:
> dedicated query

We have one internally.

jketema:
> However what happens if we tweak the example to be valid C++ by giving `a.cpp`'s `a` internal linkage (e.g. `const int a;` or `static int a;`)?

Global variables with internal linkage have different name mangling. One of the internal fixes for this PR actually corrected some issues there and includes tests.

MathiasVP:
> `ResolveGlobalVariable.qll` was resolved from `ResolveClass.qll`, so it's fairly aggressive with `pragma[noinline]`. I have not checked what happens performance-wise when I remove them. Let me know if it's worth checking this.

I think the aggressive use of `pragma[noinline]` in `ResolveClass.qll` is due to the fact that this file was written back when the optimizer was dumber. So I would guess that many of the `noinline`s could be removed. But that's probably an issue for another PR. If the performance ends up looking good I think it makes sense to just copy the style into this PR.

cpp/ql/lib/semmle/code/cpp/internal/ResolveGlobalVariable.qll
```diff
@@ -0,0 +1,60 @@
+private predicate hasDefinition(@globalvariable g) {
+  exists(@var_decl vd | var_decls(vd, g, _, _, _) | var_def(vd))
+}
+
+pragma[noinline]
+private predicate onlyOneCompleteGlobalVariableExistsWithMangledName(@mangledname name) {
+  strictcount(@globalvariable g | hasDefinition(g) and mangled_name(g, name)) = 1
```
MathiasVP, line 7: I imagine this `strictcount` could be a `unique` (which might perform _slightly_ better).

jketema: Yeah, indeed. Thanks for spotting that. I think the same holds for the one in `ResolveClass.qll` (I'll see if I can update that one in a separate PR).

jketema: I assume you're expecting:
```codeql
  exists(unique(@globalvariable g | hasDefinition(g) and mangled_name(g, name)))
```
?

MathiasVP: Yeah, something like that

jketema: Done. I'll re-do the DCA experiment just out of sheer paranoia.

jketema: Reverted, after internal discussion about `unique` and `strictcount`.

PR #9701, opened 2022-06-24
Swift: make `Element::resolve` final

redsun82:
One should only override `getResolveStep` (or `convertsFrom` for `Expr`
classes), as otherwise the resolution/conversion becomes inconsitent.


MathiasVP (approved):
Thanks!

PR #9702, opened 2022-06-24
Swift: emit intermediate build artifacts for own consumption later

AlexDenisov:
This PR adds significant changes to the way we perform extraction. Namely:
 - we actually tell the extractor not only to typecheck the code and build AST, but also to produce the build artifacts.
   This is required so that the extractor can consume the artifacts produced by an earlier run. That is, to handle cases like this (pseudo-code):
   ```
   extractor -c A.swift -o A.swiftmodule
   extractor -c B.swift -I A.swiftmodule
   ```
   Previously, the extraction of `B.swift` would fail as `A.swiftmodule` either doesn't exist, (or exists but was compiled by the original and incompatible Swift compiler).
- we rewrite all the output CLI args to point to a scratch dir. E.g.:
  `extractor -c A.swift -o A.swiftmodule` ->` extractor -c A.swift -o <codeql>/scratch-dir/A.swiftmodule`
  This is required not to conflict with the original Swift compiler.
- we redirect the extractor module lookup via [`llvm::vfs::RedirectingFileSystem`](https://llvm.org/docs/doxygen/classllvm_1_1vfs_1_1RedirectingFileSystem.html) (see all the VFS related code).
   I tried patching all the `-I` and `-F` flags, but it breaks in many different ways.
- we add a few extra `.swiftmodule` mappings to workaround the way Xcode works. See the comment around `computeModuleAliases`.

With this PR we extract much more code, and we eliminate the majority of the unresolved AST nodes (we are down to ~200 unresolved nodes from ~30k for the Signal-iOS app).


swift/extractor/SwiftOutputRewrite.cpp
```diff
@@ -0,0 +1,318 @@
...
+    }
+  }
+
+  return remapping;
+}
+
+void ensureNewPathsExist(const std::unordered_map<std::string, std::string>& remapping) {
+  for (auto& [_, newPath] : remapping) {
+    llvm::SmallString<PATH_MAX> filepath(newPath);
+    llvm::StringRef parent = llvm::sys::path::parent_path(filepath);
```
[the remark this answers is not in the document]
AlexDenisov: It could be a directory as well, yes.
So far I only observed that output files are not created if the parent directory does not exist. The directories seem to be also created, but that requires more testing on real world projects.

swift/extractor/SwiftOutputRewrite.cpp
```diff
@@ -0,0 +1,318 @@
...
+#include <llvm/Support/FileSystem.h>
+#include <llvm/Support/Path.h>
+#include <swift/Basic/OutputFileMap.h>
+#include <swift/Basic/Platform.h>
+#include <unistd.h>
+#include <unordered_set>
+#include <optional>
+#include <iostream>
+
+// Creates a copy of the output file map and updated remapping table in place
```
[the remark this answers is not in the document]
AlexDenisov: You're reading it right. 😄 
(`updated` should've been `updates`.)

swift/extractor/main.cpp
```diff
@@ -1,27 +1,32 @@
...
  public:
   explicit Observer(const codeql::SwiftExtractorConfiguration& config) : config{config} {}
 
   void parsedArgs(swift::CompilerInvocation& invocation) override {
-    // Original compiler and the extractor-compiler get into conflicts when
-    // both produce the same output files.
-    // TODO: change the final artifact destinations instead of disabling
-    // the artifact generation completely?
-    invocation.getFrontendOptions().RequestedAction = swift::FrontendOptions::ActionType::Typecheck;
+    auto& overlays = invocation.getSearchPathOptions().VFSOverlayFiles;
```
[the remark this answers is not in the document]
AlexDenisov: It can be empty or it may contain something in it already. As far as I can tell, this code adds new file system "views" on top of each other, so there is a risk that we override something here.

This also needs more testing in a real-world setting, but so far I didn't see this option used with several rather big projects.

swift/extractor/SwiftOutputRewrite.cpp
```diff
@@ -0,0 +1,318 @@
...
+      auto newPath = maybeNewPath.value();
+      CLIArgs[index] = newPath;
+      remapping[oldPath] = newPath;
+    }
+  }
+
+  // This doesn't really belong here, but we've got Xcode...
+  for (auto& [oldPath, newPath] : remapping) {
+    llvm::StringRef path(oldPath);
+    auto aliases = computeModuleAliases(path, targetTriple);
```
jketema, line 220: <strike>What if `targetTriple` is the empty string? Does this work then? Or can that never happen?</strike>

<strike>I would also like to understand why it's "safe" to map all the aliases to the same `newPath`.</strike>

This all became very clear once I started to dig into `computeModuleAliases`.

swift/extractor/SwiftOutputRewrite.cpp
```diff
@@ -0,0 +1,318 @@
...
+  // while it is not complete
+  // TODO: Pick a more robust way to not collide with files from other processes
+  auto tempVfsPath = config.tempVFSDir + '/' + std::to_string(getpid()) + "-vfs.yaml";
+  std::error_code ec;
+  llvm::raw_fd_ostream fd(tempVfsPath, ec, llvm::sys::fs::OF_None);
+  if (ec) {
+    std::cerr << "Cannot create temp VFS file: '" << tempVfsPath << "': " << ec.message() << "\n";
+    return;
+  }
+  // TODO: there must be a better API than this
```
jketema, line 273: Somewhere in the closed source XCode codebase?

swift/extractor/SwiftOutputRewrite.cpp
```diff
@@ -0,0 +1,318 @@
...
+  fd.flush();
+  auto vfsPath = config.VFSDir + '/' + std::to_string(getpid()) + "-vfs.yaml";
+  if (std::error_code ec = llvm::sys::fs::rename(tempVfsPath, vfsPath)) {
+    std::cerr << "Cannot move temp VFS file '" << tempVfsPath << "' -> '" << vfsPath
+              << "': " << ec.message() << "\n";
+    return;
+  }
+}
+
+std::vector<std::string> collectVFSFiles(const SwiftExtractorConfiguration& config) {
```
jketema, line 299: I assume this is done separately from `storeRemappingForVFS`, because we might be trying to find a file from a previous extractor invocation? If so, that might be worth a comment.

swift/extractor/SwiftOutputRewrite.cpp
```diff
@@ -0,0 +1,318 @@
...
+              << "': " << ec.message() << "\n";
+    return std::nullopt;
+  }
+
+  llvm::raw_fd_ostream fd(newPath, ec, llvm::sys::fs::OF_None);
+  newOutputMap.write(fd, keys);
+  return newPath;
+}
+
+// This is Xcode-specific workaround to produce alias names for an existing .swiftmodule file.
```
jketema, line 61:
```suggestion
// This is an Xcode-specific workaround to produce alias names for an existing .swiftmodule file.
```

swift/extractor/SwiftOutputRewrite.cpp
```diff
@@ -0,0 +1,318 @@
...
+// In the case of Xcode, it calls the Swift compiler and asks it to produce a Swift module.
+// Once it's done, Xcode moves the .swiftmodule file in another location, and the location is
+// rather arbitrary. Here are examples of such locations:
```
jketema, line 64: 😭

swift/extractor/SwiftOutputRewrite.cpp
```diff
@@ -0,0 +1,318 @@
...
+//  Project: name of a project, target, or scheme
+//  BuildType: Debug, Release, etc.
+//  Target: macOS, iphoneos, appletvsimulator, etc.
+//  Arch: arm64, x86_64, etc.
+//
+// So far we observed that Xcode can move the module into different locations, and it's not
+// entirely clear how to deduce the destination from the context available for the extractor.
+// 1. First case:
+//  DerivedData/<Project>/Build/Products/<BuiltType>-<Target>/<ModuleName>.swiftmodule/<Arch>.swiftmodule
+//  DerivedData/<Project>/Build/Products/<BuiltType>-<Target>/<ModuleName>.swiftmodule/<Triple>.swiftmodule
```
jketema, line 77: How do I read this? The original path on l. 66 can become one of two things in this first case? Either `.../<Arch>.switftmodule` or `.../<Triple>.switftmodule`

swift/extractor/SwiftOutputRewrite.cpp
```diff
@@ -0,0 +1,318 @@
...
+// 2. Second case:
+//  DerivedData/<Project>/Build/Products/<BuiltType>-<Target>/<ModuleName>/<ModuleName>.swiftmodule/<Arch>.swiftmodule
+//  DerivedData/<Project>/Build/Products/<BuiltType>-<Target>/<ModuleName>/<ModuleName>.swiftmodule/<Triple>.swiftmodule
+// 2. Third case:
+//  DerivedData/<Project>/Build/Products/<BuiltType>-<Target>/<ModuleName>/<ModuleName>.framework/Modules/<ModuleName>.swiftmodule/<Arch>.swiftmodule
+//  DerivedData/<Project>/Build/Products/<BuiltType>-<Target>/<ModuleName>/<ModuleName>.framework/Modules/<ModuleName>.swiftmodule/<Triple>.swiftmodule
+// The <Triple> here is a normalized target triple (e.g. arm64-apple-iphoneos15.4 ->
+// arm64-apple-iphoneos).
+//
+// This method construct those aliases for a module only if it comes from Xcode, which is detected
```
jketema, line 87:
```suggestion
// This method constructs those aliases for a module only if it comes from Xcode, which is detected
```

swift/extractor/SwiftOutputRewrite.cpp
```diff
@@ -0,0 +1,318 @@
...
+// by the presence of `Intermediates.noindex` directory in the module path.
```
jketema, line 88:
```suggestion
// by the presence of an `Intermediates.noindex` directory in the module path.
```

swift/extractor/SwiftOutputRewrite.cpp
```diff
@@ -0,0 +1,318 @@
...
+//
+// In the case of Swift Package Manager (`swift build`) this is not needed.
```
jketema, line 90:
```suggestion
// In the case of the Swift Package Manager (`swift build`) this is not needed.
```

jketema (commented):
Some more comments.

On the one hand this is clearly one big hack that we need and for which it's difficult to test all the corner cases. On the other hand this is sort of an elegant solution to the problem we're having here.

swift/extractor/SwiftOutputRewrite.cpp
AlexDenisov: In this case we should actually just pick up whatever default triple for the process, but the triple is always present within Xcode/SPM.

swift/extractor/SwiftOutputRewrite.cpp
AlexDenisov: In general - yes 😆.
But I meant some LLVM's `YAMLTraits` things. It's just felt easier to construct this by hand than to dig through the APIs.

swift/extractor/SwiftOutputRewrite.cpp
AlexDenisov: Yeah, it's a "collective knowledge" here for the processes that run later.

swift/extractor/SwiftOutputRewrite.cpp
AlexDenisov: Originally Xcode asks for `<Arch>/<Name>.swiftmodule` and them moves it into `<Name>.swiftmodule/<Triple>`.
So I guess we can actually ignore the `<Arch>.swiftmodule` case, but it is the format the SPM (Swift Package Manager) uses IIRC.

swift/extractor/SwiftOutputRewrite.cpp
```diff
@@ -0,0 +1,318 @@
...
+      break;
+    }
+  }
+  // Not built by Xcode, skipping
+  if (intermediatesDirIndex == 0) {
+    return {};
+  }
+  // e.g. Debug-iphoneos, Release-iphonesimulator, etc.
+  auto destinationDir = chunks[intermediatesDirIndex + 2].str();
+  auto arch = chunks[intermediatesDirIndex + 5].str();
```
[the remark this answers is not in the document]
AlexDenisov: Yeah, also bounds checking (shouldn't happen, but who knows...).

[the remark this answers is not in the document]
jketema: Good point about the bounds checking, definitely add that!

swift/extractor/SwiftOutputRewrite.cpp
```diff
@@ -0,0 +1,329 @@
...
+      continue;
+    }
+    keys.push_back(key);
+    auto& newMap = newOutputMap.getOrCreateOutputMapForInput(key);
+    newMap.copyFrom(*oldMap);
+    for (auto& entry : newMap) {
+      auto oldPath = entry.getSecond();
+      auto newPath = config.tempArtifactDir + '/' + oldPath;
+      entry.getSecond() = newPath;
+      remapping[oldPath] = newPath;
```
redsun82, line 44: maybe not a big deal for performance, but in general you should use the shiny new C++17 `insert_or_assign`, as it will avoid one default construction and one assignment in case the key is not in the map. And I would move both variables as we don't need it any more, and we then avoid lots of string copies.
```suggestion
      remapping.insert_or_assign(std::move(oldPath), std::move(newPath));
```

swift/extractor/SwiftOutputRewrite.cpp
```diff
@@ -0,0 +1,329 @@
...
+
+// Creates a copy of the output file map and updates remapping table in place
+// It does not change the original map file as it is depended upon by the original compiler
+// Returns path to the newly created output file map on success, or None in a case of failure
+static std::optional<std::string> rewriteOutputFileMap(
+    const codeql::SwiftExtractorConfiguration& config,
+    const std::string& outputFileMapPath,
+    const std::vector<std::string>& inputs,
+    std::unordered_map<std::string, std::string>& remapping) {
+  auto newPath = config.tempArtifactDir + '/' + outputFileMapPath;
```
redsun82, line 22: please rename `newPath` as it is quite confusing because of the inner loop variable of the same name.

swift/extractor/SwiftOutputRewrite.cpp
```diff
@@ -0,0 +1,329 @@
...
+
+  // TODO: do not assume absolute path for the second parameter
+  auto outputMapOrError = swift::OutputFileMap::loadFromPath(outputFileMapPath, "");
+  if (!outputMapOrError) {
+    return std::nullopt;
+  }
+  auto oldOutputMap = outputMapOrError.get();
+  swift::OutputFileMap newOutputMap;
+  std::vector<llvm::StringRef> keys;
```
redsun82, line 31: do we need `keys`? Can't we write out `newOutputMap` on line 57?

andersfugmann:
> we rewrite all the output CLI args to point to a scratch dir. E.g.:
> extractor -c A.swift -o A.swiftmodule -> extractor -c A.swift -o <codeql>/scratch-dir/A.swiftmodule
> This is required not to conflict with the original Swift compiler.

Are there cases where the compiler could be invoked with relative paths, e.g.
`extractor -c A.swift -o ../build_dir/A.swiftmodule`

AlexDenisov:
> > we rewrite all the output CLI args to point to a scratch dir. E.g.:
> > extractor -c A.swift -o A.swiftmodule -> extractor -c A.swift -o /scratch-dir/A.swiftmodule
> > This is required not to conflict with the original Swift compiler.
> 
> Are there cases where the compiler could be invoked with relative paths, e.g. `extractor -c A.swift -o ../build_dir/A.swiftmodule`

Yes, it's of course possible to invoke it like that. But I haven't seen these cases in the wild neither with Xcode nor SPM.

swift/extractor/SwiftOutputRewrite.cpp
```diff
@@ -0,0 +1,329 @@
...
+
+  std::vector<std::string> aliases;
+  for (auto& location : moduleLocations) {
+    aliases.push_back(location + arch + ".swiftmodule");
+    if (!targetTriple.empty()) {
+      llvm::Triple triple(targetTriple);
+      auto moduleTriple = swift::getTargetSpecificModuleTriple(triple);
+      aliases.push_back(location + moduleTriple.normalize() + ".swiftmodule");
+    }
+  }
```
redsun82, line 160: Slightly shorter, probably clearer, involves less characters stored in separate strings as we don't store the comon `relocateModulePath` part, and seems easier to update:
```suggestion
  std::vector<std::string> moduleLocationCases = {
    "",
    moduleName.str() + '/',
    moduleName.str() + '/' + moduleName.str() + ".framework/Modules/",
  };

  std::vector<std::string> aliases;
  for (const auto& dir : moduleLocationCases) {
    auto location = relocateModulePath + '/' + dir + moduleNameWithExt.str() + '/';
    aliases.push_back(location + arch+ ".swiftmodule");
    if (!targetTriple.empty()) {
      llvm::Triple triple(targetTriple);
      auto moduleTriple = swift::getTargetSpecificModuleTriple(triple);
      aliases.push_back(location + moduleTriple.normalize() + ".swiftmodule");
    }
  }
```

swift/extractor/SwiftOutputRewrite.cpp
```diff
@@ -0,0 +1,329 @@
...
+//  DerivedData/<Project>/Build/Products/<BuiltType>-<Target>/<ModuleName>/<ModuleName>.framework/Modules/<ModuleName>.swiftmodule/<Arch>.swiftmodule
+//  DerivedData/<Project>/Build/Products/<BuiltType>-<Target>/<ModuleName>/<ModuleName>.framework/Modules/<ModuleName>.swiftmodule/<Triple>.swiftmodule
+// The <Triple> here is a normalized target triple (e.g. arm64-apple-iphoneos15.4 ->
+// arm64-apple-iphoneos).
+//
+// This method constructs those aliases for a module only if it comes from Xcode, which is detected
+// by the presence of an `Intermediates.noindex` directory in the module path.
+//
+// In the case of the Swift Package Manager (`swift build`) this is not needed.
+static std::vector<std::string> computeModuleAliases(llvm::StringRef modulePath,
```
redsun82, line 91: this function is quite a lot to chew through... I think it's ok for now, but maybe I'd leave a todo, so that by the next time we need to do something here, if we finally have access to `std::filesystem` we should definitely use that instead of all the string operations here.

swift/extractor/SwiftOutputRewrite.cpp
```diff
@@ -0,0 +1,329 @@
...
+
+  for (auto index : outputFileMapIndexes) {
+    auto oldPath = CLIArgs[index];
+    auto maybeNewPath = rewriteOutputFileMap(config, oldPath, maybeInput, remapping);
+    if (maybeNewPath) {
+      auto newPath = maybeNewPath.value();
+      CLIArgs[index] = newPath;
+      remapping[oldPath] = newPath;
+    }
+  }
```
redsun82, line 225: there are unneeded string copies
```suggestion
  for (auto index : outputFileMapIndexes) {
    auto maybeNewPath = rewriteOutputFileMap(config, CLIArgs[index], maybeInput, remapping);
    if (maybeNewPath) {
      oldPath = std::move(CLIArgs[index]);
      CLIArgs[index] = *maybeNewPath;
      remapping.insert_or_assign(std::move(oldPath), std::move(newPath));
    }
  }
```

swift/extractor/SwiftOutputRewrite.cpp
```diff
@@ -0,0 +1,329 @@
...
+
+  // This doesn't really belong here, but we've got Xcode...
+  for (auto& [oldPath, newPath] : remapping) {
+    llvm::StringRef path(oldPath);
+    auto aliases = computeModuleAliases(path, targetTriple);
+    for (auto& alias : aliases) {
+      remapping[alias] = newPath;
```
redsun82, line 232:
```suggestion
      remapping.insert_or_assign(std::move(alias), newPath);
```

swift/extractor/SwiftOutputRewrite.cpp
```diff
@@ -0,0 +1,329 @@
...
+  }
+}
+
+void storeRemappingForVFS(const SwiftExtractorConfiguration& config,
+                          const std::unordered_map<std::string, std::string>& remapping) {
+  // Only create remapping for the .swiftmodule files
+  std::unordered_map<std::string, std::string> modules;
+  for (auto& [oldPath, newPath] : remapping) {
+    if (llvm::StringRef(oldPath).endswith(".swiftmodule")) {
+      modules[oldPath] = newPath;
```
redsun82, line 256:
```suggestion
      modules.emplace(oldPath, newPath);
```

swift/extractor/SwiftOutputRewrite.cpp
```diff
@@ -0,0 +1,329 @@
...
+      std::cerr << "Cannot create redirected directory: " << ec.message() << "\n";
+    }
+  }
+}
+
+void storeRemappingForVFS(const SwiftExtractorConfiguration& config,
+                          const std::unordered_map<std::string, std::string>& remapping) {
+  // Only create remapping for the .swiftmodule files
+  std::unordered_map<std::string, std::string> modules;
+  for (auto& [oldPath, newPath] : remapping) {
```
redsun82, line 254:
```suggestion
  for (const auto& [oldPath, newPath] : remapping) {
```

redsun82 (approved):
Great job, that's a hell of a ride :smile: 

There is nothing fundamentally wrong with the solution, at least to the best of our current knowledge. I've left lots of comments that I think might be worth addressing, but if there is pressure to merge this ASAP I'm ok with this going in as is and with the comments addressed in a separate PR

swift/extractor/main.cpp
```diff
@@ -54,12 +59,26 @@ int main(int argc, char** argv) {
   configuration.scratchDir = getenv_or("CODEQL_EXTRACTOR_SWIFT_SCRATCH_DIR", ".");
 
   configuration.tempTrapDir = configuration.scratchDir + "/swift-trap-temp";
+  configuration.VFSDir = configuration.scratchDir + "/swift-vfs";
+  configuration.tempVFSDir = configuration.scratchDir + "/swift-vfs-temp";
+  configuration.tempArtifactDir = configuration.scratchDir + "/swift-extraction-artifacts";
```
redsun82, line 64: the scratch dir is already swift-extractor specific, so I don't think we need the `swift-` prefixes here.

swift/extractor/SwiftExtractorConfiguration.h
```diff
@@ -16,7 +16,22 @@ struct SwiftExtractorConfiguration {
...
+  // Subdirectory of the scratchDir.
+  std::string VFSDir;
+  // A temporary directory that contains temp VFS files before they moved into VFSDir.
+  // Subdirectory of the scratchDir.
+  std::string tempVFSDir;
+
+  // A temporary directory that contains build artifacts generated by the extractor during the
+  // overall extraction process.
+  // Subdirectory of the scratchDir.
+  std::string tempArtifactDir;
```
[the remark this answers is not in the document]
AlexDenisov: Oh, that looks much better!

swift/extractor/SwiftOutputRewrite.cpp
```diff
@@ -0,0 +1,329 @@
...
+// It does not change the original map file as it is depended upon by the original compiler
+// Returns path to the newly created output file map on success, or None in a case of failure
+static std::optional<std::string> rewriteOutputFileMap(
+    const codeql::SwiftExtractorConfiguration& config,
+    const std::string& outputFileMapPath,
+    const std::vector<std::string>& inputs,
+    std::unordered_map<std::string, std::string>& remapping) {
+  auto newPath = config.tempArtifactDir + '/' + outputFileMapPath;
+
+  // TODO: do not assume absolute path for the second parameter
```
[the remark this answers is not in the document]
AlexDenisov: That's actually about the second argument (which is an empty string now)

swift/extractor/SwiftOutputRewrite.cpp
AlexDenisov: The `OutputFileMap`'s API requires it 🤷

swift/extractor/SwiftOutputRewrite.cpp
AlexDenisov: I agree it's shorter and more readable, but I'd keep it as is as I think it's too early to generalize. I.e., we may find a few other cases where this approach won't work.

swift/extractor/SwiftOutputRewrite.cpp
AlexDenisov: There is an ongoing pattern in this file with `remapping[oldPath] = newPath`. I'd like to stick to it (even in cases where it's not the most efficient way) just for the sake of consistency.

swift/extractor/main.cpp
redsun82: ah, I thought `CODEQL_EXTRACTOR_SWIFT_SCRATCH_DIR` meant it was extractor specific? But ok.

swift/extractor/SwiftOutputRewrite.cpp
redsun82: well, I would consistently use `insert_or_assign`, which I suggested everywhere :smile:

swift/extractor/SwiftOutputRewrite.cpp
redsun82: ok, but for readability I would still wrap it in a single assignment rather that assign, `+=`, `push_back` sequences
```suggestion
  std::vector<std::string> moduleLocations = {
    relocatedModulePath + moduleNameWithExt.str() + '/',
    relocatedModulePath + moduleName.str() + '/' + moduleNameWithExt.str() + '/',  
    relocatedModulePath + moduleName.str() + '/' + moduleName.str() + ".framework/Modules/" + moduleNameWithExt.str() + '/',
  }

  std::vector<std::string> aliases;
  for (auto& location : moduleLocations) {
    aliases.push_back(location + arch + ".swiftmodule");
    if (!targetTriple.empty()) {
      llvm::Triple triple(targetTriple);
      auto moduleTriple = swift::getTargetSpecificModuleTriple(triple);
      aliases.push_back(location + moduleTriple.normalize() + ".swiftmodule");
    }
  }
```

swift/extractor/SwiftOutputRewrite.cpp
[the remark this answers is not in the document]
redsun82: ah, so is the TODO resolved, or to be reworded?

swift/extractor/SwiftOutputRewrite.cpp
redsun82: (but not a blocker for me, just to be clear)

redsun82 (approved):
:ship:

swift/extractor/SwiftOutputRewrite.cpp
```diff
@@ -0,0 +1,329 @@
...
+      remapping[oldPath] = newPath;
+    }
+  }
+  std::error_code ec;
+  llvm::SmallString<PATH_MAX> filepath(newPath);
+  llvm::StringRef parent = llvm::sys::path::parent_path(filepath);
+  if (std::error_code ec = llvm::sys::fs::create_directories(parent)) {
+    std::cerr << "Cannot create relocated output map dir: '" << parent.str()
+              << "': " << ec.message() << "\n";
+    return std::nullopt;
```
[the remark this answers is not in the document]
AlexDenisov: We can handle this with the upcoming `diagnostics` class 😅

PR #9703, opened 2022-06-24
Kotlin: Record more kinds of elements as compiler-generated

igfoo:


java/kotlin-extractor/src/main/kotlin/KotlinFileExtractor.kt
```diff
@@ -595,6 +595,8 @@ open class KotlinFileExtractor(
                 tw.writeMethods(clinitId, "<clinit>", "<clinit>()", returnType.javaResult.id, parentId, clinitId)
                 tw.writeMethodsKotlinType(clinitId, returnType.kotlinResult.id)
 
+                tw.writeCompiler_generated(clinitId, 4)
```
smowton, line 598: Define constants / an enum for these integers?

igfoo: Done. I've put it in `KotlinFileExtractor` for now, but we might want to move it out later if we have generated things elsewhere.

smowton (approved):

PR #9705, opened 2022-06-24
C#: Implement correct behavior for `dotnet build` tracing

criemen:
For proper C# tracing, `dotnet build` needs the parameter
`/p:UseSharedCompilation=false`. However, we can't pass that to the other
subcommands of `dotnet`, therefore we need to figure out which subcommand
of `dotnet` is being invoked.

### DIFF
--- a/csharp/tools/tracing-config.lua
+++ b/csharp/tools/tracing-config.lua
@@ -2,7 +2,54 @@ function RegisterExtractorPack(id)
     local extractor = GetPlatformToolsDirectory() ..
                           'Semmle.Extraction.CSharp.Driver'
     if OperatingSystem == 'windows' then extractor = extractor .. '.exe' end
+
+    function DotnetMatcherBuild(compilerName, compilerPath, compilerArguments,
+                                _languageId)
+        if compilerName ~= 'dotnet' and compilerName ~= 'dotnet.exe' then
+            return nil
+        end
+
+        -- The dotnet CLI has the following usage instructions:
+        -- dotnet [sdk-options] [command] [command-options] [arguments]
+        -- we are interested in dotnet build, which has the following usage instructions:
+        -- dotnet [options] build [<PROJECT | SOLUTION>...]
+        -- For now, parse the command line as follows:
+        -- Everything that starts with `-` (or `/`) will be ignored.
+        -- The first non-option argument is treated as the command.
+        -- if that's `build`, we append `/p:UseSharedCompilation=false` to the command line,
+        -- otherwise we do nothing.
+        local match = false
+        local argv = compilerArguments.argv
+        if OperatingSystem == 'windows' then
+            -- let's hope that this split matches the escaping rules `dotnet` applies to command line arguments
+            -- or, at least, that it is close enough
+            argv =
+                NativeArgumentsToArgv(compilerArguments.nativeArgumentPointer)
+        end
+        for i, arg in ipairs(argv) do
+            -- dotnet options start with either - or / (both are legal)
+            local firstCharacter = string.sub(arg, 1, 1)
+            if not (firstCharacter == '-') and not (firstCharacter == '/') then
+                Log(1, 'Dotnet subcommand detected: %s', arg)
+                if arg == 'build' then match = true end
+                break
+            end
+        end
+        if match then
+            return {
+                order = ORDER_REPLACE,
+                invocation = BuildExtractorInvocation(id, compilerPath,
+                                                      compilerPath,
+                                                      compilerArguments, nil, {
+                    '/p:UseSharedCompilation=false'
+                })
+            }
+        end
+        return nil
+    end
+
     local windowsMatchers = {
+        DotnetMatcherBuild,
         CreatePatternMatcher({'^dotnet%.exe$'}, MatchCompilerName, extractor, {
             prepend = {'--dotnetexec', '--cil'},
             order = ORDER_BEFORE
@@ -10,21 +57,20 @@ function RegisterExtractorPack(id)
         CreatePatternMatcher({'^csc.*%.exe$'}, MatchCompilerName, extractor, {
             prepend = {'--compiler', '"${compiler}"', '--cil'},
             order = ORDER_BEFORE
-
         }),
         CreatePatternMatcher({'^fakes.*%.exe$', 'moles.*%.exe'},
                              MatchCompilerName, nil, {trace = false})
     }
     local posixMatchers = {
-        CreatePatternMatcher({'^mcs%.exe$', '^csc%.exe$'}, MatchCompilerName,
-                             extractor, {
-            prepend = {'--compiler', '"${compiler}"', '--cil'},
-            order = ORDER_BEFORE
-
-        }),
+        DotnetMatcherBuild,
         CreatePatternMatcher({'^mono', '^dotnet$'}, MatchCompilerName,
                              extractor, {
             prepend = {'--dotnetexec', '--cil'},
+            order = ORDER_BEFORE
+        }),
+        CreatePatternMatcher({'^mcs%.exe$', '^csc%.exe$'}, MatchCompilerName,
+                             extractor, {
+            prepend = {'--compiler', '"${compiler}"', '--cil'},
             order = ORDER_BEFORE
         }), function(compilerName, compilerPath, compilerArguments, _languageId)
             if MatchCompilerName('^msbuild$', compilerName, compilerPath,
@@ -49,7 +95,6 @@ function RegisterExtractorPack(id)
     else
         return posixMatchers
     end
-
 end
 
 -- Return a list of minimum supported versions of the configuration file format


csharp/tools/tracing-config.lua
```diff
@@ -2,29 +2,76 @@ function RegisterExtractorPack(id)
...
+        -- The dotnet CLI has the following usage instructions:
+        -- dotnet [sdk-options] [command] [command-options] [arguments]
+        -- we are interested in dotnet build, which has the following usage instructions:
+        -- dotnet [options] build [<PROJECT | SOLUTION>...]
+        -- however, `dotnet -h build`, although documented, does not work
+        -- For now, parse the command line as follows:
+        -- Everything that starts with `-` will be ignored.
+        -- The first non-option argument is treated as the command.
+        -- if that's `build`, we append `/p:UseSharedCompilation=false` to the command line,
+        -- otherwise we do nothing.
```
tamasvajk, line 21: Passing `-d` before the `build` command works. But I think ignoring everything that starts with`-` is correct and will handle this case too.

csharp/tools/tracing-config.lua
```diff
@@ -2,29 +2,76 @@ function RegisterExtractorPack(id)
...
+            return nil
+        end
+
+        -- The dotnet CLI has the following usage instructions:
+        -- dotnet [sdk-options] [command] [command-options] [arguments]
+        -- we are interested in dotnet build, which has the following usage instructions:
+        -- dotnet [options] build [<PROJECT | SOLUTION>...]
+        -- however, `dotnet -h build`, although documented, does not work
+        -- For now, parse the command line as follows:
+        -- Everything that starts with `-` will be ignored.
```
tamasvajk, line 18: This line could also state that anything prefixed `/` will be ignored.

tamasvajk (approved):
Looks plausible to me.

csharp/tools/tracing-config.lua
```diff
@@ -2,29 +2,75 @@ function RegisterExtractorPack(id)
...
+                break
+            end
+        end
+        if match then
+            return {
+                order = ORDER_REPLACE,
+                invocation = BuildExtractorInvocation(id, compilerPath,
+                                                      compilerPath,
+                                                      compilerArguments, nil, {
+                    '/p:UseSharedCompilation=false'
```
michaelnebel, line 44: Do you know, if this will cause any problems if '/p:UseSharedCompilation=false' is already set as a part of the command line arguments?

michaelnebel (approved):
Looks plausible to me!

csharp/tools/tracing-config.lua
criemen: I believe the last specified option wins, and I just tried this and it worked:
```
 ✘ criemen@Corneliuss-MacBook-Pro  ~/test  dotnet new console
The template "Console App" was created successfully.

Processing post-creation actions...
Running 'dotnet restore' on /Users/criemen/test/test.csproj...
  Determining projects to restore...
  Restored /Users/criemen/test/test.csproj (in 120 ms).
Restore succeeded.

 criemen@Corneliuss-MacBook-Pro  ~/test  dotnet clean
Microsoft (R) Build Engine version 17.1.1+a02f73656 for .NET
Copyright (C) Microsoft Corporation. All rights reserved.

Build started 07/21/2022 11:00:30.
     1>Project "/Users/criemen/test/test.csproj" on node 1 (Clean target(s)).
     1>CoreClean:
         Creating directory "obj/Debug/net6.0/".
     1>Done Building Project "/Users/criemen/test/test.csproj" (Clean target(s)).

Build succeeded.
    0 Warning(s)
    0 Error(s)

Time Elapsed 00:00:00.79
 criemen@Corneliuss-MacBook-Pro  ~/test  dotnet build /p:UseSharedCompilation=false /p:UseSharedCompilation=false
Microsoft (R) Build Engine version 17.1.1+a02f73656 for .NET
Copyright (C) Microsoft Corporation. All rights reserved.

  Determining projects to restore...
  All projects are up-to-date for restore.
  test -> /Users/criemen/test/bin/Debug/net6.0/test.dll

Build succeeded.
    0 Warning(s)
    0 Error(s)

Time Elapsed 00:00:03.00
 criemen@Corneliuss-MacBook-Pro  ~/test 
```

michaelnebel: Thank you! 😄 
Great work!

hvitved:
Thanks a lot for doing all the work that lead to this, @criemen .

jaredpar:
Why is this being done? Using `UseSharedCompilation=false` will **significantly** slow down builds. On large projects this can cause a 2-5X increase in build time. Pushing this on every CodeQL consumer is likely causing a substantial increase in build cost. This is further compounded by the [CodeQL documentation](https://docs.github.com/en/code-security/code-scanning/automatically-scanning-your-code-for-vulnerabilities-and-errors/troubleshooting-the-codeql-workflow#no-code-found-during-the-build) recommending customers set this property without providing context to the downsides of doing this. 

Further the manner in which this is done is causing downstream implications to other teams. This is being added as a global property which means it filters down to tasks, targets and sub-builds which do not anticipate seeing this. It's presently causing a number of internal teams errors because it's filtering into commands that do not support it. 

Can we get some explanation of why this was done? Did you reach out to the compiler team to get a more supported woay of working around whatever problem you were hitting?

tamasvajk:
@jaredpar CodeQL is intercepting calls to `csc` to record the arguments that are used for the call. These arguments are then passed to our extractor to create a `CSharpCompilation`. If `UseSharedCompilation=false` is not added to the build command and the compilation server instance is already started, then we’re not able to intercept the `csc` calls. (If the server is not yet started, then we might end up intercepting more `csc` calls than we should have.)

We used to have a .NET profiler based solution, but that’s no longer available. Our old build tracer wasn’t compatible with newer dotnet versions and the new tracer gives us greater flexibility over the command line.

You’re raising a valid point regarding performance. We’re aware of this issue, and investigating other approaches that could potentially improve the analysis times. We’ve created an internal issue to improve the documentation around `UseSharedCompilation=false`. Note that this PR ensures that  `UseSharedCompilation=false` is always injected into builds, prior to this change, the CodeQL documentation has always recommended this for C# projects using `dotnet build` or `msbuild` for the reasons above.

Could you give more detail on failures that are caused by `UseSharedCompilation=false`? We’ve recently received feedback on other channels regarding failures with `dotnet test`. This issue is being fixed in https://github.com/github/codeql/pull/13387. Are there more use-cases that we should look into?

baronfel:
Consider attaching a logger to the build instead to gather the CommandLineArgs passed to the Csc task. I would suggest writing and attach your own [MSBuild Logger](https://learn.microsoft.com/en-us/visualstudio/msbuild/build-loggers?view=vs-2022) (adding logger CLI arguments automatically in the same way you've added SharedCompilation arguments).  Your logger could simply listen to `TaskStarted`/`TaskFinished` events and write those out in whatever format/location you need without otherwise negatively impacting the users chosen build. Other teams in Microsoft (e.g. AzDo CI system task authors) have done this to seamlessly gather build telemetry again without impacting the users builds.

jaredpar:
> These arguments are then passed to our extractor to create a CSharpCompilation

Curious: how do you all create and use this? Just based on the code here it would seemingly need to be on the same machine as the build. In that case, given you're running an analyisis engine, why are you not just writing a Roslyn based analyzer? There are always risks and issues that come with recreating a `Compilation` after the fact so curious why you chose the current approach over just hooking into build where most of the heavy lifting is done for you. 

As @baronfel mentioned command line arguments are easy to pull from a build. There are [several][https://github.com/jaredpar/basic-compilerlog/blob/main/src/Basic.CompilerLog.Util/BinaryLogUtil.cs#L41](url) [different](https://github.com/KirillOsenkov/MSBuildStructuredLog/blob/ef41acc080228376cec614abbb27f80207d378e0/src/StructuredLogger/Analyzers/CompilerInvocations.cs#L16) ways to get command line arguments out of say a binary log.

> Could you give more detail on failures that are caused by UseSharedCompilation=false? We’ve recently received feedback on other channels regarding failures with dotnet test. 

This is source of problems I'm seeing right now.

tamasvajk:
Thank you for the pointers towards MsBuild loggers. I created an internal issue to explore this approach.

> Curious: how do you all create and use this? Just based on the code here it would seemingly need to be on the same machine as the build. In that case, given you're running an analyisis engine, why are you not just writing a Roslyn based analyzer?

Yes, the build and the analysis (extraction) need to happen on the same machine. I think the current approach was chosen based on its ease of maintenance, simplicity and robustness. We need to support most (all) versions of `csc`, which can be achieved by shipping a single variant of the extractor that depends on the latest version of Roslyn.

jaredpar:
Is the code where you do the analysis available anywhere? Interested in looking how you're creating the `Compilation` instances.

tamasvajk:
You can find the main entry point [here](https://github.com/github/codeql/blob/main/csharp/extractor/Semmle.Extraction.CSharp.Driver/Driver.cs#L12) and the compilation is created [here](https://github.com/github/codeql/blob/main/csharp/extractor/Semmle.Extraction.CSharp/Extractor/Extractor.cs#L408-L416).

If you open [this folder](https://github.com/github/codeql/tree/main/csharp) in VS Code, you can use the below `launch.json` to debug the application:
```
{
    "version": "0.2.0",
    "configurations": [
        {
            "name": ".NET Core Launch (console)",
            "type": "coreclr",
            "request": "launch",
            "preLaunchTask": "dotnet build",
            "program": "${workspaceFolder}/extractor/Semmle.Extraction.CSharp.Driver/bin/Debug/net7.0/Semmle.Extraction.CSharp.Driver.dll",
            "args": [
                "-target:library",
                "/noconfig",
                "/r:System.Private.CoreLib.dll",
                "/r:System.dll",
                "/r:System.Core.dll",
                "/r:System.Runtime.dll",
                "/r:System.Console.dll",
                "/r:System.Linq.dll",
                "${workspaceFolder}/ql/test/library-tests/csharp7/CSharp7.cs"
            ],
            "cwd": "${workspaceFolder}/ql/test/library-tests/csharp7/",
            "console": "internalConsole",
            "stopAtEntry": true
        }
    ]
}
```

PR #9716, opened 2022-06-27
JS: sanitize non-strings from unsafe-html-construction

erik-krogh:


javascript/ql/lib/semmle/javascript/dataflow/TaintTracking.qll
```diff
@@ -1126,6 +1126,21 @@ module TaintTracking {
     )
   }
 
+  /**
+   * Holds if `test` is of the form `typeof x === "something"`, preventing `x` from being a string in some cases.
```
esbena, line 1130: > ... in some cases

Are you hedging against the special case of `typeof new String("foo")`?

esbena, line 1130:
```suggestion
   * Holds if `test` is of the form `typeof x === "string"`, preventing `x` from being a string in some cases.
```

javascript/ql/lib/semmle/javascript/security/dataflow/UnsafeHtmlConstructionCustomizations.qll
```diff
@@ -180,4 +180,18 @@ module UnsafeHtmlConstruction {
 
     override string describe() { result = "Markdown rendering" }
   }
+
+  /** A test of form `typeof x === "something"`, preventing `x` from being a string in some cases. */
```
esbena, line 184:
```suggestion
  /** A test of form `typeof x === "type"`, preventing `x` from being a value with type `type` in some cases. */
```

javascript/ql/lib/semmle/javascript/security/dataflow/XssThroughDomQuery.qll
```diff
@@ -52,27 +51,13 @@ class Configuration extends TaintTracking::Configuration {
   }
 }
 
-/**
- * A test of form `typeof x === "something"`, preventing `x` from being a string in some cases.
- *
- * This sanitizer helps prune infeasible paths in type-overloaded functions.
- */
+/** A test of form `typeof x === "something"`, preventing `x` from being a string in some cases. */
```
esbena, line 54:
```suggestion
/** A test of form `typeof x === "string"`, preventing `x` from being a string in some cases. */
```

esbena:
My only issue with this is the (repeated) docstring. I only now realise that my suggestions should be loosened further to handle the case with other comparison operators. Should they perhaps be closer to:

```
A test for the value of `typeof x`, restricting the potential types of `x`.
```

javascript/ql/lib/semmle/javascript/dataflow/TaintTracking.qll
erik-krogh: I just copy-pasted the existing QLDoc from the old sanitizer-guard that the implementation came from.  

I think the "in some cases" can just be removed.

esbena (approved):

javascript/ql/lib/semmle/javascript/dataflow/TaintTracking.qll
asgerf: I think "in some cases" referred to the fact that it depends on the value of `"something"` and the branch taken.

PR #9717, opened 2022-06-27
Python: Fix bad join in MRO

tausbn:
Fixes a bad join in `list_of_linearization_of_bases_plus_bases`.

Previvously, we joined together `ConsList` and `getBase` before filtering
these out using the recursive call. Now we do the recursion first.


yoff (approved):
DCA experiment looks non-worrisome, so I am happy to merge this.

PR #9720, opened 2022-06-27
C++: Support link targets for global and namespace variables

jketema:
This can be safely merged before the internal PR. The new predicates will just be no-ops as long as the internal PR isn't merged. DCA ran on internal PR.


MathiasVP (approved):
LGTM!

PR #9729, opened 2022-06-28
Swift: extract `ProtocolCompositionType` and all `BuiltinType`s

redsun82:
There are two things of noteregarding `BuiltinType` testing:
* a new mechanism to collapse hierarchies into one single test is introduced with pragmas `qltest_collapse_hierarchy` and `qltest_uncollapse_hierarchy` to undo it on a subtree. This is used to have less tests for the builtin types with the same coverage.
* the `Builtin` module is not available in swift unless you pass `-parse-stdlib` to the frontend, so this PR introduces a possibility to specify additional extractor flags via a line comment, as we are used to do internally for the C++ extractor, namely `//codeql-extractor-options: -parse-stdlib`. This should open up new possibility to test things without necessarily using integration tests

The above two points are introduced by the first two commits, while the last commits contains the extractor code and the tests.


AlexDenisov (approved):
:shipit:

PR #9732, opened 2022-06-28
Swift: add DotSyntaxCallExpr tests

redsun82:


swift/ql/test/extractor-tests/generated/expr/DotSyntaxCallExpr/DotSyntaxCallExpr_getArgument.expected
```diff
@@ -0,0 +1,2 @@
+| dot_syntax_call.swift:6:1:6:3 | call to foo(_:_:) | 0 | : X.Type |
+| dot_syntax_call.swift:7:1:7:3 | call to bar() | 0 | : X.Type |
```
geoffw0, line 2: Why do we only see argument `0` for the `DotSyntaxCallExpr` on line 6?

redsun82: just like with instance method calls, it would seem there is some currying going on.

So `X.foo(1, 2)` gets parsed by the swift frontend library as 

```
CallExpr(
  DotSyntaxCallExpr(
    DeclRefExpr(foo), 
    [TypeExpr(X)]
  ),
  [1, 2]
)
``` 

There is an ongoing discussion to make this more user friendly using an IPA layer on top of the generated DB types ([internal link here](https://github.com/github/codeql-c-team/issues/1019))

geoffw0: Makes sense.  It would be good to make things more user friendly, though I worry there could be edge cases that are difficult to model another way.

geoffw0 (approved):
LGTM :+1:

PR #9734, opened 2022-06-28
C++: Use `exists(unique(...))` instead of `strictcount(...) = 1`

jketema:
As discussed on https://github.com/github/codeql/pull/9700


MathiasVP (approved):
LGTM if DCA is happy :shipit:.

jketema:
Closing. After internal discussion it seems that `strictcount` is the way to go.

PR #9736, opened 2022-06-28
Swift: fix QL warnings about overriding methods

redsun82:
The `getName` in `Type.qll` was issuing a warning in other generated
classes having a `getName` from a `name` property in `schema.yml`.

To fix the possible inconsistency, `diagnostic_name` is being renamed to
`name` in the schema, and more derived `name` properties are being
removed. Despite the scary doc comment on `swift::Type::getString` 
(namely `for use in diagnostics only`), that seems to be the right generic
naming mechanism for types, and it coincides with the name we were
extracting on types with an explicit `name` property.

In case we find a case where `Type::getString` gives something wrong,
we can probably just patch it on that specific type class.


geoffw0:
> My main concern with the "diagnostics" name is that I have a feeling the returned string will eventually be changed in the compiler without any notice (since it's only supposed to be used for debugging). And this will then break our queries that rely on the output.

We should have pretty good test coverage for `toString()` and `getName()` (which I assume depend on this) - if such a change did occur we will at least notice pretty quickly.

MathiasVP:
> We should have pretty good test coverage for toString() and getName() (which I assume depend on this) - if such a change did occur we will at least notice pretty quickly.

That's true. I'm also concerned about future code that lives outside our directory, though.

redsun82:
I think that is a valid concern, but between putting in the effort now to have a solid hand-rolled type naming and postponing the same effort to when and if the compiler pulls out the rug on us on this, I prefer to postpone :) Also, apart from a couple of exceptions, we were effectively already making `Type::getType` use the diagnostics string.

We do need to take care about future client code outside our control, but for that I think we will just need to make the `getName` stable even after (if) the compiler changes `swift::Type::getString`, by putting in the above mentioned effort.

MathiasVP (approved):
LGTM!

PR #9737, opened 2022-06-28
Swift: Add ArithmeticOperation.qll library

geoffw0:
Add an `ArithmeticOperation.qll` library, and make various related changes:
 - add tests for `ArithmeticOperation.qll` and `LogicalOperation.qll`.
 - fix `LogicalOperation.qll` to account for function call `getName` now including the argument pattern.
   - we'd have caught this sooner if there had been a test.
   - also simplified the QL in that library slightly.
 - change `ApplyExpr` to produce a slightly better `toString` in some cases.  I'd really like someone to review this change as I'm not sure how general what I've written is.

I'd also like to know if we're happy with the naming of `BinaryExpr`, `BinaryArithmeticOperation`, `AddExpr` etc, before this stuff becomes more established and difficult to change.

### DIFF
--- a/swift/ql/lib/codeql/swift/elements/expr/ArithmeticOperation.qll
+++ b/swift/ql/lib/codeql/swift/elements/expr/ArithmeticOperation.qll
@@ -15,6 +15,9 @@ class ArithmeticOperation extends Expr {
     this instanceof UnaryArithmeticOperation
   }
 
+  /**
+   * Gets an operand of this arithmetic operation.
+   */
   Expr getAnOperand() {
     result = this.(BinaryArithmeticOperation).getAnOperand()
     or
--- a/swift/ql/lib/codeql/swift/elements/expr/ArithmeticOperation.qll
+++ b/swift/ql/lib/codeql/swift/elements/expr/ArithmeticOperation.qll
@@ -31,7 +31,15 @@ class ArithmeticOperation extends Expr {
  * a + b
  * ```
  */
-abstract class BinaryArithmeticOperation extends BinaryExpr { }
+class BinaryArithmeticOperation extends BinaryExpr {
+  BinaryArithmeticOperation() {
+    this instanceof AddExpr or
+    this instanceof SubExpr or
+    this instanceof MulExpr or
+    this instanceof DivExpr or
+    this instanceof RemExpr
+  }
+}
 
 /**
  * An add expression.
@@ -39,7 +47,7 @@ abstract class BinaryArithmeticOperation extends BinaryExpr { }
  * a + b
  * ```
  */
-class AddExpr extends BinaryArithmeticOperation {
+class AddExpr extends BinaryExpr {
   AddExpr() { this.getFunction().(DotSyntaxCallExpr).getStaticTarget().getName() = "+(_:_:)" }
 }
 
@@ -49,7 +57,7 @@ class AddExpr extends BinaryArithmeticOperation {
  * a - b
  * ```
  */
-class SubExpr extends BinaryArithmeticOperation {
+class SubExpr extends BinaryExpr {
   SubExpr() { this.getFunction().(DotSyntaxCallExpr).getStaticTarget().getName() = "-(_:_:)" }
 }
 
@@ -59,7 +67,7 @@ class SubExpr extends BinaryArithmeticOperation {
  * a * b
  * ```
  */
-class MulExpr extends BinaryArithmeticOperation {
+class MulExpr extends BinaryExpr {
   MulExpr() { this.getFunction().(DotSyntaxCallExpr).getStaticTarget().getName() = "*(_:_:)" }
 }
 
@@ -69,7 +77,7 @@ class MulExpr extends BinaryArithmeticOperation {
  * a / b
  * ```
  */
-class DivExpr extends BinaryArithmeticOperation {
+class DivExpr extends BinaryExpr {
   DivExpr() { this.getFunction().(DotSyntaxCallExpr).getStaticTarget().getName() = "/(_:_:)" }
 }
 
@@ -79,7 +87,7 @@ class DivExpr extends BinaryArithmeticOperation {
  * a % b
  * ```
  */
-class RemExpr extends BinaryArithmeticOperation {
+class RemExpr extends BinaryExpr {
   RemExpr() { this.getFunction().(DotSyntaxCallExpr).getStaticTarget().getName() = "%(_:_:)" }
 }
 
@@ -89,7 +97,9 @@ class RemExpr extends BinaryArithmeticOperation {
  * -a
  * ```
  */
-abstract class UnaryArithmeticOperation extends PrefixUnaryExpr { }
+class UnaryArithmeticOperation extends PrefixUnaryExpr {
+  UnaryArithmeticOperation() { this instanceof UnaryMinusExpr }
+}
 
 /**
  * A unary minus expression.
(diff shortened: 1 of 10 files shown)


swift/ql/lib/codeql/swift/elements/expr/ApplyExpr.qll
```diff
@@ -9,6 +9,10 @@ class ApplyExpr extends ApplyExprBase {
     result = "call to " + this.getStaticTarget().toString()
     or
     not exists(this.getStaticTarget()) and
+    result = "call to " + this.getFunction().toString()
+    or
+    not exists(this.getStaticTarget()) and
+    not exists(this.getFunction()) and
```
MathiasVP, line 15: If there is no function then there's definitely no static call target. So the `not exists(this.getStaticTarget())` conjunction seems unnecessary.

Really, there shouldn't be any case where where `getFunction` doesn't have a result.

Also, I'm not really a big fan of the strings produced in the example below. The string "call to call to -(_:)" doesn't look like something I'd like to see in an alert message (which is the main point of `toString` in my opinion)

swift/ql/lib/codeql/swift/elements/expr/LogicalOperation.qll
```diff
@@ -6,31 +6,19 @@ private import codeql.swift.elements.expr.DeclRefExpr
...
-      .(DeclRefExpr)
-      .getDecl()
-      .(ConcreteFuncDecl)
-      .getName() = name
+  e.getFunction().(DotSyntaxCallExpr).getStaticTarget().getName() = name
 }
 
 class LogicalAndExpr extends BinaryExpr {
-  LogicalAndExpr() { binaryHasName(this, "&&") }
+  LogicalAndExpr() { binaryHasName(this, "&&(_:_:)") }
```
MathiasVP, line 17: Thanks for fixing these! I'm surprised it wasn't broken in the CFG tests after I changed the names in https://github.com/github/codeql/pull/9619 🤔.

swift/ql/lib/codeql/swift/elements/expr/ArithmeticOperation.qll
```diff
@@ -0,0 +1,99 @@
...
+  }
+}
+
+/**
+ * A binary arithmetic operation, such as:
+ * ```
+ * a + b
+ * ```
+ */
+abstract class BinaryArithmeticOperation extends BinaryExpr { }
```
MathiasVP, line 34: We shouldn't create public abstract AST classes as this creates a dangerous footgun for QL writers. If someone were to create a class that extends `BinaryArithmeticOperation` in a query it'll cause everything that depends on the AST layer to be recomputed.

We either need to make this class `private` (which doesn't seem to be the right thing to do), or list its subclasses in the charpred.

swift/ql/lib/codeql/swift/elements/expr/ApplyExpr.qll
geoffw0: Hmm, I hadn't noticed the repetition of `call to` actually.  I'll see if I can do better.

I agree about `getStaticTarget`.

swift/ql/lib/codeql/swift/elements/expr/ArithmeticOperation.qll
geoffw0: Yes, I noticed you'd done something different in `LogicalOperation.qll` and I was meaning to bring up the tradeoff... I don't *like* the way `LogicalAndExpr` doesn't extend `BinaryLogicalOperation`.  I had thought this a problem, but the more I think about it the more I'm not sure it actually is.

I guess I'm talking myself around to adopting your design...

geoffw0: Done.

swift/ql/lib/codeql/swift/elements/expr/ApplyExpr.qll
geoffw0: Done the latter, still looking at the former.

swift/ql/lib/codeql/swift/elements/expr/ApplyExpr.qll
```diff
@@ -9,6 +9,9 @@ class ApplyExpr extends ApplyExprBase {
     result = "call to " + this.getStaticTarget().toString()
     or
     not exists(this.getStaticTarget()) and
+    result = this.getFunction().toString()
```
MathiasVP, line 12: I like the output a lot more now, but I'm not sure we're getting the improved output for the right reasons. It looks like we're now generating `"call to f()"` when we can resolve the call to a static call target `f`, but we print just `"f()"` when we cannot resolve it to a target.

The reason the output looks a lot better now is that the cases where there isn't a static call target are all examples of the form where there's a strange `DotSyntaxCallExpr` (or `SelfApplyExpr`, or a `ConstructorRefCallExpr`) which all have strings that look like `"call to f()"` (because they fall into the category of calls where we can resolve the target).

For example, I think the call to `bar()` in the example below will get a confusing `toString` (since it'll just be the string `"bar"` and not `"call to bar"` even though it's a call to bar and not a reference to the variable).

```swift
func foo(_ bar: () -> ()) {
  bar()
}

foo({ print("I am a lambda :wave:") })
```

Does that make sense?

MathiasVP (commented):
I like the new `*Operation.qll` files a lot more now 👍. The only concern I have left is with the `toString` on `ApplyExpr`. If you're okay with it, maybe that should be factored out into a later PR?

geoffw0:
> The only concern I have left is with the toString on ApplyExpr. If you're okay with it, maybe that should be factored out into a later PR?

Yep, that should probably never have been part of this PR.  I've just removed those commits from here (via a rebase / force push - I didn't want to leave a confusing mess here) and will be creating a separate PR for them after this is merged (to avoid conflicts).

swift/ql/lib/codeql/swift/elements/expr/ApplyExpr.qll
geoffw0: Thanks for this, I need to think about it a bit more.  But since I've factored out this code for a future PR now, it is no longer an issue holding us up here.

MathiasVP:
> Yep, that should probably never have been part of this PR. I've just removed those commits from here (via a rebase / force push - I didn't want to leave a confusing mess here) and will be creating a separate PR for them after this is merged (to avoid conflicts).

That sounds good. In general, I expect the AST structure around function calls to be cleaned up soon, but we're still discussing how to best to this without complicating the extractor too much.

MathiasVP:
I see some of the output in the CFG tests are changing. I'll take a look at it 👀 now!

MathiasVP:
I hope you don't mind that I pushed f35ab7c2924b27137bc12e5f6bf0e98c6bd16533 to your PR. These tests changed because you fixed the names in https://github.com/github/codeql/pull/9737/commits/9e0cf62cdae127b96ef70f49d2e0f56bc1ccc959 which made us generate a better CFG (I apparently broke them in https://github.com/github/codeql/pull/9619 without noticing).

The CFG construction has special handling for `&&` and `||` (which are function calls in Swift), and by fixing the names we now properly identify those function calls again.

MathiasVP (approved):
LGTM!

geoffw0:
> I hope you don't mind that I pushed https://github.com/github/codeql/commit/f35ab7c2924b27137bc12e5f6bf0e98c6bd16533 to your PR

I don't mind, but in future you probably shouldn't merge your own changes without another set of eyes passing over them.

MathiasVP:
> I don't mind, but in future you probably shouldn't merge your own changes without another set of eyes passing over them.

Good point. I'll keep that in mind!

PR #9738, opened 2022-06-28
Swift: Add a Locatable.getFile() shortcut similar to the one in CPP.

geoffw0:
In one version of a test I tried to use this and couldn't, because it didn't exist.  Its a nice shortcut to have.


MathiasVP (approved):
Agreed. LGTM!

PR #9739, opened 2022-06-28
Swift: Set 'swift/string-length-conflation' to precision `high`

MathiasVP:
... and delete the placeholder query we had in the `queries` directory. The placeholder query was only necessary to have a non-empty query suite to run DCA on, and now that this query is on precision `high` we can create a proper Code Scanning suite.


geoffw0 (approved):
:+1:

PR #9744, opened 2022-06-28
Move contextual queries from src to lib

aeisenberg:
With this change, users are now able to run View AST command in vscode within vscode workspaces that do not include the core libraries. The relevant core library only needs to be installed in the package cache.

Question for reviewers:

1. Does this require a change note?
2. I am getting a failing documentation check. None of these predicates were documented before. Is there some way to skip this check for the failing predicates?


aibaars:
For JavaScript we probably need to move https://github.com/github/codeql/blob/main/javascript/ql/src/Declarations/Declarations.qll to `lib` as well. @asgerf what do you think?

erik-krogh:
> For JavaScript we probably need to move https://github.com/github/codeql/blob/main/javascript/ql/src/Declarations/Declarations.qll to `lib` as well. @asgerf what do you think?

It seems very reasonable to move `Declarations.qll` to `lib` :+1:

jketema:
> Does this require a change note?

I'm a bit split on this. I guess the questions is: how likely is it that someone (besides us) depends on the current location? If it's somewhat likely (I sincerely hope not), it might be good to add a change note that we moved them.

> I am getting a failing documentation check. None of these predicates were documented before. Is there some way to skip this check for the failing predicates?

It seems that these can be fixed relatively straightforwardly, so I would just fix them.

jketema (dismissed):

jketema (approved):

michaelnebel (commented):
C# Looks plausible to me

PR #9745, opened 2022-06-29
Swift: store TRAP files in a temporary folder until the extraction is complete

AlexDenisov:
Currently, we have a number of assertions in the codebase and certain assumptions about the AST. These don't always hold, sometimes leading to a crash in the extractor.
The crashes leave incomplete TRAP files that cannot be imported into the database.

With this change, we still get those incomplete TRAP files, but we also get a database in the end (even though it is also incomplete as we cannot import everything).

### DIFF
--- a/swift/extractor/SwiftExtractor.cpp
+++ b/swift/extractor/SwiftExtractor.cpp
@@ -62,13 +62,13 @@ static void extractDeclarations(const SwiftExtractorConfiguration& config,
   // TODO: find a more robust approach to avoid collisions?
   llvm::StringRef filename = primaryFile ? primaryFile->getFilename() : module.getModuleFilename();
   std::string tempTrapName = filename.str() + '.' + std::to_string(getpid()) + ".trap";
-  llvm::SmallString<PATH_MAX> tempTrapPath(config.trapDir);
+  llvm::SmallString<PATH_MAX> tempTrapPath(config.tempTrapDir);
   llvm::sys::path::append(tempTrapPath, tempTrapName);
 
-  llvm::StringRef trapParent = llvm::sys::path::parent_path(tempTrapPath);
-  if (std::error_code ec = llvm::sys::fs::create_directories(trapParent)) {
-    std::cerr << "Cannot create trap directory '" << trapParent.str() << "': " << ec.message()
-              << "\n";
+  llvm::StringRef tempTrapParent = llvm::sys::path::parent_path(tempTrapPath);
+  if (std::error_code ec = llvm::sys::fs::create_directories(tempTrapParent)) {
+    std::cerr << "Cannot create temp trap directory '" << tempTrapParent.str()
+              << "': " << ec.message() << "\n";
     return;
   }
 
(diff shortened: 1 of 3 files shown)


AlexDenisov:
With this change we can successfully extract Signal-iOS app with 40 files missing and 1865 present.

swift/extractor/SwiftExtractorConfiguration.h
```diff
@@ -9,7 +9,14 @@ struct SwiftExtractorConfiguration {
   std::string trapDir;
   // The location for storing extracted source files.
   std::string sourceArchiveDir;
-  // The arguments passed to the extractor. Used for debugging.
+  // A temporary directory that exists during database creation, but is deleted once the DB is
+  // finalized.
+  std::string scratchDir;
+  // A temporary directory that contains TRAP files before they moved into the final destination.
```
jketema, line 15:
```suggestion
  // A temporary directory that contains TRAP files before they are moved into their final destination.
```

jketema (dismissed):
LGTM (modulo a small typo).

I do wonder whether all the directory creation code should be pulled out of `extractDeclarations` into `extractSwiftFiles`. It seems sufficient to do this only once outside the main loop of `extractSwiftFiles`, or not?

swift/extractor/SwiftExtractor.cpp
```diff
@@ -62,13 +62,13 @@ static void extractDeclarations(const SwiftExtractorConfiguration& config,
...
   llvm::sys::path::append(tempTrapPath, tempTrapName);
 
-  llvm::StringRef trapParent = llvm::sys::path::parent_path(tempTrapPath);
-  if (std::error_code ec = llvm::sys::fs::create_directories(trapParent)) {
-    std::cerr << "Cannot create trap directory '" << trapParent.str() << "': " << ec.message()
-              << "\n";
+  llvm::StringRef tempTrapParent = llvm::sys::path::parent_path(tempTrapPath);
+  if (std::error_code ec = llvm::sys::fs::create_directories(tempTrapParent)) {
+    std::cerr << "Cannot create temp trap directory '" << tempTrapParent.str()
+              << "': " << ec.message() << "\n";
```
redsun82, line 71: would'n it make sense now to use the `getpid` to fill in `config.tempTrapDir`, so each process would get its own separate temp directory and we wouldn't need it on the file names? Probably does not make a lot of difference, but at least it would give configurability of `tempTrapParent` (see comment below)

AlexDenisov:
> I do wonder whether all the directory creation code should be pulled out of extractDeclarations into extractSwiftFiles. It seems sufficient to do this only once outside the main loop of extractSwiftFiles, or not?

I don't think so, as we are preserving the whole path structure e.g.
```
a/b.swift -> trap/a/b.swift.trap
c.swift -> trap/c.swift.trap
d/c.swift -> trap/d/c.swift.trap
```
Or did I misunderstood your question?

swift/extractor/SwiftExtractorConfiguration.h
```diff
@@ -9,7 +9,14 @@ struct SwiftExtractorConfiguration {
...
+  // Subdirectory of the scratchDir.
+  std::string tempTrapDir;
```
redsun82, line 17: unless we use per-process separate `tempTrapDir`, I don't see why we need this to be configurable (and not just the directly the fixed subdirectory of `scratchDir` we currently already have.

jketema:
> Or did I misunderstood your question?

No, then I apparently missed part of what is going on with `config`. I got the impression that this was shared, but it's apparently per file?

jketema:
@AlexDenisov I overlooked `llvm::sys::path::append(tempTrapPath, tempTrapName);`. So, yes the code cannot be moved.

swift/extractor/SwiftExtractor.cpp
AlexDenisov: We can do it this way, yes. What would be the benefit? 🤔

AlexDenisov:
@redsun82 I'll merge this as is, but I'm happy to discuss the temp folders issue you raised.

PR #9746, opened 2022-06-29
Swift: add a test case showing module loading problem

AlexDenisov:
Extractor fails to load separate modules that were built by another
version of an actual compiler.

The test currently fails, but I'd like to get it in separately from [the fix](https://github.com/github/codeql/pull/9702) in order to reduce the other PR diff.

### DIFF
--- a/swift/integration-tests/partial-modules/Sources/partial-modules/partial_modules.swift
+++ b/swift/integration-tests/partial-modules/Sources/partial-modules/partial_modules.swift
@@ -1,0 +1,11 @@
+import A
+import B
+
+public struct partial_modules {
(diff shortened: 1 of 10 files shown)


swift/integration-tests/partial-modules/Sources/partial-modules/partial_modules.swift
```diff
@@ -0,0 +1,11 @@
+import A
+import B
```
AlexDenisov, line 2: Currently, the extractor will fail at importing these modules and will emit a couple of unresolved AST nodes for the following `a.text`/`b.text` calls.

PR #9749, opened 2022-06-29
Swift: Possible improvement to ApplyExpr.toString

geoffw0:
Possible improvement to `ApplyExpr.toString`.

The first three commits were originally part of https://github.com/github/codeql/pull/9737 (but have since been removed from there), where @MathiasVP [said](https://github.com/github/codeql/pull/9737#discussion_r908691911):

> I like the output a lot more now, but I'm not sure we're getting the improved output for the right reasons. It looks like we're now generating `"call to f()"` when we can resolve the call to a static call target `f`, but we print just `"f()"` when we cannot resolve it to a target.
>
> The reason the output looks a lot better now is that the cases where there isn't a static call target are all examples where `getFunction` returns a strange `DotSyntaxCallExpr` (or `SelfApplyExpr`, or `ConstructorRefCallExpr`) which all have strings that look like `"call to f()"` (because they fall into the category of calls where we can resolve the target).
>
> For example, I think the call to `bar()` in the example below will get a confusing `toString` (since it'll just be the string `"bar"` and not `"call to bar"` even though it's a call to bar and not a reference to the variable).
>
> ```swift
> func foo(_ bar: () -> ()) {
>   bar()
> }
>
> foo({ print("I am a lambda :wave:") })
> ```

As a fourth commit I've added the above as a test and it works (or rather doesn't) as anticipated.  The obvious fix would be to make a special case for `DotSyntaxCallExpr` vs everything else, but that might just be a rabbit hole of special cases - ideally I would like to find a cleaner answer.

### DIFF
--- a/swift/ql/lib/codeql/swift/elements/expr/ApplyExpr.qll
+++ b/swift/ql/lib/codeql/swift/elements/expr/ApplyExpr.qll
@@ -11,8 +11,7 @@ class ApplyExpr extends ApplyExprBase {
     not exists(this.getStaticTarget()) and
     result = "call to " + this.getFunction().toString()
     or
-    not exists(this.getStaticTarget()) and
-    not exists(this.getFunction()) and
+    not exists(this.getFunction()) and // (this implies there is no `getStaticTarget` either)
     result = "call to ..."
   }
 }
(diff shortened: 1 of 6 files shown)


geoffw0:
Waiting on https://github.com/github/codeql-c-team/issues/1019 .

geoffw0:
This PR is redundant following the IPA work.

PR #9751, opened 2022-06-29
JS: add call-edge for dynamic dispatch to unknown property from an object literal

erik-krogh:
CVE-2019-10807: TP/TN

[The evaluation suggests a very slight performance regression](https://github.com/github/codeql-dca-main/tree/data/PR-9751-0-javascript__2/reports).   
No new results from the evaluation, but plenty of new call-edges (see the meta alert diff).  
The call-edges look good to me.  

I know this can be seen as a controversial addition, and I'm very open to suggestions (or arguments for why we shouldn't do this).


esbena:
> I know this can be seen as a controversial addition

I would like rein this in as much as possible.
Perhaps some of our [unpromoted route handlers](https://github.com/github/codeql-javascript-team/issues?q=is%3Aissue+is%3Aopen+label%3A%22Topic%3Aroute+handlers%22+) can inspire this line of work.

Facets to consider that further restrict the scope of this:

- receiver is an object literal that never escapes (we have an (expensive?) escape analysis around somewhere)
- the resolved function is written to a fixed property name
- every property write on the receiver is with a fixed property name
- every property write on the receiver is of the same type (function), i.e. it is a simple dictionary

erik-krogh:
> * receiver is an object literal that never escapes

I went for that one, it fit with the case I needed to support the CVE.  

[A new evaluation](https://github.com/github/codeql-dca-main/tree/data/erik-krogh/dynCall__nightly-old__code-scanning/reports) still shows some new call edges, and no new results.

asgerf:
After giving this some more thought, I think we can go ahead with this as it is and then refine it if we notice any FPs caused by this.

erik-krogh:
> After giving this some more thought, I think we can go ahead with this as it is and then refine it if we notice any FPs caused by this.

👍  

I've merged with `main` to fix a merge conflict.

asgerf (dismissed):

asgerf (approved):

PR #9755, opened 2022-06-29
Kotlin: support JvmStatic annotation

smowton:
This makes non-companion object methods into static methods, and for companion objects introduces static proxy methods that call the companion instance method.

Note this doesn't quite implement what kotlinc does, since it will also eliminate getters and setters by promoting an object field into a static field, but our translation is simpler and only differs in private members' details.


smowton:
@tamasvajk oops, test pushed

java/kotlin-extractor/src/main/kotlin/KotlinFileExtractor.kt
```diff
@@ -1349,44 +1463,48 @@ open class KotlinFileExtractor(
...
+        get() = this.willExtractAsStatic && parent is IrFile
 
-    private val IrDeclaration.isStatic: Boolean
-        get() = this is IrSimpleFunction && dispatchReceiverParameter == null ||
+    private fun isStaticAnnotatedNonCompanionMember(f: IrSimpleFunction) =
+        f.parentClassOrNull?.isNonCompanionObject == true &&
+                (f.hasAnnotation(jvmStaticFqName) || f.correspondingPropertySymbol?.owner?.hasAnnotation(jvmStaticFqName) == true)
+
+    private val IrDeclaration.willExtractAsStatic: Boolean
+        get() = this is IrSimpleFunction && (isStaticFunction(this) || isStaticAnnotatedNonCompanionMember(this)) ||
```
tamasvajk, line 1507: I find the naming of these `will...` members odd. I would have used `should...` instead.

java/kotlin-extractor/src/main/kotlin/KotlinFileExtractor.kt
```diff
@@ -458,6 +461,75 @@ open class KotlinFileExtractor(
...
+
+        // Add synthetic forwarders for any JvmStatic methods or properties:
+        val companionObject = c.companionObject() ?: return
+
+        val cType = c.typeWith()
+        val companionType = companionObject.typeWith()
+
+        fun makeProxyFunction(f: IrFunction) {
+            // Emit a function in class `c` that delegates to the same function defined on `c.CompanionInstance`.
+            val proxyFunctionId = tw.getLabelFor<DbMethod>(getFunctionLabel(f, classId, listOf()))
```
tamasvajk, line 474: Should this function be added also to `compiler_generated`?

tamasvajk (commented):
There are quite a few generated constructs in this change, I would add a `printAst` test just to visually be able to verify the generated elements.

java/kotlin-extractor/src/main/kotlin/KotlinFileExtractor.kt
smowton: Done

java/kotlin-extractor/src/main/kotlin/KotlinFileExtractor.kt
smowton: Done

tamasvajk:
> @tamasvajk added a PrintAst test.

Looks good. There are some empty `1: (Annotations)` nodes, but that's independent of this PR.

tamasvajk (approved):
LGTM

tamasvajk (approved):

PR #9766, opened 2022-06-30
Kotlin: Fix a label

igfoo:
We want the .javaResult.id of a TypeResults.


smowton:
Got a reproducer we can add as a test case?

igfoo:
No, I found it by querying the code.

Looking at the comment above, I'm not sure if it would even be possible to make one.

smowton (approved):
Fair enough

PR #9768, opened 2022-06-30
Kotlin: Mangle names of internal functions to match JVM symbols

smowton:
These are suffixed with `$main` by default, or `$modulename` if one is provided to the compiler.


tamasvajk (approved):
LGTM

PR #9769, opened 2022-07-01
C#: Fix dataflow for default constructors

tamasvajk:


hvitved (commented):
Thanks for looking into this Tamas. I realize that your PR is still in draft, but I have pushed a small code review suggestion. Feel free to discard my commit if you disagree.

tamasvajk:
> Thanks for looking into this Tamas. I realize that your PR is still in draft, but I have pushed a small code review suggestion. Feel free to discard my commit if you disagree.

Thank you. The PR is draft, because I was struggling with the performance. This approach is probably a lot more lightweight than going with `InitializerSplitting::constructorInitializes`, I'll trigger a DCA run to see if it fixed my perf regression.

csharp/ql/lib/semmle/code/csharp/dataflow/internal/DataFlowDispatch.qll
```diff
@@ -21,7 +21,13 @@ private import semmle.code.csharp.frameworks.system.collections.Generic
  */
 DotNet::Callable getCallableForDataFlow(DotNet::Callable c) {
   exists(DotNet::Callable unboundDecl | unboundDecl = c.getUnboundDeclaration() |
-    result.hasBody() and
+    (
+      result.hasBody()
+      or
+      // take synthesized bodies into account, e.g. implicit constructors
+      // with field initializer assignments
+      result = any(ControlFlow::Nodes::ElementNode n).getEnclosingCallable()
```
tamasvajk, line 29: Should we add a `not result.hasBody()` and maybe a type check `result instanceof InstanceConstructor`  to make it explicit what case we cover here?

hvitved: I would rather not, actually, as we are then also prepared for cases where we might synthesize other kinds of bodies in the future.

hvitved (approved):

michaelnebel (commented):
Looks plausible to me!

PR #9771, opened 2022-07-01
Swift: locally run integration tests

redsun82:
Minimal recreations of internal `integration-tests-runner.py` and
`create_database_utils.py` are provided to be able to run the
integration tests on the codeql repository with a released codeql CLI.

For the moment we skip the database checks by default, as we are still
producing inconsistent results.

This is also added to github workflows, but for the moment only on Linux:
there is one test that has a different result on CI, and moreover they seem
to take awfully slow on macos (last run, the integration test step took
1min 20s on Linux, and more than 17min on macOS...).


redsun82:
I was a bit surprised that I did not have to setup `swift` on the runner, it seems it's already included. @criemen maybe you know whether we can count on this or not?

AlexDenisov (approved):
❤️

criemen:
@redsun82 The available software in the base image is documented here:
https://github.com/actions/virtual-environments/blob/main/images/linux/Ubuntu2004-Readme.md#language-and-runtime
It does include an installation of Swift, so I think you'll be able to rely on that, if you don't need specific version constraints.

PR #9773, opened 2022-07-01
Swift: More improvements to swift/string-length-conflation

geoffw0:
More improvements to `swift/string-length-conflation`:
 - conflation in the `String` -> `NSString` direction is now detected 🎉 
 - slightly more data flow is permitted (in particular through operators `*`, `/`)
 - additional test cases

Notably for the `String` and `String.Index` member functions, I can't find a way of checking the associated class is correct (i.e. that we're looking at a call to `String.index` not `foo.index`).  While this would be nice to have, in practice since we match on the argument and parameter names as well, so I think false positives are unlikely.

### DIFF
--- a/swift/ql/src/queries/Security/CWE-135/StringLengthConflation.ql
+++ b/swift/ql/src/queries/Security/CWE-135/StringLengthConflation.ql
@@ -30,6 +30,14 @@ class StringLengthConflationConfiguration extends DataFlow::Configuration {
       node.asExpr() = member and
       flowstate = "String"
     )
+    or
+    // result of a call to to `NSString.length`
+    exists(MemberRefExpr member |
+      member.getBaseExpr().getType().getName() = ["NSString", "NSMutableString"] and
+      member.getMember().(VarDecl).getName() = "length" and
+      node.asExpr() = member and
+      flowstate = "NSString"
+    )
   }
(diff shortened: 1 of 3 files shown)


swift/ql/src/queries/Security/CWE-135/StringLengthConflation.ql
```diff
@@ -30,6 +30,14 @@ class StringLengthConflationConfiguration extends DataFlow::Configuration {
       node.asExpr() = member and
       flowstate = "String"
     )
+    or
+    // result of a call to to `NSString.length`
```
intrigus-lgtm, line 34: Typo: to to

geoffw0: Thanks.  Fixed.

MathiasVP (approved):
LGTM! Indeed, what you write here:

> Notably for the String and String.Index member functions, I can't find a way of checking the associated class is correct (i.e. that we're looking at a call to String.index not foo.index). While this would be nice to have, in practice since we match on the argument and parameter names as well, so I think false positives are unlikely.

is something we have to provide a better API for. Let's get this merged now so that it doesn't fall between the cracks as we're going on vacation.

PR #9783, opened 2022-07-06
Kotlin: Add MaD for stdlib

tamasvajk:


java/ql/lib/semmle/code/java/frameworks/kotlin/StdLibGenerated.qll
```diff
@@ -1796,12 +1858,16 @@ private class StdLibGeneratedSummaryCsv extends SummaryModelCsv {
...
+        "kotlin;UByte;false;toUByte;();;Argument[-1];ReturnValue;taint;generated",
         "kotlin;UByte;false;toUByte;();;Argument[-1];ReturnValue;value;generated",
         "kotlin;UByteArrayKt;false;ubyteArrayOf;(UByteArray);;Argument[0].Element;ReturnValue;taint;generated",
+        "kotlin;UInt;false;toUInt;();;Argument[-1];ReturnValue;taint;generated",
         "kotlin;UInt;false;toUInt;();;Argument[-1];ReturnValue;value;generated",
         "kotlin;UIntArrayKt;false;uintArrayOf;(UIntArray);;Argument[0].Element;ReturnValue;taint;generated",
+        "kotlin;ULong;false;toULong;();;Argument[-1];ReturnValue;taint;generated",
         "kotlin;ULong;false;toULong;();;Argument[-1];ReturnValue;value;generated",
         "kotlin;ULongArrayKt;false;ulongArrayOf;(ULongArray);;Argument[0].Element;ReturnValue;taint;generated",
+        "kotlin;UShort;false;toUShort;();;Argument[-1];ReturnValue;taint;generated",
```
tamasvajk, line 1870: These taint steps that already exist as value steps should not be added.

java/ql/lib/semmle/code/java/frameworks/kotlin/StdLibGenerated.qll
```diff
@@ -1718,11 +1777,13 @@ private class StdLibGeneratedSummaryCsv extends SummaryModelCsv {
         "kotlin.time;Duration;false;div;(int);;Argument[-1];ReturnValue;taint;generated",
         "kotlin.time;Duration;false;getAbsoluteValue;();;Argument[-1];ReturnValue;taint;generated",
         "kotlin.time;Duration;false;minus;(Duration);;Argument[-1];ReturnValue;taint;generated",
+        "kotlin.time;Duration;false;plus;(Duration);;Argument[-1];ReturnValue;taint;generated",
         "kotlin.time;Duration;false;plus;(Duration);;Argument[-1];ReturnValue;value;generated",
         "kotlin.time;Duration;false;times;(double);;Argument[-1];ReturnValue;taint;generated",
         "kotlin.time;Duration;false;times;(int);;Argument[-1];ReturnValue;taint;generated",
         "kotlin.time;DurationKt;false;times;(double,Duration);;Argument[1];ReturnValue;taint;generated",
         "kotlin.time;DurationKt;false;times;(int,Duration);;Argument[1];ReturnValue;taint;generated",
+        "kotlin.time;TimeMark;true;minus;(Duration);;Argument[-1];ReturnValue;taint;generated",
```
tamasvajk, line 1786: Why wasn't this flow found during the first MaD generation pass?

tamasvajk, line 1786: Based on the below two lines, one would expect to also see:
```
"kotlin.time;TimeMark;true;minus;(Duration);;Argument[0];ReturnValue;taint;generated",
```
This line is not generated, because we don't track flow through unary minus of double.

java/ql/lib/semmle/code/java/frameworks/kotlin/StdLibGenerated.qll
```diff
@@ -1077,6 +1112,7 @@ private class StdLibGeneratedSummaryCsv extends SummaryModelCsv {
         "kotlin.jvm.internal;ClassReference$Companion;false;getClassSimpleName;(Class);;Argument[-1];ReturnValue;taint;generated",
         "kotlin.jvm.internal;CollectionToArray;false;toArray;(Collection);;Argument[0].Element;ReturnValue;taint;generated",
         "kotlin.jvm.internal;CollectionToArray;false;toArray;(Collection,Object[]);;Argument[0].Element;ReturnValue;taint;generated",
+        "kotlin.jvm.internal;FunInterfaceConstructorReference;true;toString;();;Argument[-1];ReturnValue;taint;generated",
```
tamasvajk, line 1115: This class seems to be generated in IR, why do we find flow?

java/ql/lib/semmle/code/java/frameworks/kotlin/StdLibGenerated.qll
```diff
@@ -0,0 +1,1875 @@
+/**
+ * THIS FILE IS AN AUTO-GENERATED MODELS AS DATA FILE. DO NOT EDIT.
+ * Definitions of taint steps in the StdLibGenerated framework.
+ */
```
igfoo, line 4: Is it possible to add the Kotlin version number to this header?

java/ql/lib/semmle/code/java/frameworks/kotlin/StdLibGenerated.qll
tamasvajk: It's now found during the first pass.

java/ql/test/library-tests/dataflow/external-models/steps.expected
```diff
@@ -1,4 +1,5 @@
 invalidModelRow
+| Dubious name "getPlatformTypeUpperBound$kotlin_stdlib" in summary model. |
```
tamasvajk, line 2: This is an `internal` property. Its modifiers are `final`, `internal`, and `public`. The latter one is causing this to show up in the MaD generation. I'll have to investigate where the `public` is coming from. Is it possible that we extract the modifiers in both kotlin and java, and `internal` is actually `public` when viewed from java?

java/ql/lib/semmle/code/java/frameworks/kotlin/StdLibGenerated.qll
tamasvajk: Done in https://github.com/github/codeql/pull/9783/commits/d07effb202876c6e5053e6eae5631d1d43485655.

java/ql/test/kotlin/library-tests/dataflow/summaries/test.kt
```diff
@@ -0,0 +1,75 @@
...
+        sink(mg0.value)                             // $ hasTaintFlow=g
+        sink(mg0.component2())
+
+        val iv = IndexedValue<String>(5, taint("h"))
+        sink(iv)                                    // $ hasTaintFlow=h
+        sink(iv.index)
+        sink(iv.component2())                       // $ hasTaintFlow=h
+
+        val strings = arrayOf("", taint("i"))
+        sink(strings.withIndex())
```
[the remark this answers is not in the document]
tamasvajk: No summary is generated because the parameter is captured in a lambda, which is excluded in the MaD generator. I've added a manual model for this.

tamasvajk:
The DCA run doesn't show significant change in unique alert count. So I think it's safe to add these models as an initial approximation of flow through methods.

misc/scripts/models-as-data/generate_flow_model.py
```diff
@@ -178,7 +187,7 @@ def makeContent(self):
         return f"""
 /** 
  * THIS FILE IS AN AUTO-GENERATED MODELS AS DATA FILE. DO NOT EDIT.
- * Definitions of taint steps in the {self.shortname} framework.
+ * Definitions of taint steps in the {self.friendlyname} framework.
```
michaelnebel, line 190: Very nice!

java/ql/test/library-tests/dataflow/external-models/steps.expected
michaelnebel: Have you had a chance to look into this? In any case, it might be a good idea to exclude this from model generation of there is any doubt.

michaelnebel:
If the negative models are generated in this PR, then it matters whether this PR or https://github.com/github/codeql/pull/10170 gets merged first.
If negative models are merged prior to #10170 then it is basically dead code.

java/ql/test/library-tests/dataflow/external-models/steps.expected
tamasvajk: We extract `getPlatformTypeUpperBound$kotlin_stdlib` as internal in kotlin and public in java. So effectively it's public and therefore shows up in the model generation. I asked internally the team how to move on with this.

tamasvajk: This will be fixed by https://github.com/github/codeql/pull/10221

igfoo (dismissed):
Looks plausible to me.

michaelnebel (dismissed):
Looks good to me!

michaelnebel:
Note that https://github.com/github/codeql/pull/10170 is merged.

igfoo (approved):

PR #9785, opened 2022-07-06
C++: Nullness.qll bug fixes

bdrodes:
Corrects a bug in nullCheckExpr and validCheckExpr in Nullness.qll for logical AND expressions. Previously only checking null/valid on expressions in the right hand side of the AND operation. Now checking both left and right.


jketema:
> To answer your question though, I found this bug making my own query.

No worries. If there had been a specific query you were looking at it would have been easiest to add a test to that query. I don't think we test this functionality directly. Let me think a bit about how to best solve this.

bdrodes:
> > Would it be easier to bundle them into this PR or to do them separately?
> 
> Depends a bit on their size. If they're small, I don't have a problem with having them as part of the current PR.

I ended up only having one addition change which is now part of this PR. It's very small.

jketema:
Thanks for clarifying.

https://github.com/github/codeql/pull/9793 has been merged. Could you rebase this PR and update the tests/test results as appropriate? If you're not sure how to do the tests, please just rebase, and I'll give you a hand.

bdrodes:
@jketema Yea I could use a hand to properly update the test results.

bdrodes:
Is a merge of main sufficient or do I need to do I need to do anything fancier?

cpp/ql/lib/semmle/code/cpp/controlflow/Nullness.qll
```diff
@@ -169,7 +169,10 @@ class AnalysedExpr extends Expr {
    */
   predicate isDef(LocalScopeVariable v) {
     this.inCondition() and
-    this.(Assignment).getLValue() = v.getAnAccess()
+    (
+      this.(Assignment).getLValue() = v.getAnAccess() or
+      exists(Initializer i | this.getEnclosingStmt() = i.getEnclosingStmt() and v = i.getDeclaration())
+    )
```
jketema, line 175: Shouldn't the following suffice?
```suggestion
    (
      this.(Assignment).getLValue() = v.getAnAccess() or
      this.(ConditionDeclExpr).getVariableAccess() = v.getAnAccess()
    )
```

jketema: Done.

cpp/ql/test/library-tests/controlflow/nullness/nullness.expected
```diff
@@ -7,9 +7,14 @@
...
-| test.cpp:18:9:18:17 | ... && ... | test.cpp:5:13:5:13 | v | is not null | is not valid |
+| test.cpp:18:9:18:17 | ... && ... | test.cpp:5:13:5:13 | v | is not null | is valid |
 | test.cpp:19:9:19:18 | ... && ... | test.cpp:5:13:5:13 | v | is null | is not valid |
-| test.cpp:20:9:20:18 | ... && ... | test.cpp:5:13:5:13 | v | is not null | is not valid |
+| test.cpp:20:9:20:18 | ... && ... | test.cpp:5:13:5:13 | v | is null | is not valid |
 | test.cpp:21:9:21:14 | ... = ... | test.cpp:5:13:5:13 | v | is null | is not valid |
 | test.cpp:21:9:21:14 | ... = ... | test.cpp:7:10:7:10 | b | is not null | is valid |
-| test.cpp:22:17:22:17 | b | test.cpp:7:10:7:10 | b | is not null | is valid |
+| test.cpp:22:9:22:14 | ... = ... | test.cpp:5:13:5:13 | v | is not null | is not valid |
+| test.cpp:22:9:22:14 | ... = ... | test.cpp:7:13:7:13 | c | is not null | is not valid |
```
geoffw0, line 16: Please would you explain these two results?  It looks to me like if `c = !v` is true, `c` will be true (only after the assignment) and `v` must be false.

If this is just a limitation of the analysis, I'm happy with that and happy for it to be documented by this test.  I just want to check I'm understanding correctly.

jketema: Yeah, those are a bit weird. There's two things going on here:

1. RHSs of assignments are not covered, as they're technically not part of a condition. So, those expressions for each variable will always result in `is not null` and `is not valid`
2. The `c` case is a C++17-style if-initializer (note the `c` in the condition on the next line of the expected file). These also fall outside the condition that is evaluated in the `if`, and hence we also get `is not null` and `is not valid` there.

geoffw0 (approved):
Looks good to me. :+1: :+1:

jketema:
@bdrodes Merged! Thanks for your contribution.

PR #9795, opened 2022-07-11
Swift: extract more entities

redsun82:
This extracts `OpenedArchetypeType`, `UnresolvedPatternExpr` and `IfCondifDecl`.

Commit by commit review is encouraged.


swift/extractor/visitors/ExprVisitor.cpp
```diff
@@ -667,5 +667,11 @@ void ExprVisitor::emitLookupExpr(const swift::LookupExpr* expr, TrapLabel<Lookup
     dispatcher_.emit(LookupExprMembersTrap{label, declLabel});
   }
 }
+codeql::UnresolvedPatternExpr ExprVisitor::translateUnresolvedPatternExpr(
```
jketema, line 670: No empty line after the previous function?

swift/extractor/visitors/ExprVisitor.cpp
```diff
@@ -667,5 +667,11 @@ void ExprVisitor::emitLookupExpr(const swift::LookupExpr* expr, TrapLabel<Lookup
...
+    swift::UnresolvedPatternExpr& expr) {
+  auto entry = dispatcher_.createEntry(expr);
+  entry.sub_pattern = dispatcher_.fetchLabel(expr.getSubPattern());
+  return entry;
+}
```
jketema, line 676: Empty line?

swift/extractor/visitors/TypeVisitor.cpp
```diff
@@ -367,5 +367,10 @@ codeql::BuiltinVectorType TypeVisitor::translateBuiltinVectorType(
     const swift::BuiltinVectorType& type) {
   return createTypeEntry(type);
 }
-
```
jketema, line 370: No empty line after the previous function?

jketema (commented):
Some formatting issues 😄

swift/extractor/visitors/ExprVisitor.h
```diff
@@ -84,6 +84,11 @@ class ExprVisitor : public AstVisitorBase<ExprVisitor> {
   codeql::BridgeFromObjCExpr translateBridgeFromObjCExpr(const swift::BridgeFromObjCExpr& expr);
   codeql::DotSelfExpr translateDotSelfExpr(const swift::DotSelfExpr& expr);
   codeql::ErrorExpr translateErrorExpr(const swift::ErrorExpr& expr);
+  // following requires non-const because:
```
jketema, line 87:
```suggestion
  // The following requires non-const because:
```

swift/extractor/visitors/ExprVisitor.h
```diff
@@ -84,6 +84,11 @@ class ExprVisitor : public AstVisitorBase<ExprVisitor> {
...
+  // * `swift::UnresolvedPatternExpr::getSubPattern` gives `const swift::Pattern*` on const refs
```
jketema, line 88: I think this should be:
```suggestion
  // * `swift::UnresolvedPatternExpr::getSubPattern` gives `const swift::Pattern*` no const refs
```
or I simply don't understand what you're trying to say here.

swift/extractor/visitors/VisitorBase.h
```diff
@@ -29,7 +29,7 @@ class VisitorBase {
  public:                                                                                          \
   void visit##CLASS##KIND(swift::CLASS##KIND* e) {                                                \
     using TranslateResult = std::invoke_result_t<decltype(&CrtpSubclass::translate##CLASS##KIND), \
-                                                 CrtpSubclass, swift::CLASS##KIND>;               \
+                                                 CrtpSubclass, swift::CLASS##KIND&>;              \
```
jketema, line 32: Is this necessary for the unresolved pattern change, or is this just a bug fix?

jketema (commented):
`UnresolvedPatternExpr` related comments

jketema (commented):
Comments for `IfConfigDecl`

swift/extractor/visitors/ExprVisitor.h
redsun82: I'll reword, it is confusing.

What I wanted to say, is that this is one rare cases where the swift frontend library went for const correctness and provided a const qualified method, so `getSubPattern` gives back `const swift::Pattern*` on `const UnresolvedPatternExpr&` (practically all other getters in the library do not have a `const` overload). It would be good, if they also provided a version of the visitors working on `const` pointers, but they don't.

swift/extractor/visitors/ExprVisitor.cpp
redsun82: My IDE insists on that when generating the definition. I hoped there would be a clang-format option for this, and there is (`SeparateDefinitionBlocks`), but unfortunately only from `clang-format` version 14, while we officially use version 13 for now... I will probably install version 14 locally, if there is any formatting difference it will be caught by the pre-commit hook.

jketema: I don't understand this, because it is and was correct higher up in this file. What's different?

swift/extractor/infra/SwiftDispatcher.h
```diff
@@ -274,6 +288,7 @@ class SwiftDispatcher {
   // TODO: The following methods are supposed to redirect TRAP emission to correpsonding visitors,
   // which are to be introduced in follow-up PRs
   virtual void visit(swift::Decl* decl) = 0;
+  virtual void visit(const swift::IfConfigClause* clause) = 0;
   virtual void visit(swift::Stmt* stmt) = 0;
   virtual void visit(swift::StmtCondition* cond) = 0;
   virtual void visit(swift::CaseLabelItem* item) = 0;
```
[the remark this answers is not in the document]
redsun82: This is related to the issue commented on `ExprVisitor::translateUnresolvedPatternExpr`. If we could be consistently const correct, all should be `const` qualified here. However as long as we rely on the visitors the swift frontend library provides us, we must provide them non const pointers. As we don't rely on those visitors for `IfConfigClause`, I put the correct constness there.

Now that I notice, there is an obsolete `TODO` on line 288 here, I will update that to have consistency in `const` qualifiers, but I'd rather not fix that in this PR.

swift/extractor/visitors/SwiftVisitor.h
```diff
@@ -21,6 +21,7 @@ class SwiftVisitor : private SwiftDispatcher {
 
  private:
   void visit(swift::Decl* decl) override { declVisitor.visit(decl); }
+  void visit(const swift::IfConfigClause* clause) override { declVisitor.visit(clause); }
   void visit(swift::Stmt* stmt) override { stmtVisitor.visit(stmt); }
   void visit(swift::StmtCondition* cond) override { stmtVisitor.visitStmtCondition(cond); }
   void visit(swift::CaseLabelItem* item) override { stmtVisitor.visitCaseLabelItem(item); }
```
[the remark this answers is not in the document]
redsun82: see my answer on `SwiftDispatcher`

swift/extractor/visitors/VisitorBase.h
redsun82: If I recall correctly, yes, this was required so it could detect that implementation on a non-const reference, but I can double-check

redsun82: yep, code does not compile without this

swift/ql/test/extractor-tests/generated/decl/IfConfigClause/IfConfigClause_getCondition.expected
```diff
@@ -0,0 +1,4 @@
+| if_config.swift:1:1:1:1 | IfConfigClause | if_config.swift:1:5:1:5 | FOO (unresolved) |
+| if_config.swift:4:1:4:1 | IfConfigClause | if_config.swift:4:9:4:19 | call to ... |
+| if_config_active.swift:3:1:3:1 | IfConfigClause | if_config_active.swift:3:5:3:5 | FOO (unresolved) |
```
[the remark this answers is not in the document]
redsun82: exactly, all `#if` and `#elseif` conditions actually appear as unresolved in the frontend. As far as I can tell, they are evaluated in a different environment than the usual compiled one. We might need to get back on this at some point if we want to do some queries exploring these conditions, but for now everything in conditions (even logical operators, or the `arch` or `os` testers) is unresolved

swift/extractor/visitors/ExprVisitor.cpp
redsun82: The difference is that I was correcting it by hand and forgot it here :sweat_smile:

But now I installed `clang-format-14` and added the configuration to a git excluded `.clangformat`, so now this will be consistent (at least from my side). We can add the `SeparateDefinitionBlocks: Always` option to our `.clangformat` files if and when we decide to move to `clang-format` 14

swift/codegen/schema.yml
```diff
@@ -561,6 +570,8 @@ UnresolvedMemberExpr:
 UnresolvedPatternExpr:
   _extends: Expr
   _pragma: qltest_skip  # we should really never extract these
+  _children:
+    sub_pattern: Pattern
```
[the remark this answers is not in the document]
redsun82: well, I think I can actually reproduce this in an inactive `#if` clause, let me try it. If it's the case, I will remove the comment here

swift/extractor/infra/SwiftDispatcher.h
```diff
@@ -78,7 +78,7 @@ class SwiftDispatcher {
     waitingForNewLabel = e;
     visit(e);
     if (auto l = store.get(e)) {
-      if constexpr (!std::is_base_of_v<swift::TypeBase, E>) {
+      if constexpr (std::is_base_of_v<LocatableTag, TrapTagOf<E>>) {
```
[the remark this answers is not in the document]
redsun82: I think this was necessary at some stage of my local changes, but it is not any more. It still makes more sense to make the test about being locatable rather than not being a type, but I've actually already included that in [another PR](https://github.com/github/codeql/pull/9818) where this was necessary, so I'll revert it here

swift/codegen/schema.yml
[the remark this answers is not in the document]
redsun82: done now :+1:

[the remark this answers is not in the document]
jketema: You didn't need to update any `expected` files?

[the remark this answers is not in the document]
jketema: You still need to push?

jketema:
I think you just need to resolve the merge conflict now?

jketema:
> I think you just need to resolve the merge conflict now

What's with that awkward formatting? This really seems like a bug in the formatter.

jketema:
> What's with that awkward formatting? This really seems like a bug in the formatter.

This time in the merge commit.

jketema (approved):

PR #9797, opened 2022-07-11
Ruby: fix defining every dataflow node as a command execution sink

nickrolfe:
For any database where there is at least one call to `execute_command`, `rake`, `rails_command`, or `git` from `Rails::Generations::Actions`, we accidentally defined every dataflow node in the database as a system command execution sink.

Originally introduced in https://github.com/github/codeql/pull/9574, but presumably not spotted by the DCA run because none of our DCA projects call one of those methods.

### DIFF
--- a/ruby/ql/lib/change-notes/2022-07-11-command-execution.md
+++ b/ruby/ql/lib/change-notes/2022-07-11-command-execution.md
@@ -2,5 +2,5 @@
 category: minorAnalysis
 ---
 * Fixed a bug causing every expression in the database to be a considered a system-command execution sink when calls to any of the following methods exist:
-  * The `spawn", "fspawn", "popen4", "pspawn", "system", "_pspawn" methods and the backtick operator from the `POSIX::spawn` gem.
+  * The `spawn`, `fspawn`, `popen4`, `pspawn`, `system`, `_pspawn` methods and the backtick operator from the `POSIX::spawn` gem.
   * The `execute_command`, `rake`, `rails_command`, and `git` methods in `Rails::Generation::Actions`.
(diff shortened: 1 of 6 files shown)


ruby/ql/lib/codeql/ruby/frameworks/Railties.qll
```diff
@@ -43,7 +43,7 @@ module Railties {
 
     override DataFlow::Node getAnArgument() { result = this.getArgument([0, 1]) }
 
-    override predicate isShellInterpreted(DataFlow::Node arg) { any() }
+    override predicate isShellInterpreted(DataFlow::Node arg) { arg = this.getAnArgument() }
```
calumgrant, line 46: Could we restrict this to argument 1?

nickrolfe: A quick look at the implementation (click 'show source' on [this page](https://apidock.com/rails/v5.0.0.1/Rails/Generators/Actions/execute_command)) suggests that argument 0 would also be shell-interpreted (it gets passed through the `extify` method, but that just adds the `.bat` extension on Windows, and doesn't do any sanitization).

nickrolfe:
6632dfa includes a similar fix for `PosixSpawn::SystemCall::isShellInterpreted`, which did not adequately bind `arg`. This only affected databases where there existed at least one call to `spawn`, `fspawn`, `popen4`, `pspawn`, `system`, `_pspawn` methods and the backtick operator from the `POSIX::spawn` library.

ruby/ql/lib/change-notes/2022-07-11-command-execution.md
```diff
@@ -0,0 +1,6 @@
+---
+category: minorAnalysis
+---
+* Fixed a bug causing every expression in the database to be a considered a system-command execution sink when calls to any of the following methods exist:
```
intrigus-lgtm, line 4:
```suggestion
* Fixed a bug causing every expression in the database to be considered a system-command execution sink when calls to any of the following methods exist:
```

calumgrant (approved):

PR #9798, opened 2022-07-11
JS: use small steps in TypeBackTracker correctly

erik-krogh:
The `smallStep` predicate should be used as: `t2 = t.smallstep(result, myType(t2))`.  

Most of the uses of `smallStep` did this correctly.   
However, the documentation of `smallStep` and two uses had flipped `t` and `t2`.  

So I fixed that.  

Found while triaging CVE-2022-31127, but I don't think this fix will help us flag that CVE (the source is not library input).  

[Evaluation shows neutral performance](https://github.com/github/codeql-dca-main/tree/data/erik-krogh/pr-9798-c27ddf6__nightly-old__CustomSuite/reports) and some new TPs.


asgerf (approved):
I'm not sure the API ought to be like this, but LGTM as this clearly fixes a bug.

PR #9805, opened 2022-07-12
Swift: collapse `TypeRepr` hierarchy

redsun82:
Now `TypeRepr` is a final class in the AST, which is more or less just
a type with a location in code.

As the frontend does not provide a direct way to get a type from a
type representation, this information must be provided when fetching
the label of a type repr.

This meant:
* removing the type repr field from `EnumIsCaseExpr`: this is a virtual
  AST node introduced in place of some kinds of `IsEpxr`. The type
  repr is still available from the `ConditionalCheckedCastExpr` wrapped
  by this virtual node, and we will rebuild the original `IsExpr` with
  the IPA layer.
* adding some logic to get the type of keypath roots from
  `KeyPathExpr`. This was done to keep the `TypeRepr` to `Type` relation
  total in the DB, but goes against the design of a dumb extractor. The
  logic could be moved to QL in the future
* in the control flow library, `TypeRepr` children are now ignored. As
  far as I can tell, there is no runtime evaluation going on in
  `TypeRepr`s, so it does not make much sense to have control flow
  through them.
* when fetching a label for a `swift::ASTNode`, we cannot do it for `TypeRepr`,
  and we just skip fetching in that case (leading to an assertion). This behaviour
  relies on the fact that we only extract `swift::ASTNode` in the context of
  `BraceStmt`, where a standalone `TypeRepr` is actually impossible
  We should probably decouple `AstNode` from what can appear in a `BraceStmt`.


swift/extractor/infra/SwiftDispatcher.h
```diff
@@ -263,9 +263,15 @@ class SwiftDispatcher {
 
   template <typename Tag, typename T, typename... Ts>
   bool fetchLabelFromUnionCase(const llvm::PointerUnion<Ts...> u, TrapLabel<Tag>& output) {
-    if (auto e = u.template dyn_cast<T>()) {
-      output = fetchLabel(e);
-      return true;
+    // we rely on the fact that when we extract `ASTNode` instances (which only happens
+    // on `BraceStmt` elements), we cannot encounter a standalone `TypeRepr` there, so we skip
+    // this case, which would be problematic as we would not be able to provide the corresponding
```
jketema, line 268: I assume you mean:
```suggestion
    // this case; extracting `TypeRepr`s here would be problematic as we would not be able to provide the corresponding
```

jketema (commented):
C++ changes LGTM. One small comment.

swift/extractor/infra/SwiftDispatcher.h
jketema: You might want to reflow that comment a bit 😄

redsun82: done :slightly_smiling_face:

jketema (approved):

PR #9806, opened 2022-07-12
Kotlin: Extract an ErrorType if we fail to correctly extract a type

igfoo:
This adds an `ErrorType`, and the Kotlin extractor uses it when presented with a type that it doesn't know how to extract.


smowton (approved):

java/ql/lib/change-notes/2022-07-12-errortype.md
```diff
@@ -0,0 +1,4 @@
+---
+category: feature
+---
+* Added an `ErrorType` class. An instance of this class will be used if an extractor is unable to extract as type, or if an up/downgrade script is unable to provide a type.
```
aschackmull, line 4:
```suggestion
* Added an `ErrorType` class. An instance of this class will be used if an extractor is unable to extract a type, or if an up/downgrade script is unable to provide a type.
```

aschackmull (approved):

PR #9811, opened 2022-07-12
Kotlin: Implement JvmOverloads annotation

smowton:
This generates functions that omit parameters with default values, rightmost first, such that Java can achieve a similar experience to Kotlin (which represents calls internally as if the default was supplied explicitly, and/or uses a $default method that supplies the needed arguments).

A complication: combining JvmOverloads with JvmStatic means that both the companion object and the surrounding class get overloads.


java/kotlin-extractor/src/main/kotlin/KotlinUsesExtractor.kt
```diff
@@ -990,12 +990,12 @@ open class KotlinUsesExtractor(
...
     */
     @OptIn(ObsoleteDescriptorBasedAPI::class)
-    fun getFunctionLabel(f: IrFunction, maybeParentId: Label<out DbElement>?, classTypeArgsIncludingOuterClasses: List<IrTypeArgument>?) =
+    fun getFunctionLabel(f: IrFunction, maybeParentId: Label<out DbElement>?, classTypeArgsIncludingOuterClasses: List<IrTypeArgument>?, maybeParameterList: List<IrValueParameter>? = null) =
         getFunctionLabel(
             f.parent,
             maybeParentId,
             getFunctionShortName(f).nameInDB,
-            f.valueParameters,
+            maybeParameterList ?: f.valueParameters,
```
igfoo, line 998: Can you add an explanation of when/why `maybeParameterList` is passed in to the function's comment please?

smowton: Done

java/ql/integration-tests/posix-only/kotlin/jvmoverloads_flow/test.kt
```diff
@@ -0,0 +1,78 @@
...
+}
+ 
+public class GenericConstructorDoesNotTaintByDefault<T> @JvmOverloads constructor(before: T, s: String = "no taint", after: T, after2: T? = null) { 
+
+  init {
+    sink(s)
+  }
+
+}
+
```
tamasvajk, line 78: There's no test case added for any of the above methods being called from Kotlin. And as we discussed it on Slack, I think the calls are not working as expected.

Adding the following test case results in no new findings, but I think two new flows should be found.
```
class TestClass {
    companion object {
        @JvmOverloads @JvmStatic
        fun noTaintByDefault1(s1: String = "no taint", s2: String = "no taint") { sink(s2) }
    }

    fun noTaintByDefault2(s1: String = "no taint", s2: String = "no taint") { sink(s2) }

    fun fn() {
        noTaintByDefault1(s2 = source())
        noTaintByDefault2(s2 = source())
    }
}
```

tamasvajk (changes requested):

java/ql/integration-tests/posix-only/kotlin/jvmoverloads_flow/test.py
```diff
@@ -0,0 +1,4 @@
+from create_database_utils import *
+
+os.mkdir('kbuild')
+run_codeql_database_create(["kotlinc test.kt -d kbuild", "javac User.java -cp kbuild"], lang="java")
```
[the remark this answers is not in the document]
smowton: Done

java/ql/integration-tests/posix-only/kotlin/jvmoverloads_flow/test.kt
smowton: Indeed, I'm intentionally handling only the JvmOverloads / Java interop aspect here and will follow up to address calls from within other Kotlin code

tamasvajk (changes requested):
I added some minor comments. The PR looks generally okay, but some tests are still failing.

Also, I think we'll have problems with introducing the new parameters with type `IrValueParameterWithOverriddenIndex`. They have incorrect parents, so we might generate incorrect label for them. I haven't checked it, but in case a parameter is accessed in a default argument value, `extractExpressionExpr` doesn't have the `overloadId`:
```
fun fn(s0: String = "", s1: String = fn(s0, ""), s2: String = "") = ""
```
Maybe storing the `overloadId` in `IrValueParameterWithOverriddenIndex` would help.

smowton:
@tamasvajk now that I've had to use a piece of background state to cope with type variable references, I've changed the strategy for handling value parameters to use the same mechanism instead of creating a proxy class that wraps `IrValueParameter`. I've also converted the remaining integration test into a unit test.

java/kotlin-extractor/src/main/kotlin/KotlinFileExtractor.kt
```diff
@@ -4814,12 +4921,20 @@ open class KotlinFileExtractor(
         }
     }
 
-    private inner class DeclarationStackAdjuster(declaration: IrDeclaration): Closeable {
+    private inner class DeclarationStackAdjuster(val declaration: IrDeclaration, val overriddenAttributes: OverriddenFunctionAttributes? = null): Closeable {
```
tamasvajk, line 4926: It would probably be cleaner to have another constructor, which takes an `IrFunction` and an `OverriddenFunctionAttributes`.

java/kotlin-extractor/src/main/kotlin/KotlinUsesExtractor.kt
```diff
@@ -36,9 +40,12 @@ open class KotlinUsesExtractor(
     val externalClassExtractor: ExternalDeclExtractor,
     val primitiveTypeMapping: PrimitiveTypeMapping,
     val pluginContext: IrPluginContext,
+    val overriddenFunctionAttributes : HashMap<IrDeclaration, OverriddenFunctionAttributes>,
```
tamasvajk, line 43: Couldn't this be an `HashMap<IrFunction, ...` ?

java/kotlin-extractor/src/main/kotlin/KotlinUsesExtractor.kt
```diff
@@ -1369,6 +1380,9 @@ open class KotlinUsesExtractor(
 
     fun getTypeParameterParentLabel(param: IrTypeParameter) =
         param.parent.let {
+            (it as? IrDeclaration)?.let { decl ->
+                overriddenFunctionAttributes[decl]?.id
+            } ?:
```
tamasvajk, line 1385: Doesn't this and all other `overriddenFunctionAttributes` uses in `KotlinUsesExtractor` assume that when the access is made, we're extracting the given function with overrides? So in this case, are we sure that `getTypeParameterParentLabel` is only called when `overriddenFunctionAttributes[decl]` is being extracted? 

I think, you might be right that `getTypeParameterParentLabel` and `getValueParameterLabel` are only called within `forceExtractFunction` and `extractGeneratedOverload`, but it feels somewhat fragile. Isn't this the same as `(this as KotlinFileExtractor).declarationStack.peek().overriddenFunctionAttributes`, which doesn't make sense at the moment, because the attributes are not on the stack, but if they were, we would consider this dangerous because of the cast.

java/kotlin-extractor/src/main/kotlin/KotlinFileExtractor.kt
tamasvajk, line 4926: Wouldn't it make more sense to push `(declaration, overriddenAttributes)` onto the stack? That way we wouldn't have to maintain two sets of states (`overriddenFunctionAttributes` and `declarationStack`)?

java/kotlin-extractor/src/main/kotlin/KotlinUsesExtractor.kt
```diff
@@ -1506,13 +1520,18 @@ open class KotlinUsesExtractor(
      */
     fun getValueParameterLabel(vp: IrValueParameter, parent: Label<out DbCallable>?): String {
         val declarationParent = vp.parent
-        val parentId = parent ?: useDeclarationParent(declarationParent, false)
+        val overriddenParentAttributes = (declarationParent as? IrDeclaration)?.let {
+            overriddenFunctionAttributes[it]
+        }
+        val parentId = parent ?: overriddenParentAttributes?.id ?: useDeclarationParent(declarationParent, false)
```
tamasvajk, line 1526: Would this work with recursive calls too? `overriddenFunctionAttributes` only has a reference to one overload at a time. Could we need multiple at any time? For example when generating the overload `f(String,String)` would we need access to both `f(String,String)` and `f(String)`?
```
object X {
    @JvmStatic
    @JvmOverloads
    fun f(s0: String = f("", ""), s1: String = f(""), s2: String = "") = ""
}
```

tamasvajk (commented):
Added some questions for discussion.

java/kotlin-extractor/src/main/kotlin/KotlinUsesExtractor.kt
tamasvajk: Answering this to myself: I think this question doesn't make sense. We're accessing parameters here, I don't think it's possible to access parameters of multiple overloads, we can at most access parameters that are declared earlier in the parameter list, and those will come from the current declaration.

java/kotlin-extractor/src/main/kotlin/KotlinUsesExtractor.kt
smowton: For value parameters: I'd expect this could be seen from inside capturing local declarations and classes too. Something like

```
fun f(x: Int, y: Int = x.let {
  fun captures() = it
  captures()
})
```

For type parameters: these definitely can be seen from elsewhere, principally when extracting instantiated generic types that occur in the signature of a method with overloads. So when we extract `fun <T> f(x: List<T>, y: T? = null)` then we need to generate `List<T>` with `T = f/1::T` as well as the case with `T = f/2::T` that gets created for the "normal" version of `f`. That's why the overridden functions map gets passed to child instances of KotlinUses/FileExtractor.

java/kotlin-extractor/src/main/kotlin/KotlinFileExtractor.kt
smowton: I separated them because I needed the function override state to be propagated when extracting generic type instantiations as mentioned above and didn't want to get into the decl stack being non-empty on completing extraction of that type, but I can keep them together if that's preferred at the slight cost of searching the stack.

smowton:
@tamasvajk I've adopted your suggestion to store the overridden attributes in the declaration stack

java/kotlin-extractor/src/main/kotlin/KotlinUsesExtractor.kt
```diff
@@ -36,9 +38,27 @@ open class KotlinUsesExtractor(
     val externalClassExtractor: ExternalDeclExtractor,
     val primitiveTypeMapping: PrimitiveTypeMapping,
     val pluginContext: IrPluginContext,
+    val declarationStack: DeclarationStack,
     val globalExtensionState: KotlinExtractorGlobalState
 ) {
 
+    class DeclarationStack {
```
tamasvajk, line 45: I find it somewhat odd that the declaration stack is in the `UsesExtractor`. In my mind it belongs to the file extractor because that's where we extract the declarations.

java/kotlin-extractor/src/main/kotlin/KotlinUsesExtractor.kt
```diff
@@ -1506,13 +1533,18 @@ open class KotlinUsesExtractor(
      */
     fun getValueParameterLabel(vp: IrValueParameter, parent: Label<out DbCallable>?): String {
         val declarationParent = vp.parent
-        val parentId = parent ?: useDeclarationParent(declarationParent, false)
+        val overriddenParentAttributes = (declarationParent as? IrFunction)?.let {
```
tamasvajk, line 1536: If we kept the declaration stack in the file extractor, we could use the below:
```
val overriddenParentAttributes = if (this is KotlinFileExtractor && this.filePath == vp.file.path && declarationParent is IrFunction) {
  this.declarationStack.findOverriddenAttributes(declarationParent)
} else {
  null
}
```
I find this more explicit in what our expectations are: the overridden attributes should only be used if we're inside the extraction of the corresponding function declaration.

tamasvajk (approved):
I've added some refactoring ideas, feel free to ignore them. 

I think the PR is in a mergeable state, we know there are some missing cases, which will be covered in upcoming PRs. We could also apply the refactorings later.

java/kotlin-extractor/src/main/kotlin/KotlinUsesExtractor.kt
tamasvajk: Let me know if this is a wrong assumption.

java/kotlin-extractor/src/main/kotlin/KotlinFileExtractor.kt
```diff
@@ -4950,12 +4942,20 @@ open class KotlinFileExtractor(
...
```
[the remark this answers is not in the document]
igfoo: Don't worry about that for now. We can revisit this if/when that happens.

java/kotlin-extractor/src/main/kotlin/KotlinUsesExtractor.kt
tamasvajk: `vp.file.path` can throw, so the correct check would be `vp.fileOrNull?.path`. or `declarationParent .fileOrNull?.path`.

I think adding the check would make it explicit that we only expect to find anything by `findOverriddenAttributes` if the files match. It would only make the intent more explicit, but it wouldn't change the behaviour. Also, performance wise having the check or not probably doesn't matter, because both `fileOrNull` and `findOverriddenAttributes` walks a parent-child declaration list linearly.

tamasvajk (commented):
Added some minor comments, otherwise it looks good to me.

java/kotlin-extractor/src/main/kotlin/KotlinUsesExtractor.kt
```diff
@@ -128,18 +127,24 @@ open class KotlinUsesExtractor(
             return this
         }
 
+        val newDeclarationStack =
+            if (this is KotlinFileExtractor)
+                this.declarationStack
+            else
```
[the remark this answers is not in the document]
smowton: This actually does need to be the same stack for the case of type variables -- for example, when extracting the `List<T>` (and its downstream types, `Iterator<T>` etc) in the context of synthetic `fun <A> f(a: List<A>)` related to the real function `fun <B> f(a: List<B>, b: Int = 0)`, we need to substitute `B -> A` everywhere, and the stack of outstanding function extractions is used to find the enclosing extraction of the relevant function declaring the type variable.

java/kotlin-extractor/src/main/kotlin/KotlinUsesExtractor.kt
smowton: ok done

java/kotlin-extractor/src/main/kotlin/KotlinUsesExtractor.kt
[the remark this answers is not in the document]
tamasvajk: I see, thanks for the explanation.

java/kotlin-extractor/src/main/kotlin/KotlinUsesExtractor.kt
tamasvajk: The same should be done in `getTypeParameterParentLabel` just to keep the symmetry between the two cases.

smowton: Not done, because this is the case where we do need to look for type-parameter substitution happening up the stack, even in a different file -- for example, due to generic type specialisations transitively required from a synthetic function's parameter or return types.

tamasvajk: Makes sense. I always forget about this special case.

tamasvajk (approved):

PR #9812, opened 2022-07-12
C++: special case GVN for `std::vector::size`

rdmarsh2:
This PR adds value numbering for the `std::vector::size` function, based on the value of the `this` pointer and object. This fixes some false positives in an experimental range analysis query.

Opened as draft for discussion - I intend to generalize this with either a model or some heuristics before merging.


cpp/ql/test/library-tests/valuenumbering/GlobalValueNumbering/test.cpp
```diff
@@ -156,4 +156,27 @@ void test_read_global_different(int n) {
...
+      data = x;
+    }
+  };
+}
+
+void test_vector_size(std::vector v) {
+  int a = v.size();
+  int b = v.size();
+  v.set(10);
+  int c = v.size();
```
geoffw0, line 181: I'm not exactly sure how it works, but I'm impressed we match the result of `size()` on line 178 to line 179, but _not_ 181.

How easy would it be to fool it, by doing something subtler that changes the number of elements in `v`?

geoffw0: Sounds like the right call to me.

PR #9815, opened 2022-07-13
Swift: trap output rework

redsun82:
Firstly, this change reworks how inter-process races are resolved.
Moreover some responsability reorganization has led to merging
`TrapArena` and `TrapOutput` again into a `TrapDomain` class.
Finally, `Argument` has been made `Locatable`.

A `TargetFile` class is introduced, that is successfully created
only for the first process that starts processing a given trap output
file. From then on `TargetFile` simply wraps around `<<` stream
operations, dumping them to a temporary file. When `TargetFile::commit`
is called, the temporary file is moved on to the actual target trap
file.

Processes that lose the race can now just ignore the unneeded
extraction and go on, while previously all processes would carry out
all extractions overwriting each other at the end.

Some of the file system logic contained in `SwiftExtractor.cpp` has been
moved to this class, and two TODOs are solved:
* introducing a better inter process file collision avoidance strategy
* better error handling for trap output operations: if unable to write
  to the trap file (or carry out other basic file operations), we just
  abort.

The changes to `ExprVisitor` and `StmtVisitor` are due to wanting to
hide the raw `TrapDomain::createLabel` from them, and bring more
funcionality under the generic caching/dispatching mechanism.


swift/extractor/infra/TargetFile.h
```diff
@@ -0,0 +1,40 @@
...
+// `targetDir/target` only when `commit()` is called
+class TargetFile {
+  std::string workingPath;
+  std::string targetPath;
+  std::ofstream out;
+
+ public:
+  TargetFile(std::string_view target, std::string_view targetDir, std::string_view workingDir);
+
+  bool good() const;
```
sashabu, line 25: I think we can simplify the API:
1. Replace `commit()` with a destructor. This:
   * means not having to remember to call `commit()` and
   * makes it impossible to keep using a committed file.
2. Don't make callers check `good()` (and when we use it internally, it _might_ become clearer to just inline it). Instead make goodness an invariant (from the user's perspective):
   * Make the constructor private, and add a public factory function that returns an `optional`, with the invariant that the factory either returns a `TargetFile` with an opened file or `nullopt`. This makes it clear that the caller should check for presence.
   * Replacing `commit()` with a destructor means that once you have a `TargetFile` with an open file, that file stays open as long as the object does (RAII).
   * For other, unexpected failures, we can (and already do) handle them in `TargetFile` (by logging an error and aborting).

(As always, my review comments are open to discussion if you disagree with the suggestion!)

sashabu (changes requested):

swift/extractor/infra/TargetFile.h
redsun82: I agree with the factory function, but while I initially went with the destructor approach, I went for the explicit `commit` because it is what we actually want (for now). The temporary trap file was introduced because extractor crashes were causing problems: I don't remember if this was related to the known issue of the trap importer not liking incomplete trap files, or if we just did not want to have an inconsitent DB. In any case, we wanted the trap to be dumped only on successful execution, which would not be the case were we to commit on destruction.

We can definitely revisit that, but maybe in a separate PR, I'd keep that behaviour in this one. In any case, we should wait for @AlexDenisov because he might know more about this.

sashabu: > if this was related to the known issue of the trap importer not liking incomplete trap files

I believe it was that. I won't block the PR on it, but regarding:

> In any case, we wanted the trap to be dumped only on successful execution, which would not be the case were we to commit on destruction.

`std::abort` [does not call](https://en.cppreference.com/w/cpp/utility/program/abort) destructors (and obviously a crash doesn't either). Is there a scenario where we don't finish successfully, but do invoke the destructor?

> I agree with the factory function

I think this it's worth using a factory function and hiding `good()` from users regardless of the `commit()` changes. Does your agreement mean you plan to make that change in this PR?

redsun82: > `std::abort` [does not call](https://en.cppreference.com/w/cpp/utility/program/abort) destructors (and obviously a crash doesn't either). Is there a scenario where we don't finish successfully, but do invoke the destructor?

uh, I honestly thought it did, should have double checked :slightly_smiling_face: And `assert` which we currently use also calls `std::abort`, so I agree we can move `commit` to the destructor.

> Does your agreement mean you plan to make that change in this PR?

Yes :slightly_smiling_face:

redsun82: done :+1:

swift/extractor/infra/TargetFile.h
```diff
@@ -2,28 +2,29 @@
...
+  static std::optional<TargetFile> create(std::string_view target,
+                                          std::string_view targetDir,
+                                          std::string_view workingDir);
 
-  bool good() const;
-  void commit();
+  ~TargetFile() { commit(); }
+
+  TargetFile(TargetFile&& other) = default;
+  TargetFile& operator=(TargetFile&& other);
```
sashabu, line 27: Oof, I hadn't considered we'd need all this.

My understanding is that the clever move assignment operator never gets used. If that's correct, can we delete it so we don't have to think about the non-trivial implementation? That's the approach used by [`gsl::final_action`](https://github.com/microsoft/GSL/blob/d9fc52e20e32fcc804502480a781120b210afd41/include/gsl/util#L74). Having a move constructor but not move assignment is weird, but if we add a comment that it wasn't added simply because it's unused, I think that should be okay.

(Incidentally, using `gsl::final_action` would have made the destructor trivial too, but I don't think it's worth switching to it.)

swift/extractor/infra/TargetFile.cpp
```diff
@@ -36,31 +41,49 @@ std::string initPath(std::string_view target, std::string_view dir) {
...
+  }
+  // else we just lost the race
+  return false;
+}
+
+std::optional<TargetFile> TargetFile::create(std::string_view target,
+                                             std::string_view targetDir,
+                                             std::string_view workingDir) {
+  TargetFile ret{target, targetDir, workingDir};
+  if (ret.init()) return {std::move(ret)};
```
sashabu, line 66: Nit: We can just `return ret;` - the braces are redundant, and named return values are automatically moved from.

sashabu (approved):
I think `delete`-ing `TargetFile` move assignment is worth doing in this PR, but also happy to approve as is.

swift/extractor/infra/TargetFile.h
redsun82: yeah, I was considering it, weighing in the rule of five and whatnot, if you agree I'm happy to delete the move assignment, as I was almost about to

sashabu (approved):

PR #9817, opened 2022-07-13
Java: Model `java.util.Properties.getProperty`

smowton:


intrigus-lgtm:
There are also special `put` methods adapted for the `Properties` class, would it make sense to model them as well, like this:
```ql
        "java.util;Properties;true;setProperty;(String,String);;Argument[-1].MapValue;ReturnValue;value;manual",
        "java.util;Properties;true;setProperty;(String,String);;Argument[0];Argument[-1].MapKey;value;manual",
        "java.util;Properties;true;setProperty;(String,String);;Argument[1];Argument[-1].MapValue;value;manual",
```
?

smowton:
So there is -- mind making that into a PR?

intrigus-lgtm:
> So there is -- mind making that into a PR?

Done in #9963

PR #9818, opened 2022-07-13
Swift: cache file paths

redsun82:
This required a bit of a generalization of `TrapLabelStore` to not
work only with pointers.


AlexDenisov (approved):
:shipit:

AlexDenisov (approved):

PR #9823, opened 2022-07-14
Dataflow: Replace stage duplication with parameterised modules.

aschackmull:
Appears to work, although there are some complaints of the form
> Some errors in uninstantiated parameterised modules are not (yet) reported, so non-trivial binding sets on signatures can introduce errors that only future compiler versions will be able to detect

@ginsbach

### DIFF
--- a/java/ql/lib/semmle/code/java/dataflow/internal/DataFlowImpl.qll
+++ b/java/ql/lib/semmle/code/java/dataflow/internal/DataFlowImpl.qll
@@ -597,7 +597,7 @@ private predicate hasSinkCallCtx(Configuration config) {
   )
 }
 
-private module Stage1 {
+private module Stage1 implements StageSig {
   class ApApprox = Unit;
 
   class Ap = Unit;
@@ -944,12 +944,9 @@ private module Stage1 {
   predicate revFlow(NodeEx node, Configuration config) { revFlow(node, _, config) }
 
   bindingset[node, state, config]
-  predicate revFlow(
-    NodeEx node, FlowState state, boolean toReturn, ApOption returnAp, Ap ap, Configuration config
-  ) {
-    revFlow(node, toReturn, pragma[only_bind_into](config)) and
+  predicate revFlow(NodeEx node, FlowState state, Ap ap, Configuration config) {
+    revFlow(node, _, pragma[only_bind_into](config)) and
     exists(state) and
-    exists(returnAp) and
     exists(ap)
   }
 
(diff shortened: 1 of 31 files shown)


ginsbach:
> Appears to work, although there are some complaints of the form
> 
> > Some errors in uninstantiated parameterised modules are not (yet) reported, so non-trivial binding sets on signatures can introduce errors that only future compiler versions will be able to detect

The removal of that error message is currently still blocked by https://github.com/github/semmle-code/pull/42236. Once we have the frontend binding set checker, we can remove the warning and proceed with this.

ginsbach:
The blocking warning should be gone, now that we merged https://github.com/github/semmle-code/pull/43179 into main.

java/ql/lib/semmle/code/java/dataflow/internal/DataFlowImpl.qll
```diff
@@ -1142,867 +1139,647 @@ private predicate flowIntoCallNodeCand1(
...
-  class ApOption = BooleanOption;
+  signature module StageParam {
+    class Ap;
 
-  ApOption apNone() { result = TBooleanNone() }
+    class ApNil extends Ap;
 
-  ApOption apSome(Ap ap) { result = TBooleanSome(ap) }
+    bindingset[result, ap]
+    ApApprox getApprox(Ap ap);
```
hvitved, line 1171: We could add a ``// TODO: member predicate on `Ap` `` comment.

java/ql/lib/semmle/code/java/dataflow/internal/DataFlowImpl.qll
```diff
@@ -1142,867 +1139,647 @@ private predicate flowIntoCallNodeCand1(
...
 
-  class Cc = CallContext;
+    ApNil getApNil(NodeEx node);
 
-  class CcCall = CallContextCall;
+    bindingset[tc, tail]
+    Ap apCons(TypedContent tc, Ap tail);
 
-  class CcNoCall = CallContextNoCall;
+    Content getHeadContent(Ap ap);
```
hvitved, line 1178: We could add a ``// TODO: member predicate on `Ap` `` comment.

java/ql/lib/semmle/code/java/dataflow/internal/DataFlowImpl.qll
```diff
@@ -2011,617 +1788,430 @@ private module Stage3 {
...
+    checkCallContextReturn(innercc, c, call) and
+    if reducedViableImplInReturn(c, call) then result = TReturn(c, call) else result = ccNone()
   }
+}
 
-  bindingset[result, apa]
-  private ApApprox unbindApa(ApApprox apa) {
-    pragma[only_bind_out](apa) = pragma[only_bind_out](result)
-  }
+private module Stage2Param implements MkStage<Stage1>::StageParam {
```
hvitved, line 1856: We have two calls to `MkStage<Stage1>` (and same for other stages), here and in the module `Stage2`. Perhaps that is OK, as parameterized modules are not generative?

hvitved (commented):
LGTM. Please do a DCA run for all languages as well.

java/ql/lib/semmle/code/java/dataflow/internal/DataFlowImpl.qll
aschackmull: That is indeed perfectly fine.

aschackmull:
DCA looks good, I think.

hvitved (approved):
💪

jketema:
> DCA looks good, I think.

Agreed for C++. The vim failure is expected (we're working on it).

PR #9825, opened 2022-07-14
QL: add ql/repeated-word query

erik-krogh:
Inspired by [this comment](https://github.com/github/codeql/pull/9822#discussion_r920983468).  (Even though that's an `.md` file, which we won't detect).  

I also fixed all the alerts (as per usual).


tausbn (approved):
QL and Python stuff LGTM. :+1:

geoffw0 (approved):
I'm a bit surprised this doesn't require a whitelist of acceptable repeat words.  But the results speak for themselves. :+1:

cpp/ql/lib/semmle/code/cpp/dataflow/internal/DataFlowUtil.qll
```diff
@@ -699,7 +699,7 @@ private predicate exprToExprStep_nocfg(Expr fromExpr, Expr toExpr) {
         call.getTarget() = f and
         // AST dataflow treats a reference as if it were the referred-to object, while the dataflow
         // models treat references as pointers. If the return type of the call is a reference, then
-        // look for data flow the the referred-to object, rather than the reference itself.
+        // look for data flow the referred-to object, rather than the reference itself.
```
intrigus-lgtm, line 702: Shouldn't this be like this?
```suggestion
        // look for data flow to the referred-to object, rather than the reference itself.
```

michaelnebel (dismissed):
C# LGTM
QL for QL looks plausible to me :-)

atorralba (approved):
Java 👍

erik-krogh:
Merged with `main` to resolve a conflict.  
I'll merge on the next approving review. (@github/codeql-ruby 👀 ?).

PR #9827, opened 2022-07-14
QL: Query for detecting unused parameter in override methods

erik-krogh:
All current results are benign because the type of the unmentioned parameter is small.  
But it is very hard to statically determine that the  type is small. (E.g. one is an abstract class, where all (transitive) overrides are small).  

The fix to remove the alert for the benign results is just to insert `exists(param)` instead of `any()`.   

I did try to remove the `pred.hasAnnotation("override")` restriction, but that just gave me more noise.

### DIFF
--- a/ql/ql/src/codeql_ql/performance/VarUnusedInDisjunctQuery.qll
+++ b/ql/ql/src/codeql_ql/performance/VarUnusedInDisjunctQuery.qll
@@ -1,0 +1,27 @@
+import ql
+
+/**
+ * Holds if we assume `t` is a small type, and
+ * variables of this type are therefore not an issue in cartesian products.
+ */
+predicate isSmallType(Type t) {
+  t.getName() = "string" // DataFlow::Configuration and the like
+  or
+  exists(NewType newType | newType = t.getDeclaration() |
+    forex(NewTypeBranch branch | branch = newType.getABranch() | branch.getArity() = 0)
+  )
+  or
+  t.getName() = "boolean"
+  or
+  exists(NewType newType | newType = t.getDeclaration() |
+    forex(NewTypeBranch branch | branch = newType.getABranch() |
+      isSmallType(branch.getReturnType())
+    )
+  )
+  or
+  exists(NewTypeBranch branch | t = branch.getReturnType() |
+    forall(Type param | param = branch.getParameterType(_) | isSmallType(param))
+  )
+  or
+  isSmallType(t.getASuperType())
+}
--- a/ql/ql/src/queries/style/OverrideAny.ql
+++ b/ql/ql/src/queries/style/OverrideAny.ql
@@ -1,0 +1,40 @@
+/**
+ * @name Override with unmentioned parameter
+ * @description A predicate that overrides the default behavior but doesn't mention a parameter is suspicious.
+ * @kind problem
+ * @problem.severity warning
+ * @id ql/override-any
+ * @precision very-high
+ */
+
+import ql
+import codeql_ql.performance.VarUnusedInDisjunctQuery
+
+AstNode param(Predicate pred, string name, Type t) {
+  result = pred.getParameter(_) and
+  result.(VarDecl).getName() = name and
+  result.(VarDecl).getType() = t
+  or
+  result = pred.getReturnTypeExpr() and
+  name = "result" and
+  t = pred.getReturnType()
+}
+
+predicate hasAccess(Predicate pred, string name) {
+  exists(param(pred, name, _).(VarDecl).getAnAccess())
+  or
+  name = "result" and
+  exists(param(pred, name, _)) and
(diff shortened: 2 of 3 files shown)


ql/ql/src/codeql_ql/performance/VarUnusedInDisjunctQuery.qll
```diff
@@ -0,0 +1,27 @@
+import ql
+
+/**
+ * Holds if we assume `t` is a small type, and
+ * variables of this type are therefore not an issue in cartesian products.
+ */
+predicate isSmallType(Type t) {
+  t.getName() = "string" // DataFlow::Configuration and the like
```
kaeluka, line 8: Why are strings a small type? Small, here, means "low cardinality", right?

Edit: I see you've only moved a pre-existing predicate, but I'm still curious

ql/ql/src/codeql_ql/performance/VarUnusedInDisjunctQuery.qll
```diff
@@ -0,0 +1,27 @@
...
+  or
+  exists(NewType newType | newType = t.getDeclaration() |
+    forex(NewTypeBranch branch | branch = newType.getABranch() | branch.getArity() = 0)
```
kaeluka, line 11: Here, and in the recursive case two disjuncts below, why'd you need `forex`, rather than `forall`?

ql/ql/src/queries/style/OverrideAny.ql
```diff
@@ -0,0 +1,40 @@
+/**
+ * @name Override with unmentioned parameter
+ * @description A predicate that overrides the default behavior but doesn't mention a parameter is suspicious.
+ * @kind problem
+ * @problem.severity warning
+ * @id ql/override-any
+ * @precision very-high
```
kaeluka, line 7: Since the warnings are flagging benign cases, is precision `very-high` appropriate?

ql/ql/src/queries/style/OverrideAny.ql
```diff
@@ -0,0 +1,40 @@
...
+from Predicate pred, AstNode param, string name, Type paramType
+where
+  pred.hasAnnotation("override") and
+  param = param(pred, name, paramType) and
+  not hasAccess(pred, name) and
+  not pred.getBody() instanceof NoneCall and
+  exists(pred.getBody()) and
+  not isSmallType(pred.getParent().(Class).getType()) and
+  not isSmallType(paramType)
+select pred, "Override predicate doesn't mention $@.", param, name
```
kaeluka, line 40: Can you hint that `exists(..)` is a way to address the benign warnings in the error message?

kaeluka (approved):
I'm not sure if the `very-high` precision makes sense, since the query (in practice) tends to find benign cases. Alternatively, perhaps have the message explain the problem a bit better.

Other than that, I'd only improve the error message a bit to make the fix a bit easier.

The warnings found look good to me (JS and others).

Giving an approve as most of this is quite opinionated and I want the decision to be your's :+1:

ql/ql/src/codeql_ql/performance/VarUnusedInDisjunctQuery.qll
erik-krogh: `string` is often a `Configuration` like class, like `DataFlow::Configuration`, and those are unused in a whole lot of predicates.

ql/ql/src/codeql_ql/performance/VarUnusedInDisjunctQuery.qll
erik-krogh: I know that a branch must exists (so `forall` and `forex` are equivalent).  
And `forex` tend to avoid some horrible join-orders that can sometimes appear with `forall`, so I usually go for `forex` when they are equivalent.

ql/ql/src/queries/style/OverrideAny.ql
erik-krogh: Maybe not, but it doesn't really matter.  
If we get FP reports (I doubt it), then we can fix it.

erik-krogh: I lowered it to `high` since I touched the query anyway.

ql/ql/src/codeql_ql/performance/VarUnusedInDisjunctQuery.qll
kaeluka: 👍

ql/ql/src/codeql_ql/performance/VarUnusedInDisjunctQuery.qll
kaeluka: 🤷 ok! 👍

kaeluka (approved):
LGTM!

PR #9829, opened 2022-07-14
Kotlin: Don't extract a name for a '_' parameter

smowton:
I can't reproduce the exact circumstances, but these sometimes get "<anonymous parameter X>" names and sometimes get "$noName_X" names. Whichever way, avoiding extracting a synthetic name seems safest; anyone finding the .class file and not reading the metadata indicating it came from a `_` will extract the binary name selected, or else QL will invent a name.


igfoo:
Looks plausible, but it fails to build with 1.4.32:
```
2022-07-14T15:49:55.6699226Z ##[error]build_standalone_1.4.32/temp_src/main/kotlin/KotlinFileExtractor.kt:645:76: error: unresolved reference: UNDERSCORE_PARAMETER
2022-07-14T15:49:55.6701060Z             val syntheticParameterNames = vp.origin == IrDeclarationOrigin.UNDERSCORE_PARAMETER || ((vp.parent as? IrFunction)?.let { hasSynthesizedParameterNames(it) } ?: true)
```

smowton:
@igfoo now implemented for all supported versions.

igfoo (dismissed):

smowton:
@igfoo fix pushed

tamasvajk (approved):
LGTM

PR #9830, opened 2022-07-14
Kotlin: annotation properties should be java.lang.Class not KClass

smowton:
As documented at https://kotlinlang.org/docs/annotations.html#constructors, annotation properties of type KClass get rewritten when targeting the JVM.


tamasvajk (commented):
In the below example, we're extracting `java.lang.Class<?>` as the type of `arg1`. What is the type of `String::class`? Is that type assignable to `java.lang.Class<?>`? Does QL assume that the two types are compatible?

```kotlin
import kotlin.reflect.KClass
annotation class Ann(val arg1: KClass<*>, val arg2: KClass<out Any>)
@Ann(String::class, Int::class) class MyClass
```

smowton:
That probably is an issue; I'll take another pass at this once I'm done with Java extractors

smowton:
I've updated this to fix expression types in the annotation class' constructor and to handle arrays of classes. However I haven't addressed the expressions at annotation use site because we currently don't handle annotation uses at all (i.e., we never populate the `annotValue` table). We'll have to remember to fix this when we get around to that.

tamasvajk:
> ```kotlin
> import kotlin.reflect.KClass
> annotation class Ann(val arg1: KClass<*>, val arg2: KClass<out Any>)
> @Ann(String::class, Int::class) class MyClass
> ```

To remember this, I added this case to the issue tracking annotation extraction.

tamasvajk:
> I've updated this to fix expression types in the annotation class' constructor and to handle arrays of classes. However I haven't addressed the expressions at annotation use site because we currently don't handle annotation uses at all (i.e., we never populate the `annotValue` table). We'll have to remember to fix this when we get around to that.

To remember this, I added this case to the issue tracking annotation extraction.

tamasvajk (approved):
Looks reasonable to me.

PR #9831, opened 2022-07-14
Docs: Update supported languages page with links to CLI and pack information

adityasharad:
Include links to the CLI changelog, CLI releases, bundle releases, pack changelogs, and pack source.

Clarify that this support information applies to the current version of the CLI, bundle, query packs, and library packs.

Reviewer notes:
- I'm slightly unsure how often to mention the CodeQL bundle. We should mention it somewhere, but it feels like too much to mention it everywhere we mention the query or library packs. I've compromised by mentioning it only where we mention the CLI.

### DIFF
--- a/docs/codeql/codeql-overview/supported-languages-and-frameworks.rst
+++ b/docs/codeql/codeql-overview/supported-languages-and-frameworks.rst
@@ -11,14 +11,17 @@ CodeQL.
 Languages and compilers
 #######################
 
-CodeQL supports the following languages and compilers.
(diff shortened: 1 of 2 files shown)


mchammer01:
Docs first responder here 👋🏻 - thanks for creating this PR ✨ 
I've put it up for review on our board.

felicitymay (approved):
Thanks @adityasharad - I don't think we need to keep mentioning the bundle. This seems fine.

It does make me wonder when we'll be adding information for Ruby, but I'm guessing that will be with the GA of analysis later in the year.

PR #9837, opened 2022-07-15
Move definitions.ql back to src

aeisenberg:
LGTM requires the definitions.ql files to be in the src, not the lib pack.


adityasharad (approved):
LGTM. Will merge after we make an internal decision on whether to get this into 2.10.1 or main/2.10.2.

criemen:
Merging, as we decided that this will go into main.

PR #9838, opened 2022-07-15
Move python contextual queries to lib folders

aeisenberg:
This will ensure that python projects can use jump to ref/def in
vscode when the core libraries are not installed.

See also #9744. This PR does the same for python queries.


tausbn (approved):
I'm happy with the move, but slightly confused as to the setup here. This PR only moves the `Local...` versions, leaving e.g. `Definitions.ql` in `src/analysis/` (unless I've misread this somehow). Presumably `Definitions.ql` should also move into `/lib`?

aeisenberg:
Thanks for the review. Definitions.ql must remain in src in order for lgtm to be able to discover them. Also, ql files in *-all qlpacks are not pre-compiled.

aeisenberg:
That comment was written from my phone. Here is a deeper explanation:

The vscode extension requires the LocalDefs/Refs files to be in the lib folders (ie- `*-all` qlpacks) so that when the core libraries are not in the workspace, the appropriate contextual queries can still be resolved. This is because we can presume that if you are working with python queries, you will have `codeql/python-all` transitively available from the qlpacks in the workspace.

For LGTM, we need the Definitions.qll in the `*-queries` qlpacks because these are the only packs that LGTM can resolve. Also, if we _did_ change LGTM so that it can resolve `*-all` qlpacks, we would still have a problem in that these queries are not pre-compiled in library packs.

PR #9845, opened 2022-07-18
Ruby: skip .git folder

aibaars:
CodeQL should not scan the `.git` folder when looking for Ruby source files. Most of the time scanning the `.git` folder is harmless, but it may cause syntax errors when some branch names happen to end with `.rb`.


alexrford (approved):
👍

PR #9848, opened 2022-07-18
Swift: More improvements for the string length conflation query

geoffw0:
More improvements for the string length conflation query:
 - added a test for a potential issue with implementations of `NSString` that has been discussed before, where member functions are defined in a parent class rather than directly in `NSString` itself.  At the moment I believe we can't *fix* this issue because we don't have information about base classes, but at least now it is documented with a test.
 - clean up the QL for `isSink` (less code duplication).

### DIFF
--- a/swift/ql/src/queries/Security/CWE-135/StringLengthConflation.ql
+++ b/swift/ql/src/queries/Security/CWE-135/StringLengthConflation.ql
@@ -41,88 +41,87 @@ class StringLengthConflationConfiguration extends DataFlow::Configuration {
   }
 
   override predicate isSink(DataFlow::Node node, string flowstate) {
-    // arguments to method calls...
-    exists(
-      string className, string methodName, string paramName, ClassDecl c, AbstractFunctionDecl f,
-      CallExpr call, int arg
-    |
-      (
-        // `NSRange.init`
-        className = "NSRange" and
-        methodName = "init(location:length:)" and
-        paramName = ["location", "length"]
-        or
-        // `NSString.character`
-        className = ["NSString", "NSMutableString"] and
-        methodName = "character(at:)" and
-        paramName = "at"
-        or
-        // `NSString.character`
-        className = ["NSString", "NSMutableString"] and
-        methodName = "substring(from:)" and
-        paramName = "from"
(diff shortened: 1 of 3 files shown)


swift/ql/src/queries/Security/CWE-135/StringLengthConflation.ql
```diff
@@ -41,88 +41,80 @@ class StringLengthConflationConfiguration extends DataFlow::Configuration {
...
-        methodName = "substring(to:)" and
-        paramName = "to"
-        or
-        // `NSMutableString.insert`
-        className = "NSMutableString" and
-        methodName = "insert(_:at:)" and
-        paramName = "at"
-      ) and
-      c.getName() = className and
-      c.getAMember() = f and // TODO: will this even work if its defined in a parent class?
```
jketema, line 76: I think I need a voice over for the removal of the comment. My (likely flawed) intuition is that the relevant case here is:
```
let nstr2 = ns.substring(from: s.count - 1) // BAD: String length used in NSString [NOT DETECTED]
```
However, that says "not detected"

jketema (commented):
One question. Otherwise this looks very nice.

swift/ql/src/queries/Security/CWE-135/StringLengthConflation.ql
rdmarsh2: That's what I'd expect - we typically don't keep TODO comments in the code once there's a failing test case for them.

jketema: Thanks for clearing that up.

jketema (approved):

swift/ql/src/queries/Security/CWE-135/StringLengthConflation.ql
geoffw0: Yep, sorry for not replying sooner.  We generally don't like to keep TODO comments in the code, but sadly this is unfixable at the moment, so its documented with a NOT DETECTED test case now.

Should be an easy fix when we extract information about class hierarchies.

PR #9851, opened 2022-07-19
Ruby: Recognise more AR write accesses

hmac:
This change means we recognise calls like

```rb
User.create(params)
User.update(id, params)
```

as instances of `PersistentWriteAccess`.


alexrford (approved):
LGTM, DCA results look viable

PR #9852, opened 2022-07-19
Python: Add `Str` class

### LINKED ISSUE #9833: [Python AST] AST Viewer shows [Str] but no such type exists
**Description of the issue**

I expect the type names present on the AST viewer to match a real CodeQL type, but this is not the case here.
The type could be either Str_ or StrConst, both printing "Str" as their toString().

The Str_ class also mentions the Str class which I cannot find: https://github.com/github/codeql/blob/d50816a284e71f50d0c937928ec5a8521c9dba57/python/ql/lib/semmle/python/AstGenerated.qll#L1240

tausbn:
This makes the AST viewer (which annotates string constant nodes as `Str`, and not `StrConst`) a bit more consistent.

Fixes #9833. 

It might be even better to instead flip the inheritance and make `Str` the lowest class in the hierarchy. That way, it would have access to all of the methods on `StrConst` (and indeed, we might consider just deprecating that class).

---
This is a very minor change, so I don't think a change note is required.

### DIFF
--- a/python/ql/lib/semmle/python/Exprs.qll
+++ b/python/ql/lib/semmle/python/Exprs.qll
@@ -617,10 +617,10 @@ private string non_byte_prefix() {
 }
 
 /** A string constant. This is a placeholder class -- use `StrConst` instead. */
-class Str extends Str_ { }
+class Str = StrConst;
 
 /** A string constant. */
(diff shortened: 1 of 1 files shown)


RasmusWL (approved):
As a fix for people using AST viewer, this LGTM! :+1:

PR #9854, opened 2022-07-19
Python: Add `toString` method to `StrConst`

fcasal:
Override StrConst's toString to show the correct type on the AST viewer. Currently, the AST shows the type "Str", which does not exist.

Fixes https://github.com/github/codeql/issues/9833.


fcasal:
@tausbn could you add the no change note required tag so that the tests run?

tausbn:
Thank you for your submission. After discussing this in the Python team, in the end, we went for a slightly less invasive solution to this issue: #9852.

PR #9866, opened 2022-07-20
Ruby: handle magic coding: comments

aibaars:
This PR adds support for magic character `coding: [CHARSET]` comments in Ruby source files. If such a comment is detected and `CHARSET` is not `UTF-8` then the source file's contents needs to be converted from `CHARSET` into UTF-8. In addition we should store the converted contents in the source archive, because files in the source archives have to be in UTF-8 format.

See also: https://en.wikibooks.org/wiki/Ruby_Programming/Encoding#Using_Encodings

### DIFF
--- a/ruby/extractor/src/main.rs
+++ b/ruby/extractor/src/main.rs
@@ -48,17 +48,13 @@ lazy_static! {
 }
 
 fn encoding_from_name(encoding_name: &str) -> Option<&(dyn encoding::Encoding + Send + Sync)> {
-    match encoding::label::encoding_from_whatwg_label(&encoding_name) {
-        Some(e) => return Some(e),
-        None => {
-            if let Some(cap) = CP_NUMBER.captures(&encoding_name) {
-                return encoding::label::encoding_from_windows_code_page(
-                    str::parse(cap.get(1).unwrap().as_str()).unwrap(),
-                );
-            } else {
-                return None;
-            }
-        }
+    match encoding::label::encoding_from_whatwg_label(encoding_name) {
+        s @ Some(_) => s,
+        None => CP_NUMBER.captures(encoding_name).and_then(|cap| {
+            encoding::label::encoding_from_windows_code_page(
+                str::parse(cap.get(1).unwrap().as_str()).unwrap(),
+            )
+        }),
     }
 }
 
--- a/ruby/extractor/src/main.rs
+++ b/ruby/extractor/src/main.rs
@@ -1,10 +1,14 @@
 mod extractor;
 
+#[macro_use]
+extern crate lazy_static;
 extern crate num_cpus;
 
 use clap::arg;
+use encoding::{self};
 use flate2::write::GzEncoder;
 use rayon::prelude::*;
+use std::borrow::Cow;
 use std::fs;
 use std::io::{BufRead, BufWriter};
 use std::path::{Path, PathBuf};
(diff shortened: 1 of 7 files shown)


ruby/extractor/src/main.rs
```diff
@@ -168,6 +170,16 @@ fn main() -> std::io::Result<()> {
                 }
                 code_ranges = ranges;
             } else {
+                if let Some(encoding) = scan_coding_comment(&source) {
+                    is_utf8 = encoding.whatwg_name().unwrap_or_default() == "utf-8";
+                    if !is_utf8 {
+                        source = encoding
+                            .decode(&source, encoding::types::DecoderTrap::Replace)
+                            .expect("character decoding failure")
```
nickrolfe, line 178: `expect` will panic if `decode` returns `Err`, but we would want to just skip processing of that file, rather than aborting extraction entirely. Does the use of `DecoderTrap::Replace` guarantee that `decode` will always return `Ok`?

ruby/extractor/src/main.rs
```diff
@@ -299,3 +315,151 @@ fn path_for(dir: &Path, path: &Path, ext: &str) -> PathBuf {
     }
     result
 }
+
+fn skip_space(content: &Vec<u8>, index: usize) -> usize {
```
nickrolfe, line 319: Clippy suggests
```suggestion
fn skip_space(content: &[u8], index: usize) -> usize {
```

ruby/extractor/src/main.rs
```diff
@@ -299,3 +315,151 @@ fn path_for(dir: &Path, path: &Path, ext: &str) -> PathBuf {
...
+    let mut index = index;
+    while index < content.len() {
+        let c = content[index] as char;
+        // white space except \n
+        let is_space = c == ' ' || c >= '\t' && c <= '\r' && c != '\n';
```
nickrolfe, line 324: Clippy suggests this, but I'm not sure I prefer one over the other.
```suggestion
        let is_space = c == ' ' || ('\t'..='\r').contains(&c) && c != '\n';
```

ruby/extractor/src/main.rs
```diff
@@ -299,3 +315,151 @@ fn path_for(dir: &Path, path: &Path, ext: &str) -> PathBuf {
...
+        if !is_space {
+            break;
+        }
+        index += 1;
+    }
+    index
+}
+
+fn scan_coding_comment(content: &Vec<u8>) -> std::option::Option<&'static dyn encoding::Encoding> {
```
nickrolfe, line 333:
```suggestion
fn scan_coding_comment(content: &[u8]) -> std::option::Option<&'static dyn encoding::Encoding> {
```

nickrolfe (commented):
Very nice. I have one concern and a few nitpicks courtesy of `cargo clippy`.

ruby/extractor/src/main.rs
aibaars: It should not fail, but I rewrote the code anyway to print a warning instead.

aibaars:
@nickrolfe This is ready for review now.

ruby/extractor/src/main.rs
```diff
@@ -39,6 +43,25 @@ fn num_codeql_threads() -> usize {
...
+        None => {
+            if let Some(cap) = CP_NUMBER.captures(&encoding_name) {
+                return encoding::label::encoding_from_windows_code_page(
+                    str::parse(cap.get(1).unwrap().as_str()).unwrap(),
+                );
+            } else {
+                return None;
+            }
+        }
+    }
```
nickrolfe, line 62: We can make a few simplifications here to make it more Rust-y.

```suggestion
    match encoding::label::encoding_from_whatwg_label(encoding_name) {
        s @ Some(_) => s,
        None => CP_NUMBER.captures(encoding_name).and_then(|cap| {
            encoding::label::encoding_from_windows_code_page(
                str::parse(cap.get(1).unwrap().as_str()).unwrap(),
            )
        }),
    }
```

1.  No need to `return x;` when you can just write `x`.
2. `s @ Some(_) => s` lets you avoid destructuring and recreating the `Some(...)` value (not a big problem, I admit).
3. What you wrote in the `None` case can be written more succinctly with `Option::and_then`.

nickrolfe (approved):
🚀

PR #9867, opened 2022-07-20
C#: Negative summaries (ie. no flow through)

michaelnebel:
In this PR we introduce the concept of `negative` flow summaries, which is a summary of a callable stating that there is no flow via this callable.
The implementation introduces a new class NegativeSummaryModelCsv class, which re-uses parts of the syntax for summaries. The CSV row syntax is `namespace;type;name;signature;provenance`.

- The implementation is made for C# and Java (and dummy/empty implementation has been made for Swift and Ruby)
- The Framework code can parse the negative summaries and new predicates have been introduced to extract the negative summaries from NegativeSummaryModelCsv subclasses.
- Validation of the negative summaries is included in the CsvValidation.
- The model generator(s) have been improved to also output a negative summary in case no flow through summary exists.
    - Furthermore it turned out that we need to discard Finalizers (destructors) in the model generation.
- Telemetry queries now respect the negative summaries. That is, if a callable has a negative flow summary it will not be reported as missing by the telemetry queries.
- Negative summaries are not included and not counted towards flow summaries in general (in the model coverage queries).

Open questions
- Should we extends the model coverage functionality to include these summaries or alternatively make another category of coverage where this is reported?
- Should we make some DCA support for this as well?

@hvitved Any insights about the open questions and/or the implementation itself will be valuable before continuing the work.


michaelnebel:
DCA tests have executed.
(1) Performance of nightly queries show no performance regressions or change in alerts.
(2) Performance of capture model show no performance regressions, but a significant change in the number of output rows. For .NET Runtime there is an increase in around 40k rows (which is probably acceptable as we have 10k positive rows). For .NET EFCore we don't know how many positive rows we have but there is a total of around 70k rows.
Will try and run the Capture model queries again on all DCA projects such that we get an understanding of the magnitude of rows.

csharp/ql/lib/semmle/code/csharp/dataflow/CsvValidation.qll
```diff
@@ -0,0 +1,140 @@
+/** Provides a query predicate to check the CSV data for validation errors. */
+
+import csharp
```
hvitved, line 3: As discussed offline, let's move this back into `ExternalFlow.qll`.

csharp/ql/lib/semmle/code/csharp/dataflow/NegativeSummary.qll
```diff
@@ -0,0 +1,9 @@
+/** Provides modules for importing negative summaries. */
+
+/**
+ * A module importing the frameworks that provide external flow data,
+ * ensuring that they are visible to the taint tracking / data flow library.
+ */
+private module Frameworks {
```
hvitved, line 7: I think I would move this file into `internal`.

csharp/ql/lib/semmle/code/csharp/dataflow/internal/DataFlowPrivate.qll
```diff
@@ -2129,18 +2129,37 @@ module Csv {
...
           + getCallableOverride(c) + ";" //
           + name + ";" //
-          + "(" + parameterQualifiedTypeNamesToString(c) + ")" + ";" //
+          + parameters + ";" //
           + /* ext + */ ";" //
     )
   }
+
+  /** Computes the first 4 columns for negative CSV rows of `c`. */
+  string asPartialNegativeModel(DotNet::Callable c) {
```
hvitved, line 2155: I wonder if we should add `bindingset[c]` to all three predicates, to effectively inline them. Otherwise we will in fact be constructing strings for all callables, and not just the relevant ones.

csharp/ql/lib/semmle/code/csharp/dataflow/internal/FlowSummaryImplSpecific.qll
```diff
@@ -114,6 +114,18 @@ predicate summaryElement(Callable c, string input, string output, string kind, b
   )
 }
 
+/**
+ * Holds is an external flow summary exists for `c` which means that there is no
```
hvitved, line 118: is -> if

csharp/ql/src/utils/model-generator/CaptureNegativeSummaryModels.ql
```diff
@@ -0,0 +1,15 @@
...
+ * @id cs/utils/model-generator/negative-summary-models
+ * @tags model-generator
+ */
+
+private import semmle.code.csharp.dataflow.ExternalFlow
+private import internal.CaptureModels
+private import internal.CaptureSummaryFlow
+
+from TargetApi api, string noflow
+where noflow = captureNoFlow(api) and not hasSummary(api, false)
```
hvitved, line 14: Why is the `not hasSummary(api, false)` bit needed?

csharp/ql/lib/semmle/code/csharp/dataflow/CsvValidation.qll
michaelnebel: Sure thing - I will get on it!

csharp/ql/lib/semmle/code/csharp/dataflow/NegativeSummary.qll
michaelnebel: Good idea!

csharp/ql/lib/semmle/code/csharp/dataflow/internal/DataFlowPrivate.qll
michaelnebel: Yes that sounds plausible!

csharp/ql/lib/semmle/code/csharp/dataflow/internal/FlowSummaryImplSpecific.qll
michaelnebel: Will fix

csharp/ql/src/utils/model-generator/CaptureNegativeSummaryModels.ql
michaelnebel: The CaptureFlow predicate might return a less specialized version of a handwritten summary or no summary at all (if flow was not identified programmatically). To avoid creating less specialized versions of the summaries or create negative summaries, when there might exist a handwritten one this has been made explicit here.
One could argue, whether we care at all about "conflicts" because at least for positive summaries we discard the less specialized version (due to the heuristic of trusting handwritten summaries over generated ones).
Should we remove the hasSummary(...) check completely? (I would like to do it as a follow up and include a re-gen of the models if we decide to do so).

csharp/ql/lib/semmle/code/csharp/dataflow/internal/DataFlowPrivate.qll
michaelnebel: As discussed offline, we decide not to do this for now.

csharp/ql/src/utils/model-generator/CaptureNegativeSummaryModels.ql
hvitved: Thanks for the reminder; let's keep it in.

go/ql/test/library-tests/semmle/go/dataflow/ExternalFlowVarArgs/Flows.ql
```diff
@@ -1,7 +1,7 @@
 import go
 import semmle.go.dataflow.ExternalFlow
-import CsvValidation
 import TestUtilities.InlineExpectationsTest
+import CsvValidation
```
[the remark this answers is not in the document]
michaelnebel: No particular reason. Side effect of refactoring in two directions.
I will make sure that the import of the CsvValidation module follows the import ExternalFlow module instead.
Will fix.

hvitved (approved):
LGTM. Let's give @aschackmull a chance to review as well.

hvitved (approved):

csharp/ql/lib/semmle/code/csharp/dataflow/ExternalFlow.qll
```diff
@@ -343,58 +345,123 @@ module CsvValidation {
...
+    )
+  }
+
+  private string getInvalidModelSubtype() {
+    exists(string pred, string row, int expect |
+      sourceModel(row) and expect = 9 and pred = "source"
+      or
+      sinkModel(row) and expect = 9 and pred = "sink"
+      or
+      summaryModel(row) and expect = 10 and pred = "summary"
```
aschackmull, line 379: `expect` is unused

csharp/ql/lib/semmle/code/csharp/dataflow/ExternalFlow.qll
```diff
@@ -343,58 +345,123 @@ module CsvValidation {
...
 ) {
-  sourceModel(namespace, type, subtypes, name, signature, ext, _, _, _) or
-  sinkModel(namespace, type, subtypes, name, signature, ext, _, _, _) or
+  sourceModel(namespace, type, subtypes, name, signature, ext, _, _, _)
+  or
+  sinkModel(namespace, type, subtypes, name, signature, ext, _, _, _)
+  or
   summaryModel(namespace, type, subtypes, name, signature, ext, _, _, _, _)
+  or
+  negativeSummaryModel(namespace, type, name, signature, _) and ext = "" and subtypes = false
```
aschackmull, line 464: This makes anything depending on the other kinds of models also depend on negative summaries. This might be fine for now, but if they grow too large that we don't want to scan them for regular queries then we'd likely need to split this predicate here (and do the implementation sharing bits with parameterised modules instead). But it's probably fine to leave that as potential future work.

csharp/ql/lib/semmle/code/csharp/dataflow/internal/FlowSummaryImpl.qll
```diff
@@ -240,6 +240,16 @@ module Public {
      */
     predicate isAutoGenerated() { none() }
   }
+
+  /** A callable with a flow summary stating there is no flow via the callable. */
+  class NegativeSummarizedCallable extends SummarizedCallableBase {
+    NegativeSummarizedCallable() { negativeSummaryElement(this, _) }
```
aschackmull, line 246: I did not expect to see negative summaries as part of `SummarizedCallableBase`. That seems wrong to me.

csharp/ql/lib/semmle/code/csharp/dataflow/internal/FlowSummaryImplSpecific.qll
```diff
@@ -114,6 +114,18 @@ predicate summaryElement(Callable c, string input, string output, string kind, b
   )
 }
 
+/**
+ * Holds if an external flow summary exists for `c` which means that there is no
+ * flow through `c` and a flag `generated` stating whether the summary is autogenerated.
```
aschackmull, line 119: There's a grammatical mistake here that really messes with the meaning of the sentence.
```suggestion
 * Holds if a negative flow summary exists for `c`, which means that there is no
 * flow through `c`. The flag `generated` states whether the summary is autogenerated.
```

csharp/ql/src/utils/model-generator/internal/CaptureSummaryFlow.qll
```diff
@@ -79,3 +79,16 @@ string captureFlow(TargetApi api) {
...
+
+/**
+ * Gets the negative summary for `api`, if any.
+ * A negative summary is generated, if there does not exist any positive flow that
+ * shares the same summary prefix - otherwise these models will be indistinguishable.
+ */
+string captureNoFlow(TargetApi api) {
+  not exists(TargetApi other, string flow |
+    flow = captureFlow(other) and
+    partialNegativeModel(other) = partialNegativeModel(api)
```
aschackmull, line 91: Does this join perform well enough? It looks suspicious.

aschackmull: Hmm, is `partialNegativeModel` a bijection? I assume not, since it's used here, but I can't quite see why.

csharp/ql/lib/semmle/code/csharp/dataflow/ExternalFlow.qll
michaelnebel: The idea is that the modules containing the negative summaries are included near the queries that relies on them. Are you worried that if a query that imports the negative summaries are included in query package containing many other queries that these other queries will be slower?

csharp/ql/lib/semmle/code/csharp/dataflow/ExternalFlow.qll
michaelnebel: Will fix - copy paste refactor mistake.

csharp/ql/lib/semmle/code/csharp/dataflow/internal/FlowSummaryImpl.qll
michaelnebel: Not exactly sure, what you mean.
Maybe it is the naming of the class SummarizedCallableBase that is misleading - maybe SummarizedCallableBase should be renamed to something like SyntheticCallable(Base). Note that summarized callables (SummarizedCallable) extends SummarizedCallableBase.

csharp/ql/src/utils/model-generator/internal/CaptureSummaryFlow.qll
michaelnebel: Good catch. This was needed in one of previous versions of the Negative summaries (the syntax and information level has changed since this was introduced). I will remove it.

csharp/ql/lib/semmle/code/csharp/dataflow/internal/FlowSummaryImplSpecific.qll
michaelnebel: Yes - I will fix it.

csharp/ql/src/utils/model-generator/internal/CaptureSummaryFlow.qll
michaelnebel: It turns out when generating models for .NET runtime, the source of the reference assemblies interferes with the model generation. Eg. we have
```csharp
System.Linq.Expressions.Expression System.Linq.Expressions.IArgumentProvider.GetArgument(int index) { throw null; }
public Expression GetArgument(int index) => _arguments[index];
```
Which means that both a positve and negative flow summary is generated.

csharp/ql/lib/semmle/code/csharp/dataflow/ExternalFlow.qll
aschackmull: Currently inclusion is based on file module activation (aka transitive import), and the negative models aren't activated for regular queries, so they'll be fine. But all of the positive models and their dependencies will be recomputed once negative models are included, so any queries referring to negative models need to do extra work. Again, this is probably fine for now.
I originally didn't notice that the negative models were excluded from the computation by way of not being activated, since the picture in my mind was a more explicit split at the predicate level. And once we move the MaD rows to external files then we need to refactor this, since at that point the exclusion-by-no-import won't work.

csharp/ql/lib/semmle/code/csharp/dataflow/internal/FlowSummaryImpl.qll
aschackmull: But negative summaries don't need data flow node synthesis. The `SummarizedCallable` class is the set of callables for which we synthesise a number of nodes to enable flow-through.

csharp/ql/lib/semmle/code/csharp/dataflow/ExternalFlow.qll
michaelnebel: Great! Thank you very much for the explanation! 😃

csharp/ql/lib/semmle/code/csharp/dataflow/internal/FlowSummaryImpl.qll
michaelnebel: Not exactly sure how that could happen.
Note that there are three classes `SummarizedCallable`, `SummarizedCallableBase` and `NegativeSummarizedCallable`. SummarizedCallable is an abstract class that extends `SummarizedCallableBase` and `NegativeSummarizedCallable` extends `SummarizedCallableBase` and not `SummarizedCallable`.
Should we change the same of `SummarizedCallableBase` to something else?

aschackmull: Never mind. This was me misreading and thinking that `SummarizedCallableBase` was an abstract class.

aschackmull (dismissed):

michaelnebel:
(1) DCA for nightly looks fine. No problem with query execution.
(2) DCA run for the Models as Data queries themselves hit an OOM exception. This should not block merging this PR, but I will investigate.
@aschackmull : Since you last review'ed there has been a rebase, but no other changes.

aschackmull (dismissed):

michaelnebel:
> (1) DCA for nightly looks fine. No problem with query execution. (2) DCA run for the Models as Data queries themselves hit an OOM exception. This should not block merging this PR, but I will investigate. @aschackmull : Since you last review'ed there has been a rebase, but no other changes.

@aschackmull : Had to rebase yet again. No other changes. Please "review" again. I will keep an eye on the CI and merge ASAP if it becomes possible.

aschackmull (approved):

PR #9868, opened 2022-07-20
Ruby: update tree-sitter-ruby

aibaars:
Update to https://github.com/tree-sitter/tree-sitter-ruby/pull/222

### DIFF
--- a/ruby/Cargo.lock
+++ b/ruby/Cargo.lock
@@ -556,7 +556,7 @@ dependencies = [
 [[package]]
 name = "tree-sitter-ruby"
 version = "0.19.0"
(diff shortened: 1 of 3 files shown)


nickrolfe:
Looking at DCA, do you know why the "Lines of code in the database" metric changed?

aibaars:
> Looking at DCA, do you know why the "Lines of code in the database" metric changed?

Good question. It turns out that all differences are in files with `__END__` blocks. Before this PR the `uninterpreted` block would start right after the `__END__` marker and as a result include the newline after the `__END__`. In the updated grammar the newline is consumed as part of the `__END__[\r\n]` token. As a result the `uninterpreted` block starts on the line after the `__END__`. I think that is actually better.

nickrolfe (approved):
Excellent, that makes sense.

PR #9875, opened 2022-07-21
Kotlin: Special-case String.charAt naming

smowton:
In the Kotlin universe this is called `get` so that Kotlin programmers can use the `[]` operator on `String`s.


atorralba (dismissed):

smowton:
@atorralba could you re-approve?

atorralba (approved):
Done :)

PR #9876, opened 2022-07-21
Kotlin: implement default interface forwarding

smowton:
Kotlin's implementation of defaults depends on the -Xjvm-default setting (or the `@JvmDefault` deprecated annotation, not implemented here): by default, actual interface class files don't use default methods, and any class that would inherit one instead implements the interface calling a static method defined on `TheInterface$DefaultImpls`. With
`-Xjvm-default=all` or `=all-compatibility`, real interface default methods are emitted, with the latter retaining the `DefaultImpls` methods so that other Kotlin can use it.

Here I adopt a hybrid solution: create a real default method implementation, but also emit a forwarding method like `@override int f(int x) { return super.TheInterface.f(x); }`, because the Java extractor will see `MyClass.f` in the emitted class file and try to dispatch directly to it. The only downside is that we emit a default interface
method body for a prototype that will appear to be `abstract` to the Java extractor and which it will extract as such. I work around this by tolerating the combination `default abstract` in QL. The alternative would be to fully mimic the DefaultImpls approach, giving 100% fidelity to kotlinc's strategy and therefore no clash with the Java
extractor's view of the world.


igfoo:
Generally looks plausible to me, but the tests are failing.

java/kotlin-extractor/src/main/kotlin/KotlinFileExtractor.kt
```diff
@@ -823,10 +824,60 @@ open class KotlinFileExtractor(
         }
     }
 
+    private fun isKotlinDefinedInterface(cls: IrClass?) =
+        cls != null && cls.isInterface && cls.origin != IrDeclarationOrigin.IR_EXTERNAL_JAVA_DECLARATION_STUB
```
tamasvajk, line 828: Do we need to cover annotation classes?

java/kotlin-extractor/src/main/kotlin/KotlinFileExtractor.kt
```diff
@@ -1195,7 +1246,9 @@ open class KotlinFileExtractor(
 
     private fun extractProperty(p: IrProperty, parentId: Label<out DbReftype>, extractBackingField: Boolean, extractFunctionBodies: Boolean, extractPrivateMembers: Boolean, typeSubstitution: TypeSubstitution?, classTypeArgsIncludingOuterClasses: List<IrTypeArgument>?) {
         with("property", p) {
-            if (isFake(p)) return
+            fun needsInterfaceForwarderQ(f: IrFunction?) = f?.let { needsInterfaceForwarder(f) } ?: false
```
tamasvajk, line 1249: What does the `Q` postfix mean?

smowton: Question

java/kotlin-extractor/src/main/kotlin/KotlinFileExtractor.kt
smowton: No, because they cannot have non-Object supertypes or members, so the circumstance should never be able to arise.

smowton:
This should now be in a mergeable state.

java/ql/test/library-tests/structure/structure/A.java
```diff
@@ -7,7 +7,7 @@ public class A {
 	int y;
 	class MemberClass { }
 	void m() {
-		class LocalClass { int n; }
+		class LocalClass { int n; void localClassMethod() { class DoublyLocalClass { void doublyLocalClassMethod() { } } } }
```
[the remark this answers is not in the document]
smowton: No, this is relating to the commit https://github.com/github/codeql/pull/9876/commits/b763c406b68d83dbc9f7400be854b544b3a741c8 which was necessary to avoid a spurious test failure

java/ql/test/kotlin/library-tests/reflection/reflection.expected
```diff
@@ -263,29 +263,78 @@ modifiers
...
+| file://<external>/IntProgression.class:0:0:0:0 | forEach | Forwarder for a Kotlin class inheriting an interface default method |
+| file://<external>/IntProgression.class:0:0:0:0 | spliterator | Forwarder for a Kotlin class inheriting an interface default method |
+| file://<external>/IntRange.class:0:0:0:0 | forEach | Forwarder for a Kotlin class inheriting an interface default method |
+| file://<external>/IntRange.class:0:0:0:0 | spliterator | Forwarder for a Kotlin class inheriting an interface default method |
+| file://<external>/KTypeProjection.class:0:0:0:0 | contravariant | Proxy static method for a @JvmStatic-annotated function or property |
+| file://<external>/KTypeProjection.class:0:0:0:0 | copy$default | Forwarder for Kotlin calls that need default arguments filling in |
+| file://<external>/KTypeProjection.class:0:0:0:0 | covariant | Proxy static method for a @JvmStatic-annotated function or property |
+| file://<external>/KTypeProjection.class:0:0:0:0 | invariant | Proxy static method for a @JvmStatic-annotated function or property |
+| file://<external>/List.class:0:0:0:0 | forEach | Forwarder for a Kotlin class inheriting an interface default method |
```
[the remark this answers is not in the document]
smowton: Looks like this is the Kotlin Iterable equivalent of https://docs.oracle.com/javase/8/docs/api/java/lang/Iterable.html#forEach-java.util.function.Consumer- -- there's also the extension function defined in CollectionsKt, which takes a `Function1<T, Unit>` instead of a `Consumer<T>`. Not sure why they bothered to do that instead of just SAM converting the function to Consumer.

java/ql/integration-tests/posix-only/kotlin/kotlin-interface-inherited-default/test.kt
```diff
@@ -0,0 +1,12 @@
...
+  fun f() = 1
+
+  fun g(x: Int) = x
+
+  val x : Int
+    get() = 3
+
+}
+
+class Real : Test { }
```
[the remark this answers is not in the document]
smowton: Good point -- just pushed a commit that avoids synthesising forwarders in abstract types, and makes sure the `super` qualifier is a direct supertype like it's supposed to be.

tamasvajk (approved):

PR #9891, opened 2022-07-25
Swift: first prototype of a generated IPA layer

redsun82:
This introduces a first version of a generated IPA layer.

## Schema
In `schema.yml`, types can be marked with either
```yml
MyType:
  _synth:
    from: OtherType
```
or
```yml
MyType:
  _synth:
    on:
      arg1: OtherType
      arg2: int
```
The former are "derived" IPA types, built on top of DB entities, the latter are "fresh" ones, built from scratch.

## New generated files
`Raw.qll` defines a thin non customizable QL class layer over DB types, inside a `Raw` module. So for example `Raw::Expr` is a simple wrapper of `@expr`.

`Synth.qll` defines the base algebraic type `Synth::TElement` with all pure DB types and IPA types as defined in `schema.yml`. All `Synth` types get characteristic predicates (except for parameterless derived IPA types) to be customized in stubs (`MyTypeConstructor.qll`) alongside the other stubs already created. The default provided constructors are:
* for derived and fresh IPA types: `none()` 
* for raw types: `not constructX1(id) and ... and no constructXn(id)` where `Xi` are the derived IPA types built on top of the given raw type (`any()` if there are none)

The rest of the type hierarchy is recreated via `class = ... or ...` declarations. The `Synth` module is fully cached, and also provides conversions to and from raw entities when it makes sense (for types that are not derived IPA ones).

## Class customization
This works more or less like before, with one notable exception: the overrideable property getters are now named `getImmediateX` in case the result is a class (not a built-in type), and are used to return the result without element resolution kicking in. For pure raw types, properties are implemented as before based on DB relations (using the property getters defined in `Raw.qll`). For fresh or derived IPA types, properties are left unimplemented. Inherited properties will however be implemented if the type is a derived IPA with the source compatible with the inherited type.

## Example
This PR contains a simple example of usage, introducing singleton IPA-synthesized unknown file and location instances.

## Class fields
It was initially planned to have underlying DB entities or derived IPA constructor arguments as fields in the classes. This turned out to not work because that would have required extensive use of characteristic predicates, which were exhausting the internal limit of 2^25 for BDD tree nodes...

## Further work
This is a prototype with some corners being cut:
* IPA types are still used for `dbscheme`, `trap` and `cpp` generation, they should be removed from there
* `qlgen.py` has become too big and with too many things being generated, it should probably be split
* while codegen unit tests have been kept passing, the additional functionality is not covered yet
* there is room for quality of life improvements, like generating getters for derived IPA type constructor arguments
* `getImmediateX` might be too noisy to have in the public interface of the classes. Maybe we can rethink the resolution mechanism that we have used mostly for hiding conversion AST nodes, or maybe we can just hide away this detail better. (a `protected` access modifier would help here...)


swift/ql/lib/codeql/swift/generated/File.qll
```diff
@@ -3,6 +3,6 @@ private import codeql.swift.generated.IpaTypes
 private import codeql.swift.generated.Db
 import codeql.swift.elements.Element
 
-class FileBase extends Cached::TFile, Element {
-  string getName() { result = asDbInstance().(Db::File).getName() }
+class FileBase extends Ipa::TFile, Element {
+  string getName() { result = Ipa::toDbInstance(this).(Db::File).getName() }
```
MathiasVP: Ruby calls these `toDbInstance` predicates `toGenerated`. Should we do the same?

swift/ql/lib/codeql/swift/generated/Db.qll
```diff
@@ -0,0 +1,1500 @@
+module Db {
```
MathiasVP, line 1: More name bikeshedding: How about calling this `Raw` instead?

MathiasVP (commented):
This looks super cool, @redsun82! Here are some initial comments. I think I'll have more later this week.

swift/ql/lib/codeql/swift/generated/IpaTypes.qll
```diff
@@ -0,0 +1,442 @@
...
+    TArrowExpr(@arrow_expr id) or
+    TAssignExpr(@assign_expr id) or
+    TAutoClosureExpr(@auto_closure_expr id) or
+    TAwaitExpr(@await_expr id) or
+    TBinaryExpr(@binary_expr id) or
+    TBindOptionalExpr(@bind_optional_expr id) or
+    TBooleanLiteralExpr(@boolean_literal_expr id) or
+    TBridgeFromObjCExpr(@bridge_from_obj_c_expr id) or
+    TBridgeToObjCExpr(@bridge_to_obj_c_expr id) or
+    TCallExpr(@call_expr id) { not constructMethodCallExpr(id) } or
```
MathiasVP: Thinking about this a bit more, I don't think we need to automatically generate these `not x1(...) and not x2(...) and ... not xn(...)` branches because:
1. This forces us to map all db entities to _some_ IPA branch (which we might not want). For instance, I think we want to hide away the [SelfApplyExpr](https://github.com/apple/swift/blob/main/include/swift/AST/Expr.h#L4703) entity since we want to provide a different API for obtaining qualifiers.
2. If we do need a catch-all IPA branch like this we can just create it in the `schema.yml` file with some `_ipa: ... on:` fields, I think

swift/ql/lib/codeql/swift/controlflow/internal/ControlFlowGraphImpl.qll
```diff
@@ -33,6 +33,7 @@
 
 private import swift
 private import codeql.swift.controlflow.ControlFlowGraph
+private import codeql.swift.generated.Ipa
```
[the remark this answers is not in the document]
redsun82: I like `Synth` :slightly_smiling_face:

swift/ql/lib/codeql/swift/elements/Locatable.qll
```diff
@@ -1,12 +1,13 @@
 private import codeql.swift.generated.Locatable
 private import codeql.swift.elements.File
+private import codeql.swift.elements.UnknownLocation
 
 class Locatable extends LocatableBase {
   pragma[nomagic]
-  override Location getLocation() {
-    result = LocatableBase.super.getLocation()
+  override Location getImmediateLocation() {
```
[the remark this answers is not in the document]
redsun82: That makes a lot of sense for its usage! The downside I see is that when overriding it in stubs one then needs to recreate the same mechanism and put in the resolution by hand, while here one should (and can) only override the `Immediate` version and the resolution is done at the level of the `final` predicate...

I would be less concerned if we could mark the `Immediate` predicates as protected...

swift/ql/lib/codeql/swift/generated/IpaTypes.qll
```diff
@@ -1,324 +1,295 @@
...
-    TPlaceholderTypeRepr(@placeholder_type_repr id) or
-    TProtocolTypeRepr(@protocol_type_repr id) or
-    TSharedTypeRepr(@shared_type_repr id) or
-    TSilBoxTypeRepr(@sil_box_type_repr id) or
-    TSimpleIdentTypeRepr(@simple_ident_type_repr id) or
-    TTupleTypeRepr(@tuple_type_repr id)
-
-  class TAstNode =
-    TCaseLabelItem or TDecl or TExpr or TPattern or TStmt or TStmtCondition or TTypeRepr;
+    TArgument(Db::Argument id) or
```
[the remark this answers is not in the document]
redsun82: I did not add it in yet, as while I was playing with it, I was not completely sure IPAing it was the best call. In the end I think that for this specific case we can get away with just extending `CallExpr`, something like

```codeql
class MethodCallExpr extends CallExpr {
  MethodCallExpr() { this.getFunction() instanceof SelfApplyExpr }

  Expr getInstance() { result = this.getFunction().(SelfApplyExpr).getBase() }
  Expr getMethod() { result = this.getFunction().(SelfApplyExpr).getFunction() }
}
```

swift/ql/lib/codeql/swift/generated/File.qll
redsun82: I'm not sure. We generate more than they do, including the synthesised layer, so while for them `Generated` unambiguously refers to stuff coming from the treesitter generation and that is in the dbscheme, for us the term would not be unambiguous. What do you think?

swift/ql/lib/codeql/swift/generated/Db.qll
redsun82: I like that! So maybe we can also rename `from/toDbInstance` to `from/toRawInstance`?

swift/ql/lib/codeql/swift/generated/IpaTypes.qll
redsun82: I don't completely get it, I guess we can chat about it later :slightly_smiling_face: This avoids the same db instances appearing in separate synthesized classes, wouldn't that be a problem?

swift/ql/lib/codeql/swift/generated/IpaTypes.qll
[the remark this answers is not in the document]
MathiasVP: Hm. Maybe we can start with that, yes 🤔. We also have to give it a better `toString`, though.

Of course, there are still the annoying `SelfApplyExpr`s floating around the AST, which is kinda annoying. But if we proceed with this PR (and go the route I suggested in [this comment](https://github.com/github/codeql/pull/9891#discussion_r940047173)) I guess we can get rid of those.

swift/ql/lib/codeql/swift/generated/Db.qll
MathiasVP: Sounds good. I'm not sure I like the word "instance" since this is not a word we really use in QL. Ruby calls these `fromGenerated/toGenerated`. Could we do the same maybe?

swift/ql/lib/codeql/swift/elements/Locatable.qll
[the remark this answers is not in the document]
redsun82: As discussed, we can come back to it later, but for the moment we leave it like this but rename `Immediate` to `Unresolved`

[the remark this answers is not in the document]
redsun82: uh, `Unresolved` risks being confusing with swift's `Unresolved` AST nodes... I'll keep it like that, but we will get back to this later.

swift/ql/lib/codeql/swift/generated/Db.qll
redsun82: I went with `fromRaw`/`toRaw` :slightly_smiling_face:

swift/ql/lib/codeql/swift/generated/IpaTypes.qll
redsun82: I'm trying out generalising constructors to all types, with useful defaults (`none()` for actual IPA type, and `not constructX` disjunctions for DB types, with `any()` if no ipa type is built from it), but want to keep an eye if it impacts test run time.

swift/ql/lib/codeql/swift/generated/Comment.qll
```diff
@@ -1,8 +1,10 @@
...
+private import codeql.swift.generated.Synth
+private import codeql.swift.generated.Raw
 import codeql.swift.elements.Locatable
 
-class CommentBase extends @comment, Locatable {
+class CommentBase extends Synth::TComment, Locatable {
   override string getAPrimaryQlClass() { result = "Comment" }
 
-  string getText() { comments(this, result) }
+  string getText() { result = Synth::toRawComment(this).(Raw::Comment).getText() }
```
MathiasVP, line 9: Why does `Synth::toRawComment` not specify a `Raw::Comment` return type?

redsun82: I'll try it out though, maybe it can work

MathiasVP: Yeah, fair enough. Ruby just has a single predicate called `to/fromGenerated` (in which case the generic result type kinda makes sense). Should we do the same?

redsun82: I think it could work, but it would require a bit of an effort and even more complexity on the generation side, so I'd rather not do it unless we have reason to. As far as I can tell we won't be calling these conversion functions a lot in non-generated code.

redsun82: (I will rename the conversion functions thought to avoid confusion)

redsun82: The "I think it could work" comment above was written without seeing your last comment and was referring to specialising the raw types on conversions.

As to


> Yeah, fair enough. Ruby just has a single predicate called `to/fromGenerated` (in which case the generic result type kinda makes sense). Should we do the same (just define a giant disjunction in a single predicate, that is)?

I had that version at one point, it was working ok. I then tried the version with separate conversion functions and it turned out it was performing a bit better. Not by a lot, but still noticeably, probably exactly because avoiding the giant disjunction in many cases.

MathiasVP (approved):
LGTM!

PR #9892, opened 2022-07-25
C++: Global value numbering for function calls

rdmarsh2:
This PR adds global value numbering for function calls, taking into account side effects. This doesn't do any new analysis to eliminate side effects, so it will only apply to functions that we have an existing model for.


cpp/ql/lib/semmle/code/cpp/ir/implementation/aliased_ssa/gvn/internal/ValueNumberingInternal.qll
```diff
@@ -41,8 +41,75 @@ newtype TValueNumber =
...
+newtype TCallPartialValueNumber =
+  TNilArgument() or
+  TArgument(TCallPartialValueNumber head, TValueNumber arg) {
+    exists(CallInstruction call, int index |
+      callArgValueNumber(call, index, arg) and
+      callPartialValueNumber(call, index, head)
+    )
+  }
+
+predicate callValueNumber(CallInstruction call, int index, TCallPartialValueNumber vn) {
```
MathiasVP, line 56: Should we maybe exclude `CallInstruction`s that has a void return type? That's a slightly less drastic change, and it'll suffice for our vector size use-case.

cpp/ql/lib/semmle/code/cpp/ir/implementation/aliased_ssa/gvn/internal/ValueNumberingInternal.qll
```diff
@@ -92,6 +92,20 @@ predicate callArgValueNumber(CallInstruction call, int index, TValueNumber arg)
...
+        isEffect = false
+      )
+      or
+      exists(CallReadSideEffectInstruction cse |
+        cse.getPrimaryInstruction() = call and
+        cse.getSideEffectOperand().getAnyDef() = instr and
+        argIndex = -2 and
+        isEffect = false
+      )
+      or
```
MathiasVP, line 108: It probably doesn't matter a lot, but shouldn't both of these be `isEffect = true`?

cpp/ql/lib/semmle/code/cpp/ir/implementation/internal/TInstructionInternal.qll
```diff
@@ -2,3 +2,4 @@ import semmle.code.cpp.ir.internal.IRCppLanguage as Language
 import semmle.code.cpp.ir.implementation.raw.internal.IRConstruction as IRConstruction
 import semmle.code.cpp.ir.implementation.unaliased_ssa.internal.SSAConstruction as UnaliasedSsa
 import semmle.code.cpp.ir.implementation.aliased_ssa.internal.SSAConstruction as AliasedSSA
+import semmle.code.cpp.ir.implementation.aliased_ssa.internal.SideEffectElimination as Elim
```
MathiasVP, line 5: I don't think you've checked `SideEffectElimination.qll` in yet. That explains all the consistency errors in ql-for-ql.

cpp/ql/lib/semmle/code/cpp/ir/implementation/aliased_ssa/gvn/internal/ValueNumberingInternal.qll
MathiasVP: Also, I think all of the new definitions can be marked as `private`, no?

cpp/ql/lib/semmle/code/cpp/ir/implementation/aliased_ssa/gvn/internal/ValueNumberingInternal.qll
```diff
@@ -41,8 +41,75 @@ newtype TValueNumber =
...
+    vn = TArgument(head, arg)
+  )
+  or
+  not exists(int n | callArgRank(call, n, _)) and
+  index = -1 and
+  vn = TNilArgument()
+}
+
+predicate callPartialValueNumber(CallInstruction call, int index, TCallPartialValueNumber head) {
+  index = 1 and head = TNilArgument()
```
MathiasVP, line 70: I would really prefer this to be `index = 0`. Is there a reason it has to be `index = 1`?

cpp/ql/lib/semmle/code/cpp/ir/implementation/aliased_ssa/gvn/internal/ValueNumberingInternal.qll
MathiasVP: Alternatively, we could choose to only GVN functions that are side-effect free (which I think would cover `std::vector`'s `size` function, right?)

MathiasVP (commented):
The changes to .expected files LGTM! I've left a couple of comments regarding how many calls we really want to GVN.

I'm eagerly waiting for the missing `SideEffectElimination` file. Once that has been checked in we should definitely do a DCA run on this. If there's a sign of an unacceptable slowdown we could consider GVN'ing fewer calls like I've suggested in a couple of comments.

cpp/ql/lib/semmle/code/cpp/ir/implementation/aliased_ssa/internal/SideEffectElimination.qll
```diff
@@ -0,0 +1,42 @@
...
+  exists(Unaliased::IRFunction irFunc | irFunc.getFunction() = func) and
+  noLocalSideEffectWrite(func) and
+  forall(Unaliased::CallInstruction call | call.getEnclosingFunction() = func |
+    exists(call.getStaticCallTarget())
+  ) and
+  forall(Unaliased::CallInstruction call, CPP::Function callee |
+    call.getStaticCallTarget() = callee and
+    call.getEnclosingFunction() = func
+  |
+    noTransitiveSideEffectWrite(callee)
```
MathiasVP, line 27: I'm always scared of recursion through `forall` as it's very easy to get a quadratic blowup. There's a general performance trick to this described in this internal issue: https://github.com/github/codeql-core/issues/416. We might want to apply that here.

cpp/ql/lib/semmle/code/cpp/ir/implementation/aliased_ssa/gvn/internal/ValueNumberingInternal.qll
rdmarsh2: Actually the place to do this is in CallPartialValueNumber

cpp/ql/lib/semmle/code/cpp/ir/implementation/aliased_ssa/gvn/internal/ValueNumberingInternal.qll
rdmarsh2: I'd prefer to GVN all functions if we can, and eventually GVN the memory side effects as well (but that's work for another PR)

cpp/ql/lib/semmle/code/cpp/ir/implementation/aliased_ssa/gvn/internal/ValueNumberingInternal.qll
rdmarsh2: ... Apparently I got my recursion wrong here - this is actually skipping the first item in the rank.

MathiasVP:
Small update on this PR. There's a very large performance regression. I would have guessed that it was due to the recursion through `forall`, but it looks like `callPartialValueNumber` is to blame. There may just be too much non-linear recursion.

PR #9896, opened 2022-07-26
Ruby: port js/hardcoded-data-interpreted-as-code

nickrolfe:


aibaars:
Should this query flag up https://github.com/hahwul/mad-metasploit/blob/master/archive/exploits/windows/local/17177.rb#L97 ? The file is truncated, see the bottom of the file at https://raw.githubusercontent.com/hahwul/mad-metasploit/master/archive/exploits/windows/local/17177.rb.

nickrolfe:
That's a useful example. The query won't flag that, for a few reasons:

1. The string won't be considered a `Source`, given the current regexp.
2. I assume we don't model flow through the call to `Zlib::Inflate.inflate`.
3. The decompressed data is not actually executed. It's just written to a file.

I should at least fix 1.

alexrford (approved):
LGTM overall - I agree that it would be good to extend the `DefaultSource` to cover sequences of arbitrary character codes as in the metasploit example.

alexrford (approved):
👍

PR #9898, opened 2022-07-26
Kotlin: implement super-method calls

smowton:
If we only look at the dispatch receiver, these show up like `this` references rather than `super` references, preventing flow through super-calls. The super-interface case requires properly noting that interface methods with a body get a `default` modifier in order to avoid QL discarding the method as a possible callee.


atorralba (approved):
Looks plausible to me.

PR #9903, opened 2022-07-27
C++: Support additional compiler builtin operations

jketema:
These builtins are supported by our compiler frontend, and occur in internal testing. Test failures are expected, because this depends on an internal PR.


cpp/ql/lib/semmle/code/cpp/exprs/Expr.qll
```diff
@@ -596,9 +596,12 @@ class ParenthesisExpr extends Conversion, @parexpr {
...
 
 /**
- * A C/C++ expression that has not been resolved.
+ * A C/C++ expression that could not be resolved, or that can no longer be
+ * represented due to a database upgrade or downgrade.
  *
- * It is assigned `ErroneousType` as its type.
+ * If the expression could not be resolved, it has type `ErroneousType`. In the
+ * case of a database upgrade or downgrade, the original type from before the
+ * upgrade or downgrade is kept if that type can be represented.
```
geoffw0, line 604: This is great detail, thanks for adding it.

redsun82 (approved):
:rocket:

PR #9905, opened 2022-07-27
Control Flow: extend ordering

redsun82:
In Swift we encountered a situation where ordering of nodes and edges in the control flow graph was not stable between test runs.

This patch:
* adds the successor string representation to the edge ordering
* adds an overridable `getOrderDisambiguation` method to both node and edge orderings

This is used in https://github.com/github/codeql/pull/9891

This is currently only used in tests.

### DIFF
--- a/csharp/ql/lib/semmle/code/csharp/controlflow/internal/ControlFlowGraphImplShared.qll
+++ b/csharp/ql/lib/semmle/code/csharp/controlflow/internal/ControlFlowGraphImplShared.qll
@@ -916,7 +916,7 @@ module TestOutput {
             s
             order by
               l.getFile().getBaseName(), l.getFile().getAbsolutePath(), l.getStartLine(),
-              l.getStartColumn(), l.getEndLine(), l.getEndColumn(), t.toString()
+              l.getStartColumn(), l.getEndLine(), l.getEndColumn(), t.toString(), s.toString()
           )
       ).toString()
   }
--- a/csharp/ql/lib/semmle/code/csharp/controlflow/internal/ControlFlowGraphImplShared.qll
+++ b/csharp/ql/lib/semmle/code/csharp/controlflow/internal/ControlFlowGraphImplShared.qll
@@ -881,7 +881,12 @@ import Cached
  * graph is restricted to nodes from `RelevantNode`.
  */
 module TestOutput {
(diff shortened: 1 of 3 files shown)


nickrolfe:
Do I understand correctly that you've observed CFG nodes that have two or more successors with the exact same locations and the same `SuccessorType`, but are distinguished by their `toString()`?

nickrolfe (dismissed):
Either way, the change looks fine, and I can see for all three languages it's only used by the CFG tests, so I don't think we need a change note.

redsun82:
I was seeing flakiness in these lines on a branch of mine https://github.com/github/codeql/blob/2bc6c987c9c7cdd2c1bf5880d10625ccca1d7638/swift/ql/test/library-tests/controlflow/graph/Cfg.expected#L1514-L1535 with the successors changing order betwen runs. This change seemed to help, but now I have started seeing flakiness again, although this time it might rather be related to graph node ordering. I see an alternation between
```
#  141| =~ ...
#-----|  -> 0

#  141| =~ ...
#-----| no-match -> =~ ...
#-----| match -> true

#  141| 1
#-----|  -> =~ ...

#  141| =~ ...
#-----|  -> 1

#  141| =~ ...
#-----| match -> true
#-----| no-match -> case ...
```
and
```
#  141| =~ ...
#-----| no-match -> =~ ...
#-----| match -> true

#  141| =~ ...
#-----|  -> 0

#  141| 1
#-----|  -> =~ ...

#  141| =~ ...
#-----| match -> true
#-----| no-match -> case ...

#  141| =~ ...
#-----|  -> 1
```

I'll research this a bit more maybe before merging this

nickrolfe:
It looks like you have multiple nodes at line 141 (with the same start/end line/column) that all have `=~ ...` as their `toString()` result. So I'm not surprised you're seeing flakiness - they are equal for all the `order by` expressions in the `nodes` predicate.

michaelnebel (dismissed):
LGTM!

redsun82:
As the edge ordering was not enough to solve the flakiness I had, I added a further customizable string key. In https://github.com/github/codeql/pull/9891 I use it with `getPrimaryQlClasses` which seems to finally solve my flakiness issues.

michaelnebel (approved):
LGTM

PR #9909, opened 2022-07-28
Swift: Understand String.utf8.count etc in the string length conflation CVE query

geoffw0:
Understand `String.utf8.count`, `String.utf16.count` and `String.unicodeScalars.count` in the unicode / string length conflation CVE query.  It turns out `String.utf16.count` is exactly the same as `NSString.length` (e.g. scroll down to some of the examples [here](https://talk.objc.io/episodes/S01E80-swift-string-vs-nsstring)) - but apart from that each is a unique thing and any conflation should be flagged.


geoffw0:
... I had to merge `main` into this and got something wrong, give me a moment to fix it ...

geoffw0:
... all fixed (a predicate had gone missing in the merge; probably my fault for fixing a merge conflict badly).

PR #9912, opened 2022-07-28
Swift: make trap key prefixes readable

redsun82:
This replaces numeric tag-based prefixes with the actual tag name.
While this means in general slightly larger trap files, it aids
debugging them for a human.

In the future we can make this conditional on some kind of trap debug
option, but for the moment it does not seem detrimental.

### DIFF
--- a/swift/codegen/generators/trapgen.py
+++ b/swift/codegen/generators/trapgen.py
@@ -86,11 +86,10 @@ def generate(opts, renderer):
         renderer.render(cpp.TrapList(entries, opts.dbscheme), out / dir / "TrapEntries")
 
     tags = []
-    for index, tag in enumerate(toposort_flatten(tag_graph)):
+    for tag in toposort_flatten(tag_graph):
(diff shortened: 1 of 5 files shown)


redsun82:
For reference, here's how a DB inconsistency from an integration test looks with this change:
```
[INVALID_KEY_SET] predicate iterable_decl_context_members(@iterable_decl_context id, int index, @decl member): The key set {id, index} does not functionally determine all fields.
Here is a pair of tuples that agree on the key set but differ at index 2:
Tuple 1 in row 4525: (77649,2,77699)
Tuple 2 in row 4526: (77649,2,230376)
	Relevant element: Tuple 1: id=77649
		Full ID for 77649: @"StructDecl_So10HeapObjectV"
	Relevant element: Tuple 2: id=77649
		Full ID for 77649: @"StructDecl_So10HeapObjectV"
```
The `StructDecl` instead of a number is the difference.

AlexDenisov (approved):
Love it!

Issue #3710, opened 2020-06-13
Taint is stuck in exception variable declaration of a catch clause: `catch(Exception e)`


intrigus-lgtm:
**Description of the issue**
Taint can not flow out of the caught exceptions in catch clauses.
E.g.
```java
try {
} catch(Exception e) {
System.out.println(e.getMessage());
}
```
Here taint will not flow to `e.getMessage()`.

```codeql
import java
import semmle.code.java.dataflow.TaintTracking

class T extends TaintTracking::Configuration {
  T() { this = "T" }

  override predicate isSource(DataFlow::Node source) {
    exists(CatchClause c | c.getVariable() = source.asExpr())
  }

  override predicate isSink(DataFlow::Node sink) { any() }
}

from DataFlow::Node source, DataFlow::Node sink, T t
where t.hasFlow(source, sink)
select source, sink
```

Link to query:
https://lgtm.com/query/3909102737996232443/

aschackmull:
Thanks for reporting this. We are indeed missing a flow-edge here.  In the meantime you may use an access of the variable as your source definition instead to get the relevant flow:
```
exists(CatchClause c | c.getVariable().getAnAccess() = source.asExpr())
```

PR #9914, opened 2022-07-28
Java: Add support for data flow through thrown exceptions.

aschackmull:
This adds support for exception flow.

A test is included that shows some of what works and what doesn't.
Currently we get spurious flow due to imprecisions in type pruning.
There's also some missing, since I've ignored MaD-synthesised method bodies for now.


intrigus-lgtm:
Will this also fix https://github.com/github/codeql/issues/3710?

aschackmull:
> Will this also fix #3710?

Excellent question. I am indeed introducing dedicated data flow nodes corresponding to variable declarations in catch clauses along with (among others) the outgoing edges you ask for in the issue. However, I was not initially planning to expose those nodes, as I hadn't yet imagined use-cases. Do you think it is useful to be able to e.g. declare such a variable declarations as a source in the way that you do in the issue? With the changes in this PR I'd hoped that you didn't need to.

intrigus-lgtm:
> > Will this also fix #3710?
> 
> Excellent question. I am indeed introducing dedicated data flow nodes corresponding to variable declarations in catch clauses along with (among others) the outgoing edges you ask for in the issue. However, I was not initially planning to expose those nodes, as I hadn't yet imagined use-cases. Do you think it is useful to be able to e.g. declare such a variable declaration as a source in the way that you do in the issue? With the changes in this PR I'd hoped that you didn't need to.

#3710 was originally created in response to the "Step 3: Errors and Exceptions" part of https://securitylab.github.com/ctf/codeql-and-chill/.

```java
try {
    parse(tainted);
} catch (Exception e) {
    sink(e.getMessage())
}
```
The goal was to catch flow from `parse` to `sink`, but the code of `parse` is _unavailable_.
Will something like this be modelable with this PR?
(I still have to say that this feels like a really rare use case)

aschackmull:
> #3710 was originally created in response to the "Step 3: Errors and Exceptions" part of https://securitylab.github.com/ctf/codeql-and-chill/.
> 
> ```java
> try {
>     parse(tainted);
> } catch (Exception e) {
>     sink(e.getMessage())
> }
> ```
> 
> The goal was to catch flow from `parse` to `sink`, but the code of `parse` is _unavailable_. Will something like this be modelable with this PR? (I still have to say that this feels like a really rare use case)

Yes. In the context of this PR, the way to do that would be to make `parse` a taint step that transfers taint from its argument to its returned exception. No need to match up the parse call with the enclosing try-catch, that's all something that the feature in this PR will handle (including the case when `parse` would be nested inside some wrapper call).

PR #9918, opened 2022-07-28
Ruby: Model Mime::Type

hmac:
`Mime::Type#match?` and `Mime::Type#=~` convert their argument to a regular expression, so they're relevant for queries like ReDoS. This PR adds modeling for `Mime::Type` so we recognise instances of it and consider arguments these two methods to be regular expressions.


ruby/ql/lib/codeql/ruby/Regexp.qll
```diff
@@ -92,20 +92,32 @@ private class ParsedStringRegExp extends RegExp {
...
+   * A node that is not a regular expression literal, but is used in places that
+   * may interpret it as one. Instances of this class are typically strings that
+   * flow to method calls like `RegExp.new`.
+   */
+  abstract class Range extends DataFlow::Node { }
+}
+
 /**
- * Holds if `source` may be interpreted as a regular expression.
+ * Nodes interpreted as regular expressions via various standard library methods.
```
aibaars, line 106: Let's address `The QLDoc for a class should start with 'A', 'An', or 'The'.` warning.

ruby/ql/lib/codeql/ruby/frameworks/ActionDispatch.qll
```diff
@@ -6,12 +6,55 @@
...
+   * Type summaries for the `Mime::Type` class, i.e. method calls that produce new
+   * `Mime::Type` instances.
+   */
+  private class MimeTypeTypeSummary extends ModelInput::TypeModelCsv {
+    override predicate row(string row) {
+      // package1;type1;package2;type2;path
+      row =
+        [
+          // Mime[type] : Mime::Type (omitted)
+          // Method names with brackets like [] cannot be represented in MaD.
```
aibaars, line 28: @asgerf Is there a way to escape `[` and `]` symbols in a member name?

aibaars (approved):
Looks good to me. Let's add a change note and fix the QLDoc warning.

aibaars (approved):

PR #9919, opened 2022-07-29
Ruby: ActiveRecord associations

hmac:
Add modelling for AR associations, so we recognise things like

```rb
user.posts.create(args)
```

as a method call that creates an AR record.

### DIFF
--- a/ruby/ql/lib/codeql/ruby/frameworks/ActiveRecord.qll
+++ b/ruby/ql/lib/codeql/ruby/frameworks/ActiveRecord.qll
@@ -618,7 +618,9 @@ private class ActiveRecordAssociationMethodCall extends DataFlow::CallNode {
       (
         assoc.isCollection() and
         (
-          this.getMethodName() = pluralize(model) + ["", "=", "<<"]
+          this.getMethodName() = pluralize(model) + ["", "="]
+          or
+          this.getMethodName() = "<<"
           or
           this.getMethodName() = model + ["_ids", "_ids="]
         )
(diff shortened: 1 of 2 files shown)


ruby/ql/lib/codeql/ruby/frameworks/ActiveRecord.qll
```diff
@@ -516,3 +516,175 @@ private module Persistence {
...
+private class ActiveRecordAssociationMethodCall extends DataFlow::CallNode {
+  ActiveRecordAssociation assoc;
+
+  ActiveRecordAssociationMethodCall() {
+    exists(string model | model = assoc.getTargetModelName() |
+      this.getReceiver().(ActiveRecordInstance).getClass() = assoc.getSourceClass() and
+      (
+        assoc.isCollection() and
+        (
+          this.getMethodName() = pluralize(model) + ["", "=", "<<"]
```
alexrford, line 621: It's not clear to me from the tests that the `others<<` method is registering as a an `ActiveRecordAssociationMethodCall` in practice - in the test file on line 29 we have:
```
author2.posts << post2
```
but the `.expected` only has a single result
```
| associations.rb:29:1:29:13 | call to posts | associations.rb:5:1:9:3 | Post |
```
for the call to `posts` - I would expect another one for the call to `posts<<`.

I think that we might parse this as a call to a method named `<<` where the receiver is the result of a call to `posts`? If this is the case, then moving `<<` to `ActiveRecordCollectionProxyMethodCall` should fix this. 

`=` is probably worth having a test case for as well, although I think that one does work as it's written here.

alexrford (commented):
Looks great, really nicely documented. Just one question/suggestion regarding `<<`.

alexrford (approved):
👍

PR #9920, opened 2022-07-29
Swift: put all the PCM traps into the same place

AlexDenisov:
This removes duplicated TRAPs for precompiled clang modules.

### DIFF
--- a/swift/extractor/SwiftExtractor.cpp
+++ b/swift/extractor/SwiftExtractor.cpp
@@ -17,6 +17,7 @@
 #include "swift/extractor/infra/TargetFile.h"
 
 using namespace codeql;
+using namespace std::string_literals;
 
 static void archiveFile(const SwiftExtractorConfiguration& config, swift::SourceFile& file) {
   if (std::error_code ec = llvm::sys::fs::create_directories(config.trapDir)) {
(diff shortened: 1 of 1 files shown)


swift/extractor/SwiftExtractor.cpp
```diff
@@ -53,12 +54,16 @@ static std::string getFilename(swift::ModuleDecl& module, swift::SourceFile* pri
...
-  // Several modules with different name might come from .pcm (clang module) files
-  // In this case we want to differentiate them
-  std::string filename = module.getModuleFilename().str();
-  filename += "-";
-  filename += module.getName().str();
-  return filename;
+  // PCM clang module
+  if (module.isNonSwiftModule()) {
+    // Several modules with different name might come from .pcm (clang module) files
+    // In this case we want to differentiate them
```
redsun82, line 60: this does not explain why we strip the directory. We should probably say `pcm` files come from caches that might be located in different places, but are unambigouously identified by the base filename

redsun82 (commented):
I was trying this out myself :smile: I would just put some more explanation in the comment, otherwise I'm ok with merging this!

swift/extractor/SwiftExtractor.cpp
redsun82: I need to rebase https://github.com/github/codeql/pull/9922 on this one any way, so we can maybe just merge this as it is and I'll update the comment in mine

redsun82 (approved):
:ship:

PR #9921, opened 2022-07-29
Swift: mangle TypeAliasDecls differently

AlexDenisov:


swift/extractor/visitors/DeclVisitor.cpp
```diff
@@ -296,6 +296,9 @@ std::string DeclVisitor::mangledName(const swift::ValueDecl& decl) {
   if (decl.getKind() == swift::DeclKind::Module) {
     return static_cast<const swift::ModuleDecl&>(decl).getRealName().str().str();
   }
+  if (decl.getKind() == swift::DeclKind::TypeAlias) {
+    return mangler.mangleEntity(&decl);
+  }
```
redsun82, line 301: I guess this warrants a comment explaining why we treat this differently. Even if we don't know the root cause requiring this, it'd be good to give a short description of the example requiring this (mentioning `CFXML` type aliases)

redsun82: or maybe, as `mangledName` is already on `swift::ValueDecl`, we can just ditch `mangleAnyDecl` and only use `mangleEntity`?

redsun82 (approved):
:ship:

PR #9923, opened 2022-07-29
C#: SQL Injection improvements for SQLite.

michaelnebel:
The content on this PR is to address the issue: https://github.com/github/codeql-csharp-team/issues/172
Following the link to the original ticket it is worth noticing that
1. The project referenced on the ticket doesn't build, which could lead to the missing results.
2. The project referenced on the ticket is fork of another project which does build and
    - The claimed missing SQL injection is found by our query even though we don't directly support the SQLite framework.
    - The claimed missing Second Order SQL injection is in parts of the code that isn't included in the build. That being said, the Second Order SQL injection query will not find this.

In this PR we
1. Provide better support for the SQLite framework.
   - Explicit summaries for this Framework has been made, including Adapters (which was not supported before).
   - Stubs for the SQLite framework has been included.
   - Coding examples with weaknesses discovered by the SQL injection query are included.
2. FileStream and StreamReader summaries.
   - Flow summaries for FileStream and StreamReader has been included.
   - Example similar to the Second Order SQL injection has been provided and will no be caught as a first order SQL injection, if the filename is tainted.
4. FileStreams are now considered StoredFlowSources.
   - FileStreams are now considered stored flow sources.
   - Second Order SQL injection now catches an example similar to the one provided in the referenced ticket.

A change note will be added, when the exact scope of the PR is decided.


csharp/ql/lib/semmle/code/csharp/frameworks/Sql.qll
```diff
@@ -67,18 +68,46 @@ private class IDbCommandConstructionSinkModelCsv extends SinkModelCsv {
...
 }
 
-/** A construction of an `SqlDataAdapter` object. */
+/** Data flow for SqlCommand and friends. */
+private class SqlCommandSummaryModelCsv extends SummaryModelCsv {
+  override predicate row(string row) {
+    row =
+      [
+        // SqlCommand
+        "System.Data.SqlClient;SqlCommand;false;SqlCommand;(System.String);;Argument[0];Argument[Qualifier];taint;manual",
```
hvitved, line 86: Why is the output for all constructors `Argument[Qualifier]` and not `ReturnValue`?

michaelnebel: Constructors are speciel in the way that they are considered to be side effect based, ie. that they manipulate `this`. As far as I remember the reasoning behind this is related to inheritance or maybe I am mistaken?

hvitved: For other constructors, like `Tuple<>`, we use `ReturnValue` (see https://github.com/github/codeql/blob/main/csharp/ql/lib/semmle/code/csharp/frameworks/System.qll#L1323).

michaelnebel: Oh no - you are right. We need to fix all those manual written summaries - I don't think they work at all. Constructors are considered to be side effect based (ie. they manipulate an implicit this argument).
Fortunately, the model generator already handles constructors correctly.
Should I make a tech debt PR and start to fix all the hand written summaries (we (and by we  - I mean I) must have converted constructors incorrectly when removing the legacy data flow library))?

hvitved: > Should I make a tech debt PR and start to fix all the hand written summaries

Yes, we should definitely fix those broken summaries. Thanks :-)

hvitved: As discussed offline, `Argument[Qualifier]` is arguably the right thing to write. Sorry about the confusion.

hvitved (changes requested):
Unit tests are failing; otherwise LGTM.

michaelnebel:
> Unit tests are failing; otherwise LGTM.

Yes, this is due to the rebase after the new stubs were generated. Will fix.
I will also add a release note and start DCA.

hvitved (dismissed):

csharp/ql/src/change-notes/2022-08-10-sqlinjection-queries.md
```diff
@@ -0,0 +1,5 @@
+---
+category: minorAnalysis
+---
+* Added better support for the SQ Lite framework in the SQL injection query.
```
tamasvajk, line 4: Nitpicking: There's an extra space in `SQ Lite`.

michaelnebel: Thanks for catching it! Will update it!

michaelnebel:
Lots of projects failed in DCA. Will re-execute DCA as there is an alerts diff on the projects that succeeded (which is not totally unexpected).

michaelnebel:
Comments on the DCA execution
- There are no performance regressions.
- The failure of ASP.NET Core is most likely unrelated (it fails with out of memory - the same was the case for the nightly run)
- The new alerts are all for queries that use StoredFlowSource as a source of taint and the first step is opening a FileStream, which makes sense, since we are now considered a file a StoredFlowSource. Are these alerts something we want?

hvitved:
> * The new alerts are all for queries that use StoredFlowSource as a source of taint and the first step is opening a FileStream, which makes sense, since we are now considered a file a StoredFlowSource. Are these alerts something we want?

AFAICT, there are no new results on WebGoat. Were we expecting that?

michaelnebel:
> > * The new alerts are all for queries that use StoredFlowSource as a source of taint and the first step is opening a FileStream, which makes sense, since we are now considered a file a StoredFlowSource. Are these alerts something we want?
> 
> AFAICT, there are no new results on WebGoat. Were we expecting that?

No, we were not expecting more results.
(1) The first order SQL injection result was already found without any changes.
(2) The second SQL injection result will not be a found as the references code is not included in the compilation of the project.

hvitved:
> the references code is not included in the compilation of the project

🤦

hvitved (approved):

PR #9924, opened 2022-07-29
Revert "Swift: deduplicate `VarDecl`"

AlexDenisov:
Reverts github/codeql#9911

Reverting for now as it crashes in certain cases when mangles some variables coming from PCMs.


AlexDenisov:
Better workaround https://github.com/github/codeql/pull/9925

PR #9928, opened 2022-07-29
JS: Simplify type hierarchy for SourceNode

asgerf:
The charpred caused spurious type tests to appear in the DIL.

On `main` the problem is not so severe, but I have seen it can cause bigger issues during development.


asgerf:
[Evaluation](https://github.com/github/codeql-dca-main/tree/data/asgerf/source-node-typ__default__code-scanning/reports) looks fine

erik-krogh (approved):

PR #9932, opened 2022-07-31
Ruby: RBI library changes to support models-as-data model generation

alexrford:
These changes are mostly in service of better parameter type modelling - they should make more sense in the context of the models-as-data manager PR for a Ruby typegraph implementation.

### DIFF
--- a/ruby/ql/lib/codeql/ruby/ast/Constant.qll
+++ b/ruby/ql/lib/codeql/ruby/ast/Constant.qll
@@ -294,6 +294,11 @@ class ConstantReadAccess extends ConstantAccess {
   Expr getValue() { result = getConstantReadAccessValue(this) }
 
   final override string getAPrimaryQlClass() { result = "ConstantReadAccess" }
+
+  /**
+   * Gets a fully qualified name for this constant.
+   */
+  string getAQualifiedName() { TResolved(result) = resolveConstantReadAccess(this) }
 }
 
 /**
--- a/ruby/ql/lib/codeql/ruby/ast/Constant.qll
+++ b/ruby/ql/lib/codeql/ruby/ast/Constant.qll
@@ -208,11 +208,6 @@ class ConstantAccess extends Expr, TConstantAccess {
    */
   predicate hasGlobalScope() { none() }
 
-  /**
-   * Gets a fully qualified name for this constant.
-   */
-  string getAQualifiedName() { none() }
-
   override string toString() { result = this.getName() }
 
   override AstNode getAChild(string pred) {
@@ -299,8 +294,6 @@ class ConstantReadAccess extends ConstantAccess {
   Expr getValue() { result = getConstantReadAccessValue(this) }
 
   final override string getAPrimaryQlClass() { result = "ConstantReadAccess" }
-
-  override string getAQualifiedName() { TResolved(result) = resolveConstantReadAccess(this) }
 }
 
 /**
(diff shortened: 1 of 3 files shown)


ruby/ql/lib/codeql/ruby/ast/Constant.qll
```diff
@@ -208,6 +208,11 @@ class ConstantAccess extends Expr, TConstantAccess {
    */
   predicate hasGlobalScope() { none() }
 
+  /**
+   * Gets a fully qualified name for this constant.
+   */
+  string getAQualifiedName() { none() }
```
aibaars, line 214: The change looks fine; however `getAQualifiedName` is not used in this pull request so I wonder why you are changing this here.

ruby/ql/lib/codeql/ruby/experimental/Rbi.qll
```diff
@@ -150,22 +150,14 @@ module Rbi {
...
 
-      /**
-       * Gets the type of a parameter of this type signature.
-       */
-      override ParameterType getAParameterType() {
-        result = this.getParamsCall().getAParameterType()
-      }
-      // TODO: get associated method to which this can be passed
+      // TODO: widen type for procs/blocks
+      override MethodBase getAssociatedMethod() { none() }
```
aibaars, line 160: Should this return `Callable` for procs/blocks/lambdas?

ruby/ql/lib/codeql/ruby/experimental/Rbi.qll
```diff
@@ -313,19 +311,64 @@ module Rbi {
...
-     * Gets the type of a parameter defined by this call.
+     * Gets the type of a positional parameter defined by this call.
      */
-    ParameterType getAParameterType() { result = this.getArgument(_) }
+    ParameterType getPositionalParameterType(int i) {
+      result = this.getArgument(i) and
+      // explicitly exclude keyword parameters
+      not this.getAssociatedParameter(result.getName()) instanceof KeywordParameter and
+      // and exclude block arguments
+      not isBlockParamName(result.getName())
```
aibaars, line 331: Shouldn't this be `not result instanceof BlockArgument` instead? Or is using `blk` and `block` the way that RBI models block arguments and parameters?

aibaars (commented):
Looks fine, but I have some questions still.

ruby/ql/lib/codeql/ruby/experimental/Rbi.qll
alexrford: Good point - I think in this case we have something like:
```
  sig do
    params(
      a: String,
      block: T.nilable(T.proc.params(b: Integer).void)
    ).void
  end
  def blk_nilable_method(a, &block); end
```
so the argument here is `block: T.nilable(T.proc.params(b: Integer).void)` which is not a `BlockArgument`, but the associated parameter is `&block` so we can check if that is a `BlockParameter` instead which seems neater and less fragile.

ruby/ql/lib/codeql/ruby/ast/Constant.qll
alexrford: Ah, I was using this in an earlier revision of https://github.com/github/codeql-model-manager/pull/5 to get qualified names for general `ConstantAccess`es, but changed how this is implemented so this predicate isn't required. I'll drop it from this PR.

ruby/ql/lib/codeql/ruby/experimental/Rbi.qll
alexrford: We'd probably want this to return a `BlockParameter`, e.g. in:
```
  sig do
    params(
      a: String,
      block: T.nilable(T.proc.params(b: Integer).void)
    ).void
  end
  def blk_nilable_method(a, &block); end
```
the `T.proc` call corresponds to the `&block` parameter of `blk_nilable_method`, whereas the `sig` call corresponds to the `blk_nilable_method` definition itself. So we maybe need something like `newtype TTypedCallable = TypedMethod(MethodBase m) or TypedBlock(BlockParameter b)` or maybe to add another predicate `BlockParameter getAssociatedBlockParameter()` to the `SignatureCall` abstract class and in practice only one of `getAssociatedMethod()` and `getAssociatedBlockParameter()` will have a value.

aibaars (approved):
Looks good to me.

PR #9936, opened 2022-08-01
Use 'gh codeql' with the nightly release for CI jobs

aibaars:
This pull request changes the default CodeQL version used by CI jobs to be the `nightly` release. Before we used the latest stable release which could lag behind the bleeding edge main branch of CodeQL quite a bit causing friction.


hvitved (approved):
Makes sense to me.

PR #9942, opened 2022-08-01
C++: Remove internal `microsoft` tags from queries

MathiasVP:
This PR removes the internal `microsoft` tags on the remaining seven queries. Now, `grep`'ing for `"external/microsoft"` gives no more results in `.ql` files.


jketema (commented):
LGTM. I assume you'll notify the MS contact before we merge this?

MathiasVP:
> LGTM. I assume you'll notify the MS contact before we merge this?

Yep. I'm writing them an email now.

jketema (approved):
Got the ok from MS.

PR #9945, opened 2022-08-02
Java: Improve join-order.

aschackmull:
Minor join-order improvement. This was taking a lot of slots in the join-order badness reports.


atorralba (approved):
Looks plausible to me as long as DCA is happy.

PR #9947, opened 2022-08-02
Go: note that numeric-typed nodes can't cause path traversal

smowton:


go/ql/lib/semmle/go/security/TaintedPathCustomizations.qll
```diff
@@ -70,6 +70,15 @@ module TaintedPath {
     PathAsSink() { this = any(FileSystemAccess fsa).getAPathArgument() }
   }
 
+  /**
+   * A numeric-typed node, considered a sanitizer for path traversal.
```
intrigus-lgtm, line 74: Nitpick:
Is `Bool` considered a numeric type in Go land?
If not, isn't the class name and the doc a bit misleading?

atorralba (approved):

PR #9949, opened 2022-08-02
JS: Change how TRAP cache is configured

edoardopirovano:
This PR refactors how we configure the TRAP cache for the JavaScript extractor. Previously, we had two different mechanisms:

1. In the autobuilder, via the `LGTM_TRAP_CACHE` environment variables.
2. In a direct invocation of the extractor, via the `--trap-cache` flags.

As far as I can tell, neither of these mechanisms is actually in use at the moment. So, before we create new uses for it, let's take the opportunity to refactor both those code paths to flow through the same place, and be configured by the new extractor options functionality documented [here](https://codeql.github.com/docs/codeql-cli/extractor-options/), which is how the Action's upcoming TRAP caching feature will configure the cache.

As part of this, I've also added a new option that configures whether the cache should be written to. This will be needed in the new usage of this cache in the Action, where we will distinguish runs on `main` (which will update the cache) and runs on PRs (which may read a cache, but will never bother writing to it since it will be discarded).

### DIFF
--- a/javascript/extractor/src/com/semmle/js/extractor/trapcache/ITrapCache.java
+++ b/javascript/extractor/src/com/semmle/js/extractor/trapcache/ITrapCache.java
@@ -1,7 +1,11 @@
 package com.semmle.js.extractor.trapcache;
 
+import static com.semmle.js.extractor.ExtractorOptionsUtil.readExtractorOption;
+
 import com.semmle.js.extractor.ExtractorConfig;
 import com.semmle.js.extractor.FileExtractor;
+import com.semmle.js.extractor.Main;
+import com.semmle.util.exception.UserError;
 import java.io.File;
 
 /** Generic TRAP cache interface. */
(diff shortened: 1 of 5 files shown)


esbena:
>  (which may read a cache, but will never bother writing to it since it will be discarded).

It sounds like we deliberately prevent ourselves from using "our own" cache on a PR where codeql is run multiple times. Is that cache effect considered insignificant compared to the cache on main?

esbena:
Great, thanks for confirming and explaining.

The implementation LGTM, but I'll defer to @asgerf for the final decision.

Long shot: perhaps @max-schaefer has some LGTM_TRAP_CACHE usage that we do not know about.

max-schaefer:
> Long shot: perhaps @max-schaefer has some LGTM_TRAP_CACHE usage that we do not know about.

I do not, but many thanks for considering it!

edoardopirovano:
https://github.com/github/codeql-action/pull/1172 is the PR on the Action that will start hitting this code path. It's gated behind a feature flag.

javascript/extractor/src/com/semmle/js/extractor/trapcache/ITrapCache.java
```diff
@@ -18,4 +22,29 @@ public interface ITrapCache {
...
   public File lookup(String source, ExtractorConfig config, FileExtractor.FileType type);
+
+  /**
+  * Build a TRAP cache as defined by the extractor options, which are read from the corresponding
+  * environment variables as defined in
+  * https://github.com/github/codeql-core/blob/main/design/spec/codeql-extractors.md
+  *
+ * @return a TRAP cache
+ */
+  public static ITrapCache fromExtractorOptions() {
```
asgerf, line 33:
```suggestion
  *
  * @return a TRAP cache
  */
  public static ITrapCache fromExtractorOptions() {
```
Not sure what happened to the indentation there

asgerf (approved):
LGTM, just a weird indentation issue

javascript/extractor/src/com/semmle/js/extractor/trapcache/ITrapCache.java
edoardopirovano: Good spot, fixed.

asgerf (approved):

PR #9953, opened 2022-08-02
Update tree-sitter-ruby

aibaars:
See: 
* https://github.com/tree-sitter/tree-sitter-ruby/pull/224
* https://github.com/tree-sitter/tree-sitter-ruby/pull/225


nickrolfe (commented):
Looks quite reasonable (and like it was quite a pain to figure out the upgrades!)

I've started a new DCA run, since you modified the stats since the last one.

I trust you did the usual trick of manually testing the downgrades and upgrades by running the tests with them?

nickrolfe (approved):

PR #9956, opened 2022-08-02
Make java/path-injection recognise create-file MaD sinks

smowton:


java/ql/src/Security/CWE/CWE-022/TaintedPath.ql
```diff
@@ -44,9 +49,7 @@ class TaintedPathConfig extends TaintTracking::Configuration {
...
 }
 
-from DataFlow::PathNode source, DataFlow::PathNode sink, PathCreation p, TaintedPathConfig conf
-where
-  sink.getNode().asExpr() = p.getAnInput() and
-  conf.hasFlowPath(source, sink)
-select p, source, sink, "$@ flows to here and is used in a path.", source.getNode(),
+from DataFlow::PathNode source, DataFlow::PathNode sink, TaintedPathConfig conf
+where conf.hasFlowPath(source, sink)
+select sink, source, sink, "$@ flows to here and is used in a path.", source.getNode(),
```
atorralba, line 54: Will this change in the `select` columns dismiss all existing alerts and create new ones for projects that already have results?

smowton: I suppose so -- I can have this report a `Top` so I can preserve the existing reporting location

atorralba: This query needs an extensive refactor anyway, so I'm not against doing that for the time being. I think it's better than cause friction with all the alert changes.

smowton: Amended as suggested

atorralba (dismissed):
LGTM now, thanks for the change. I guess at some point we should add tests for all the MaD `create-file` sinks though.

Edit: needs a change note.

smowton:
Yeah, was surprised to find they're all untested. I won't take that on right here :)

atorralba (approved):

PR #9963, opened 2022-08-03
Model `java.util.Properties.setProperty`

intrigus-lgtm:
Models `java.util.Properties.setProperty` as mentioned in https://github.com/github/codeql/pull/9817#issuecomment-1183619419


intrigus-lgtm:
I don't know whether this deserves a change note and whether this could be merged with the existing change note ([2022-07-13-properites.md](https://github.com/github/codeql/pull/9817/files#diff-a17ad305a88c8668d2027c7bdc6d9040272de64507161b1979d890f9eb404cfa)) as the change note seems to have already been released (https://github.com/github/codeql/blob/main/java/ql/lib/change-notes/released/0.3.1.md#minor-analysis-improvements).
Note that the existing change note both has a typo in the filename "properites" and in the content "java.util.Properites".

java/ql/test/library-tests/dataflow/collections/Test.java
```diff
@@ -84,7 +84,17 @@ public void run4() {
...
     sink(p.getProperty("key", "defaultValue")); // Flow
-    
+
+    Properties clean = new Properties();
+    sink(clean.getProperty("key", tainted)); // Flow
+  }
+
+  public void run5() {
+    Properties p = new Properties();
+    p.setProperty("key", tainted);
```
smowton, line 94: You should test that the return value of setProperty returns taint from the map too

smowton (commented):
Can't merge change notes, so yes give this a new one

java/ql/test/library-tests/dataflow/collections/Test.java
intrigus-lgtm, line 94: Like this?
```suggestion
    sink(p.setProperty("key", tainted)); // Flow
```

java/ql/test/library-tests/dataflow/collections/Test.java
```diff
@@ -84,7 +84,17 @@ public void run4() {
     p.put("key", tainted);
```
intrigus-lgtm, line 84: This method is similar to `setProperty`, should I also change it like this?
```suggestion
    sink(p.put("key", tainted)); // Flow
```

java/ql/test/library-tests/dataflow/collections/Test.java
smowton: I believe it returns the *old* value, so you actually want to store some taint, then store something clean and check that the return value of the clean store shows taint

java/ql/test/library-tests/dataflow/collections/Test.java
intrigus-lgtm: Done in https://github.com/github/codeql/pull/9963/commits/8eebe710ce568c49f57c51d958b211e4e0e4fa63

java/ql/test/library-tests/dataflow/collections/Test.java
intrigus-lgtm: Yes, you're right, `setProperty` returns the *old* value.
Done in https://github.com/github/codeql/pull/9963/commits/8eebe710ce568c49f57c51d958b211e4e0e4fa63

intrigus-lgtm: Done in https://github.com/github/codeql/pull/9963/commits/8eebe710ce568c49f57c51d958b211e4e0e4fa63

java/ql/test/library-tests/dataflow/collections/Test.java
```diff
@@ -84,8 +84,30 @@ public void run4() {
...
     sink(p.getProperty("key")); // Flow
     sink(p.getProperty("key", "defaultValue")); // Flow
-    
+
+    Properties clean = new Properties();
+    sink(clean.getProperty("key", tainted)); // Flow
```
smowton, line 89: This is accidentally duplicated on lines 98-99

intrigus-lgtm: Deduplicated :)

smowton (approved):

PR #9964, opened 2022-08-03
Swift: Query for CWE-79 / CWE-95

geoffw0:
Query for CWE-79 / CWE-95 / very loosely equivalent of MobSF `ios_load_html_string`.  The issue is that calling `UIWebView.loadHTMLString` or `WKWebView.loadHTMLString` (or a similar method) without specifying a base URL leaves the app potentially open to cross-site scripting attacks and extraction of data from the local file system.
```
webview.loadHTMLString(htmlData, baseURL: nil) // BAD
```

---

The query only reports cases where the `baseURL` is `nil` _and_ taint reaches `htmlData`.
 - if the `baseURL` is `nil` but the `htmlData` is not tainted, it seems unlikely an attacker would be able to get a URL they control resolved.
   - ideally we would be quite inclusive about the definition of 'taint' here (e.g. data from a database might be a likely source for a practical attack).
 - on the flip side if the `htmlData` is tainted but the `baseURL` is specified, the app could be vulnerable to attack but it seems just as likely its doing something complicated and dynamic with user data in a correct and safe way.  I'm not very keen to flag these cases.

---

The query does reasonably well in the tests, though there are certainly improvements to be made.  There are also taint extensions (`isAdditionalTaintStep`) that should be generalized to all Swift taint flow.  I will create follow-up issues to address these improvements to the libraries.


swift/ql/src/queries/Security/CWE-079/UnsafeWebViewFetch.ql
```diff
@@ -0,0 +1,143 @@
...
+import codeql.swift.dataflow.TaintTracking
+import codeql.swift.dataflow.FlowSources
+import DataFlow::PathGraph
+import codeql.swift.frameworks.StandardLibrary.String
+
+/**
+ * A taint source that is `String(contentsOf:)`.
+ * TODO: this shouldn't be needed when `StringSource` in `String.qll` is working.
+ */
+class StringContentsOfURLSource extends RemoteFlowSource {
```
MathiasVP, line 26: Oh, right. I see why you need this class. This is essentially the same issue as in https://github.com/github/codeql/pull/9673#discussion_r905171464. I have a fix for our MaD library so that we can delete this class, I'll put up a PR that fixes this once we've merged this PR.

swift/ql/src/queries/Security/CWE-079/UnsafeWebViewFetch.ql
```diff
@@ -0,0 +1,143 @@
...
+    // TODO: this should probably be part of DataFlow / TaintTracking.
+    node1.asExpr() = node2.asExpr().(AnyTryExpr).getSubExpr()
+    or
+    // allow flow through `!`
+    // TODO: this should probably be part of DataFlow / TaintTracking.
+    node1.asExpr() = node2.asExpr().(ForceValueExpr).getSubExpr()
+    or
+    // allow flow through string concatenation.
+    // TODO: this should probably be part of TaintTracking.
+    node2.asExpr().(AddExpr).getAnOperand() = node1.asExpr()
```
MathiasVP, line 114: The first two should definitely be taint steps that are included in `TaintTracking`, yeah.

I also think the string concatenation rule should be a taint step in `TaintTracking`. Maybe it should only be a taint step when the other operand is a constant string?

swift/ql/test/query-tests/Security/CWE-079/UnsafeWebViewFetch.swift
```diff
@@ -0,0 +1,217 @@
...
+
+	webview.loadHTMLString(localString, baseURL: nil) // GOOD: the HTML data is local
+	webview.loadHTMLString(getRemoteData(), baseURL: nil) // BAD: HTML contains remote input, may access local secrets
+	webview.loadHTMLString(remoteString, baseURL: nil) // BAD
+
+	webview.loadHTMLString("<html>" + localStringFragment + "</html>", baseURL: nil) // GOOD: the HTML data is local
+	webview.loadHTMLString("<html>" + remoteString + "</html>", baseURL: nil) // BAD
+
+	webview.loadHTMLString("<html>\(localStringFragment)</html>", baseURL: nil) // GOOD: the HTML data is local
+	webview.loadHTMLString("<html>\(remoteString)</html>", baseURL: nil) // BAD [NOT DETECTED]
```
MathiasVP, line 127: I'm not surprised this isn't caught, currently. We do extract the relevant parts, but it's not hooked up properly in the CFG/dataflow library yet ([here's](https://github.com/github/codeql-c-team/issues/1078) the internal issue for this).

Once that library issue is resolved, I imagine this missing flow will appear in this query.

swift/ql/src/queries/Security/CWE-079/UnsafeWebViewFetch.ql
```diff
@@ -0,0 +1,143 @@
...
+where
+  config.hasFlowPath(sourceNode, sinkNode) and
+  sink = sinkNode.getNode() and
+  (
+    // base URL is nil
+    sink.getBaseURL() instanceof NilLiteralExpr and
+    message = "Tainted data is used in a WebView fetch without restricting the base URL."
+    or
+    // base URL is tainted
+    config.hasFlow(_, any(DataFlow::Node n | n.asExpr() = sink.getBaseURL())) and
```
MathiasVP, line 140: Dataflow configurations actually have a couple of convenient wrappers around `hasFlow`, and this is a great place to use one of those.
```suggestion
    config.hasFlowToExpr(sink.getBaseURL()) and
```

swift/ql/src/queries/Security/CWE-079/UnsafeWebViewFetch.ql
```diff
@@ -0,0 +1,143 @@
...
+    message = "Tainted data is used in a WebView fetch with a tainted base URL."
+  )
+select sinkNode, sourceNode, sinkNode, message
```
MathiasVP, line 143: We normally use a `DataFlow::Node` as the first column (and not the `DataFlow::PathNode`).
```suggestion
select sink, sourceNode, sinkNode, message
```

swift/ql/src/queries/Security/CWE-079/UnsafeWebViewFetch.ql
```diff
@@ -0,0 +1,143 @@
...
+    // TODO: this should probably be part of TaintTracking.
+    node2.asExpr().(AddExpr).getAnOperand() = node1.asExpr()
+    or
+    // allow flow through `URL.init`.
+    exists(CallExpr call, ClassDecl c, AbstractFunctionDecl f |
+      c.getName() = "URL" and
+      c.getAMember() = f and
+      f.getName() = ["init(string:)", "init(string:relativeTo:)"] and
+      call.getFunction().(ApplyExpr).getStaticTarget() = f and
+      node1.asExpr() = call.getArgument(_).getExpr() and
```
MathiasVP, line 122:
```suggestion
      node1.asExpr() = call.getAnArgument().getExpr() and
```

MathiasVP (commented):
This looks great! Mostly some stylistic things. Many of my comments are just "You're right. This is a library thing that needs to be fixed". I can do those pretty quickly once this PR is merged. Do you want to run a DCA for this as well before we merge it?

swift/ql/src/queries/Security/CWE-079/UnsafeWebViewFetch.ql
```diff
@@ -0,0 +1,143 @@
...
+  override predicate isSource(DataFlow::Node node) { node instanceof RemoteFlowSource }
+
+  override predicate isSink(DataFlow::Node node) {
+    node instanceof Sink or
+    node.asExpr() = any(Sink s).getBaseURL()
+  }
+
+  override predicate isAdditionalTaintStep(DataFlow::Node node1, DataFlow::Node node2) {
+    // allow flow through `try!` and similar constructs
+    // TODO: this should probably be part of DataFlow / TaintTracking.
```
geoffw0, line 105: Issue created for this here: https://github.com/github/codeql-c-team/issues/1108

swift/ql/src/queries/Security/CWE-079/UnsafeWebViewFetch.ql
geoffw0: Fantastic!  I was going to create an issue for this, but it sounds like you're already on it and that would just be a duplicate, so I'll l eave it in your hands.

swift/ql/src/queries/Security/CWE-079/UnsafeWebViewFetch.ql
geoffw0: I'm actually suspicious flow through `try` and `!` should be *data* flow, never mind taint flow.

Why do you think we might not want taint flow through appending non-constant strings?

swift/ql/test/query-tests/Security/CWE-079/UnsafeWebViewFetch.swift
geoffw0: I've made it part of https://github.com/github/codeql-c-team/issues/1108, though obviously it depends on the issue you cited.

swift/ql/src/queries/Security/CWE-079/UnsafeWebViewFetch.ql
geoffw0: I just created https://github.com/github/codeql-c-team/issues/1108 to track this by the way, so ideally continue discussion of future taint steps on there.

swift/ql/src/queries/Security/CWE-079/UnsafeWebViewFetch.ql
geoffw0: Done.

swift/ql/src/queries/Security/CWE-079/UnsafeWebViewFetch.ql
geoffw0: Done.

swift/ql/src/queries/Security/CWE-079/UnsafeWebViewFetch.ql
geoffw0: Done, thanks!

swift/ql/src/queries/Security/CWE-079/UnsafeWebViewFetch.ql
```diff
@@ -0,0 +1,143 @@
...
+ * for this query.
+ */
+class UnsafeWebViewFetchConfig extends TaintTracking::Configuration {
+  UnsafeWebViewFetchConfig() { this = "UnsafeWebViewFetchConfig" }
+
+  override predicate isSource(DataFlow::Node node) { node instanceof RemoteFlowSource }
+
+  override predicate isSink(DataFlow::Node node) {
+    node instanceof Sink or
+    node.asExpr() = any(Sink s).getBaseURL()
```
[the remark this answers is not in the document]
geoffw0: It was a conflict caused by applying a suggestion after making another change.  Fixed.

swift/ql/test/query-tests/Security/CWE-079/UnsafeWebViewFetch.swift
```diff
@@ -0,0 +1,217 @@
...
+func testSimpleFlows() {
+	let webview = UIWebView()
+
+	webview.loadHTMLString(try! String(contentsOf: URL(string: "http://example.com/")!), baseURL: nil) // BAD [NOT DETECTED]
+
+	let data = try! String(contentsOf: URL(string: "http://example.com/")!)
+	webview.loadHTMLString(data, baseURL: nil) // BAD [NOT DETECTED]
+
+	let url = URL(string: "http://example.com/")
+	webview.loadHTMLString(try! String(contentsOf: url!), baseURL: nil) // BAD [NOT DETECTED]
```
geoffw0, line 109: I've created https://github.com/github/codeql-c-team/issues/1110 to debug these three omissions.  Likely an issue with DataFlow for Swift, rather than an issue in this query.

MathiasVP: Hm, indeed that is odd. I can take a look at this after this PR is merged.

MathiasVP:
Also, I'm happy to merge this as soon as CI is happy. I don't think we need to wait for a DCA run.

MathiasVP (approved):
LGTM!

PR #9966, opened 2022-08-04
C#: Disable CLR tracer

hvitved:
~Also remove old tracer configs, as we now use the Lua tracer.~


csharp/tools/linux64/compiler-tracing.spec
```diff
@@ -1,16 +0,0 @@
-**/mcs.exe:
```
criemen, line 1: Can we keep those files around for a bit longer?
Right now, we still have the option to switch to the legacy tracer, either via a command-line flag, or via a code-scanning wide feature flag. For that, we still need the legacy configs for some weeks longer.
The whole thing is only as a backup, in case we have some fatal issue with Lua tracing.

criemen (approved):

michaelnebel (approved):
LGTM!

PR #9970, opened 2022-08-04
Java: Improve performance of ConfusingOverloading.

aschackmull:
The
```
  exists(RefType commonSubtype | hasSubtypeOrInstantiation*(t1, commonSubtype) |
    hasSubtypeOrInstantiation*(t2, commonSubtype)
  )
```
construction is very expensive, so this change magics that as much as possible, including by noting that the above snippet implies `haveIntersection(t1, t2)` (the converse doesn't hold).

There's also a minor bugfix to account for `t1 = t2` for primitive types.


aschackmull:
dca looks good.

atorralba (approved):
Looks plausible to me.

PR #9971, opened 2022-08-04
C++: Handle all forms of constant attribute arguments

jketema:


cpp/ql/lib/semmle/code/cpp/Specifier.qll
```diff
@@ -256,9 +256,13 @@ class AttributeArgument extends Element, @attribute_arg {
 
   /**
    * Gets the text for the value of this argument, if its value is
-   * a string or a number.
+   * a constant or token.
```
MathiasVP, line 259: I'd probably write this as `or a token`, but I won't force you to do another submodule bump just for that. So let's just wait with that 😂.

MathiasVP (approved):
LGTM!

cpp/ql/lib/semmle/code/cpp/Specifier.qll
jketema: Thanks

PR #9975, opened 2022-08-05
Java: Improve AsyncTask data flow support

atorralba:
Uses jump steps to model the life-cycle described here: https://developer.android.com/reference/android/os/AsyncTask\#the-4-steps

### DIFF
--- a/java/ql/lib/semmle/code/java/frameworks/android/AsyncTask.qll
+++ b/java/ql/lib/semmle/code/java/frameworks/android/AsyncTask.qll
@@ -77,8 +77,6 @@ private class AsyncTaskInit extends Callable {
 
 /** A call to the `execute` or `executeOnExecutor` methods of the `android.os.AsyncTask` class. */
 private class ExecuteAsyncTaskMethodAccess extends MethodAccess {
-  Argument paramsArgument;
-
   ExecuteAsyncTaskMethodAccess() {
     this.getMethod().hasName(["execute", "executeOnExecutor"]) and
     this.getMethod().getDeclaringType().getSourceDeclaration().getASourceSupertype*() instanceof
--- a/java/ql/lib/semmle/code/java/frameworks/android/AsyncTask.qll
+++ b/java/ql/lib/semmle/code/java/frameworks/android/AsyncTask.qll
@@ -4,15 +4,58 @@ import java
 private import semmle.code.java.dataflow.DataFlow
 private import semmle.code.java.dataflow.FlowSteps
 
-/**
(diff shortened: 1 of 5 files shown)


java/ql/lib/semmle/code/java/frameworks/android/AsyncTask.qll
```diff
@@ -4,49 +4,103 @@ import java
...
   ExecuteAsyncTaskMethodAccess() {
-    exists(Method m |
-      this.getMethod() = m and
-      m.getDeclaringType().getSourceDeclaration().getASourceSupertype*() instanceof AsyncTask
-    |
-      m.getName() = "execute" and not m.isStatic() and paramsArgument = this.getArgument(0)
-      or
-      m.getName() = "executeOnExecutor" and paramsArgument = this.getArgument(1)
-    )
+    this.getMethod().hasName(["execute", "executeOnExecutor"]) and
```
smowton, line 83: Member `paramsArgument` is now unbound?

smowton (commented):
Looks plausible

java/ql/lib/semmle/code/java/frameworks/android/AsyncTask.qll
atorralba: You're absolutely right, thanks for catching this. Fixed in b75b073.

smowton (approved):

java/ql/lib/semmle/code/java/frameworks/android/AsyncTask.qll
```diff
@@ -4,49 +4,101 @@ import java
...
  */
 private class AsyncTask extends RefType {
   AsyncTask() { this.hasQualifiedName("android.os", "AsyncTask") }
 }
 
+/** The constructor or initializer method of the `android.os.AsyncTask` class. */
+private class AsyncTaskInit extends Callable {
+  AsyncTaskInit() {
+    this.getDeclaringType().getSourceDeclaration().getASourceSupertype*() instanceof AsyncTask and
+    (this instanceof Constructor or this instanceof InitializerMethod)
```
intrigus-lgtm, line 74: What is `or this instanceof InitializerMethod` doing?
If I remove it, the tests still seem to pass.

atorralba: Good point, I was missing a test for the initializer method. Thanks for bringing this up! See 03b854a.

atorralba:
@smowton can you re-approve? I only added a test case.

smowton (approved):

PR #9976, opened 2022-08-05
Ruby: Simplify flow summaries for hash literals

hvitved:
The existing summaries could be simplified using the models-as-data token `Argument[hash-splat]`, which refers to both keyword arguments wrapped in an implicit hash, as well as explicit hash splat arguments.


hvitved:
Thanks, I'll add the additional tests.

PR #9977, opened 2022-08-05
Swift: extract comments

AlexDenisov:


MathiasVP (approved):
LGTM! Do we want to run a DCA for this to measure the extractor performance impact before merging this?

geoffw0 (approved):
Library QL and test LGTM. :+1:

In C/C++, we have `CStyleComment` and `CppStyleComment` identifying the type of the comment (`//` or `/*`).  But it turns out this is just done by pattern matching on the comment string, so we can easily add this feature later.

AlexDenisov:
> LGTM! Do we want to run a DCA for this to measure the extractor performance impact before merging this?

@MathiasVP it's a great idea!

PR #9978, opened 2022-08-05
Swift: Fix flow summaries for methods

MathiasVP:
The `getCallTarget` predicate in our flow summary instantiation was suffering from the same problem as the one I described [here](https://github.com/github/codeql/pull/9673#discussion_r905171464): The call to `init(contentsOf:)` is really two calls (one providing the qualifier, and another one providing the `contentsOf` argument. Prior to this PR, the `getCallTarget` predicate was giving us the function that provided the qualifier.

cc @geoffw0.

Note: Once we merge https://github.com/github/codeql/pull/9891 this class of issues will be gone (since we can provide a much better API for function calls then)


rdmarsh2 (approved):

geoffw0:
Great, thanks!

PR #9979, opened 2022-08-05
Swift: Add InlineExpectationsTest framework

MathiasVP:
Now that we have [comment extraction](https://github.com/github/codeql/pull/9977) we can start to make use of the `InlineExpectationsTest` framework. This PR adds the framework and instantiates it for Swift, but it doesn't actually _use_ it for anything yet.

Once we've merged this PR we can start to make use of it for writing dataflow tests.


AlexDenisov (approved):
:shipit:

PR #9980, opened 2022-08-05
C#: Also disable shared compilation in the tracer for `dotnet msbuild`

hvitved:


criemen (approved):

criemen:
It'd be great if we had an (internal) integration test, just like we do have for dotnet build.

michaelnebel (approved):
LGTM!

PR #9982, opened 2022-08-05
Java: Add query for RSA without OAEP

joefarebrother:


java/ql/lib/semmle/code/java/security/RsaWithoutOaepQuery.qll
```diff
@@ -0,0 +1,15 @@
+/** Definitions for the RSE without OAEP query */
```
intrigus-lgtm, line 1:
```suggestion
/** Definitions for the RSA without OAEP query */
```

java/ql/src/Security/CWE/CWE-780/RsaWithoutOaep.qhelp
```diff
@@ -0,0 +1,27 @@
+<!DOCTYPE qhelp PUBLIC "-//Semmle//qhelp//EN" "qhelp.dtd">
+<qhelp>
+
+  <overview>
+    <p>Cryptographic algorithms often use padding schemes to make the plaintext less predictable. The OAEP scheme (Optimal Asymmetric Encryption Padding) should used with RSA encryption.
```
intrigus-lgtm, line 5:
```suggestion
    <p>Cryptographic algorithms often use padding schemes to make the plaintext less predictable. The OAEP scheme (Optimal Asymmetric Encryption Padding) should be used with RSA encryption.
```

java/ql/lib/semmle/code/java/security/RsaWithoutOaepQuery.qll
```diff
@@ -0,0 +1,15 @@
...
+
+import java
+import Encryption
+import semmle.code.java.dataflow.DataFlow
+
+/** Holds if `c` is a call which initialises an RSA cipher without using OAEP padding. */
+predicate rsaWithoutOaepCall(CryptoAlgoSpec c) {
+  exists(CompileTimeConstantExpr specExpr, string spec |
+    specExpr.getStringValue() = spec and
+    DataFlow::localExprFlow(specExpr, c.getAlgoSpec()) and
```
atorralba, line 11: Do you think it would make sense to use a full-blown `DataFlow::Configuration` here so that we catch some more cases (e.g. field flow or inter-procedural flow)?

atorralba:
Looks good to me (only needs rebasing to fix the failing test I think).

Asking for a docs review.

java/ql/src/Security/CWE/CWE-780/RsaWithoutOaep.ql
```diff
@@ -0,0 +1,20 @@
...
+ */
+
+import java
+import semmle.code.java.security.RsaWithoutOaepQuery
+import DataFlow::PathGraph
+
+from RsaWithoutOaepConfig conf, DataFlow::Node source, DataFlow::Node sink
+where conf.hasFlow(source, sink)
+select source, source, sink,
+  "This specification is used to initialize an RSA cipher without OAEP padding $@.", sink, "here"
```
atorralba, line 20: As per the failed test: either change the query back to a `problem` kind (even if you're using a `DataFlow::Configuration`), or use `PathNode` here:

```suggestion
from RsaWithoutOaepConfig conf, DataFlow::PathNode source, DataFlow::PathNode sink
where conf.hasFlowPath(source, sink)
select source.getNode(), source, sink,
  "This specification is used to initialize an RSA cipher without OAEP padding $@.", sink.getNode(),
  "here"
```

saritai:
👋 Docs content first responder here! Adding this to our review board so a writer can review it! 🙇‍♀️

java/ql/src/Security/CWE/CWE-780/RsaWithoutOaep.ql
```diff
@@ -0,0 +1,20 @@
+/**
+ * @name Use of RSA algorithm without OAEP
+ * @description Using RSA encryption without OAEP padding can lead to a padding oracle attack, weakening the encryption.
```
mchammer01, line 3: Perhaps, to make the second clause flow more naturally. What do you think?
```suggestion
 * @description Using RSA encryption without OAEP padding can result in a padding oracle attack, leading to a weaker encryption.
```

java/ql/src/Security/CWE/CWE-780/RsaWithoutOaep.qhelp
```diff
@@ -0,0 +1,27 @@
+<!DOCTYPE qhelp PUBLIC "-//Semmle//qhelp//EN" "qhelp.dtd">
+<qhelp>
+
+  <overview>
+    <p>Cryptographic algorithms often use padding schemes to make the plaintext less predictable. The OAEP scheme (Optimal Asymmetric Encryption Padding) should be used with RSA encryption.
+    Using no padding or an outdated padding scheme such as PKCS1 can weaken the encryption by making it vulnerable to a padding oracle attack.
```
mchammer01, line 6: Switching the options so that we don't start with the negative one (which makes the rest of the sentence difficult to parse)
```suggestion
    Using an outdated padding scheme such as PKCS1, or no padding at all, can weaken the encryption by making it vulnerable to a padding oracle attack.
```

java/ql/src/Security/CWE/CWE-780/RsaWithoutOaep.qhelp
```diff
@@ -0,0 +1,27 @@
+<!DOCTYPE qhelp PUBLIC "-//Semmle//qhelp//EN" "qhelp.dtd">
+<qhelp>
+
+  <overview>
+    <p>Cryptographic algorithms often use padding schemes to make the plaintext less predictable. The OAEP scheme (Optimal Asymmetric Encryption Padding) should be used with RSA encryption.
```
mchammer01, line 5: Perhaps?
```suggestion
    <p>Cryptographic algorithms often use padding schemes to make the plaintext less predictable. The OAEP (Optimal Asymmetric Encryption Padding) scheme should be used with RSA encryption.
```

mchammer01 (dismissed):
@joefarebrother - this LGTM ✨ 
I've added a few minor comments or suggestions for your consideration.

aschackmull (approved):

atorralba:
@sidshank I think we were waiting for performance evaluation in DCA, but since @aschackmull approved it, it means we're good in that sense and @joefarebrother went ahead with the merge 😄

aschackmull:
Luckily the query looks sufficiently innocent that I'm not too worried about its performance. But if you were waiting for dca then you could just have waited on the merge - I approved the PR, as the QL looks fine, so you could merge whenever you were ready to do so.

atorralba:
Well, I assumed that was the case because of the DCA runs linked to this PR, but @joefarebrother will know for sure.

joefarebrother:
I was trying to do a DCA run but it wasn't working (now that I think about it I probably just needed to rebase); merged because it was approved and that it seemed unlikely to have a performance issue.

PR #9988, opened 2022-08-08
C#: Update .NET Core and ASP.NET Core Stubs.

michaelnebel:
In this PR we update .NET Core and ASP.NET Core stubs and replace some hand written stubs with the generated ones.
The stubs for ASP.NET Core and a .NET Core was produced by making stubs for ServiceStack 6.2.0 and its dependencies.

There were a couple of issues with the code produced by the stub generator.
1. A couple of <error>'s in delegate pointer return types (the delegate and corresponding method has just been deleted from the stub).
2. A generation issue with an AdminUI type (confusion whether it was an enum or class). I just deleted one of them.
3. Again we saw the issue with enum's not preserving the value order of it's elements. In two cases I re-arranged the Order of the generated elements as this was needed for some test cases.
4. Some unwanted project references in the generated stubs that I have manually cleaned up (to reduce the number of needed stubs - but also to make the code compile to avoid duplicate implementations).

IMO this solution still beats that we continue to extend the hand written stubs for ASP.NET Core and use the generated ones with these minor changes. I think it is a good idea, if we make some improvements to the generator (at least the improvement of the enum value ordering), but in any case I think we need to consider the generator a "helper" tool and not a fully automatic solution for stub generation.


hvitved (approved):
This is really great, thanks for getting rid of the manual stubs.

PR #9989, opened 2022-08-08
C#: Handle `dotnet exec csc.dll` and the likes in the Lua tracer

hvitved:
This PR moves logic for handling `dotnet exec csc.dll` and friends from [the extractor](https://github.com/github/codeql/blob/main/csharp/extractor/Semmle.Extraction.CSharp.Driver/Driver.cs#L18) into the new Lua tracer spec. The benefit of doing so is that we can avoid unnecessary process creations, and logically doing compiler matching belongs to the tracer spec, and not the extractor.

Note that the logic has not yet been removed from the extractor; we will do so once the legacy tracer is retired.

Thanks to @criemen for helping out.

### DIFF
--- a/csharp/tools/tracing-config.lua
+++ b/csharp/tools/tracing-config.lua
@@ -1,6 +1,6 @@
 function RegisterExtractorPack(id)
     local extractor = GetPlatformToolsDirectory() ..
-                          'Semmle.Extraction.CSharp.Driver'
+        'Semmle.Extraction.CSharp.Driver'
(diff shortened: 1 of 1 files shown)


michaelnebel (dismissed):
Looks plausible to me!

criemen (approved):
Looks good to me as well, thanks!

hvitved:
Rebased onto latest `main` to hopefully get a successful DCA run.

michaelnebel (approved):
Looks plausible to me!

PR #9992, opened 2022-08-09
Fix path of `fetch-codeql`

tamasvajk:
This PR fixes the path of `.github/actions/fetch-codeql` where the CodeQL `actions/checkout@v3` is using a custom `path`.


aibaars (approved):
Looks good to me. I wonder why these workflows didn't fail when using the wrong path.

tamasvajk:
> Looks good to me. I wonder why these workflows didn't fail when using the wrong path.

These are jobs that run on a schedule or run on manual triggering. (I'm not sure if the manually triggered ones are still in use or not. They were not run for quite some time.)

PR #9995, opened 2022-08-09
Swift: CFG for `any!`

MathiasVP:
We were not constructing CFGs for `any!` expressions. This led to unreachable CFG nodes, which led to missing dataflow nodes (as observed by @geoffw0).


geoffw0 (approved):
Fantastic!  LGTM, and thanks for investigating this.

PR #9997, opened 2022-08-09
C++: Experimental product flow library

rdmarsh2:
This library performs a value-flow analysis of pairs of values from a pair of related sources to a pair of related sinks, using two underlying `DataFlow::Configuration`s and a post-analysis on their resulting `PathNode`s.


cpp/ql/test/experimental/query-tests/Security/CWE/CWE-119/OverrunWriteProductFlow.expected
```diff
@@ -0,0 +1,2 @@
+| test.cpp:19:19:19:24 | call to malloc | test.cpp:18:17:18:20 | size | test.cpp:26:18:26:23 | string | test.cpp:26:31:26:39 | (size_t)... |
+| test.cpp:19:19:19:24 | call to malloc | test.cpp:18:17:18:20 | size | test.cpp:30:18:30:23 | string | test.cpp:30:31:30:39 | (size_t)... |
```
MathiasVP, line 2: Could you annotate in the .cpp file whether these results are `// BAD` or `// BAD [FALSE POSITIVE]`?

MathiasVP (commented):
A couple of small things. Since this is really a prototype solution I don't think we need to give this too deep of a review, but I'll give @hvitved a chance to comment on it as well before we merge anything.

cpp/ql/test/experimental/query-tests/Security/CWE/CWE-119/OverrunWriteProductFlow.expected
rdmarsh2: At this point the query isn't mature enough to consider them bad or good - I'm just verifying that the library actually finds these flows.

MathiasVP: Fair enough 👍

cpp/ql/lib/experimental/semmle/code/cpp/dataflow/ProductFlow.qll
```diff
@@ -0,0 +1,115 @@
...
+    reachablePair1(conf, source1, source2, node1, node2) // TODO: restrict more
+    or
+    exists(DataFlow2::PathNode mid2 |
+      reachablePair2(conf, source1, source2, node1, mid2) and
+      mid2.getASuccessor() = node2 and
+      mid2.getNode().getFunction() = node2.getNode().getFunction()
+    )
+  }
+
+  predicate interprocStep(
```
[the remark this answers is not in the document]
rdmarsh2: This might be premature optimization, but `reachablePair1` will take only steps in the first and `reachablePair2` will take only steps in the second. That makes `reachablePair2` include all the locally-reachable pairs, but should reduce the number of duplicate tuples compared to putting both in the same predicate.

cpp/ql/lib/experimental/semmle/code/cpp/dataflow/ProductFlow.qll
```diff
@@ -0,0 +1,111 @@
...
+      DataFlow::PathNode source1, DataFlow2::PathNode source2, DataFlow::PathNode sink1,
+      DataFlow2::PathNode sink2
+    ) {
+      isSourcePair(source1.getNode(), source2.getNode()) and
+      isSinkPair(sink1.getNode(), sink2.getNode()) and
+      reachablePair2(this, source1, source2, sink1, sink2)
+    }
+  }
+
+  class Conf1 extends DataFlow::Configuration {
```
hvitved, line 33: private

cpp/ql/lib/experimental/semmle/code/cpp/dataflow/ProductFlow.qll
```diff
@@ -0,0 +1,111 @@
...
+    override predicate isSource(DataFlow::Node source) {
+      exists(Configuration conf | conf.isSourcePair(source, _))
+    }
+
+    override predicate isSink(DataFlow::Node sink) {
+      exists(Configuration conf | conf.isSinkPair(sink, _))
+    }
+  }
+
+  class Conf2 extends DataFlow2::Configuration {
```
hvitved, line 45: private

cpp/ql/lib/experimental/semmle/code/cpp/dataflow/ProductFlow.qll
```diff
@@ -0,0 +1,111 @@
...
+    Conf2() { this = "Conf2" }
+
+    override predicate isSource(DataFlow::Node source) {
+      exists(Configuration conf | conf.isSourcePair(_, source))
```
hvitved, line 49:
```suggestion
      exists(Configuration conf, DataFlow::Node source1 |
        conf.isSourcePair(source1, source) and
        any(Conf1 c).hasFlow(source1, _)
      )
```

cpp/ql/lib/experimental/semmle/code/cpp/dataflow/ProductFlow.qll
```diff
@@ -0,0 +1,111 @@
...
+    }
+
+    override predicate isSink(DataFlow::Node sink) {
+      exists(Configuration conf | conf.isSinkPair(_, sink))
```
hvitved, line 53:
```suggestion
      exists(Configuration conf, DataFlow::Node sink1 |
        conf.isSinkPair(sink1, sink) and
        any(Conf1 c).hasFlow(_, sink1)
      )
```

cpp/ql/lib/experimental/semmle/code/cpp/dataflow/ProductFlow.qll
```diff
@@ -0,0 +1,111 @@
...
+      mid2.getASuccessor() = node2 and
+      mid1.getNode().getEnclosingCallable() = funcMid and // TODO: recursive function weirdness?
+      mid2.getNode().getEnclosingCallable() = funcMid and
+      node1.getNode().getEnclosingCallable() = func and
+      node2.getNode().getEnclosingCallable() = func and
+      funcMid != func
+    )
+  }
+
+  predicate reachablePair(
```
hvitved, line 99: private

cpp/ql/lib/experimental/semmle/code/cpp/dataflow/ProductFlow.qll
```diff
@@ -0,0 +1,111 @@
...
+    reachablePair1(conf, source1, source2, node1, node2) // TODO: restrict more
+    or
+    exists(DataFlow2::PathNode mid2 |
+      reachablePair2(conf, source1, source2, node1, mid2) and
+      mid2.getASuccessor() = node2 and
+      mid2.getNode().getEnclosingCallable() = node2.getNode().getEnclosingCallable()
+    )
+  }
+
+  predicate interprocStep(
```
hvitved, line 83: private

cpp/ql/lib/experimental/semmle/code/cpp/dataflow/ProductFlow.qll
```diff
@@ -0,0 +1,111 @@
...
+    reachablePair(conf, source1, source2, node1, node2)
+    or
+    exists(DataFlow::PathNode mid1 |
+      reachablePair1(conf, source1, source2, mid1, node2) and
+      mid1.getASuccessor() = node1 and
+      mid1.getNode().getEnclosingCallable() = node1.getNode().getEnclosingCallable()
+    )
+  }
+
+  predicate reachablePair2(
```
hvitved, line 70: private

cpp/ql/lib/experimental/semmle/code/cpp/dataflow/ProductFlow.qll
```diff
@@ -0,0 +1,111 @@
...
+    override predicate isSource(DataFlow::Node source) {
+      exists(Configuration conf | conf.isSourcePair(_, source))
+    }
+
+    override predicate isSink(DataFlow::Node sink) {
+      exists(Configuration conf | conf.isSinkPair(_, sink))
+    }
+  }
+
+  predicate reachablePair1(
```
hvitved, line 57: private

cpp/ql/lib/experimental/semmle/code/cpp/dataflow/ProductFlow.qll
```diff
@@ -0,0 +1,111 @@
...
+    Configuration conf, DataFlow::PathNode source1, DataFlow2::PathNode source2,
+    DataFlow::PathNode node1, DataFlow2::PathNode node2
+  ) {
+    reachablePair(conf, source1, source2, node1, node2)
+    or
+    exists(DataFlow::PathNode mid1 |
+      reachablePair1(conf, source1, source2, mid1, node2) and
+      mid1.getASuccessor() = node1 and
+      mid1.getNode().getEnclosingCallable() = node1.getNode().getEnclosingCallable()
```
hvitved, line 66: This is always a candidate for a bad join (i.e., joining all nodes with same enclosing callable first).

cpp/ql/lib/experimental/semmle/code/cpp/dataflow/ProductFlow.qll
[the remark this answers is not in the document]
hvitved: You may also want to do some pre-big-step calculations before computing product reachability, much [like in the `ContentDataFlow` library](https://github.com/github/codeql/blob/main/csharp/ql/lib/semmle/code/csharp/dataflow/internal/ContentDataFlow.qll#L302).

cpp/ql/lib/experimental/semmle/code/cpp/dataflow/ProductFlow.qll
```diff
@@ -0,0 +1,111 @@
...
+     *
+     * `sink1` and `sink2` must belong to the same callable.
+     */
+    abstract predicate isSinkPair(DataFlow::Node sink1, DataFlow::Node sink2);
+
+    predicate hasFlowPath(
+      DataFlow::PathNode source1, DataFlow2::PathNode source2, DataFlow::PathNode sink1,
+      DataFlow2::PathNode sink2
+    ) {
+      isSourcePair(source1.getNode(), source2.getNode()) and
```
hvitved, line 27: This should not be needed, right?

cpp/ql/lib/experimental/semmle/code/cpp/dataflow/ProductFlow.qll
```diff
@@ -0,0 +1,111 @@
...
+    abstract predicate isSourcePair(DataFlow::Node source1, DataFlow::Node source2);
+
+    /**
+     * Holds if `(sink1, sink2)` is a relevant data flow sink.
+     *
+     * `sink1` and `sink2` must belong to the same callable.
+     */
+    abstract predicate isSinkPair(DataFlow::Node sink1, DataFlow::Node sink2);
+
+    predicate hasFlowPath(
```
hvitved, line 23: I am not sure whether we want to expose the underlying path nodes for the two helper configurations in the final API. Since we cannot visualize two paths simultaneously, my initial thinking was to expose only the path graph for the first component, but restricted to nodes that are actually relevant (i.e. reachable in both forwards and backwards product flow).

hvitved (commented):
Glad to see that the approach seems to work.

cpp/ql/lib/experimental/semmle/code/cpp/dataflow/ProductFlow.qll
rdmarsh2: Oh, right, that's now implied by `reachablePair2`

cpp/ql/lib/experimental/semmle/code/cpp/dataflow/ProductFlow.qll
```diff
@@ -0,0 +1,115 @@
...
+  predicate reachablePair2(
+    Configuration conf, DataFlow::PathNode source1, DataFlow2::PathNode source2,
+    DataFlow::PathNode node1, DataFlow2::PathNode node2
+  ) {
+    reachablePair1(conf, source1, source2, node1, node2) // TODO: restrict more
+    or
+    exists(DataFlow2::PathNode mid2 |
+      reachablePair2(conf, source1, source2, node1, mid2) and
+      mid2.getASuccessor() = node2 and
+      mid2.getNode().getFunction() = node2.getNode().getFunction()
```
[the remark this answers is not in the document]
rdmarsh2: Done

cpp/ql/lib/experimental/semmle/code/cpp/dataflow/ProductFlow.qll
```diff
@@ -0,0 +1,115 @@
...
+
+module ProductFlow {
+  abstract class Configuration extends string {
+    bindingset[this]
+    Configuration() { any() }
+
+    /**
+     * Holds if `(source1, source2)` is a relevant data flow source.
+     *
+     * `source1` and `source2` must belong to the same callable.
```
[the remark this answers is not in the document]
rdmarsh2: Be contained in the same function/constructor/lambda body

cpp/ql/src/experimental/Likely Bugs/OverrunWriteProductFlow.ql
```diff
@@ -0,0 +1,31 @@
...
+import semmle.code.cpp.valuenumbering.GlobalValueNumbering
+import semmle.code.cpp.models.interfaces.Allocation
+import semmle.code.cpp.models.interfaces.ArrayFunction
+
+class StringSizeConfiguration extends ProductFlow::Configuration {
+  StringSizeConfiguration() { this = "StringSizeConfiguration" }
+
+  override predicate isSourcePair(DataFlow::Node bufSource, DataFlow::Node sizeSource) {
+    exists(
+      GVN sizeGVN // TODO: use-use flow instead of GVN
```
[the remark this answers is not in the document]
rdmarsh2: Two issues: one is that GVN will produce multiple data flow sources for each allocation site, where we only really want one, and the source(s) that actually reaches the sink won't be directly related to the allocation site. The second is that in the overall flow path, we may have multiple layers where the allocation size is passed down through a function call, but we don't want to take these undirected GVN steps at every call site in the path.

cpp/ql/lib/experimental/semmle/code/cpp/dataflow/ProductFlow.qll
```diff
@@ -0,0 +1,119 @@
+import semmle.code.cpp.ir.dataflow.DataFlow
+import semmle.code.cpp.ir.dataflow.DataFlow2
+
+module ProductFlow {
+  abstract class Configuration extends string {
```
hvitved, line 5: May as well make this a parameterized module.

cpp/ql/lib/experimental/semmle/code/cpp/dataflow/ProductFlow.qll
```diff
@@ -0,0 +1,119 @@
...
+      mid1.getASuccessor() = node1 and
+      mid1.getNode().getEnclosingCallable() = node1.getNode().getEnclosingCallable()
+    )
+  }
+
+  private predicate reachablePair2(
+    Configuration conf, DataFlow::PathNode source1, DataFlow2::PathNode source2,
+    DataFlow::PathNode node1, DataFlow2::PathNode node2
+  ) {
+    reachablePair1(conf, source1, source2, node1, node2) // TODO: restrict more
```
aschackmull, line 82: I get what you're trying to do here - and if it works as a prototype that's great. But it looks quite problematic from a performance point of view.

Here's what I'd do instead:
Consider the set of steps from one of the two underlying flow graphs.
Join those with `.getEnclosingCallable()` for both ends and filter to only keep the inter-procedural cases. (Let's leave flow through recursive function calls as future work and ignore it for now).
Think about how well those pairs of callables can be expected to identify the step - I'd guess pretty well.
So the same for the other underlying flow.
Given our assumption about low callable-pair-to-PathNode-pair fanout in the above, we can join on the callable-pairs and produce the set of inter-procedural steps in the product flow graph.
With these we can avoid materialising relations like `reachablePair2`, which includes almost the cartesian product of co-callable-located reachable nodes, and instead only step to the relevant node-pairs.
I don't think there's a reason not to just use fastTC on the local steps, so I'd just do that.

cpp/ql/test/experimental/query-tests/Security/CWE/CWE-119/test.cpp
```diff
@@ -0,0 +1,37 @@
...
+
+void strncpy_wrapper(string_t *str, char *buf) {
+    strncpy(str->string, buf, str->size);
+}
+
+void test2(int size, char *buf) {
+    string_t *str = mk_string_t(size);
+    strncpy_wrapper(str, buf);
+}
+
```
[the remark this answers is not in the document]
rdmarsh2: Done now

cpp/ql/lib/experimental/semmle/code/cpp/dataflow/ProductFlow.qll
rdmarsh2: Can we do that yet? When I try the straightforward thing I get an error: `Classes may not extend across parameterised module boundaries`.

MathiasVP: Yep, that's still a restriction that's yet to be lifted.

rdmarsh2: Is it even going to be lifted? I think extending an abstract class within a parameterized module will have weird implications for the "active module" concept. I'm guessing this will need to wait until the dataflow API is a parameterized module.

cpp/ql/lib/experimental/semmle/code/cpp/dataflow/ProductFlow.qll
rdmarsh2: I've implemented this suggestion in [b1de54e](https://github.com/github/codeql/pull/9997/commits/b1de54e9aafa485579ac57fc3e4d41c16386219f)

cpp/ql/lib/experimental/semmle/code/cpp/dataflow/ProductFlow.qll
```diff
@@ -0,0 +1,119 @@
...
+    exists(Declaration predDecl, Declaration succDecl |
+      DataFlow::PathGraph::edges(pred1, succ1) and
+      DataFlow2::PathGraph::edges(pred2, succ2) and
+      predDecl != succDecl and
+      pred1.getNode().getEnclosingCallable() = predDecl and
+      pred2.getNode().getEnclosingCallable() = predDecl and
+      succ1.getNode().getEnclosingCallable() = succDecl and
+      succ2.getNode().getEnclosingCallable() = succDecl
+    )
+  }
```
aschackmull, line 107: Yes, this predicate is exactly what I had in mind. But as written, its best join-order isn't pipelineable. So it'll very likely need to be factored like this:
```suggestion
  pragma[nomagic]
  private predicate interprocEdge1(Declaration predDecl, Declaration succDecl, DataFlow::PathNode pred1, DataFlow::PathNode succ1) {
    DataFlow::PathGraph::edges(pred1, succ1) and
    predDecl != succDecl and
    pred1.getNode().getEnclosingCallable() = predDecl and
    succ1.getNode().getEnclosingCallable() = succDecl
  }

  pragma[nomagic]
  private predicate interprocEdge2(Declaration predDecl, Declaration succDecl, DataFlow2::PathNode pred2, DataFlow2::PathNode succ2) {
    DataFlow2::PathGraph::edges(pred2, succ2) and
    predDecl != succDecl and
    pred2.getNode().getEnclosingCallable() = predDecl and
    succ2.getNode().getEnclosingCallable() = succDecl
  }

  private predicate interprocEdgePair(
    DataFlow::PathNode pred1, DataFlow2::PathNode pred2, DataFlow::PathNode succ1,
    DataFlow2::PathNode succ2
  ) {
    exists(Declaration predDecl, Declaration succDecl |
      interprocEdge1(predDecl, succDecl, pred1, succ1) and
      interprocEdge2(predDecl, succDecl, pred2, succ2)
    )
  }
```

cpp/ql/lib/experimental/semmle/code/cpp/dataflow/ProductFlow.qll
```diff
@@ -0,0 +1,119 @@
...
+      reachableInterprocEntry(conf, source1, source2, midEntry1, midEntry2) and
+      interprocEdgePair(midExit1, midExit2, node1, node2) and
+      localPathStep1*(midEntry1, midExit1) and
+      localPathStep2*(midEntry2, midExit2)
+    )
+  }
+
+  private predicate localPathStep1(DataFlow::PathNode pred, DataFlow::PathNode succ) {
+    DataFlow::PathGraph::edges(pred, succ) and
+    pred.getNode().getEnclosingCallable() = succ.getNode().getEnclosingCallable()
```
aschackmull, line 86: In case of a bad join-order here, you may need to unbind:
```suggestion
    pragma[only_bind_out](pred.getNode().getEnclosingCallable()) = pragma[only_bind_out](succ.getNode().getEnclosingCallable())
```

cpp/ql/lib/experimental/semmle/code/cpp/dataflow/ProductFlow.qll
```diff
@@ -0,0 +1,119 @@
...
+        exists(Configuration conf, DataFlow::Node sink1 |
+          conf.isSinkPair(sink1, sink) and any(Conf1 c).hasFlow(_, sink1)
+        )
+      }
+    }
+  }
+
+
+  private predicate reachableInterprocEntry(
+    Configuration conf, DataFlow::PathNode source1, DataFlow2::PathNode source2,
```
aschackmull, line 66: Including the source pair in the reachability predicate like this is probably fine for a prototype, but if you find that it performs poorly (and it likely will in some cases), then it'll need to be reachability independent of the source pair in order to construct relevant local pair-to-pair edges, the reachable pairs need to be wrapped in an IPA construction, and the final end-to-end reachability done by fastTC.

cpp/ql/lib/experimental/semmle/code/cpp/dataflow/ProductFlow.qll
aschackmull: Great, this looks much better now. There are still some things that can be improved performance-wise, but it might be good enough for now.

cpp/ql/lib/experimental/semmle/code/cpp/dataflow/ProductFlow.qll
aschackmull: No, that achieves something slightly different - that forces `edges` to be first, but doesn't rule out the bad `edges; getNode; getEnclosing; getEnclosing; getNode` order.

MathiasVP: Ah, right. That would indeed be bad!

MathiasVP: Fixed in e0a5d18d7d9a0e0807a3a1191b94d73ae6af7bd3.

cpp/ql/lib/experimental/semmle/code/cpp/dataflow/ProductFlow.qll
MathiasVP: Fixed in e0a5d18d7d9a0e0807a3a1191b94d73ae6af7bd3.

cpp/ql/lib/experimental/semmle/code/cpp/dataflow/ProductFlow.qll
MathiasVP: Indeed, I'd definitely like to see this as well. Let's delay that to a future PR, though @rdmarsh2.

aschackmull:
How is performance now? Things to check:
* Inspect the join-order of `reachableInterprocEntry`.
* Inspect the join-order of `reachable`.
* Compare the total tuple count of `reachableInterprocEntry` with the tuple count of its projection `reachableInterprocEntry(c, _, _, n1, n2)`. This should tell you something about the room for improvement through fastTC.
* Compare the above tuple counts with the counts of `DataFlow::PathGraph::edges` and `DataFlow2::PathGraph::edges`. This should tell you something about your amortized expected complexity.

MathiasVP:
Thanks for those questions! I'll attempt to answer them when I get the new range analysis to terminate on a project that has some interesting product flows 😭.

MathiasVP:
@aschackmull for your entertainment while I try to fix other issues:
- `reachableInterprocEntry`: https://gist.github.com/MathiasVP/e13237772586a5d48dc787ce0fb7c61a
- `reachable` (it was inlined into `select`): https://gist.github.com/MathiasVP/f284878847e568e8b9139fc81d93c0a7

aschackmull:
Those join-orders look plausibly acceptable, but have definite room for improvement through some hand-tuned manual magic. As a small tweak, you might consider `nomagic`ing `interprocEdgePair` as that is being inlined, which wasn't exactly the idea. Further improvements are definitely possible if `reachableInterprocEntry` takes a non-trivial amount of time to compute.
The join-order for `reachable` looks fine (but also has some room for improvement).

PR #9998, opened 2022-08-09
Use StrcpyFunction in cpp/bad-strncpy-size

d10c:
This PR:

- Uses the [StrcpyFunction](https://github.com/github/codeql/blob/main/cpp/ql/lib/semmle/code/cpp/models/implementations/Strcpy.qll#L14) class in the [StrncpyFlippedArgs](https://github.com/github/codeql/blob/main/cpp/ql/src/Likely%20Bugs/Memory%20Management/StrncpyFlippedArgs.ql) query instead of an ad-hoc predicate for finding strcpy-like functions.
- Tests this by adding one previously unsupported strcpy-like function (`wcsxfrm_l`) to StrncpyFlippedArgs's test.cpp.


cpp/ql/test/query-tests/Likely Bugs/Memory Management/StrncpyFlippedArgs/test.cpp
```diff
@@ -93,3 +95,17 @@ void test8(char x[], char y[]) {
...
   strncpy(x, y, 32);
 }
+
+void test9()
+{
+	wchar_t buf1[10];
+	wchar_t buf2[20];
+	const wchar_t *str = L"01234567890123456789";
+
+	wcsxfrm_l(buf1, str, sizeof(buf1), nullptr);					// BAD (but not a StrncpyFlippedArgs bug)
```
geoffw0, line 105: > (but not a StrncpyFlippedArgs bug)

I'm not sure what you're saying here, the query appears to have flagged this line.

geoffw0 (commented):
This looks great.

I would include a short change note, as the query now finds more results.

cpp/ql/test/query-tests/Likely Bugs/Memory Management/StrncpyFlippedArgs/test.cpp
d10c: Yeah, good point... `test9()` is a copy of `test2()` pretty much, including the comments, and has the same behaviour. Namely, that line is flagged because (I'm guessing) it falls under the ["Fixed sized case"](https://github.com/github/codeql/blob/a144828266e87d4a4cdac02f4b546245453ce11e/cpp/ql/src/Likely%20Bugs/Memory%20Management/StrncpyFlippedArgs.ql#L88), which is not exactly a case of "flipped" args. At least, that's how I interpreted the original comment.

geoffw0 (dismissed):
Looks good to me!

geoffw0 (approved):

PR #9999, opened 2022-08-09
Go: implement conservative cross-thread dataflow

smowton:
Steps into captured variables are moved into jumpStep where they always should have been, and the store/load step implementation for channels is completed.

For the time being this takes a very conservative approach to identify channels that are likely connected: if there is exactly one receive site and one send site for a field, the two are presumed connected.

### DIFF
--- a/go/ql/lib/semmle/go/controlflow/ControlFlowGraph.qll
+++ b/go/ql/lib/semmle/go/controlflow/ControlFlowGraph.qll
@@ -169,10 +169,10 @@ module ControlFlow {
     /**
      * Holds if this node writes `rhs` to `channel`.
      */
-    predicate writesToChannel(DataFlow::Node channel, DataFlow::Node rhs) {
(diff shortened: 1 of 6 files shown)


go/ql/lib/semmle/go/controlflow/ControlFlowGraph.qll
```diff
@@ -166,11 +166,21 @@ module ControlFlow {
       )
     }
 
+    /**
+     * Holds if this node writes `rhs` to `channel`.
+     */
+    predicate writesToChannel(DataFlow::Node channel, DataFlow::Node rhs) {
+      exists(SendStmt send |
+        send.getChannel() = channel.(DataFlow::ExprNode).asExpr() and
+        send.getValue() = rhs.(DataFlow::ExprNode).asExpr()
```
igfoo, line 175: Can you give the parameters type `DataFlow::ExprNode` and avoid the need to cast these?

igfoo (dismissed):
Looks plausible to me.

igfoo (approved):

PR #10002, opened 2022-08-10
Ruby: Model protected methods

hmac:
Add two new predicates to `MethodBase`: `isPublic` and `isProtected`.

As part of this I've refactored the code around method visibility. This is easiest to review commit-by-commit.

### DIFF
--- a/ruby/ql/lib/codeql/ruby/ast/Method.qll
+++ b/ruby/ql/lib/codeql/ruby/ast/Method.qll
@@ -53,16 +53,68 @@ class MethodBase extends Callable, BodyStmt, Scope, TMethodBase {
    * This is either 'public', 'private' or 'protected'.
    */
   string getVisibility() {
-    result = this.getVisibilityModifier().getVisibility()
+    result = getVisibilityModifier(this).getVisibility()
     or
-    not exists(this.getVisibilityModifier()) and result = "public"
+    not exists(getVisibilityModifier(this)) and result = "public"
   }
+}
 
-  /**
-   * Gets the visibility modifier that defines the visibility of this method, if
-   * any.
-   */
-  VisibilityModifier getVisibilityModifier() { none() }
+/**
+ * Gets the visibility modifier that explicitly sets the visibility of method
+ * `m`.
+ *
+ * Examples:
+ * ```rb
+ * def f
+ * end
+ * private :f
+ *
+ * private def g
+ * end
+ * ```
+ */
+private VisibilityModifier getExplicitVisibilityModifier(Method m) {
+  result.getMethodArgument() = m
+  or
+  exists(ModuleBase n, string name |
+    methodIsDeclaredIn(m, n, name) and
+    modifiesIn(result, n, name)
+  )
+}
+
+/**
+ * Gets the visibility modifier that defines the visibility of method `m`, if
+ * any.
+ */
+private VisibilityModifier getVisibilityModifier(MethodBase mb) {
+  mb =
+    any(Method m |
+      result = getExplicitVisibilityModifier(m)
+      or
+      not exists(getExplicitVisibilityModifier(m)) and
+      exists(ModuleBase n, int methodPos | isDeclaredIn(m, n, methodPos) |
+        // The relevant visibility modifier is the closest call that occurs before
+        // the definition of `m` (typically this means higher up the file).
+        result =
+          max(int modifierPos, VisibilityModifier modifier |
+            modifier.modifiesAmbientVisibility() and
+            isDeclaredIn(modifier, n, modifierPos) and
+            modifierPos < methodPos
+          |
+            modifier order by modifierPos
+          )
+      )
+    )
+  or
+  mb =
+    any(SingletonMethod m |
+      result.getMethodArgument() = m
+      or
+      exists(ModuleBase n, string name |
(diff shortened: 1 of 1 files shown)


ruby/ql/lib/codeql/ruby/ast/Method.qll
```diff
@@ -47,45 +64,40 @@ class MethodBase extends Callable, BodyStmt, Scope, TMethodBase {
...
+  predicate modifiesAmbientVisibility() {
+    this.getMethodName() = ["public", "private", "protected"] and
+    this.getNumberOfArguments() = 0
   }
-}
 
-/** A call to `private_class_method`. */
-private class PrivateClassMethod extends MethodModifier {
-  PrivateClassMethod() { this.getMethodName() = "private_class_method" }
+  string getVisibility() {
```
hvitved, line 90: Could add a `private newtype Visibility = Public() or Private() or Protected()` and use that instead of `string`.

ruby/ql/lib/codeql/ruby/ast/Method.qll
```diff
@@ -140,28 +152,41 @@ class Method extends MethodBase, TMethod {
...
-  n = m.getEnclosingModule() and name = m.getName()
+  override VisibilityModifier getVisibilityModifier() {
+    result.getEnclosingModule() = this.getEnclosingModule() and
+    (
+      result.getMethodArgument() = this
+      or
+      result.getMethodArgument().getConstantValue().getStringlikeValue() = this.getName()
+    )
+    or
+    exists(Namespace n, int methodPos | n.getStmt(methodPos) = this |
```
hvitved, line 180: Should there be a check that this method does not have an explicit overriding modifier?

ruby/ql/lib/codeql/ruby/ast/Method.qll
```diff
@@ -140,28 +152,41 @@ class Method extends MethodBase, TMethod {
...
-}
 
-/**
- * Holds if the method `m` has name `name` and is declared in namespace `n`.
- */
-pragma[noinline]
-private predicate isDeclaredIn(MethodBase m, Namespace n, string name) {
-  n = m.getEnclosingModule() and name = m.getName()
+  override VisibilityModifier getVisibilityModifier() {
+    result.getEnclosingModule() = this.getEnclosingModule() and
```
hvitved, line 173: This is a candidate for a bad join.

hvitved: The original `modifiesMethod` was a way to avoid such a bad join.

ruby/ql/lib/codeql/ruby/ast/Method.qll
```diff
@@ -216,10 +241,14 @@ class SingletonMethod extends MethodBase, TSingletonMethod {
...
    * ```
    */
-  override predicate isPrivate() {
-    exists(Namespace n, string name |
-      any(PrivateClassMethod p).modifiesMethod(n, name) and
-      isDeclaredIn(this, n, name)
+  override predicate isPrivate() { super.isPrivate() }
+
+  override VisibilityModifier getVisibilityModifier() {
+    result.getEnclosingModule() = this.getEnclosingModule() and
```
hvitved, line 250: Candidate for a bad join.

ruby/ql/lib/codeql/ruby/ast/Method.qll
```diff
@@ -151,21 +159,26 @@ class Method extends MethodBase, TMethod {
...
    */
-  override predicate isPrivate() {
-    this.getVisibilityModifier().getVisibility() = "private"
-    or
-    // Top-level methods are private members of the Object class
-    this.getEnclosingModule() instanceof Toplevel
-  }
-
-  override predicate isPublic() { super.isPublic() and not this.isPrivate() }
+  override predicate isPrivate() { super.isPrivate() }
```
hvitved, line 152: Why is this override needed?

hvitved (commented):
Looks good!

ruby/ql/lib/codeql/ruby/ast/Method.qll
hvitved: Should be OK if we make `getVisibilityModifier` private as well (`VisibilityModifier` already is).

ruby/ql/lib/codeql/ruby/ast/Method.qll
hvitved: I have pushed two join-order fixes.

calumgrant:
Any updates on this @hmac?

ruby/ql/lib/codeql/ruby/ast/Method.qll
```diff
@@ -36,58 +36,68 @@ class MethodBase extends Callable, BodyStmt, Scope, TMethodBase {
...
+    result = this.getVisibilityModifier().getVisibility()
+    or
+    not exists(this.getVisibilityModifier()) and result = "public"
+  }
+
+  /**
+   * Gets the visibility modifier that defines the visibility of this method, if
+   * any.
+   */
+  VisibilityModifier getVisibilityModifier() { none() }
```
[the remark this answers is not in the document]
hvitved: I have pushed a change that does this. Let me know what you think.

hvitved (approved):

PR #10003, opened 2022-08-10
C#: Stub generator improvements.

michaelnebel:
In this PR we introduce the following improvements to the the stub generator.
- Enums are now explicitly typed and enum values are assigned.
- Support for function pointers types.

This brings the stub generator in state where we *almost* can generate stubs for the ASP.NET Core and .NET Core frameworks without making manual changes to the generated code.
The only part that is now *missing* are some `unsafe` keywords (whether this is due to extraction errors or issues with the stub generator is unknown).

Furthermore, the subs for ASP.NET Core and .NET Core has been regenerated based on the newest version of the stub generator.

Closes https://github.com/github/codeql-csharp-team/issues/260


hvitved (approved):

csharp/ql/src/Stubs/Stubs.qll
```diff
@@ -513,12 +513,30 @@ private string stubClassName(Type t) {
                                   stubClassName(element), "," order by i
                                 ) + ")"
                         else
-                          if t instanceof ValueOrRefType
+                          if t instanceof FunctionPointerType
```
hvitved, line 516: You can remove this if, and instead replace the matching `else` with `or not t instanceof FunctionPointerType and`

csharp/ql/src/Stubs/Stubs.qll
```diff
@@ -513,12 +513,30 @@ private string stubClassName(Type t) {
...
                           then
-                            result =
-                              stubQualifiedNamePrefix(t) + t.getUndecoratedName() +
-                                stubGenericArguments(t)
-                          else result = "<error>"
+                            exists(CallingConvention callconvention, string calltext |
```
hvitved, line 518: Perhaps add `FunctionPointerType fpt`, assign `fpt = this`, and replace the three `t.(FunctionPointerType)` with `fpt`.

csharp/ql/test/query-tests/Stubs/All/Test.cs
```diff
@@ -134,6 +134,34 @@ public class Nested : Class9
 
         public Class9.Nested NestedInstance { get; } = new Class9.Nested(1);
     }
+
+    public enum Enum1
```
hvitved, line 138: Add a test case with non-int enum?

csharp/ql/src/Stubs/Stubs.qll
michaelnebel: Sure - we can do it like that instead - I just continued the pattern that was already there.
However, it might make things slightly more complicated, when adding new cases 😄

csharp/ql/src/Stubs/Stubs.qll
michaelnebel: Yes, I thought about that as well, but decided against it because last time I did it, the review comment recommend to do it the other way around 😄  But on the other hand, here we already have an exists, so it will make it nicer 👍

csharp/ql/test/query-tests/Stubs/All/Test.cs
michaelnebel: Oh, definitely yes!
and also a testcase for function pointers!

csharp/ql/src/Stubs/Stubs.qll
hvitved: It makes sense to introduce a new variable (either with `exists` or `any`) when there are multiple identical casts, as in this case.

michaelnebel: Great - that is the rule of 👍 I needed 😸

csharp/ql/src/Stubs/Stubs.qll
```diff
@@ -726,7 +744,9 @@ pragma[noinline]
 private string stubEnumConstant(EnumConstant ec, Assembly assembly) {
   ec instanceof GeneratedMember and
   ec.getALocation() = assembly and
-  result = "    " + escapeIfKeyword(ec.getName()) + ",\n"
+  result =
+    "    " + escapeIfKeyword(ec.getName()) + " = " + "(" +
+      ec.getDeclaringEnum().getUnderlyingType().toStringWithTypes() + ")" + ec.getValue() + ",\n"
```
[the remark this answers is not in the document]
michaelnebel: You can have enums of different types (eg. long). In that case an explicit cast is required. I will also make a testcase.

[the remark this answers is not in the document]
michaelnebel: Wait, you are right!
We might not need the explicit cast (the compiler will figure it out, if we just type the enum correctly)!
Thank you - I will try that!

hvitved (approved):

PR #10006, opened 2022-08-10
Java: Remove SensitiveLoggingQuery results that flow through a source.

aschackmull:
This query was particularly prone to source-to-source flow due to its heuristic nature of its source definition. So adding an `inBarrier` will remove a lot of results that are effectively useless duplicates. In one observed case this removed several thousand results.


atorralba (approved):
Love it, let's merge if DCA is happy.

PR #10007, opened 2022-08-10
Dataflow: Fix identification of source PathNodes in the presence of source-to-source flow

aschackmull:
If one source could flow to another with a different tracked type, we would end up duplicating the reported results for the latter source.


hvitved (approved):
Good catch!

atorralba (approved):

PR #10008, opened 2022-08-10
Ruby: Add `rb/log-injection` query 

alexrford:
Detects some cases where a logging call may allow a malicious user to forge new log entries. The concrete implementation is mostly identical to the JS version of this query.


alexrford:
Having some issues starting a DCA run right now - I'll trigger one tomorrow.

PR #10012, opened 2022-08-10
QL: update codeql-action in QL-for-QL

erik-krogh:
The previous version was from back in March. 

The updated version of `codeql-action` is the latest `main` + a commit that adds `ql` as a supported language. 

I had to update how we do the config file to support the changes that has happened in `codeql-action`.

### DIFF
--- a/.github/workflows/ql-for-ql-build.yml
+++ b/.github/workflows/ql-for-ql-build.yml
@@ -163,6 +163,11 @@ jobs:
           languages: ql
           db-location: ${{ runner.temp }}/db
           config-file: ./ql-for-ql-config.yml
+      - name: Move pack cache
+        run: |
+          cp -r ${PACK}/.cache ql/ql/src/.cache
+        env:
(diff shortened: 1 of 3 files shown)


.github/workflows/ql-for-ql-build.yml
```diff
@@ -17,7 +17,7 @@ jobs:
       - uses: actions/checkout@v3
       - name: Find codeql
         id: find-codeql
-        uses: github/codeql-action/init@aa93aea877e5fb8841bcb1193f672abf6e9f2980
+        uses: github/codeql-action/init@71a8b35ff4c80fcfcd05bc1cd932fe3c08f943ca
```
esbena, line 20: Could we point this at a tag to make similar updates simpler.

I.e. create a tag for 71a8b35ff4c80fcfcd05bc1cd932fe3c08f943ca on gitub/codeql-action titled `v2-with-ql-language`, and use that instead of the explicit sha.

esbena (approved):

PR #10014, opened 2022-08-10
Swift: Add an inline expectations test for taint flow

geoffw0:
Add an inline expectations test for taint flow ([discussed here](https://github.com/github/codeql/pull/10000#discussion_r942264663)).  This makes it a little easier to see what's going on in the source files, and ensures things stay up-to-date.


swift/ql/test/library-tests/dataflow/taint/TaintInline.ql
```diff
@@ -0,0 +1,20 @@
...
+
+  override string getARelevantTag() { result = "taintedFromLine" }
+
+  override predicate hasActualResult(Location location, string element, string tag, string value) {
+    exists(TestConfiguration config, Node source, Node sink, Expr sinkExpr |
+      config.hasFlow(source, sink) and
+      sinkExpr = sink.asExpr() and
+      location = sinkExpr.getLocation() and
+      element = sinkExpr.toString() and
+      tag = "taintedFromLine" and
```
MathiasVP, line 16: `taintedFromLine` is quite a mouthful. Could we just call it `tainted` instead? If you think it's too unclear what `tainted=42` is supposed to mean we could write a comment at the top of the file saying that `tainted=42` means that taint is coming from line 42.

MathiasVP (commented):
Thanks a lot for the quick fix, @geoffw0! I only have a slightly subjective comment.

swift/ql/test/library-tests/dataflow/taint/TaintInline.ql
geoffw0: I lean towards being slightly more verbose, but I'm happy with just `tainted` if you are.  Done.

MathiasVP (approved):

PR #10019, opened 2022-08-11
delete the dead TypeRepr files

erik-krogh:
It looks like https://github.com/github/codeql/pull/9805 forgot to delete some files, and those files don't compile 
(QL-for-QL didn't like that). 

So I deleted them.


jketema (approved):
Thanks.

PR #10020, opened 2022-08-11
C++: Improve QLDoc based on earlier review

jketema:
See here: https://github.com/github/codeql/pull/9971#discussion_r943280458


MathiasVP (approved):
Thanks!

PR #10021, opened 2022-08-11
C++: Add internal metrics query for IR consistency

jketema:
Turns out there were more of these than I expected. So, for DCA we might still sum these up to a single number.

This is internal, so I do not think a change note is required.


MathiasVP (approved):
LGTM!

PR #10026, opened 2022-08-11
Go: Partial URLs should not sanitize against SSRF

pwntester:
As an example:

```go
	urlPath := ctx.Req.URL.Path
	hash := urlPath[strings.LastIndex(urlPath, "/")+1:]
        req, _ := http.NewRequest("GET", source+hash, nil)
```

### DIFF
--- a/go/ql/lib/change-notes/2023-04-14-partial-URLs-should-not-sanitize-against-SSRF.md
+++ b/go/ql/lib/change-notes/2023-04-14-partial-URLs-should-not-sanitize-against-SSRF.md
@@ -1,0 +1,4 @@
+---
+category: minorAnalysis
+---
+* Taking a slice is now considered a sanitizer for `SafeUrlFlow`.


go/ql/lib/semmle/go/security/SafeUrlFlowCustomizations.qll
```diff
@@ -35,4 +35,10 @@ module SafeUrlFlow {
   private class UnsafeUrlMethodEdge extends SanitizerEdge {
     UnsafeUrlMethodEdge() { this = any(UnsafeUrlMethod um).getAnInputNode(_) }
   }
+  
+  /** Any slicing of the URL, considered as a sanitizer for safe URL flow. */
+  private class StringSlicingEdge extends SanitizerEdge {
+    StringSlicingEdge() { this = any(DataFlow::SliceNode sn) }
+  }
+
```
owen-mc, line 43:
```suggestion
```

go/ql/lib/semmle/go/security/SafeUrlFlowCustomizations.qll
```diff
@@ -35,4 +35,9 @@ module SafeUrlFlow {
   private class UnsafeUrlMethodEdge extends SanitizerEdge {
     UnsafeUrlMethodEdge() { this = any(UnsafeUrlMethod um).getAnInputNode(_) }
   }
+
```
owen-mc, line 38:
```suggestion

```

owen-mc, line 38:
```suggestion

```

owen-mc:
@pwntester Sorry this PR didn't get more attention. I've fixed up the formatting problem that was stopping CI from running properly (though it did take 3 attempts). Do you think this should be merged? Could you add a test, possibly based on the code you put in the PR description? Also, I find it confusing that the title says "should not sanitize" but the change adds a sanitizer.

atorralba:
@owen-mc As I understand it, this is adding a sanitizer to `SafeUrlFlow`, which is used to discard sinks considered "safe" in the SSRF query. Safe sinks are those that have flow coming from a part of a request that is not user-controllable (e.g. the host or full URL, since those are determined by the destination server that hosts the application, and thus can't be arbitrarily set by an attacker). In other words, adding a sanitizer to `SafeUrlFlow` is _removing_ a "sanitizer" from the SSRF query.

In this case, @pwntester wants to mark flows coming from a _part_ of the request URL as unsafe. So, for an incoming request like:

```
https://application.server.safe.com/path?can=be&user=controlled#this-too
```

The scheme, authority, and possibly path of the URL can't be arbitrarily set, so they are sources of `SafeUrlFlow`. On the contrary, the query and fragment parts of the URL _are_ user-controllable, so slice accesses on the full URL path can be considered sanitizers of `SafeUrlFlow`.

@pwntester or anyone, please correct me if I'm wrong, since this is my first time looking at the Go SSRF query and libraries 😄.

pwntester:
Thanks @atorralba (could not have explained it better with my own words 😄)

PR #10031, opened 2022-08-11
C++: Handle block assignments

jketema:
Block assignments may occur as part of default copy/move constructors and assignment operators that are generated by the frontend. The assignments operate like `memcpy`, where the type of the source is used determine the size of what needs to be copied.


cpp/ql/lib/semmle/code/cpp/ir/implementation/raw/internal/TranslatedExpr.qll
```diff
@@ -1493,6 +1491,75 @@ class TranslatedAssignExpr extends TranslatedNonConstantExpr {
...
+  override BlockAssignExpr expr;
+
+  final override TranslatedElement getChild(int id) {
+    id = 0 and result = this.getLeftOperand()
+    or
+    id = 1 and result = this.getRightOperand()
+  }
+
+  final override Instruction getFirstInstruction() {
+    // The operand evaluation order should since block assignments behave like memcpy.
```
MathiasVP, line 1504: This comment reads a bit weird. Did you mean "[...] should be left-to-right [...]"?

cpp/ql/lib/semmle/code/cpp/ir/implementation/raw/internal/TranslatedExpr.qll
```diff
@@ -1450,8 +1450,6 @@ class TranslatedAssignExpr extends TranslatedNonConstantExpr {
       result = this.getLeftOperand().getResult()
   }
 
-  abstract Instruction getStoredValue();
```
MathiasVP, line 1453: Wat. This was just an unused abstract member predicate lying around? Well spotted!

MathiasVP (approved):
LGTM! I only have a small comment nitpick. Feel free to fix that up in a later PR to avoid an unnecessary submodule dance.

cpp/ql/lib/semmle/code/cpp/ir/implementation/raw/internal/TranslatedExpr.qll
jketema: Woops:
```suggestion
    // The operand evaluation order should not matter since block assignments behave like memcpy.
```

jketema: Does that make more sense?

MathiasVP: That makes a lot of sense, yes!

PR #10034, opened 2022-08-12
Kotlin: Change handling of version variants in build script

tamasvajk:


java/kotlin-extractor/kotlin_plugin_versions.py
```diff
@@ -21,6 +21,7 @@ def version_string_to_tuple(version):
     m = re.match(r'([0-9]+)\.([0-9]+)\.([0-9]+)(.*)', version)
     return tuple([int(m.group(i)) for i in range(1, 4)] + [m.group(4)])
 
+# Version numbers in the list need to be in semantically increasing order
 many_versions = [ '1.4.32', '1.5.0', '1.5.10', '1.5.21', '1.5.31', '1.6.10', '1.6.20', '1.7.0' ]
```
igfoo, line 25: The order is getting a little overloaded, as `get_latest_url` assumes that the last one is the one that CI should use.

Perhaps defining a `single_version` that CI uses would be simpler than trying to sort these.

tamasvajk:
The linked internal PR contains a run of the "Check Kotlin versions" workflow, which shows that this change has no functional impact.

java/kotlin-extractor/kotlin_plugin_versions.py
igfoo: Yes, perhaps with something like
```
if ci_version not in many_versions:
    raise Exception('CI version must be one of many_versions')
```

igfoo (approved):

PR #10036, opened 2022-08-12
JS: More precise handling of "exports"

asgerf:
https://github.com/github/codeql/pull/9548 caused problems for generated typings, since it does not distinguish between the main entry point of a package, versus a file that is exported under a non-trivial relative path.

For example, with these exports
```
{
  "exports": {
    ".": "./main.js",
    "./foo": "./foo.js"
  }
}
```
the typings for `foo.d.ts` would be incorrectly merged with those from `main.d.ts`.

### DIFF
--- a/javascript/ql/lib/semmle/javascript/NPM.qll
+++ b/javascript/ql/lib/semmle/javascript/NPM.qll
@@ -56,7 +56,7 @@ class PackageJson extends JsonObject {
    * This can be given by the `main` or `module` property, or via the
    * `exports` property with the relative path `"."`.
    */
-  string getMain() { result = getExportedPath(".") }
+  string getMain() { result = this.getExportedPath(".") }
 
   /**
    * Gets the path to the file exported with the given relative path.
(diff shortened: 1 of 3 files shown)


javascript/ql/lib/semmle/javascript/NPM.qll
```diff
@@ -50,8 +50,23 @@ class PackageJson extends JsonObject {
...
 
-  /** Gets the main module of this package. */
-  string getMain() { result = MainModulePath::of(this).getValue() }
+  /**
+   * Gets the main module of this package.
+   *
+   * This can be given by the `main` or `module` property, or via the
+   * `exports` property with the relative path `"."`.
+   */
+  string getMain() { result = MainModulePath::of(this, ".").getValue() }
```
erik-krogh, line 59:
```suggestion
  string getMain() { result =getExportedPath(".") }
```
?

erik-krogh (approved):
👍  
With an optional comment.

erik-krogh (approved):

PR #6716, opened 2021-09-18
Java: Additional hardcoded credentials candidates 3rd-party api calls

bananabr:
Included the methods that might receive hard-coded credential material in their parameters from some of the most used SSH, FTP, and MongoDB Java libraries in the Maven official repository to the SensitiveApi.qll library.

PR #10041, opened 2022-08-13
Java: support more libraries in hardcoded-credentials queries

smowton:
This is abandoned external PR https://github.com/github/codeql/pull/6716, but I have also written tests and ported the hardcoded-creds queries to use inline-expectations tests so it's practically possible to verify that the results are as expected.


atorralba (approved):
LGTM! Thanks for doing this.

I always wonder, though — how much of this should be covered by secret scanning and not CodeQL?

java/ql/lib/semmle/code/java/security/HardcodedCredentialsComparison.qll
```diff
@@ -0,0 +1,25 @@
+/**
+ * Provides classes and predicates to detect comparing a parameter to a hard-coded credential.
+ */
+
```
atorralba, line 4: Oh, and a little NIT: maybe we should have `import java` here for consistency. I guess the imports should be `private` too, but this doesn't seem like it's going to be used anywhere except in the query itself, so 🤷

smowton: Added the imports. Re: who should cover this, my impression was the involvement of dataflow analysis makes it appropriate for CodeQL, but it's possible I'm underestimating what secret scanning can achieve.

atorralba: Well, that makes sense actually! 👍

atorralba (approved):

PR #10045, opened 2022-08-15
Swift: CWE-757: insecure TLS configuration

AlexDenisov:


swift/ql/test/query-tests/Security/CWE-757/TLS13IsNotUsed.swift
```diff
@@ -0,0 +1,173 @@
...
+}
+
+/// Indirect assignment (global vars)
+
+let badGlobalVersion = tls_protocol_version_t.TLSv10
+let goodGlobalVersion = tls_protocol_version_t.TLSv12
+
+func case_10() {
+  let config = URLSessionConfiguration()
+  config.tlsMinimumSupportedProtocolVersion = badGlobalVersion // BAD [not detected]
```
MathiasVP, line 91: Indeed, we don't yet have global dataflow for Swift. So this makes sense.

swift/ql/test/query-tests/Security/CWE-757/TLS13IsNotUsed.swift
```diff
@@ -0,0 +1,173 @@
...
+/// Indirect assignment (via external entity)
+
+struct BadDefault {
+  let TLSVersion: tls_protocol_version_t = tls_protocol_version_t.TLSv11
+}
+
+func case_16() {
+  let def = BadDefault()
+  let config = URLSessionConfiguration()
+  config.tlsMinimumSupportedProtocolVersion = def.TLSVersion // BAD [not detected]
```
MathiasVP, line 144: Yeah, I think this makes sense as well. I don't think we recognize the assignment from line 138 as an SSA definition.

Thanks for some awesome tests!

swift/ql/src/queries/Security/CWE-757/TLS13IsNotUsed.ql
```diff
@@ -0,0 +1,58 @@
...
+
+/**
+ * TODO
+ */
+class TLS13IsNotUsedConfig extends TaintTracking::Configuration {
+  TLS13IsNotUsedConfig() { this = "TLS13IsNotUsedConfig" }
+
+  override predicate isSink(DataFlow::Node node) {
+    exists(AssignExpr assign, MemberRefExpr member, string memberName |
+      assign.getSource() = node.asExpr() and
```
MathiasVP, line 27: Currently, this will only be seen as a sink if the write to the configuration is through an `AssignExpr`. If we want to generalize this we could let the sink be `sink.(DataFlow::PostUpdateNode).getPreUpdateNode().asExpr() = member.getBase()` (or whatever we call `a` in `a.b`). This would mean we recognized the write to the config no matter where it came from (i.e., in case it doesn't come from an `AssignExpr`).

However, we'd also need to be careful about the fields involved. We might need to also register `sink.(DataFlow::PostUpdateNode).getPreUpdateNode()` as implicitly reading the fields enumerated by `memberName` in `isSink` 🤔.

TLDR: In any case, this is good enough for now! But we can probably make this a bit less syntax-dependent in a future PR.

swift/ql/src/queries/Security/CWE-757/TLS13IsNotUsed.ql
```diff
@@ -0,0 +1,58 @@
...
+        memberName = "tlsMinimumSupportedProtocol" or
+        memberName = "tlsMaximumSupportedProtocolVersion" or
+        memberName = "tlsMaximumSupportedProtocol"
+      )
+    )
+  }
+
+  override predicate isSource(DataFlow::Node node) {
+    exists(MethodRefExpr expr, EnumElementDecl enum, string enumName |
+      node.asExpr() = expr and
```
MathiasVP, line 41: Hmmm, wait. Is `tls_protocol_version_t.TLSv12` represented as a `MethodRefExpr` now @redsun82? Is that ... intentional?

swift/ql/src/queries/Security/CWE-757/TLS13IsNotUsed.ql
```diff
@@ -0,0 +1,58 @@
...
+        enumName = "tlsProtocol11"
+      )
+    )
+  }
+}
+
+from TLS13IsNotUsedConfig config, DataFlow::PathNode sourceNode, DataFlow::PathNode sinkNode
+where config.hasFlowPath(sourceNode, sinkNode)
+select sinkNode.getNode(), sourceNode, sinkNode,
+  "TLS 1.3 should be used (from " + sourceNode.getNode().asExpr().getEnclosingFunction().getName() +
```
MathiasVP, line 57: I don't know if the name of the enclosing function from the source is helpful in the alert message. Afterall, the query already visualizes a path that starts at the source for the user to inspect. What do you think @AlexDenisov?

MathiasVP (commented):
Looks great so far! A couple of comments, but nothing that requires any immediate attention.

swift/ql/src/queries/Security/CWE-757/TLS13IsNotUsed.ql
AlexDenisov: Yeah, I totally agree! This is mostly for debugging and should be updated (together with the rest of the user facing text).

swift/ql/src/queries/Security/CWE-757/TLS13IsNotUsed.ql
AlexDenisov: This one got me confused as well, I've been looking for `DeclRefExpr`s and was surprised to not see it there.
I'm glad we have PrintAst though 😅

swift/ql/src/queries/Security/CWE-757/TLS13IsNotUsed.ql
AlexDenisov: OOC, what else could it be if not assignment? 🤔

geoffw0: I'm still not totally familiar with the Swift AST, but in CPP initializers are not assignment for example.

MathiasVP: maybe a write coming out a function like:
```swift
func set(_ protocol : inout SSLProtocol) {
  procotol = SSLProtocol.tlsProtocol10
}

func test() {
  let config = URLSessionConfiguration()
  set(&config.tlsMaximumSupportedProtocol); // BAD
}
```

(Assuming that compiles, obviously 😂)

swift/ql/src/queries/Security/CWE-757/TLS13IsNotUsed.ql
```diff
@@ -0,0 +1,58 @@
...
+  override predicate isSource(DataFlow::Node node) {
+    exists(MethodRefExpr expr, EnumElementDecl enum, string enumName |
+      node.asExpr() = expr and
+      expr.getMember() = enum and
+      enumName = enum.getName() and
+      (
+        enumName = "TLSv10" or
+        enumName = "TLSv11" or
+        enumName = "tlsProtocol10" or
+        enumName = "tlsProtocol11"
```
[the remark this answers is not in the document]
geoffw0: This message could be more helpful.  It means `["TLSv10", "TLSv11", "tlsProtocol10", "tlsProtocol11"]`.

swift/ql/src/queries/Security/CWE-757/TLS13IsNotUsed.ql
```diff
@@ -0,0 +1,58 @@
+/**
+ * @name TODO
+ * @description TODO
+ * @kind path-problem
+ * @problem.severity error
+ * @security-severity TODO
+ * @precision high
+ * @id swift/tls13-is-not-used
```
geoffw0, line 8: Consider a slightly more generic `@id`, in case requirements are increased at some point.

geoffw0 (commented):
Looks good to me, mostly just needs a bit of docs / polish.

swift/ql/src/queries/Security/CWE-757/InsecureTLS.ql
```diff
@@ -0,0 +1,55 @@
+/**
+ * @name Insecure TLS configuration
+ * @description TLS v1.0 and v1.1 versions are known to be vulnerable.. TLS v1.2 or v1.3 should be used instead.
+ * @kind path-problem
+ * @problem.severity error
+ * @security-severity TODO
```
geoffw0, line 6: You can generate this value using the script in the [CWE-Scores](https://github.com/github/cwe-scores) repo.  There's instructions in the repo readme, but let me know if you have problems with that at all.

geoffw0 (commented):
LGTM apart from a few tiny corrections.  I'm happy with query and the thorough tests.

After reviewing my comments I suggest you request a docs review (using the `ready-for-doc-review` tag).

martin389:
Docs first responder here 👋 I've now added this PR to our queue for another writer to pickup for docs review  👍

swift/ql/src/queries/Security/CWE-757/InsecureTLS.qhelp
```diff
@@ -0,0 +1,28 @@
...
+<p>TLS v1.0 and v1.1 versions are known to be vulnerable.</p>
+</overview>
+<recommendation>
+
+<p>Use <code>tls_protocol_version_t.TLSv12</code> or <code>tls_protocol_version_t.TLSv13</code> when configuring <code>URLSession</code>.</p>
+
+</recommendation>
+<example>
+
+<p>Specify newer <code>tls_protocol_version_t</code> explicitly, or omit it completely as the OS will use secure defaults.</p>
```
hubwriter, line 15:
```suggestion
<p>Specify a newer <code>tls_protocol_version_t</code> explicitly, or omit it completely as the OS will use secure defaults.</p>
```

hubwriter (approved):
LGTM. Just the one tiny comment.

MathiasVP (approved):
LGTM!

PR #10050, opened 2022-08-15
C#: ASP.NET MapGet Routing endpoints (Remote Flow Sources)

michaelnebel:
Parameters of delegates passed to routing endpoint calls like `MapGet` in ASP.NET Core are now considered remote flow sources.

### DIFF
--- a/csharp/ql/test/library-tests/dataflow/flowsources/aspremote/AspRemoteFlowSource.cs
+++ b/csharp/ql/test/library-tests/dataflow/flowsources/aspremote/AspRemoteFlowSource.cs
@@ -24,7 +24,6 @@ namespace Testing
 
     public class AspRoutingEndpoints
     {
-
         public void M1(string[] args)
         {
             var builder = WebApplication.CreateBuilder(args);
@@ -32,6 +31,7 @@ namespace Testing
 
             // The delegate parameters are considered flow sources.
             app.MapGet("/api/redirect/{newUrl}", (string newUrl) => { });
+            app.MapGet("/{myApi}/redirect/{myUrl}", (string myApi, string myUrl) => { } );
 
             app.Run();
         }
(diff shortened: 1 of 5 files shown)


csharp/ql/test/library-tests/dataflow/flowsources/aspremote/AspRemoteFlowSource.cs
```diff
@@ -20,4 +22,33 @@ public object MyAction(ViewModel viewModel)
...
+            app.MapGet("/api/redirect/{newUrl}", (string newUrl) => { });
+            app.MapGet("/{myApi}/redirect/{myUrl}", (string myApi, string myUrl) => { });
+
+            Action<string> handler = (string lambdaParam) => { };
+            app.MapGet("/api/redirect/{lambdaParam}", handler);
+
+            MapGetHandler handler2 = HandlerMethod;
+            app.MapGet("/api/redirect/{mapGetParam}", handler2);
+
+            app.MapPost("/api/redirect/{mapPostParam}", (string mapPostParam) => { });
```
tamasvajk, line 47: I think `MapPost` (and maybe others too) works with complex models too. Can you add a test for that? Also, properties of complex models are marked as remote flow sources in https://github.com/github/codeql/pull/9393. Would that work here too?

hvitved (dismissed):
LGTM. Let's see what DCA says.

csharp/ql/test/library-tests/dataflow/flowsources/aspremote/AspRemoteFlowSource.cs
michaelnebel: To make sure I understand; By complex models do you mean user defined types with fields and properties? Are you asking whether something like in the example below will be caught (I haven't tried but I don't think so)?

```csharp
public class Item
{
    public string Id { get; set; }
}

Minimal API:
...
app.MapPost("/items", (Item item) => Sink(item.Id))
app.Run();
```

tamasvajk: Yep, exactly. I think this might be used frequently, because even the MS samples show these at https://docs.microsoft.com/en-us/aspnet/core/tutorials/min-web-api?view=aspnetcore-6.0&tabs=visual-studio#add-the-api-code.

hvitved (approved):

michaelnebel:
DCA Looks good (1) No performance regression(s) (2) Only one alert diff, which looks like the usual wobly alert.
Merging now.

PR #10053, opened 2022-08-15
QL: add ql/consistent-alert-message

erik-krogh:
And drive-by remove some alerts on deprecated predicates/classes.  

[I got a WIP branch were I've fixed most of the alerts introduced by the new query](https://github.com/github/codeql/pull/9816).

### DIFF
--- a/ql/ql/src/queries/style/ConsistentAlertMessage.ql
+++ b/ql/ql/src/queries/style/ConsistentAlertMessage.ql
@@ -31,13 +31,10 @@ string getMessage(Select sel) {
  * This fingerprint avoid false positives where two queries with the same ID behave differently (which is OK).
  */
 string getSelectFingerPrint(Select sel) {
-  exists(File file, QLDoc doc |
-    sel.getLocation().getFile() = file and
-    any(TopLevel top | top.getLocation().getFile() = file).getQLDoc() = doc
-  |
+  exists(QueryDoc doc | doc = sel.getQueryDoc() |
     result =
-      doc.getContents().regexpCapture("(?s).*@id (\\w+)/([\\w\\-]+)\\s.*", 2) // query ID (without lang)
-        + "-" + doc.getContents().regexpCapture("(?s).*@kind (\\w+)\\s.*", 1) // @kind
+      doc.getQueryId() // query ID (without lang)
+        + "-" + doc.getQueryKind() // @kind
         + "-" +
         strictcount(String e | e.getParent*() = sel.getExpr(any(int i | i % 2 = 1))) // the number of string constants in the select
         + "-" + count(sel.getExpr(_)) // and the total number of expressions in the select
@@ -49,10 +46,10 @@ string getSelectFingerPrint(Select sel) {
  * The query-id (without language), the language, the message from the select, and a language agnostic fingerprint.
  */
 Select parseSelect(string id, string lang, string msg, string fingerPrint) {
-  exists(File file, QLDoc doc | result.getLocation().getFile() = file |
-    any(TopLevel top | top.getLocation().getFile() = file).getQLDoc() = doc and
-    id = doc.getContents().regexpCapture("(?s).*@id (\\w+)/([\\w\\-]+)\\s.*", 2) and
-    lang = doc.getContents().regexpCapture("(?s).*@id (\\w+)/([\\w\\-]+)\\s.*", 1) and
+  exists(QueryDoc doc |
+    doc = result.getQueryDoc() and
+    id = doc.getQueryId() and
+    lang = doc.getQueryLanguage() and
     fingerPrint = getSelectFingerPrint(result) and
     msg = getMessage(result).toLowerCase() // case normalize, because some languages upper-case methods.
   ) and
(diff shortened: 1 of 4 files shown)


ql/ql/src/queries/style/ConsistentAlertMessage.ql
```diff
@@ -0,0 +1,77 @@
...
+import ql
+
+/**
+ * Gets a string representation of the entire message in `sel`.
+ * Ignores everything that is not a string constant.
+ */
+string getMessage(Select sel) {
+  result =
+    strictconcat(String e, Location l |
+      // is child of an expression in the select (in an uneven position, that's where the message is)
```
tausbn, line 20: >  ... in an uneven position ...

That's an _odd_ way to phrase that. :wink: 

(Though I'm not sure "odd position" is actually more intuitive. Maybe "odd-indexed position" would be better?)

ql/ql/src/queries/style/ConsistentAlertMessage.ql
```diff
@@ -0,0 +1,77 @@
...
+/**
+ * Gets a language agnostic fingerprint for a Select.
+ * The fingerPrint includes e.g. the query-id, the @kind of the query, and the number of expressions in the select.
+ *
+ * This fingerprint avoid false positives where two queries with the same ID behave differently (which is OK).
+ */
+string getSelectFingerPrint(Select sel) {
+  exists(File file, QLDoc doc |
+    sel.getLocation().getFile() = file and
+    any(TopLevel top | top.getLocation().getFile() = file).getQLDoc() = doc
```
tausbn, line 36: I feel like this could be convenient to push into a `getQLDoc` predicate on `File`.

ql/ql/src/queries/style/ConsistentAlertMessage.ql
```diff
@@ -0,0 +1,77 @@
...
+        + "-" + count(sel.getExpr(_)) // and the total number of expressions in the select
+  )
+}
+
+/**
+ * Gets information about the select.
+ * The query-id (without language), the language, the message from the select, and a language agnostic fingerprint.
+ */
+Select parseSelect(string id, string lang, string msg, string fingerPrint) {
+  exists(File file, QLDoc doc | result.getLocation().getFile() = file |
```
tausbn, line 52: Suggestion: why not create a subclass of `QLDoc` for the kind of QLDoc that applies to a whole file, and then expose the relevant tags as member predicates?
(I think it's fine to only implement the predicates you need for this query. We can always add more later, and some of them may be a bit fiddly to implement.)

ql/ql/src/queries/style/ConsistentAlertMessage.ql
```diff
@@ -0,0 +1,77 @@
...
+  not result.getLocation().getFile().getRelativePath().matches("%/experimental/%") and
+  not lang = "ql" // excluding QL-for-QL
+}
+
+from Select sel, string id, string lang, string msg, string fingerPrint, string badLangs
+where
+  // for a select with a fingerprint
+  sel = parseSelect(id, lang, msg, fingerPrint) and
+  // there exists other languages with the same fingerprint, but other message
+  badLangs =
```
tausbn, line 69: The way the alert message is formulated, it sounds like `lang` is bad compared to everything in `badLangs`, but then wouldn't it make more sense to call this `goodLangs`?
Or perhaps just `otherLangs` to be more neutral.

(In the case where `lang` is the only outlier, it seems strange to classify the other languages as `bad`.)

ql/ql/src/queries/style/ConsistentAlertMessage.ql
```diff
@@ -0,0 +1,77 @@
...
+    |
+      e.getValue(), " | " order by l.getStartLine(), l.getStartColumn()
+    ).trim()
+}
+
+/**
+ * Gets a language agnostic fingerprint for a Select.
+ * The fingerPrint includes e.g. the query-id, the @kind of the query, and the number of expressions in the select.
+ *
+ * This fingerprint avoid false positives where two queries with the same ID behave differently (which is OK).
```
tausbn, line 31:
```suggestion
 * This fingerprint avoids false positives where two queries with the same ID behave differently (which is OK).
```

tausbn (changes requested):
A few comments and suggestions, but otherwise this looks good. :+1:

ql/ql/src/queries/style/ConsistentAlertMessage.ql
erik-krogh: > Though I'm not sure "odd position" is actually more intuitive.

That's not an odd position to have 😉  

I'll go for `odd-indexed position`.

ql/ql/src/queries/style/ConsistentAlertMessage.ql
erik-krogh: I've made a class `QueryDoc`, and it felt more appropriate to have an accessor for that on `Select`.

tausbn (approved):
Very nice! :+1:

PR #10056, opened 2022-08-16
Ruby: Recognise Rails render calls as HTTP responses

hmac:
What it says on the tin. We now recognise calls to `render`, `render_to_string` and `render_to_body` such as

```rb
def show
  render json: { some: "data" }
  render @user
  render html: "<p>hello</p>"
end
```

as instances of `HTTP::Server::HttpResponse`.

@nickrolfe hopefully this will be useful for you!


nickrolfe (commented):
Looks great, thank you! Just one nitpick on the change note.

nickrolfe (approved):

PR #10059, opened 2022-08-16
Ruby: Fix deprecation warning

hvitved:


erik-krogh (approved):
I'll take a look if I can re-deprecate it after this PR is merged.

PR #10061, opened 2022-08-16
Swift: Queries for CWE-311 (originally CWE-200)

geoffw0:
The idea is to have two queries, one for "Cleartext storage of sensitive information in a local database" and one for "Cleartext transmission of sensitive information".  Both will use the same heuristic framework for determining what is sensitive information, but each will have different taint sinks.  Right now there are no queries, just test cases covering the following:
 - storage using the Core Data database library
 - storage using the Realm database library
 - transmission as part of a `URL`
 - transmission using `NWConnection.send`

At this stage I would appreciate a review of these tests, with a particular focus on whether I've modelled Core Data and Realm in a reasonably realistic way (albeit with a _lot_ of setup + result handling code missing as its long winded and should be irrelevant to the queries).  I'm also open to other suggestions, including other sinks that might be worth considering.


geoffw0:
I've just pushed a load of progress.
 - `SensitiveExprs.qll` library, which detects likely sensitive / private data heuristically (based on the proven CPP implementation; though we'll probably want to refine things a bit more for the Swift eco-system when we have more data).
 - written the queries.
 - `.qhelp` for the queries.

There are some refinements I still want to make (to deal with some potential false positives; and spot certain types of sensitive data more reliably), but I'd rather do these things as follow-up.  This PR is already quite big and I believe complete enough to be reviewed and merged now.

swift/ql/src/queries/Security/CWE-311/CleartextStorageDatabase.ql
```diff
@@ -0,0 +1,109 @@
...
+    isSource(node)
+  }
+
+  override predicate isAdditionalTaintStep(DataFlow::Node node1, DataFlow::Node node2) {
+    // TODO: the following special case flows are required to catch any of the Realm test
+    // cases, I hope we'll be able to remove them once we have field flow???
+    // flow out from field accesses, i.e. `a.b` -> `a`
+    exists(MemberRefExpr m |
+      node1.asExpr() = m and // `a.b`
+      node2.asExpr() = m.getBaseExpr() // `a`
```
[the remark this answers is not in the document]
geoffw0: This QL-for-QL query has no useful information in the message or query description, and has no `.qhelp` either.  It is saying nothing.  Dismissed.

geoffw0:
The merge with main has caused some non-trivial problems in the taint flow.  I'll have to fix them next week.

I've requested a docs review for the two `.qhelp` files.

mchammer01:
@geoffw0 👋🏻 - just wondering if this is ready for editorial review. I am confused by the presence of the `ready-for-doc-review` label in this draft PR. Thanks for clarifying 🙏🏻 🙂

geoffw0:
> just wondering if this is ready for editorial review.

The docs are finished and ready for review.  Something needs fixing with the query itself running on latest `main`, which I need to look into, but the fix won't affect docs.

mchammer01:
Great, thanks for clarifying @geoffw0 ⚡

mchammer01 (dismissed):
@geoffw0 - this LGTM ✨ 
I've made minor comments on punctuation. You can ignore any you don't agree with, but it would be great to add the full stops to the list items in the references.
Also, perhaps I misunderstood but I thought these two queries related to CWE 311 🤔 
The previews failed so I wasn't able to preview the qhelp files. Also, do we need to add something to the changelog?

swift/ql/src/queries/Security/CWE-311/CleartextTransmission.ql
```diff
@@ -0,0 +1,86 @@
...
+ * @name Cleartext transmission of sensitive information
+ * @description Transmitting sensitive information across a network in
+ *              cleartext can expose it to an attacker.
+ * @kind path-problem
+ * @problem.severity warning
+ * @security-severity 7.5
+ * @precision high
+ * @id swift/cleartext-transmission
+ * @tags security
+ *       external/cwe/cwe-319
```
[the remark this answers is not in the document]
geoffw0: CWE-319 is a child of CWE-311.  I believe this is a more accurate label (and we generally only apply the most precise CWE label), whereas CWE-311 is an appropriate classification for both queries.  In the PR title, path and other places I'm talking about both so I use CWE-311.

swift/ql/src/queries/Security/CWE-311/CleartextStorageDatabase.ql
```diff
@@ -0,0 +1,109 @@
...
+ * @name Cleartext storage of sensitive information in a local database
+ * @description Storing sensitive information in a non-encrypted
+ *              database can expose it to an attacker.
+ * @kind path-problem
+ * @problem.severity warning
+ * @security-severity 7.5
+ * @precision medium
+ * @id swift/cleartext-storage-database
+ * @tags security
+ *       external/cwe/cwe-312
```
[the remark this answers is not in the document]
geoffw0: Again, CWE-312 is a child of CWE-311.  The relevant part of the hierarchy looks something like this:
```
CWE-693: Protection Mechanism Failure
 |-- ...
 \-- CWE-311: Missing Encryption of Sensitive Data
      |-- CWE-312: Cleartext Storage of Sensitive Information
      \-- CWE-319: Cleartext Transmission of Sensitive Information
```

swift/ql/src/queries/Security/CWE-311/CleartextStorageDatabase.qhelp
```diff
@@ -0,0 +1,28 @@
+<!DOCTYPE qhelp PUBLIC "-//Semmle//qhelp//EN" "qhelp.dtd">
+<qhelp>
+<overview>
+
+<p>Sensitive information that is stored unencrypted in a database is accessible to an attacker who gains access to that database. For example the information could be accessed by any process or user in a rooted device, or exposed through another vulnerability.</p>
+
+</overview>
+<recommendation>
+
+<p>Either encrypt the entire database, or ensure that each piece of sensitive information is encrypted before being stored. In general, decrypt sensitive information only at the point where it is necessary for it to be used in cleartext. Avoid storing sensitive information at all if you do not need to keep it.</p>
```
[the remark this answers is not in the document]
geoffw0: I think just "cleartext", it just means "written in plain text" i.e. "not encrypted".  I would consider "clear text" but that isn't what the CWE docs use.

geoffw0:
Thanks for the review @mchammer01 !

swift/ql/src/queries/Security/CWE-311/CleartextStorageDatabase.ql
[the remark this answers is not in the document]
mchammer01: Thanks for clarifying 🙇🏻‍♀️

swift/ql/src/queries/Security/CWE-311/CleartextTransmission.ql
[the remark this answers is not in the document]
mchammer01: I see, thanks for explaining!

geoffw0:
> The previews failed so I wasn't able to preview the qhelp files.

Where is this?

> Also, do we need to add something to the changelog?

Swift isn't released to the public yet, so we're not writing change notes until then.

geoffw0:
The merge issues are fixed, the query is now working again.  And ready to review.

Some of the results are less than perfect, that will be improved by one or more follow-up PRs at some stage.  Right now I just want to get the essentials in.

swift/ql/src/queries/Security/CWE-311/CleartextStorageDatabase.ql
```diff
@@ -0,0 +1,109 @@
...
+  }
+
+  override predicate isAdditionalTaintStep(DataFlow::Node node1, DataFlow::Node node2) {
+    // TODO: the following special case flows are required to catch any of the Realm test
+    // cases, I hope we'll be able to remove them once we have field flow???
+    // flow out from field accesses, i.e. `a.b` -> `a`
+    exists(MemberRefExpr m |
+      node1.asExpr() = m and // `a.b`
+      node2.asExpr() = m.getImmediateBase() // `a`
+    )
```
MathiasVP, line 94: This does object/field conflation for all nodes in the database which seems very aggressive. I can see why you want to do this given examples such as:
```swift
  let a = MyRealmSwiftObject()
  a.data = myPassword
  realm.add(a) // BAD
```
Field flow won't help us out-of-the-box since there's no corresponding read of `a.data`. However, you can add implicit read steps at the sinks by overriding [allowImplicitRead](https://github.com/github/codeql/blob/main/swift/ql/lib/codeql/swift/dataflow/internal/DataFlowImpl.qll#L125) in your configuration. Be sure to limit this predicate as much as possible though. Maybe based on things like:
- The node should only be the sink nodes
- The `ContentSet`s should only be fields from objects of type RealmSwiftObject

Also, note that none of these things will work before we merge https://github.com/github/codeql/pull/10157.

swift/ql/src/queries/Security/CWE-311/CleartextStorageDatabase.ql
```diff
@@ -0,0 +1,109 @@
...
+    or
+    // flow through assignment (!)
+    exists(AssignExpr ae |
+      node1.asExpr() = ae.getSource() and
+      node2.asExpr() = ae.getDest()
+    )
```
MathiasVP, line 100: This one surprises me. What additional flow does this give us? 😮

MathiasVP (dismissed):
A couple of questions. LGTM otherwise!

swift/ql/src/queries/Security/CWE-311/CleartextStorageDatabase.ql
geoffw0: I'll check.  For context, these cases were developed before merging several weeks worth of changes in `main`, so they may reflect problems we no longer have...

geoffw0: It _does_ still matter, its required to find the results in `testRealm.swift` lines 35 and 43, both of which look something like this:
```
	a.data = myPassword // source: myPassword
	realm.add(a) // sink: a
```
I was as surprised as you are when I encountered this!

MathiasVP: Ah, yes. That makes sense. This should also be solved by field flow + https://github.com/github/codeql/pull/10061#discussion_r958756974

swift/ql/src/queries/Security/CWE-311/CleartextStorageDatabase.ql
geoffw0: I've merged in `main` again and done what you suggest.  This seems like much less of a hack now.

There's currently a hardcoded class name from the test because we either don't have `ClassDecl.getParentClass()` or I don't know what its called.  That will need fixing.

swift/ql/src/queries/Security/CWE-311/CleartextStorageDatabase.ql
geoffw0: Ah you're right, with the other stuff in place this special case is no longer needed either.  Removed.  Thanks!

swift/ql/src/queries/Security/CWE-311/CleartextStorageDatabase.ql
geoffw0: I've fixed the hardcoded class name.

geoffw0:
All comments addressed (thanks for the review and help).

swift/ql/src/queries/Security/CWE-311/CleartextStorageDatabase.ql
```diff
@@ -0,0 +1,105 @@
...
+    isSink(node) and
+    exists(ClassDecl cd |
+      c.getAReadContent().(DataFlow::Content::FieldContent).getField() = cd.getAMember() and
+      cd.getType().(NominalType).getABaseType*().getName() = "RealmSwiftObject"
+    )
+    or
+    // any default implicit reads
+    super.allowImplicitRead(node, c)
+  }
+}
```
MathiasVP, line 98: Fantastic! I'm very happy to see that we could get rid of all the additional taint steps 😍. I jus opened https://github.com/github/codeql/pull/10254 which will hopefully make it possible to get a base type decl without going the types themselves. But we can always do that change as a follow-up to avoid an unnecessary merge from main.

MathiasVP (approved):
LGTM!

PR #10064, opened 2022-08-16
Swift: Fix SSA for calls to getter functions

MathiasVP:
This fixes some FNs we were seeing on https://github.com/github/codeql/pull/10045.

cc {@AlexDenisov, @redsun82}.


AlexDenisov (approved):
:shipit:

PR #10065, opened 2022-08-16
Kotlin: Add support for version 1.7.20-Beta

tamasvajk:
This PR adds support for 1.7.20-Beta. After adding that version, 1.7.10 support was broken. So the internal PR changes a bit how the interceptor selects compatible plugin versions. As a result, some version numbers were also changed in this PR too.


java/kotlin-extractor/build.gradle
```diff
@@ -30,9 +30,22 @@ sourceSets {
             // change the excludes for building with other versions.
             // Currently 1.7.0 is configured:
             excludes = [
-                    "utils/versions/v_1_4_32/*.kt",
-                    "utils/versions/v_1_5_21/Descriptors.kt",
-                    "utils/versions/v_1_6_10/Descriptors.kt",
+                    // For 1.7.20-Beta, the below two files should be included, and the corresponding v_1_7_20-Beta ones should be excluded.
```
igfoo, line 33: I don't understand this comment

java/kotlin-extractor/src/main/kotlin/utils/versions/v_1_7_20-Beta/allOverridden.kt
```diff
@@ -0,0 +1,6 @@
+package com.github.codeql.utils.versions
+
+import org.jetbrains.kotlin.ir.declarations.IrSimpleFunction
+import org.jetbrains.kotlin.ir.util.allOverridden
+
+fun IrSimpleFunction.allOverridden(includeSelf: Boolean = false) = this.allOverridden(includeSelf)
```
igfoo, line 6: Does anything stop this recursively calling itself? I think it would be clearer if we changed the name of our function.

tamasvajk: It doesn't call itself recursively, but I only know this from the IDE, and don't know the exact symbol resolution rules.

I changed the name to be more explicit.

igfoo (approved):

PR #10066, opened 2022-08-16
C#: Include both source code and hand-written summaries in data flow

hvitved:
As discussed with @aschackmull offline, it makes sense to include source code also when there is a manually written flow summary, since there might be a source or a sink in the source code that was not modeled.

### DIFF
--- a/csharp/ql/lib/semmle/code/csharp/dataflow/internal/DataFlowPrivate.qll
+++ b/csharp/ql/lib/semmle/code/csharp/dataflow/internal/DataFlowPrivate.qll
@@ -65,9 +65,11 @@ abstract class NodeImpl extends Node {
 
 private class ExprNodeImpl extends ExprNode, NodeImpl {
   override DataFlowCallable getEnclosingCallableImpl() {
-    result.getUnderlyingCallable() = this.getExpr().(CIL::Expr).getEnclosingCallable()
-    or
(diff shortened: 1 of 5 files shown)


csharp/ql/lib/semmle/code/csharp/dataflow/internal/DataFlowPrivate.qll
```diff
@@ -1330,9 +1330,10 @@ private module ReturnNodes {
  * In this case we adjust it to instead be a return node.
  */
 private predicate summaryPostUpdateNodeIsOutOrRef(SummaryNode n, Parameter p) {
-  exists(ParameterNode pn |
+  exists(ParameterNodeImpl pn, DataFlowCallable c, ParameterPosition pos |
     FlowSummaryImpl::Private::summaryPostUpdateNode(n, pn) and
-    pn.getParameter() = p and
+    pn.isParameterOf(c, pos) and
+    p = c.getUnderlyingCallable().getParameter(pos.getPosition()) and
```
michaelnebel, line 1336: Would the more narrow predicate c.asSummarizedCallable()... be sufficient?

hvitved: I would think so, yes. Good catch.

michaelnebel (approved):
LGTM!
It is a bit surprising that we don't see any real performance regressions on .NET runtime.

PR #10067, opened 2022-08-16
QL: support signature parameters in QL-for-QL

erik-krogh:
See the test-case for an example.  

The syntax is not yet supported in the latest version of CodeQL, but that will happen soon. 
(See the backref).  

I don't quite like how the grammar is done, but I don't see a better way. 
I can't in the grammar always determine if something is a reference to a module, or to some other type.  
One solution would just be to combine it all into one big `TypeExpr`.  
But in some places we know that the reference cannot be a module, so I want to keep those grammar rules separate.   

In the QL side it all becomes a `TypeRef`, so it all ends up working, even though some references to modules might be parsed as a `TypeExpr` instead of a `ModuleExpr`.  

A review of https://github.com/tree-sitter/tree-sitter-ql/pull/18 would also be nice 🥺

### DIFF
--- a/.github/workflows/ql-for-ql-tests.yml
+++ b/.github/workflows/ql-for-ql-tests.yml
@@ -44,7 +44,7 @@ jobs:
           CODEQL: ${{ steps.find-codeql.outputs.codeql-path }}
       - name: Check QL formatting
         run: |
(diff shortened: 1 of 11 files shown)


kaeluka (approved):
@erik-krogh walked me through it in a pairing session. It looks good to me!
Fun that QL-for-QL is ahead :)

PR #10068, opened 2022-08-16
Java: Make synthesized method bodies disjoint from source code.

aschackmull:
Distinguish MaD synthesized method bodies from any source code callables that happen to otherwise overlap.
This is essentially a bugfix, since we would otherwise get some unintended interaction between source code flow and synthesized flow occurring in the same callable.


aschackmull:
Dca looks fine. A few alert differences, but that is to be expected.

hvitved (approved):
LGTM

PR #10069, opened 2022-08-16
C++: Expose `PresentIRFunction` and override in `cpp/count-ir-inconsistencies`

jketema:
The `toString` implementtion that `PresentIRFunction` uses may result in very long strings that may crash the evaluator. Overriding allows is to limit the string size and still suffices when just counting the number of inconsistencies.

### DIFF
--- a/cpp/ql/lib/semmle/code/cpp/ir/implementation/aliased_ssa/IRConsistency.qll
+++ b/cpp/ql/lib/semmle/code/cpp/ir/implementation/aliased_ssa/IRConsistency.qll
@@ -22,7 +22,7 @@ module InstructionConsistency {
     abstract Language::Location getLocation();
   }
 
-  private class PresentIRFunction extends OptionalIRFunction, TPresentIRFunction {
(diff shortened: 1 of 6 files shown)


jketema:
I don't think this code will be hit in a DCA experiment, unless it's in combination with my DCA PR.

MathiasVP:
> I don't think this code will be hit in a DCA experiment, unless it's in combination with my DCA PR.

Yeah, that's what I was thinking. I'm totally okay with just merging this without it, though, as this is really only internal debugging code.

jketema:
> Yeah, that's what I was thinking. I'm totally okay with just merging this without it, though, as this is really only internal debugging code.

I think. I'll just merge if all the tests pass. Then, bump the submodule, and try running a new DCA experiment with the IRConsistency summary enabled. If that works, we can merge that PR tomorrow, so we don't have a nightly run where only half has the updated code from this PR.

jketema:
Note that was was able to reproduce the error I saw locally, and I confirmed that it went away locally with this change.

PR #10070, opened 2022-08-16
Swift: add more `SelfApplyExpr` testing

redsun82:


MathiasVP (approved):

swift/ql/test/extractor-tests/generated/expr/ConstructorRefCallExpr/ConstructorRefCallExpr_getArgument.expected
```diff
@@ -0,0 +1,3 @@
+| constructor_ref_calls.swift:7:9:7:9 | call to init | 0 | constructor_ref_calls.swift:7:9:7:9 | : X.Type |
+| constructor_ref_calls.swift:8:10:8:10 | call to init | 0 | constructor_ref_calls.swift:8:10:8:10 | : Y.Type |
```
geoffw0, line 2: Doesn't line 8 have an argument 1 as well?

geoffw0 (commented):
More tests, always good!

swift/ql/test/extractor-tests/generated/expr/ConstructorRefCallExpr/ConstructorRefCallExpr_getArgument.expected
redsun82: that's the usual confusing currying... with a twist for constructors. The init constructor is first applied to a `TypeExpr` (this call is a `ConstructorRefCallExpr`), and then applied to the constructor arguments (in a `CallExpr`).

So `Y(0)` has the AST
```
CallExpr(
  function=ConstructorRefCallExpr(
    function=DeclRefExpr(init(_:)),
    arguments=[TypeExpr(Y.Type)]
  ),
  arguments=[IntegerLiteralExpr(0)]
)
```
 The twist is that for normal methods, the first application is to the object instance, but for constructor I guess the instance does not exist. The metatype `Y.Type` is passed though, I guess to handle inheritance tricks (like using the `Self` type in the constructor).

geoffw0: Got it, thanks for explaining.

:+1:

PR #10071, opened 2022-08-16
Kotlin: Refactor TypeResults

igfoo:
We statically know when we expect to have no signature, so now we tell the type system what we know, rather than having signature always be nullable.


tamasvajk (approved):
Looks reasonable to me.

PR #10073, opened 2022-08-16
rename all occurrences of XML to Xml

erik-krogh:
[Acronyms should use normal PascalCase/camelCase](https://github.com/github/codeql/blob/main/docs/ql-style-guide.md#naming).  

written automatically using my patching utility 

The CI here is failing, and that is intended. Look at the CI checks in the internal PR (see the first backref below).  

I've also deleted a dead method in [java/kotlin-extractor/src/main/kotlin/utils/List.kt](https://github.com/github/codeql/pull/10073/commits/98acb1db01fb0574a1ba7ef3a4f520e16d66a616#diff-b2dd6f8c69711edb07aa0cbbc5686fe64c73958188af34e90807439c60423d75), because that method being dead caused the internal CI to fail. 

This PR has to be merged at the same time as the internal PR.

### DIFF
--- a/cpp/ql/lib/semmle/code/cpp/XML.qll
+++ b/cpp/ql/lib/semmle/code/cpp/XML.qll
@@ -36,12 +36,12 @@ class XmlLocatable extends @xmllocatable, TXmlLocatable {
 deprecated class XMLLocatable = XmlLocatable;
 
 /**
- * An `XMLParent` is either an `XMLElement` or an `XMLFile`,
+ * An `XmlParent` is either an `XmlElement` or an `XmlFile`,
  * both of which can contain other elements.
  */
 class XmlParent extends @xmlparent {
   XmlParent() {
-    // explicitly restrict `this` to be either an `XMLElement` or an `XMLFile`;
+    // explicitly restrict `this` to be either an `XmlElement` or an `XmlFile`;
     // the type `@xmlparent` currently also includes non-XML files
     this instanceof @xmlelement or xmlEncoding(this, _)
   }
--- a/cpp/ql/lib/semmle/code/cpp/XML.qll
+++ b/cpp/ql/lib/semmle/code/cpp/XML.qll
@@ -132,7 +132,7 @@ class XmlFile extends XmlParent, File {
   XmlElement getARootElement() { result = this.getAChild() }
 
   /** Gets a DTD associated with this XML file. */
-  XMLDTD getADTD() { xmlDTDs(result, _, _, _, this) }
+  XmlDTD getADTD() { xmlDTDs(result, _, _, _, this) }
 }
 
 /** DEPRECATED: Alias for XmlFile */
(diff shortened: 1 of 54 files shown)


smowton:
AFAIK generally we would keep the old name for public symbols and deprecate it to avoid a breaking change? Given that, maybe it's best to restrict the rewrite to private symbols?

erik-krogh:
> AFAIK generally we would keep the old name for public symbols and deprecate it to avoid a breaking change for 3rd-party queries? Given that, maybe it's best to restrict the rewrite to private symbols?

I've left a deprecated alias with the old name for every rename, isn't that good enough?  
People will get a deprecation warning for at least a year before we can delete the deprecated name.

smowton:
Oh sorry I somehow overlooked that, ignore me.

michaelnebel (approved):
C# 👍

tausbn (approved):
~LGTM~ Lgtm! :+1:

cpp/ql/lib/semmle/code/cpp/XML.qll
```diff
@@ -32,12 +32,15 @@ class XMLLocatable extends @xmllocatable, TXmlLocatable {
...
+
 /**
  * An `XMLParent` is either an `XMLElement` or an `XMLFile`,
  * both of which can contain other elements.
  */
-class XMLParent extends @xmlparent {
-  XMLParent() {
+class XmlParent extends @xmlparent {
+  XmlParent() {
     // explicitly restrict `this` to be either an `XMLElement` or an `XMLFile`;
```
jketema, line 44:
```suggestion
    // explicitly restrict `this` to be either an `XmlElement` or an `XmlFile`;
```

jketema (changes requested):

cpp/ql/lib/semmle/code/cpp/XML.qll
erik-krogh: The `XMLDTD` had also been missed by my patch (probably because of the all upper-case name), so I also updated that one.

jketema (dismissed):
C++ 👍

erik-krogh:
QL-for-QL is complaining that the `DTD` in `XmlDTD` is not PascalCase.  
I will fix that, but in a later PR.

javascript/ql/lib/semmle/javascript/XML.qll
```diff
@@ -165,6 +174,9 @@ class XMLDTD extends XMLLocatable, @xmldtd {
   }
 }
 
+/** DEPRECATED: Alias for XmlDTD */
+deprecated class XMLDTD = XmlDTD;
```
kaeluka, line 178: while we're at it, can we name this one `XmlDtd`?

kaeluka (approved):
Javascript: 👍, with an optional comment

javascript/ql/lib/semmle/javascript/XML.qll
erik-krogh: I going to rename more things in a later PR.
I'll include DTD in that one.

igfoo (approved):

atorralba (approved):
Java 👍

michaelnebel (commented):
C# Lgtm

PR #10075, opened 2022-08-16
delete old deprecations

erik-krogh:
The first commit was made using the same script as last time.   
The following commits were mostly made to fix CI errors and QL-for-QL warnings. 

All the deleted deprecations are at least 14 months old according to `git blame`.


hvitved (dismissed):
C# 👍 Thanks

javascript/ql/lib/semmle/javascript/dataflow/Configuration.qll
```diff
@@ -523,22 +523,6 @@ abstract class LabeledBarrierGuardNode extends BarrierGuardNode {
...
- * `SharedFlowStep::` module, such as `SharedFlowStep::step`.
- *
- * A data flow edge that should be added to all data flow configurations in
- * addition to standard data flow edges.
- *
- * Note: For performance reasons, all subclasses of this class should be part
- * of the standard library. Override `Configuration::isAdditionalFlowStep`
- * for analysis-specific flow steps.
- */
-deprecated class AdditionalFlowStep = LegacyAdditionalFlowStep;
```
kaeluka, line 540: I suspect, legacyadditionalflowstep can also be removed? (or at least be deprecated)

javascript/ql/lib/semmle/javascript/dataflow/TaintTracking.qll
```diff
@@ -456,22 +456,6 @@ module TaintTracking {
...
-   * predicate.
-   *
-   * A taint-propagating data flow edge that should be added to all taint tracking
-   * configurations in addition to standard data flow edges.
-   *
-   * Note: For performance reasons, all subclasses of this class should be part
-   * of the standard library. Override `Configuration::isAdditionalTaintStep`
-   * for analysis-specific taint steps.
-   */
-  deprecated class AdditionalTaintStep = InternalAdditionalTaintStep;
```
kaeluka, line 473: can `InternalAdditionalTaintStep` also be removed? (or at least be deprecated)

kaeluka (approved):
LGTM, but minor nitpicks

geoffw0 (dismissed):
C++ LGTM.

I actually wonder if 14 months is a little on the aggressive side, given that deprecated code is doing little harm until there's a lot of it cluttering up the libraries.  Not all of our users use CodeQL every day.  I expect others will disagree though, and I'm not going to protest any number >= 12 months.

RasmusWL:
> I actually wonder if 14 months is a little on the aggressive side, given that deprecated code is doing little harm until there's a lot of it cluttering up the libraries. Not all of our users use CodeQL every day. I expect others will disagree though, and I'm not going to protest any number >= 12 months.

Context: Erik made a similar PR some time ago where we discussed that 12 months was not enough (since commit will happen before PR is merged), but agreed that 14 months was probably fine.

RasmusWL (approved):
:+1: for Python. Thanks!

erik-krogh:
I deleted some more code thanks to the review by @kaeluka, and I added some change-notes.  
([It's the same change-note as last time](https://github.com/github/codeql/pull/8347)).

cpp/ql/lib/change-notes/2022-08-17-deleted-deprecations.md
```diff
@@ -0,0 +1,6 @@
+---
+category: minorAnalysis
+---
+* All deprecated predicates/classes/modules that have been deprecated for over a year have been 
+deleted.
```
geoffw0, line 5: Is this true?  A quick search of the CPP code base suggests there are a number of exceptions where we have not deleted older deprecated predicates, e.g. `File.getShortName()`???

erik-krogh: According to `git blame` the line containing the `deprecated` annotation for `File.getShortName()` was last touched 12 months ago.  (In [this commit](https://github.com/github/codeql/commit/6c32b92929473d4b020c18c07ebe60d3c6d0090c)). 
That is less than the 14 months required by my script, so it was therefore not deleted.  

That predicate has been deprecated for 4 years, but my script just looks at when the line containing the `deprecated` annotation was last touched.  

I could change the change-note to say "Some" or "Most"?

geoffw0: Sounds good.  It does look like `File.getShortName()` might just be an unusual case that got my attention.

kaeluka (approved):
JS: 👍 

Thanks for iterating!

smowton (approved):
Go lgtm

PR #10076, opened 2022-08-16
various QL-for-QL fixes

erik-krogh:
The first 4 commits were written automatically by my patching tool

### DIFF
--- a/cpp/ql/lib/experimental/semmle/code/cpp/semantic/SemanticExprSpecific.qll
+++ b/cpp/ql/lib/experimental/semmle/code/cpp/semantic/SemanticExprSpecific.qll
@@ -292,22 +292,20 @@ module SemanticExprConfig {
 
   class Guard = IRGuards::IRGuardCondition;
 
-  predicate guard(Guard guard, BasicBlock block) {
-    block = guard.(IRGuards::IRGuardCondition).getBlock()
-  }
+  predicate guard(Guard guard, BasicBlock block) { block = guard.getBlock() }
 
   Expr getGuardAsExpr(Guard guard) { result = guard }
 
   predicate equalityGuard(Guard guard, Expr e1, Expr e2, boolean polarity) {
-    guard.(IRGuards::IRGuardCondition).comparesEq(e1.getAUse(), e2.getAUse(), 0, true, polarity)
+    guard.comparesEq(e1.getAUse(), e2.getAUse(), 0, true, polarity)
   }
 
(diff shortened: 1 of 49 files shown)


aschackmull (dismissed):
LGTM. Although I wonder whether the import-removals are based on the import being unused or merely superfluous due to being transitively imported. In the latter case the proper fix was probably to remove the transitive import by making more imports private.

geoffw0 (approved):
C/C++ LGTM.

RasmusWL (dismissed):
Python :+1:

erik-krogh:
> LGTM. Although I wonder whether the import-removals are based on the import being unused or merely superfluous due to being transitively imported. In the latter case the proper fix was probably to remove the transitive import by making more imports private.

It's the latter. The query looks for imports that have already been transitively imported by an earlier import  (only 1 level down).

aschackmull:
> It's the latter. The query looks for imports that have already been transitively imported by an earlier import (only 1 level down).

Then I think I mildly disagree with the QL style that that query pushes towards. I'd rather have a QL-for-QL query that made those imports relevant by making the transitive imports private.

erik-krogh:
> Then I think I mildly disagree with the QL style that that query pushes towards. I'd rather have a QL-for-QL query that made those imports relevant by making the transitive imports private.

I'll look into that and update this PR.

aschackmull (dismissed):

smowton (dismissed):
Go lgtm

erik-krogh:
> > Then I think I mildly disagree with the QL style that that query pushes towards. I'd rather have a QL-for-QL query that made those imports relevant by making the transitive imports private.
> 
> I'll look into that and update this PR.

@aschackmull: Making the transitive imports private changes behaviour because some code relied on those imports being public.   
The [last commit](https://github.com/github/codeql/pull/10076/commits/4bc10f9b5c9489c5d7b19503468afc91af0eafd3) was required to things to compile again.  
I don't think there's a way to make those public imports deprecated without introducing a deprecation warning.   
This change will definitely require a change-note.   

But do you want this? Or should I just revert the changing of imports in `java/`?  
I'm leaning towards the latter.

aschackmull:
> Or should I just revert the changing of imports in `java/`?

That's fine (do leave the changes+revert in the commit history, though). Cc @atorralba in case you want to cherry-pick some of these changes into a separate PR.

aschackmull (approved):

PR #10077, opened 2022-08-16
Add query for tainted `wordexp` calls.

intrigus-lgtm:
This is my first query for C/C++ I think, so there is likely some improvement possible.

### DIFF
--- a/cpp/ql/src/experimental/Security/CWE/CWE-078/WordexpTainted.qhelp
+++ b/cpp/ql/src/experimental/Security/CWE/CWE-078/WordexpTainted.qhelp
@@ -9,7 +9,7 @@ vulnerable to attack by command injection, because <code>wordexp</code> performs
 </overview>
 <recommendation>
 
-<p>When calling <code>wordexp</code>, pass the <code>WRDE_NOCMD</code> flag to to prevent command substitution.</p>
+<p>When calling <code>wordexp</code>, pass the <code>WRDE_NOCMD</code> flag to prevent command substitution.</p>
 
 </recommendation>
 <example>
--- a/cpp/ql/src/experimental/Security/CWE/CWE-078/WordexpTainted.qhelp
+++ b/cpp/ql/src/experimental/Security/CWE/CWE-078/WordexpTainted.qhelp
@@ -4,7 +4,10 @@
 <qhelp>
 <overview>
 <p>The code passes user input to <code>wordexp</code>. This leaves the code
-vulnerable to attack by command injection, because <code>wordexp</code> performs command substitution.</p>
+vulnerable to attack by command injection, because <code>wordexp</code> performs command substitution.
+Command substitution is a feature that replaces <code>$(command)</code> or <code>`command`</code> with the
+output of the given command, allowing the user to run arbitrary code on the system.
+</p>
 
 </overview>
 <recommendation>
--- a/cpp/ql/src/experimental/Security/CWE/CWE-078/WordexpTainted.qhelp
+++ b/cpp/ql/src/experimental/Security/CWE/CWE-078/WordexpTainted.qhelp
@@ -1,0 +1,39 @@
+<!DOCTYPE qhelp PUBLIC
+  "-//Semmle//qhelp//EN"
+  "qhelp.dtd">
+<qhelp>
+<overview>
+<p>The code passes user input to <code>wordexp</code>. This leaves the code
+vulnerable to attack by command injection, because <code>wordexp</code> performs command substitution.</p>
+
+</overview>
+<recommendation>
+
+<p>When calling <code>wordexp</code>, pass the <code>WRDE_NOCMD</code> flag to to prevent command substitution.</p>
+
+</recommendation>
+<example>
+<p>The following example passes a user-supplied file path to <code>wordexp</code> in two ways. The
+first way uses <code>wordexp</code> with no specified flags. As such, it is vulnerable to command
+injection.
+The second way uses <code>wordexp</code> with the <code>WRDE_NOCMD</code> flag. As such, no command substitution
+is performed, making this safe from command injection.</p>
+<sample src="WordexpTainted.c" />
+
+</example>
+<references>
+
+<li>CERT C Coding Standard:
+<a href="https://www.securecoding.cert.org/confluence/display/c/STR02-C.+Sanitize+data+passed+to+complex+subsystems">STR02-C.
+Sanitize data passed to complex subsystems</a>.</li>
+<li>
+OWASP:
+<a href="https://www.owasp.org/index.php/Command_Injection">Command Injection</a>.
+</li>
+
+
+<!--  LocalWords: CWE STR
+ -->
+
+</references>
+</qhelp>
--- a/cpp/ql/test/experimental/query-tests/Security/CWE/CWE-078/WordexpTainted.expected
+++ b/cpp/ql/test/experimental/query-tests/Security/CWE/CWE-078/WordexpTainted.expected
@@ -1,0 +1,11 @@
+edges
+| test.cpp:23:20:23:23 | argv | test.cpp:29:13:29:20 | (const char *)... |
+| test.cpp:23:20:23:23 | argv | test.cpp:29:13:29:20 | filePath |
+nodes
+| test.cpp:23:20:23:23 | argv | semmle.label | argv |
+| test.cpp:29:13:29:20 | (const char *)... | semmle.label | (const char *)... |
+| test.cpp:29:13:29:20 | filePath | semmle.label | filePath |
+subpaths
+#select
+| test.cpp:29:13:29:20 | (const char *)... | test.cpp:23:20:23:23 | argv | test.cpp:29:13:29:20 | (const char *)... | Using user-supplied data in a `wordexp` command, without disabling command substitution, can make code vulnerable to command injection. |
+| test.cpp:29:13:29:20 | filePath | test.cpp:23:20:23:23 | argv | test.cpp:29:13:29:20 | filePath | Using user-supplied data in a `wordexp` command, without disabling command substitution, can make code vulnerable to command injection. |
--- a/cpp/ql/src/experimental/Security/CWE/CWE-078/WordexpTainted.c
+++ b/cpp/ql/src/experimental/Security/CWE/CWE-078/WordexpTainted.c
@@ -1,0 +1,19 @@
+
+int main(int argc, char** argv) {
+  char *filePath = argv[2];
+
+  {
+    // BAD: the user-controlled string is injected
+    // directly into `wordexp` which performs command substitution
+
+    wordexp_t we;
+    wordexp(filePath, &we, 0);
+  }
+
+  {
+    // GOOD: command substitution is disabled
+
+    wordexp_t we;
+    wordexp(filePath, &we, WRDE_NOCMD);
+  }
+}
--- a/cpp/ql/src/experimental/Security/CWE/CWE-078/WordexpTainted.ql
+++ b/cpp/ql/src/experimental/Security/CWE/CWE-078/WordexpTainted.ql
@@ -1,0 +1,53 @@
+/**
+ * @name Uncontrolled data used in `wordexp` command
+ * @description Using user-supplied data in a `wordexp` command, without
+ *              disabling command substitution, can make code vulnerable
+ *              to command injection.
+ * @kind path-problem
+ * @problem.severity error
+ * @precision high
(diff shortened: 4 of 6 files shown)


cpp/ql/src/experimental/Security/CWE/CWE-078/WordexpTainted.qhelp
```diff
@@ -0,0 +1,39 @@
...
+  "qhelp.dtd">
+<qhelp>
+<overview>
+<p>The code passes user input to <code>wordexp</code>. This leaves the code
+vulnerable to attack by command injection, because <code>wordexp</code> performs command substitution.</p>
+
+</overview>
+<recommendation>
+
+<p>When calling <code>wordexp</code>, pass the <code>WRDE_NOCMD</code> flag to to prevent command substitution.</p>
```
intrigus-lgtm, line 12:
```suggestion
<p>When calling <code>wordexp</code>, pass the <code>WRDE_NOCMD</code> flag to prevent command substitution.</p>
```

intrigus-lgtm:
This is inspired by https://github.com/syoyo/tinygltf/issues/368.
Sadly, my query can not find that specific instance.
The taint seems to get lost in `LoadFromString` around lines 5648.
`LoadFromString` -> `ParseBuffer` -> `LoadExternalFile` -> `FindFile` -> `fs->ExpandFilePath` (`ExpandFilePathFunction`) -> `ExpandFilePath` (line 2611) -> `wordexp` (line 2640)

(The used database is from https://lgtm.com/projects/g/syoyo/tinygltf/ci/#ql)

cpp/ql/src/experimental/Security/CWE/CWE-078/WordexpTainted.qhelp
```diff
@@ -0,0 +1,39 @@
+<!DOCTYPE qhelp PUBLIC
+  "-//Semmle//qhelp//EN"
+  "qhelp.dtd">
+<qhelp>
+<overview>
+<p>The code passes user input to <code>wordexp</code>. This leaves the code
+vulnerable to attack by command injection, because <code>wordexp</code> performs command substitution.</p>
```
geoffw0, line 7: It might be nice to expand on this for those unaware of the feature, maybe something along the lines of:
```
Command substitution is a feature that replaces $(command) or `command` with the
output of the given command, allowing the user to run arbitrary code on the system.
```

cpp/ql/src/experimental/Security/CWE/CWE-078/WordexpTainted.qhelp
```diff
@@ -0,0 +1,39 @@
...
+<a href="https://www.securecoding.cert.org/confluence/display/c/STR02-C.+Sanitize+data+passed+to+complex+subsystems">STR02-C.
+Sanitize data passed to complex subsystems</a>.</li>
+<li>
+OWASP:
+<a href="https://www.owasp.org/index.php/Command_Injection">Command Injection</a>.
+</li>
+
+
+<!--  LocalWords: CWE STR
+ -->
```
geoffw0, line 36: What is this?

geoffw0 (commented):
Hi @intrigus-lgtm, this look to be a nice little query, with great qldoc and a clear example.

I've made a couple of comments, I also intend to look into why taint is getting lost in the inspiring instance.

geoffw0:
Hi @intrigus-lgtm.  `syoyo/tinygltf` contains a call to `wordexp` in `tiny_gltf.h` line 2661.  However the LGTM build does not include this line (presumably because of the `#if` preprocessor condition a few lines up combined with the specific compliation flags / environment that was used in the build) - so the call is not in the database and your query can't be expected to find it as a result on this particular snapshot.  I wouldn't worry about it, but if you're really determined you could attempt to build a snaphot of the project with different settings, or write test cases that are closer to what you see there (e.g. exercising taint flow through a `std::string`).

I'm also suspicious you're right about taint not reaching the function, but I'm not sure where it stops or why (or whether its correct to).  Again, the query is not at fault.

intrigus-lgtm:
> Hi @intrigus-lgtm. `syoyo/tinygltf` contains a call to `wordexp` in `tiny_gltf.h` line 2661. However the LGTM build does not include this line (presumably because of the `#if` preprocessor condition a few lines up combined with the specific compliation flags / environment that was used in the build) - so the call is not in the database and your query can't be expected to find it as a result on this particular snapshot. I wouldn't worry about it, but if you're really determined you could attempt to build a snaphot of the project with different settings, or write test cases that are closer to what you see there (e.g. exercising taint flow through a `std::string`).
> 
> I'm also suspicious you're right about taint not reaching the function, but I'm not sure where it stops or why (or whether its correct to). Again, the query is not at fault.

I guess the problem is that lgtm.com received the new commits that patched the problem (by commenting out the vulnerable code), so the db does not contain any calls to `wordexp` anymore.
You can try this slightly older database instead that works for me:
https://github.com/intrigus-lgtm/file-dump/blob/master/syoyo_tinygltf_cpp-srcVersion_81f7dbe53a112d05217a79bb1c986f9ada3b6631-dist_codeql-bundle-linux64-20220811-2843018186.zip?raw=true

I could successfully find the call to `wordexp` in that database yesterday, but taint is getting lost as described in https://github.com/github/codeql/pull/10077#issuecomment-1217220464

cpp/ql/src/experimental/Security/CWE/CWE-078/WordexpTainted.qhelp
intrigus-lgtm: Artifact from the query help I based this on. Should I remove it?
I don't know what exactly it is doing, but some other query help files also contain this part:
https://github.com/github/codeql/search?q=LocalWords

geoffw0:
Thanks for that update ... I think the taint issue might be that the flow into `ParseBuffer` is in a lambda expression, its likely such flows are less well tested that other more common situations.  For some reason (perhaps to do with capturing the `base_dir` variable, or uncertainty that the lambda is called???) taint flow isn't working there.  I'll file an internal issue so we get this fixed at some point.

cpp/ql/src/experimental/Security/CWE/CWE-078/WordexpTainted.qhelp
geoffw0: I asked around and we think they're there to make some spell checking software happy with words that aren't in its dictionary.  I don't think it matters much but it seems harmless to leave them in.

cpp/ql/src/experimental/Security/CWE/CWE-078/WordexpTainted.qhelp
intrigus-lgtm, line 7:
```suggestion
vulnerable to attack by command injection, because <code>wordexp</code> performs command substitution.
Command substitution is a feature that replaces <code>$(command)</code> or <code>`command`</code> with the
output of the given command, allowing the user to run arbitrary code on the system.
</p>
```

cpp/ql/test/experimental/query-tests/Security/CWE/CWE-078/WordexpTainted.expected
```diff
@@ -0,0 +1,11 @@
+edges
+| test.cpp:23:20:23:23 | argv | test.cpp:29:13:29:20 | (const char *)... |
+| test.cpp:23:20:23:23 | argv | test.cpp:29:13:29:20 | filePath |
+nodes
+| test.cpp:23:20:23:23 | argv | semmle.label | argv |
+| test.cpp:29:13:29:20 | (const char *)... | semmle.label | (const char *)... |
+| test.cpp:29:13:29:20 | filePath | semmle.label | filePath |
+subpaths
+#select
+| test.cpp:29:13:29:20 | (const char *)... | test.cpp:23:20:23:23 | argv | test.cpp:29:13:29:20 | (const char *)... | Using user-supplied data in a `wordexp` command, without disabling command substitution, can make code vulnerable to command injection. |
```
intrigus-lgtm, line 10: If I read this right, CodeQL identifies two different sinks at the same position and therefore generates two alerts, doesn't it?
I expected only one alert, is there a way to fix this?

MathiasVP: That's usually due to implicit conversions. In this case, I think it's because of the implicit conversion from `char*` to `const char*`. You can try replacing
```ql
fc.getArgument(0) = sink.asExpr() and
```
with
```ql
fc.getArgument(0).getFullyConverted() = sink.asConvertedExpr() and
```

Alternatively, you can make the `filePath` in your example a `const char*` to avoid the implicit conversion.

intrigus-lgtm: In general, when looking for sinks, should one always use the  `foo.getFullyConverted() = sink.asConvertedExpr()` pattern to get unique results?
In Java I always just wrote `sink.asExpr()` :D

geoffw0: Note that the code scanning / LGTM interfaces will collapse the duplicate results somewhat, so while its nice to fix them they should not affect end users too badly.

MathiasVP: > In general, when looking for sinks, should one always use the foo.getFullyConverted() = sink.asConvertedExpr() pattern to get unique results?
In Java I always just wrote sink.asExpr() :D

That's because Java represents conversions differently from how it's done in C/C++ (AFAIK, they're more explicit in Java). Since we can't do that for C/C++ since basically any expression would be hidden behind layers of nested implicit and explicit conversions.

So in C/C++, when you have snippet like:
```cpp
void test(void* ptr) {
  int* x = (int*)ptr;
}
```
and you ask for the right-hand side of the initialization, you'll get back `ptr` (and _not_ `(int*)ptr`). You can then ask for the fully converted expression of `ptr` and get back `(int*)ptr`.

As @geoffw0 said, this isn't a super big deal as Code Scanning (and LGTM) will collapse some of these results together.

intrigus-lgtm: Okay, then I'll leave the query as-is, since this is just in a test and won't be a problem for end users :)

geoffw0:
I've started the checks, and run a moderately large LGTM run of the query here: https://lgtm.com/query/1704064894746891004/ (results look good at a glance)

geoffw0 (dismissed):
I think this is ready to merge.  Thank you for your contribution!

intrigus-lgtm:
Just to be sure, this does not need a review of the security lab?

MathiasVP:
> Just to be sure, this does not need a review of the security lab?

Yes, it will. It will be merged once Security Lab has reviewed it and are happy with it :)

JarLob:
One FP pattern I have noticed is when the taint from `char*` is transferred to `int` through `strlen` and causes false positives. I think for this query a string to integer sanitizer is need.

I may add more comments in the thread as I review the query.

geoffw0:
> I think for this query a string to integer sanitizer is need.

Its fairly common to simply block flow through all integer typed expressions, in order to stop this sort of thing.  Something like:
```
  node.asExpr().getUnspecifiedType() instanceof IntegralType
}
```

JarLob:
The query doesn't find the vulnerability in tinygltf built from changset 0fa56e239c77cc864dc248842e8887d985cf8e3f.

JarLob:
I see why it doesn't find. Please, disregard.

JarLob:
I switched the sink to `any()` to see all reachable nodes (I think it was Pavel who shown the technique in one of CodeQL videos) and it looks to me the taint stops at `bool fileread = fs.ReadWholeFile(&data, &fileerr, filename, fs.user_data);` which is tinygltf specific function. I think CodeQL doesn't understand the taint flows to the first `std::vector<unsigned char> *out` parameter. I think it is ok to not include tinygltf specific steps in the general query.

JarLob:
@intrigus-lgtm Could you please add the sanitizer?

intrigus-lgtm:
@JarLob done, sorry for the delay.

geoffw0 (approved):
Sorry I lost track of this.  I think its ready to merge into experimental.

geoffw0:
Oh, and I'll run the checks now...

intrigus-lgtm:
@geoffw0 anything else missing?

geoffw0 (approved):
> @geoffw0 anything else missing?

I don't think so.  Lets merge this.

PR #10082, opened 2022-08-17
JS: Handle nested conditions in "exports" section

asgerf:
Follow-up to https://github.com/github/codeql/pull/10036, to support nested conditions in `"exports"` again.

This did not affect any of the project in the default suite, but when [evaluated](https://github.com/github/codeql-dca-main/blob/data/asgerf/exports-handlin__sql-libraries1__CustomSuite/reports/alert-meta-comparison.md) on one of the SQL-specific suites, there are some affected projects.


erik-krogh (approved):
👍

PR #10083, opened 2022-08-17
Java: Add some type-based sanitizers to SensitiveInfoLog.ql.

aschackmull:
The source definition is rather ad-hoc in this query, so filtering nodes based on some safe types can help remove FPs.


smowton (approved):

aschackmull:
Dca looks good. Added change note.

atorralba (approved):

PR #10084, opened 2022-08-17
Java: Move sink-constraints into the configuration in NumericCastTainted.ql.

aschackmull:
The exclusion of right-shifted variables on sinks is probably better as a sanitizer, so move it there. That should help performance a bit and possibly also remove some FPs.

Also, there was a somewhat arbitrary-looking restriction that the sink has to be a variable access. I'm not going to consider whether that restriction ought to lifted (it's probably fine to keep), but we might as well make it explicit and move it into the configuration to gain some performance.


aschackmull:
Dca looks very uneventful.

atorralba (approved):
LGTM. I have the gut feeling that restricting the sinks to `VarAccess`es wasn't intentional, so removing it would probably be fine.

PR #10086, opened 2022-08-17
C#: CIL Extractor option.

michaelnebel:
In this PR we introduce a new C# extractor option, which indicates, whether CIL extraction should be enabled or not.
Furthermore, CIL extraction is now enabled by default (the converse held before).
To disable CIL extraction you need to use the new extractor option.

codeql database create --language=csharp -Ocil=true db
codeql database create --language=csharp --extractor-option=cil=true db

In tests CIL extraction is still disabled by default.


csharp/ql/src/change-notes/2022-08-24-remove-cil-flag.md
```diff
@@ -0,0 +1,4 @@
+---
+category: minorAnalysis
+---
+* The `--cil` flag that could be provided to the `codeql` CLI has been replaced by an extactor option. Furthermore, the extractor default is now set to do CIL extraction. To disable CIL extraction provide `-Ocil=false` to the CLI.
```
sidshank, line 4: In the sentence beginning with `To disable CIL extraction provide...`, should the following command line switch be:

`--cil=false` 

instead of

`-0cil=false`

at the end?

michaelnebel: The syntax for specifying extractor options can be seen in `Specifying extractor options directly` in https://github.com/github/codeql-core/blob/main/design/draft/codeql-cli-extractor-options.md
That is we need to provide `-O' or `--extractor-option` to provide an extractor option.
It was good that you brought this to my attention, as the documentation also contains some information that we might be able to use to get rid of the `--cil` flag completely (also for testing purposes). I will look into this.

michaelnebel: It seems that the .yml files mentioned in the documentation are not respected during `codeql test run`. Will stick to the solution as is.

hvitved: It was actually not possible to disable CIL extraction before, so I would go with
```
A new extractor option has been introduced for disabling CIL extraction. Either pass `-Ocil=false` to the `codeql` CLI or set the environment variable `CODEQL_EXTRACTOR_CSHARP_OPTION_CIL=false`.
```

hvitved (changes requested):
LGTM, only one suggestion.

hvitved (approved):

PR #10087, opened 2022-08-17
Ruby: Get rid of warning in `getUnknownMember`

hvitved:
Fixes https://github.com/github/codeql-team/issues/929.

### DIFF
--- a/ruby/ql/lib/codeql/ruby/ApiGraphs.qll
+++ b/ruby/ql/lib/codeql/ruby/ApiGraphs.qll
@@ -718,6 +718,10 @@ module API {
         or
         succ = MkMethodAccessNode(entry.getACall())
       )
+      or
+      // to get rid of an incompatible types error in `getUnknownMember`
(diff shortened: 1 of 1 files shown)


hvitved:
> do we actually use `UnknownMember` at all

No. I think we just like to have it there for future use, right @erik-krogh ?

erik-krogh:
> No. I think we just like to have it there for future use, right @erik-krogh ?

I don't remember having anything to do with the decision to add the `getUnknownMember()` predicate in Ruby.  

But I think it was done to be consistent with the other languages.  
Both Python and JavaScript have plenty of member reads where the name is unknown.  
If that isn't really a thing in Ruby then I think you should just delete the predicate.

hvitved:
> If that isn't really a thing in Ruby then I think you should just delete the predicate.

OK, I'll remove it instead then.

nickrolfe (approved):

PR #10089, opened 2022-08-17
Ruby: Reduce size of `isLocalSourceNode`

hvitved:
Alternative to https://github.com/github/codeql/pull/9175.

I have verified, locally, that the analysis is now able to run on https://github.com/rapid7/metasploit-framework without timing out.


nickrolfe (approved):
Thanks for fixing this.

I can't claim to understand the how you figured out those two new disjuncts are required, but the tests pass, DCA looks good, and you say it fixes the performance problems on `metasploit-framework`, so I'm happy.

The query simplifications look sensible.

PR #10090, opened 2022-08-18
Ruby: Model Activestorage

hmac:
Add modelling for the [activestorage](https://rubygems.org/gems/activestorage) Ruby library. This library provides a framework for uploading files to cloud services and performing image, audio and video transformations on them.

This involves adding instances for several Concepts: `FileSystemWriteAccess`, `HTTP::Client::Request`, `SystemCommandExecution` and `CodeExecution`.

The meat of it is in `SystemCommandExecution`, as there are many ways to inject a command into an image transformation that ultimately gets executed by a subshell.

As part of this I've modified the `EntryPoint` class, adding an `edge` member predicate which allows you to construct custom API entry points at arbitrary points in the graph, rather than just at the root. This is explained in more detail in the code.

### DIFF
--- a/ruby/ql/test/library-tests/frameworks/active_storage/ActiveStorage.ql
+++ b/ruby/ql/test/library-tests/frameworks/active_storage/ActiveStorage.ql
@@ -1,6 +1,5 @@
 import ruby
 import codeql.ruby.ApiGraphs
-import codeql.ruby.DataFlow
 import codeql.ruby.Concepts
 import codeql.ruby.frameworks.ActiveStorage
 


ruby/ql/lib/codeql/ruby/frameworks/ActiveStorage.qll
```diff
@@ -8,23 +8,236 @@ private import codeql.ruby.Concepts
...
+          assoc.isSingular() and this.getMethodName() = model
+          or
+          assoc.isCollection() and this.getMethodName() = model
+        )
+      )
+    }
+  }
+
+  /**
+   * A call on an ActiveStorage  object that results in an image transformation.
```
nickrolfe, line 139:
```suggestion
   * A call on an ActiveStorage object that results in an image transformation.
```

ruby/ql/lib/codeql/ruby/frameworks/ActiveStorage.qll
```diff
@@ -8,23 +8,236 @@ private import codeql.ruby.Concepts
...
+
+    override DataFlow::Node getCode() { result = this.getArgument(0) }
+  }
+
+  /**
+   * Adds ActiveStorage instances to the API graph.
+   * Source code may not mention `ActiveStorage` or `ActiveStorage::Attachment`,
+   * so we add synthetic nodes for them.
+   */
+  private module ApiNodes {
```
asgerf, line 196: Before MaD, we would generally handle this by introducing a predicate returning the set of API nodes of a certain type. For example, we'd have:
```ql
API::Node activeStorageStatic() {
  result = API::getTopLevelMember("ActiveStorage")
}

API::Node attachmentStatic() {
  result = activeStorageStatic().getMember("Attachment")
}

API::Node attachmentInstance() {
  result = attachmentStatic().getInstance()
  or
  result = any(AttachmentInstanceEntryPoint e).getANode()
}

class AttachmentInstanceEntryPoint extends API::EntryPoint {
  AttachmentInstanceEntryPoint() { this = "AttachmentInstanceEntryPoint" }

  DataFlow::LocalSourceNode getAUse() { result instanceof AttachmentInstance }
}
```

And then instead of writing out `API::getTopLevelMember("ActiveStorage").getMember("Attachment").getInstance()` we'd use a call to `attachmentInstance()`. (With MaD, this all becomes type-definitions rows instead, but there is no pressure to use MaD right now)

Is there a reason to add this new way of doing it? One thing I worry about is that the choice of access path is arbitrary, and there's no way to know which path you want to use at the use-site. For example, suppose `ABC::Attachment` was an alias for `ActiveStorage::Attachment`, would you expect `API::getTopLevelMember("ABC").getMember("Attachment")` to work as well?

ruby/ql/lib/codeql/ruby/frameworks/ActiveStorage.qll
```diff
@@ -8,23 +8,236 @@ private import codeql.ruby.Concepts
...
+        [
+          // ActiveStorage::Blob.new : Blob
+          "activestorage;Blob;activestorage;;Member[ActiveStorage].Member[Blob].Instance",
+          // ActiveStorage::Blob.create_and_upload! : Blob
+          "activestorage;Blob;activestorage;;Member[ActiveStorage].Member[Blob].Method[create_and_upload!].ReturnValue",
+          // ActiveStorage::Blob.create_before_direct_upload! : Blob
+          "activestorage;Blob;activestorage;;Member[ActiveStorage].Member[Blob].Method[create_before_direct_upload!].ReturnValue",
+          // ActiveStorage::Blob.compose(blobs : [Blob]) : Blob
+          "activestorage;Blob;activestorage;;Member[ActiveStorage].Member[Blob].Method[compose].ReturnValue",
+          // gives error: Invalid name 'Element' in access path
```
alexrford, line 56: Does this work with an unparameterized `Element` (i.e. `"activestorage;Blob;activestorage;;Member[ActiveStorage].Member[Blob].Method[compose].Argument[0].Element"`) or with `ArrayElement` instead (i.e. `"activestorage;Blob;activestorage;;Member[ActiveStorage].Member[Blob].Method[compose].Argument[0].ArrayElement"`)?

ruby/ql/lib/codeql/ruby/frameworks/ActiveStorage.qll
```diff
@@ -8,23 +8,236 @@ private import codeql.ruby.Concepts
...
+   * Method calls on `ActiveStorage::Blob` that send HTTP requests.
+   */
+  private class BlobRequestCall extends HTTP::Client::Request::Range {
+    BlobRequestCall() {
+      this =
+        [
+          // Class methods
+          API::getTopLevelMember("ActiveStorage")
+              .getMember("Blob")
+              .getASubclass*()
```
alexrford, line 78: Is the transitive closure necessary here? `getASubclass()` is defined as `Node getASubclass() { result = this.getAnImmediateSubclass*() }`

ruby/ql/lib/codeql/ruby/frameworks/ActiveStorage.qll
```diff
@@ -8,23 +8,236 @@ private import codeql.ruby.Concepts
...
+   *
+   * user = User.find(id)
+   * user.avatar
+   * ```
+   */
+  private class AttachmentInstance extends DataFlow::CallNode {
+    AttachmentInstance() {
+      exists(Association assoc, string model | model = assoc.getTargetModelName() |
+        this.getReceiver().(ActiveRecordInstance).getClass() = assoc.getSourceClass() and
+        (
```
alexrford, line 129: Are there associations that are neither singular or collections? Otherwise this looks equivalent to just `this.getMethodName() = model`.

alexrford (changes requested):
Very minor nitpicks/questions, otherwise looks good!

ruby/ql/lib/codeql/ruby/frameworks/ActiveStorage.qll
asgerf: After thinking about this some more, I realised that the API graph edge relation this now mutually recursive with things like `getTopLevelMember` and `getMember` and each of these `EntryPoint.edge` predicates.

For instance, this means this implementation,
```codeql
      override predicate edge(API::Node pred, API::Label::ApiLabel lbl) {
        pred = API::getTopLevelMember("ActiveStorage").getMember("Attachment").getMethod("new") and
        lbl = API::Label::return()
      }
```
is a recursive predicate with a three-way recursive join (2x `getMember` and 1x `getTopLevelMember`). This is pretty risky performance-wise, and even if it appears to scale now it's the kind of thing that can become a problem later.

asgerf: Thanks! I'm [working on adding](https://github.com/github/codeql/pull/10276) the missing hook so you can contribute to the MaD type-def from CodeQL. I expect it would look something like this,
```codeql
class AttachmentInstanceType extends ModelInput::TypeModel {
  override DataFlow::LocalSourceNode getASource(string package, string type) {
    package = "activestorage" and
    type = "Attachment" and
    result instanceof AttachmentInstance
  }
}
```
Then you could write the MaD lines like this
```js
// ActiveStorage::Attachment#blob : Blob    
"activestorage;Blob;activestorage;Attachment;Method[blob].ReturnValue",
// ActiveStorage::Attachment delegates method calls to its associated Blob       
"activestorage;Blob;activestorage;Attachment;",
```

But if you can work around it for now that would be great.

asgerf: [The PR with TypeModel](https://github.com/github/codeql/pull/10276) has been merged, it would be great if you could use it in this PR, but I'll leave it up to you.

alexrford (approved):
Nice, LGTM

PR #10091, opened 2022-08-18
Kotlin: Identify data classes during extraction

tamasvajk:


java/downgrades/81ccfabe82e696953268e784979262e56871ce86/upgrade.properties
```diff
@@ -0,0 +1,3 @@
+description: Remove ktDataClasses relation
```
tamasvajk, line 1: The SHA of this folder is probably not good. And maybe the old and new schemas need to be swapped.

tamasvajk:
The downgrade script passes the new CI check: https://github.com/github/semmle-code/runs/7917300759?check_suite_focus=true#step:18:1

tamasvajk:
I checked the difference in successfully extracted files in the DCA run. I'm unable to repro the reported results. I'm getting 164 files locally, which is different from both the before (171) and after (167) results.

There are 4 files which show valid extraction errors with source file based locations. I'm fixing this extraction error in a separate PR.

igfoo:
Does rerunning DCA produce the same results?

I think the debug artifacts should contain enough to work out which files were missing, and whether there were any log messages about them. Are there any clues in there?

igfoo (approved):
We found that the DCA issue was unrelated to this PR.

PR #10093, opened 2022-08-18
Java: pick an arbitrary representative location when an entity has many candidate locations.

smowton:


aschackmull (dismissed):
LGTM.

smowton:
@aschackmull I needed to do some join-order engineering after this perturbed InlineExpectationsTest into a bad enough order that tests were timing out.

michaelnebel (approved):
Looks plausible to me!

erik-krogh (approved):
QL-for-QL 👍

MathiasVP (approved):

PR #10096, opened 2022-08-18
make acronyms camelcase

erik-krogh:
[Acronyms should use normal PascalCase/camelCase](https://github.com/github/codeql/blob/main/docs/ql-style-guide.md#naming).  

This is a followup to https://github.com/github/codeql/pull/10073, and there might be more later. 

Most of this PR was written automatically by my patching tool.  

A QLDoc check was complaining about some missing QLDoc, so I fixed that (even though I didn't introduce the code 🤷).


java/ql/lib/semmle/code/xml/XML.qll
```diff
@@ -174,7 +177,10 @@ class XmlDTD extends XmlLocatable, @xmldtd {
   }
 }
 
-/** DEPRECATED: Alias for XmlDTD */
+/** DEPRECATED: Alias for XmlDtd */
+deprecated class XmlDTD = XmlDtd;
+
+/** DEPRECATED: Alias for XmlDtd */
 deprecated class XMLDTD = XmlDTD;
```
aschackmull, line 184: This was just  introduced in your prioir PR, so I think we can combine these aliases.
```suggestion
/** DEPRECATED: Alias for XmlDtd */
deprecated class XMLDTD = XmlDtd;
```

michaelnebel (dismissed):
C# LGTM

java/ql/src/experimental/Security/CWE/CWE-327/SslLib.qll
```diff
@@ -94,6 +94,9 @@ class UnsafeTlsVersion extends StringLiteral {
...
 }
 
-class SSLServerSocket extends RefType {
-  SSLServerSocket() { hasQualifiedName("javax.net.ssl", "SSLServerSocket") }
+class SslServerSocket extends RefType {
+  SslServerSocket() { hasQualifiedName("javax.net.ssl", "SSLServerSocket") }
 }
+
+/** DEPRECATED: Alias for SslServerSocket */
+deprecated class SSLServerSocket = SslServerSocket;
```
aschackmull, line 102: I don't think we need a deprecated alias in the `experimental` dir.

aschackmull (commented):
A couple of comments, otherwise Java LGTM.

smowton:
In Go I notice a couple of cases where a `newtype` constructor doesn't get a deprecated alias (though the type name itself does), though perhaps that's too technically tricky to recreate?

smowton:
Otherwise Go LGTM

python/ql/src/Security/CWE-327/PyOpenSSL.qll
```diff
@@ -80,6 +86,9 @@ class PyOpenSSL extends TlsLibrary {
   override ProtocolRestriction protocol_restriction() { result instanceof SetOptionsCall }
 
   override ProtocolUnrestriction protocol_unrestriction() {
-    result instanceof UnspecificPyOpenSSLContextCreation
+    result instanceof UnspecificPyOpenSslContextCreation
   }
 }
+
+/** DEPRECATED: Alias for PyOpenSsl */
+deprecated class PyOpenSSL = PyOpenSsl;
```
RasmusWL, line 94:
```suggestion
```

this class is not importable by anything outside the CWE-327 directory, so doesn't need a deprecation.

python/ql/src/Security/CWE-327/PyOpenSSL.qll
```diff
@@ -51,12 +54,15 @@ class SetOptionsCall extends ProtocolRestriction, DataFlow::CallCfgNode {
...
-  UnspecificPyOpenSSLContextCreation() { library instanceof PyOpenSSL }
+class UnspecificPyOpenSslContextCreation extends PyOpenSslContextCreation, UnspecificContextCreation {
+  UnspecificPyOpenSslContextCreation() { library instanceof PyOpenSsl }
 }
 
-class PyOpenSSL extends TlsLibrary {
-  PyOpenSSL() { this = "pyOpenSSL" }
+/** DEPRECATED: Alias for UnspecificPyOpenSslContextCreation */
+deprecated class UnspecificPyOpenSSLContextCreation = UnspecificPyOpenSslContextCreation;
+
```
RasmusWL, line 63:
```suggestion
```

python/ql/src/Security/CWE-327/Ssl.qll
```diff
@@ -181,6 +190,9 @@ class UnspecificSSLDefaultContextCreation extends SSLDefaultContextCreation, Pro
   }
 }
 
+/** DEPRECATED: Alias for UnspecificSslDefaultContextCreation */
+deprecated class UnspecificSSLDefaultContextCreation = UnspecificSslDefaultContextCreation;
+
```
RasmusWL, line 195:
```suggestion
```

java/ql/lib/semmle/code/xml/XML.qll
jketema: Same for C++.

RasmusWL (changes requested):
Python :+1: besides the requested changes

java/ql/lib/semmle/code/xml/XML.qll
erik-krogh: > Same for C++.

And the other languages, I'm fixing all in one go.

erik-krogh:
> In Go I notice a couple of cases where a `newtype` constructor doesn't get a deprecated alias (though the type name itself does), though perhaps that's too technically tricky to recreate?

It's because it's `private`. It's not accessible from the outside, so it doesn't need a deprecated alias.

smowton (approved):
Yeah ok on closer inspection the deprecated alias was for a nearby entity not the newtype itself

jketema (approved):
C++ 👍

python/ql/lib/semmle/python/frameworks/Stdlib.qll
```diff
@@ -1845,7 +1845,7 @@ private module StdlibPrivate {
   deprecated module SimpleHTTPServer = SimpleHttpServer;
 
   // ---------------------------------------------------------------------------
-  // CGIHTTPServer (Python 2 only)
+  // CgiHttpServer (Python 2 only)
```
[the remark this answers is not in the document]
erik-krogh: I had reverted a bunch of comment rewrites, but I missed that one.

RasmusWL (approved):
Python :+1:

aschackmull (approved):
Java 👍

cpp/ql/lib/semmle/code/cpp/XML.qll
```diff
@@ -174,8 +177,8 @@ class XmlDTD extends XmlLocatable, @xmldtd {
   }
 }
 
-/** DEPRECATED: Alias for XmlDTD */
-deprecated class XMLDTD = XmlDTD;
+/** DEPRECATED: Alias for XmlDtd */
+deprecated class XMLDTD = XmlDtd;
```
geoffw0, line 181: I take it we haven't released with `XmlDTD`?  So we don't also need
```
deprecated class XmlDTD= XmlDtd;
```
as well?

erik-krogh: Correct, we don't need that.   
I've removed the redundant deprecation in [this commit](https://github.com/github/codeql/commit/82d9180892314c87e6e778f209edea1c26efcd59).

alexrford (approved):
Ruby 👍

javascript/ql/lib/change-notes/2022-08-22-xml-rename.md
```diff
@@ -1,5 +1,5 @@
 ---
 category: deprecated
 ---
-* Classes/predicates that had upper-case acronym XML in their name have been renamed to Xml to follow our style-guide. 
+* Many classes/predicates/modules with upper-case acronyms in their name have been renamed to follow our style-guide.
```
kaeluka, line 4:
```suggestion
Many classes/predicates/modules with upper-case acronyms in their name have been renamed to follow our style-guide. The old spelling is still supported, but now deprecated.
```

kaeluka (approved):
JS: 👍 LGTM. One comment is that the change note(s) make it sound as if this would be breaking a lot of customer code, when the old spelling is actually still supported in the form of a deprecated alias.

PR #10098, opened 2022-08-18
QL: Add redundant overrides query

hvitved:
Inspired by https://github.com/github/codeql/pull/10094 (which it finds, in addition to a few others).


erik-krogh (approved):
Nice 👍  

You could probably add the check that the arguments to the super call are unmodified.  
But this is definitely good enough.

PR #10099, opened 2022-08-18
C#: Add a cshtml-based XSS test

hvitved:
Thanks to @michaelnebel for the newly introduced AspNetCore stubs, which makes this possible.


csharp/ql/test/query-tests/Security Features/CWE-079/XSS/Index.cshtml.g.cs
```diff
@@ -0,0 +1,62 @@
...
+  
+    ViewData["Title"] = "ASP.NET Core";
+    var message = Request.Query["m"];
+
+#line default
+#line hidden
+#nullable disable
+            WriteLiteral("<div class=\"cli\">\n    <div class=\"cli-example\">        \n");
+#nullable restore
+#line 14 "Index.cshtml"
```
michaelnebel, line 40: It doesn't look like the Index.cshtml file contains any code that would have generated this (even though the "generated" code here i handwritten).

hvitved: The contents of the `.cshtml` file is not important; it is only added because the extractor would otherwise fail when it attempts to add the file to the source archive (because the file is referenced with a `#line` directive).

michaelnebel: Alright - understood. Maybe it is of less importance, but I was just worried that it might be confusing when revisited later that the content of the .cs file doesn't correspond to the content of the .cshtml file (even though it is stated that it is handwritten). One could consider, if we should add a comment in the .cshtml file with your statement from above.

michaelnebel (approved):
Looks good to me!

PR #10103, opened 2022-08-18
Go: support go 1.19

smowton:


go/ql/test/library-tests/semmle/go/frameworks/StdlibTaintFlow/Fmt.go
```diff
@@ -1,4 +1,5 @@
-// Code generated by https://github.com/gagliardetto/codebox. DO NOT EDIT.
+// Code generated by https://github.com/gagliardetto/codebox, with manual additions for
+// go 1.19's new Append[f, ln] functions.
```
igfoo, line 2: Can we modify the generator instead?

igfoo (approved):
Otherwise looks plausible to me

go/ql/test/library-tests/semmle/go/frameworks/StdlibTaintFlow/Fmt.go
smowton: The generator is a third-party tool that we don't routinely make use of, so as of now I don't want to wade into figuring it out, updating it for Go 1.18+, etc

PR #10104, opened 2022-08-18
RB: introduce the cached stages pattern for Ruby

erik-krogh:
When you look at the changes and relate that to how the CodeQL evaluator works, then you might think that there should be no performance change at all from these changes.   

But there is: [Evaluation shows ~1.5% performance increase](https://github.com/github/codeql-dca-main/tree/data/erik-krogh/rbPerf__nightly__code-scanning__6/reports).  

The change there to be less stages at compile time, which can cause more predicates to be compiled together, which might allow the optimizer to share more intermediate results. 
That's probably not the whole explanation, but it might be some of it. 
At least that's what I think. I haven't been able to confirm why it's faster, just that it definitely is faster. 

There are some drawbacks:  
 - More cached predicates are evaluated together, meaning that in some cases predicates will be computed without their result ever being used. 
 - It's more code.   
 
 So I'll leave it to the Ruby team whether you want this PR merged or not.

### DIFF
--- a/ruby/ql/lib/codeql/ruby/AST.qll
+++ b/ruby/ql/lib/codeql/ruby/AST.qll
@@ -17,27 +17,21 @@ private import ast.internal.AST
 private import ast.internal.Scope
 private import ast.internal.Synthesis
 private import ast.internal.TreeSitter
+private import codeql.ruby.internal.CachedStages
 
-cached
-private module Cached {
-  cached
-  ModuleBase getEnclosingModule(Scope s) {
-    result = s
-    or
-    not s instanceof ModuleBase and result = getEnclosingModule(s.getOuterScope())
-  }
-
-  cached
-  MethodBase getEnclosingMethod(Scope s) {
-    result = s
-    or
-    not s instanceof MethodBase and
-    not s instanceof ModuleBase and
-    result = getEnclosingMethod(s.getOuterScope())
-  }
+private ModuleBase getEnclosingModule(Scope s) {
+  result = s
+  or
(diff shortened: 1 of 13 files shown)


hvitved:
I am not convinced about the performance change, I would usually consider such a small improvement a flux.

@hmakholm : I have not yet watched your presentation on the evaluator, which probably answers my question, but I was under the impression that stages no longer had an impact on `codeql analyze`, as we are doing whole-suite compilation to make sure that all things are shared and only evaluated once?

erik-krogh:
> I was under the impression that stages no longer had an impact on `codeql analyze`, as we are doing whole-suite compilation to make sure that all things are shared and only evaluated once?

I agree completely. But when I tried to revert the cached-stages-pattern for JS I saw a 4% performance regression, so something is happening.  

This is the full explanation I got from Henning the last time I asked him:  
_I think that has to do with how the optimizer works on one stage at a time. If two of the predicates in your collapsed stage share some intermediate results, having them in the same stage will make the optimizer create a single version of them which is compiled to work in both contexts. Otherwise the shared part will be optimized twice, and context from one of the calls can seep into the RA versions during optimization and mean that the evaluator will need to do almost-but-not-quite the same computation for the two stages._ 

And the performance increase is not just a fluke (each source has run multiple times).  
The exact number might be slightly off, but performance is better.

hmakholm:
FWIW, we're not doing "whole-suite-compilation".

_Compilation_ is still done query for query, stage by stage. The _evaluator_ shares computations if and only if parts of the RA the compiler produced happen to be identical between several queries/stages.

hvitved:
Let's close this, for now.

PR #10108, opened 2022-08-19
Remove redundant overrides

hvitved:
Created using an automatic patch based on https://github.com/github/codeql/pull/10102.

### DIFF
--- a/ql/ql/test/callgraph/Foo.qll
+++ b/ql/ql/test/callgraph/Foo.qll
@@ -10,9 +10,7 @@ class Foo extends AstNode {
   predicate baz() { bar() }
 }
 
-class Sub extends Foo {
-  override predicate baz() { super.baz() }
-}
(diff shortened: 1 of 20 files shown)


ql/ql/test/callgraph/Foo.qll
```diff
@@ -10,9 +10,7 @@ class Foo extends AstNode {
   predicate baz() { bar() }
 }
 
-class Sub extends Foo {
```
erik-krogh, line 13: You should revert the change to this file.  
It's a QL-for-QL test file. 

(Or you could make that override non-redundant, and update the expected test output).

erik-krogh (approved):
QL / JS 👍

MathiasVP (approved):

tausbn (approved):
Python :+1:

alexrford (approved):
Ruby 👍

PR #10111, opened 2022-08-19
Swift: synthesize `MethodRefExpr`

redsun82:
This introduces a `MethodRefExpr` node synthesized out of
`DotSyntaxCallExpr` under the `LookupExpr` hierarchy. This means that
much like
```free_function(1, 2)```
is a `CallExpr` with `getFunction` giving a `DeclRefExpr`,
```foo.method(1, 2)```
is now a `CallExpr` with `getFunction` giving a `MethodRefExpr`.

`ApplyExpr::getStaticTarget` has been made to work with it (as well as
with `ConstructorRefCallExpr` which for the moment has been left where it
is), a new `MethodApplyExpr` has been introduced deriving from it,
and control and data flow libraries have adapted.

A small bug was fixed in `qlgen` where the default constructor for DB
types was not correctly subtracting derived IPA types depending on the
order of definitions in `schema.yml`.

There are still some occurrences of `DotSyntaxCallExpr`, and as already
mentioned the other `SelfApply` class (`ConstructorRefCallExpr`) was
left alone. Their treatment is left for a future PR.

### CODE REVIEW

File: swift/ql/lib/codeql/swift/elements/expr/ApplyExpr.qll [hunk 1]
```diff
@@ -1,9 +1,25 @@
...
-  AbstractFunctionDecl getStaticTarget() { result = this.getFunction().(DeclRefExpr).getDecl() }
+  AbstractFunctionDecl getStaticTarget() {
+    exists(Expr f |
+      f = this.getFunction() and
+      (
+        result = f.(DeclRefExpr).getDecl()
+        or
+        result = f.(MethodRefExpr).getMethod()
+        or
+        result = f.(ConstructorRefCallExpr).getFunction().(DeclRefExpr).getDecl()
```
MathiasVP, line 16: I imagine we can remove this case once `MethodRefExpr` includes `ConstructorRefCallExpr`.

swift/ql/lib/codeql/swift/elements/expr/ApplyExpr.qll
```diff
@@ -12,3 +28,11 @@ class ApplyExpr extends ApplyExprBase {
...
   }
 }
+
+class MethodApplyExpr extends ApplyExpr {
+  MethodApplyExpr() { this.getFunction() instanceof MethodRefExpr }
+
+  AbstractFunctionDecl getMethod() { result = this.getFunction().(MethodRefExpr).getMethod() }
+
+  override Expr getQualifier() { result = this.getFunction().(MethodRefExpr).getBase() }
+}
```
MathiasVP, line 38: No need to repeat the `this.getFunction()(MethodRef)` part over and over:

```suggestion
class MethodApplyExpr extends ApplyExpr {
  MethodRefExpr method;
  
  MethodApplyExpr() { method = this.getFunction() }

  AbstractFunctionDecl getMethod() { result = method.getMethod() }

  override Expr getQualifier() { result = method.getBase() }
}
```

Now a `MethodApplyExpr` has both a `getFunction` and a `getMethod`, and they return different things, right? That seems a bit weird. Is this something we could improve on?

MathiasVP (commented):
This looks great! A couple of minor comments/suggestions, but nothing blocking this PR.

swift/ql/lib/codeql/swift/elements/expr/ApplyExpr.qll
redsun82: maybe renaming `getMethod` to `getMethodDeclaration` would clear up things a bit?

swift/ql/lib/codeql/swift/elements/expr/ApplyExpr.qll
redsun82: it depends how we want to do that. If we want constructors to be in the same bag as methods, then yes, though it will mean having to make `MethodRefExpr` a superclass of two distinct concrete IPA classes (one synthesized from `DotSyntaxCallExpr` and the other from `ConstructorRefCallExpr`). The other approach would be to make `ConstructorRefExpr` a sibling of `MethodRefExpr` under `LookupExpr`, then we would edit this line rather than suppress it

swift/ql/lib/codeql/swift/elements/expr/ApplyExpr.qll
redsun82: Thanks, that look way better!

Maybe renaming `getMethod` to `getMethodDeclaration` would clear up things a bit?

redsun82: > Hm, actually. Since `getStaticTarget` has been specialized for `MethodRefExpr`, do we even need `getMethod` here (except to implement `getStaticTarget`, of course)? Won't `getStaticTarget()` give back the same `AbstractFunctionDecl`?

you are absolutely right!

MathiasVP: I'd really prefer just killing `getMethod`/`getMethodDeclaration` 😂. Is there a reason you want to keep it around?

MathiasVP: Awesome 😍!

swift/ql/test/extractor-tests/generated/expr/DotSyntaxCallExpr/dot_syntax_call.swift
```diff
@@ -1,16 +1,9 @@
...
-  init() {
-    let f = baz
-  }
 }
 
 X.foo(1, 2)
 X.bar()
-X().baz(42)
 
 let f = X.bar
```
MathiasVP, line 9: Did you mean to delete these tests?

swift/ql/lib/codeql/swift/elements/expr/ApplyExpr.qll
MathiasVP: Good point. In this case, I think we should follow whatever precedence Java/C# has set. Do you know if their version of `MethodRef` also includes constructors (assuming you can actually do this in those languages 🤔)?

swift/ql/test/extractor-tests/generated/expr/DotSyntaxCallExpr/dot_syntax_call.swift
MathiasVP: Yeah, I think we should just leave it in for completeness.

MathiasVP (approved):
LGTM!

PR #10114, opened 2022-08-19
Ruby/Python: Shared HTTP client request concept

RasmusWL:
[Not too long ago](https://github.com/github/codeql/pull/9463) we ported the Python version of `request-without-cert-validation` to use the `HTTP::Client::Request` concept. In doing that, we used a slightly different setup, so you could see where the value comes from that causes the validation to be disabled. The reason is alerts on code such as the snippet below, which causes me to ask, so WHERE does the dangerous `validation` argument then come from? :thinking: 

```py
def make_req(url, validation):
    requests.get(url, verify=validation)
```

I think it's a much nicer experience to see exactly why our analysis concluded the argument could be false, instead of leaving that detective work for the end-user.

@nickrolfe I've requested a review from you now, since you [seem](https://github.com/github/codeql/commit/1ce458fa333296ea94c689e65e207e4aa5ae96c8) to have written the original Ruby version for this. (so it's only in draft, since we might have some discussion before it makes sense for the JS people to look at it). -- you can review it commit-by-commit.

I hated every minute of doing the library modeling rewrite in 0ac36243420e1accda8d867e94b23224ee3cb74f, so I added a helper predicate in 9eda63096510786349486d0ecdebcb66a4f430ca. That's probably the most controversial part of this PR. If there is already such a helper method available, I'm happy to rewrite my PR to use it instead. I'm also very open to naming it differently, but being able to replace that copied-logic with a helper method was just so sweet :sparkles: 

Would also be interested to hear your thoughts on 97905949846707c88a15a9a33db04a1556119324 -- (see commit msg)

---

### Review from JS

Note: This is mostly a Ruby PR. We do need some kind of approval from `@github/codeql-javascript` on the concept, before merging.


ruby/ql/lib/change-notes/2022-08-18-change-http-client-request.md
```diff
@@ -0,0 +1,4 @@
+---
+category: breaking
+---
+* Changed the `HTTP::Client::Request` concept from using `MethodCall` as base class, to using `DataFlow::Node` as base class. Any class that extend `HTTP::Client::Request::Range` must be changed, but if you only uses the member predicates of `HTTP::Client::Request` no changes are required.
```
nickrolfe, line 4:
```suggestion
* Changed the `HTTP::Client::Request` concept from using `MethodCall` as base class, to using `DataFlow::Node` as base class. Any class that extends `HTTP::Client::Request::Range` must be changed, but if you only use the member predicates of `HTTP::Client::Request`, no changes are required.
```

ruby/ql/lib/codeql/ruby/Concepts.qll
```diff
@@ -532,20 +529,13 @@ module HTTP {
...
-         * Gets a data-flow node that contributes to the URL of the request.
-         * Depending on the framework, a request may have multiple nodes which contribute to the URL.
-         */
-        abstract DataFlow::Node getAUrlPart();
-
-        /** Gets a string that identifies the framework used for this request. */
-        abstract string getFramework();
-
-        /**
+         * DEPRECATED: overwrite `disablesCertificateValidation/2` instead.
```
nickrolfe, line 532:
```suggestion
         * DEPRECATED: override `disablesCertificateValidation/2` instead.
```

ruby/ql/lib/codeql/ruby/dataflow/internal/DataFlowPublic.qll
```diff
@@ -87,6 +87,46 @@ class CallNode extends LocalSourceNode, ExprNode {
...
+      pair =
+        this.getArgument(_)
+            .getALocalSource()
+            .asExpr()
+            .(CfgNodes::ExprNodes::HashLiteralCfgNode)
+            .getAKeyValuePair() and
+      exprNode(pair.getKey())
+          .getALocalSource()
+          .asExpr()
+          .getExpr()
```
nickrolfe, line 124: You can remove this call to `getExpr()`, since `getConstantValue()` is also defined on `ExprCfgNode`, so there's no need to drop to the AST layer.

(I know you were just copying/moving existing code, but we may as well simplify while we're here).

ruby/ql/lib/codeql/ruby/frameworks/http_clients/Excon.qll
```diff
@@ -56,96 +55,77 @@ class ExconHttpRequest extends HTTP::Client::Request::Range {
...
+    // before the request, and no `ssl_verify_peer: true` in the explicit options hash
+    // for the request call.
+    exists(DataFlow::CallNode disableCall, BooleanLiteral value |
+      // Excon.defaults[:ssl_verify_peer]
+      disableCall = API::getTopLevelMember("Excon").getReturn("defaults").getAMethodCall("[]=") and
+      disableCall
+          .getArgument(0)
+          .getALocalSource()
+          .asExpr()
+          .getExpr()
```
nickrolfe, line 90: Same here.

ruby/ql/lib/codeql/ruby/frameworks/http_clients/Faraday.qll
```diff
@@ -38,125 +38,65 @@ class FaradayHttpRequest extends HTTP::Client::Request::Range {
...
-      // Faraday.new(..., { ssl: {...} })
-      exists(DataFlow::LocalSourceNode optionsNode, CfgNodes::ExprNodes::PairCfgNode p |
-        p = optionsNode.asExpr().(CfgNodes::ExprNodes::HashLiteralCfgNode).getAKeyValuePair() and
-        isSslOptionsPairDisablingValidation(p) and
-        optionsNode.flowsTo(arg) and
-        disablingNode.asExpr() = p
+      exists(CfgNodes::ExprNodes::PairCfgNode p, DataFlow::Node key |
+        p = sslValue.asExpr().(CfgNodes::ExprNodes::HashLiteralCfgNode).getAKeyValuePair() and
+        key.asExpr() = p.getKey() and
+        key.getALocalSource().asExpr().getExpr().getConstantValue().isStringlikeValue(argName) and
```
nickrolfe, line 68: Same here.

ruby/ql/lib/codeql/ruby/frameworks/http_clients/OpenURI.qll
```diff
@@ -56,72 +61,64 @@ class OpenUriRequest extends HTTP::Client::Request::Range {
...
+    exists(
+      DataFlow::LocalSourceNode optionsNode, CfgNodes::ExprNodes::PairCfgNode p, DataFlow::Node key
+    |
+      // can't flow to argument 0, since that's the URL
+      optionsNode.flowsTo(this.getArgument(any(int i | i > 0))) and
+      p = optionsNode.asExpr().(CfgNodes::ExprNodes::HashLiteralCfgNode).getAKeyValuePair() and
+      key.asExpr() = p.getKey() and
+      key.getALocalSource()
+          .asExpr()
+          .getExpr()
```
nickrolfe, line 91: Same here.

ruby/ql/src/queries/security/cwe-295/RequestWithoutValidation.ql
```diff
@@ -15,6 +15,20 @@ import ruby
...
+  // be the `disablingNode`, and the `origin` would be the place were `arg` got its
+  // value from.
+  //
+  // NOTE: We compare the locations instead of DataFlow::Nodes directly, since for
+  // snippet `Excon.defaults[:ssl_verify_peer] = false`, `disablingNode = argumentNode`
+  // does NOT hold.
+  if disablingNode.getLocation() = origin.getLocation()
+  then ending = "."
+  else ending = " by the value from $@."
+select request, "This request may run without certificate validation because it is $@" + ending,
```
nickrolfe, line 33: I think this will result in an invalid results pattern error if we ever hit the `else` case, since there will be two `$@` components in the message string.

ruby/ql/lib/codeql/ruby/security/InsecureDownloadCustomizations.qll
```diff
@@ -143,7 +143,7 @@ module InsecureDownload {
       hasUnsafeExtension(req.getAUrlPart().asExpr().getConstantValue().getString())
     }
 
-    override DataFlow::Node getDownloadCall() { result.asExpr().getExpr() = req }
+    override DataFlow::Node getDownloadCall() { result = req }
```
nickrolfe, line 146: If you're worried, we could also override the return type to be `DataFlow::CallNode`?

ruby/ql/lib/codeql/ruby/security/InsecureDownloadCustomizations.qll
```diff
@@ -193,6 +193,6 @@ module InsecureDownload {
 
     override DataFlow::FlowState getALabel() { result instanceof Label::Insecure }
 
-    override DataFlow::Node getDownloadCall() { result.asExpr().getExpr() = request }
+    override DataFlow::Node getDownloadCall() { result = request }
```
nickrolfe, line 196: Same here.

nickrolfe (commented):
I went through commit-by-commit. Looks like it was a lot of work, but it's a nice improvement, especially the helper method for keyword arguments!

ruby/ql/lib/codeql/ruby/security/InsecureDownloadCustomizations.qll
RasmusWL: No I was worried the other way around, such that someone makes a subclass of `HTTP::Client::Request::Range` they may not be a `DataFlow::CallNode`... but looking closer at this, it seems like it's just used in the [select of the query](https://github.com/github/codeql/blob/61bf2154cdd6f85ba29df12e2035458d0adf53a6/ruby/ql/src/queries/security/cwe-829/InsecureDownload.ql#L22), so it doesn't really matter.

ruby/ql/src/queries/security/cwe-295/RequestWithoutValidation.ql
RasmusWL: It actually just works :sparkles: Using the Python version as example, and the [Python test-file](https://github.com/github/codeql/blob/9cb249fc2f7bc1043d50f7d9424026a3a68e4f2a/python/ql/test/query-tests/Security/CWE-295-RequestWithoutValidation/make_request.py).

# VS Code

image

# Code Scanning

image


image

nickrolfe: Oh cool, sorry for doubting you!

RasmusWL: > Oh cool, sorry for doubting you!

No worries. It's good to question whether things work before just blindly accepting suspicious stuff like this. I also remember questioning whether I could even do this (I had not seen a query do this before), but it just works :muscle:

nickrolfe (dismissed):
Nice. I'm happy with the Ruby changes 👍

RasmusWL:
Requested review directly from you @hvitved to check on the use of `DataFlowImplForLibraries` -- just wanted to make 100% sure my use is OK, since I could only find a use of the taint-tracking build on top, not a direct use.

yoff (dismissed):
Python 👍

erik-krogh:
> ### Review from JS
> Note: This is mostly a Ruby PR. We do need some kind of approval from `@github/codeql-javascript` on the concept, before merging.

This new concept seems very similar to [ClientRequest from JS](https://github.com/github/codeql/blob/c6f89aac0af506f5182105113cc378f92541ba3f/javascript/ql/lib/semmle/javascript/frameworks/ClientRequests.qll).   
I don't like having a new empty concept. If someone starts using the new concept they'll be confused why they get no results for JS.   
Perhaps we should do a `class ClientRequstAsHttpRequest extends Http::Request::Range instanceof ClientRequest`?    
Let me know if you want me to do that for you.  

As for the `disablesCertificateValidation()` predicate.  
We don't have that as part of the models in JS. We only have that in the `js/disabling-certificate-validation` query. 

@asgerf thoughts?

RasmusWL:
> This new concept seems very similar to [ClientRequest from JS](https://github.com/github/codeql/blob/c6f89aac0af506f5182105113cc378f92541ba3f/javascript/ql/lib/semmle/javascript/frameworks/ClientRequests.qll). I don't like having a new empty concept. If someone starts using the new concept they'll be confused why they get no results for JS. Perhaps we should do a `class ClientRequstAsHttpRequest extends Http::Request::Range instanceof ClientRequest`? Let me know if you want me to do that for you.
>
> As for the `disablesCertificateValidation()` predicate.
> We don't have that as part of the models in JS. We only have that in the `js/disabling-certificate-validation` query.

For now, this only adds the concept to the internal file in `javascript/ql/lib/semmle/javascript/internal/ConceptsShared.qll`, which currently isn't used anywhere.

Long term, I think it would be nice to move modeling and queries in JS to use this concept instead. When that time comes, it would be sad if the concept we're trying to add now is in some way incompatible with what JS needs, such that we will have to go through changes for both Python and Ruby as well :grimacing:

So what I want from the JS review right now is a: Is this concept something you would accept was used in the modeling/queries in the future (potentially with more member-predicates, but not less).

erik-krogh (approved):
> So what I want from the JS review right now is a: Is this concept something you would accept was used in the modeling/queries in the future (potentially with more member-predicates, but not less).

Ahh, I see now.  

That I can give a 👍

We don't currently have `getFramework()` or `disablesCertificateValidation()` in JS, but the former seems reasonable, and the latter we already got inside a query.

ruby/ql/lib/codeql/ruby/frameworks/http_clients/Excon.qll
```diff
@@ -7,6 +7,7 @@ private import codeql.ruby.CFG
 private import codeql.ruby.Concepts
 private import codeql.ruby.ApiGraphs
 private import codeql.ruby.DataFlow
+private import codeql.ruby.dataflow.internal.DataFlowImplForLibraries as DataFlowImplForLibraries
```
hvitved, line 10: A `DataFlowImpl` copy should only be used multiple times in library code, when we can guarantee that either (a) those copies will _never_ be evaluated together (unlikely), or (b) that they are _always_ evaluated together. Both can be difficult to guarantee, so it is often easier to simply add more copies of `DataFlowImpl` (e.g. `DataFlowImplForLibraries2`, `DataFlowImplForLibraries3`, etc.

ruby/ql/lib/codeql/ruby/dataflow/internal/DataFlowPublic.qll
```diff
@@ -87,6 +87,41 @@ class CallNode extends LocalSourceNode, ExprNode {
...
+    result = this.getKeywordArgument(name)
+    or
+    exists(CfgNodes::ExprNodes::PairCfgNode pair |
+      pair =
+        this.getArgument(_)
+            .getALocalSource()
+            .asExpr()
+            .(CfgNodes::ExprNodes::HashLiteralCfgNode)
+            .getAKeyValuePair() and
+      exprNode(pair.getKey()).getALocalSource().asExpr().getConstantValue().isStringlikeValue(name) and
```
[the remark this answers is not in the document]
RasmusWL: Done in 25d09cd

ruby/ql/lib/codeql/ruby/frameworks/http_clients/Excon.qll
RasmusWL: Done in 528ef0e as discussed offline

hvitved (approved):

RasmusWL:
The new performance evaluation no longer shows a slowdown on ruby/ruby. Overall less than 1% slowdown, which seems acceptable. So this PR seems good to go :rocket:

PR #10115, opened 2022-08-19
Java: Add support for androidx.fragment.app.Fragment

atorralba:
`android.app.Fragment` is [deprecated](https://developer.android.com/reference/android/app/Fragment). This PR adds support for the new version, [`androidx.fragment.app.Fragment`](https://developer.android.com/reference/androidx/fragment/app/Fragment.html), which should help with our Android analysis by detecting more fragments.


aschackmull (approved):

atorralba:
@aschackmull if you don't mind re-approving, please 🙏  I needed to fix some tests that were affected by the new stubs.

aschackmull (approved):

PR #10116, opened 2022-08-19
Java: Improve Static Initialization Vector query

atorralba:
No longer require a mode in the sink (`Cipher.init`). The fact that a bytearray flows through an `AlgorithmParameterSpec` into the sink seems enough to determine that an IV is being used, be it for encryption or decryption. Other subclasses of `AlgorithmParameterSpec` that don't work with IVs don't seem to receive bytearrays anyway.

This helps covering CVE-2020-5408, which was covered by the experimental version of this query but was lost when we transitioned from global taint tracking to local data flow to keep track of the `Cipher` mode.


atorralba:
### FP evaluation 

After 3c6c09b0f89366ee95f1e669905339ed4f4ac416, 50 new projects with results popped up in our analysis.

From evaluating those results, new `ArrayUpdate` expressions (which act as sanitizers) were added in c35fbf9.

That reduced the new results to 29. From those, 8 are FPs and 21 are TPs.

The FPs are due to:
* Using custom dependencies to update an array, which means adding those wouldn't add value to our query other than in that specific project.
* `AsyncTask` spurious flow that comes from https://github.com/github/codeql/pull/9975, and should be addressed by the eventual solution of synthesized callables.
* Classic virtual dispatch spurious flow.

None were actually caused by removing the mode requirement in `Cipher.init` (but rather because we now see more paths), which gives me confidence that the change is a net improvement.

Launching DCA now to see if this impacts performance.

atorralba:
DCA doesn't show any interesting impact on performance.

aschackmull (dismissed):
LGTM, thanks for the details. Only one minor comment - the change note probably belongs in `src` rather than `lib`.

aschackmull:
> the change note probably belongs in `src` rather than `lib`.

I know that the only change is in a `lib` file, but maybe we ought to move all of the query-specific `*Query.qll` files into the `src` pack? (As a separate PR of course).

atorralba:
> the change note probably belongs in src rather than lib.

I doubted about where to place it, but in the end I thought that, since the change was in `lib` it should be there. The change is only interesting to one query though, so it makes sense the change note is in `src`. Moved it there, thanks.

> maybe we ought to move all of the query-specific *Query.qll files into the src pack? 

Although that would make sense, I think we wanted them in `lib` to be re-usable in the tests.

aschackmull:
> Although that would make sense, I think we wanted them in `lib` to be re-usable in the tests.

Hmm, but the query tests that use `qlref`s reference the queries in `src` just fine. I know that we can't put the `*Query.qll` files in the same dir as the query they belong to due to the directory name, but they could just go in a separate subdirectory within `src`.

aschackmull (approved):

PR #10117, opened 2022-08-19
C++: Improve performance of `cpp/cleartext-storage-buffer`

MathiasVP:
This PR contains two unrelated performance fixes to prevent the `cpp/cleartext-storage-buffer` query from blowing up some strange codebases.

The first one is a general fix to global variable dataflow: We prune the relation so that we don't consider `use2(x)` to be a jump step from the global variable to the use of `x` in:
```c
int x;

void test1() {
  use1(x); // This is a jump step.
  x = 42;
  use2(x); // But this is no longer a jump step.
}
```
and similarly, for writes to global variables we only register the "final" write as a write to the global variable. So for instance, `x = 42` in:
```c
int x;

void test2() {
  x = 42; // No longer a jump step.
  x = 99; // This is still a jump step, though.
}
```

The second performance fix is to block flow through integral typed expressions in the query. We have used similar sanitizers in other queries to fix FPs, so this is a natural thing to do in this query as well, and it seems to dramatically improve the query's performance.

### DIFF
--- a/cpp/ql/lib/semmle/code/cpp/ir/dataflow/internal/DataFlowPrivate.qll
+++ b/cpp/ql/lib/semmle/code/cpp/ir/dataflow/internal/DataFlowPrivate.qll
@@ -239,6 +239,71 @@ OutNode getAnOutNode(DataFlowCall call, ReturnKind kind) {
     )
 }
 
+private predicate writesToGlobal(StoreInstruction store, GlobalOrNamespaceVariable v) {
+  v = store.getResultAddress().(VariableAddressInstruction).getAstVariable()
+}
+
+private predicate loadsFromGlobal(LoadInstruction load, GlobalOrNamespaceVariable v) {
+  v = load.getSourceAddress().(VariableAddressInstruction).getAstVariable()
+}
+
+/** Holds if `i1` is always executed after `i2` */
+bindingset[i1, i2]
+private predicate instrStrictlyPostDominates(Instruction i1, Instruction i2) {
+  i1.getBlock().strictlyPostDominates(i2.getBlock())
+  or
+  exists(IRBlock block, int index1, int index2 |
+    block = i1.getBlock() and
+    block = i2.getBlock() and
+    block.getInstruction(index1) = i1 and
+    block.getInstruction(index2) = i2 and
+    index1 > index2
+  )
+}
+
+/** Holds if `i1` is always executed before `i2` */
+bindingset[i1, i2]
+private predicate instrStrictlyDominates(Instruction i1, Instruction i2) {
+  i1.getBlock().strictlyDominates(i2.getBlock())
+  or
+  exists(IRBlock block, int index1, int index2 |
+    block = i1.getBlock() and
+    block = i2.getBlock() and
+    block.getInstruction(index1) = i1 and
+    block.getInstruction(index2) = i2 and
+    index1 < index2
+  )
+}
+
+predicate jumpInto(GlobalOrNamespaceVariable v, Node n1, Node n2) {
+  exists(StoreInstruction store |
+    store = n1.asInstruction() and
+    writesToGlobal(store, v) and
+    // Only do a jump step if there is no write that is guarenteed to
+    // overwrite the global variable.
+    not exists(StoreInstruction anotherStore |
+      store != anotherStore and
+      writesToGlobal(anotherStore, v) and
+      instrStrictlyPostDominates(anotherStore, store)
+    )
+  ) and
+  v = n2.asVariable()
+}
+
+predicate jumpOutOf(GlobalOrNamespaceVariable v, Node n1, Node n2) {
+  exists(LoadInstruction load |
+    load = n2.asInstruction() and
+    loadsFromGlobal(load, v) and
+    // Only do a jump step if there is no dominating write that has overwritten
+    // the global variable.
+    not exists(StoreInstruction store |
+      writesToGlobal(store, v) and
+      instrStrictlyDominates(store, load)
+    )
+  ) and
+  v = n1.asVariable()
+}
+
 /**
  * Holds if data can flow from `node1` to `node2` in a way that loses the
  * calling context. For example, this would happen with flow through a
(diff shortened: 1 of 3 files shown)


cpp/ql/lib/semmle/code/cpp/ir/dataflow/internal/DataFlowPrivate.qll
```diff
@@ -239,28 +239,81 @@ OutNode getAnOutNode(DataFlowCall call, ReturnKind kind) {
...
+}
+
+/** Holds if `i1` is always executed after `i2` */
+bindingset[i1, i2]
+private predicate instrStrictlyPostDominates(Instruction i1, Instruction i2) {
+  i1.getBlock().strictlyPostDominates(i2.getBlock())
+  or
+  exists(IRBlock block, int index1, int index2 |
+    block = i1.getBlock() and
+    block = i2.getBlock() and
```
geoffw0, line 257: Are these two lines necessary?

cpp/ql/lib/semmle/code/cpp/ir/dataflow/internal/DataFlowPrivate.qll
```diff
@@ -239,28 +239,81 @@ OutNode getAnOutNode(DataFlowCall call, ReturnKind kind) {
...
+}
+
+predicate jumpInto(GlobalOrNamespaceVariable v, Node n1, Node n2) {
+  exists(StoreInstruction store |
+    store = n1.asInstruction() and
+    writesToGlobal(store, v) and
+    // Only do a jump step if there is no write that is guarenteed to
+    // overwrite the global variable.
+    not exists(StoreInstruction anotherStore |
+      store != anotherStore and
```
geoffw0, line 285: Doesn't the "strictly" in `instrStrictlyPostDominates` already give us this?

geoffw0 (commented):
I'm very glad to see some performance improvements for this query.  Two concerns:

(1) is this stuff tested somewhere?

(2) I'd like to know what happens in the case where there is a read between the two writes, or a write between the two reads?  For example:

```
int x = 0; // our global

void test()
{
  x = 10;
  printf("x = %i\n", x); // does x = 10 flow to here?
  x = 20;
}
```

In that case `x` could have been a local and maybe the "jump step" isn't needed anyway; but we can obscure the read, for example:

```
int x = 0; // our global

void useX()
{
  printf("x = %i\n", x); // does x = 10 flow to here?
}

void test()
{
  x = 10;
  useX();
  x = 20;
}
```

cpp/ql/lib/semmle/code/cpp/ir/dataflow/internal/DataFlowPrivate.qll
MathiasVP: Oh yes, you're totally right 👍. Will fix!

cpp/ql/lib/semmle/code/cpp/ir/dataflow/internal/DataFlowPrivate.qll
MathiasVP: You're right. They're implied by the next to lines. Good catch!

cpp/ql/lib/semmle/code/cpp/ir/dataflow/internal/DataFlowPrivate.qll
MathiasVP: Fixed in 7d1bfba1bef7933450afdec7088a21386b6fe729.

MathiasVP:
> (1) is this stuff tested somewhere?

In terms of test coverage: We already had a test with an incorrect annotation (it was actually a FP) that's fixed by the `jumpStep` change. See e4debf3edf67f5491c13f74e6745e868834edcf4 for that change to our test. I can definitely add more tests that specifically exercise this PR.

In terms of DCA: I did a DCA run over the weekend (it should be backlinked from this PR), but it looks like that run failed. I'll start a new run.

> (2) I'd like to know what happens in the case where there is a read between the two writes, or a write between the two reads? For example:
>
> ```c
> int x = 0; // our global
> 
> void test()
> {
>   x = 10;
>   printf("x = %i\n", x); // does x = 10 flow to here?
>   x = 20;
> }
> ```
> In that case x could have been a local and maybe the "jump step" isn't needed anyway; but we can obscure the read, for example:
> 
> ```c
> int x = 0; // our global
> 
> void useX()
> {
>   printf("x = %i\n", x); // does x = 10 flow to here?
> }
> 
> void test()
> {
>   x = 10;
>   useX();
>   x = 20;
> }
> ```

For the first example, we'll have:
- a normal SSA step from `x = 10` to `printf("x = %i\n", x);`.
- A jump step from `x = 20` to the global variable.
In particular, we don't have any of the following:
- A jump step from `x = 10` to `printf("x = %i\n", x);`
- A jump step from the global variable to `printf("x = %i\n", x);`

So to answer the first question in the comment: yes, `x = 10` should flow there by a local SSA step.

The second example is more sinister. I _think_ we're now losing the flow from `x = 10` to `printf("x = %i\n", x);`. I'll confirm this with a test, and think of a fix for this.

MathiasVP:
cb56c1e262e31a47e020c02f85da233cef0c5ab9 adds your tests @geoffw0. As you correctly pointed out, the last of your examples demonstrates a problem with the pruning. I'll see if I can come up with a fix for this.

aschackmull:
You should also add the dual test, i.e. the case when a write is reached by some call-chain in-between a local write-read sequence.
```
x = safe;
callThatMayUpdateX();
use(x); // <- ought to be a jump-step target to account for callThatMayUpdateX
```
Note that the implementations in Java and C# accounts for these sorts of updates in their field-based SSA by depending on virtual dispatch and making any call that may transitively reach a write into a may-write ssa definition.

MathiasVP:
> You should also add the dual test, i.e. the case when a write is reached by some call-chain in-between a local write-read sequence.
> 
> ```
> x = safe;
> callThatMayUpdateX();
> use(x); // <- ought to be a jump-step target to account for callThatMayUpdateX
> ```
> 
> Note that the implementations in Java and C# accounts for these sorts of updates in their field-based SSA by depending on virtual dispatch and making any call that may transitively reach a write into a may-write ssa definition.

Hmm, yes. Good point. Thanks!

Since this turns out to be kinda difficult to do on an ad-hoc basis I may delay this change until we've merged another upcoming big change to C++ SSA.

cpp/ql/lib/semmle/code/cpp/ir/dataflow/internal/DataFlowPrivate.qll
geoffw0: The same change can be applied to `instrStrictlyPostDominates` as well.

geoffw0:
Thanks for adding the tests (and explaining why my first example is not an issue).

PR #6559, opened 2021-08-26
JS: Support a taint tracking for arguments of `.apply()` function call

yuske:
Implemented DataFlow::Node's for arguments of `.apply` function calls and propagation of tainted values from the `argsArray` parameter to these arguments. Fixed small issues in the array initialization and improved unit tests, see [arrays-init.js](javascript/ql/test/library-tests/Arrays/arrays-init.js) and [call-apply.js](javascript/ql/test/library-tests/InterProceduralFlow/call-apply.js). 

@erik-krogh, take a look at this PR. We discussed this one in Slack.

PR #10126, opened 2022-08-22
JS: precise flow through calls to `.apply()`

erik-krogh:
For some context see this external contribution that never got merged: https://github.com/github/codeql/pull/6559 

The [improve performance of global dataflow by inlining a step predicate](https://github.com/github/codeql/commit/354eca001fa2715e62b0fb8c09e3273ff669d5c1) commit got the predicate from 4m8s to 1m36s on `bwip-js`.  

[An evaluation looks good](https://github.com/github/codeql-dca-main/tree/data/erik-krogh/myApply__nightly-old__code-scanning/reports). There are a few new results, and the performance looks good.  
(The new `js/unsafe-jquery-plugin` results are from the new `arguments` node, the others are from the improved flow through arrays).  

In an earlier evaluation I did a comparison where I disabled `ReflectiveParametersNode`: https://github.com/github/codeql-dca-main/tree/data/erik-krogh/myApply-no-parms-node-nightly-codescanning/reports 
The query-suite is bigger, so it finds more results.   
If you look at the `alert-meta-comparison` in the two evaluations you can see that most of the new flow doesn't have anything to do with `ReflectiveParametersNode`.   
(Library inputs worked slightly different back when I did that evaluation, so those numbers are off).  

/cc @yuske This is inspired by your attempt to do the same thing: https://github.com/github/codeql/pull/6559

### DIFF
--- a/javascript/ql/lib/semmle/javascript/dataflow/DataFlow.qll
+++ b/javascript/ql/lib/semmle/javascript/dataflow/DataFlow.qll
@@ -1030,6 +1030,32 @@ module DataFlow {
   }
 
   /**
+   * A data flow node representing the arguments object given to a function.
+   */
+  class ReflectiveParametersNode extends DataFlow::Node, TReflectiveParametersNode {
+    Function function;
+
+    ReflectiveParametersNode() { this = TReflectiveParametersNode(function) }
+
+    override string toString() { result = "the arguments object of " + function.describe() }
+
+    override predicate hasLocationInfo(
+      string filepath, int startline, int startcolumn, int endline, int endcolumn
+    ) {
+      function.getLocation().hasLocationInfo(filepath, startline, startcolumn, endline, endcolumn)
+    }
+
+    override BasicBlock getBasicBlock() { result = function.getEntry().getBasicBlock() }
+
+    /**
+     * Gets the function corresponding to this reflektive parameters node.
+     */
+    Function getFunction() { result = function }
+
+    override File getFile() { result = function.getFile() }
+  }
+
+  /**
    * A data flow node representing the exceptions thrown by the callee of an invocation.
    */
   class ExceptionalInvocationReturnNode extends DataFlow::Node, TExceptionalInvocationReturnNode {
@@ -1627,6 +1653,26 @@ module DataFlow {
     exists(Function f | not f.isAsyncOrGenerator() |
       DataFlow::functionReturnNode(succ, f) and pred = valueNode(f.getAReturnedExpr())
     )
+    or
+    // from a reflective params node to a reference to the arguments object.
+    exists(DataFlow::ReflectiveParametersNode params, Function f | f = params.getFunction() |
+      succ = f.getArgumentsVariable().getAnAccess().flow() and
+      pred = params
+    )
+  }
+
+  /**
+   * A step from a reflective parameter node to each parameter.
+   */
+  private class ReflectiveParamsStep extends PreCallGraphStep {
+    override predicate loadStep(DataFlow::Node obj, DataFlow::Node element, string prop) {
+      exists(DataFlow::ReflectiveParametersNode params, DataFlow::FunctionNode f, int i |
+        f.getFunction() = params.getFunction() and
+        obj = params and
+        prop = i + "" and
+        element = f.getParameter(i)
+      )
+    }
   }
 
   /**
--- a/javascript/ql/lib/semmle/javascript/dataflow/DataFlow.qll
+++ b/javascript/ql/lib/semmle/javascript/dataflow/DataFlow.qll
@@ -1037,7 +1037,7 @@ module DataFlow {
 
     ReflectiveParametersNode() { this = TReflectiveParametersNode(function) }
 
-    override string toString() { result = "the arguments object of " + function.describe() }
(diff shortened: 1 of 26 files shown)


javascript/ql/lib/semmle/javascript/dataflow/DataFlow.qll
```diff
@@ -1029,6 +1029,32 @@ module DataFlow {
...
+    override predicate hasLocationInfo(
+      string filepath, int startline, int startcolumn, int endline, int endcolumn
+    ) {
+      function.getLocation().hasLocationInfo(filepath, startline, startcolumn, endline, endcolumn)
+    }
+
+    override BasicBlock getBasicBlock() { result = function.getEntry().getBasicBlock() }
+
+    /**
+     * Gets the function corresponding to this reflektive parameters node.
```
asgerf, line 1051:
```suggestion
     * Gets the function whose `arguments` object is represented by this node.
```

javascript/ql/lib/semmle/javascript/security/dataflow/PrototypePollutingAssignmentCustomizations.qll
```diff
@@ -68,7 +68,7 @@ module PrototypePollutingAssignment {
   /**
    * A parameter of an exported function, seen as a source prototype-polluting assignment.
    */
-  class ExternalInputSource extends Source, DataFlow::SourceNode {
```
asgerf, line 71: Wouldn't it make more sense to treat `ReflectiveParmaetersNode` as a `SourceNode`?

asgerf (commented):
Looks great! I really like the results of those evaluations! 👍 

It seems we lost three `tainted-nodes` (open the raw document to see them). They all seem related to flow through `forEach`. Could you check if this is a regression or if this was spurious flow?

erik-krogh:
> It seems we lost three `tainted-nodes` (open the raw document to see them). They all seem related to flow through `forEach`. Could you check if this is a regression or if this was spurious flow?

The cause of those lost nodes is the change of `ArrayCreationStep` to be a `PreCallGraphStep`, which causes us to track [this array precisely](https://github.com/ywzhaiqi/userscript/blob/c399eee598f36b22bddffccf842118fbc32f413d/scripts/Kindle%E4%BA%BA%E5%8A%A9%E6%89%8B.user.js#L209). 
I'm not sure how that causes is to loose the tainted nodes.   

But we still have taint on those lines.  
E.g. [here](https://github.com/ywzhaiqi/userscript/blob/c399eee598f36b22bddffccf842118fbc32f413d/scripts/Kindle%E4%BA%BA%E5%8A%A9%E6%89%8B.user.js#L242) we still mark `url` and `url.match(/&tid=(\d+)/i)` as tainted, we just loose `url.match`, which is benign.    

So I think it's OK that we loose those tainted nodes.  

I'll start an evaluation on `default.yml` with meta queries to confirm that we don't loose more.

javascript/ql/test/library-tests/TaintTracking/call-apply.js
```diff
@@ -0,0 +1,85 @@
+function foo1(arg1, arg2) {
```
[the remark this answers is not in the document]
erik-krogh: It didn't change any results 👍

javascript/ql/lib/semmle/javascript/security/dataflow/PrototypePollutingAssignmentCustomizations.qll
erik-krogh: The `ReflectiveParametersNode` is a source node (but when I try to put that into the `extends` clause I get a non-monotonic recursion).  

But the source should still not be a `SourceNode`, because the `getALibraryInputParameter()` predicate does not return a `ReflectiveParametersNode`, that predicate returns either a parameter or a reference to the `arguments` variable.   
The location of `ReflectiveParametersNode` looks bad as a source (the location is the entire function), so I avoid using `ReflectiveParametersNode` directly as a source.

asgerf (approved):
👍

yuske:
Hi @erik-krogh, 
It is great that you finished this idea! Unfortunately, I was busy with a paper where we used CodeQL for JS static analysis and did not have enough time to do the performance measures of the two approaches that we discussed at #6559. Btw, the paper now is accepted at USENIX Security'23 and the early version is already available https://arxiv.org/pdf/2207.11171.pdf. We increased the recall metrics for Prototype Pollution queries from 33% and 25% to 88% - 97% (also thanks to this `.apply()` and `.call()` functions support, other changes in the JS stdlib and PP queries). We detected many new Prototype Pollution vulnerabilities that lead to RCE (not all are described in the paper yet, a part of them is under review).

erik-krogh:
> We increased the recall metrics for Prototype Pollution queries from 33% and 25% to 88% - 97% (also thanks to this `.apply()` and `.call()` functions support, other changes in the JS stdlib and PP queries). We detected many new Prototype Pollution vulnerabilities that lead to RCE (not all are described in the paper yet, a part of them is under review).

Great! You're very welcome to open pull requests with those changes.  
I recommend doing one small change at a time, otherwise it might be hard for us to evaluate.

PR #10128, opened 2022-08-22
C++: Add a sanitizer to `cpp/cleartext-storage-buffer`

MathiasVP:
This is the easy part of https://github.com/github/codeql/pull/10117:
> The second performance fix is to block flow through integral typed expressions in the query. We have used similar sanitizers in other queries to fix FPs, so this is a natural thing to do in this query as well, and it seems to dramatically improve the query's performance.

I pulled this out to a separate PR so it's not needlessly delayed by the unrelated `jumpStep` change.

### DIFF
--- a/cpp/ql/src/Security/CWE/CWE-311/CleartextBufferWrite.ql
+++ b/cpp/ql/src/Security/CWE/CWE-311/CleartextBufferWrite.ql
@@ -26,6 +26,10 @@ class ToBufferConfiguration extends TaintTracking::Configuration {
 
   override predicate isSource(DataFlow::Node source) { source instanceof FlowSource }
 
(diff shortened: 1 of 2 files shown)


MathiasVP:
DCA looks fine. Merging 🎉

geoffw0:
I wonder if the same improvement would make sense in the `cpp/cleartext-storage-database` and other `cpp/cleartext-storage-file` queries?

MathiasVP:
> I wonder if the same improvement would make sense in the `cpp/cleartext-storage-database` and other `cpp/cleartext-storage-file` queries?

I would imagine that to be the case, yes.

geoffw0:
I might look into it when I have a spare hour then.

PR #10130, opened 2022-08-22
Kotlin: Fix some more not-null-exprs in the extractor

igfoo:
The non-null-expressions were in `getFunctionalInterfaceType` and `getFunctionalInterfaceTypeWithTypeArgs`. Everything else is just indenting code inside `if null` conditionals, and reordering blocks of code so that those conditionals make sense.

### DIFF
--- a/java/kotlin-extractor/src/main/kotlin/KotlinFileExtractor.kt
+++ b/java/kotlin-extractor/src/main/kotlin/KotlinFileExtractor.kt
@@ -1468,14 +1468,22 @@ open class KotlinFileExtractor(
                     val extractionMethod = if (isFunctionInvoke) {
                         // For `kotlin.FunctionX` and `kotlin.reflect.KFunctionX` interfaces, we're making sure that we
                         // extract the call to the `invoke` method that does exist, `kotlin.jvm.functions.FunctionX::invoke`.
-                        val interfaceType = getFunctionalInterfaceTypeWithTypeArgs(drType.arguments).classOrNull!!.owner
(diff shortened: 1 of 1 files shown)


java/kotlin-extractor/src/main/kotlin/KotlinFileExtractor.kt
```diff
@@ -1484,10 +1492,15 @@ open class KotlinFileExtractor(
...
                     }
                 }
-                else
+                else {
                     useFunction<DbCallable>(callTarget)
+                }
 
-            tw.writeCallableBinding(id, methodId)
+            if (methodId == null) {
+                logger.warn("No method to bind call to for raw method access")
```
tamasvajk, line 1500: Minor: Can't we use `locId: Label<DbLocation>` as the location of this warning?

igfoo: Do you mean using `warnElement`? If so, we can't get any location information for a `Label`, as the compiler doesn't know what its location is.

We will get the stack of context from https://github.com/github/codeql/blob/main/java/kotlin-extractor/src/main/kotlin/utils/Logger.kt#L133-L137 though

tamasvajk: I was thinking that we could add an overload of `warn` that takes a `Label<DbLocation>`. But I see that `warnElement` also creates `val locationString = tw.getLocationString(element)`, which we can't do from a label.

tamasvajk (approved):

PR #10131, opened 2022-08-22
Java: Add new java.nio.Path{,s} summary models

atorralba:
Adds new summary models for the `java.nio.Path` and `java.nio.Paths` classes.


eldeloro75017:
> Adds new summary models for the `java.nio.Path` and `java.nio.Paths` classes.

aschackmull (approved):

java/ql/lib/change-notes/2022-08-22-path-summaries.md
```diff
@@ -0,0 +1,4 @@
+---
+category: minorAnalysis
+---
+* Added new flow steps for the classes `java.io.Path` and `java.nio.Paths`.
```
Marcono1234, line 4: It looks like the class names are incorrect here:
```suggestion
* Added new flow steps for the classes `java.nio.file.Path` and `java.nio.file.Paths`.
```
Should this be corrected in the release notes?

java/ql/lib/semmle/code/java/dataflow/ExternalFlow.qll
```diff
@@ -317,7 +317,12 @@ private predicate summaryModelCsv(string row) {
       "java.io;File;true;getAbsolutePath;;;Argument[-1];ReturnValue;taint;manual",
       "java.io;File;true;getCanonicalPath;;;Argument[-1];ReturnValue;taint;manual",
       "java.nio;ByteBuffer;false;array;();;Argument[-1];ReturnValue;taint;manual",
+      "java.nio.file;Path;true;normalize;;;Argument[-1];ReturnValue;taint;manual",
+      "java.nio.file;Path;true;resolve;;;Argument[-1..0];ReturnValue;taint;manual",
       "java.nio.file;Path;false;toFile;;;Argument[-1];ReturnValue;taint;manual",
+      "java.nio.file;Path;true;toString;;;Argument[-1];ReturnValue;taint;manual",
+      "java.nio.file;Path;true;toUri;;;Argument[-1];ReturnValue;taint;manual",
+      "java.nio.file;Paths;true;get;;;Argument[0..1];ReturnValue;taint;manual",
```
Marcono1234, line 325: Since Java 11 there are the static methods [`Path.of(String, String...)`](https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/nio/file/Path.html#of(java.lang.String,java.lang.String...)) and [`Path.of(URI)`](https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/nio/file/Path.html#of(java.net.URI)) which as per `Paths` documentation should be preferred over the static `Paths` methods.
Should these methods therefore be covered here as well?

Marcono1234 (commented):
Sorry for this late review; I hope it is useful nonetheless.

java/ql/lib/change-notes/2022-08-22-path-summaries.md
atorralba: Thanks @Marcono1234, that was indeed a mistake. I opened https://github.com/github/codeql/pull/10437 to address this.

java/ql/lib/semmle/code/java/dataflow/ExternalFlow.qll
atorralba: We should add those, yes. I'll do it in a subsequent PR. Thanks for pointing this out!

PR #10138, opened 2022-08-23
Java: Add summaries for ContentResolver and adjacent classes

atorralba:
Adds summaries for the following classes:
* [ContentResolver](https://developer.android.com/reference/android/content/ContentResolver)
* [ContentProviderClient](https://developer.android.com/reference/android/content/ContentProviderClient)
* [ContentProviderOperation](https://developer.android.com/reference/android/content/ContentProviderOperation)
* [ContentProviderOperation$Builder](https://developer.android.com/reference/android/content/ContentProviderOperation.Builder)
* [ContentProviderResult](https://developer.android.com/reference/android/content/ContentProviderResult)
* [Cursor](https://developer.android.com/reference/android/database/Cursor)

This is needed because, if an attacker controls the `Uri` that is used to interact with a Content Provider through a `ContentResolver`, the response can be arbitrarily set by the attacker.

This helps covering CVE-2020-6828.


aschackmull (approved):
Looks plausible.

PR #10141, opened 2022-08-23
Ruby: Perf fix for trackUseNode.

aschackmull:
Let's see if this does the trick. The problem was the following join-order:
```
Evaluated recursive predicate ApiGraphs#3116a2f3::API::Impl::trackUseNode#2#fff@ce9ac9do in 51821ms on iteration 43 (delta size: 4222306).
Evaluated relational algebra for predicate ApiGraphs#3116a2f3::API::Impl::trackUseNode#2#fff@ce9ac9do on iteration 43 running pipeline standard with tuple counts:
          3770177   ~0%    {3} r1 = SCAN ApiGraphs#3116a2f3::API::Impl::trackUseNode#2#fff#prev_delta OUTPUT In.2, In.0, In.1
        461567525  ~64%    {5} r2 = JOIN r1 WITH ApiGraphs#3116a2f3::API::Impl::trackUseNode#2#fff#join_rhs ON FIRST 1 OUTPUT Lhs.2, Rhs.3, Rhs.2, Rhs.1, Lhs.1
         10791924  ~66%    {5} r3 = JOIN r2 WITH TypeTracker#6f8aa30b::Cached::append#2#fff ON FIRST 3 OUTPUT Lhs.3, Lhs.1, Lhs.4, Lhs.1, Lhs.2
          4936050  ~21%    {5} r4 = r3 AND NOT ApiGraphs#3116a2f3::API::Impl::trackUseNode#2#fff#prev(Lhs.2, Lhs.4, Lhs.0)
          4936050  ~21%    {3} r5 = SCAN r4 OUTPUT In.2, In.4, In.0
                           return r5
```
which was clearly joining `append` too late.

### DIFF
--- a/ruby/ql/lib/codeql/ruby/ApiGraphs.qll
+++ b/ruby/ql/lib/codeql/ruby/ApiGraphs.qll
@@ -619,7 +619,7 @@ module API {
       exists(TypeBackTracker tb |
         result = mid.track(tmid, t) and
         pragma[only_bind_into](result) = useCandRev(pragma[only_bind_into](tb)) and
-        pragma[only_bind_out](t) = tb.getACompatibleTypeTracker()
+        pragma[only_bind_out](t) = pragma[only_bind_into](tb).getACompatibleTypeTracker()
       )
     }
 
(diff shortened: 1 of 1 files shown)


nickrolfe (approved):
I finally got some successful DCA runs. No change on our regular nightly suite, and a 38% reduction in analysis time on the latest `ruby/ruby`.

PR #10144, opened 2022-08-23
Swift: Extract `self` parameter declarations

MathiasVP:
Before this PR we were not extracting the `self` parameter of member function declarations because `Decl::getParameters` doesn't include the `self` parameter (see the comment [here](https://github.com/apple/swift/blob/main/include/swift/AST/Decl.h#L6719)).

(This is the extractor part of https://github.com/github/codeql-c-team/issues/903)


MathiasVP:
Note that there's a slight inconsistency test change accepted in 0257f09d9a0baac74e102f5f2a860a9f82f425b5. I don't think we should let this block this PR as I have another PR ready locally that's blocked on this one. I'll be happy to investigate the double-parent issue when my next PR has been merged.

swift/extractor/visitors/DeclVisitor.cpp
```diff
@@ -305,6 +305,8 @@ void DeclVisitor::fillAbstractFunctionDecl(const swift::AbstractFunctionDecl& de
   entry.name = !decl.hasName() ? "(unnamed function decl)" : constructName(decl.getName());
   entry.body = dispatcher_.fetchOptionalLabel(decl.getBody());
   entry.params = dispatcher_.fetchRepeatedLabels(*decl.getParameters());
+  auto self = const_cast<swift::ParamDecl* const>(decl.getImplicitSelfDecl());
```
AlexDenisov, line 308: Can we have a TODO here? Something like "TODO: figure out const correctness" would work 😄

AlexDenisov (approved):
Feel free to include the TODO in a follow-up PR

swift/extractor/visitors/DeclVisitor.cpp
MathiasVP: Indeed. I'll add such a comment in another PR 👍.

PR #10146, opened 2022-08-23
Remove upgrade database docs

aeisenberg:
Non-destructive database upgrades happens implicitly now as part of analysis v2.4.0, so I do not think we need to continue documenting user-driven upgrades. There is a caveat that very, very old databases (not sure exactly how old) may still need to have `database upgrade` called on it explicitly.

@jf205 is v2.4.0 old enough that we no longer need to document its behaviour?
@alexet can you let me know if there are any edge cases where db upgrades still need to be run manually?


jf205:
>is v2.4.0 old enough that we no longer need to document its behaviour?
@alexet can you let me know if there are any edge cases where db upgrades still need to be run manually?

[Version 2.4.0](https://github.com/github/codeql-cli-binaries/blob/main/CHANGELOG.md#release-240-2020-11-25) was released almost 2 years ago, so my gut feeling is that removing this article shouldn't be a problem. In the unlikely event that a user encounters a very old db, i suppose the CLI will return a helpful error message?

alexet:
A bunch of the tutorial database are so old that they need to run `database upgrade`. The compile will say to use upgrade in thsoe cases. The problem is that those really old databases have nothing that makes them look like a database.  

I know internally some people upgrade their databases so that they can measure performance on a clear cache without having to run the upgrades each time however this should be internal only.

aeisenberg:
> I know internally some people upgrade their databases so that they can measure performance on a clear cache without having to run the upgrades each time however this should be internal only.

I'm not concerned about internal users.

> A bunch of the tutorial databases are so old that they need to run `database upgrade`. The compiler will say to use upgrade in those cases.

Hmmm...yes. Maybe we can add a smaller section in the `analyzing-databases-with-the-codeql-cli.rst` page that mentions database upgrades for very old databases.?

aeisenberg:
Added a section on upgrading very old databases.

docs/codeql/codeql-cli/analyzing-databases-with-the-codeql-cli.rst
```diff
@@ -79,17 +80,21 @@ You can also specify:
...
 
-
-.. pull-quote::
+. pull-quote::
 
    Upgrading databases
 
-   If the CodeQL queries you want to use are newer than the
-   extractor used to create the database, then you may see a message telling you
+   For databases that were creaated by CodeQL CLI v2.3.3 or earlier, you will need
```
adityasharad, line 87:
```suggestion
   For databases that were created by CodeQL CLI v2.3.3 or earlier, you will need
```

docs/codeql/codeql-cli/analyzing-databases-with-the-codeql-cli.rst
```diff
@@ -79,17 +80,21 @@ You can also specify:
...
+   to explicitly upgrade the database before you can run an analysis with a newer
+   version of the CodeQL CLI. If this step is necessary, then you will see a message telling you
    that your database needs to be upgraded when you run ``database analyze``.
-   You can quickly upgrade a database by running the ``database upgrade``
-   command. For more information, see ":doc:`Upgrading CodeQL databases
-   <upgrading-codeql-databases>`."
+
+   CodeQL databases are upgraded by running the following command::
```
adityasharad, line 92: Add a sentence before this explaining that for databases created by newer versions of CodeQL, CodeQL will implicitly perform any necessary upgrades, and an explicit command is not required.

adityasharad (commented):
Thanks!

adityasharad (approved):

docs/codeql/codeql-cli/analyzing-databases-with-the-codeql-cli.rst
```diff
@@ -79,17 +80,24 @@ You can also specify:
...
+
+   For databases that were created by CodeQL CLI v2.3.4 or later, the CLI will implicitly run any
+   required upgrades. Explicitly running the upgrade command is not necessary.
+
+   CodeQL databases are upgraded by running the following command::
+
+       codeql database upgrade <database>
+
+   For full details of all the options you can use when upgrading databases,
+   see the "`database upgrade <../manual/database-upgrade>`__"  reference documentation.
```
jf205, line 100: I think we can probably drop these lines altogether as the first two sentences of this note tell users everything they need to know.

```suggestion
```

docs/codeql/codeql-cli/analyzing-databases-with-the-codeql-cli.rst
```diff
@@ -79,17 +80,18 @@ You can also specify:
 
 - .. include:: ../reusables/threads-query-execution.rst
 
-
-.. pull-quote::
+. pull-quote::
```
felicitymay, line 83: It looks as if we've accidentally lost a period here, resulting in bad formatting:

image

```suggestion
.. pull-quote::
```

felicitymay (commented):
Apart from one small typo introduced in this PR, it looks ready to merge 💖

docs/codeql/codeql-cli/analyzing-databases-with-the-codeql-cli.rst
aeisenberg: Thanks. Yes.

felicitymay (dismissed):

aeisenberg:
Looks like the previous approvals have been dismissed. Could someone re-approve?

adityasharad:
There are some changes about workspaces and lock files -- are those from a different PR?

aeisenberg:
Yes. They were from a downstream PR https://github.com/github/codeql/pull/10182 that got merged into this one.

adityasharad (approved):

PR #10148, opened 2022-08-23
Python: New call-graph based on type-trackers [still WIP]

RasmusWL:
While developing, things changed majorly 2-3 times, and it just wasn't feasible to go back and write a really nice commit history... so this is what you get instead. I've spent some time (maybe too much) making some of the history nice to review, but doing it for all of it just wasn't feasible.

There might be a few things failing now, but before I started doing rebases, I had a branch without test failures :sparkles:

note: the reason so many lines are added, is that I copied the old `dataflow/internal` folder to keep the points-to based version around for comparison while developing. As the very last thing on this branch, that copy should be removed again.


RasmusWL:
Closing in favor of https://github.com/github/codeql/pull/11376

PR #10154, opened 2022-08-24
Kotlin: Implement and use fun <T,reified S: T> Iterable<T>.findSubType

igfoo:
We had a lot of uses of find which looked for a subtype, but then had to cast (seemingly unsafely) the result to that subtype. This pulls all the casts out into a utility function.

### DIFF
--- a/java/kotlin-extractor/src/main/kotlin/utils/Iterable.kt
+++ b/java/kotlin-extractor/src/main/kotlin/utils/Iterable.kt
@@ -1,0 +1,13 @@
+package com.github.codeql
+
+/**
+ * This behaves the same as Iterable<T>.find, but requires
+ * that the value found is of the subtype S, and it casts
+ * the result for you appropriately.
+ */
+inline fun <T,reified S: T> Iterable<T>.findSubType(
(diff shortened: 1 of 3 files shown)


java/kotlin-extractor/src/main/kotlin/utils/Iterable.kt
```diff
@@ -0,0 +1,13 @@
...
+/**
+ * This behaves the same as Iterable<T>.find, but requires
+ * that the value found is of the subtype S, and it casts
+ * the result for you appropriately.
+ */
+inline fun <T,reified S: T> Iterable<T>.findSubType(
+    predicate: (S) -> Boolean
+): S? {
+    return this.find { it is S && predicate(it) } as S?
+}
```
tamasvajk, line 12: There are only `findSubType<IrDeclaration,...>()` calls to this function. We could remove `T` and replace with `IrDeclaration`.

igfoo: Done!

tamasvajk (approved):

PR #10156, opened 2022-08-24
Swift: make `ImportDecl:imported_module` optional

redsun82:
`ImportDecl::getModule` can return `nullptr` for `import` statements in inactive `#if` clauses.


AlexDenisov (approved):

redsun82:
@AlexDenisov this might also have been a cause for extractor crashes on DCA...

PR #10157, opened 2022-08-24
Swift: Add field flow

MathiasVP:
(Part of https://github.com/github/codeql-c-team/issues/903.)

This is the final PR needed for adding field flow for Swift. We replace the old `TInOutUpdateNode` with an IPA branch for post-update nodes, and include the various things that need post-update nodes (arguments to functions, and base expressions for field lookups and stores). Then we can pretty much steal Ruby's `readStep` and `storeStep` for fields 🎉.

The PR also implements `clearsContent` using the trivial implementation that depends on use-use flow.


AlexDenisov (approved):

MathiasVP:
:shipit:

PR #10158, opened 2022-08-24
C#: Narrow collection like types in model generation.

michaelnebel:
The purpose of this PR is to
- Align the relevant types part of the model generation, such that we no longer accept collection like types in read/store steps (except for bulk datatypes that can be used for string representation).

To see the impact of the change above, models were regenerated based on main and committed as a part of this PR, such that we can pinpoint the exact diff in model generation that this PR produces.

This revealed some interesting things
- Only very few summaries will no longer be produced for .NET Runtime when excluding collection like types in model generation.
- There appears to be an issue with the source model generation (the same has been observed for java). That is, source model generation based on "file" kind has been disabled for C#. This needs to be investigated (will add this as an item to the ongoing issue with model generation work)
- Some of the tuple related summaries that has previously disappeared has re-appeared in the generated models. Doing a case analysis is way to time-consuming, but this hints that we need much better tooling around the model generation to observe which changes impacts the generated models instead of having to figure it out later.

Will try to run DCA to check if the exclusion of collection like types narrows the search space in way that is measurable in terms of performance.


michaelnebel:
> The purpose of this PR is to
> 
> * Align the relevant types part of the model generation, such that we no longer accept collection like types in read/store steps (except for bulk datatypes that can be used for string representation).
> 
> To see the impact of the change above, models were regenerated based on main and committed as a part of this PR, such that we can pinpoint the exact diff in model generation that this PR produces.
> 
> This revealed some interesting things
> 
> * Only very few summaries will no longer be produced for .NET Runtime when excluding collection like types in model generation.
> * There appears to be an issue with the source model generation (the same has been observed for java). That is, source model generation based on "file" kind has been disabled for C#. This needs to be investigated (will add this as an item to the ongoing issue with model generation work)
> * Some of the tuple related summaries that has previously disappeared has re-appeared in the generated models. Doing a case analysis is way to time-consuming, but this hints that we need much better tooling around the model generation to observe which changes impacts the generated models instead of having to figure it out later.
> 
> Will try to run DCA to check if the exclusion of collection like types narrows the search space in way that is measurable in terms of performance.

DCA did not show any improvements or regressions (but shows that there is a bunch of source models that will no longer be produced (as expected)).
In any case, we should consider to merge this change to be aligned with the Java implementation.

PR #10033, opened 2022-08-12
C++: Missing return-value check for scanf-like functions #1076

d10c:
- Update .gitignore for .vscode/*.log temporaries
- C++: Add test and placeholder query.
- C++: First working. We now prefer flagging the cases where the variable was initialized, as in real world cases we haven't seen it done safely.
- Add more (false-negative) MissingCheckScanf tests
- Add more MissingCheckScanf test cases

PR #5930, opened 2021-05-20
C++: Promote `cpp/incorrect-allocation-error-handling`

MathiasVP:
This PR moves `cpp/incorrect-allocation-error-handling` out of experimental with `@precision medium`. This means it won't go into the Code Scanning suite yet, nor will it be displayed by default on LGTM. It will, however, be run on LGTM. This will give us some helpful data on whether we can safely move this to a higher precision.

<b>Note:</b> The documentation for this query was improved in https://github.com/github/codeql/pull/5804, but it has not received a docs review yet. I made some conscious whitespace changes in the first commit (95045929096f4b96eaedefb816ed42515b079583) to help @github/docs-content-codeql review it (I hope that works 🤞 ).

The results on LGTM look good: https://lgtm.com/query/600361054747246847/

PR #10163, opened 2022-08-24
C++: New Query: missing return-value check for scanf-like functions

d10c:
Introduces `cpp/missing-check-scanf`, a query with a similar idea as [this external contribution](https://github.com/github/codeql/pull/8246): find `scanf` output arguments that are read without checking that they were written in the first place. The main differences are an IR-based flow and a more conservative treatment of unguarded reads of already-initialized `scanf` output variables.

This PR squashes and supercedes [draft PR #10033](https://github.com/github/codeql/pull/10033).


cpp/ql/src/Critical/MissingCheckScanf.ql
```diff
@@ -0,0 +1,123 @@
...
+  /**
+   * Any subsequent use of this argument should be surrounded by a
+   * check ensuring that the `scanf`-like function has returned a value
+   * equal to at least `getMinimumGuardConstant()`.
+   */
+  int getMinimumGuardConstant() {
+    result =
+      varargIndex + 1 -
+        count(ScanfFormatLiteral f, int n |
+          n <= varargIndex and f.getUse() = call and f.parseConvSpec(n, _, _, _, "n")
```
MathiasVP, line 48: Ha! I was not aware of this `%n` specifier. Good catch 😂. I think that deserves a comment, though.

cpp/ql/src/Critical/MissingCheckScanf.ql
```diff
@@ -0,0 +1,123 @@
...
+    valueNumber(i) = valNum and
+    exists(Expr e | i.getAst() = e |
+      i = any(StoreInstruction s).getDestinationAddress()
+      or
+      [e, e.getParent().(AddressOfExpr)] instanceof ScanfOutput
+    )
+  }
+}
+
+/** Returns a block guarded by the assertion of $value $op $call */
```
MathiasVP, line 100: We usually don't include $ in QLDoc to highlight the specific parameters.

```suggestion
/** Returns a block guarded by the assertion of `value op call`. */
```

cpp/ql/src/Critical/MissingCheckScanf.ql
```diff
@@ -0,0 +1,123 @@
...
+
+  /**
+   * Get a subsequent access of the same underlying storage,
+   * but before it gets reset or reused in another `scanf` call.
+   */
+  Access getAnAccess() {
+    exists(Instruction dst |
+      this.bigStep(instr) = dst and
+      dst.getAst() = result and
+      valueNumber(dst) = valNum
```
MathiasVP, line 74: Could we remove this conjunct into `smallStep` instead? i.e., require that every small step preserves the value number? I think that will reduce the size of the `smallStep` relation (and more importantly, the `bigStep` relation!) quite a bit.

cpp/ql/test/query-tests/Critical/MissingCheckScanf/.clang-format
```diff
@@ -0,0 +1 @@
+DisableFormat: true
```
MathiasVP, line 1: I think you committed this file by mistake?

cpp/ql/src/Critical/MissingCheckScanf.ql
```diff
@@ -0,0 +1,123 @@
...
+
+from ScanfOutput output, ScanfFunctionCall call, ScanfFunction fun, Access access
+where
+  call.getTarget() = fun and
+  output.getCall() = call and
+  output.hasGuardedAccess(access, false)
+select access,
+  "$@ is read here, but may not have been written. " +
+    "It should be guarded by a check that $@() returns at least " + output.getMinimumGuardConstant()
+    + ".", access, access.toString(), call, fun.getName()
```
MathiasVP, line 123: I'm not sure I like the `()` suffix since it implies that the call had no arguments. How about this instead?

```suggestion
    "It should be guarded by a check that $@ returns at least " + output.getMinimumGuardConstant()
    + ".", access, access.toString(), call, call.toString()
```

MathiasVP (commented):
Looks really good! Here are my first round of comments.

change-notes/2022-08-24-missing-check-scanf.md
```diff
@@ -0,0 +1,4 @@
+---
+category: newQuery
+---
+* Added a new query, `cpp/missing-check-scanf`, to detect `scanf` output variables that are used without a proper return-value check to see that they were actually written.
```
MathiasVP, line 4: We should add that this query was initially submitted as an experimental query. You can see the phrasing I used for this [here](https://github.com/github/codeql/pull/5930).

cpp/ql/src/Critical/MissingCheckScanf.ql
```diff
@@ -0,0 +1,123 @@
...
+/** An expression appearing as an output argument to a `scanf`-like call */
+class ScanfOutput extends Expr {
+  ScanfFunctionCall call;
+  int varargIndex;
+  Instruction instr;
+  ValueNumber valNum;
+
+  ScanfOutput() {
+    this = call.getArgument(call.getTarget().getNumberOfParameters() + varargIndex) and
+    varargIndex >= 0 and
```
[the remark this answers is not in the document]
d10c: Yeah, I was seriously considering moving part of this ScanfOutput class into a the Scanf.qll library. `getAnOutputArgument()` would be a useful function, as well as, potentially, `getMinimumGuardConstant()`, due to having special cases that you don't want duplicated elsewhere (assuming other queries are interested in this information).

cpp/ql/src/Critical/MissingCheckScanf.ql
```diff
@@ -0,0 +1,123 @@
...
+    instr.getUnconvertedResultExpression() = this and
+    valueNumber(instr) = valNum and
+    // The following line is a kludge to prohibit more than one associated `instr` field,
+    // as would occur, for example, when `this` is an access to an array variable.
+    not instr instanceof ConvertInstruction
```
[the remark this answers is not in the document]
d10c: Ok, I'll try this out. Note, I was specifically filtering out the ConvertInstruction because it happens _right after_ the array variable access instruction, and I didn't want the second one to be interpreted as a barrier. But maybe this won't be an issue.

cpp/ql/src/Critical/MissingCheckScanf.ql
d10c: True, a comment would be wise.

cpp/ql/src/Critical/MissingCheckScanf.ql
d10c: That would be tricky, because while we do want the final Accesses to have the same ValueNumber, we don't want all intermediate instructions to have to have the same ValueNumber too, because then some crucial paths will be severed. Currently, `bigStep` is the same the `.getASuccessor+()` closure starting from `instr`, minus all paths that go through a barrier as defined by `isBarrier()`. I would hesitantly wager that this is as compact as it gets.

d10c: Actually, on second thought, maybe `bigStep(i)` can be optimized by having `i = instr`, always :) Because none of the `i != instr` tuples are ever needed.

cpp/ql/test/query-tests/Critical/MissingCheckScanf/.clang-format
d10c: I didn't. But I guess I should keep this kind of file to myself? What do you think of .gitignore-ing it?

jketema: We could add it to `.gitignore`, but that requires some care, as there are places where we have added `.clang-format` files on purpose. Probably best to leave this out of this PR and discuss separately. In general, I don't think we want to litter the `cpp/ql/test` directory with `.clang-format` files. 

I'm not sure what triggered you to add the `.clang-format` file, but if it was some post-commit hook, it might also be an indication that the hook needs some fine tuning.

Personally, I would also leave the change you made to the `.gitignore` file out of this PR, as that affects people outside our team. However, if you think it's a non-controversial change, then feel free to leave it as is.

change-notes/2022-08-24-missing-check-scanf.md
jketema: The change note is in the wrong place by the way, which is why the CI is still complaining. Should be in `cpp/ql/src/change-notes`. I get why you added it where you added it. It's probably the most obvious place for it, and holds change notes for pretty old versions 😄

cpp/ql/src/Critical/MissingCheckScanf.ql
MathiasVP: > That would be tricky, because while we do want the final Accesses to have the same ValueNumber, we don't want all intermediate instructions to have to have the same ValueNumber too, because then some crucial paths will be severed. Currently, bigStep is the same the .getASuccessor+() closure starting from instr, minus all paths that go through a barrier as defined by isBarrier(). I would hesitantly wager that this is as compact as it gets.

Ah, right. Yeah, I do see what you mean here.

MathiasVP: > Actually, on second thought, maybe bigStep(i) can be optimized by having i = instr, always :) Because none of the i != instr tuples are ever needed.

Yes, I think that's a good idea.

Speaking of optimizations: Currently, you're big step is defined something like:

```ql
predicate bigStep(Instruction from, Instruction to) { source(from) and smallStep*(from, to) and sink(to) }
```

The `smallStep*(from, to)` computation produces a lot of intermediate tuples that you don't really care about. We have a lot of experience in how to do this efficiently in QL, and I'd be happy to chat tomorrow about how to compute this properly. Let me just dump it here for you to read at your own pace:

The trick is to define a unary predicate to avoid producing a lot of unnecessary tuples:
```ql
// Holds if `i` can be reached starting at a source.
predicate reachedFromSource(Instruction i) {
  source(i)
  or
  exists(Instruction mid | reachedFromSource(mid) | step(mid, i))
}

// Holds if `i` can be reached starting at a source, and can eventually reach a sink.
predicate reachesSink(Instruction i) {
  reachedFromSource(i) and
  (
    sink(i)
    or
    exists(Instruction mid | reachesSink(mid) | step(i, mid))
  )
}
```ql
Then, you can use this predicate to define a smaller step relation:
```
predicate prunedSmallStep(Instruction from, Instruction to) {
  reachesSink(from) and reachesSink(to) and smallStep(from, to)
}
```
and finally, you can rewrite your `bigStep` as:
```ql
predicate bigStep(Instruction from, Instruction to) { source(from) and prunedSmallStep*(from, to) and sink(to) }
```

cpp/ql/src/Critical/MissingCheckScanf.ql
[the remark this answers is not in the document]
geoffw0: I agree it would be nice to have `ScanfFunctionCall.getAnOutputArgument()` (or `ScanfFunctionCall.getOutputArgument(int index)`, or both).  I'm not entirely clear what `getMinimumGuardConstant()` would be exactly, is this the number of vararg arguments that should have been written (and thus also the return value) if the `scanf` is entirely successful?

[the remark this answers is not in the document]
d10c: Yes, it's the smallest return value that indicates success in writing the given argument.

change-notes/2022-08-24-missing-check-scanf.md
d10c: Ah, ok, I was wondering about that. Because the check-change-note GitHub action seemed buggy to me. It does `gh api 'repos/github/codeql/pulls/10163/files' --paginate --jq 'any(.[].filename ; test("/change-notes/.*[.]md$"))'`, which returns false; but if only it did `test("change-notes/.*[.]md$")`, it would have returned true. But now I see it's a matter of moving the change note to a subdirectory.

cpp/ql/src/Critical/MissingCheckScanf.ql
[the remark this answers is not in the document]
d10c: Btw, is there a convention that "parameter" refers to the variable in the parameter list from the perspective of the body of a function, while "argument" refers to an expression passed at the call site of a function?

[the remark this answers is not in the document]
MathiasVP: Yep.

cpp/ql/src/Critical/MissingCheckScanf.ql
d10c: I have saved the best for last, implementing your proposed changes. Here are the before-and-after MRVA runs... Let's see how much of a difference it makes :) 
Before: https://github.com/github/semmle-code/actions/runs/2927651707
After: https://github.com/github/semmle-code/actions/runs/2927920561

d10c: Hmm... Nope, I'm calling it. It's not worth it. I've had to inline some predicates to avoid the runaway you see in the "After", and still it's not a significant difference. This is my WIP, so it could be improved in the future: https://github.com/github/codeql/commit/7cc0d7111e8544d98c45fe83ac6cccd18456c8bf 

P.S. MRVA top 100 run on that WIP code: https://github.com/github/semmle-code/actions/runs/2929372003

MathiasVP: The fact that the new code is _not_ timing out on opencv suggest that's it's a worthwhile improvement to make. I've left a comment on your branch. We can also discuss them tomorrow if you want :)

cpp/ql/src/Critical/MissingCheckScanf.qhelp
```diff
@@ -0,0 +1,51 @@
...
+
+<overview>
+<p>
+This query finds calls of <tt>scanf</tt>-like functions with missing or
+improper return-value checking.
+</p>
+<p>
+Specifically, the query flags uses of variables that may have been modified by
+<tt>scanf</tt> and subsequently are used without being guarded by a correct
+return-value check. A proper check is one that asserts the corresponding
```
geoffw0, line 15: Consider replacing "asserts" with another word ("ensures"?), as some developers may read "asserts" as a reference to the `assert` macro and miss the point of this sentence.

cpp/ql/src/Critical/MissingCheckScanf.ql
```diff
@@ -0,0 +1,119 @@
...
+
+  /**
+   * Returns the smallest possible `scanf` return value that would indicate
+   * success in writing this output argument.
+   */
+  int getMinimumGuardConstant() {
+    result =
+      varargIndex + 1 -
+        count(ScanfFormatLiteral f, int n |
+          // Special case: %n writes to an argument without reading any input.
```
geoffw0, line 43: This is great attention to detail.

geoffw0 (commented):
Query and `.qhelp` look great.  It looks like this query became a fair bit more complicated than we had expected!

cpp/ql/src/Critical/MissingCheckScanf.ql
```diff
@@ -0,0 +1,119 @@
+/**
+ * @name Missing return-value check for a 'scanf'-like function
+ * @description Without checking that a call to 'scanf' actually wrote to an
```
mchammer01, line 3: This is a little convoluted. Suggesting something into these lines:

`Failing to check that a call to 'scanf' actually writes to an output variable can lead to unexpected behavior at reading time`

What do you think? 🙂

cpp/ql/src/Critical/MissingCheckScanf.qhelp
```diff
@@ -0,0 +1,51 @@
...
+<example>
+<p>This example shows different ways of guarding a <tt>scanf</tt> output:
+</p>
+<sample src="MissingCheckScanf.cpp" />
+</example>
+
+<references>
+<li>SEI CERT C++ Coding Standard: <a href="https://wiki.sei.cmu.edu/confluence/display/cplusplus/ERR62-CPP.+Detect+errors+when+converting+a+string+to+a+number">ERR62-CPP. Detect errors when converting a string to a number</a>.</li>
+<li>SEI CERT C Coding Standard: <a href="https://wiki.sei.cmu.edu/confluence/display/c/ERR33-C.+Detect+and+handle+standard+library+errors">ERR33-C. Detect and handle standard library errors</a>.</li>
+<li>cppreference.com: <a href="https://en.cppreference.com/w/c/io/fscanf">scanf, fscanf, sscanf, scanf_s, fscanf_s, sscanf_s</a></li>
```
mchammer01, line 49: Missing full stop
```suggestion
<li>cppreference.com: <a href="https://en.cppreference.com/w/c/io/fscanf">scanf, fscanf, sscanf, scanf_s, fscanf_s, sscanf_s</a>.</li>
```

cpp/ql/src/Critical/MissingCheckScanf.qhelp
```diff
@@ -0,0 +1,51 @@
...
+<p>
+Functions in the <tt>scanf</tt> family return either EOF (a negative value)
+in case of IO failure, or the number of items successfully read from the
+input. Consequently, a simple check that the return value is truthy (nonzero)
+is not enough.
+</p>
+<warning>
+This query has medium precision because, in the current implementation, it
+takes a strict stance on unguarded uses of output variables, and flags them
+as problematic even if they had already been initialized.
```
mchammer01, line 27: Not sure about this one, but suggesting the following:
```suggestion
as problematic even if they have already been initialized.
```

mchammer01 (approved):
@d10c 👋🏻 - I've made a few comments and suggestions, but in generally this looks great to me ✨ 
It would be awesome if we could simplify the query description. Feel free to ignore any feedback you don't agree with 😃

cpp/ql/src/Critical/MissingCheckScanf.ql
d10c: Thanks! I think there are even more special cases, that I'm parking for a follow-up issue for now (namely, the *_s variants of scanf introduced in C11).

cpp/ql/src/Critical/MissingCheckScanf.ql
d10c: Yeah, adding the `pragma[inline]` to the cartesian-product-prone predicate removed the timeout on opencv 😛 Maybe a `bindingset[this]` annotation would have done the trick too..? And you're right, I missed the first line from `reachesSink`, which might make all the difference...

So along with removing the `this` argument from these predicates (which might be tricky in some cases, since `isBarrier` _does_ depend on `this`), I'll leave the exploration of this type of optimization for a follow-up issue, unless you think the performance improvements would be so drastic that they would be worth having in the initial release of this query (that is, if you think we can analyze opencv in significantly less than an hour on GitHub Actions [Locally, it's just a few minutes, fyi])

MathiasVP: > Yeah, adding the `pragma[inline]` to the cartesian-product-prone predicate removed the timeout on opencv 😛 Maybe a `bindingset[this]` annotation would have done the trick too..? And you're right, I missed the first line from `reachesSink`, which might make all the difference...

Yep, I think `bindingset` would do the same thing as a `bindingset` annotation implies inlining (as far as I know, at least). I also think the missing conjunct in `reachesSink` might make all the difference 😄, indeed.

I'm still not completely sure how to read the MRVA run output. I assume the 50 minutes of runtime on `opencv` is because we also have to compute the IR and all the other underlying things that the query makes use of. If it only takes a few minutes once the `cache`d libraries are computed that sounds fine.

MathiasVP:
I'm quite happy with this PR now. I don't see a DCA run anywhere, though. I think we should run one to be safe.

d10c:
Another thing, currently the query metadata reads:

```
@problem.severity recommendation
@security-severity 4.5
@precision medium
```

I wasn't clear on how to set these numbers, so I just winged the security-severity to "some number in the medium range", and the severity as "recommendation" because the MissingNullCheck query was also a recommendation. But maybe the security properties of this query are sufficiently different that these numbers merit more scrutiny. Maybe @geoffw0 can chime in?

geoffw0:
> `@problem.severity recommendation`

Severity is about how bad the problem is when you have a true positive.
 - `recommendation` is for bad coding style, hard to maintain code, something that's not quite ideal but there is no reason to believe the code actually behaves incorrectly as a result.  Often subjective.
 - `warning` is for something that's generally considered a bad idea and might cause problems.  Anything that doesn't fit the above or below categories.
 - `error` is for an issue that objectively causes incorrect program behaviour, crash, or vulnerability.  No sensible developer thinks doing this thing is OK.

I think I could justify `error` in this case but `recommendation` is fine.

> `@security-severity 4.5`

Discussed elsewhere.

> `@precision medium`

This is about what proportion of results can be expected to be true positives.  The exact meanings are somewhat controversial, I tend to think them as:
 - `medium`: in most projects at least 50% of results are plausibly true positives.
 - `high`: at least 75% (some people will probably give a higher number here) of results are plausibly true positives, and some effort has been made to confirm this.
 - `very high`: basically all (95%+) results are true positive and this has been confirmed extensively in real world code.

Its a bit tricky, because it isn't always easy to determine if a result in someone else's code is a false positive, and in high quality projects FP rates will be higher (as there are fewer TPs).

We usually target `high` precision but sometimes fall short and end up at `medium`.  Given the work you've done on false positives I suspect you can set this to `high`, or maybe upgrade it to `high` after fixing a few more false positives in a follow-up issue.

See https://github.com/github/codeql/blob/main/docs/query-metadata-style-guide.md for more info about this stuff.

cpp/ql/src/Critical/MissingCheckScanf.ql
```diff
@@ -3,11 +3,11 @@
...
  *              output variable can lead to unexpected behavior at reading time.
  * @kind problem
- * @problem.severity recommendation
- * @security-severity 4.5
+ * @problem.severity warning
+ * @security-severity 7.5
  * @precision medium
  * @id cpp/missing-check-scanf
- * @tags security
+ * @tags security cwe-252 cwe-253
```
MathiasVP, line 10: I _think_ each tag should be on a new line. That's what we've done in the past, and it looks like that's the format also used in our [style guide](https://github.com/github/codeql/blob/main/docs/query-metadata-style-guide.md#metadata-area).

geoffw0 (approved):
LGTM.

PR #10170, opened 2022-08-25
Java: Update models for commons-io and add negative models.

michaelnebel:
This this PR
- Update existing models for commons-io.
- Generate Negative models for commons-io.
- The non-supported external api telemetry query no will no longer report any APIs with negative summaries.
- Models are not produced for static initializers.

The output was produced by:
1. Delete the models content from IOGenerated.qll
2. Run our generation script
`time python3 GenerateFlowModel.py --with-sinks --with-summaries --with-negative-summaries ~/Work/databases/java/commons-io "apache/IOGenerated.qll"`


java/ql/src/utils/model-generator/internal/CaptureModelsSpecific.qll
```diff
@@ -55,7 +55,8 @@ private predicate isJdkInternal(J::CompilationUnit cu) {
 private predicate isRelevantForModels(J::Callable api) {
   not isInTestFile(api.getCompilationUnit().getFile()) and
   not isJdkInternal(api.getCompilationUnit()) and
-  not api instanceof J::MainMethod
+  not api instanceof J::MainMethod and
+  not api instanceof J::StaticInitializer
```
tamasvajk, line 59: Why are static initializers not relevant for model generation?

tamasvajk (commented):
Looks plausible to me. It would be great to understand why there are changes in the generated models.

java/ql/src/utils/model-generator/internal/CaptureModelsSpecific.qll
michaelnebel: They showed up in the negative generated summary models and I don't think it makes sense to include them for positive models either as the static initializers are not explicitly called.

michaelnebel:
> Looks plausible to me. It would be great to understand why there are changes in the generated models.

Yes, but I think that needs to be a separate project where we need to improve the tooling such that we can "see" what changes causes the models to change.

tamasvajk (approved):

java/ql/lib/semmle/code/java/frameworks/apache/NegativeIOGenerated.qll
```diff
@@ -0,0 +1,769 @@
...
+        "org.apache.commons.io.output;AbstractByteArrayOutputStream;toInputStream;();generated",
+        "org.apache.commons.io.output;AbstractByteArrayOutputStream;write;(InputStream);generated",
+        "org.apache.commons.io.output;AbstractByteArrayOutputStream;writeTo;(OutputStream);generated",
+        "org.apache.commons.io.output;BrokenOutputStream;BrokenOutputStream;();generated",
+        "org.apache.commons.io.output;BrokenOutputStream;BrokenOutputStream;(IOException);generated",
+        "org.apache.commons.io.output;BrokenWriter;BrokenWriter;();generated",
+        "org.apache.commons.io.output;BrokenWriter;BrokenWriter;(IOException);generated",
+        "org.apache.commons.io.output;ByteArrayOutputStream;ByteArrayOutputStream;();generated",
+        "org.apache.commons.io.output;ByteArrayOutputStream;ByteArrayOutputStream;(int);generated",
+        "org.apache.commons.io.output;ByteArrayOutputStream;toBufferedInputStream;(InputStream);generated",
```
intrigus-lgtm, line 376: It may be that I'm misunderstanding negative models, but why should there be no flow here?
> public static [InputStream](http://docs.oracle.com/javase/7/docs/api/java/io/InputStream.html?is-external=true) [toBufferedInputStream](https://commons.apache.org/proper/commons-io/javadocs/api-2.5/src-html/org/apache/commons/io/IOUtils.html#line.518)([InputStream](http://docs.oracle.com/javase/7/docs/api/java/io/InputStream.html?is-external=true) input)
                                         throws [IOException](http://docs.oracle.com/javase/7/docs/api/java/io/IOException.html?is-external=true)
>
> Fetches entire contents of an InputStream and represent same data as result InputStream.

michaelnebel: Yes, that would seem reasonable. The negative models are generated in the following way:
If the model generator does not discover flow, then we generate a negative model (stating there is no flow). The model generator is far from perfect and this is probably a case, where "positive" flow is missing.
In any case, the negative models are only used for telemetry purposes at the moment, so they don't interfere with dataflow analysis as such.

intrigus-lgtm: I didn't know that they are only for telemetry. 
Thank you for the explanation.

PR #10171, opened 2022-08-25
Python: Fixes for variable access

RasmusWL:
I found a problem with what we could track with type-trackers... but the problem ran much deeper, and required a fixup in the extractor.

(review commit by commit)


python/ql/lib/change-notes/2022-08-25-variable-fix.md
```diff
@@ -0,0 +1,4 @@
+---
+category: minorAnalysis
+---
+* Reads of global/non-local variables (without annotations) inside function defined on classes now works properly in the case where the class had an attribute defined with the same name as the non-local variable.
```
tausbn, line 4:
```suggestion
* Reads of global/non-local variables (without annotations) inside functions defined on classes now works properly in the case where the class had an attribute defined with the same name as the non-local variable.
```

tausbn (approved):
One small typo fix suggestion in the change note, otherwise this looks good to me! :+1:

tausbn (dismissed):

tausbn (approved):

PR #10175, opened 2022-08-25
Swift: fix missing extractions

redsun82:
There were missing extractions from the Builtin (and other) modules.

This was actually caused by two issues:
* we did not visit all required modules, as for example the `Builtin`
  module does not appear as being imported by anybody (together with
  another mysterious `__Objc` module)
* moreover the `Builtin` module works internally by only creating
  declarations on demand, and does not provide a list of its top level
  declarations.

The first problem was solved by moving the bulk of module collection to
the actual visiting.

The second one was solved by explicitly listing the builtin symbols
encountered during a normal extraction. This does mean this list needs
to be kept up to date, or possibly made truly exhaustive in the future.

With this change the currently running integration tests only have
the inconsistency caused by https://github.com/github/codeql-c-team/issues/1105.

Commit by commit reviewing is encouraged.


swift/extractor/SwiftExtractor.cpp
```diff
@@ -176,10 +186,16 @@ void codeql::extractSwiftFiles(const SwiftExtractorConfiguration& config,
...
-      extractDeclarations(config, compiler, *module, sourceFile);
+      encounteredModules = extractDeclarations(config, compiler, *module, sourceFile);
     }
     if (!isFromSourceFile) {
-      extractDeclarations(config, compiler, *module);
+      encounteredModules = extractDeclarations(config, compiler, *module);
+    }
+    processed.insert(module);
+    for (auto encountered : encounteredModules) {
+      if (processed.count(encountered) == 0) {
```
redsun82, line 196: This check should actually use seen modules, not already processed ones, or we can end up processing the same module multiple times. I will fix this tomorrow.

swift/extractor/SwiftExtractor.cpp
```diff
@@ -65,25 +66,119 @@ static std::string getFilename(swift::ModuleDecl& module, swift::SourceFile* pri
     filename += module.getName().str();
     return filename;
   }
+  if (module.isBuiltinModule()) {
+    // The Builtin module has an empty filename, let's fix that
+    return "/<Builtin>";
```
jketema, line 71: Drive-by comment: Probably not of much concern right now, but this would be an illegal file name on Windows due to `<` and `>` (see [here](https://docs.microsoft.com/en-us/windows/win32/fileio/naming-a-file#naming-conventions)). In practise this might also give some usability problems on other platforms, due to `<` and `>` begin used for redirected input and output (that's assuming the user starts digging into the generated trap files)..

AlexDenisov: I think this is a very good point: right now you can download an existing database and run analysis on Windows, if the db was not cleaned, then it might be problematic for Windows users.
This is of course theoretical, but I think it's worth fixing this.

AlexDenisov (approved):
Fantastic investigation!

swift/extractor/SwiftExtractor.cpp
redsun82: good point! I got inspiration by another module (the mysterious `__Objc`) that actually reports `<imports>` as a filename, leading to `<imports>.trap`. So with the same thing in mind, we might want to also fix that path, right?

swift/extractor/SwiftBuiltinSymbols.h
```diff
@@ -0,0 +1,71 @@
+#pragma once
+
+#include <array>
+
+namespace codeql {
+constexpr std::array swiftBuiltins = {
```
redsun82, line 6: C++17 deduction guidelines are so cool :slightly_smiling_face:

AlexDenisov (approved):

AlexDenisov (approved):

PR #10176, opened 2022-08-25
Python: Add testcase for import problem

RasmusWL:
Apparently ordering is currently important when using `import *` :grimacing: We should fix that, but here is a test-case to reproduce :blush:


tausbn (approved):
Thank you for this comprehensive test case! :+1:

PR #10177, opened 2022-08-25
Java: Promote `PathSanitizer.qll` from experimental

atorralba:
Promotes `PathSanitizer.qll` from experimental and uses it in `java/tainted-path`, `java/tainted-path-local` and `java/zipslip`. The deprecation of `PathTraversalBarrierGuard` wasn't necessary since it was previously in experimental and thus it was not importable from the main query pack, so I removed it.

Also, applied the changes from #9956 into `java/tainted-path-local` for consistency.

Note that the ZipSlip query already had a nice sanitizer in place, the features of which were merged into `PathSanitizer.qll`.

I recommend commit-by-commit review.


java/ql/lib/semmle/code/java/security/PathSanitizer.qll
```diff
@@ -123,7 +122,7 @@ private predicate blockListGuard(Guard g, Expr e, boolean branch) {
   )
 }
 
-private class BlockListSanitizer extends PathTraversalSanitizer {
+private class BlockListSanitizer extends PathInjectionSanitizer {
   BlockListSanitizer() { this = DataFlow::BarrierGuard<blockListGuard/3>::getABarrierNode() }
```
lcartey, line 126: For the `blockListGuard` we require that the user has mitigated for URL decoding issues. Is this really a universal requirement to use a block list?

As far as I know, the only impact URI decoding has on tainted path issues is that decoding should not occur after path checks, because the path may only be malicious after decoding.

However, requiring the user _decode_ the path before checking can be outright harmful if using a library or framework that already URI decodes remote flow sources, or for remote flow sources that don't require URI decoding.

lcartey (commented):
Provided some feedback on what I think we should consider a valid sanitizer for these queries.

java/ql/lib/semmle/code/java/security/PathSanitizer.qll
```diff
@@ -149,38 +148,51 @@ private class UrlEncodingSanitizer extends PathTraversalSanitizer {
  */
 private predicate isStringPartialMatch(MethodAccess ma) {
   ma.getMethod().getDeclaringType() instanceof TypeString and
-  ma.getMethod().getName() =
-    ["contains", "startsWith", "matches", "regionMatches", "indexOf", "lastIndexOf"]
```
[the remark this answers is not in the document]
atorralba: These are good suggestions, thanks! I'm actually modifying the logic of this sanitizer even further, so I'll try to fit this ideas in, too.

Keep in mind that, at some point, we need to compromise. Overfitting to one case may leave others out (either introducing FNs or FPs). For instance, in your example:

```java
Path path = Paths.get("/etc/passwd");
if (!path.contains("..")) {
  if (path.contains("mySpecialCase")) {
    sink(path);
  }
}
```

The payload would not actually reach the sink because `/etc/passwd` doesn't contain `mySpecialCase`. I agree the payload would go through if it looked like `/anotherDirectory/mySpecialCase/*`, whereas the intention of the developer probably was trying to restrict the path to `/mySpecialCase/*`, but the attacker's options are way more limited in this case (it funnily starts to look like a potential [partial path traversal](https://github.com/github/codeql/blob/e39475d7789ab4bea196e7510f75e5353af706bd/java/ql/src/Security/CWE/CWE-023/PartialPathTraversalFromRemote.ql#L4) instead).

Just saying this because some times `contains` et al might be enough to avoid exploitation — so not considering them would introduce FPs, but adding them could introduce FNs in other specific scenarios.

java/ql/lib/semmle/code/java/security/PathSanitizer.qll
atorralba: In general, block lists are less than ideal sanitizers, and allow lists should be preferred. One of the reasons for that is that it's easy to forget about a bad case, or to not consider different encodings when checking the externally-provided data.

So this condition is there to force block lists approaches to be a little more robust if we want them to be considered valid sanitizers. I agree that the requirement of URL decoding prior to the check might be too much, but theoretically your examples:

> if using a library or framework that already URI decodes remote flow sources, or for remote flow sources that don't require URI decoding.

could be fixed by marking those specific remote flow sources as `UrlDecoderSanitizer`s themselves.

The alternative would be that a simple `payload.contains("..")` would be considered a sanitizer, and then users would complain as well, since we all know that that doesn't work at all. Again, at some point we need to trade FNs for FPs, or vice versa.

java/ql/lib/semmle/code/java/security/PathSanitizer.qll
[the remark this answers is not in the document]
lcartey: Sorry, I chose a bad example for the `allowListGuard`, should have been something like this:
```java
String path = "/etc/passwd";
if (!path.contains("..")) {
  if (path.contains("pass")) {
    sink(path);
  }
}
```

> I agree the payload would go through if it looked like /anotherDirectory/mySpecialCase/*, whereas the intention of the developer probably was trying to restrict the path to /mySpecialCase/*, but the attacker's options are way more limited in this case (it funnily starts to look like a potential [partial path traversal](https://github.com/github/codeql/blob/e39475d7789ab4bea196e7510f75e5353af706bd/java/ql/src/Security/CWE/CWE-023/PartialPathTraversalFromRemote.ql#L4) instead).

Although the `contains` may limit the attackers choice of full path, it doesn't prevent an attacker from creating a path with any directory prefix on the system[1]. That always seems undesirable and unintended and although it may be more difficult to exploit that a vanilla tainted path example, it's also hard to convince yourself that it's _not_ exploitable because it depends on paths accessible to the process, software installed on the system and the exact set of allowed patterns.

Out of interest, do we have some examples of this type of allow list sanitizer?

> Keep in mind that, at some point, we need to compromise. Overfitting to one case may leave others out (either introducing FNs or FPs).

fwiw, for this rule I think poor sanitizers that cause false negatives are more of a problem than good sanitizers that cause false positives.


---
[1]: Assuming the value used as a path is created in a way that's vulnerable to absolute path traversal e.g. `new File(taintedValue);` and not `new File(parent, taintedValue)`. In an ideal world I think we would actually separate absolute and relative path traversal with flow states, and only consider the latter for relative path traversal, but I think that's probably a separate issue. 🙂

java/ql/lib/semmle/code/java/security/PathSanitizer.qll
lcartey: > The alternative would be that a simple payload.contains("..") would be considered a sanitizer

I agree that we don't want that, but I don't think requiring a URI decoding step beforehand makes `payload.contains("..")` a sufficient sanitizer for absolute path traversal, for the same reasons as above.

> could be fixed by marking those specific remote flow sources as UrlDecoderSanitizers themselves.

I don't think requiring URI decoding _before_ the check prevents any security issues. The URI decoding issue is only a problem if you URI decode _between_ the check and the access, e.g.
```
x = decode(source())
if (check(x)) {
  new File(x); // fine
}

x = source()
if (check(x)) {
  new File(x); // also fine, File won't URI decode
}

x = source()
if (check(x)) {
  new File(decode(x)); // bad!
}
```

For example, in the Spring vulnerability the issue wasn't that they forgot to URI decode before their tainted path check. Instead, the problem was that they ran second decode step _after_ the check, but before the value was used.

java/ql/lib/semmle/code/java/security/PathSanitizer.qll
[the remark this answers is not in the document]
atorralba: > fwiw, for this rule I think poor sanitizers that cause false negatives are more of a problem than good sanitizers that cause false positives.

You have more context from users, so let's follow your approach. See 60b31c6: `AllowListGuard` now only considers prefix checks (and not partial matching checks), and `regionMatches` and `matches` apply heuristics to make sure a prefix is being checked.

See the other thread below for the analogous changes in `BlockListGuard`.

> In an ideal world I think we would actually separate absolute and relative path traversal with flow states, and only consider the latter for relative path traversal, but I think that's probably a separate issue. 🙂

This sounds like a good improvement to do in a future PR. Noted, thanks!

java/ql/lib/semmle/code/java/security/PathSanitizer.qll
atorralba: Clearly, requiring taking double-encoding into account only makes sense if the sink itself implicitly URL-decodes the value, which happens in some dispatch/forward/redirect vulnerabilities. I think this could actually be another interesting use case for flow states (see the comment I added in the now FN test in https://github.com/github/codeql/pull/10177/files#diff-6bc44c5765a6e460db430d2c13a69b22afe818b46bc96d630508b5e1d1cd6eb1R103-R106).

So, with that in mind, I followed your advice and removed the URL-decode requirement for `blockListGuard`. Instead, I added the same canonicalization/path traversal check requirement that `allowListGuard` has, assuming that the block list will check for dangerous prefixes (like `/data` in Android). The only case that would not need normalization would be `contains(File.seperator)` (or `contains("/")` or `contains("\\")`), but to special-case those bad words we would need to create yet another guard. We can discuss if you think that'd be worth it.

atorralba:
@aschackmull I introduced a new parameterized module [here](https://github.com/github/codeql/pull/10177/files#diff-38978229f15815dd860ec9f3696f469770f85b8082736b808072ef64be18a834R15-R43). If you have the chance to review it, let me know if something doesn't look right.

aschackmull:
> @aschackmull I introduced a new parameterized module [here](https://github.com/github/codeql/pull/10177/files#diff-38978229f15815dd860ec9f3696f469770f85b8082736b808072ef64be18a834R15-R43). If you have the chance to review it, let me know if something doesn't look right.

Had a quick look - looks fine to me, I think. Just make sure you have some test coverage for that - we may want to fold some of it into the BarrierGuard module at some point.

java/ql/lib/semmle/code/java/security/PathSanitizer.qll
```diff
@@ -0,0 +1,260 @@
...
+        this = eq and
+        ma.getMethod().hasName(["indexOf", "lastIndexOf"]) and
+        eq.getAnOperand() = ma and
+        eq.getAnOperand().(CompileTimeConstantExpr).getIntValue() = -1
+      )
+    )
+  }
+
+  Expr getCheckedExpr() {
+    exists(MethodAccess ma | ma = this.(BinaryExpr).getAnOperand() or ma = this |
```
atorralba, line 227:
```suggestion
    exists(MethodAccess ma | ma = this.(EqualityTest).getAnOperand() or ma = this |
```

vlkl-sap:
Hey. Thanks for tackling this! It's a real improvement. I particularly like the tests as they serve as a kind of spec document.

Would it make sense to rename `AllowListGuard` and `BlockListGuard` to `AllowPrefixGuard` and `BlockPrefixGuard` respectively?

vlkl-sap:
Another thing: I'm struggling to understand the `WEB-INF` part. It is OK to block a prefix with "WEB-INF" somewhere in it but it's not OK to block "WEB-INF" in the middle of a string? Why is that?

atorralba:
Hey @vlkl-sap, thanks for the feedback!

> Would it make sense to rename `AllowListGuard` and `BlockListGuard` to `AllowPrefixGuard` and `BlockPrefixGuard` respectively?

`BlockListGuard` actually handles both prefixes and a few restricted substrings that can appear anywhere in the path, so the name seems appropriate. It's true though that `AllowListGuard` could use a better name, and the documentation needs to reflect this too.

> It is OK to block a prefix with "WEB-INF" somewhere in it but it's not OK to block "WEB-INF" in the middle of a string? Why is that?

Hmm, that seems wrong indeed. Requiring prefix checks instead of substring checks is something that makes sense in the allow list approach, but for the block list approach substring checks work fine (actually, even better from a security POV since they cover more cases). Thanks for pointing this out, I'll fix it.

vlkl-sap:
Are these duplicates?
https://github.com/github/codeql/blob/c5abbbae93ee83b7a4767fcaa2ed832f7e258699/java/ql/test/library-tests/pathsanitizer/Test.java#L252-L265
Or did I fail the eye test?

atorralba:
> Are these duplicates?

Indeed they are! One of those two should have been the `!= -1` case.

See the recent commits for fixes to all the issues you mentioned.

Thanks again for your detailed review @vlkl-sap!

yo-h:
> This branch has conflicts that must be resolved
> Conflicting files
> java/ql/src/Security/CWE/CWE-022/TaintedPathCommon.qll

atorralba:
> > This branch has conflicts that must be resolved
> > Conflicting files
> > java/ql/src/Security/CWE/CWE-022/TaintedPathCommon.qll

Thanks @yo-h! Conflict resolved.

java/ql/lib/semmle/code/java/security/PathSanitizer.qll
```diff
@@ -0,0 +1,254 @@
...
+
+/** A sanitizer that protects against path injection vulnerabilities. */
+abstract class PathInjectionSanitizer extends DataFlow::Node { }
+
+/**
+ * Provides a set of nodes validated by a method that uses a validation guard.
+ */
+private module ValidationMethod<DataFlow::guardChecksSig/3 validationGuard> {
+  /** Gets a node that is safely guarded by a method that uses the given guard check. */
+  DataFlow::Node getAValidatedNode() {
```
aschackmull, line 17: This might need an unbind on `pos`. Have you checked the join-order?

java/ql/lib/semmle/code/java/security/PathSanitizer.qll
```diff
@@ -0,0 +1,254 @@
...
+}
+
+/**
+ * Holds if `g` is a guard that considers a path safe because it is checked against trusted prefixes.
+ * This requires additional protection against path traversal, either another guard (`PathTraversalGuard`)
+ * or a sanitizer (`PathNormalizeSanitizer`), to ensure any internal `..` components are removed from the path.
+ */
+private predicate allowedPrefixGuard(Guard g, Expr e, boolean branch) {
+  branch = true and
+  TaintTracking::localExprTaint(e, g.(AllowedPrefixGuard).getCheckedExpr()) and
```
aschackmull, line 88: Why the taint-flow here? Why not simply `e = g.(AllowedPrefixGuard).getCheckedExpr()`?

java/ql/lib/semmle/code/java/security/PathSanitizer.qll
```diff
@@ -0,0 +1,254 @@
...
+  exists(Expr previousGuard |
+    TaintTracking::localExprTaint(previousGuard.(PathNormalizeSanitizer),
+      g.(AllowedPrefixGuard).getCheckedExpr())
+    or
+    previousGuard
+        .(PathTraversalGuard)
+        .controls(g.getBasicBlock().(ConditionBlock), previousGuard.(PathTraversalGuard).getBranch())
```
aschackmull, line 95: Why the cast to `ConditionBlock` here?

java/ql/lib/semmle/code/java/security/PathSanitizer.qll
```diff
@@ -0,0 +1,254 @@
...
+  }
+}
+
+/**
+ * Holds if `g` is a guard that considers a path safe because it is checked for `..` components, having previously
+ * been checked for a trusted prefix.
+ */
+private predicate dotDotCheckGuard(Guard g, Expr e, boolean branch) {
+  branch = g.(PathTraversalGuard).getBranch() and
+  TaintTracking::localExprTaint(e, g.(PathTraversalGuard).getCheckedExpr()) and
```
aschackmull, line 112: Why taint flow here? Why not just `e = g.(PathTraversalGuard).getCheckedExpr()`?

java/ql/lib/semmle/code/java/security/PathSanitizer.qll
```diff
@@ -0,0 +1,254 @@
...
+  exists(Guard previousGuard |
+    previousGuard.(AllowedPrefixGuard).controls(g.getBasicBlock().(ConditionBlock), true)
+    or
+    previousGuard.(BlockListGuard).controls(g.getBasicBlock().(ConditionBlock), false)
```
aschackmull, line 116: Why cast to `ConditionBlock` here?

java/ql/lib/semmle/code/java/security/PathSanitizer.qll
```diff
@@ -0,0 +1,254 @@
...
+}
+
+/**
+ * Holds if `g` is a guard that considers a string safe because it is checked against a blocklist of known dangerous values.
+ * This requires additional protection against path traversal, either another guard (`PathTraversalGuard`)
+ * or a sanitizer (`PathNormalizeSanitizer`), to ensure any internal `..` components are removed from the path.
+ */
+private predicate blockListGuard(Guard g, Expr e, boolean branch) {
+  branch = false and
+  TaintTracking::localExprTaint(e, g.(BlockListGuard).getCheckedExpr()) and
```
aschackmull, line 143: same question

java/ql/lib/semmle/code/java/security/PathSanitizer.qll
```diff
@@ -0,0 +1,254 @@
...
+  exists(Expr previousGuard |
+    TaintTracking::localExprTaint(previousGuard.(PathNormalizeSanitizer),
+      g.(BlockListGuard).getCheckedExpr())
+    or
+    previousGuard
+        .(PathTraversalGuard)
+        .controls(g.getBasicBlock().(ConditionBlock), previousGuard.(PathTraversalGuard).getBranch())
```
aschackmull, line 150: same question

java/ql/lib/semmle/code/java/security/PathSanitizer.qll
```diff
@@ -0,0 +1,254 @@
...
+      this.getMethod().hasName("normalize")
+    )
+    or
+    this.getMethod().getDeclaringType() instanceof TypeFile and
+    this.getMethod().hasName(["getCanonicalPath", "getCanonicalFile"])
+  }
+}
+
+/** A node with path normalization. */
+class NormalizedPathNode extends DataFlow::Node {
```
aschackmull, line 250: Unused code? Also, this class seems a bit weird as it defines the set of nodes that precedes a path normalization.

java/ql/lib/semmle/code/java/security/PathSanitizer.qll
aschackmull: Ah, I see it's copied from zipslip, probably fine then.

java/ql/lib/semmle/code/java/security/PathSanitizer.qll
atorralba: Hm, probably a leftover of the code before promotion. I'll remove it.

java/ql/lib/semmle/code/java/security/PathSanitizer.qll
atorralba: Similarly, this is to cover:

```java
Path path = source();
String strPath = path.toString();
if (!strPath.contains("/"))
  sink(path);
```

java/ql/lib/semmle/code/java/security/PathSanitizer.qll
atorralba: Same answer, probably a leftover, I'll remove it.

java/ql/lib/semmle/code/java/security/PathSanitizer.qll
atorralba: This is used in the experimental query `FilePathInjection.ql`, and I didn't want to break it:

```
override predicate isSink(DataFlow::Node sink) {
    sink.asExpr() = any(PathCreation p).getAnInput() and
    not sink instanceof NormalizedPathNode
  }
```

I can move this class back to the experimental folder if we don't feel like promoting it.

java/ql/lib/semmle/code/java/security/PathSanitizer.qll
atorralba: See answer below.

java/ql/lib/semmle/code/java/security/PathSanitizer.qll
atorralba: See answer below.

java/ql/lib/semmle/code/java/security/PathSanitizer.qll
aschackmull: Let's move it back for now, as I think it looks a bit weird.

java/ql/lib/semmle/code/java/security/PathSanitizer.qll
aschackmull: Gotcha.

java/ql/lib/semmle/code/java/security/PathSanitizer.qll
aschackmull: Maybe actually explain this in a comment for the benefit of future readers.

aschackmull:
Looks like there's a performance problem somewhere.

atorralba:
> Looks like there's a performance problem somewhere.

I've rebased main and run DCA again, to make sure that a) it wasn't a one-time thing and b) we benefit from your recent virtual dispatch changes (because I was seeing _more_ path injection alerts, which is weird). Also I added the tuple-counting option this time.

atorralba:
DCA looks good now, with the exception of some increased tuple counts (I'm not sure we can do much about that?). There's some increase (and also some decrease) in some projects' path injection alerts, which seems logic — now we are accounting for incomplete path injection sanitization/validation attempts, so in some cases we'll now be flagging new results that we previously considered safe.

PR #10180, opened 2022-08-25
Add missing security tags

erik-krogh:
I stumbled upon a query that had a missing `@tag security`.  
So I took a look at the results of `ql/missing-security-metadata`, and added the missing tags were I could.  

[There still some queries with missing @security-severity](https://github.com/github/codeql/security/code-scanning?query=pr%3A10180+tool%3ACodeQL+is%3Aopen+rule%3Aql%2Fmissing-security-metadata), but they are also not tagged with any CWEs, so I'm not sure what to set the value to.

### DIFF
--- a/java/ql/src/Frameworks/JavaEE/EJB/EjbSetSocketOrUrlFactory.ql
+++ b/java/ql/src/Frameworks/JavaEE/EJB/EjbSetSocketOrUrlFactory.ql
@@ -4,7 +4,7 @@
  *              or the stream handler factory used by URL. Such operations could
  *              compromise security or interfere with the EJB container's operation.
  * @kind problem
- * @security-severity 5.8
+ * @problem.severity error
  * @precision low
  * @id java/ejb/socket-or-stream-handler-factory
  * @tags reliability
--- a/java/ql/src/Frameworks/JavaEE/EJB/EjbSetSocketOrUrlFactory.ql
+++ b/java/ql/src/Frameworks/JavaEE/EJB/EjbSetSocketOrUrlFactory.ql
@@ -4,7 +4,6 @@
  *              or the stream handler factory used by URL. Such operations could
  *              compromise security or interfere with the EJB container's operation.
  * @kind problem
- * @problem.severity error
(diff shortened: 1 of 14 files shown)


atorralba:
According to https://github.com/github/codeql/pull/7613, I think the Java queries should have the `security-severity` tag removed, rather than the tag `security` added (or just be ignored).

Also, I think this needs a `queryMetadata` change note, akin to https://github.com/github/codeql/pull/8426/files#diff-5b76977c943c81b88d65fad2c4cad38a86773b31ddd1a3a744040eb4b81b5aaa.

RasmusWL (dismissed):
Python :+1: (thanks)

java/ql/src/Frameworks/JavaEE/EJB/EjbSetSocketOrUrlFactory.ql
```diff
@@ -4,7 +4,6 @@
  *              or the stream handler factory used by URL. Such operations could
  *              compromise security or interfere with the EJB container's operation.
  * @kind problem
- * @problem.severity error
  * @security-severity 5.8
```
atorralba, line 7: I think the intention here was to remove the line below instead.

java/ql/src/change-notes/2022-08-29-remove-security-tag.md
```diff
@@ -0,0 +1,4 @@
+---
+category: queryMetadata
+---
+* Removed the `security` tag from several queries not in the `Security/` folder that also had missing `@security-severity` tags.
```
atorralba, line 4: It's the other way around, isn't it?

```suggestion
* Removed the `@security-severity` tag from several queries not in the `Security/` folder that also had missing `security` tags.
```

java/ql/src/Frameworks/JavaEE/EJB/EjbSetSocketOrUrlFactory.ql
erik-krogh: Yes it was 👍

PR #10181, opened 2022-08-25
QL: remove consistency errors related to resolving multiple predicates from parameterized modules

erik-krogh:
Predicate references in prameterized modules will resolve to multiple predicates, that's a feature (for now).  
This fixes a FP in the consistency test related to that.

### DIFF
--- a/ql/ql/src/codeql_ql/ast/internal/Predicate.qll
+++ b/ql/ql/src/codeql_ql/ast/internal/Predicate.qll
@@ -211,7 +211,7 @@ module PredConsistency {
     c > 1 and
     resolvePredicateExpr(pe, p) and
     // parameterized modules are expected to resolve to multiple.
-    not exists(Predicate sig | not exists(sig.getBody()) and resolvePredicateExpr(pe, sig))
+    not exists(ClasslessPredicate sig | not sig.isSignature() and resolvePredicateExpr(pe, sig))
   }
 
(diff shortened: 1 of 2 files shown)


ql/ql/src/codeql_ql/ast/internal/Predicate.qll
```diff
@@ -209,7 +209,9 @@ module PredConsistency {
         not exists(p0.getAlias())
       ) and
     c > 1 and
-    resolvePredicateExpr(pe, p)
+    resolvePredicateExpr(pe, p) and
+    // parameterized modules are expected to resolve to multiple.
+    not exists(Predicate sig | not exists(sig.getBody()) and resolvePredicateExpr(pe, sig))
```
kaeluka, line 214: `not exists(sig.getBody)` is the way you're checking that the predicate is part of a parameterized module interface? If so, I think that would well deserve a named predicate or a named class. A name would make this line easier to read, and it would also make it less likely that others in the future have to re-invent/re-discover this logic.

kaeluka (approved):
I've tried to understand this PR and my interpretation is that you're seeing consistency errors due to this line:
https://github.com/github/codeql/blob/e39475d7789ab4bea196e7510f75e5353af706bd/ql/ql/src/queries/diagnostics/EmptyConsistencies.ql#L31

If that interpretation is correct and you've read my other comment, merge away.

kaeluka (approved):
LGTM!

PR #10184, opened 2022-08-26
JavaScript: Update to TypeScript 4.8

erik-krogh:
https://devblogs.microsoft.com/typescript/announcing-typescript-4-8

The most tricky part was [a change to the syntax tree representation](https://devblogs.microsoft.com/typescript/announcing-typescript-4-8/#decorators-are-placed-on-modifiers-on-typescripts-syntax-trees).   
The `modifiers` property that we rely on in `TypeScriptASTConverter.java` might disappear at a later time.  
The fix at that point will be to use `ts.canHaveModifiers()` and `ts.getModifiers()`.   

TypeScript changed how illegal decorators are stored between 4.8.1-RC and 4.8.2 ([this PR](https://github.com/microsoft/TypeScript/pull/50343)). So I had to fix that.  

Otherwise it was an uneventful update.  
There's not much new, and what's new doesn't affect us, except that we get slightly more precise types from TypeScript.

Evaluation will appear [here](https://github.com/github/codeql-dca-main/tree/data/erik-krogh/ts48__typescript-full__code-scanning__1/reports).

### DIFF
--- a/javascript/extractor/lib/typescript/src/ast_extractor.ts
+++ b/javascript/extractor/lib/typescript/src/ast_extractor.ts
@@ -202,6 +202,12 @@ export function augmentAst(ast: AugmentedSourceFile, code: string, project: Proj
             }
         }
 
+        // This PR: https://github.com/microsoft/TypeScript/pull/50343/ changed how illegal decorators are stored, this hack reverts it.
+        // Those decorator uses are syntax errors, but people use them anyway.
+        if ((node as any).illegalDecorators) {
+            (node as any).decorators = (node as any).illegalDecorators;
+        }
+
         if (typeChecker != null && insideConditionalTypes === 0) {
             if (isTypedNode(node)) {
                 let contextualType = isContextuallyTypedNode(node)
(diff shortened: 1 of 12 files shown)


javascript/extractor/lib/typescript/src/ast_extractor.ts
```diff
@@ -202,6 +202,12 @@ export function augmentAst(ast: AugmentedSourceFile, code: string, project: Proj
             }
         }
 
+        // This PR: https://github.com/microsoft/TypeScript/pull/50343/ changed how illegal decorators are stored, this hack reverts it.
+        // Those decorator uses are syntax errors, but people use them anyway.
+        if ((node as any).illegalDecorators) {
+            (node as any).decorators = (node as any).illegalDecorators;
```
asgerf, line 208: I'd like to avoid modifying the `decorators` property directly. Injecting our own properties (prefixed with `$`) is one thing, but modifying properties owned by TypeScript compiler is too brittle.

I'd rather to something like `node.decorators ?? node.illegalDecorators` at the point where we access the decorators.

asgerf (approved):
👍

PR #10190, opened 2022-08-26
C++: Switch to use-use flow in IR dataflow

MathiasVP:
This is a very large redesign of the C++ IR dataflow library. Most importantly, it switches C++ to use-use flow (instead of def-use flow) like all the other languages that use the shared dataflow library.

Previously, this has been impossible to do for C++. To see why, consider the following example:
```cpp
const char* path = getenv("PATH"); // *path is tainted
const char* path2 = path;
sink(path);
```
Most other languages would have flow from `path` on line 1 to `path` on line 2, and from `path` on line 2 to `path` on line 3. However, if we just step from the expression node on line 1 to the expression node on line 2 we conflate the pointed-to value with the pointer value.

Instead, C++ currently has flow from `path` on line 1 to `path` on both lines 2 and 3. This prevents pointee-pointer conflation, but it also has a number of bad side effects:
- It makes barriers terribly difficult to use.
- It will sometimes explode on codebases where a single definition have thousands of uses leading to bad performance.
- It makes it impossible to implement `clearsContent` for C/C++ (whereas this is a trivial one-liner in languages with use-use flow).

All of these issues will be fixed by this PR.

I expect this PR will take quite a while to go through. While each commit doesn't compile on its own, it's meant to be reviewed commit-by-commit.


cpp/ql/lib/semmle/code/cpp/ir/dataflow/internal/SsaInternals.qll
```diff
@@ -1,636 +1,522 @@
...
-  iTo.(FieldAddressInstruction).getObjectAddress() = iFrom
-  or
-  // We traverse `LoadInstruction`s since we want to conclude that the
-  // destination of the store operation `*x = source()` is derived from `x`.
-  iTo.(LoadInstruction).getSourceAddress() = iFrom
+private predicate defToNode(Node nodeFrom, Def def) {
+  nodeHasInstruction(nodeFrom, def.getDefiningInstruction(), def.getIndex())
+}
+
+private predicate nodeToDefOrUse(Node nodeFrom, SsaDefOrUse defOrUse) {
```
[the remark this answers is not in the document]
MathiasVP: Correct

cpp/ql/lib/semmle/code/cpp/ir/dataflow/internal/SsaInternalsCommon.qll
```diff
@@ -0,0 +1,167 @@
...
+    result = 1 + countIndirections(type.getUnspecifiedType())
+  )
+  or
+  exists(Type type | langType.hasType(type, false) |
+    result = countIndirections(type.getUnspecifiedType())
+  )
+}
+
+class AllocationInstruction extends CallInstruction {
+  AllocationInstruction() { this.getStaticCallTarget() instanceof Cpp::AllocationFunction }
```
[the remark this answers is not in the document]
MathiasVP: Yeah, this also handles `new` properly. This snippet:
```cpp
void test() {
    int* x = new int;
}
```
has the following IR:
```cpp
void test()
  Block 0
    v4_1(void)           = EnterFunction                 :
    m4_2(unknown)        = AliasedDefinition             :
    m4_3(unknown)        = InitializeNonLocal            :
    m4_4(unknown)        = Chi                           : total:m4_2, partial:m4_3
    r5_1(glval<int *>)   = VariableAddress[x]            :
    r5_2(glval<unknown>) = FunctionAddress[operator new] :
    r5_3(unsigned long)  = Constant[4]                   :
    r5_4(void *)         = Call[operator new]            : func:r5_2
    m5_5(unknown)        = ^CallSideEffect               : ~m4_4
    m5_6(unknown)        = Chi                           : total:m4_4, partial:m5_5
    m5_7(unknown)        = ^InitializeDynamicAllocation  : &:r5_4
    r5_8(int *)          = Convert                       : r5_4
    m5_9(int *)          = Store[x]                      : &:r5_1, r5_8
    v6_1(void)           = NoOp                          :
    v4_5(void)           = ReturnVoid                    :
    v4_6(void)           = AliasedUse                    : ~m5_6
    v4_7(void)           = ExitFunction                  :
```
and `r5_4` is correctly identified as an `AllocationInstruction`.

cpp/ql/lib/semmle/code/cpp/ir/dataflow/internal/SsaInternals.qll
```diff
@@ -1,636 +1,522 @@
...
+
+  /**
+   * Gets the index (i.e., the number of loads required) of this
+   * definition or use.
+   *
+   * Note that this is _not_ the definition's (or use's) index in
+   * the enclosing basic block. To obtain this index, use
+   * `DefOrUseImpl::hasIndexInBlock/2` or `DefOrUseImpl::hasIndexInBlock/3`.
+   */
+  abstract int getIndex();
```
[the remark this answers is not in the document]
MathiasVP: Yeah, that might be better 👍

cpp/ql/lib/semmle/code/cpp/ir/dataflow/internal/SsaInternalsCommon.qll
```diff
@@ -0,0 +1,167 @@
...
+          instr.(InitializeParameterInstruction).getAnOperand(),
+          instr.(InitializeDynamicAllocationInstruction).getAllocationAddressOperand(),
+          instr.(UninitializedInstruction).getAnOperand()
+        ]
+    |
+      isDefImpl(address, base, ind0) and
+      type = getLanguageType(address) and
+      m = countIndirectionsForCppType(type) and
+      ind = [ind0 + 1 .. ind0 + m] and
+      index = ind - (ind0 + 1)
```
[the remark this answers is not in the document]
MathiasVP: Yeah, this is a bit strange. The case for `isUse` is easier to understand, I think. For `isUse` it's:
- `ind = [ind0 .. ind0 + m]` (which might be easier read as `ind = ind0 + [0 .. m]`?) is the set of possible number of loads that can be done on `address`. `ind0` is the number of loads already performed on `address`, and we can only perform `m` number of loads due to the type of the address (i.e., for `int*` we can do 1 load).
- `index` is useful number to work with for the rest of SSA. `index = 0` means that no additional loads need to be performed to read the value represented by this use (and so the use can be represented directly by the `operand` in the IR). Any index greater than zero means that a number of loads are required to reach the value pointed to by the operand (and these uses will be represented by `IndirectOperand` and `IndirectInstruction` in a later commit in this PR).

For `isDef` we add `1` to `ind` to account for rvalue-ness (so that a definition like `x = 42` is a definition of the value pointed to by `&x`), and we modify `index` accordingly (so that `index = 0` means that this is the value represented directly by the operand in the IR).

I hope that makes sense. if not, I'd be happy to discuss this further in our chat later.

cpp/ql/lib/semmle/code/cpp/ir/dataflow/internal/DataFlowUtil.qll
```diff
@@ -17,52 +17,37 @@ cached
...
-    TSsaPhiNode(Ssa::PhiNode phi)
+    TPostFieldUpdateNode(FieldAddress operand, int index) {
+      index =
+        [0 .. Ssa::countIndirectionsForCppType(operand.getObjectAddress().getResultLanguageType()) -
+            1]
+    } or
+    TSsaPhiNode(Ssa::PhiNode phi) or
+    TIndirectArgumentOutNode(ArgumentOperand operand, int index) {
+      Ssa::isModifiableByCall(operand) and
+      index = [0 .. Ssa::countIndirectionsForCppType(operand.getLanguageType()) - 1]
```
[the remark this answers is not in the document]
MathiasVP: Possibly, yeah. We need to be a bit careful, though. As we do want an `IndirectArgumentOutNode` for `S * const *` where `S` is some struct since you can still do:
```cpp
struct S {
  int x;
};

void foo(S * const * s) {
  (*s)->x = source();
}

void test() {
  S* s;
  foo(&s);
  sink(s->x);
}
```

cpp/ql/lib/semmle/code/cpp/ir/dataflow/internal/DataFlowPrivate.qll
```diff
@@ -154,53 +150,146 @@ class ReturnKind extends TReturnKind {
...
+ * out of a function. However, the IR will unconditionally create those for a
+ * variable passed to a function even though the variable was never updated by
+ * the function. And if a function has too many `ReturnNode`s the dataflow
+ * library lowers its precision for that function by disabling field flow.
+ * 
+ * So we those eliminate `ReturnNode`s that would have otherwise been created
+ * by this unconditional `ReturnIndirectionInstruction` by requiring that there
+ * must exist an SSA definition of the IR variable in the function.
+ */
+private predicate hasNonInitializeParameterDef(IRVariable v) {
```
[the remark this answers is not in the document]
MathiasVP: Oh, yes. That would be fantastic!

cpp/ql/lib/semmle/code/cpp/ir/dataflow/internal/SsaInternalsCommon.qll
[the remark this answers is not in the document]
rdmarsh2: Yeah, that makes sense. I was just confused by the redundancy between `ind` and `index` as predicate parameters.

cpp/ql/lib/semmle/code/cpp/ir/dataflow/internal/DataFlowUtil.qll
```diff
@@ -211,260 +223,412 @@ class OperandNode extends Node, TOperandNode {
...
 
-  override Location getLocation() { result = op.getLocation() }
+  final override Location getLocationImpl() { result = op.getLocation() }
 
-  override string toString() { result = this.getOperand().toString() }
+  override string toStringImpl() { result = this.getOperand().toString() }
 }
 
 /**
  * INTERNAL: do not use.
```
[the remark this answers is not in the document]
MathiasVP: It probably could, yeah. It's a bit annoying since that would imply removing the private annotation on the cached module `Cached` that wraps the IPA types.

cpp/ql/lib/semmle/code/cpp/ir/dataflow/internal/SsaInternalsCommon.qll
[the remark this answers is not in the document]
MathiasVP: Yeah, we might be able to remove the `ind` parameter altogether. I'll see if I can make that work.

cpp/ql/lib/semmle/code/cpp/ir/dataflow/internal/DataFlowUtil.qll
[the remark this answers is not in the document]
rdmarsh2: Couldn't that also go in DataFlowPrivate?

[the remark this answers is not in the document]
MathiasVP: Hmmm... Good question. Maybe it could, yeah.

cpp/ql/src/Security/CWE/CWE-129/ImproperArrayIndexValidation.ql
```diff
@@ -50,8 +46,12 @@ predicate hasUpperBoundsCheck(Variable var) {
...
 
 predicate nodeIsBarrierEqualityCandidate(DataFlow::Node node, Operand access, Variable checkedVar) {
-  readsVariable(node.asInstruction(), checkedVar) and
-  any(IRGuardCondition guard).ensuresEq(access, _, _, node.asInstruction().getBlock(), true)
+  exists(VariableAccess va |
+    va = node.asExpr() and
+    va.getTarget() = checkedVar and
+    access.getDef().getAst() = va and
+    any(IRGuardCondition guard).ensuresEq(access, _, _, access.getDef().getBlock(), true)
+  )
```
rdmarsh2, line 54: Is this actually equivalent? It looks like `access` and `node` were only tied together by the `Variable` previously, but now they're tied by the `VariableAccess`.

aschackmull:
Here's something to aspire to: Put qldoc on _all_ predicates. See e.g. `cpp/ql/lib/semmle/code/cpp/ir/dataflow/internal/SsaInternalsCommon.qll`, which doesn't have even a single line of qldoc.

cpp/ql/lib/semmle/code/cpp/ir/dataflow/internal/SsaInternalsCommon.qll
```diff
@@ -0,0 +1,167 @@
...
+    call.getArgumentOperand(index) = operand and
+    if index = -1
+    then not call.getStaticCallTarget() instanceof Cpp::ConstMemberFunction
+    else not SideEffects::isConstPointerLike(any(Type t | type.hasType(t, _)))
+  )
+}
+
+cached
+private module DefCached {
+  cached
```
aschackmull, line 84: There's only one cached predicate in here - no need to use a cached module.

cpp/ql/lib/semmle/code/cpp/ir/dataflow/internal/SsaInternalsCommon.qll
```diff
@@ -0,0 +1,167 @@
...
+      ind0 = ind - 1
+    )
+  }
+}
+
+import DefCached
+
+cached
+private module UseCached {
+  cached
```
aschackmull, line 133: Same. Did you mean to put both of these in the same cached module aka same compilation stage?

cpp/ql/lib/semmle/code/cpp/ir/dataflow/internal/SsaInternals.qll
```diff
@@ -1,636 +1,526 @@
...
+
+  /**
+   * Gets the index (i.e., the number of loads required) of this
+   * definition or use.
+   *
+   * Note that this is _not_ the definition's (or use's) index in
+   * the enclosing basic block. To obtain this index, use
+   * `DefOrUseImpl::hasIndexInBlock/2` or `DefOrUseImpl::hasIndexInBlock/3`.
+   */
+  abstract int getIndirectionIndex();
```
hvitved, line 147: Would `getIndirectionCount` be a better name?

cpp/ql/lib/semmle/code/cpp/ir/dataflow/internal/SsaInternals.qll
```diff
@@ -1,636 +1,526 @@
...
-private class ReturnParameterIndirection extends Use, TReturnParamIndirection {
-  ReturnParameterIndirection() { this = TReturnParamIndirection(use) }
+  predicate isCertain() { isUse(true, operand, _, _, ind) }
+}
 
-  override SourceVariable getSourceVariable() {
-    exists(ReturnIndirectionInstruction ret |
-      returnParameterIndirection(use, ret) and
-      result.(SourceIRVariableIndirection).getUnderlyingIRVariable() = ret.getIRVariable()
+predicate adjacentDefRead(DefOrUse defOrUse1, UseOrPhi use) {
```
hvitved, line 258: I think this predicate deserves a comment.

cpp/ql/lib/semmle/code/cpp/ir/dataflow/internal/SsaInternals.qll
```diff
@@ -1,636 +1,526 @@
...
+  exists(IRBlock bb1, int i1, SourceVariable v |
+    defOrUse1.asDefOrUse().hasIndexInBlock(bb1, i1, v)
+  |
+    exists(IRBlock bb2, int i2 |
+      adjacentDefRead(_, pragma[only_bind_into](bb1), pragma[only_bind_into](i1),
```
hvitved, line 263: I _think_ this could in principle go wrong, if `adjacentDefRead(def, bb1, i1, bb2, i2)` happens to hold for a _different_ SSA definition (with another source variable) than the one underlying `defOrUse1`. But only if multiple variables could be referenced at the same index (which it can in C# for psuedo-reads).

cpp/ql/lib/semmle/code/cpp/ir/dataflow/internal/SsaInternals.qll
```diff
@@ -1,636 +1,526 @@
...
-      instr.(LoadInstruction).getSourceAddressOperand(),
-      instr.(ReadSideEffectInstruction).getArgumentOperand()
-    ]
+/** Holds if `nodeTo` receives flow from the phi node `nodeFrom`. */
+predicate fromPhiNode(SsaPhiNode nodeFrom, Node nodeTo) {
+  exists(UseOrPhi use, IRBlock bb2, int i2, PhiNode phi, SourceVariable v |
+    use.asDefOrUse().hasIndexInBlock(bb2, i2, v) and
+    phi = nodeFrom.getPhiNode() and
+    phi.getSourceVariable() = v and
+    adjacentDefRead(phi, _, -1, bb2, i2) and
```
hvitved, line 369: I don't think you should rely on `-1`, that is an internal implementation detail.

cpp/ql/lib/semmle/code/cpp/ir/dataflow/internal/SsaInternals.qll
```diff
@@ -1,636 +1,526 @@
...
+    useToNode(use, nodeTo)
+  )
+  or
+  exists(PhiNode phiFrom, PhiNode phiTo |
```
hvitved, line 373: Did you deliberately intend to not use `lastRefRedef` here? As is, there could be an intermediate read between `nodeFrom` and `nodeTo`.

cpp/ql/lib/semmle/code/cpp/ir/dataflow/internal/SsaInternals.qll
```diff
@@ -1,636 +1,526 @@
...
   }
+}
 
-  private predicate fromPhiNode(SsaPhiNode nodeFrom, Node nodeTo) {
-    exists(PhiNode phi, Use use, IRBlock block, int rnk |
-      phi = nodeFrom.getPhiNode() and
-      adjacentDefRead(phi, _, _, block, rnk) and
-      use.hasIndexInBlock(block, rnk) and
-      flowOutOfAddressStep(use.getOperand(), nodeTo)
+private newtype TSsaDefOrUse =
```
hvitved, line 442: Should this have been cached?

cpp/ql/lib/semmle/code/cpp/ir/dataflow/internal/DataFlowUtil.qll
```diff
@@ -11,77 +11,87 @@ private import semmle.code.cpp.ir.IR
...
-   *   the final address of the chain.
-   * - Flow can enter a `SsaPhiNode` from an `InstructionNode`, a `StoreNode` or another `SsaPhiNode`
-   *   (in `toPhiNode`), depending on which node provided the previous definition of the underlying
-   *   variable. Flow leaves a `SsaPhiNode` (in `fromPhiNode`) by using the shared SSA library to
-   *   determine the next use of the variable.
+   * - `IndirectArgumentOutNode`, which represents the value of an argument (and its indirections) after
+   * it leaves a function call.
+   * - `IndirectOperand`, which represents the value of `operand` after loading the address a number
+   * of times.
+   * `IndirectInstruction`, which represents the value of `instr` after loading the address a number
```
hvitved, line 31: Missing `-`

cpp/ql/lib/semmle/code/cpp/ir/dataflow/internal/DataFlowUtil.qll
```diff
@@ -11,77 +11,87 @@ private import semmle.code.cpp.ir.IR
...
-    TStoreNodeInstr(Instruction i) { Ssa::explicitWrite(_, _, i) } or
-    TStoreNodeOperand(ArgumentOperand op) { Ssa::explicitWrite(_, _, op.getDef()) } or
-    TReadNode(Instruction i) { needsPostReadNode(i) } or
-    TSsaPhiNode(Ssa::PhiNode phi)
+    TPostFieldUpdateNode(FieldAddress operand, int indirectionIndex) {
+      indirectionIndex =
+        [0 .. Ssa::countIndirectionsForCppType(operand.getObjectAddress().getResultLanguageType()) -
+            1]
+    } or
+    TSsaPhiNode(Ssa::PhiNode phi) or
```
hvitved, line 44: So for non-phi definitions, you rely on the underlying instructions instead?

cpp/ql/lib/semmle/code/cpp/ir/dataflow/internal/DataFlowUtil.qll
```diff
@@ -226,260 +350,414 @@ class OperandNode extends Node, TOperandNode {
...
 
   /**
-   * Gets the `StoreNode` that computes the address used by this `StoreNode`.
+   * Holds if this phi node has input from the definition `input` (which is the `rnk`'th write
+   * operation in block `block`).
    */
-  abstract StoreNode getInner();
+  final predicate hasInputAtRankInBlock(IRBlock block, int rnk, Ssa::Definition input) {
+    Ssa::phiHasInputFromBlock(phi, input, _) and
+    input.definesAt(_, block, rnk)
```
hvitved, line 429: If this is from the shared SSA library, then it is not actually ranked, but instead the original index provided by `variableWrite`.

hvitved (commented):
I only did a light review of the SSA parts.

cpp/ql/lib/semmle/code/cpp/ir/dataflow/internal/DataFlowUtil.qll
MathiasVP: Correct

cpp/ql/lib/semmle/code/cpp/ir/dataflow/internal/DataFlowUtil.qll
```diff
@@ -754,158 +1039,43 @@ predicate localFlowStep = localFlowStepCached/2;
  * data flow. It may have less flow than the `localFlowStep` predicate.
  */
 predicate simpleLocalFlowStep(Node nodeFrom, Node nodeTo) {
+  // Post update node -> Node flow
+  Ssa::defUseFlow(nodeFrom.(PostUpdateNode).getPreUpdateNode(), nodeTo)
+  or
+  // Def-use flow
+  Ssa::defUseFlow(nodeFrom, nodeTo)
```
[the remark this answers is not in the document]
MathiasVP: Good point. Maybe it should be named `ssaFlow` (which is actually what the equivalent predicate is currently called on `main`)?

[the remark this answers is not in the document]
hvitved: Either that, or two separate predicates.

cpp/ql/lib/semmle/code/cpp/ir/dataflow/internal/SsaInternals.qll
MathiasVP: Well spotted! Initially, I used `phiHasInputFromBlock` (before learning about `lastRefRedef`) in `adjacentDefRead`, which led to missing flow, and I fixed it by using `lastRefRedef`. I imagine this call to `phiHasInputFromBlock` can also be replaced by `lastRefRedef`. Thanks!

cpp/ql/lib/semmle/code/cpp/ir/dataflow/internal/SsaInternalsCommon.qll
MathiasVP: Oh, yeah. They should definitely just be in a shared `cached` module.

cpp/ql/lib/semmle/code/cpp/ir/dataflow/internal/SsaInternals.qll
MathiasVP: Isn't this okay because we're also requiring that:
```
defOrUse1.asDefOrUse().hasIndexInBlock(bb1, i1, v) and
use.asDefOrUse().(UseImpl).hasIndexInBlock(bb2, i2, v)
```
(i.e., that they read the same source variable.)

Would you be more confident in the correctness if the `Ssa::Definition` column wasn't ignored, and instead checked the source variable of that definition?

cpp/ql/lib/semmle/code/cpp/ir/dataflow/internal/SsaInternals.qll
MathiasVP: Oh yes, it definitely should be.

cpp/ql/lib/semmle/code/cpp/ir/dataflow/internal/SsaInternalsCommon.qll
MathiasVP: Fixed in 3aa825b904a8d07dd07c0f7e88d9f3e57416b518.

cpp/ql/lib/semmle/code/cpp/ir/dataflow/internal/SsaInternals.qll
MathiasVP: Fixed in 471e08bc11a76fac1178e7826e351ee9d50f9d5c.

cpp/ql/lib/semmle/code/cpp/ir/dataflow/internal/SsaInternals.qll
MathiasVP: Fixed in b692aed874732e8c584dd761ec773cd826f73665. Indeed, `lastRefRedef` was the correct predicate to use. With that change, all of our uses of `phiHasInputFromBlock` disappeared 🎉.

cpp/ql/lib/semmle/code/cpp/ir/dataflow/internal/SsaInternals.qll
MathiasVP: Good point. Fixed in b692aed874732e8c584dd761ec773cd826f73665.

cpp/ql/lib/semmle/code/cpp/ir/dataflow/internal/DataFlowUtil.qll
MathiasVP: Oops. Thanks! Fixed in 53b58ac7d89097a3edcba22568d1ec606e203a7b.

cpp/ql/lib/semmle/code/cpp/ir/dataflow/internal/DataFlowUtil.qll
MathiasVP: Yeah, I think the parameter was badly named. In any case, b692aed874732e8c584dd761ec773cd826f73665 removed the predicate since it was dead code after resolving https://github.com/github/codeql/pull/10190#discussion_r960528476.

cpp/ql/lib/semmle/code/cpp/ir/dataflow/internal/DataFlowUtil.qll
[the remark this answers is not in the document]
MathiasVP: e1f7cb65b2f784dc06ce51d0c8d96533e4211466 renames the predicate. I went for that solution since it was the solution with the smallest diff 😂.

cpp/ql/lib/semmle/code/cpp/ir/dataflow/internal/SsaInternals.qll
MathiasVP: Yeah, that would also work. I'll think about whether that might be a more appropriate name. Thanks!

cpp/ql/lib/semmle/code/cpp/ir/dataflow/internal/SsaInternals.qll
hvitved: Ah yes, for some reason I failed to spot the `v`, so all should be good.

MathiasVP:
> Here's something to aspire to: Put qldoc on all predicates. See e.g. cpp/ql/lib/semmle/code/cpp/ir/dataflow/internal/SsaInternalsCommon.qll, which doesn't have even a single line of qldoc.

Thanks for the callout on this. I definitely intended to write QLDoc for the predicates in that file, but it somehow slipped my mind 🤦. 89cafeee54fa9d1eac9aa91feebb0ffb3e0550a8 adds the required QLDoc to that file.

MathiasVP:
Superseded by https://github.com/github/codeql/pull/10366.

PR #10191, opened 2022-08-26
Java: Add test regarding the type of an implicit `this` expression

smowton:


atorralba (dismissed):

smowton:
@atorralba @aschackmull this has a bunch of new changes compensating for implicit `this` calls now potentially having a callee declared on a specialised-type (e.g. `List<String>.add(String)`)

aschackmull (approved):
LGTM!

PR #1295, opened 2019-04-30
Typescript full mode fixes

asger-semmle:
Fixes some crashes in the 14 projects that started failing after TypeScript full mode became the default on LGTM.com this week.

Some issues that came from the TypeScript compiler itself:
- Some internal debug assertions failed, but went way after upgrading to 3.4.5.
- Calling `getProperties()` on the type `typeof globalThis | {}` caused it to crash. I've worked around it by catching the exception.
- Calling `typeSignatureToString` on [this type signature](https://github.com/GoogleChromeLabs/comlink/blob/master/dist/comlink.d.ts#L63) caused an out-of-memory error for some reason. I suspect this has something to do with the use of `infer` types, so I've disabled type extraction inside conditional types (which are where `infer` types are introduced).

Lastly, we didn't handle `export=` correctly when the right-hand side was an expression. I've fixed that.

I'll run an evaluation as well.

PR #10202, opened 2022-08-29
JS: Enable type extraction inside conditional types again

asgerf:
This re-enables type extraction inside conditional types, i.e. types of form `X extends Y ? A : B`, which are relevant for extracting MaD typings from `.d.ts` files.

This was originally disabled in https://github.com/github/codeql/pull/1295 to avoid divergence issues when extracting the type property graph. However, in https://github.com/github/codeql/pull/3781 we stopped extracting the type property graph entirely, so the previous mitigation should no longer be needed.

I verified that [this code](https://github.com/GoogleChromeLabs/comlink/blob/41eb437b877ce943b7e41636f7ce555b59c1d3a5/dist/comlink.d.ts#L63), which originally caused problems, still works after this change. [Evaluation](https://github.com/github/codeql-dca-main/issues/6483) also looks fine.

### DIFF
--- a/javascript/extractor/lib/typescript/src/ast_extractor.ts
+++ b/javascript/extractor/lib/typescript/src/ast_extractor.ts
@@ -168,20 +168,9 @@ export function augmentAst(ast: AugmentedSourceFile, code: string, project: Proj
         }
     }
 
-    // Number of conditional type expressions the visitor is currently inside.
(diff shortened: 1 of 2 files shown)


erik-krogh (dismissed):
👍

erik-krogh (approved):

PR #10203, opened 2022-08-29
SSA: Make shared library a parameterized module

hvitved:
This PR changes `SsaImplCommon.qll` to be a parameterized module. This means that we no longer need 4 copies for C#, but we still need a copy per language (I plan to change that in a follow-up PR).


MathiasVP (dismissed):
The changes LGTM! I'll delay the merge until @ginsbach (or someone else from the core team) has looked at the parameterized modules usage.

MathiasVP (approved):

alexet:
Are these predicates meant to be publically usable? If they are then they will disappear from the online docs.

hvitved:
> Are these predicates meant to be publically usable? If they are then they will disappear from the online docs.

Nope. Each language should provide its own public layer on top.

PR #10208, opened 2022-08-29
Java: A couple of small virtual dispatch fixes

aschackmull:
The first commit allows dispatch to methods defined on abstract classes for which we don't have any subclasses. Normally, ruling out abstract classes as the type to dispatch to is fine, as the dispatch targets are inherited by one or more subclasses and thus allowing the dispatch, but for abstract classes intended to be extended outside the analysed source this isn't the case, which left a few glaring holes in our dispatch relation.

The second commit fixes a minor bug in TypeFlow (included as separate commit for visibility, as the code is immediately refactored).

The third commit refactors the code a bit, and also includes upcasts to raw types and unbound parameterised types as interesting starting points for TypeFlow, thus improving our dispatch precision a little bit.


aschackmull:
Nothing to note from dca.

atorralba (approved):
Looks plausible to me.

PR #10209, opened 2022-08-29
QL: add query detecting consistent casing of names

erik-krogh:
E.g. having both `InstanceOfExpr` and `InstanceofExpr` is confusing.  The should use consistent casing.  

The query detects when the same kind of node (e.g. class, predicate, newtype) have different names only differ in casing.  

I changed `NodeName.qll` to treat more things as having the same "kind". 

You can see the new alerts here: https://github.com/github/codeql/security/code-scanning?query=pr%3A10209+tool%3ACodeQL+is%3Aopen+rule%3Aql%2Fnames-only-differing-by-case


tausbn (approved):
Looks reasonable to me. :+1:

erik-krogh:
> Looks reasonable to me. 👍

Great 😃  

You know what's coming next 😉

PR #10210, opened 2022-08-29
Swift: fix double parent

redsun82:
`RebindSelfInConstructorExpr` marking the `self` `VarDecl` as a child was causing a double parent problem.

The change in order in the `Callable` children is mainly cosmetic now, but will have more meaning after generating `PrintAst` (where we want `self` before the other params).


MathiasVP (approved):
Thanks!

MathiasVP (approved):

PR #10211, opened 2022-08-29
Swift: generate indexed parent/child relationship

redsun82:
This is the first step for closing https://github.com/github/codeql-c-team/issues/862, the second one being creating extendable `PrintAst.qll` and `PrintAst.ql` files using this generated information. Customization should probably happen at that level in hand-written code.

Commit by commit review is encouraged. The first changes the script and the related generated file (which was renamed from `GetImmediateParent.qll` to `ParentChild.qll`). The second introduces the other hand written changes required to fix library and tests. The third introduces more tests. The fourth makes the `Impl` module private in `ParentChild.qll`.


swift/ql/lib/codeql/swift/generated/ParentChild.qll
```diff
@@ -0,0 +1,2182 @@
+// generated by codegen/codegen.py
+import codeql.swift.elements
+
+module Impl {
```
MathiasVP: Could this module be `private`?

swift/ql/lib/codeql/swift/generated/ParentChild.qll
```diff
@@ -0,0 +1,2182 @@
...
+        or
+        index = n and result = e.(WhileStmt).getImmediateBody() and partialAccessor = "Body()"
+      )
+    )
+  }
+}
+
+/**
+ * Gets the "immediate" parent of `e`. "Immediate" means not taking into account node resolution: for example
+ * if `e` has conversions, `getImmediateParent(e)` will give the bottom conversion in the hidden AST.
```
MathiasVP, line 2156: Should I read "bottom conversion" as the outer-most conversion?

swift/ql/lib/codeql/swift/generated/ParentChild.qll
```diff
@@ -0,0 +1,2182 @@
...
+ */
+Element getImmediateParent(Element e) {
+  // `unique` is used here to tell the optimizer that there is in fact only one result
+  // this is tested by the `library-tests/parent/no_double_parents.ql` test
+  result = unique(Element x | e = Impl::getImmediateChild(x, _, _) | x)
+}
+
+/**
+ * Gets the immediate child indexed at `index`. Indexes are not guaranteed to be contiguous, but are guaranteed to be distinct. `accessor` is bound the the method giving the given child.
```
MathiasVP, line 2165: QL doesn't have methods 😄. Could we say "member-predicate name" instead?

swift/ql/lib/codeql/swift/generated/ParentChild.qll
```diff
@@ -0,0 +1,2182 @@
...
+ */
+Element getImmediateChildAndAccessor(Element e, int index, string accessor) {
+  exists(string partialAccessor |
+    result = Impl::getImmediateChild(e, index, partialAccessor) and
+    accessor = "getImmediate" + partialAccessor
+  )
+}
+
+/**
+ * Gets the child indexed at `index`. Indexes are not guaranteed to be contiguous, but are guaranteed to be distinct. `accessor` is bound the the method giving the given child. Node resolution is carried out.
```
MathiasVP, line 2175: Same here: "method" should probably be "member-predicate name".

swift/ql/lib/codeql/swift/generated/ParentChild.qll
redsun82: yeah, I wanted to, then noticed I didn't do it and fixed it with [3f4a330](https://github.com/github/codeql/pull/10211/commits/3f4a33065865ea257e845c52d0856a4e25eeea3d) :slightly_smiling_face:

swift/ql/lib/codeql/swift/generated/ParentChild.qll
redsun82: good point. Is it a member predicate call then? It's not just for example `getElement`, it's `getElement(3)`

swift/ql/lib/codeql/swift/generated/ParentChild.qll
redsun82: to prove the point, it's actually the innermost conversion. In C++ terms, for
```
(double)(int)x.i
```
* `getImmediateParent` called on the `.i` accessor would give uncoverted `x.i`
* `getImmediateParent` called on `x.i` would give the innermost `(int)x.i`

swift/ql/lib/codeql/swift/generated/ParentChild.qll
MathiasVP: Yeah, I think that sounds good 👌

swift/ql/lib/codeql/swift/generated/ParentChild.qll
```diff
@@ -0,0 +1,2182 @@
+// generated by codegen/codegen.py
+import codeql.swift.elements
+
+private module Impl {
+  int getMaximumChildrenIndexElement(Element e) {
+    e = e and
```
MathiasVP, line 6: Is this to fix a "unused" parameter warning? If so, we normally prefer writing `exists(e)` for this.

swift/ql/lib/codeql/swift/generated/ParentChild.qll
```diff
@@ -0,0 +1,2182 @@
...
+    result = 0
+  }
+
+  int getMaximumChildrenIndexCallable(Callable e) {
```
MathiasVP, line 10: Can all of these predicates be made private?

swift/ql/test/library-tests/parent/child.expected
```diff
@@ -0,0 +1,1319 @@
+| declarations.swift:1:8:1:8 | init | 0 | getImmediateSelfParam() | declarations.swift:1:8:1:8 | self |
+| declarations.swift:1:8:1:8 | init | 0 | getImmediateSelfParam() | declarations.swift:1:8:1:8 | self |
```
redsun82, line 2: These are actually two different `init` functions. Once this gets merged, I have another change ready to use functions default naming for constructors and distinguish the two (namely into `init()` and `init(x:)`)

swift/ql/lib/codeql/swift/generated/ParentChild.qll
redsun82: ah, good to know! in the end I've bypassed that and other warnings by just using `none()` for `Element` (we don't want global children :slightly_smiling_face: )

swift/ql/lib/codeql/swift/generated/ParentChild.qll
redsun82: sure, but they really cannot leak outside this file as they are in a private module, right?

MathiasVP: They can't leak outside the file, no. But if we mark them as `private` they also can't leak outside the module either. So that's even better 😄.

MathiasVP (approved):
LGTM!

PR #10213, opened 2022-08-29
put a limit on the length of the equivalent range in overly-large-range

erik-krogh:
The overly-large-range query would produce an extremely big alert message on some regexps from `natalie-lang/natalie`.  
This fixes it.  

I capped the range at 50 chars, because that broke no tests in the existing test-suite.

### DIFF
--- a/python/ql/lib/semmle/python/security/OverlyLargeRangeQuery.qll
+++ b/python/ql/lib/semmle/python/security/OverlyLargeRangeQuery.qll
@@ -238,8 +238,13 @@ module RangePrinter {
 
 /** Gets a char range that is overly large because of `reason`. */
 RegExpCharacterRange getABadRange(string reason, int priority) {
+  result instanceof OverlyWideRange and
   priority = 0 and
-  reason = "is equivalent to " + result.(OverlyWideRange).printEquivalent()
+  exists(string equiv | equiv = result.(OverlyWideRange).printEquivalent() |
+    if equiv.length() <= 50
+    then reason = "is equivalent to " + equiv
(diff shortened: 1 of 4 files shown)


python/ql/lib/semmle/python/security/OverlyLargeRangeQuery.qll
```diff
@@ -238,8 +238,13 @@ module RangePrinter {
 
 /** Gets a char range that is overly large because of `reason`. */
 RegExpCharacterRange getABadRange(string reason, int priority) {
+  result instanceof OverlyWideRange and
```
tausbn, line 241: I'm surprised that this `instanceof` check is needed. Isn't this implied by the inline cast on line 243? :thinking:

tausbn (approved):
Looks reasonable to me. I take it that constructing the long string (`equiv`) is not an issue, just displaying it.

python/ql/lib/semmle/python/security/OverlyLargeRangeQuery.qll
erik-krogh: You're right, that check is not needed, but I'm going this merge this PR as is anyway.  
The cast was needed in one of the first versions of my fix, but it's not needed with the version I ended up with.  

Maybe a QL-for-QL query in the style of `ql/superfluous-exists`?

erik-krogh:
The JavaScript language tests are failing, that is unrelated and will be fixed here: https://github.com/github/codeql/pull/9751 

I'm merging this PR now, and I'll cherry-pick the commit to a new PR against the 3.7 branch.

PR #10215, opened 2022-08-30
put a limit on the length of the equivalent range

erik-krogh:
Same as this PR: https://github.com/github/codeql/pull/10213. 

But against the `rc/3.7` branch.


aibaars (approved):

erik-krogh:
The Java tests are failing because we are using a semmle-code version that includes: https://github.com/github/semmle-code/pull/43400, but the `github/codeql` doesn't include https://github.com/github/codeql/pull/9816.  

The fix would be to include https://github.com/github/codeql/pull/9816 in the `rc/3.7` branch, but I'm not sure if we want to do that.   
I'm merging this PR for now.

PR #10216, opened 2022-08-30
SSA: Create a new `codeql/shared-ssa` library pack and move implementation there

hvitved:
~This PR creates a new QL pack for shared, language-agnostic code. I called it `codeql/shared-all` to be aligned with the names of library packs for our languages, but TBH I don't know why it is called `-all` and e.g. not `-lib`.~

Update: We decided to use separate packs for separate libraries, so this PR adds a new `codeql/shared-ssa` pack for the shared SSA library.

The new pack initially consists of the shared SSA implementation, which means that we now only have one (!) copy of that library, as opposed to one per language that uses it.

Commit-by-commit review is suggested.


shared/ssa/qlpack.yml
```diff
@@ -0,0 +1,4 @@
+name: codeql/shared-ssa
+version: 0.0.1
```
aeisenberg, line 2: No need to make a decision for this PR, but we may want to make the version `0.0.1-dev` like the other packs and this prevents accidental publishing. Our release process would strip the `-dev` publish the version and bump the version after publishing.

shared/README.md
```diff
@@ -0,0 +1,17 @@
...
+
+This folder contains shared, language-agnostic CodeQL libraries.
+
+Libraries are organized into separate query packs, in order to allow for
+individual versioning. For example, the shared static single assignment (SSA)
+library exists in the `codeql/shared-ssa` pack, which can be referenced by adding
+
+```
+dependencies:
+  codeql/shared-ssa: "*"
```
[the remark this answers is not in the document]
aeisenberg: I hope that with some tooling, we can make this easy to work with.

Maybe one possibility is that we always make sure all the core libraries depend on the latest version of the shared libraries and we can have a check that enforces this.

cpp/ql/lib/qlpack.yml
```diff
@@ -6,4 +6,4 @@ extractor: cpp
 library: true
 upgrades: upgrades
 dependencies:
-  codeql/shared-ssa: "*"
+  codeql/shared-ssa: 0.0.1-dev
```
aeisenberg, line 9: Here and below. I think we want the version without the `-dev` suffix since when this pack is published (and the shared packs are published), the packs won't have a `-dev` suffix. If we publish this pack with `-dev` in the version constraint, it can never be depended on by third parties since transitive dependencies can't be resolved.

```suggestion
  codeql/shared-ssa: 0.0.1
```

shared/ssa/qlpack.yml
```diff
@@ -1,4 +1,4 @@
 name: codeql/shared-ssa
-version: 0.0.1
+version: 0.0.1-dev
```
aeisenberg, line 2: This one should stay as it is, and before publishing `-dev` needs to be removed. This happens automatically for the core libraries, but will need to be manual for the shared libraries until we augment the existing automation.

shared/README.md
```diff
@@ -0,0 +1,17 @@
...
+library exists in the `codeql/shared-ssa` pack, which can be referenced by adding
+
+```
+dependencies:
+  codeql/shared-ssa: 0.0.1-dev
+```
+
+to `qlpack.yml`.
+
+All shared libraries will belong to a `codeql/shared-<name>` pack, and live in the
```
aibaars, line 16: I'm not convinced about putting the word `shared` all over the place. The whole point of packs and libraries is to share things, so there should be no need to state that in the name of a pack or package.

Why not simply use  `import codeql.<name>` for importing language-agnostic things? Language specific stuff can go into `codeql.<lang>` (or `semmle.code.<lang>`).

hvitved: Works for me.

.github/workflows/ruby-build.yml
```diff
@@ -95,6 +95,7 @@ jobs:
         uses: ./.github/actions/fetch-codeql
       - name: Build Query Pack
         run: |
+          codeql pack create ../shared/ssa --output target/packs
```
aibaars, line 98: Is this line necessary? I think the right thing would be to run run `codeql pack install ql/lib` instead. Although that probably requires `codeql/ssa` to be published previously, so there might be a bootstrapping problem.

@aeisenberg what do you think?

aibaars (approved):
This looks good to me. Just a small remark about the "ruby build" workflow.

MathiasVP (approved):
😍

.github/workflows/ruby-build.yml
hvitved: The job fails if I don't add the line.

aibaars: Let's leave it in for now. I guess things go wrong at the `codeql test run` step later in the workflow, right?

aibaars: Hmm, perhaps we should actually update the `qlpack.lock.yml` so it includes the new dependency. Perhaps as a follow-up once the shared pack has been published:
```
codeql pack install ql/lib --mode=update
A fatal error occurred: Package version failure. No compatible version found: codeql/ssa@>=0.0.1 <=0.0.1.
```

shared/ssa/qlpack.yml
```diff
@@ -1,4 +1,4 @@
-name: codeql/shared-ssa
+name: codeql/ssa
 version: 0.0.1-dev
```
aeisenberg, line 2: This will have to be manually changed to `0.0.1` before releasing and then manually bumped to `0.0.2` (or whatever) after releasing. We can implement automation for this in a similar way to how we automate the core libraries. 

For now, the steps should be:

1. remove the `-dev` suffix
2. commit the change
3. `codeql pack publish`
4. bump the version and add the `-dev` back
5. bump all dependencies in the libraries as well
6. commit the change

This simulates the automation will do.

aeisenberg (approved):
Looks good to me. Please take note of the manual release process.

PR #10218, opened 2022-08-30
Swift: make `ConstructorDecl`'s name include params

redsun82:
In order to distinguish overloads of the constructor and for consistency with other function calls, `ConstructorDecl` string representation uses the name which includes parentheses and parameter labels.

For consistency also the destructor got the same change, which means all `DestructorDecl`s will now show as `deinit()` rather than `deinit`.


MathiasVP (approved):
Indeed, this is much better!

PR #10220, opened 2022-08-30
print a correct range for ranges that doesn't contain any alpha-numeric chars

erik-krogh:


aschackmull (approved):

aibaars:
LGTM, should this also go into `rc/3.7` ?

aibaars (approved):

PR #10223, opened 2022-08-30
Java: New Android query to detect unsafe content URI resolution

atorralba:
Adds a query to detect uncontrolled data being used in `ContentProvider` methods that resolve URIs. Normally this is done to allow third party applications to provide URIs pointing to external files, but an attacker can provide a URI pointing to internal files or content providers of the victim application, potentially bypassing security controls and accessing sensitive information.

This query adds coverage for the following CVEs:

* CVE-2022-20144
* CVE-2021-0604
* CVE-2020-0460
* CVE-2018-9587

Also, 1ce3aef is there only for safe-keeping: it was a previous version of the query that restricted the sinks to be write operations, so that we made sure that the resolved file was used for something dangerous. But that left out other interesting cases, so in the end I removed that restriction and changed the severity to `warning` to reflect that the exploitation depends on how the file is used after the URI resolution.


yo-h:
> This branch has conflicts that must be resolved
> Conflicting files
> java/ql/src/Security/CWE/CWE-022/TaintedPathCommon.qll

atorralba:
> > This branch has conflicts that must be resolved
> > Conflicting files
> > java/ql/src/Security/CWE/CWE-022/TaintedPathCommon.qll

Fixed, thanks @yo-h! Although this will need another rebase once https://github.com/github/codeql/pull/10177 is merged.

java/ql/lib/semmle/code/java/security/UnsafeContentUriResolution.qll
```diff
@@ -0,0 +1,83 @@
...
+abstract class ContentUriResolutionSink extends DataFlow::Node { }
+
+/** A sanitizer for content URIs. */
+abstract class ContentUriResolutionSanitizer extends DataFlow::Node { }
+
+/**
+ * A unit class for adding additional taint steps to configurations related to
+ * content URI resolution vulnerabilities.
+ */
+abstract class ContentUriResolutionAdditionalTaintStep extends Unit {
```
aschackmull, line 17: It's a unit class, so it shouldn't be abstract.
```suggestion
class ContentUriResolutionAdditionalTaintStep extends Unit {
```

aschackmull (approved):
Looks plausible to me.

java/ql/src/Security/CWE/CWE-441/UnsafeContentUriResolution.qhelp
```diff
@@ -0,0 +1,49 @@
...
+  If possible, avoid using externally-provided data to determine URIs used by a <code>ContentResolver</code>.
+  If that is not an option, validate that the incoming URI can only reference trusted components, like an allow list
+  of content providers and/or applications, or alternatively make sure that the URI does not reference private
+  directories like <code>/data/</code>.
+</p>
+</recommendation>
+<example>
+<p>
+  This example shows two ways of opening a file using a <code>ContentResolver</code>. In the first case, externally-provided
+  data coming from an intent is directly used in the file-reading operation, allowing an attacker to provide a URI
```
felicitymay, line 33: Recommend splitting up this rather long sentence.

```suggestion
  data from an intent is used directly in the file-reading operation. This allows an attacker to provide a URI
```

java/ql/src/Security/CWE/CWE-441/UnsafeContentUriResolution.qhelp
```diff
@@ -0,0 +1,49 @@
...
+  of the form <code>/data/data/(vulnerable app package)/(private file)</code> to trick the application into reading it and
+  copying it to the external storage. In the second case, the URI is validated before being used, making sure it does not reference
```
felicitymay, line 35: It looks as if the example contains three cases, two `BAD` and one `GOOD`, so this might need some adjustment.

java/ql/src/Security/CWE/CWE-441/UnsafeContentUriResolution.java
```diff
@@ -0,0 +1,41 @@
...
+public class Example extends Activity {
+    public void onCreate() {
+        // BAD: Externally-provided URI directly used in content resolution
+        {
+            ContentResolver contentResolver = getContentResolver();
+            Uri uri = (Uri) getIntent().getParcelableExtra("URI_EXTRA");
+            InputStream is = contentResolver.openInputStream(uri);
+            copyToExternalCache(is);
+        }
+        // BAD: input Uri is not normalized, and check can be bypassed with ".." characters
```
felicitymay, line 13: Suggest using the same capitalization as in the previous comment:

```suggestion
        // BAD: input URI is not normalized, and check can be bypassed with ".." characters
```

java/ql/src/Security/CWE/CWE-441/UnsafeContentUriResolution.java
```diff
@@ -0,0 +1,41 @@
...
+        {
+            ContentResolver contentResolver = getContentResolver();
+            Uri uri = (Uri) getIntent().getParcelableExtra("URI_EXTRA");
+            String path = uri.getPath();
+            if (path.startsWith("/data"))
+                throw new SecurityException();
+            InputStream is = contentResolver.openInputStream(uri);
+            copyToExternalCache(is);
+        }
+        // GOOD: URI gets properly validated to avoid access to internal files
```
felicitymay, line 23:
```suggestion
        // GOOD: URI is properly validated to block access to internal files
```

felicitymay (commented):
@atorralba - thanks for writing a clear query help file and including examples ✨ 

I've added a couple of suggestions and a question on the docs.

I notice that [java/ql/test/query-tests/security/CWE-441/UnsafeContentUriResolutionTest.expected](https://github.com/github/codeql/pull/10223/files#diff-ed37b6f66a93b152b7676e264e092a52de80d489c0f5a84614a269a0dc0d4ed4) is an empty file. Is this intended?

atorralba:
Thanks for the review @felicitymay! 🙏 All comments addressed.

> I notice that [java/ql/test/query-tests/security/CWE-441/UnsafeContentUriResolutionTest.expected](https://github.com/github/codeql/pull/10223/files#diff-ed37b6f66a93b152b7676e264e092a52de80d489c0f5a84614a269a0dc0d4ed4) is an empty file. Is this intended?

Yes, this file is needed so that tests execute correctly, and it must be empty because the actual expectations are in the source code files themselves.

atorralba:
@aschackmull would you mind re-approving please?

felicitymay:
@atorralba - it looks as if I've missed a few typos, just having another look.

java/ql/src/Security/CWE/CWE-441/UnsafeContentUriResolution.qhelp
```diff
@@ -0,0 +1,50 @@
...
+<qhelp>
+<overview>
+<p>
+  When an Android application wants to access data in a content provider, it uses the <code>ContentResolver</code>
+  object. <code>ContentResolver</code>s communicate with an instance of a class that implements the 
+  <code>ContentProvider</code> interface via URIs with the <code>content://</code> scheme.
+
+  The authority part (the first path segment) of the URI passed as parameter to the <code>ContentResolver</code> 
+  determines which content provider is contacted for the operation. Specific operations that act on files also
+  support the <code>file://</code> scheme, in which case the local filesystem is queried instead.
```
felicitymay, line 13:
```suggestion
  The authority part (the first path segment) of the URI, passed as parameter to the <code>ContentResolver</code>, 
  determines which content provider is contacted for the operation. Specific operations that act on files also
  support the <code>file://</code> scheme, in which case the local filesystem is queried instead.
```

java/ql/src/Security/CWE/CWE-441/UnsafeContentUriResolution.qhelp
```diff
@@ -0,0 +1,50 @@
...
+
+  If an external component, like a malicious or compromised application, controls the URI that is used in a
+  <code>ContentResolver</code> operation, it can trick the vulnerable application into accessing its own private
+  files or non-exported content providers. Depending on what the vulnerable application does after accessing the file,
+  the attacking application might get access to the file by forcing it to be copied to a public directory like the
+  external storage, or tamper with it by making the application overwrite it with unexpected data.
```
felicitymay, line 19: The last sentence here is difficult to parse, suggest simplifying it.

```suggestion
  If an external component, like a malicious or compromised application, controls the URI for a
  <code>ContentResolver</code> operation, it can trick the vulnerable application into accessing its own private
  files or non-exported content providers. The attacking application might be able to get access to the file by forcing it to be copied to a public directory, like
  external storage, or tamper with the contents by making the application overwrite the file with unexpected data.
```

java/ql/src/Security/CWE/CWE-441/UnsafeContentUriResolution.qhelp
```diff
@@ -0,0 +1,50 @@
...
+</p>
+</overview>
+<recommendation>
+<p>
+  If possible, avoid using externally-provided data to determine URIs used by a <code>ContentResolver</code>.
```
felicitymay, line 24:
```suggestion
  If possible, avoid using externally-provided data to determine the URI for a <code>ContentResolver</code> to use.
```

felicitymay (dismissed):
Thanks for applying my earlier suggestions.

Apologies for missing these small things on my first pass. I think these additional suggested changes will make the help text easier to read.

atorralba:
Thanks for the thorough review @felicitymay! Suggestions applied.

felicitymay (approved):

PR #10226, opened 2022-08-30
Swift: add `PrintAst`

redsun82:
This includes integration with VSCode.


MathiasVP:
> We should maybe consider either removing `base` from the schema and add it as a shortcut on QL, or finish up replacing all `SelfApplyExpr` instances with synthesized types.

FWIW, I'm definitely in favor of replacing all `SelfApplyExpr`s with synthesized IPA branches.

swift/ql/lib/codeql/swift/printast/PrintAst.qll
```diff
@@ -0,0 +1,131 @@
...
+
+private predicate shouldPrint(Locatable e) { any(PrintAstConfiguration config).shouldPrint(e) }
+
+/**
+ * An AST node that should be printed.
+ */
+private newtype TPrintAstNode =
+  TLocatable(Locatable ast) {
+    // Only consider resolved nodes (that is not within the hidden conversion AST)
+    ast = ast.resolve()
```
MathiasVP, line 35: Hmm. Shouldn't this be:
```ql
ast = any(Locatable loc).resolve()
```
?

That is, I think _any_ element targeted by a call to `resolve` should be a possible PrintAST node.

MathiasVP (commented):
One question. LGTM otherwise.

swift/ql/lib/codeql/swift/printast/PrintAst.qll
redsun82: Hmm, I think they are equivalent, because resolve is idempotent.

MathiasVP: Ah, right!

MathiasVP (approved):
LGTM!

PR #10233, opened 2022-08-31
Swift: show conversions in `PrintAst`

redsun82:
While the uncoverted node is still shown as a direct child, its conversions are listed in a new `conversions` virtual child.

Moreover, `VarDecl` (including `ParamDecl`) instance show their type as a property, and so do `AbstractFunctionDecl`.

This also showcases the customizability of the `PrintAst` infrastructure resting on generated code.


swift/ql/lib/codeql/swift/printast/PrintAstNode.qll
```diff
@@ -74,13 +80,86 @@ class PrintLocatable extends PrintAstNode, TLocatable {
...
+
+  PrintConversionContainer() { this = TConversionContainer(convertee) }
+
+  override string toString() { result = "" }
+
+  final override predicate shouldBePrinted() { shouldPrint(convertee) }
+
+  override predicate hasChild(PrintAstNode child, int index, string accessor) {
+    child = TConversion(convertee.getConversion(index)) and
+    accessor = "getConversion(" + index + ")"
```
MathiasVP, line 125: I think there's a usability question to be resolved here.
The `accessor` string is meant to inform users about which predicate they should be calling to get the corresponding AST element, and I'm not sure we want to advocate calls to `getConversion(1)`, `getConversion(2)`, ... `getConversion(n)`.

Instead, I think we should have a `getFullyConverted()` child under which the entire AST is laid out (including hidden AST nodes and whatever). IIRC, this is what we did for C/C++.

Do you think that could work for Swift? And would it be better?

MathiasVP (commented):
The code LGTM, but I'm wondering if this is how we want the conversions to be laid out in the viewer.

swift/ql/lib/codeql/swift/printast/PrintAstNode.qll
redsun82: hmm, as far as I understand the PrintAst on C++ only shows one conversion, which I found quite confusing when multiple convertions are present.

For more similarity to the C++ case, we could maybe go with a repeated `Conversion` label, starting with the `getFullyConverted` one, like (I wonder how will that work, and also if there is a `semmle.order` thing to do on properties as well to get that)
```
#  155|       getInit(0): [CallExpr] call to init()
#  155|           Conversion: [InjectIntoOptionalExpr] (Baz?) ...
#  155|           Conversion: [CoerceExpr] (Baz) ...
#  155|           Conversion: [DerivedToBaseExpr] (Baz) ...
#  155|         getFunction(): [ConstructorRefCallExpr] call to init()
#  155|           getFunction(): [DeclRefExpr] init()
#  155|           getArgument(0): [Argument] : Derived.Type
#  155|             getExpr(): [TypeExpr] Derived.Type
#  155|               getTypeRepr(): [TypeRepr] Derived
```
When there is one conversion this would look exactly like C++

jketema: That's an incorrect observation regarding C++. You'll get nested conversions:
```
# 1685|           getExpr().getFullyConverted(): [ReferenceToExpr] (reference to)
# 1685|               Type = [LValueReferenceType] const CapturedLambdaMyObj &
# 1685|               ValueCategory = prvalue
# 1685|             getExpr(): [CStyleCast] (const CapturedLambdaMyObj)...
# 1685|                 Conversion = [GlvalueConversion] glvalue conversion
# 1685|                 Type = [SpecifiedType] const CapturedLambdaMyObj
# 1685|                 ValueCategory = lvalue
# 1685|               getExpr(): [TemporaryObjectExpr] temporary object
# 1685|                   Type = [Class] CapturedLambdaMyObj
# 1685|                   ValueCategory = lvalue
```

MathiasVP: Yeah, I think I like the C/C++ visualization better since it's clearer which predicate should be called to get to the conversion of interest through `getFullyConverted`.

With that said, if you think this is difficult to model in the current Swift setup then I'm perfectly happy with leaving that reordering as future work as this PR is clearly better than no PrintAST support for conversions 😂.

redsun82: ah, ok I copmletely misunderstood the C++ format then. I can try reproduce that.

redsun82: it's done. If we fix the fact that contrary to C++ the getters in the hidden AST also resolve nodes in Swift, we can get rid of the `Immediate`, but I'd do it in another PR

swift/ql/lib/codeql/swift/elements/expr/Expr.qll
```diff
@@ -7,6 +7,12 @@ class Expr extends ExprBase {
 
   Expr getConversion() { result.convertsFrom(this) }
 
+  Expr getConversion(int n) {
+    n = 0 and result = this.getConversion()
+    or
+    result = this.getConversion(n - 1).getConversion()
+  }
+
```
MathiasVP, line 15: I suppose this predicate is unused now, but maybe it'll be a useful thing to have in the future. If so, we should maybe add a `Expr.getAConversion()` predicate as well in the future.

MathiasVP (approved):
LGTM!

swift/ql/lib/codeql/swift/printast/PrintAstNode.qll
MathiasVP: Sounds good to me!

PR #10234, opened 2022-08-31
JS: Make API graph more sparse for escaping class instance members

asgerf:
Fixes a timeout in the upgrade caused by the N^2 behaviour when a class with N members has an instance that escapes at N different def-nodes.

Previously, each def-node would have edges to each of the class's members, so that patterns like this would work:
```ts
class C {
  foo(x) {} // 'x' found by API::moduleImport("testlib").getParameter(0).getMember("foo").getParameter(0)
  bar() {}
  baz() {}
}
const c = new C();
c.field = x;
require('testlib')(c);
```

Now, instead of adding edges from the `c` def-node, we mark the `MkClassInstance` API node as another def-node for that argument. That is, when you do `API::moduleImport("testlib").getParameter(0)`, you now get two def-nodes: `MkDef(c)` for the argument and `MkClassInstance(C)` for the class instance. The `MkDef` node will have the edges corresponding to properties directly added on the object, while the `MkClassInstance` node has the edges corresponding to the members of that class.

### Before
```mermaid
graph LR
  testlib["API::moduleImport(testlib)"]
  testlib -- "Parameter[0]" --> c
  c["MkDef(c)"]
  foo["MkDef(foo)"]
  bar["MkDef(bar)"]
  baz["MkDef(baz)"]
  x["MkDef(x)"]
  c -- "Member[field]" --> x
  c -- "Member[foo]" --> foo
  c -- "Member[bar]" --> bar
  c -- "Member[baz]" --> baz
```

### After
```mermaid
graph LR
  testlib["API::moduleImport(testlib)"]
  testlib -- "Parameter[0]" --> c
  c["MkDef(c)"]
  foo["MkDef(foo)"]
  bar["MkDef(bar)"]
  baz["MkDef(baz)"]
  x["MkDef(x)"]
  c -- "Member[field]" --> x
  cls -- "Member[foo]" --> foo
  cls -- "Member[bar]" --> bar
  cls -- "Member[baz]" --> baz
  cls["MkClassInstance(C)"]
  testlib -- "Parameter[0]" --> cls
```

This means when a class has instances that escapes at many def-nodes, the edges for instance members are shared.

[Evaluation](https://github.com/github/codeql-dca-main/issues/6576) looks quiet. I've confirmed that it fixes some of the timeouts observed.

### DIFF
--- a/javascript/ql/lib/semmle/javascript/ApiGraphs.qll
+++ b/javascript/ql/lib/semmle/javascript/ApiGraphs.qll
@@ -663,7 +663,14 @@ module API {
         or
         any(Type t).hasUnderlyingType(m, _)
       } or
-      MkClassInstance(DataFlow::ClassNode cls) { cls = trackDefNode(_) and hasSemantics(cls) } or
+      MkClassInstance(DataFlow::ClassNode cls) {
+        hasSemantics(cls) and
+        (
+          cls = trackDefNode(_)
+          or
+          cls.getAnInstanceReference() = trackDefNode(_)
+        )
+      } or
       MkAsyncFuncResult(DataFlow::FunctionNode f) {
         f = trackDefNode(_) and f.getFunction().isAsync() and hasSemantics(f)
       } or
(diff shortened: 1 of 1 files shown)


erik-krogh (approved):
I love it ❤️ 

But I think there's a typo in your example, the `x` in `x.field` is not declared anywhere.

asgerf:
> But I think there's a typo in your example, the x in x.field is not declared anywhere.

Right, that was supposed to say `c.field = x;`

PR #10235, opened 2022-08-31
C#: Remove legacy tracer support

hvitved:
This PR removes the legacy tracer specifications, and moves the responsibility of disabling shared compilation (and pre-compilation of ASP.NET MVC code) from the auto-builder to the tracer, which means that it will also work with custom build commands.

Commit-by-commit review is suggested.


michaelnebel (approved):
Looks plausible to me!

felickz:
@hvitved awesome fix!  Missing the shared compilation flag is not always obvious as it just creates a small DB w/o errors.  Just [hit this this week in a POC]( https://github.slack.com/archives/CNG82545A/p1661776759536149?thread_ts=1661776759.536149&cid=CNG82545A )!!

PR #10237, opened 2022-08-31
Swift: add generic type parameters to AST children

redsun82:
This required a fix in codegen, as order of bases as written in `schema.yml` was not preserved, so the order of children coming from different bases was off in the AST printer. This implied changing the type of `bases` in `schema.Class` from `Set` to `List`, and removing sorting of `bases` in `ql.Class`.

Commit by commit review is encouraged.

### DIFF
--- a/swift/ql/test/extractor-tests/declarations/declarations.swift
+++ b/swift/ql/test/extractor-tests/declarations/declarations.swift
@@ -143,3 +143,8 @@ extension Int {
 
 42.id()
 
+class GenericClass<A, B: Baz, C: MyProtocol> {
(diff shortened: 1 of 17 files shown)


swift/ql/test/extractor-tests/declarations/declarations.swift
```diff
@@ -1,3 +1,4 @@
+
```
redsun82, line 1: whoops, this newline is unintended, and changes the test results too much, I will revert this

MathiasVP (approved):
LGTM!

PR #10238, opened 2022-08-31
C#: Theorems for Free - Model generation

michaelnebel:
In this PR we introduce a new model generation query for summary flow models based on the *Theorems for Free* approach. 
Based on the type information of a callable we derive a summary that we assume is sensible.
The idea is that for polymorphic methods (generics) there is a limit to what the method can do with a value of the generic type as the generic has to be valid for all types (unless there is a type constraint/restriction).

We implement the following rules to generate summaries based on the type of a generic method `f` where `f` is a member of class parameterized over the type variable `T`.

Some examples are
- `f : T -> unit`: Assumed to be a setter, that is, the value of type T is assigned to a field/property. This means that taint is propagated from the argument into the a field in the object.
- `f : unit -> T`: Assumed to be a getter, that is, the value of type T is fetched from a field/property on the object. This means that taint is propagated from the field on the object to the return.
- `f : (T -> S) -> S`: Assumed to be an apply like function that takes the value of type T stored in the object and apply the provided function and return the generated value. That is, taint is propagated from the field/property on the object to the parameter of the supplied function and taint is propagated from the return of the supplied function to the return of `f`.

The above rules are generalised to functions with multiple arguments and collection like types, but the above illustrates the principles.

Applying the query to .NET Standard library yields approximately 4k results. Below are a couple of screenshots with produced summaries, which look accurate.
image
image

Lots of testcases has been included
- The classes `TypeBased.*` are add hoc test cases.
- The `SystemLinqEnumerable` class contains a large subset of all the LINQ extension methods in the .NET runtime framework. Methods where correct summaries are not produced are commented out.


csharp/ql/test/utils/model-generator/theorems/TheoremSummaries.cs
```diff
@@ -0,0 +1,49 @@
...
+
+    public int Apply(Func<T, int> f) { throw null; }
+}
+
+public class Theorems2 {
+
+    public T Transform<T>(T x) { throw null; }
+}
+
+public class CollectionTheorems1<T> : IEnumerable<T> {
```
aschackmull, line 42: If it extends a known interface like this, which already has a known relation between its type variable and the way it's represented as contents, then we'd like the generated summaries to be compatible with the existing summaries. I.e. we'd like
```
var c = new CollectionTheorems1<String>();
c.Add(source());
sink(c.Last());
```
to have flow. Therefore we'll need the generated summary to just store in `Element` - not a freshly minted synthetic field.

csharp/ql/test/utils/model-generator/theorems/CaptureTheoremsForFreeSummaryModels.expected
```diff
@@ -0,0 +1,14 @@
+| Summaries;CollectionTheorems1<>;false;Add;(T);;Argument[0];Argument[this].SyntheticField[ArgT].Element;taint;generated |
```
aschackmull, line 1: I think all of these generated steps should be value steps rather than taint steps.

csharp/ql/test/utils/model-generator/theorems/CaptureTheoremsForFreeSummaryModels.expected
```diff
@@ -0,0 +1,14 @@
...
+| Summaries;CollectionTheorems1<>;false;First;();;Argument[this].SyntheticField[ArgT].Element;ReturnValue;taint;generated |
+| Summaries;Theorems1<>;false;Apply;(System.Func<T,System.Int32>);;Argument[this].SyntheticField[ArgT];Argument[0].Parameter[0];taint;generated |
+| Summaries;Theorems1<>;false;Get;();;Argument[this].SyntheticField[ArgT];ReturnValue;taint;generated |
+| Summaries;Theorems1<>;false;Get;(System.Object);;Argument[this].SyntheticField[ArgT];ReturnValue;taint;generated |
+| Summaries;Theorems1<>;false;Id;(T);;Argument[0];ReturnValue;taint;generated |
```
aschackmull, line 6: I don't think this is the summary we want. I'd expect a `T -> T` method on a `Type<T>` to be a simultaneous getter and setter, i.e. something like "replace" or "addAndReturnExisting". The Java `V Map::put(K, V)` method e.g. returns an existing `MapValue` while also adding a `MapKey` and a `MapValue`.

csharp/ql/test/utils/model-generator/theorems/TheoremSummaries.cs
```diff
@@ -0,0 +1,49 @@
...
+
+    // No summary as S is unrelated to T
+    public void Set<S>(S x) { throw null; }
+
+    public int Apply(Func<T, int> f) { throw null; }
+}
+
+public class Theorems2 {
+
+    public T Transform<T>(T x) { throw null; }
```
aschackmull, line 39: I'm not sure this is a case we'll want/need to cover.

csharp/ql/test/utils/model-generator/theorems/TheoremSummaries.cs
```diff
@@ -0,0 +1,49 @@
+using System;
+using System.Linq;
+using System.Collections;
+using System.Collections.Generic;
+
+namespace Summaries;
+
+public class Theorems1<T> {
```
aschackmull, line 8: We'll also want test cases for `void addMany(IList<T> xs)` and `IList<T> getMany()`. These should result in
```
Arg[0].Element -> Arg[this].SynthField[T]
Arg[this].SynthField[T] -> ReturnValue.Element
```
respectively.

csharp/ql/test/utils/model-generator/theorems/TheoremSummaries.cs
```diff
@@ -0,0 +1,49 @@
...
+
+    public void Set(T x) { throw null; }
+
+    public void Set(int x, T y) { throw null; }
+
+    // No summary as S is unrelated to T
+    public void Set<S>(S x) { throw null; }
+
+    public int Apply(Func<T, int> f) { throw null; }
+}
```
aschackmull, line 35: Also, I'd like the test case
```
Theorems1<S> FlatMap<S>(Func<T, IEnumerable<S>> f)
```
for which I'd expect two summaries:
```
Arg[this].SynthField[T] -> Arg[0].Param[0]
Arg[0].ReturnValue.Element -> ReturnValue.SynthField[T]
```
It's crucial here that the two type variables are different. Had they both been `T` then we probably also ought to get
```
Arg[0].ReturnValue.Element -> Arg[this].SynthField[T]
```

csharp/ql/test/utils/model-generator/theorems/CaptureTheoremsForFreeSummaryModels.expected
michaelnebel: Makes sense. In general this must mean that our assumptions is that `T -> S` is a setter no matter the type of `S`: If `S = T` then your reasoning holds and if S != T (eg. S=unit or S is just something unrelated) then the only thing we can do with the value of type `T` is to set it (at least as long as there are no type constraints on `T`). The dual reasoning can be performed for signatures like `S -> T` (these must then be getter like).

csharp/ql/test/utils/model-generator/theorems/TheoremSummaries.cs
michaelnebel: I will remove it, but I am not entirely sure that it is not needed.
If something has a type signature like this, then it could likely be the case that there are some type constraints on T that makes it reasonable to assume that flow is propagated from argument to return - either because a new value of type T is generated based on the input or that the existing value of T is manipulated by changing its state. One could imagine a generic with a type constraint `T : I`, where I is an interface with members that has side effects on the value of T when being called.

aschackmull: But if `T` is constrained in that way, then I'm not sure we'll want the theorems-for-free approach at all - the theoretical foundation certainly becomes much weaker.

michaelnebel: Alright. Let's leave it out then :-)

csharp/ql/test/utils/model-generator/theorems/TheoremSummaries.cs
michaelnebel: Yes, you are right!

csharp/ql/test/utils/model-generator/theorems/TheoremSummaries.cs
michaelnebel: I will add these, but I suppose the more realistic versions are `void AddMany(IList<T> xs)` and `IList<T> GetMany()` in `CollectionTheorems<T>`, where they will get summaries like
```
AddMany: Argument[0].Element -> Argument[this].Element
GetMany: Argument[this].Element -> ReturnValue.Element
```

aschackmull: Yes. Might as well add both variants as tests.

csharp/ql/test/utils/model-generator/theorems/TheoremSummaries.cs
michaelnebel: I intentionally only made rudimentary support for higher order methods as I would like to do that as a follow up.

michaelnebel:
@hvitved @aschackmull : Just a naming nitpick; Would it be better that we just name logic like this *Type based summaries* and just say in the description say that this is based on *Theorems for Free*?

michaelnebel:
Finally DCA looks good:

image

csharp/ql/src/utils/model-generator/internal/CaptureTypeBasedSummaryModels.qll
```diff
@@ -0,0 +1,231 @@
...
+    callable.getDeclaringType() = return and unboundGeneric(return, tp)
+  ) and
+  result = getSyntheticField(tp)
+}
+
+/**
+ * Holds if `input` is a models as data string representation of, how a value of type `tp`
+ * (or a generic parameterized over `tp`) can be generated by a delegate parameter of `callable`.
+ */
+private predicate source(DotNet::Callable callable, string input, TypeParameter tp) {
```
hvitved, line 123: Perhaps `delegateSource` is a better name.

csharp/ql/src/utils/model-generator/internal/CaptureTypeBasedSummaryModels.qll
```diff
@@ -0,0 +1,231 @@
...
+    output = "ReturnValue" + access
+  )
+}
+
+/**
+ * Holds if `callable` has a delegate parameter that accepts a value of type `tp`
+ * and `output` is the models as data string representation of, how data is routed to
+ * the delegate parameter.
+ */
+private predicate sink(DotNet::Callable callable, TypeParameter tp, string output) {
```
hvitved, line 165: `delegateSink`

csharp/ql/src/utils/model-generator/internal/CaptureTypeBasedSummaryModels.qll
```diff
@@ -0,0 +1,231 @@
...
+  exists(DelegateType dt, int position, Type t, Parameter p |
+    delegate(callable, dt, position) and
+    p = dt.getAParameter() and
+    t = p.getType() and
+    t = tp and
```
hvitved, line 170: Remove the variable `t`.

csharp/ql/src/utils/model-generator/internal/CaptureTypeBasedSummaryModels.qll
```diff
@@ -0,0 +1,231 @@
...
+ * on the Theorems for Free approach.
+ */
+class TypeBasedFlowTargetApi extends Specific::TargetApiSpecific {
+  TypeBasedFlowTargetApi() { Specific::isRelevantForTypeBasedFlowModels(this) }
+
+  /**
+   * Gets the string representation of all type based summaries inspired by
+   * the Theorems for Free approach.
+   *
+   * Basic example signatures could be
```
hvitved, line 202: I don't quite understand this; I think it would be better with concrete C# examples.

csharp/ql/src/utils/model-generator/internal/CaptureTypeBasedSummaryModels.qll
```diff
@@ -0,0 +1,231 @@
...
+   * Important special cases are \alpha = unit (setter),
+   * \alpha = T (getter, setter and id) and \beta = unit (getter).
+   *
+   * Complex example signatures could be
+   * this : (T -> S) -> S
+   * this : S1 x (S1 -> S2) -> S2
+   * where T is type parameter of the class declaring `this` and S, S1 and S2 are type parameters
+   * of `this`.
+   */
+  string getSummaries() {
```
hvitved, line 216: What if we have
```csharp
void M<T>(Func<T,T> f)
```
do we then get a flow summary that passes the output of calling `f` into `f` itself? I would think we shouldn't generate such a summary.

hvitved (commented):
Looks great (and surprisingly simple). A couple of minor comments.

csharp/ql/src/utils/model-generator/internal/CaptureTypeBasedSummaryModels.qll
michaelnebel: Yes, we would get a summary for that. 
Do you dislike summaries for the example above because (1) there is no other source that could *generate* a value of type `T` or (2) we don't expect fix-point calculations 😂 ?
In the case you mention here it might not make sense, but we do need to generate summaries like this under some circumstances (think of `fold` or in terms of C# `Aggregate`).

csharp/ql/src/utils/model-generator/internal/CaptureTypeBasedSummaryModels.qll
```diff
@@ -0,0 +1,231 @@
...
+}
+
+/**
+ * Gets models as data input/output access relative to the type parameter `tp` in the
+ * type `t` in the scope of `callable`.
+ *
+ * Note: This predicate has to be inlined as `callable` is not related to `return` or `tp`
+ * in every disjunction.
+ */
+pragma[inline]
```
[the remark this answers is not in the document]
michaelnebel: Ok, will that also have the effect of inlining the predicate? If it is not inline, we might get a large tuple count for every callable as it is not related to the other predicate parameters in all disjunkts.

csharp/ql/src/utils/model-generator/internal/CaptureTypeBasedSummaryModels.qll
michaelnebel: Yes, that is a good idea!

csharp/ql/src/utils/model-generator/internal/CaptureTypeBasedSummaryModels.qll
```diff
@@ -196,22 +185,24 @@ class TypeBasedFlowTargetApi extends Specific::TargetApiSpecific {
...
+   * Examples could be (see C# psuedo code below)
+   * (1) `Get` returns a value of type `T`. We assume that the returned
+   *     value was fetched from a (synthetic) field.
+   * (2) `Set` consumes a value of type `T`. We assume that the value is stored in
+   *     a (synthetic) field.
+   * (3) `Apply<S>` is assumed to apply the provided function to a value stored in
+   *     a (synthetic) field and return the result.
+   * (4) `Apply<S1,S2>` is assumed to apply the provided function to provided value
+   *     and return the result.
+   * public class MyGeneric<T> {
```
hvitved, line 200: Enclose in ```csharp

csharp/ql/src/utils/model-generator/internal/CaptureTypeBasedSummaryModels.qll
hvitved: Ah, yes, it is indeed needed for `Aggregate`.

hvitved (approved):

PR #10239, opened 2022-08-31
Kotlin: Fix declaration stack

tamasvajk:


java/kotlin-extractor/src/main/kotlin/KotlinFileExtractor.kt
```diff
@@ -4702,6 +4708,30 @@ open class KotlinFileExtractor(
...
+        fun peek() = stack.peek()
+
+        fun peekAsDeclarationParent(): IrDeclarationParent {
+            val trapWriter = tw
+            if (isEmpty() && trapWriter is SourceFileTrapWriter) {
+                // If the current declaration is used as a parent, we might end up with an empty stack. In this case, the source file is the parent.
+                return trapWriter.irFile
+            }
+
+            return peek() as IrDeclarationParent
```
igfoo, line 4732: Can this cast ever fail?

java/kotlin-extractor/src/main/kotlin/KotlinFileExtractor.kt
```diff
@@ -4681,7 +4687,7 @@ open class KotlinFileExtractor(
         addVisibilityModifierToLocalOrAnonymousClass(id)
         extractClassSupertypes(superTypes, listOf(), id, inReceiverContext = true)
 
-        extractEnclosingClass(currentDeclaration, id, locId, listOf())
+        extractEnclosingClass(declarationParent, id, null, locId, listOf())
```
igfoo, line 4690: Should the `null` be `currentDeclaration as? IrClass`?

java/kotlin-extractor/src/main/kotlin/KotlinFileExtractor.kt
```diff
@@ -551,10 +555,15 @@ open class KotlinFileExtractor(
         }
     }
 
-    // If `parentClassTypeArguments` is null, the parent class is a raw type.
-    private fun extractEnclosingClass(innerDeclaration: IrDeclaration, innerId: Label<out DbClassorinterface>, innerLocId: Label<DbLocation>, parentClassTypeArguments: List<IrTypeArgument>?) {
-        with("enclosing class", innerDeclaration) {
-            var parent: IrDeclarationParent? = innerDeclaration.parent
+    private fun extractEnclosingClass(
```
igfoo, line 558: It's not clear to me exactly what this function is meant to do. Could you add a doc comment for the function as a whole, please?

java/kotlin-extractor/src/main/kotlin/KotlinFileExtractor.kt
tamasvajk: I think you mean `elementToReportOn as? IrClass`, but that will always be `null`, because the element is either a function reference, a property reference, a local function or an `IrTypeOperatorCall`.

java/kotlin-extractor/src/main/kotlin/KotlinFileExtractor.kt
tamasvajk: Added a comment. Maybe we could change the function name to `processEnclosedClass` so that one wouldn't expect to only extract the `enclInReftype` relation. What do you think?

java/kotlin-extractor/src/main/kotlin/KotlinFileExtractor.kt
tamasvajk: Pushed https://github.com/github/codeql/pull/10239/commits/2728517c7f6d8deda20c8206ec27c38951662661 to handle the error.

PR #10241, opened 2022-08-31
Java: Add query for WebView debugging enabled

joefarebrother:


java/ql/lib/semmle/code/java/security/WebviewDubuggingEnabledQuery.qll
```diff
@@ -0,0 +1,47 @@
...
+
+  override predicate isSink(DataFlow::Node node) {
+    exists(MethodAccess ma |
+      ma.getMethod().hasQualifiedName("android.webkit", "WebView", "setWebContentsDebuggingEnabled") and
+      node.asExpr() = ma.getArgument(0)
+    )
+  }
+
+  override predicate isBarrier(DataFlow::Node node) {
+    not node.getType() instanceof BooleanType
```
atorralba, line 41: Wouldn't this be already implied by the fact this is a `DataFlow::Configuration` and the source is a boolean? My concern is that we will have a gigantic `isBarrier` predicate and impact performance, so I think we should remove this line.

java/ql/src/Security/CWE/CWE-489/WebviewDebuggingEnabled.qhelp
```diff
@@ -0,0 +1,38 @@
+<!DOCTYPE qhelp PUBLIC
+  "-//Semmle//qhelp//EN"
+  "qhelp.dtd">
+<qhelp>
+
+<overview>
+<p>The <code>WebView.setWebContentsDebuggingEnabled</code> method enables or disables the contents of any <code>WebView</code> in the application to be debugged.</p>
+
+<p>Enabling debugging featues could allow for additional entry points or leaking sensitive information.
```
atorralba, line 9:
```suggestion
<p>Enabling debugging features could allow for additional entry points or leaking sensitive information. 
```

java/ql/src/Security/CWE/CWE-489/WebviewDebuggingEnabled.qhelp
```diff
@@ -0,0 +1,38 @@
+<!DOCTYPE qhelp PUBLIC
+  "-//Semmle//qhelp//EN"
+  "qhelp.dtd">
+<qhelp>
+
+<overview>
+<p>The <code>WebView.setWebContentsDebuggingEnabled</code> method enables or disables the contents of any <code>WebView</code> in the application to be debugged.</p>
+
+<p>Enabling debugging featues could allow for additional entry points or leaking sensitive information. 
+As such, debugging should only be anabled during development, and disabled during production builds.</p>
```
atorralba, line 10:
```suggestion
As such, debugging should only be enabled during development, and disabled in production builds.</p>
```

java/ql/src/Security/CWE/CWE-489/WebviewDebuggingEnabled.qhelp
```diff
@@ -0,0 +1,38 @@
...
+</overview>
+<recommendation>
+<p>Ensure that debugging features are not enabled during production builds.
```
atorralba, line 13:
```suggestion
<p>Ensure that debugging features are not enabled in production builds. 
```

java/ql/src/Security/CWE/CWE-489/WebviewDebuggingEnabled.qhelp
```diff
@@ -0,0 +1,38 @@
...
+As such, debugging should only be anabled during development, and disabled during production builds.</p>
+</overview>
+<recommendation>
+<p>Ensure that debugging features are not enabled during production builds. 
+If <code>WebView.setWebContentsDebuggingEnabled(true)</code> is used, ensure that it is guarded by a flag indicating that this is a debug build.</p>
+
+</recommendation>
+<example>
+
+<p>In the code below, the BAD case shows debugging always being enabled,
```
atorralba, line 19:
```suggestion
<p>In the code below, the BAD case shows WebView debugging always being enabled, 
```

java/ql/src/Security/CWE/CWE-489/WebviewDebuggingEnabled.qhelp
```diff
@@ -0,0 +1,38 @@
...
+</overview>
+<recommendation>
+<p>Ensure that debugging features are not enabled during production builds. 
+If <code>WebView.setWebContentsDebuggingEnabled(true)</code> is used, ensure that it is guarded by a flag indicating that this is a debug build.</p>
+
+</recommendation>
+<example>
+
+<p>In the code below, the BAD case shows debugging always being enabled, 
+whereas the GOOD case only enables debugging if the <code>android:debuggable</code> attribute is set to <code>true</code>.</p>
```
atorralba, line 20:
```suggestion
whereas the GOOD case only enables it if the <code>android:debuggable</code> attribute is set to <code>true</code>.</p>
```

java/ql/lib/semmle/code/java/security/WebviewDubuggingEnabledQuery.qll
```diff
@@ -0,0 +1,47 @@
...
+import semmle.code.java.security.SecurityTests
+
+/** Holds if `ex` looks like a check that this is a debug build. */
+private predicate isDebugCheck(Expr ex) {
+  exists(Expr subex, string debug |
+    (
+      debug.toLowerCase().matches("%debug%")
+      or
+      debug.toLowerCase().matches("%test%")
+    ) and
```
atorralba, line 15: Can be simplified:

```suggestion
    debug.toLowerCase().matches(["%debug%", "%test%"]) and
```

atorralba (commented):
Hey @joefarebrother, I added some inline comments.

java/ql/lib/semmle/code/java/security/WebviewDubuggingEnabledQuery.qll
joefarebrother: I was thining about the potential for flow to reach a nodes of more general types (e.g. to `Boolean` via implicit boxing, and then to `Object`) and then have a lot of flow due to there being a lot of sources. Can check performance either way.

martin389:
Hello from this week's docs first responder 👋 Thanks for creating this PR! I've added it to our review board for another writer to look at. They'll be able to advise on any additional considerations for this change.

java/ql/src/Security/CWE/CWE-489/WebviewDebuggingEnabled.ql
```diff
@@ -0,0 +1,19 @@
+/**
+ * @name Android Webview debugging enabled
+ * @description Webview debugging should not be enabled in production builds.
```
subatoi, line 3:
```suggestion
 * @description Enabling Webview debugging in production builds can expose entry points or leak sensitive information.
```

Just a small change to bring in line with the [query style guide](https://github.com/github/codeql/blob/main/docs/query-metadata-style-guide.md#query-descriptions-description).

subatoi (commented):
Hi @joefarebrother! I've just joined the Docs team, nice to meet you 👋  

I've just left a few minor comments based around language and bringing in line with our style guide, and asked one question. I'll be happy to follow up afterwards. Many thanks!

java/ql/lib/semmle/code/java/security/WebviewDubuggingEnabledQuery.qll
joefarebrother: It looks like it's a bit better without this line (in particular, no 'high cache usage' flagged)

java/ql/lib/semmle/code/java/security/WebviewDubuggingEnabledQuery.qll
```diff
@@ -0,0 +1,41 @@
...
+
+  override predicate isSink(DataFlow::Node node) {
+    exists(MethodAccess ma |
+      ma.getMethod().hasQualifiedName("android.webkit", "WebView", "setWebContentsDebuggingEnabled") and
+      node.asExpr() = ma.getArgument(0)
+    )
+  }
+
+  override predicate isBarrier(DataFlow::Node node) {
+    exists(Guard debug | isDebugCheck(debug) and debug.controls(node.asExpr().getBasicBlock(), _))
```
atorralba, line 37: We are sanitizing both branches here, and I think that might be a little too much. Is this done this way to handle the `0 != (getApplicationInfo().flags & ApplicationInfo.FLAG_DEBUGGABLE` case?

If that's the case, would it make sense to implement more precise guards? Namely, something like this: https://github.com/github/codeql/pull/10177/files#diff-38978229f15815dd860ec9f3696f469770f85b8082736b808072ef64be18a834R202

So if the `debug` variable is of type boolean, we can assume it sanitizes the branch `true`, whereas if it's an `int` (or anything else, really), we can expect there'll be some `EqualityTest` involved and then sanitize the branch indicated by `EqualityTest::polarity`. What do you think?

joefarebrother: How much benefit is there to making the guard more precise? It doesn't seem that likely to me that somebody would write a check for a debug flag but then enable debug features in the wrong branch. And there are different possible variations; e.g. an integer might be checked that it's >0 instead, which we might not catch by only looking for equality tests.

atorralba: I ran a query to find out how many sinks were sanitized by this guard in 1k projects, and only 4 results came back. Those 4 results were under the correct branch, so I think your assumption makes sense and we don't need to increase the precision of the guard. We can always go back and do it if we spot many FNs caused by this in the future.

atorralba (approved):

PR #10242, opened 2022-08-31
Kotlin: Refactor kotlinFunctionToJavaEquivalent

igfoo:
Avoids a cast.

### DIFF
--- a/java/kotlin-extractor/src/main/kotlin/KotlinUsesExtractor.kt
+++ b/java/kotlin-extractor/src/main/kotlin/KotlinUsesExtractor.kt
@@ -1195,17 +1195,15 @@ open class KotlinUsesExtractor(
                             decl.valueParameters.size == f.valueParameters.size
                         } ?:
                         // Or check property accessors:
-                        if (f.isAccessor) {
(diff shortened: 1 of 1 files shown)


java/kotlin-extractor/src/main/kotlin/KotlinUsesExtractor.kt
```diff
@@ -1195,17 +1195,15 @@ open class KotlinUsesExtractor(
                             decl.valueParameters.size == f.valueParameters.size
                         } ?:
                         // Or check property accessors:
-                        if (f.isAccessor) {
-                            val prop = javaClass.declarations.findSubType<IrProperty> { decl ->
-                                decl.name == (f.propertyIfAccessor as IrProperty).name
+                        (f.propertyIfAccessor as? IrProperty?)?.let { kotlinProp ->
```
tamasvajk, line 1198: Isn't there an extra `?` in `as? IrProperty?`? Wouldn't `as? IrProperty` be enough?

igfoo: Good point, thanks!

tamasvajk (approved):

PR #10252, opened 2022-09-01
PY: change alert messages of path queries to use the same template

erik-krogh:
I tried to rewrite the path queries.   
- I tried to fit the message to the template: `This path depends on a user-provided value.` 
- If that didn't work I tried `User-provided value flows to this location and is used in a path.`

**WIP**


martin389:
Hello from this week's docs first responder 👋 Thanks for the ping! I've added this PR to our review board for another writer to look at. They'll be able to advise on any additional considerations for this change.

python/ql/src/Functions/ModificationOfParameterWithDefault.ql
```diff
@@ -19,5 +19,5 @@ from
   ModificationOfParameterWithDefault::Configuration config, DataFlow::PathNode source,
   DataFlow::PathNode sink
 where config.hasFlowPath(source, sink)
-select sink.getNode(), source, sink, "$@ flows to here and is mutated.", source.getNode(),
-  "Default value"
+select sink.getNode(), source, sink, "This value modification depends on a $@.", source.getNode(),
+  "default value"
```
tausbn, line 23: This one sounds a bit strange to me ("'value modification'? What's that?"), but I couldn't come up with a better phrasing.

Also, "depends on a default value" seems too weak of a statement to me. It not only _depends_ on a default value, it actively _mutates_ said default value. 

Edit: Perhaps "`This expression mutates a default value.`" would be better?

tausbn (changes requested):
One comment/suggestion, otherwise these all look good to me. :+1:

python/ql/src/Functions/ModificationOfParameterWithDefault.ql
tausbn: Actually, `s/mutates/modifies/` might be even better.

erik-krogh: > Also, "depends on a default value" seems too weak of a statement to me. It not only _depends_ on a default value, it actively _mutates_ said default value.

You're right, it's a `DataFlow::Configuration` and not a `TaintTracking::Configuration`, so _depends_ is too weak a statement.   
"This expression mutates a default value" sounds good to me 👍

python/ql/src/Security/CWE-022/PathInjection.ql
```diff
@@ -22,5 +22,5 @@ import DataFlow::PathGraph
 
 from Configuration config, DataFlow::PathNode source, DataFlow::PathNode sink
 where config.hasFlowPath(source, sink)
-select sink.getNode(), source, sink, "$@ flows to here and is used in a path.", source.getNode(),
-  "User-provided value"
+select sink.getNode(), source, sink, "This path depends on $@.", source.getNode(),
+  "a user-provided value"
```
felicitymay, line 26: I'm rather rusty on how these select statements look in results. Am I right to think that this is shown as:

> This path depends on [a user-provided value](source.getNode).

If so, this looks good to me 👍🏻

felicitymay (approved):
As I mentioned in a comment, I'm rather rusty on select statements.

Assuming that I'm reading these correctly, these changes look clearer to me ✨

python/ql/src/Security/CWE-022/PathInjection.ql
tausbn: Yep, that's right!

tausbn (approved):

PR #10206, opened 2022-08-29
JS: Some JS-specific MaD changes

asgerf:
Some changes to the JS-specific MaD library in support of generated typings.

[Evaluation](https://github.com/github/codeql-dca-main/issues/6509) looks quiet

PR #10253, opened 2022-09-01
JS: Add generated typings to SQL models

asgerf:
This refactors most of the SQL and NoSQL models to rely on MaD type-definitions generated from `.d.ts` files from their respective libraries.

Most of the infrastructure behind this was merged in https://github.com/github/codeql/pull/10205 and https://github.com/github/codeql/pull/10206.

### Model.json files
Each MaD model lives in a folder containing a `model.json` and `Model.qll` file. The `.qll` file is entirely generated from the `.json` file and isn't something you'd normally look at.

Some information about the `model.json` files:
- It lists the exact library versions used to produce the model.
- The `generatedModel` field contains CSV rows entirely generated by madman, which can be regenerated from the rest of the file. It only contains type definitions and type variables (no sources and sinks).
- The `model` field contains user-maintained CSV rows to be merged with those from `generatedModel`. Anything added via the interactive mode in `madman` gets added here as well (i.e. sinks).
- The `usedTypes` field lists `package;type` pairs that should be considered relevant (i.e. not pruned), since we expect to use from from `ModelOutput::getATypeNode`
- The `replaceTypeParameters` field identifies type parameters that should be seen as an alias for another type. This can help deal with typings that are parameterized in ways that we don't care about.

### Comparison to hand-written models
The initial version of the generated models were not a perfect improvement over hand-written models, as both models covered some cases that were missing in the other one (due to incomplete typings). I manually added the missing parts to the `model` section, or via the `replaceTypeParameters` feature, to ensure a final model that covers everything.

Previously we'd have roughly one predicate per type we cared about, for example,
```codeql
  /** Gets a data flow node referring to a MongoDB collection. */
  private API::Node getACollection() {
    exists(API::Node collection |
      collection = getAMongoClientOrDatabase().getMember("collection").getReturn()
    |
      result = collection
      or
      result = collection.getParameter(1).getParameter(0)
    )
    or
    result = API::Node::ofType("mongodb", "Collection")
  }
```
Now, only the last clause is relevant. Simply doing `API::Node::ofType("mongodb", "Collection")` should give everything from the above, because the MaD tables contain the information previously hand-written in the bodies of these predicates.

### Evaluations

[Evaluation 1](https://github.com/github/codeql-dca-main/issues/7109) and [Evaluation 2](https://github.com/github/codeql-dca-main/issues/7108) cover 200 projects that depend on various SQL/NoSQL libraries, and show in total:
- 637 new SQL injection sinks
- 8 new SQL injection alerts


RasmusWL (approved):
Pyhton/Ruby changes looks fine :blush:

javascript/ql/lib/semmle/javascript/frameworks/NoSQL.qll
```diff
@@ -17,490 +17,73 @@ module NoSql {
...
-      }
-
-      override API::Node getQueryArgument() {
-        exists(int n |
-          MethodSignatures::interpretsArgumentAsQuery(methodName, n) and
-          result = this.getParameter(n)
-        )
-      }
+      this =
+        ModelOutput::getATypeNode("mongodb", ["Db", "MongoClient"])
```
erik-krogh, line 49:
```suggestion
        ModelOutput::getATypeNode("mongodb", "Db")
```

In the above you alias MongoClient as DB (the `mongodb;Db;mongodb;MongoClient;` row).  
So this suggestion should be equivalent? 

(Maybe you did it like you've done on purpose, and in that case dismiss this suggestion).

javascript/ql/test/library-tests/frameworks/SQL/SqlString.expected
```diff
@@ -41,7 +41,11 @@
 | sequelize-types.ts:7:24:7:35 | 'SELECT 123' |
 | sequelize.js:8:17:8:118 | 'SELECT ... Y name' |
 | sequelizeImport.js:3:17:3:118 | 'SELECT ... Y name' |
+| spanner2.js:4:19:4:32 | "not SQL code" |
```
erik-krogh, line 44: The test recognizes this "not SQL code"-string as an SQL-string?  

(Same as few lines below).

erik-krogh (commented):
This is much more than "just" adding generated typings.  
You've greatly improved beyond adding generated typings.  

Solid work 💪  

A single optional suggestion, and a comment about the test-output.

javascript/ql/test/library-tests/frameworks/SQL/SqlString.expected
asgerf: That's right, I deliberately loosened the model a bit to not distinguish between endpoint that expect `string`, `{sql: string}`, or both. That tends to work out better IMO. I've updated the test to clarify on this.

javascript/ql/lib/semmle/javascript/frameworks/NoSQL.qll
asgerf: Yes, this was on purpose. The 2.x adapter was meant to stand on its own so the rest of the model wouldn't need to worry about it too much. If we later change how the 2.x adapter works, it would be a hassle to have to fix up places where we assumed they would be aliased.

erik-krogh (approved):
👍

PR #10266, opened 2022-09-01
python: rewrite CatchingBaseException from `points-to` to API graph

yoff:


tausbn (approved):
Looks good to me. We should figure out whether a change note is required (I would lean towards doing a single change note once we've done a bunch of these rewrites).

yoff:
> We should figure out whether a change note is required (I would lean towards doing a single change note once we've done a bunch of these rewrites).

Agreed. Can we do just one? Should we make it a single PR, then?

tausbn:
> Agreed. Can we do just one? Should we make it a single PR, then?

I think keeping the query rewrites as separate PRs is probably best. We should just add the https://github.com/github/codeql/labels/no-change-note-required label, and then put the change note in a separate PR at a later date.

PR #10267, opened 2022-09-01
python: Rewrite EmptyExcept from `points-to` to API graph

yoff:


python/ql/src/Exceptions/EmptyExcept.ql
```diff
@@ -61,27 +62,32 @@ predicate subscript(Stmt s) {
...
+    type = API::builtin("UnicodeDecodeError").getAValueReachableFromSource().asExpr()
   )
 }
 
-predicate small_handler(ExceptStmt ex, Stmt s, ClassValue type) {
+predicate small_handler(ExceptStmt ex, Stmt s, Expr type) {
   not exists(ex.getTry().getStmt(1)) and
   s = ex.getTry().getStmt(0) and
-  ex.getType().pointsTo(type)
+  ex.getType() = type
```
tausbn, line 78: This is (at a surface level) somewhat weaker than the previous version, but I guess it might be okay?

tausbn (approved):
Lovely to see these rewrites!

I'm tempted to suggest pushing most of the "action" to a higher level (so as to avoid having to write `asExpr` everywhere), but I think that might as well wait for a later round of clean-up.

python/ql/src/Exceptions/EmptyExcept.ql
yoff: The logic is reshuffled a bit. In this predicate, we no longer consider a `ClassValue` but merely an expression. The relation between the expression and an appropriate class is now simpler here and more complicated in `focussed_handler` (and `encode_decode`) where the API graphs kicks in.

tausbn: Oh, good point! Thanks for clarifying. :slightly_smiling_face:

PR #10268, opened 2022-09-02
Kotlin: fix doc comment extraction for local functions

tamasvajk:


java/kotlin-extractor/src/main/kotlin/comments/CommentExtractor.kt
```diff
@@ -104,7 +105,11 @@ class CommentExtractor(private val fileExtractor: KotlinFileExtractor, private v
                             val existingLabel = if (ownerIr is IrVariable) {
                                 label = "variable ${ownerIr.name.asString()}"
                                 tw.getExistingVariableLabelFor(ownerIr)
-                            } else {
+                            } else if (ownerIr is IrFunction && ownerIr.isLocalFunction()) {
+                                label = "local function ${ownerIr.name.asString()}"
+                                fileExtractor.getLocallyVisibleFunctionLabels(ownerIr).function
+                            }
```
igfoo, line 111: I'm a bit confused by this, and the `IrVariable` case above. Can't we move them in to `getLabel`?

tamasvajk: `getLabel` returns the string representation of a label. In case of a local function or a variable, we don't have a meaningful unique string name, only `*`. Let me do a small refactoring to pull these special cases to a properly named function to ease readability.

tamasvajk: See https://github.com/github/codeql/pull/10268/commits/bb5f0d4afbeeda5318b96475336f3d34e5f3051c.

igfoo (approved):

PR #10274, opened 2022-09-02
Kotlin: Remove a cast in array iterator call extraction

igfoo:


java/kotlin-extractor/src/main/kotlin/KotlinFileExtractor.kt
```diff
@@ -2175,13 +2175,18 @@ open class KotlinFileExtractor(
                         if (dispatchReceiver == null) {
                             logger.errorElement("No dispatch receiver found for array iterator call", c)
                         } else {
-                            val typeArgs = (dispatchReceiver.type as IrSimpleType).arguments.map {
-                                when(it) {
-                                    is IrTypeProjection -> it.type
-                                    else -> pluginContext.irBuiltIns.anyNType
+                            val drType = dispatchReceiver.type
+                            if (drType !is IrSimpleType) {
+                                logger.errorElement("Dispatch receiver with unexpected type rep found for array iterator call: ${drType.javaClass}", c)
```
tamasvajk, line 2180: What is `type rep`?

igfoo: "type representation". i.e. if `drType.javaClass` is `IrSomeType`, then that isn't the type of the dispatch receiver, but it's the type of the thing that describes the type of the dispatch receiver.

I'm happy to take suggestions for a better phrase!

tamasvajk: I see. I would probably use `with unexpected type kind found`, but `representation` is also good, I understand now.
BTW, we already have a couple of error messages that include the phrases `simple type` or `non-simple type`. (We could change those to match the wording used here)

tamasvajk (approved):

PR #10276, opened 2022-09-02
Add TypeModel hook for adding MaD type-defs from CodeQL

asgerf:
Adds the following hook in `ModelInput` as way to contribute type-definitions from CodeQL.
```ql
  /**
   * A unit class for adding additional type model rows from CodeQL models.
   */
  class TypeModel extends Unit {
    /**
     * Gets a data-flow node that is a source of the type `package;type`.
     */
    DataFlow::Node getASource(string package, string type) { none() }

    /**
     * Gets a data flow node that is a sink of the type `package;type`,
     * usually because it is an argument passed to a parameter of that type.
     */
    DataFlow::Node getASink(string package, string type) { none() }
  }
```

For example, one of the use-cases could be in https://github.com/github/codeql/pull/10090 where it would look something like this
```ql
class AttachmentInstanceType extends ModelInput::TypeModel {
  override DataFlow::LocalSourceNode getASource(string package, string type) {
    package = "activestorage" and
    type = "Attachment" and
    result instanceof AttachmentInstance
  }
}
```

Also adds `API::EntryPoint` to Python and unifies its API across the three API graph implementations (Ruby and JS had drifted apart here).

The QLDoc check failure is spurious. I've opened an internal PR that should fix it.

### DIFF
--- a/javascript/ql/lib/semmle/javascript/frameworks/data/internal/ApiGraphModels.qll
+++ b/javascript/ql/lib/semmle/javascript/frameworks/data/internal/ApiGraphModels.qll
@@ -72,7 +72,7 @@ private class Unit = Specific::Unit;
 
 private module API = Specific::API;
 
-private module DataFlow = Specific::DF;
+private module DataFlow = Specific::DataFlow;
 
 private import Specific::AccessPathSyntax
 
--- a/javascript/ql/lib/semmle/javascript/frameworks/data/internal/ApiGraphModels.qll
+++ b/javascript/ql/lib/semmle/javascript/frameworks/data/internal/ApiGraphModels.qll
@@ -72,6 +72,8 @@ private class Unit = Specific::Unit;
 
 private module API = Specific::API;
 
+private module DataFlow = Specific::DF;
+
 private import Specific::AccessPathSyntax
 
 /** Module containing hooks for providing input data to be interpreted as a model. */
@@ -154,6 +156,22 @@ module ModelInput {
      * indicates that `(package2, type2, path)` should be seen as an instance of `(package1, type1)`.
      */
     abstract predicate row(string row);
+  }
+
+  /**
+   * A unit class for adding additional type model rows from CodeQL models.
+   */
+  class TypeModel extends Unit {
+    /**
+     * Gets a data-flow node that is a source of the type `package;type`.
+     */
+    DataFlow::Node getASource(string package, string type) { none() }
+
+    /**
+     * Gets a data flow node that is a sink of the type `package;type`,
+     * usually because it is an argument passed to a parameter of that type.
+     */
+    DataFlow::Node getASink(string package, string type) { none() }
   }
 
   /**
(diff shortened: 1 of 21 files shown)


javascript/ql/lib/semmle/javascript/frameworks/data/internal/ApiGraphModels.qll
```diff
@@ -368,6 +386,58 @@ private predicate invocationMatchesCallSiteFilter(Specific::InvokeNode invoke, A
...
+    typeModel(package, type, package2, type2, path2) and
+    result = getNodeFromPath(package2, type2, path2, path2.getNumToken())
+  )
+  or
+  result = any(TypeModelUseEntry e).getNodeForType(package, type)
+  or
+  result = any(TypeModelDefEntry e).getNodeForType(package, type)
+  or
+  isRelevantFullPath(package, type, _) and
+  result = Specific::getExtraNodeFromPath(package, type, _, 0)
```
aibaars, line 438: I wonder if this should be like:
```suggestion
  exists(string path | isRelevantFullPath(package, type, path) and
  result = Specific::getExtraNodeFromPath(package, type, path, 0))
```
Also these lines overlap with 
```
result = Specific::getExtraNodeFromPath(package, type, path, n)
```
on line [385](https://github.com/github/codeql/pull/10276/files#diff-f6a72b8d40fb2a6a2b8b273c53c19086292a8efa538591bcfb41204f07a327a8L385)

ruby/ql/lib/codeql/ruby/frameworks/data/internal/ApiGraphModels.qll
```diff
@@ -368,6 +386,58 @@ private predicate invocationMatchesCallSiteFilter(Specific::InvokeNode invoke, A
...
+  exists(string package2, string type2, AccessPath path2 |
+    typeModel(package, type, package2, type2, path2) and
+    result = getNodeFromPath(package2, type2, path2, path2.getNumToken())
+  )
+  or
+  result = any(TypeModelUseEntry e).getNodeForType(package, type)
+  or
+  result = any(TypeModelDefEntry e).getNodeForType(package, type)
+  or
+  isRelevantFullPath(package, type, _) and
```
aibaars, line 437: Same question here.

aibaars (commented):
This looks good to me from a Ruby point of view. I have a question though.

ruby/ql/lib/codeql/ruby/frameworks/data/internal/ApiGraphModels.qll
```diff
@@ -368,6 +386,58 @@ private predicate invocationMatchesCallSiteFilter(Specific::InvokeNode invoke, A
...
+}
+
+/**
+ * Gets an API node identified by the given `(package,type)` pair.
+ */
+pragma[nomagic]
+private API::Node getNodeFromType(string package, string type) {
+  exists(string package2, string type2, AccessPath path2 |
+    typeModel(package, type, package2, type2, path2) and
+    result = getNodeFromPath(package2, type2, path2, path2.getNumToken())
```
hvitved, line 430:
```suggestion
    result = getNodeFromPath(package2, type2, path2)
```

hvitved (commented):
One suggestion. Otherwise LGTM

javascript/ql/lib/semmle/javascript/frameworks/data/internal/ApiGraphModels.qll
asgerf: Using `getExtraNodeFromPath` with `n=0` was a bit of a hack, actually. The real fix is to introduce `getExtraNodeFromType` instead, PTAL.

ruby/ql/lib/codeql/ruby/frameworks/data/internal/ApiGraphModels.qll
asgerf: Same answer as above (this is shared code).

hvitved (dismissed):
Let's see what DCA says.

erik-krogh (approved):
JS 👍

asgerf:
There was previously an [evaluation for Ruby](https://github.com/github/codeql-dca-main/issues/6728), but I'm running another one for JS due to the drive-by fix while moving some code into `getExtraNodeFromType`.

python/ql/lib/semmle/python/frameworks/data/internal/ApiGraphModelsSpecific.qll
```diff
@@ -28,6 +28,9 @@ class Unit = PY::Unit;
 
 // Re-export libraries needed by ApiGraphModels.qll
 import semmle.python.frameworks.data.internal.AccessPathSyntax as AccessPathSyntax
+
+module DF = DataFlow;
```
RasmusWL, line 32: is there a reason you export this a `DF`, and not as `DataFlow` (which would seem more natural to me)?

RasmusWL (approved):
Python: Looks plausible :+1: (one minor nitpick)

python/ql/lib/semmle/python/frameworks/data/internal/ApiGraphModelsSpecific.qll
asgerf: Not a terribly good reason, actually, it was too fiddly to figure out how to export `DataFlow` without accidentally importing as well for each language.

asgerf: Changed to export it as `DataFlow`

RasmusWL (dismissed):
Python: Looks plausible :+1:

erik-krogh (approved):

erik-krogh (approved):

asgerf:
[JS evaluation](https://github.com/github/codeql-dca-main/issues/6822) looks fine

PR #10277, opened 2022-09-02
C#: Also inject `/p:UseSharedCompilation=false` into `dotnet publish`

hvitved:
Disable shared compilation for `dotnet publish` (like `dotnet build` and `dotnet msbuild`).

Update: Extended to also cover `dotnet {pack,run,test}`.

### DIFF
--- a/csharp/tools/tracing-config.lua
+++ b/csharp/tools/tracing-config.lua
@@ -31,7 +31,7 @@ function RegisterExtractorPack(id)
             local firstCharacter = string.sub(arg, 1, 1)
             if not (firstCharacter == '-') and not (firstCharacter == '/') then
                 Log(1, 'Dotnet subcommand detected: %s', arg)
-                if arg == 'build' or arg == 'msbuild' then match = true end
+                if arg == 'build' or arg == 'msbuild' or arg == 'publish' then match = true end
                 break
             end
         end
--- a/csharp/tools/tracing-config.lua
+++ b/csharp/tools/tracing-config.lua
@@ -19,6 +19,7 @@ function RegisterExtractorPack(id)
         -- if that's `build`, we append `/p:UseSharedCompilation=false` to the command line,
         -- otherwise we do nothing.
         local match = false
+        local needsSeparator = false;
         local argv = compilerArguments.argv
         if OperatingSystem == 'windows' then
             -- let's hope that this split matches the escaping rules `dotnet` applies to command line arguments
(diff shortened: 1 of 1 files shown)


alexet:
Do we want to go through all the commands and see if they might build code. There are others that look like they could build code (e.g. `pack` and `test`)

hvitved:
> Do we want to go through all the commands and see if they might build code. There are others that look like they could build code (e.g. `pack` and `test`)

Good idea. Looks like `dotnet run` is also relevant.

csharp/tools/tracing-config.lua
```diff
@@ -31,7 +31,8 @@ function RegisterExtractorPack(id)
             local firstCharacter = string.sub(arg, 1, 1)
             if not (firstCharacter == '-') and not (firstCharacter == '/') then
                 Log(1, 'Dotnet subcommand detected: %s', arg)
-                if arg == 'build' or arg == 'msbuild' then match = true end
+                if arg == 'build' or arg == 'msbuild' or arg == 'publish' or arg == 'pack' or arg == 'test' or
+                    arg == 'run' then match = true end
```
adityasharad, line 35: Minor: update the comment above to be more generic. Currently it only mentions `build`.

hvitved: Done.

adityasharad:
LGTM once you're happy with the DCA results. Let's also add some internal integration tests that exercise the different dotnet commands.

csharp/tools/tracing-config.lua
```diff
@@ -31,19 +32,30 @@ function RegisterExtractorPack(id)
...
+            end
+            if arg == '--' then
+                needsSeparator = false
                 break
             end
         end
         if match then
+            local injections = { '/p:UseSharedCompilation=false' }
+            if needsSeparator then
+                table.insert(injections, '--')
```
michaelnebel, line 54: For `dotnet run` you append `--` after the UseSharedCompilation flag (if not otherwise provided in the list of arguments). Does `--` have special semantic meaning, when executing `dotnet run`. That is, are everything *after* `--` considered arguments to the program that is run?

hvitved: Yes, indeed:
```
Description:
  .NET Run Command

Usage:
  dotnet run [options] [[--] <additional arguments>...]]

Additional Arguments:
  Arguments passed to the application that is being run.
```

michaelnebel (dismissed):
Looks plausible to me!

michaelnebel (approved):
Looks plausible to me!

PR #10278, opened 2022-09-02
Kotlin: Extract error expression for `enumValues<T>` calls

tamasvajk:
A call to `enumValues<MyEnum>()` is translated to `MyEnum.values()` in JVM. This transformation is also implemented in the extractor. However, the special case of `enumValues<T>()` was not handled. 

`T` can be `reified T : Enum<T>` of an `inline` function. In this case, the above `T.values()` conversion is done only on the call site, where `T` is already a known enum type. The method body in byte code contains a `new Enum[0]` array creation. This PR doesn't mimic this empty array creation, instead it extracts an `ErrorExpr`.

### DIFF
--- a/java/kotlin-extractor/src/main/kotlin/KotlinFileExtractor.kt
+++ b/java/kotlin-extractor/src/main/kotlin/KotlinFileExtractor.kt
@@ -1778,13 +1778,34 @@ open class KotlinFileExtractor(
                     return
                 }
 
-                val func = ((c.getTypeArgument(0) as? IrSimpleType)?.classifier?.owner as? IrClass)?.declarations?.findSubType<IrFunction> { it.name.asString() == fnName }
-                if (func == null) {
-                    logger.errorElement("Couldn't find function $fnName on enum type", c)
+                val enumType = (c.getTypeArgument(0) as? IrSimpleType)?.classifier?.owner
+                if (enumType == null) {
+                    logger.errorElement("Couldn't find type of enum type", c)
                     return
                 }
 
-                extractMethodAccess(func, false)
+                if (enumType is IrClass) {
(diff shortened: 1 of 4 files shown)


java/kotlin-extractor/src/main/kotlin/KotlinFileExtractor.kt
```diff
@@ -1778,13 +1778,34 @@ open class KotlinFileExtractor(
...
+                    // A call to `enumValues<T>()` is being extracted, where `T` is a reified type parameter of an `inline` function.
+                    // We can't generate a valid expression here, because we would need to know the type of T on the call site.
+                    val id = tw.getFreshIdLabel<DbErrorexpr>()
+                    val type = useType(c.type)
+
+                    tw.writeExprs_errorexpr(id, type.javaResult.id, parent, idx)
+                    tw.writeExprsKotlinType(id, type.kotlinResult.id)
+                    tw.writeHasLocation(id, tw.getLocation(c))
+                    tw.writeCallableEnclosingExpr(id, callable)
+                    tw.writeStatementEnclosingExpr(id, enclosingStmt)
```
igfoo, line 1805: I would say that having an errorexpr in the database means that there is an extractor bug. This is fine for now, but I think we at least want a TODO comment here to replace it with something else in the future.

I'm not sure what that would be exactly. Perhaps a new `@unrepresentableexpr`?

igfoo (approved):

java/kotlin-extractor/src/main/kotlin/KotlinFileExtractor.kt
tamasvajk: Done in https://github.com/github/codeql/pull/10278/commits/f84e62e16b767fd963d9cd2cbaf4c1f300146eeb

PR #10256, opened 2022-09-01
Swift: Use TypeDecl.getABaseTypeDecl().

geoffw0:
Very minor follow-up to https://github.com/github/codeql/pull/10061, taking advantage of https://github.com/github/codeql/pull/10254.

@MathiasVP

PR #10284, opened 2022-09-02
Swift: Improve swift/string-length-conflation

geoffw0:
Improve `swift/string-length-conflation`:
 - catch an edge case we were missing where a method was declared in a base class of the type we had expected it to be in.
 - clean up the logic a bit.


swift/ql/src/queries/Security/CWE-135/StringLengthConflation.ql
```diff
@@ -58,39 +58,41 @@ class StringLengthConflationConfiguration extends DataFlow::Configuration {
   override predicate isSource(DataFlow::Node node, string flowstate) {
     // result of a call to `String.count`
     exists(MemberRefExpr member |
-      member.getBase().getType().getName() = "String" and
+      member.getBase().getType().(NominalType).getABaseType*().getName() = "String" and
```
MathiasVP: Could you not use the predicate I added in https://github.com/github/codeql/pull/10254 and rewrite this to:
```ql
member.getBase().(ClassDecl).getABaseTypeDecl*().getName()
```
(or something along those lines)?

MathiasVP (commented):
One small comment. It also looks like this PR includes commits already merged into `main`? (i.e., [this one](https://github.com/github/codeql/pull/10256) and [this one](https://github.com/github/codeql/pull/10254))

swift/ql/src/queries/Security/CWE-135/StringLengthConflation.ql
geoffw0: `getBase()` returns an `Expr`, not a `ClassDecl`, so I think an extra step of some form is required either way.

MathiasVP: Doh, right 🤦 . Yeah, I see there are multiple ways of doing this, but none of them are _obviously_ shorter than the rest.

MathiasVP (approved):

geoffw0:
> It also looks like this PR includes commits already merged into main?

Yes, I believe that commit was required for this change (or at least I thought it was at one point).  I was having trouble with latest `main` so I pulled it in specifically instead.  It should come out in the merge.

PR #10286, opened 2022-09-02
JS: change alert messages of path queries to use the same template

erik-krogh:
I just did this for Python: https://github.com/github/codeql/pull/10252, now I'm doing the same thing for JavaScript. 

I'm rewriting alert messages for the taint tracking queries:  
- I tried to fit the message to the template: `This path depends on a user-provided value.` 
- If that didn't work I tried `User-provided value flows to this location and is used in a path.`

I also generally tried to use "this location" instead of "here". 
I tried to formulate messages to use "depends on" and use the term "a user-provided value". 

The previous change-note for changing the alert messages still cover this PR.


martin389:
Hello from this week's docs first responder 👋 Thanks for creating this PR! I've added it to our review board for another writer to look at. They'll be able to advise on any additional considerations for this change.

javascript/ql/src/Security/CWE-094/UnsafeDynamicMethodAccess.ql
```diff
@@ -18,4 +18,4 @@ from Configuration cfg, DataFlow::PathNode source, DataFlow::PathNode sink
 where cfg.hasFlowPath(source, sink)
 select sink, source, sink,
   "Invocation of method derived from $@ may lead to remote code execution.", source.getNode(),
```
felicitymay, line 20: Should this be more like:

```suggestion
  "This method is invoked using $@, which may allow remote code execution.", source.getNode(),
```

javascript/ql/src/Security/CWE-312/CleartextLogging.ql
```diff
@@ -38,5 +38,5 @@ where
   cfg.hasFlowPath(source, sink) and
   // ignore logging to the browser console (even though it is not a good practice)
   not inBrowserEnvironment(sink.getNode().asExpr().getTopLevel())
-select sink.getNode(), source, sink, "Sensitive data returned by $@ is logged here.",
-  source.getNode(), source.getNode().(Source).describe()
+select sink.getNode(), source, sink, "$@ is logged here.", source.getNode(),
```
felicitymay, line 41: Is this a place where "this location" was unsuitable?

javascript/ql/src/Security/CWE-400/RemotePropertyInjection.ql
```diff
@@ -18,5 +18,5 @@ import DataFlow::PathGraph
 
 from Configuration cfg, DataFlow::PathNode source, DataFlow::PathNode sink
 where cfg.hasFlowPath(source, sink)
-select sink.getNode(), source, sink, "A $@ is used as" + sink.getNode().(Sink).getMessage(),
+select sink.getNode(), source, sink, sink.getNode().(Sink).getMessage() + " depends on $@.",
   source.getNode(), "user-provided value"
```
felicitymay, line 22:
```suggestion
  source.getNode(), "a user-provided value"
```

javascript/ql/src/Security/CWE-834/LoopBoundInjection.ql
```diff
@@ -20,4 +20,4 @@ from Configuration dataflow, DataFlow::PathNode source, DataFlow::PathNode sink
 where dataflow.hasFlowPath(source, sink)
 select sink, source, sink,
   "Iterating over user-controlled object with a potentially unbounded .length property from $@.",
-  source, "here"
+  source, "a user-provided value"
```
felicitymay, line 23: Maybe more like this (it won't let me add a suggestion).

```
"Iteration over a user-controlled object with a potentially unbounded .length property from $@.",
  source, "a user-provided value"
```

felicitymay (approved):
@erik-krogh - many thanks for continuing to improve the consistency and clarity of our alert comments ✨ 

As you work on this, it would be great if you could update the [Style guide](https://github.com/github/codeql/blob/main/docs/query-metadata-style-guide.md#alert-messages) with the patterns you've implemented for common situations.

This generally looks good to me, as far as I can interpret the statements. I've added a few questions and comments for your consideration.

javascript/ql/src/Security/CWE-078/CommandInjection.ql
```diff
@@ -28,5 +28,5 @@ where
     else highlight = sink.getNode()
   ) and
   sourceNode = source.getNode()
-select highlight, source, sink, "$@ flows to here and is used in a command.", source.getNode(),
-  sourceNode.getSourceType()
+select highlight, source, sink, "This command line depends on $@.", source.getNode(),
```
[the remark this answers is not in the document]
erik-krogh: > I notice that for this query you include "This" but that this is omitted in TemplateObjectInjection.ql. Do you have a rule of thumb for deciding when to include "This" and when to omit it.

I don't have a rule of thumb, I think it was a coincidence.  
I think I prefer the more concise version without "This", so I'll remove the "This" from `CommandInjection.ql`.  

> I'm wondering if we should update https://github.com/github/codeql/blob/main/docs/query-metadata-style-guide.md#alert-messages, based on your work on select statements. 

We should! I'll get on that when I rewrite the alert-messages for the next language (after merging this PR).

javascript/ql/src/Security/CWE-079/UnsafeHtmlConstruction.ql
```diff
@@ -18,6 +18,6 @@ import semmle.javascript.security.dataflow.UnsafeHtmlConstructionQuery
 
 from DataFlow::Configuration cfg, DataFlow::PathNode source, DataFlow::PathNode sink, Sink sinkNode
 where cfg.hasFlowPath(source, sink) and sink.getNode() = sinkNode
-select sinkNode, source, sink, "$@ based on $@ might later cause $@.", sinkNode,
+select sinkNode, source, sink, "$@ which depends on $@ might later cause $@.", sinkNode,
```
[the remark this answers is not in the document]
erik-krogh: The query detects something that appears unsafe, but that's not a full exploitable vulnerability.   
The full vulnerability requires that the library (flagged by the query) is used with unsafe inputs in some other project.  

So that's why there is a "might" in the alert message.

javascript/ql/src/Security/CWE-312/CleartextLogging.ql
erik-krogh: I was trying to avoid "this location" in general.

javascript/ql/src/Security/CWE-079/UnsafeHtmlConstruction.ql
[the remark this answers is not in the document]
felicitymay: I'd suggest replacing "cause" since the code doesn't cause cross-site scripting, it seems more as if it "allows" a hacker to perform cross-site scripting.

```suggestion
select sinkNode, source, sink, "$@ which depends on $@ might later allow $@.", sinkNode,
```

javascript/ql/src/Security/CWE-312/CleartextLogging.ql
erik-krogh: Well, yes, I also tried that.  
I tried to avoid "here" and "this location" that doesn't describe what is happening, and use "Sensitive data", "command execution", or something else that describes what is happening.  
But for this alert-message I didn't find a good message that avoided the "here".

felicitymay (approved):
Thanks for the updates. Happy for this to be merged once it's okayed by the JavaScript team.

asgerf (approved):

PR #10292, opened 2022-09-05
QL4QL: Add `--search-path` to `codeql resolve languages` calls

edoardopirovano:
See also https://github.com/github/codeql-action/pull/1234

Since the introduction of TRAP caching, the CodeQL Action now does `codeql resolve languages` to find languages that support TRAP caching. This is failing for QL4QL because we aren't passing the `--search-path` to find the custom extractor in this call. This PR adds that in.


erik-krogh (approved):
👍

PR #10295, opened 2022-09-05
C#: No longer manually disable shared compilation in `codeql-analysis.yml`

hvitved:
Instead, rely on the tracer to inject the flag.


criemen (approved):
Yay!

PR #10297, opened 2022-09-05
Kotlin: Lookup getter methods based on special JVM method mapping

tamasvajk:


java/kotlin-extractor/src/main/kotlin/KotlinUsesExtractor.kt
```diff
@@ -1191,20 +1191,22 @@ open class KotlinUsesExtractor(
         else
             f.parentClassOrNull?.let { parentClass ->
                 getJavaEquivalentClass(parentClass)?.let { javaClass ->
-                    if (javaClass != parentClass)
+                    if (javaClass != parentClass) {
+                        val jvmName = getSpecialJvmName(f) ?: f.name.asString()
```
igfoo, line 1195: Are you sure we don't want `getJvmName`?

tamasvajk: It doesn't make a difference, because these functions have no annotation on them. I pushed a change to use `getJvmName`, and reverted its visibility change.

igfoo (approved):

PR #10300, opened 2022-09-05
C++: Multiple minor improvements to the cpp/cleartext-* queries

geoffw0:
Multiple minor improvements to the `cpp/cleartext-*` queries:
 - add a sanitizer for `IntegralType`s to all queries, as was done for `cpp/cleartext-storage-buffer` in https://github.com/github/codeql/pull/10128
   - this should improve performance and may get rid of some dubious / FP results.
 - rewrite `cpp/cleartext-storage-database` using AST taint (replacing security taint).  This is an end in itself, but also makes the query more readable and makes it possible to introduce the sanitizer.

I will run start a DCA run on all the changes, and probably an LGTM/MRVA run specifically for the changes to `cpp/cleartext-storage-database`.


geoffw0:
LGTM run of `cpp/cleartext-storage-database` on 18 relevant projects before: https://lgtm.com/query/227505127859308963/ and after these changes: https://lgtm.com/query/3389769774679136194/
 - 3 results on `stellar/stellar-core` are preserved
	- they look reasonable to me
 - 10 results on `bloomberg/comdb2` are lost
	- they involve a flow from a `clnt->current_user.password` to the `clnt` structure itself, which appears to be stored in the database (I think).  If so, these results appear legitimate as well.
	- I will look into using `allowImplicitRead` to catch these cases in the new version (the right way).

geoffw0:
It looks like the DCA run hit disk space issues.  I'm hopefully going to get an improvement in before I re-run it on this PR in any case.

TODO:
 - [ ] additional test cases showing the issue in `bloomberg/comdb2`.
 - [ ] `allowImplicitRead` improvement?
 - [ ] re-run DCA.

geoffw0:
I've just pushed the fix using `allowImplicitRead`.  It took a few iterations to get it right (and efficient).  The same LGTM run of the new version is [here](https://lgtm.com/query/8225880443163334969/) and has slightly more results than the original now (new seemingly good results on `bloomberg/comdb2`).

DCA run to follow...

geoffw0:
DCA is misbehaving right now (`Unhandled promise rejection`), I'll come back to this next week probably.

MathiasVP:
> DCA is misbehaving right now (`Unhandled promise rejection`), I'll come back to this next week probably.

There's been a recent large refactoring of DCA, so it might be worth asking the DCA people on Slack if they can identify the issue (since it might be a new bug that needs to be fixed)

cpp/ql/src/Security/CWE/CWE-313/CleartextSqliteDatabase.ql
```diff
@@ -34,25 +28,51 @@ predicate sqlite_encryption_used() {
...
 
-class Configuration extends TaintTrackingConfiguration {
-  override predicate isSource(Expr source) {
-    super.isSource(source) and source instanceof SensitiveExpr
+/**
+ *  Gets a field of the class `c`, or of another class contained in `c`.
+ */
+Field getRecField(Class c) {
+  result = c.getAField() or
+  result = getRecField(c.getAField().getUnspecifiedType().stripType())
```
MathiasVP, line 36: - Does this recursive field digging actually work? I don't see it changing any query results in our tests.

cpp/ql/src/change-notes/2022-09-05-cleartext-queries.md
```diff
@@ -0,0 +1,4 @@
+---
+category: minorAnalysis
+---
+* Modernizations from "Cleartext storage of sensitive information in buffer" (`cpp/cleartext-storage-buffer`) have been ported to the "Cleartext storage of sensitive information in file" (`cpp/cleartext-storage-file`), "Cleartext transmission of sensitive information" (`cpp/cleartext-transmission`) and "Cleartext storage of sensitive information in an SQLite database" (`cpp/cleartext-storage-database`) queries. These changes may reduce false positive results.
```
MathiasVP, line 4: This PR also overrides `allowImplicitRead`, which might give _more_ results. Should the possibility of "more results" also be mentioned in the change-note?

MathiasVP (commented):
A couple of small comments. Otherwise, LGTM once DCA is happy.

cpp/ql/src/Security/CWE/CWE-313/CleartextSqliteDatabase.ql
geoffw0: The second case is needed for a couple of test cases, though that only uses one recursive step so I think this is a fair criticism.  I've added another test case (to the internal test) with two recursive steps.

cpp/ql/src/change-notes/2022-09-05-cleartext-queries.md
geoffw0: Done.

geoffw0:
I don't seem to be able to start a DCA run today ("Unhandled promise rejection"...) :(

MathiasVP:
> I don't seem to be able to start a DCA run today ("Unhandled promise rejection"...) :(

I'll start one for you. One sec.

geoffw0:
Thanks!

geoffw0:
DCA shows a substantial speedup of `cpp/cleartext-storage-database` (250s -> 61s).  That's the one with the most rewriting here - so I'm very pleased to see it!  The other queries aren't much affected, which is fine.

MathiasVP (approved):

PR #10236, opened 2022-08-31
C++: Workaround for missing `DeclarationEntry` for `DeclStmt`

MathiasVP:
This is a simple IR fix for [missing association between DeclarationEntry and DeclStmt in template instantiations](https://github.com/github/codeql-c-team/issues/1009). Draft while investigating the implications on DCA.

PR #10305, opened 2022-09-05
C++: Synthesize `DeclarationEntry`s for IR construction

MathiasVP:
Naively ignoring functions containing `DeclStmt`s with missing `DeclarationEntry`s was attempted in https://github.com/github/codeql/pull/10236. Unfortunately, this removed too much IR to be a feasible solution.

Instead, this PR synthesizes `PseudoDeclarationEntry` (feel free to suggest a new name! Maybe `IRDeclarationEntry`) to handle the places where the IR expected a declaration entry.

With this PR, I expect we don't need to turn on those scary front-end flags. @rdmarsh2 I'd be happy to hear about whether this fixes the performance issues you've encountered.

Note: As a bonus, this PR actually also _deletes_ a bunch of code as well 😍. Turns out we had special-cased code to handle `DeclarationEntry`s generated from range-based for loops (which also suffered from the missing-declaration-entry-in-declaration-statements issue). With these new `PseudoDeclarationEntry`s we can actually completely delete this specialized fix, and just rely on `PseudoDeclarationEntry` and the `DeclStmt`s that we already have for the desugared range-based for loops. This is done in f6654e3d69fce7c1454d6ae2d6c018f246e5f702.

- 2da46ad691579e271940756b452d0c144af33dc2 adds two failing tests that cause bad IR.
- fc85ef76ed831aa215e7481bd33c0e0027c17f75 introduces `PseudoDeclarationEntry` and adds them where the IR generation expects `DeclarationEntry`s.
- f6654e3d69fce7c1454d6ae2d6c018f246e5f702 deletes the special-case code we had for missing `DeclarationEntry`s in range-based for loops.


cpp/ql/lib/semmle/code/cpp/ir/implementation/raw/internal/TranslatedElement.qll
```diff
@@ -458,6 +458,100 @@ private predicate translateDeclarationEntry(DeclarationEntry entry) {
...
+    TMissingDeclarationEntry(DeclStmt stmt, Declaration d, int index) {
+      not exists(stmt.getDeclarationEntry(index)) and
+      stmt.getDeclaration(index) = d
+    }
+
+  /**
+   * An entity that represents a declaration entry in the database.
+   *
+   * This class exists to work around the fact that `DeclStmt`s in
+   * template instantiations do not have `DeclarationEntry`s.
```
jketema, line 473: This should probably also mention range-based for-loops.

cpp/ql/lib/semmle/code/cpp/ir/implementation/raw/internal/TranslatedElement.qll
```diff
@@ -458,6 +458,100 @@ private predicate translateDeclarationEntry(DeclarationEntry entry) {
...
+   *
+   * So instead, the IR works with `PseudoDeclarationEntry`s that synthesize
+   * missing `DeclarationEntry`s when there is no result for `DeclStmt::getDeclarationEntry`.
+   */
+  abstract class PseudoDeclarationEntry extends TPseudoDeclarationEntry {
```
jketema, line 478: I would indeed just go with `IRDeclarationEntry`, as I doubt this will be fixed for ranged-based for loops.

cpp/ql/lib/semmle/code/cpp/ir/implementation/raw/internal/TranslatedElement.qll
```diff
@@ -458,6 +458,100 @@ private predicate translateDeclarationEntry(DeclarationEntry entry) {
...
+   * result for `DeclStmt::getDeclarationEntry`
+   */
+  private class MissingDeclarationEntry extends PseudoDeclarationEntry, TMissingDeclarationEntry {
+    DeclStmt stmt;
+    Declaration d;
+    int index;
+
+    MissingDeclarationEntry() { this = TMissingDeclarationEntry(stmt, d, index) }
+
+    override string toString() { result = "pseudo declaration of " + d.getName() }
```
jketema, line 526:
```suggestion
    override string toString() { result = "missing declaration of " + d.getName() }
```

jketema (commented):
Some small comments. Otherwise, LGTM. I assume we want to wait for confirmation from @rdmarsh2.

MathiasVP:
> Some small comments. Otherwise, LGTM. I assume we want to wait for confirmation from @rdmarsh2.

Thanks! Yeah, let's do that.

cpp/ql/lib/semmle/code/cpp/ir/implementation/raw/internal/TranslatedElement.qll
MathiasVP: Renamed to `IRDeclarationEntry` in 10f962f341ef071d02f392d524802256e06184c3.

MathiasVP:
cc @rdmarsh2 do you mind checking if this solves your performance issues?

rdmarsh2:
Yes, it does.

PR #10310, opened 2022-09-06
Docs: No longer mention required `/p:UseSharedCompilation=false`

hvitved:
No longer needed with the new tracer.


criemen (approved):
LGTM!

PR #10311, opened 2022-09-06
C++: Support more builtin operations

jketema:
<strike>This still requires a stats file update, but otherwise is ready to be reviewed.</strike>

Note that the bulk of the changes is in [cpp/ql/lib/semmle/code/cpp/exprs/BuiltInOperations.qll](https://github.com/github/codeql/pull/10311/files#diff-4dbaa606f9785b755097586a4255243c74e2ff507be5ebc1e3fb2675b84529ab), which doesn't show a diff by default.


MathiasVP (approved):
LGTM! Do we want to run a DCA to make sure the new stats file doesn't break anything?

jketema:
> LGTM! Do we want to run a DCA to make sure the new stats file doesn't break anything?

That's the last two I ran on the internal PR. The second is a re-run I'm doing at the moment, because I don't completely trust the results of the first one. I will not merge this until the experiment complete and the results look ok.

jketema:
Results of the second DCA experiment looked fine. Rebased this, so needs re-approval.

PR #10312, opened 2022-09-06
ensure consistent casing of names

erik-krogh:
[I recently added a QL-for-QL query to detect when two names only differ in casing inside a single language](https://github.com/github/codeql/pull/10209).  

E.g. Python used both `Multidict` and `MultiDict`.  
This can cause confusion, so I'm fixing it. 

I did the following renamings: 
- `SSA` -> `Ssa`
- `LDAP` -> `Ldap`
- `HTTP` -> `Http`
- `Multidict` -> `MultiDict`
- `URL` -> `Url`
- `Newobj` -> `NewObj`
- `Rethrow` -> `ReThrow`

In Python I had to do a few renamings in order to avoid naming conflicts. 

I skipped `HTML` and `API`, so those are still used both as `HTML/Html` and `API/Api`.

### DIFF
--- a/python/ql/lib/semmle/python/frameworks/Multidict.qll
+++ b/python/ql/lib/semmle/python/frameworks/Multidict.qll
@@ -16,7 +16,7 @@ private import semmle.python.frameworks.internal.InstanceTaintStepsHelper
  * Provides models for the `multidict` PyPI package.
  * See https://multidict.readthedocs.io/en/stable/.
  */
-module Multidict {
+module MultiDict {
   /**
    * Provides models for a `MultiDictProxy` class:
    * - `multidict.MultiDictProxy`
@@ -88,3 +88,6 @@ module Multidict {
     }
   }
 }
+
+/** DEPRECATED: Alias for MultiDict */
+deprecated module Multidict = MultiDict;
--- a/python/ql/lib/semmle/python/frameworks/Multidict.qll
+++ b/python/ql/lib/semmle/python/frameworks/Multidict.qll
@@ -16,7 +16,7 @@ private import semmle.python.frameworks.internal.InstanceTaintStepsHelper
  * Provides models for the `multidict` PyPI package.
  * See https://multidict.readthedocs.io/en/stable/.
  */
-module MultiDict {
+module Multidict {
   /**
    * Provides models for a `MultiDictProxy` class:
    * - `multidict.MultiDictProxy`
@@ -88,6 +88,3 @@ module MultiDict {
     }
   }
 }
-
-/** DEPRECATED: Alias for MultiDict */
-deprecated module Multidict = MultiDict;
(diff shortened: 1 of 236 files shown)


tausbn (dismissed):
Python :+1:

michaelnebel (approved):
C# 👍

geoffw0 (dismissed):
C++ :+1:

python/ql/lib/semmle/python/frameworks/Multidict.qll
```diff
@@ -16,7 +16,7 @@ private import semmle.python.frameworks.internal.InstanceTaintStepsHelper
  * Provides models for the `multidict` PyPI package.
  * See https://multidict.readthedocs.io/en/stable/.
  */
-module Multidict {
+module MultiDict {
```
RasmusWL, line 19: This naming was intentional (to match the PyPI package name), so lets keep it!

```suggestion
module Multidict {
```

python/ql/lib/semmle/python/frameworks/Multidict.qll
```diff
@@ -88,3 +88,6 @@ module Multidict {
     }
   }
 }
+
+/** DEPRECATED: Alias for MultiDict */
+deprecated module Multidict = MultiDict;
```
RasmusWL, line 93:
```suggestion
```

python/ql/lib/semmle/python/frameworks/Yarl.qll
```diff
@@ -102,7 +102,7 @@ module Yarl {
     }
 
     /** An attribute read on a `yarl.URL` that is a `MultiDictProxy` instance. */
-    class YarlUrlMultiDictProxyInstance extends Multidict::MultiDictProxy::InstanceSource {
+    class YarlUrlMultiDictProxyInstance extends MultiDict::MultiDictProxy::InstanceSource {
```
RasmusWL, line 105:
```suggestion
    class YarlUrlMultiDictProxyInstance extends Multidict::MultiDictProxy::InstanceSource {
```

python/ql/lib/semmle/python/frameworks/Stdlib.qll
```diff
@@ -1877,7 +1877,7 @@ private module StdlibPrivate {
   API::Node http() { result = API::moduleImport("http") }
 
   /** Provides models for the `http` module. */
-  module Http {
+  module StdHttp {
```
RasmusWL, line 1880:
```suggestion
  module StdlibHttp {
```

RasmusWL (changes requested):
I recently looked at changing the casing to `Http` in Concepts.qll, but ran into the same problems as you have with the `Http` modules in Django/Stdlib and so on. I guess renaming them is fine, since they're inside `Private...` modules that are intended for only internal use.

I was a little surprised about the naming `StdHttp`, and I think it would more consistent to use `StdlibHttp`.

MathiasVP (dismissed):

asgerf (dismissed):
JS 👍

MathiasVP (approved):

aibaars (dismissed):
Ruby changes look good to me.

RasmusWL (dismissed):
Python :+1:

RasmusWL (approved):
Python still :+1: (sorry for the late review, thought this might have been waiting for someone else)

PR #10314, opened 2022-09-06
Python: Revert changes to sensitive data query alert messages

RasmusWL:
This partly reverts the changes from https://github.com/github/codeql/pull/10252

Although consistency is nice, the new messages didn't sound as natural.

New alert message would read

> Insecure hashing algorithm (md5) depends on sensitive data (password). (...)

I'm not sure what it means that a hashing algorithm depends on data. So
for me, the original text below is much easier to understand.

> Sensitive data (password) is used in a hashing algorithm (md5) that is insecure (...)

Same goes for the other sensitive data queries.


erik-krogh (approved):
👍

tausbn (approved):
Nice! These all sound better to me! 👍

PR #10316, opened 2022-09-06
Ruby: Model ActionView

hmac:
Model ActionView.

### DIFF
--- a/ruby/ql/lib/codeql/ruby/frameworks/ActionView.qll
+++ b/ruby/ql/lib/codeql/ruby/frameworks/ActionView.qll
@@ -291,9 +291,9 @@ module ActionView {
     }
 
     /**
-     * `ActionView::Helpers::TagHelper#tag`.
+     * `ActionView::Helpers::TagHelper#content_tag`.
      *
-     * `tag(x, x, y, false)`.
+     * `content_tag(x, x, y, false)`.
      */
     private class ContentTag extends ActionViewContextCall, RawHelperCall {
       ContentTag() {


ruby/ql/lib/codeql/ruby/frameworks/ActionView.qll
```diff
@@ -216,4 +216,138 @@ class FileSystemResolverAccess extends DataFlow::CallNode, FileSystemAccess::Ran
...
+   */
+  module Helpers {
+    /**
+     * A call to ActionView helpers which renders its argument without escaping.
+     * The argument should be treated as an XSS sink. In the documentation for
+     * classes in this module, the vulnerable argument is named `x`.
+     */
+    abstract class RawHelperCall extends MethodCall {
+      /**
+       * Get an argument which is rendered without escaping.
```
nickrolfe, line 235: The [style guide](https://github.com/github/codeql/blob/main/docs/qldoc-style-guide.md#predicates-with-result) suggests:
```suggestion
       * Gets an argument which is rendered without escaping.
```

ruby/ql/lib/codeql/ruby/frameworks/ActionView.qll
```diff
@@ -216,4 +216,138 @@ class FileSystemResolverAccess extends DataFlow::CallNode, FileSystemAccess::Ran
...
 // TODO: model flow in/out of template files properly,
+// TODO: Move the classes and predicates above inside this module.
+/** Modeling for `ActionView`. */
+module ActionView {
+  /**
+   * Action view helper methods which are XSS sinks.
+   */
+  module Helpers {
+    /**
+     * A call to ActionView helpers which renders its argument without escaping.
```
nickrolfe, line 229: Do you think this reads better?
```suggestion
     * A call to an ActionView helper which renders its argument without escaping.
```

ruby/ql/lib/codeql/ruby/frameworks/ActionView.qll
```diff
@@ -216,4 +216,138 @@ class FileSystemResolverAccess extends DataFlow::CallNode, FileSystemAccess::Ran
...
+        this.getKeywordArgument("escape").getConstantValue().isBoolean(false)
+      }
+
+      override Expr getRawArgument() { result = this.getArgument(0) }
+    }
+
+    /**
+     * `ActionView::Helpers::TextHelper#highlight`.
+     *
+     * `truncate(x, y, sanitize: false)`.
```
nickrolfe, line 271:
```suggestion
     * `highlight(x, y, sanitize: false)`.
```

nickrolfe (commented):
Just a few qldoc nitpicks, but otherwise looks like a nice improvement.

ruby/ql/lib/codeql/ruby/frameworks/ActionView.qll
```diff
@@ -216,4 +216,138 @@ class FileSystemResolverAccess extends DataFlow::CallNode, FileSystemAccess::Ran
...
+     *
+     * `tag(x, x, y, false)`.
+     */
+    private class Tag extends ActionViewContextCall, RawHelperCall {
+      Tag() {
+        this.getMethodName() = "tag" and
+        this.getArgument(3).getConstantValue().isBoolean(false)
+      }
+
+      override Expr getRawArgument() { result = this.getArgument(0) }
```
alexrford, line 318: It seems like the `false` argument to `escape` disables escaping of attribute values, so this should maybe find keyword argument values instead? I used the `Legacy syntax` section of https://api.rubyonrails.org/v5.1/classes/ActionView/Helpers/TagHelper.html#method-i-tag as reference.

alexrford (changes requested):
Just a couple of minor points, otherwise looks good.

ruby/ql/lib/codeql/ruby/frameworks/ActionView.qll
alexrford: Interesting, thanks for testing that.

alexrford (approved):
👍

ruby/ql/lib/codeql/ruby/frameworks/ActionView.qll
```diff
@@ -216,4 +216,138 @@ class FileSystemResolverAccess extends DataFlow::CallNode, FileSystemAccess::Ran
...
+  /**
+   * Action view helper methods which are XSS sinks.
+   */
+  module Helpers {
+    /**
+     * A call to an ActionView helper which renders its argument without escaping.
+     * The argument should be treated as an XSS sink. In the documentation for
+     * classes in this module, the vulnerable argument is named `x`.
+     */
+    abstract class RawHelperCall extends MethodCall {
```
nickrolfe, line 233: We should avoid introducing public abstract classes when we don't want/expect users to provide their own implementations. However, I'm already working on a branch to remove the other abstract classes in our framework libraries, so I suggest merging this PR as-is, and then I can fix this one as well.

PR #10317, opened 2022-09-06
PY: deprecate a bunch of unused code 

erik-krogh:
I'm looking through some of the results of this yet-to-be-merged QL-for-QL query: https://github.com/github/codeql/pull/8454  

I know the Python teams wants to deprecate the points-to code, so I tried to see if I could deprecate some unused code in Python. 

No query/test depend on any of the code I've deprecated. 

There was some more code I initially tried to deprecate. However, it was being used in internal queries (that QL-for-QL can't see).

### DIFF
--- a/python/ql/lib/change-notes/2022-09-12-deprecate-pointsto.md
+++ b/python/ql/lib/change-notes/2022-09-12-deprecate-pointsto.md
@@ -1,0 +1,4 @@
+---
+category: deprecated
+---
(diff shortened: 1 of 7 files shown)


tausbn (commented):
Changes look good to me. I'm debating whether this needs a change note or not. Some of these predicates are in principle visible externally, though I wouldn't expect anyone to actually depend on them.

erik-krogh:
> Changes look good to me. I'm debating whether this needs a change note or not. Some of these predicates are in principle visible externally, though I wouldn't expect anyone to actually depend on them.

I've added a change-note.

tausbn (approved):
Nice! Thank you! :heart:

PR #10319, opened 2022-09-06
Swift: Add encryption barriers to the two swift/cleartext-* queries

geoffw0:
This is a simple improvement that didn't make it into the original PR of the two `cleartext-*` queries.  Removes some false positive results.


MathiasVP (approved):
LGTM!

PR #10322, opened 2022-09-06
Kotlin: Rewrite array type extraction

igfoo:
It's now simpler, and doesn't require we ask the Kotlin compiler to build as many types for us.

### DIFF
--- a/java/kotlin-extractor/src/main/kotlin/KotlinUsesExtractor.kt
+++ b/java/kotlin-extractor/src/main/kotlin/KotlinUsesExtractor.kt
@@ -539,21 +539,52 @@ open class KotlinUsesExtractor(
                 )
         }
 
-    private fun useArrayType(arrayType: IrSimpleType, componentType: IrType, elementType: IrType, dimensions: Int, isPrimitiveArray: Boolean): TypeResults {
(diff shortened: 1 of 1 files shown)


java/kotlin-extractor/src/main/kotlin/KotlinUsesExtractor.kt
```diff
@@ -539,31 +539,62 @@ open class KotlinUsesExtractor(
...
+     * `t` is somewhere in a stack of array types, or possibly the
+     * element type of the innermost array. For example, in
+     * `Array<Array<Int>>`, we will be called with `t` being
+     * `Array<Array<Int>>`, then `Array<Int>`, then `Int`.
+     * `isPrimitiveArray` is true if we are immediately nested
+     * inside a primitive array.
+     */
+    private fun useArrayType(t: IrType, isPrimitiveArray: Boolean): ArrayInfo {
+
+        if (!t.isBoxedArray && !t.isPrimitiveArray()) {
```
tamasvajk, line 556: I think we already have a helper function `isArray()` that does exactly this check.

tamasvajk (approved):
LGTM

PR #10332, opened 2022-09-07
JS: bugfixes in TypeThroughThroughParameterTampering

asgerf:
Fixes a bunch of issues with the TypeThroughThroughParameterTampering query:
- `Array.isArray` was incorrectly recognized (kudos to @erik-krogh for spotting that)
- We override `Configuration.isBarrier` without calling `super`, meaning that all barrier guards stop working. This footgun should really be fixed in general, but for now I'm just fixing the query.
- The test case was wrong. It used `req.query` as the source, which is always an object and can not itself be the cause of type confusion (only its properties can).

### DIFF
--- a/javascript/ql/test/query-tests/Security/CWE-843/tst.js
+++ b/javascript/ql/test/query-tests/Security/CWE-843/tst.js
@@ -101,6 +101,7 @@ express().get('/foo', function (req, res) {
 
 express().get('/foo', function (req, res) {
     let data = req.query.data;
+    data.indexOf(); // NOT OK
     if (Array.isArray(data)) {
         data.indexOf(); // OK
     } else {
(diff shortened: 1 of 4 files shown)


javascript/ql/test/query-tests/Security/CWE-843/tst.js
```diff
@@ -100,7 +100,7 @@ express().get('/foo', function (req, res) {
 });
 
 express().get('/foo', function (req, res) {
-    let data = req.query;
+    let data = req.query.data;
```
erik-krogh, line 103:
```suggestion
    let data = req.query.data;
    data.indexOf(); // NOT OK
```
Let us include an example we actually flag. 
That's why the bug wasn't found originally.

asgerf: Good point 👍

erik-krogh (approved):

asgerf:
[Evaluation](https://github.com/github/codeql-dca-main/issues/6829) looks quiet

PR #10334, opened 2022-09-07
Java: Implement union type flow and replace ad-hoc variable tracking in dispatch

aschackmull:
This provides a proper replacement for the ad-hoc pre-SSA variable tracking that dispatch used to rely on.

Union types arising from disjunctive instanceof checks are ~left as future work~ also supported.


java/ql/test/library-tests/typeflow/uniontypeflow.expected
```diff
@@ -0,0 +1,24 @@
+| UnionTypes.java:11:5:11:5 | m | 2 | ConcurrentHashMap<String,String> | false |
+| UnionTypes.java:11:5:11:5 | m | 2 | LinkedHashMap<String,String> | false |
+| UnionTypes.java:26:10:26:10 | x | 2 | A2 | true |
+| UnionTypes.java:26:10:26:10 | x | 2 | A3 | false |
```
hvitved, line 4: Why is this not `true`?

java/ql/lib/semmle/code/java/dataflow/internal/DataFlowDispatch.qll
```diff
@@ -63,15 +64,21 @@ private module DispatchImpl {
...
         or
         ctx.getArgument(i) = arg
       |
-        src = VirtualDispatch::variableTrack(arg) and
-        srctype = getPreciseType(src) and
-        if src instanceof ClassInstanceExpr then exact = true else exact = false
+        exprTypeFlow(arg, srctype, exact)
+        or
+        not exprTypeFlow(arg, _, _) and
+        exprUnionTypeFlow(arg, srctype, exact)
```
hvitved, line 76: It would be nice with a data flow test that exhibits this; i.e., a call context that provides a disjunctive type bound, which is better than the disjunctive type bound inside the callee.

java/ql/lib/semmle/code/java/dataflow/TypeFlow.qll
```diff
@@ -589,6 +622,175 @@ private predicate bestTypeFlow(TypeFlowNode n, RefType t) {
...
+ * being contained in another part of the union type bound.
+ */
+private predicate irrelevantUnionTypePart(TypeFlowNode n, RefType t, boolean exact) {
+  unionTypeFlow(n, t, exact) and
+  not irrelevantUnionType(n) and
+  exists(RefType weaker |
+    unionTypeFlow(n, weaker, false) and
+    t.getASupertype*() = weaker
+  |
+    exact = true or not weaker.getASupertype*() = t
```
hvitved, line 778: I don't quite understand why this restriction is there.

hvitved (commented):
Looks awesome! I have some questions.

java/ql/lib/semmle/code/java/dataflow/TypeFlow.qll
```diff
@@ -509,7 +520,29 @@ private predicate typeFlowBase(TypeFlowNode n, RefType t) {
   )
 }
 
+/**
+ * Holds if `n` has type `t` and this information is discarded, such that `t`
+ * might be a better type bound for nodes where `n` flows to. This only includes
+ * the best such bound for each node.
```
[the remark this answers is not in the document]
aschackmull: It is possible to have multiple best bounds, yes. But it is fairly rare, to the point where we mostly ignore the possibility.

java/ql/lib/semmle/code/java/dataflow/TypeFlow.qll
```diff
@@ -509,7 +520,29 @@ private predicate typeFlowBase(TypeFlowNode n, RefType t) {
...
+ */
+private predicate typeFlowBase(TypeFlowNode n, RefType t) {
+  exists(RefType te |
+    typeFlowBaseCand(n, t) and
+    te = t.getErasure() and
+    not exists(RefType better |
+      typeFlowBaseCand(n, better) and
+      better != t and
```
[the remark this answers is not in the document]
aschackmull: The trickiness is that there can be loops in the subtyping relation.

java/ql/lib/semmle/code/java/dataflow/TypeFlow.qll
```diff
@@ -589,6 +622,175 @@ private predicate bestTypeFlow(TypeFlowNode n, RefType t) {
...
+  )
+}
+
+/**
+ * Holds if `ioe` checks `v`, its true-successor is `bb`, and `bb` has 2 or more
+ * predecessors.
+ */
+private predicate instanceofDisjunct(InstanceOfExpr ioe, BasicBlock bb, BaseSsaVariable v) {
+  ioe.getExpr() = v.getAUse() and
+  strictcount(bb.getABBPredecessor()) > 1 and
```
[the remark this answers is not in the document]
aschackmull: With the additional negation, I'd guess not. And `strictcount` is certainly efficient enough that I think this sort of tweak would lie firmly in the realm of micro-optimisations that's best handled by the optimiser.

java/ql/test/library-tests/typeflow/uniontypeflow.expected
aschackmull: Because `A3` has no subtypes - in those cases we just report upper bounds and don't bother with the lower bounds.

java/ql/lib/semmle/code/java/dataflow/TypeFlow.qll
```diff
@@ -589,6 +622,175 @@ private predicate bestTypeFlow(TypeFlowNode n, RefType t) {
...
+    exists(TypeFlowNode scc |
+      sccRepr(n, scc) and
+      // Optimized version of
+      // `forex(TypeFlowNode mid | sccJoinStep(mid, scc) | unionTypeFlowBaseCand(mid, _, _) or hasUnionTypeFlow(mid))`
+      ForAll<RankedSccJoinStep, HasUnionTypePropagation>::flowJoin(scc, _)
+    )
+    or
+    exists(TypeFlowNode mid | step(mid, n) and hasUnionTypeFlow(mid))
+    or
+    instanceofDisjunctionGuarded(n, _)
```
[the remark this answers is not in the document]
aschackmull: No, because `unionTypeFlowBaseCand` consists of nodes that have a single bound that may downstream be combined to disjunctive bounds, whereas `instanceofDisjunctionGuarded` is already a disjunctive bound.

java/ql/lib/semmle/code/java/dataflow/TypeFlow.qll
aschackmull: We're trying to determine whether we can reduce `A | B | ...` to just `B | ...`. We can do this either if `B` is a strictly weaker type than `A` or if `A` and `B` are similar types only distinguished by `A` being exact and `B` being an upper bound. Note that we have to be careful not to drop both `A` and `B` at the same time if they are equivalent types (in that case we could drop one of them, but that requires an arbitrary choice, which is tricky, so we don't bother with that case).

aschackmull:
> Looks awesome! I have some questions.

I think I've addressed everything.

java/ql/lib/semmle/code/java/dataflow/TypeFlow.qll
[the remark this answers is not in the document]
hvitved: 😱

java/ql/lib/semmle/code/java/dataflow/TypeFlow.qll
hvitved: Ok, I guess this is again because sub typing is not acyclic.

hvitved (approved):

PR #10338, opened 2022-09-07
Ruby: Treat ActiveRecord::Base.create as a model instantiation

hmac:
Calls such as `Post.create` are now recognised as model instantiations.

[Evaluation](https://github.com/github/codeql-dca-main/issues/7318) shows no noticeable change (good or bad).


alexrford (approved):
👍

PR #10340, opened 2022-09-07
C++: further work on experimental buffer overflow queries

rdmarsh2:
With these changes, the buffer overflow queries now rely on use-use flow (#10190), so I'm opening as a draft for review and comment.


MathiasVP:
I've moved this work into a PR that we can actually merge: https://github.com/github/codeql/pull/10398

PR #10341, opened 2022-09-07
Ruby: Inline `getAValueReachableFromSource`

hvitved:
Computing this predicate in full can sometimes be expensive, so inline instead (like in JavaScript).


aibaars (approved):
Looks fine to me. Did you perform a DCA run?

hvitved:
> Did you perform a DCA run?

Yes (see link above).

PR #9974, opened 2022-08-05
Ruby: Model ActiveResource

hmac:
https://github.com/rails/activeresource

PR #10348, opened 2022-09-08
Ruby: Fix ActiveResource HTTP client request modeling

RasmusWL:
This was a conflict between the merge of https://github.com/github/codeql/pull/9974 and https://github.com/github/codeql/pull/10114


asgerf (approved):
LGTM

The QLDoc check failure occurs because the baseline commit has compilation errors (unlike the other checks, the QLDOc check needs to be compile the baseline commit). Should be safe to ignore

PR #10349, opened 2022-09-08
C++: Create a `ZeroBound` per function

MathiasVP:
When we call `ZeroBound::getExpr` to get an expression that matches a given bound we eventually reach https://github.com/github/codeql/blob/8b8e74cc9a11f3cb5bedd0aed42ec26de179dc61/cpp/ql/lib/experimental/semmle/code/cpp/rangeanalysis/Bound.qll#L57 which will match _any_ instruction with a constant value of `0`. As @rdmarsh2 noted, this gives a very large tuple explosion.

This PR creates a per-function `ZeroBound` and restricts the expressions returned by `ZeroBound::getExpr` to be only instructions in the same callable as the `ZeroBound`.

### DIFF
--- a/cpp/ql/lib/experimental/semmle/code/cpp/rangeanalysis/Bound.qll
+++ b/cpp/ql/lib/experimental/semmle/code/cpp/rangeanalysis/Bound.qll
@@ -3,7 +3,7 @@ private import semmle.code.cpp.ir.IR
 private import semmle.code.cpp.ir.ValueNumbering
 
 private newtype TBound =
-  TBoundZero() or
+  TBoundZero(IRFunction func) or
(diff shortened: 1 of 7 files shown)


aschackmull:
Why is this relevant? Doesn't this just indicate that you have a poor join-order somewhere else?

MathiasVP:
> Why is this relevant? Doesn't this just indicate that you have a poor join-order somewhere else?

Hmm, actually you might be right. I'll turn this into a draft and discuss it with @rdmarsh2 on what his use-case for it was. Looking at my evaluation log file (which I thought was blowing up due to too many expressions) is really just blowing up due to a bad join in `unequalIntegralSsa` 🤦. Thanks for the quick comment!

PR #10351, opened 2022-09-08
JS: find a main module in more cases

erik-krogh:
Gets a TP/TN for CVE-2022-36036.  

Recognizes files that aren't exported under the "." path as a "main module", if we can't find anything else.  

Also looks in all subfolders if we can't directly find the file.  
The basename must match, and both the file and the path are equally deep (and exactly one level down).   
This handles if e.g. Rollup moves the compiled files from one folder to another (e.g. `lib` or `dist`).  

---- 

[Evaluation looks OK. ](https://github.com/github/codeql-dca-main/blob/data/erik-krogh/moreMains__nightly-old__code-scanning__1/reports) 
There are 2 new resolved imports. One is an FP, but that's due to us not extracting the file that is supposed to be imported (I'll check that later).  
There are a bunch of new library-inputs, they seem reasonable.  
The new alerts also look reasonable.

### DIFF
--- a/javascript/ql/lib/semmle/javascript/NodeModuleResolutionImpl.qll
+++ b/javascript/ql/lib/semmle/javascript/NodeModuleResolutionImpl.qll
@@ -93,14 +93,28 @@ private string getStem(string name) { result = name.regexpCapture("(.+?)(?:\\.([
  * Gets the main module described by `pkg` with the given `priority`.
  */
 File resolveMainModule(PackageJson pkg, int priority) {
-  exists(PathExpr main | main = MainModulePath::of(pkg, ".") |
-    result = main.resolve() and priority = 0
+  exists(PathExpr main, int subPriority, string mainPath |
+    main = MainModulePath::of(pkg, mainPath) and
+    if mainPath = "." then subPriority = priority else priority = subPriority + 1000
+  |
+    result = main.resolve() and subPriority = 0
     or
-    result = tryExtensions(main.resolve(), "index", priority)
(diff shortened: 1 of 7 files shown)


javascript/ql/lib/semmle/javascript/NodeModuleResolutionImpl.qll
```diff
@@ -93,14 +93,28 @@ private string getStem(string name) { result = name.regexpCapture("(.+?)(?:\\.([
...
+    )
+    or
+    // assuming the files get moved from one dir to another during compilation:
+    not exists(main.resolve()) and // didn't resolve
+    count(int i, string comp | comp = main.getComponent(i) and not comp = "." | i) = 2 and // is down one folder
+    exists(Folder subFolder | subFolder = pkg.getFile().getParentContainer().getAFolder() |
+      // is in one folder below the package.json, and has the right basename
+      result =
+        tryExtensions(subFolder, getStem(main.getComponent(main.getNumComponent() - 1)),
+          subPriority - 999)
```
asgerf, line 117: This `subPriority - 999` seems like it will cause confusion every time we come back to look at this.

Could you split the `exists` clause into two parts, so the second one just uses `priority` directly? I'd find it easier to understand that this back-and-forth adjustment of priority.

erik-krogh: I've moved most of the exists into a helper predicate. I think that should help readability.

asgerf (approved):

PR #5935, opened 2021-05-20
Java : Add SSTI query

porcupineyhairs:
This adds a query to detect server side template injections in Java. 

This is a continuation of #3353. Since, that one is quite stale, I am closing that and opening a new one .

PR #10352, opened 2022-09-08
Java: Promote Server-side template injection from experimental

atorralba:
Promotes https://github.com/github/codeql/pull/5935 to the main query pack.

DCA and MRVA runs look uneventful.


java/ql/lib/semmle/code/java/security/TemplateInjectionQuery.qll
```diff
@@ -0,0 +1,40 @@
...
+import java
+import semmle.code.java.dataflow.TaintTracking
+import semmle.code.java.dataflow.FlowSources
+import semmle.code.java.security.TemplateInjection
+
+/** A taint tracking configuration to reason about server-side template injection (SST) vulnerabilities */
+class TemplateInjectionFlowConfig extends TaintTracking::Configuration {
+  TemplateInjectionFlowConfig() { this = "TemplateInjectionFlowConfig" }
+
+  override predicate isSource(DataFlow::Node source) { this.isSource(source, _) }
```
aschackmull, line 12: Should probably remove this line (and the corresponding `isSink`). What this currently means is that any source with a non-empty flow state will also be a source with an empty flow state, which probably wasn't intended.

java/ql/lib/semmle/code/java/security/TemplateInjectionQuery.qll
```diff
@@ -0,0 +1,40 @@
...
+
+  override predicate isSink(DataFlow::Node sink, DataFlow::FlowState state) {
+    sink.(TemplateInjectionSink).hasState(state)
+  }
+
+  override predicate isSanitizer(DataFlow::Node sanitizer) { this.isSanitizer(sanitizer, _) }
+
+  override predicate isSanitizer(DataFlow::Node sanitizer, DataFlow::FlowState state) {
+    sanitizer.(TemplateInjectionSanitizer).hasState(state)
+  }
```
aschackmull, line 28: The case with sanitizers is different, although there's also something wrong here. The stateless sanitizer predicate applies to _all_ states, so if you want both regular sanitizers and stateful sanitizers to be configurable, then they'll need a separate extension point.

atorralba: Hm, I think I'm missing something. In the current implementation, the stateless `isSanitizer` actually applies to _all_ states as intended, doesn't it? Well, to any `state` where `any(TemplateInjectionSanitizer).hasState(state)` holds, but I think that's functionally the same (?).

Should we just change the default state of the sanitizer to be `any()`:

```
abstract class TemplateInjectionSanitizer extends DataFlow::Node {
  /** Holds if this sanitizer has the specified `state`. */
  predicate hasState(DataFlow::FlowState state) { any() }
}
```

so that its extensions apply to all states by default, and then users can create an extensions that overrides `hasState` to create state-specific sanitizers?

java/ql/lib/semmle/code/java/security/TemplateInjectionQuery.qll
aschackmull: I think removing them makes the most sense (your code snippet with `this.isSource(source, any(DataFlow::FlowStateEmpty s))` ought to be equivalent to removing them).

java/ql/lib/semmle/code/java/security/TemplateInjectionQuery.qll
aschackmull: `FlowState` is an infinite type, so making changing the code to `any()` like that is equivalent to making it `abstract` (if it even compiles, which I doubt).
The problem is that if you currently add a sanitizer by extending `TemplateInjectionSanitizer` then it's going to apply to all states regardless of whether you override `hasState` or not.
And there's really no way of fixing that with just a single predicate extension point to represent both state-oblivious sanitizers and state-based sanitizers, since the former can't be represented as part of a binary predicate when the set of possible states is infinite.

atorralba:
Thanks for the review @aschackmull. Comments addressed.

aschackmull (dismissed):
LGTM, but please consider whether this needs a doc review prior to merging.

hubwriter:
Docs Team first responder here.

Thanks for adding the docs review label. I've added this to our Docs Team review board with a due date of next Wednesday. If you need it turned around before then please drop a comment into the #docs-content channel.

java/ql/src/Security/CWE/CWE-094/TemplateInjection.qhelp
```diff
@@ -0,0 +1,32 @@
+<!DOCTYPE qhelp PUBLIC "-//Semmle//qhelp//EN" "qhelp.dtd">
+<qhelp>
+<overview>
+<p>
+Template Injection occurs when user input is embedded in a template's code in an unsafe manner.
```
mchammer01, line 5: I don't think we require this capitalization here
```suggestion
Template injection occurs when user input is embedded in a template's code in an unsafe manner.
```

java/ql/src/Security/CWE/CWE-094/TemplateInjection.qhelp
```diff
@@ -0,0 +1,32 @@
...
+</recommendation>
+<example>
+<p>
+In the example given below, an untrusted HTTP parameter <code>code</code> is used as a Velocity template string. 
+This can lead to remote code execution.
+</p>
+<sample src="SSTIBad.java" />
+
+<p>
+In the next example the problem is avoided by using a fixed template string <code>s</code>.
```
mchammer01, line 24: Nit
```suggestion
In the next example, the problem is avoided by using a fixed template string <code>s</code>.
```

java/ql/src/Security/CWE/CWE-094/TemplateInjection.qhelp
```diff
@@ -0,0 +1,32 @@
...
+Since the template's code is not attacker-controlled in this case, the untrusted code execution is prevented.
```
mchammer01, line 25: To try and avoid the use of the passive voice
```suggestion
Since the template's code is not attacker-controlled in this case, this solution prevents the execution of untrusted code.
```

java/ql/src/Security/CWE/CWE-094/TemplateInjection.qhelp
```diff
@@ -0,0 +1,32 @@
...
+</p>
+<sample src="SSTIGood.java" />
+</example>
+<references>
+<li>Portswigger: <a href="https://portswigger.net/web-security/server-side-template-injection">Server Side Template Injection</a></li>
```
mchammer01, line 30: References need full stops.
```suggestion
<li>Portswigger: <a href="https://portswigger.net/web-security/server-side-template-injection">Server Side Template Injection</a>.</li>
```

mchammer01 (approved):
@atorralba - this LGTM ✨ 
I've made a few comments/suggestions for your consideration. Feel free to ignore those you don't agree with.

atorralba:
Thanks for the review @mchammer01! All suggestions applied :)

@aschackmull could you re-approve?

aschackmull (approved):

PR #10357, opened 2022-09-08
make a shared library of the typo database

erik-krogh:
The typo database was shared between JS and QL-for-QL.  
With the new shared libraries we can avoid storing 2 copies of the same file.


ql/ql/src/qlpack.yml
```diff
@@ -5,3 +5,5 @@ dbscheme: ql.dbscheme
 suites: codeql-suites
 defaultSuiteFile: codeql-suites/ql-code-scanning.qls
 extractor: ql
+dependencies:
+  codeql/typos: "*"
```
aeisenberg, line 9: For now, can you be explicit about the dependency here? This will require manually bumping when you change versions.

```suggestion
  codeql/typos: 0.0.1
```

We are working on something to make this simpler. See https://github.com/github/codeql-core/issues/2741

aeisenberg (commented):
Nice. A few small things.

tausbn (approved):
Apart from `codeql/typos` being a slightly odd name for a package (it doesn't immediately communicate what it's actually about), everything looks good to me. :+1:

PR #10358, opened 2022-09-08
Ruby: Context sensitive instance method resolution

hvitved:
This PR implements call-sensitive instance method resolution, using [the data-flow library's built-in support](https://github.com/github/codeql/blob/main/docs/ql-libraries/dataflow/dataflow.md#virtual-dispatch-with-call-context).

For example, in
```rb
class A
  def method1 x
    sink x
  end

  def method2 x
    method1 x
  end
end

class B < A
  def method1 x
    non_sink x
  end
end

B.new.method2(taint)
```
we will no longer report flow from `taint` to `sink` (but still to `non_sink`).

### DIFF
--- a/ruby/ql/lib/codeql/ruby/ast/Module.qll
+++ b/ruby/ql/lib/codeql/ruby/ast/Module.qll
@@ -14,6 +14,9 @@ class Module extends TModule {
   /** Gets the super class of this module, if any. */
   Module getSuperClass() { result = getSuperClass(this) }
 
+  /** Gets a sub class of this module, if any. */
+  Module getASubClass() { this = getSuperClass(result) }
+
   /** Gets a `prepend`ed module. */
   Module getAPrependedModule() { result = getAPrependedModule(this) }
 
--- a/ruby/ql/lib/codeql/ruby/ast/Module.qll
+++ b/ruby/ql/lib/codeql/ruby/ast/Module.qll
@@ -14,7 +14,7 @@ class Module extends TModule {
   /** Gets the super class of this module, if any. */
   Module getSuperClass() { result = getSuperClass(this) }
 
-  /** Gets a sub class of this module, if any. */
+  /** Gets an immediate sub class of this module, if any. */
   Module getASubClass() { this = getSuperClass(result) }
 
   /** Gets a `prepend`ed module. */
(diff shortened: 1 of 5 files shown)


ruby/ql/lib/codeql/ruby/controlflow/CfgNodes.qll
```diff
@@ -444,6 +444,9 @@ module ExprNodes {
 
     /** Gets the body of this `when`-clause. */
     final ExprCfgNode getBody() { e.hasCfgChild(e.getBody(), this, result) }
+
+    /** Gets the `i`th pattern this `when`-clause. */
```
intrigus-lgtm, line 448:
```suggestion
    /** Gets the `i`th pattern of this `when`-clause. */
```

ruby/ql/lib/codeql/ruby/ast/Module.qll
```diff
@@ -14,6 +14,9 @@ class Module extends TModule {
   /** Gets the super class of this module, if any. */
   Module getSuperClass() { result = getSuperClass(this) }
 
+  /** Gets a sub class of this module, if any. */
+  Module getASubClass() { this = getSuperClass(result) }
```
aibaars, line 18: This predicate returns direct sub classes only, doesn't it? Let document that and perhaps also make it more clear from its name.

ruby/ql/lib/codeql/ruby/dataflow/internal/DataFlowDispatch.qll
```diff
@@ -438,90 +448,103 @@ private DataFlow::LocalSourceNode trackModuleAccess(Module m) {
...
-          not m instanceof SingletonMethod and
-          if m.getEnclosingModule() instanceof Toplevel then exact = true else exact = false
-        )
-        or
-        selfInToplevel(self.getVariable(), tp) and
-        exact = true
-      )
+  or
+  // `in C => c then c.foo`
+  asModulePattern(n, tp) and
```
aibaars, line 535: Perhaps also add `C === x` tests. If I'm not mistaken, both types of `case` expressions are implemented using that operator.

aibaars: Ah I see you just moved these predicate. Perhaps we should post-pone `===` for a future PR. We might want to desugar `case` statements into `if` at some point too.

aibaars (commented):
Looks good to me.

ruby/ql/lib/codeql/ruby/dataflow/internal/DataFlowDispatch.qll
hvitved: Yes, we can do that in another PR. I guess is could be added to `hasAdjacentTypeCheckedReads`.

aibaars (approved):

PR #10359, opened 2022-09-08
Python: Clean up module resolution

tausbn:
Part 1: Moving the current logic into its own internal module. Hopefully this will make the PR that removes the dependence on points-to easier to review (and less likely to conflict with other changes).

### DIFF
--- a/python/ql/lib/semmle/python/dataflow/new/internal/DataFlowPrivate.qll
+++ b/python/ql/lib/semmle/python/dataflow/new/internal/DataFlowPrivate.qll
@@ -335,11 +335,7 @@ predicate runtimeJumpStep(Node nodeFrom, Node nodeTo) {
   nodeFrom = nodeTo.(ModuleVariableNode).getAWrite()
   or
   // Setting the possible values of the variable at the end of import time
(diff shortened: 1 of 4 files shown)


python/ql/lib/semmle/python/dataflow/new/internal/DataFlowPrivate.qll
```diff
@@ -2,6 +2,7 @@ private import python
 private import DataFlowPublic
 private import semmle.python.essa.SsaCompute
 private import semmle.python.dataflow.new.internal.ImportStar
```
RasmusWL, line 4: Could we delete this import now then?

```suggestion
```

RasmusWL (approved):
:+1: besides a minor suggestion

python/ql/lib/semmle/python/dataflow/new/internal/DataFlowPrivate.qll
tausbn: Yep. Good catch!

RasmusWL (approved):

PR #10360, opened 2022-09-08
Dataflow: Fix implicit reads in taint tracking when FlowStates are used

atorralba:
The first commit adds a test that fails because implicit reads weren't allowed in sinks or additional taint steps that used FlowStates.

The second commit fixes the issue.

### DIFF
--- a/java/ql/lib/change-notes/2022-09-08-implicit-read-flowstates.md
+++ b/java/ql/lib/change-notes/2022-09-08-implicit-read-flowstates.md
@@ -1,0 +1,4 @@
+---
+category: fix
+---
+* Fixed an issue in the taint tracking analysis where implicit reads were not allowed by default in sinks or additional taint steps that used flow states.
--- a/cpp/ql/lib/change-notes/2022-09-08-implicit-read-flowstates.md
+++ b/cpp/ql/lib/change-notes/2022-09-08-implicit-read-flowstates.md
@@ -1,0 +1,4 @@
+---
+category: fix
+---
+* Fixed an issue in the taint tracking analysis where implicit reads were not allowed by default in sinks or additional taint steps that used flow states.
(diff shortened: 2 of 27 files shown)


MathiasVP (dismissed):

java/ql/lib/change-notes/2022-09-08-implicit-read-flowstates.md
```diff
@@ -0,0 +1,4 @@
+---
+category: fix
+---
+* Fixed an issue in the taint tracking analysis where implicit reads were not allowed by default in sinks or additional taint steps that used flow states.
```
aschackmull, line 4: If we think this needs a change note, then it probably needs to be for all the languages.

atorralba: Do you think the change note isn't needed?

aschackmull: I could be convinced either way - no strong feelings. It's fairly technical. Argument in favour: If someone is writing queries using flow state and had trouble with this, then it could be relevant. Argument against: If you're mostly just running queries out-of-the-box then it's probably not that likely to affect you, and as-is it's likely hard to understand from that perspective.

atorralba: Let's risk being verbose: 6f9b856.

javascript/ql/lib/change-notes/2022-09-08-implicit-read-flowstates.md
```diff
@@ -0,0 +1,4 @@
+---
+category: fix
+---
+* Fixed an issue in the taint tracking analysis where implicit reads were not allowed by default in sinks or additional taint steps that used flow states.
```
aschackmull, line 4: JS doesn't use the shared lib, so we can remove this change note.

atorralba: Woops 👍

aschackmull (approved):

PR #10364, opened 2022-09-09
Go: Remove the legacy tracer configuration files.

criemen:
These are not needed anymore, and from 2.11.0 on we will not ship a CLI that can read these files.


smowton (approved):
I'll take your word for it :)

PR #10365, opened 2022-09-09
Kotlin: Fix `isUnspecialised` to handle generic classes inside generic methods

tamasvajk:


java/kotlin-extractor/src/main/kotlin/utils/TypeSubstitution.kt
```diff
@@ -215,27 +217,36 @@ private fun matchingTypeParameters(l: IrTypeParameter?, r: IrTypeParameter): Boo
...
 
-    val parentClass = paramsContainer.parents.firstIsInstanceOrNull<IrClass>()
+    val parentTypeContainer = paramsContainer.parents.firstIsInstanceOrNull<IrTypeParametersContainer>()
 
     val parentUnspecialised = when {
         remainingArgs.isEmpty() -> true
-        parentClass == null -> false
-        else -> isUnspecialised(parentClass, remainingArgs)
+        parentTypeContainer == null -> {
+            if (remainingArgs.isNotEmpty()) {
```
igfoo, line 239: Isn't this always true given the previous branch?

igfoo (commented):
Otherwise LGTM

java/kotlin-extractor/src/main/kotlin/utils/TypeSubstitution.kt
tamasvajk: Thanks, good point.

igfoo (approved):

PR #10366, opened 2022-09-09
C++: Use-use flow in `experimental`

MathiasVP:
Due to the performance regressions on https://github.com/github/codeql/pull/10190 we've decided to put the new use-use dataflow code in the experimental directory so that we can use it while we're working on experimental buffer-overflow queries.

[The first commit](6d313ace2d4b04f38aba75c91125a2d5b8d5b0df) just copies over all the already-reviewed code from https://github.com/github/codeql/pull/10190 (along with a couple of copies of the dataflow library). So it doesn't really require any deep reviewing (since it's already received a lot of awesome 👀s!).

The last two commits are required changes to non-experimental code that should hopefully be semantically equivalent from the point of view of the existing non-experimental dataflow library.

### DIFF
--- a/cpp/ql/lib/semmle/code/cpp/models/implementations/Iterator.qll
+++ b/cpp/ql/lib/semmle/code/cpp/models/implementations/Iterator.qll
@@ -223,7 +223,7 @@ private class IteratorCrementMemberOperator extends MemberFunction, DataFlowFunc
     output.isQualifierObject()
     or
     input.isQualifierObject() and
-    output.isReturnValueDeref()
+    output.isReturnValue()
   }
 
   override predicate hasTaintFlow(FunctionInput input, FunctionOutput output) {
--- a/cpp/ql/lib/semmle/code/cpp/models/implementations/Iterator.qll
+++ b/cpp/ql/lib/semmle/code/cpp/models/implementations/Iterator.qll
@@ -223,7 +223,7 @@ private class IteratorCrementMemberOperator extends MemberFunction, DataFlowFunc
     output.isQualifierObject()
     or
     input.isQualifierObject() and
-    output.isReturnValue()
+    output.isReturnValueDeref()
   }
 
   override predicate hasTaintFlow(FunctionInput input, FunctionOutput output) {
--- a/cpp/ql/lib/semmle/code/cpp/models/implementations/StdString.qll
+++ b/cpp/ql/lib/semmle/code/cpp/models/implementations/StdString.qll
@@ -176,7 +176,7 @@ private class StdStringAppend extends TaintFunction {
     ) and
     (
       output.isQualifierObject() or
-      output.isReturnValueDeref()
+      output.isReturnValue()
     )
     or
     // reverse flow from returned reference to the qualifier (for writes to
@@ -543,11 +543,11 @@ private class StdOStreamOutNonMember extends DataFlowFunction, TaintFunction {
 
   override predicate hasTaintFlow(FunctionInput input, FunctionOutput output) {
     // flow from second parameter to first parameter
-    input.isParameter(1) and
+    input.isParameterDeref(1) and
     output.isParameterDeref(0)
     or
     // flow from second parameter to return value
-    input.isParameter(1) and
+    input.isParameterDeref(1) and
     output.isReturnValueDeref()
     or
     // reverse flow from returned reference to the first parameter
--- a/cpp/ql/lib/semmle/code/cpp/models/implementations/StdString.qll
+++ b/cpp/ql/lib/semmle/code/cpp/models/implementations/StdString.qll
@@ -176,7 +176,7 @@ private class StdStringAppend extends TaintFunction {
     ) and
     (
       output.isQualifierObject() or
-      output.isReturnValue()
+      output.isReturnValueDeref()
     )
     or
     // reverse flow from returned reference to the qualifier (for writes to
--- a/cpp/ql/lib/semmle/code/cpp/models/implementations/Strcat.qll
+++ b/cpp/ql/lib/semmle/code/cpp/models/implementations/Strcat.qll
@@ -61,7 +61,7 @@ class StrcatFunction extends TaintFunction, DataFlowFunction, ArrayFunction, Sid
     input.isParameterDeref(0) and
     output.isParameterDeref(0)
     or
-    input.isParameter(1) and
+    input.isParameterDeref(1) and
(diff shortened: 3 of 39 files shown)


cpp/ql/lib/semmle/code/cpp/models/implementations/StdString.qll
```diff
@@ -543,11 +543,11 @@ private class StdOStreamOutNonMember extends DataFlowFunction, TaintFunction {
 
   override predicate hasTaintFlow(FunctionInput input, FunctionOutput output) {
     // flow from second parameter to first parameter
-    input.isParameter(1) and
+    input.isParameterDeref(1) and
     output.isParameterDeref(0)
     or
     // flow from second parameter to return value
-    input.isParameter(1) and
+    input.isParameterDeref(1) and
```
geoffw0, line 550: These two assume we're using `operator<<` to output a string (not an `int`, `char` etc) but I think that's reasonable.

cpp/ql/lib/semmle/code/cpp/models/implementations/Strcat.qll
```diff
@@ -61,7 +61,7 @@ class StrcatFunction extends TaintFunction, DataFlowFunction, ArrayFunction, Sid
     input.isParameterDeref(0) and
     output.isParameterDeref(0)
     or
-    input.isParameter(1) and
+    input.isParameterDeref(1) and
```
geoffw0, line 64: :+1:

cpp/ql/lib/semmle/code/cpp/models/interfaces/FunctionInputsAndOutputs.qll
```diff
@@ -46,6 +46,26 @@ class FunctionInput extends TFunctionInput {
...
+   * void func(int n, char* p, float& r);
+   * ```
+   * - `isParameterDeref(1, 1)` holds for the `FunctionInput` that represents the value of `*p` (with
+   *   type `char`) on entry to the function.
+   * - `isParameterDeref(2, 1)` holds for the `FunctionInput` that represents the value of `r` (with type
+   *   `float`) on entry to the function.
+   * - There is no `FunctionInput` for which `isParameterDeref(0, _)` holds, because `n` is neither a
+   *   pointer nor a reference.
+   */
+  predicate isParameterDeref(ParameterIndex index, int ind) {
```
geoffw0, line 65: I don't understand what the `ind` parameter means here.

geoffw0 (commented):
(reviewing just the changes to models)

Some of these deserve a bit of discussion, though I understand they're motivated by test regressions so its very likely you're right or at least partly right in each case.

cpp/ql/lib/semmle/code/cpp/models/implementations/Iterator.qll
```diff
@@ -223,7 +223,7 @@ private class IteratorCrementMemberOperator extends MemberFunction, DataFlowFunc
     output.isQualifierObject()
     or
     input.isQualifierObject() and
-    output.isReturnValueDeref()
+    output.isReturnValue()
```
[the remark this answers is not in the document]
MathiasVP: Hm, yes. This might be an issue with how the use-use flow PR did flow through iterators (because they're not actually a pointer type, but we want to model them as such). I think I'll revert this change (+ the one in `StdStringAppend` since it's also related to iterators). It may mean that we have less flow in the use-use code, but since this is now in experimental it doesn't really matter. And I'd rather risk losing flow in experimental than in our production code.

cpp/ql/lib/semmle/code/cpp/models/implementations/StdString.qll
```diff
@@ -176,7 +176,7 @@ private class StdStringAppend extends TaintFunction {
     ) and
     (
       output.isQualifierObject() or
-      output.isReturnValueDeref()
+      output.isReturnValue()
```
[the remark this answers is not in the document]
MathiasVP: Indeed, I think I'll revert this one as well for the same reason as in https://github.com/github/codeql/pull/10366#discussion_r968459935.

cpp/ql/lib/semmle/code/cpp/models/interfaces/FunctionInputsAndOutputs.qll
MathiasVP: Yeah, sorry. I've added a small amount of comments to it now. Here's the backstory: Currently on `main`, we cannot model a function like:
```cpp
int f(int** p) {
  return **p;
}
```
since we cannot say that there's flow from `**p` to the return value (we can only say that there's flow from `*p` by using `isParameterDeref(0)`). The use-use flow PR gives us the ability to model this correctly by adding another parameter to `isParameterDeref` that specifies the number of indirections. So we can say that there's flow from `*p` by using `isParameterDeref(0, 1)`, or flow from `**p` by using `isParameterDeref(0, 2)`.

The use-use flow PR requires these two argument predicates to be present (so that the code compiles), but all the code in non-experimental main will still just use the old predicates.

Does that make sense?

cpp/ql/lib/semmle/code/cpp/models/implementations/StdString.qll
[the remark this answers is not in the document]
MathiasVP: Done in bb1c088fe08330879da7aa74dcbf8751a02674eb.

cpp/ql/lib/semmle/code/cpp/models/interfaces/FunctionInputsAndOutputs.qll
geoffw0: Yes, that's fantastic, thanks for expanding on this!

geoffw0 (approved):
I'm happy with the changes to models / everything outside of experimental now.  I haven't reviewed the changes within experimental in any detail, but I understand it to be a modified version of the dataflow library that's been reviewed elsewhere, and will be under examination in `experimental`.  I can't see anything scary in the DCA run.

:+1:

MathiasVP:
> I'm happy with the changes to models / everything outside of experimental now. I haven't reviewed the changes within experimental in any detail, but I understand it to be a modified version of the dataflow library that's been reviewed elsewhere, and will be under examination in experimental. I can't see anything scary in the DCA run.

Thanks a lot! Yeah, all of the code in experimental has already been thoroughly reviewed in the initial PR. So this should be pretty safe

PR #10369, opened 2022-09-10
Ruby: add `rb/sensitive-get-query` query

alexrford:
Finds cases where an HTTP GET request handler takes sensitive input, such as a password or other credential, from the query string of the request. The bulk of this PR concerns porting `SensitiveNode` to Ruby.

This uses for `HTTP::Server::RequestHandler`s, which is currently implemented for Rails `ActionController` and the ruby `graphql` gem. In the case of GraphQL, both `POST` and `GET` requests are by default both accepted and handled uniformly - the `string getAnHttpMethod()` implementation is trivial here as I couldn't find a documented way to change this behaviour.


ruby/ql/src/change-notes/2022-09-10-sensitive-get-query.md
```diff
@@ -0,0 +1,4 @@
+---
+category: newQuery
+---
```
ThanhVy1121, line 3: 1

ruby/ql/lib/codeql/ruby/security/SensitiveActions.qll
```diff
@@ -11,6 +11,135 @@
...
+}
+
+/** A method name that might return sensitive credential data. */
+class CredentialsMethodName extends SensitiveDataMethodName {
+  SensitiveDataClassification classification;
+
+  CredentialsMethodName() { nameIndicatesSensitiveData(this, classification) }
+
+  override SensitiveDataClassification getClassification() { result = classification }
+}
```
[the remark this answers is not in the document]
alexrford: Yeah, I think that would be sensible. When I was working on this PR, I ran into some issues where the JS version of this class had been moved from the AST layer to the DataFlow layer. I only noticed this change by chance, but it would have been easier to notice and stay synched on if there had been some shared Concept used by both.

ruby/ql/src/queries/security/cwe-598/SensitiveGetQuery.ql
```diff
@@ -0,0 +1,44 @@
...
+  src.flowsTo(to)
+  or
+  exists(DataFlow::Node midRecv, DataFlow::LocalSourceNode mid, ElementReference ref |
+    src.flowsTo(midRecv) and
+    midRecv.asExpr().getExpr() = ref.getReceiver() and
+    mid.asExpr().getExpr() = ref
+  |
+    localFlowWithElementReference(mid, to)
+  )
+}
```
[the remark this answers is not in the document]
alexrford: I should probably try this out with a full `TaintTracking` configuration as well to see if there's any difference in performance or results. In the JS version it seems like using global dataflow/taint tracking didn't improve results, but this may be different for Ruby. FWIW Java uses full taint tracking for its version of this query.

alexrford:
Finally got around to updating this PR. There are a couple of main changes:
- Switched to using full taint tracking rather than local flow. I've kicked off a DCA run to see how this affects results/performance. I've used the Query/Customization pattern to implement this, which leaves us with an unfortunately named `SensitiveGetQueryQuery.qll` library. I converted the query to a `path-problem` as part of this change, which is not consistent with the JS version, but it seems worth using the path graph if we have it.
- Fixed an issue where data from headers in the request headers were also treated as if they were transmitted as plaintext, where the other versions of this query are correctly limited to just query params. This required adding a `string RequestInputAccess#getKind()` predicate.

alexrford:
I've rolled the taint tracking changes back since this had a fairly large performance impact and the results that it found should all be picked up with local flow anyway.

alexrford:
> I've rolled the taint tracking changes back since this had a fairly large performance impact and the results that it found should all be picked up with local flow anyway.

I spoke too soon - the performance impact seems to be due to something else. I'll look into this properly tomorrow.

alexrford:
I think the bulk of the performance impact (~3.5% increase in analysis time) is from computing the new `SensitiveNode`s. I've not been able to make any significant improvements here. I'm conflicted over if this regression is acceptable - the `SensitiveNode` supports 5 different queries in JS of which I think 4 can be ported to Ruby whilst reusing this class. As it stands we only use this class for `rb/sensitive-get-query` in this PR.

ruby/ql/src/queries/security/cwe-598/SensitiveGetQuery.qhelp
```diff
@@ -0,0 +1,43 @@
+<!DOCTYPE qhelp PUBLIC "-//Semmle//qhelp//EN" "qhelp.dtd">
+<qhelp>
+<overview>
+<p>
+Sensitive information such as user passwords should not be transmitted within the query string of the requested URL.
```
aibaars, line 5: Did you mean `password` or `user names and passwords`. While `user passwords` is a correct term, I think transmitting non-user passwords would be equally problematic.

ruby/ql/lib/codeql/ruby/Concepts.qll
```diff
@@ -304,6 +314,14 @@ module Http {
        * This is typically the name of the method that gives rise to this input.
        */
       string getSourceType() { result = super.getSourceType() }
+
+      /**
+       * Gets the kind of the accessed input,
+       * Can be one of "parameter", "header", "body", "url", "cookie".
+       *
+       * Note that this predicate is functional.
```
aibaars, line 322: What is the purpose of this `Note` ? I guess `functional` in this context indicates that the predicate returns exactly one result. This is kind of implied in the name of the predicate `get` instead of `getA` and the `Gets the` and lack of `if any` suffix in the comment. 

I think the note may confuse users and make them wonder if there are any predicates that do not work/function somehow.

ruby/ql/lib/codeql/ruby/Concepts.qll
```diff
@@ -321,6 +339,14 @@ module Http {
          * This is typically the name of the method that gives rise to this input.
          */
         abstract string getSourceType();
+
+        /**
+         * Gets the kind of the accessed input,
+         * Can be one of "parameter", "header", "body", "url", "cookie".
+         *
+         * Note that this predicate is functional.
```
aibaars, line 347: same here.

ruby/ql/lib/codeql/ruby/security/SensitiveActions.qll
```diff
@@ -11,6 +11,135 @@
...
+/**
+ * Holds if `node` is a write to a variable or hash value named `name`.
+ *
+ * Helper predicate factored out for performance,
+ * to filter `name` as much as possible before using it in
+ * regex matching.
+ */
+pragma[nomagic]
+private predicate writesProperty(DataFlow::Node node, string name) {
+  exists(VariableWriteAccess vwa | vwa.getVariable().getName() = name |
```
aibaars, line 78: If I recall correctly instance and class variables have `@` and `@@` prefixes in their names, do we need to account for those here?

ruby/ql/test/query-tests/security/cwe-598/app/controllers/users_controller.rb
```diff
@@ -0,0 +1,22 @@
+class UsersController < ApplicationController
+
+  def login_get
+    password = params[:password] # BAD: route handler uses GET query parameters to receive sensitive data
```
aibaars, line 4: could you also add a test case with an instance variable? Like 
```
@password = params[:some_thing]
authenticate_user(params[:username], @password)
```

aibaars, line 4: It think you best change the test case in one that uses `params[:password]` but with a non-sensitive variable name, and another with a non-sensitive query parameter name and a sensitive variable name.

ruby/ql/lib/codeql/ruby/Concepts.qll
alexrford: I'm going to back out of the `kind` related changes here as https://github.com/github/codeql/pull/10602 implements this in a neater way. FWIW this was more or less copied from the JS version of this, my understanding of `functional` in this sense was that the return values could be relied on to have a precise semantic meaning. This is in contrast to `string getSourceType()` which is a bit fuzzier in what it might return.

ruby/ql/src/queries/security/cwe-598/SensitiveGetQuery.qhelp
alexrford: Yeah, just `passwords` makes more sense here. We don't report user names here as they're not considered sensitive in this context.

ruby/ql/lib/codeql/ruby/security/SensitiveActions.qll
alexrford: Sorry, I take that back - the heuristics exclude anything containing non-alphanumerics. It might be worth changing `nameIndicatesSensitiveData` to ignore leading `@` characters as we deal with variable names in multiple places in this file.

alexrford:
I'll update this again tomorrow to reflect the `kind` changes from the other PR before running DCA again.

ruby/ql/lib/codeql/ruby/Concepts.qll
alexrford: > Sorry for conflicting with your changes here! When we talked about it in the meeting, I misunderstood and thought your changes were already merged, so when I couldn't find them I assumed they were for a different concept and so I would have to add them in my PR.

No worries, thanks for the improved implementation. I've updated this PR to use it.

ruby/ql/lib/codeql/ruby/Concepts.qll
alexrford: This is gone now due to pulling in the changes from the other PR.

ruby/ql/lib/codeql/ruby/security/SensitiveActions.qll
alexrford: `nameIndicatesSensitiveData` is in a shared file dealing with cross language heuristics, so I've updated the places in this file that deal with variable accesses instead.

alexrford:
Updated to address review comments. DCA has a couple of FPs that are interesting, as they look to be safe by virtue of appearing in an if statement body that checks for `request.post?` or `request.get?`.

aibaars (approved):
LGTM, perhaps we should implement some sanitizers/barriers for `if request.post?` checks.

alexrford:
> LGTM, perhaps we should implement some sanitizers/barriers for `if request.post?` checks.

Do you think that this is a blocker for the current PR?

aibaars:
> > LGTM, perhaps we should implement some sanitizers/barriers for `if request.post?` checks.
> 
> Do you think that this is a blocker for the current PR?

No problem ;-)

PR #10375, opened 2022-09-12
Ruby: type-tracking and API edges through simple library callables

asgerf:
This PR ensures that simple library callables whose callee does not depend on API graphs can be seen by type-tracking and API graphs. "Simple" in this case means it can be reduced to a store or load edge at the call site, though in the future we may want to add load-store and basic edges as well.

For example, a summary of form `Argument[0].Element[0];ReturnValue` will be seen as:
- a load step in type-tracking, reading the content set corresponding to `Element[0]`, and
- an edge in the API graph labelled with the content corresponding to `Element[0]`.

Ruby is the first language to use the string-based flow summary approach combined with type-tracking and API graphs, but the two systems hadn't been fully integrated. In particular, Ruby has so far been modelling the standard library using summaries that are invisible to type-tracking and API graphs - this would lead to incompleteness compared to just generating the flow edges directly. I believe this PR bridges the gap, so we can get the best of both worlds while modelling everything with flow summaries.

[Evaluation](https://github.com/github/codeql-dca-main/issues/7203) shows 216 new call edges, at a modest performance cost. I've investigated the performance overhead a bit, but couldn't get it any lower than it is now - partly because we are just tracking more flow now.

I'm running an evaluation for Python as well, since some shared code was affected.


ruby/ql/lib/codeql/ruby/typetracking/TypeTracker.qll
```diff
@@ -33,48 +12,72 @@ private module Cached {
...
     )
+    or
+    exists(TypeTrackerContentSet contents, boolean hasReturn |
+      step = StoreStep(pragma[only_bind_into](contents)) and
+      tbt = MkTypeBackTracker(hasReturn, contents.getAReadContent()) and
+      result = noContentTypeBackTracker(hasReturn)
+      or
+      step = LoadStep(pragma[only_bind_into](contents)) and
+      tbt = noContentTypeBackTracker(hasReturn) and
+      result = MkTypeBackTracker(hasReturn, contents.getAStoreContent())
```
hvitved, line 79: This should be `getAReadContent`.

ruby/ql/lib/codeql/ruby/typetracking/TypeTrackerSpecific.qll
```diff
@@ -136,50 +140,65 @@ predicate returnStep(Node nodeFrom, Node nodeTo) {
...
  * `Foo` object created on the first line of the function. This means we will
  * track the fact that `x.content` can have the type of `y` into the assignment
  * to `z` inside `bar`, even though this content write happens _after_ `bar` is
  * called.
  */
-predicate basicStoreStep(Node nodeFrom, Node nodeTo, string content) {
+predicate basicStoreStep(Node nodeFrom, Node nodeTo, TypeTrackerContentSet contents) {
+  postUpdateStoreStep(nodeFrom, nodeTo, contents)
+  or
+  exists(
```
hvitved, line 153: You should be able to use `summarySetterStep` from `FlowSummaryImpl` instead.

ruby/ql/lib/codeql/ruby/typetracking/TypeTrackerSpecific.qll
```diff
@@ -136,50 +140,65 @@ predicate returnStep(Node nodeFrom, Node nodeTo) {
...
+predicate basicLoadStep(Node nodeFrom, Node nodeTo, TypeTrackerContentSet contents) {
   exists(ExprNodes::MethodCallCfgNode call |
     call.getExpr().getNumberOfArguments() = 0 and
-    content = call.getExpr().getMethodName() and
+    contents.isSingleton(DataFlowPublic::Content::getAttributeName(call.getExpr().getMethodName())) and
     nodeFrom.asExpr() = call.getReceiver() and
     nodeTo.asExpr() = call
   )
+  or
+  exists(
```
hvitved, line 193: You should be able to use `summaryGetterStep` from `FlowSummaryImpl` instead.

ruby/ql/test/library-tests/modules/calls.rb
```diff
@@ -126,13 +126,13 @@ def funny
 1.bit_length
 1.abs
 
-["a","b","c"].foreach { |i, v| puts "#{i} -> #{v.capitalize}"} # TODO should resolve to String.capitalize
+["a","b","c"].foreach { |i, v| puts "#{i} -> #{v.capitalize}"}
```
hvitved, line 129: Great that this is fixed, why is that?

hvitved (changes requested):
Looks great Asger! A few comments.

ruby/ql/lib/codeql/ruby/ApiGraphs.qll
```diff
@@ -257,6 +258,29 @@ module API {
      */
     Node getAnImmediateSubclass() { result = this.getASuccessor(Label::subclass()) }
 
+    /**
+     * Gets a node representing the `content` stored on the base object.
+     */
+    Node getContent(DataFlow::Content content) {
+      result = this.getASuccessor(Label::content(content))
```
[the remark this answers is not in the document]
asgerf: It is done like this exactly to avoid the fan-out from `getAReadContent`.

Here's a break down of how the graph is generated and what ends up being matched:
| Description | Edge label  | Matched by  |
|---|---|---|
| Use node reading a known element `n`   | `TKnownElementContent(n)`  | `Element[n]` and `Element[any]`  |
| Def node storing a known element `n`   | `TKnownElementContent(n)`  | `Element[n]` and `Element[any]`  |
| Use node reading an unknown element   | `TUnknownElementContent`  | `Element[?]` and `Element[any]`  |
| Def node storing an unknown element   | `TUnknownElementContent`  | `Element[?]` and `Element[any]`  |

`Element[n]` does not match `TUnknownElementContent`, and the model must thus use `Element[?,n]` if this behaviour is wanted. This seems consistent with how existing Ruby models are required to mention `Element[?]` explicitly.

I believe this is the best integration we can hope for without changing how `ContentSet` works.

I tried storing `ContentSet` on the edge label but it doesn't really work out nicely. `getAReadContent()` returns a `Content`, but the edges are labelled with a `ContentSet` so we have to map it back to a `ContentSet` to look up in the `edge` relation. To ensure an efficient join there we'd end up factoring out a relation indexed on the `Content`, which basically corresponds the edge relation we're using now.

ruby/ql/lib/codeql/ruby/ApiGraphs.qll
```diff
@@ -539,27 +579,12 @@ module API {
     /** Gets a node reachable from a use-node. */
     private DataFlow::LocalSourceNode useCandFwd() { result = useCandFwd(TypeTracker::end()) }
 
-    private DataFlow::Node useCandRev(TypeBackTracker tb) {
```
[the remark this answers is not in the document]
asgerf: After applying the changes in this PR, it didn't perform well anymore.

The pipeline looked like this on `openproject` (manually aborted after a while)
```
(310s) Tuple counts for ApiGraphs#3116a2f3::API::Impl::useNodeStep#3#ffff/4@022f8a1e after 3m48s:
189468     ~0%      {2} r1 = JOIN ApiGraphs#3116a2f3::API::Impl::useCandRev#1#ff_10#join_rhs WITH DataFlowPrivate#462ff392::Cached::isLocalSourceNode#1#f ON FIRST 1 OUTPUT Lhs.1, Lhs.0 'result'
189468     ~0%      {3} r2 = JOIN r1 WITH num#TypeTracker#6f8aa30b::MkTypeBackTracker#fff_2#join_rhs ON FIRST 1 OUTPUT Lhs.1 'result', Lhs.1 'result', Lhs.0
51903      ~3%      {4} r3 = JOIN r2 WITH TypeTracker#6f8aa30b::Cached::stepCall#3#fff_102#join_rhs ON FIRST 1 OUTPUT Lhs.2, Lhs.1 'result', Rhs.1 'mid', Rhs.2
49779      ~2%      {4} r4 = JOIN r2 WITH TypeTracker#6f8aa30b::Cached::stepNoCall#3#fff_102#join_rhs ON FIRST 1 OUTPUT Lhs.2, Lhs.1 'result', Rhs.1 'mid', Rhs.2
101682     ~0%      {4} r5 = r3 UNION r4
189161     ~6%      {4} r6 = JOIN r5 WITH TypeTracker#6f8aa30b::TypeBackTracker::getACompatibleTypeTracker#0#dispred#ff ON FIRST 1 OUTPUT Lhs.3, Rhs.1 't', Lhs.1 'result', Lhs.2 'mid'
1322764717 ~12%     {4} r7 = JOIN r6 WITH TypeTracker#6f8aa30b::Cached::append#2#fff_120#join_rhs ON FIRST 2 OUTPUT Lhs.3 'mid', Rhs.2 'tmid', Lhs.1 't', Lhs.2 'result'
                    return r7
```

When I reverted to idiomatic type tracking, it seemed to perform quite well again, as witnessed by the performance evaluation.

[This evaluation](https://github.com/github/codeql-dca-main/tree/data/asgerf/summarize-loads-trackUseNodeComparison/reports) shows the effect of the commit that reverts to idiomatic type tracking (i.e. the whole branch without that commit, versus the brach with that commit).

We could of course try to hack on the join ordering again, but I couldn't find an example where the pruning seemed beneficial. If you remember a specific project where the pruning was important, please let me know and I'll investigate.

ruby/ql/lib/codeql/ruby/typetracking/TypeTracker.qll
```diff
@@ -33,48 +12,72 @@ private module Cached {
...
-      or
-      step = StoreStep(content) and result = MkTypeBackTracker(hasReturn, "")
-      or
       step = JumpStep() and
       result = MkTypeBackTracker(false, content)
     )
+    or
+    exists(TypeTrackerContentSet contents, boolean hasReturn |
+      step = StoreStep(pragma[only_bind_into](contents)) and
+      tbt = MkTypeBackTracker(hasReturn, contents.getAReadContent()) and
```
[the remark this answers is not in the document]
asgerf: As discussed offline, this was done to avoid the fan-out in `getAReadContent()`. However, it caused false negatives when backtracking. I have added a test that produces the missed backtracking flow.

I have changed type trackers to store a `ContentSet` instead, and added `compatibleContents(ContentSet, ContentSet)` to explicitly match up the content sets of a store and a load.

ruby/ql/lib/codeql/ruby/typetracking/TypeTracker.qll
asgerf: See answer above.

ruby/ql/lib/codeql/ruby/typetracking/TypeTrackerSpecific.qll
asgerf: That gives us a step inside the synthesised callable, whereas we need a step at the call site.

If we start doing bottom-up summarisation of type-tracking steps then we could use `summarySetterStep` instead. However, API graphs would still only want the step at the original call site so the logic is still needed somewhere.

ruby/ql/lib/codeql/ruby/typetracking/TypeTrackerSpecific.qll
asgerf: See answer above.

ruby/ql/test/library-tests/modules/calls.rb
asgerf: Turns out it was fixed for the wrong reasons. The summary seems to depend on having store-load steps so I'll revisit it when adding those (which are relatively simple in type-tracking but more complex in API graphs, so I'd rather not handle it in this PR).

ruby/ql/lib/codeql/ruby/dataflow/internal/DataFlowPrivate.qll
```diff
@@ -383,7 +385,14 @@ private module Cached {
...
       )
-    }
+    } or
+    // Only used by type-tracking
+    TAttributeName(string name) { name = any(AST::SetterMethodCall c).getTargetName() }
+
+  cached
+  class TContent =
+    TKnownElementContent or TUnknownElementContent or TKnownPairValueContent or
+        TUnknownPairValueContent or TFieldContent or TAttributeName;
```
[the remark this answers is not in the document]
asgerf: Is it a problem to have it here? I don't really see how it would help to have a new union type for this.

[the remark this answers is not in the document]
asgerf: Actually now that type-trackers contain `ContentSet` instead of `Content`, it's actually not that easy to factor out anymore.

ruby/ql/lib/codeql/ruby/dataflow/internal/DataFlowPublic.qll
```diff
@@ -235,15 +235,18 @@ predicate localExprFlow(CfgNodes::ExprCfgNode e1, CfgNodes::ExprCfgNode e2) {
   localFlow(exprNode(e1), exprNode(e2))
 }
 
-/** A reference contained in an object. */
-class Content extends TContent {
+/** A reference contained in an object, or the `noContent()` value. */
+class OptionalContent extends TOptionalContent {
```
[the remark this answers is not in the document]
asgerf: Moved to `TypeTrackerSpecific` 👍

asgerf:
[New evaluation](https://github.com/github/codeql-dca-main/issues/7203)

ruby/ql/lib/codeql/ruby/ApiGraphs.qll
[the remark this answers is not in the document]
hvitved: FWIW, `useNodeStep` only [got introduced recently](https://github.com/github/codeql/pull/10141), and I was never convinced that it was right to join with `append` first. I wonder what it would look like if that PR got reverted.

ruby/ql/lib/codeql/ruby/typetracking/TypeTrackerSpecific.qll
hvitved: `summarySetterStep` is a step at the call site.

ruby/ql/lib/codeql/ruby/ApiGraphs.qll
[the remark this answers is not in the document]
hvitved: Thanks for the detailed explanation.

ruby/ql/lib/codeql/ruby/dataflow/internal/DataFlowPrivate.qll
```diff
@@ -378,12 +378,16 @@ private module Cached {
...
   cached
-  newtype TContentSet =
+  newtype TOptionalContentSet =
     TSingletonContent(Content c) or
     TAnyElementContent() or
     TElementLowerBoundContent(int lower) {
       FlowSummaryImplSpecific::ParsePositions::isParsedElementLowerBoundPosition(_, lower)
-    }
+    } or
+    TNoContentSet()
```
hvitved, line 386: Couldn't this rather be defined as a `newtype` inside `TypeTracker.qll`:
```ql
newtype TOptionalContentSet =
  TContentSetSome(ContentSet cs) or
  TContentSetNone()
```

ruby/ql/lib/codeql/ruby/ApiGraphs.qll
```diff
@@ -268,6 +268,7 @@ module API {
     /**
      * Gets a node representing the `contents` stored on the base object.
      */
+    pragma[inline]
```
hvitved, line 271: I wouldn't expect `getAReadContent` to have large fan-in (only large fan-out).

ruby/ql/lib/codeql/ruby/dataflow/internal/DataFlowPrivate.qll
asgerf: You're asking if I want to revert https://github.com/github/codeql/pull/10375/commits/b1c18986483520926ea004c81d7b8b7bebbef983?

Repeatedly boxing values is just less efficient than putting them in the same universe. The `Optional` wrapper is the most obvious one to avoid.

ruby/ql/lib/codeql/ruby/ApiGraphs.qll
asgerf: As discussed in chat, this just gives a better join order, it's not that it's too large to materialize.

ruby/ql/lib/codeql/ruby/dataflow/internal/DataFlowPublic.qll
```diff
@@ -314,6 +317,34 @@ module Content {
   class UnknownPairValueContent extends PairValueContent, TUnknownPairValueContent {
     override string toString() { result = "pair" }
   }
+
+  /**
```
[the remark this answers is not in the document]
asgerf: I'm not sure which class you're referring to
- `TOptionalContentSet` is entirely hidden in `DataFlowPrivate`.
- `OptionalTypeTrackerContent` is defined in `TypeTrackerSpecific.qll`
- `Content::Attribute` is not hidden as it can be wrapped in a `ContentSet`. We could hide the class, but the values can still leak out to the user.

ruby/ql/lib/codeql/ruby/typetracking/TypeTrackerSpecific.qll
asgerf: Ah, sorry I was too a bit quick to conclude what was going on in that predicate.

It does seem to include taint steps though, which we don't want here.

Would you prefer if I add `boolean valuePreserving` as a parameter to that predicate so we can use it from here?

ruby/ql/lib/codeql/ruby/ApiGraphs.qll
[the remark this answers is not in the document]
asgerf: Reverting my commit and then reverting that PR seems to handle `openproject` without blowing up, albeit with larger tuple counts than with idiomatic type-tracking. I've [started an evaluation](https://github.com/github/codeql-dca-main/issues/7299) of the two branches to see which performs better.

asgerf:
We've got quite a few unresolved threads open. @hvitved could you 'mark as resolved' any conversations you feel have been addressed.

From my perspective the two outstanding items are:
- [trackUseNode pruning](https://github.com/github/codeql/pull/10375#discussion_r979958693): Waiting for [this evaluation](https://github.com/github/codeql-dca-main/issues/7299).
- [summarySetterStep](https://github.com/github/codeql/pull/10375#discussion_r979890000): Waiting for your feedback in  on whether to add `boolean valuePreserving` to the relevant predicates.

ruby/ql/lib/codeql/ruby/typetracking/TypeTrackerSpecific.qll
hvitved: 👍

ruby/ql/lib/codeql/ruby/typetracking/TypeTrackerSpecific.qll
hvitved: Remember to also update this bit.

hvitved:
> @hvitved could you 'mark as resolved' any conversations you feel have been addressed.

Done.

ruby/ql/lib/codeql/ruby/ApiGraphs.qll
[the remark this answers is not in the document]
asgerf: Evaluation was messed up, trying again [here](https://github.com/github/codeql-dca-main/issues/7328).

ruby/ql/lib/codeql/ruby/typetracking/TypeTrackerSpecific.qll
asgerf: Unfortunately this leads to non-monotonic recursion, since that predicate ultimately depends on `SummarizedCallable.getACall` whereas we need something that only depends on `getACallSimple`.

ruby/ql/lib/codeql/ruby/ApiGraphs.qll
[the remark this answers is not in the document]
asgerf: [Evaluation](https://github.com/github/codeql-dca-main/issues/7334) comparing pruning with https://github.com/github/codeql/pull/10141 reverted, versus idiomatic type-tracking with no pruning.

The evaluation does not show any benefit from pruning. It's possible that it originally capitalised on incompleteness, and as the call graph and API graph are now more complete, many more nodes can now be tracked to a use site.

asgerf:
Force-pushed to remove some revert commits that I had accidentally pushed as part of the investigation in [this thread](https://github.com/github/codeql/pull/10375#discussion_r979958693).

ruby/ql/lib/codeql/ruby/typetracking/TypeTrackerSpecific.qll
hvitved: Ah, right.

asgerf:
Rebased to fix conflict in `DataFlowDispatch.qll`: the call to `getACallSimple` now appears in `viableLibraryCallable`.

ruby/ql/lib/codeql/ruby/ApiGraphs.qll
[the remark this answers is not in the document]
asgerf: After merging with https://github.com/github/codeql/pull/10574 I believe the table is updated as follows:
- the first two rows can also be matched by `Element[n!]`
- the last two rows can also be matched by `Element[k]` for any constant `k`

| Description | Edge label  | Matched by  |
|---|---|---|
| Use node reading a known element `n`   | `TKnownElementContent(n)`  | `Element[n]`, `Element[n!]`, and `Element[any]`  |
| Def node storing a known element `n`   | `TKnownElementContent(n)`  | `Element[n]`, `Element[n!]`, and `Element[any]`  |
| Use node reading an unknown element   | `TUnknownElementContent`  | `Element[k]` for any `k`, `Element[?]` and `Element[any]`  |
| Def node storing an unknown element   | `TUnknownElementContent`  | `Element[k]` for any `k`, `Element[?]` and `Element[any]`  |

- Previously it was the model's responsibility to mention `Element[?]` if unknown reads/stores should be found.
- Now models are now responsible for using `Element[n!]` if this behaviour is _not_ wanted.

Again this seems consistent with how Ruby models work after https://github.com/github/codeql/pull/10574, so I'm cautiously optimistic that everything just works after this change.

ruby/ql/lib/codeql/ruby/dataflow/internal/DataFlowPrivate.qll
```diff
@@ -373,18 +373,24 @@ private module Cached {
...
+  newtype TOptionalContentSet =
     TSingletonContent(Content c) or
     TAnyElementContent() or
     TKnownOrUnknownElementContent(Content::KnownElementContent c) or
     TElementLowerBoundContent(int lower, boolean includeUnknown) {
       FlowSummaryImplSpecific::ParsePositions::isParsedElementLowerBoundPosition(_, includeUnknown,
         lower)
-    }
+    } or
+    TNoContentSet()
```
hvitved, line 388: Perhaps add a `// Only used by type-tracking` comment here as well.

ruby/ql/lib/codeql/ruby/typetracking/TypeTrackerSpecific.qll
```diff
@@ -190,3 +227,53 @@ predicate basicLoadStep(Node nodeFrom, Node nodeTo, string content) {
...
+    exists(string name |
+      pos.isKeyword(name) and
+      result = call.getKeywordArgument(name)
+    )
+    or
+    pos.isBlock() and
+    result = call.getBlock()
+    or
+    pos.isSelf() and
+    result = call.getReceiver()
```
hvitved, line 274: I think this can be replaced with
```ql
argumentPositionMatch(call, result, pos)
```

hvitved (commented):
Two last comments (sorry for not catching earlier).

ruby/ql/lib/codeql/ruby/typetracking/TypeTrackerSpecific.qll
asgerf: Thanks! 👍  I was a little sad about having to do this here.

asgerf:
The evaluation shows a slightly increased cost for ruby/ruby after merging with https://github.com/github/codeql/pull/10574. I think part of the reason is that there are about twice as many content sets now, and ruby/ruby has a lot of content sets. I tried restricting the content sets we care about with [this commit](https://github.com/asgerf/codeql/commit/1c4c05810fc9a3d1f893b3b07d9781bfbf13c55e) and although the tuple counts showed a great reduction, [it did not seem to help](https://github.com/github/codeql-dca-main/tree/data/asgerf/rb/summarize-loads-v2-tt-domains/reports). I'd prefer to move ahead with the PR as it is.

hvitved:
> I tried restricting the content sets we care about with [this commit](https://github.com/asgerf/codeql/commit/1c4c05810fc9a3d1f893b3b07d9781bfbf13c55e) and although the tuple counts showed a great reduction, [it did not seem to help](https://github.com/github/codeql-dca-main/tree/data/asgerf/rb/summarize-loads-v2-tt-domains/reports). I'd prefer to move ahead with the PR as it is.

I do like that commit, in particular because `TStepSummary` is cached. We could even make the two branches even smaller via
```ql
private predicate compatibleStoreLoadContents(TypeTrackerContent storeContents, TypeTrackerContent loadContents) {
  basicStoreStep(_, _, storeContents) and
  basicLoadStep(_, _, loadContents) and
  compatibleContents(storeContents, loadContents)
}
...
    StoreStep(TypeTrackerContent content) { compatibleStoreLoadContents(content, _)  } or
    LoadStep(TypeTrackerContent content) { compatibleStoreLoadContents(_, content) } or
```

asgerf:
I measured the size of `StepSummary` with each change on `ruby/ruby`:

| Commit | Size |
|--------|--------|
| Baseline dc03557 | 4,111,472 |
| Size restriction from https://github.com/asgerf/codeql/commit/1c4c05810fc9a3d1f893b3b07d9781bfbf13c55e | 4,633 |
| Additional size restriction from @hvitved's comment | 4,428 |

Unfortunately neither version is compatible with `TypeTracker.startInContent` as a model might start type-tracking in a property that has a load in the database but not a store. The predicate isn't currently used in Ruby, but has uses in JS and Python.

I've pushed another commit that makes the restriction on type-tracker compatible with `startInContent` and start another evaluation (not that I expect it look better than last time).

hvitved (approved):
Thanks for you hard work, and patience, on this one!

asgerf:
[Evaluation](https://github.com/github/codeql-dca-main/tree/data/asgerf/summarize-loads__nightly__code-scanning__25/reports) looks similar to the last one.

The [last Python evaluation](https://github.com/github/codeql-dca-main/tree/data/asgerf/summarize-loads-python/reports) shows a median slow-down of 1% which is a bummer given that this PR does nothing for Python. We can't do langauge-dependent join order tweaks in shared code, and the best join order is slightly different since Ruby has `ContentSet` and Python doesn't.

asgerf:
[Another evaluation](https://github.com/github/codeql-dca-main/issues/7444) of Python including the latest changes shows the same result, and I got a 👍 from the Python team that we're ok with this.

PR #3980, opened 2020-07-27
JavaScript: Make access paths more reusable

max-schaefer:
As discussed, this factors out a few reusable predicates from the access-path computation. The API graphs currently don't use this yet, so there is no particular rush to merge this, I just wanted to put it up for visibility and discussion.

An [evaluation](https://github.com/max-schaefer/dist-compare-reports/blob/master/js/make-access-paths-more-reusable/report.md) (internal link) shows a slight performance degradation. The massive apparent slowdown on gecko for the first run was due to the base SHA timing out on `definitions.ql` and the tip SHA timing out on a later query; on rerun, both timed out on `definitions.ql`.

PR #10378, opened 2022-09-12
JS: expand localFieldStep to use access-paths, and build access-paths in more cases

erik-krogh:
Gets a TP/FP for CVE-2022-29078.  

[Latest performance evaluation shows a ~1.5% performance regression](https://github.com/github/codeql-dca-main/tree/data/erik-krogh/aliasFlow-try-more-threads-again__1/reports).  
And [an evaluation on default.yml](https://github.com/github/codeql-dca-main/tree/data/erik-krogh/aliasFlow__default__code-scanning__7/reports) shows about the same performance regression.   

I've battled the performance evaluations on this [for a while](https://github.com/github/codeql-dca-main/issues?q=is%3Aissue+aliasFlow), and now I'm opening this PR even though I'm not that satisfied with the performance.  
But I can't find any reason for the performance regression, and the performance regression disappears when I try to pinpoint it.   
(E.g. [this evaluation showed a performance regression](https://github.com/github/codeql-dca-main/tree/data/erik-krogh/aliasFlow__default__code-scanning__5/reports), but that disappeared when I did a [single-threaded experiment on the worst offenders](https://github.com/github/codeql-dca-main/tree/data/erik-krogh/aliasFlow-single-threaded/reports)).  

I think the performance regression is from us finding more flow, and thus lots of predicates get ever so slightly slower. 

[An evaluation on the meta queries show some nice new call-edges](https://github.com/github/codeql-dca-main/blob/data/erik-krogh/aliasFlow-with-meta/reports/alert-meta-comparison.md).


asgerf (commented):
I'd like to share a bit of background info in the access paths library:

Nodes should in principle only have one access path, but this rule is currently broken when the recursive call to `fromRhs` inside `fromRhs` returns a node that is itself a root node (usually an object literal on the RHS of an assignment):
```js
BASE.bar = { baz: x } // BAD: 'x' has two access paths, {}.baz and BASE.bar.baz
```

The solution I wanted to use for this is based on what @max-schaefer was doing in [this PR](https://github.com/github/codeql/pull/3980) - sadly I never got around to finishing it as it got bundled up with an ill-fated attempt for access paths to step through imports.

The idea was that once we have factored out `ref2ref` and `ref2rhs`, we can explicitly analyze the graph for cycles and branches. We then determine which object literals can be seen as an extension of another access path (and therefore should not be roots), and which assignments can safely be used to extend an access path.


---

I mention this because I believe the above solution would ultimately subsume the changes you're making to the access path computation, and it may be worth going straight to the general solution.

For this PR, I think the required change to access path computation can be summarized by this snippet:
```js
let temp = { baz: x }; // 'x' should have the access path BASE.bar.baz
BASE.bar = temp;
```

This PR focuses on the point in time in which the assignment to `baz` happens, but it's merely a way to ensure no cycles are created. If we add explicit cycle detection, none of this control flow analysis would be needed.

If you're up for it, I think we should give the other strategy a try, though let me know if you'd rather just go ahead with it like this.

asgerf (approved):

PR #10384, opened 2022-09-12
Python: Allow `CallNode.getArgByName` for keyword args after `**kwargs`

RasmusWL:
I was rather surprised when I learned we don't allow this currently. I think this is a minor thing, as it's not a very common pattern to pass your arguments like this, but no reason we shouldn't handle this correctly.

see https://github.com/github/codeql/blob/1c15fc5600c0a566c1938f6a4c84319a6bb0eba3/python/ql/lib/semmle/python/Exprs.qll#L185-L197

We might want to consider deprecating or changing `getKeyword`/`getAKeyword` on the AST `Call` class.


yoff (approved):
LGTM

PR #10387, opened 2022-09-12
Python: `getStarArg` gives first `*args` argument

RasmusWL:
Besides my intended use in the new call-graph PR, these member-predicates only seems to be used in this part of points-to:
https://github.com/github/codeql/blob/2436b060f1a330b02f333bf99d716bff1f02f0da/python/ql/lib/semmle/python/pointsto/PointsTo.qll#L1142-L1167

I couldn't see any reason that we should give up altogether if there are multiple `*args` arguments. Including the first one looks like a win to me!

### DIFF
--- a/python/ql/lib/change-notes/2022-09-12-getStarArg-first.md
+++ b/python/ql/lib/change-notes/2022-09-12-getStarArg-first.md
@@ -1,0 +1,4 @@
+---
+category: minorAnalysis
+---
+* `getStarArg` member-predicate on `Call` and `CallNode` has been changed for calls that have multiple `*args` arguments (for example `func(42, *my_args, *other_args)`): Instead of producing no results, it will always have a result for the _first_ such `*args` argument.
--- a/python/ql/lib/semmle/python/Exprs.qll
+++ b/python/ql/lib/semmle/python/Exprs.qll
@@ -245,10 +245,12 @@ class Call extends Call_ {
     result = count(Expr arg | arg = this.getAPositionalArg() and not arg instanceof Starred)
   }
 
-  /** Gets the tuple (*) argument of this call, provided there is exactly one. */
+  /** Gets the first tuple (*) argument of this call, if any. */
   Expr getStarArg() {
-    count(this.getStarargs()) < 2 and
-    result = this.getStarargs()
+    exists(int firstStarArgIndex |
+      firstStarArgIndex = min(int i | this.getPositionalArg(i) instanceof Starred | i) and
+      result = this.getPositionalArg(firstStarArgIndex).(Starred).getValue()
+    )
   }
 }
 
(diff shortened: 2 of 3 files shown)


RasmusWL:
I can see that this caused no tests to change. Normally, if I had been a reviewer, I would have asked for some tests to be written for this :thinking: However, this does affect my local version of #10148 -- I can certainly just put this change into that PR as well, I just thought it was nice to try to reduce the amount of stuff going on in the call-graph PR :sweat_smile:

yoff (commented):
This change clearly invalidates the origin computation you show, but we are of course moving away from that. Is giving the first arg the right thing here, though? Why not all of them? Or, why not add a new predicate `getFirstStarArg` with obvious behaviour if the first one is what you want?

RasmusWL:
> This change clearly invalidates the origin computation you show

It will actually not change anything, due to the implementation of `getPositionalArgumentCount()` seen below. So that's why no test results were changed by this! -- I'll admit I actually had expected to fix up some results, and was very surprised when I didn't have to do that.

https://github.com/github/codeql/blob/1c15fc5600c0a566c1938f6a4c84319a6bb0eba3/python/ql/lib/semmle/python/Exprs.qll#L242-L246

Actually, the code is also wrong, since `positional_argument_points_to` assumes that a `*args` always HAS to be the last argument, which isn't true.

> Is giving the first arg the right thing here, though? Why not all of them?

```py
func(*args1, *args2)
```

The problem with mapping the arguments in the code above to parameters is that if you don't know the size of `args1`, you don't know at what index to start putting in the elements from `args2`. But you can always know the _starting_ position of the _first_ `*args` argument.

So overall, it seemed much better to at least get the first one.

> Or, why not add a new predicate `getFirstStarArg` with obvious behaviour if the first one is what you want?

we can certainly do that as well... the change just seemed like it would always be an improvement of the old behavior. If you feel like it, we can certainly deprecate the old member-predicate, and introduce the new one (personally, I'm not feeling very strongly about any of this, probably 1/10, I just wanted the new behavior :sweat_smile:)

yoff:
> > This change clearly invalidates the origin computation you show
> 
> It will actually not change anything, due to the implementation of `getPositionalArgumentCount()` seen below. So that's why no test results were changed by this! -- I'll admit I actually had expected to fix up some results, and was very surprised when I didn't have to do that.
> 

Aha. Ok, so now I am much less concerned about the behaviour not being apparent from the name.
 
> Actually, the code is also wrong, since `positional_argument_points_to` assumes that a `*args` always HAS to be the last argument, which isn't true.

I think it might be correct, actually, given https://github.com/github/codeql/blob/2436b060f1a330b02f333bf99d716bff1f02f0da/python/ql/lib/semmle/python/Exprs.qll#L211-L220

It seems the whole points-to analysis is trying to only operate on positional arguments before the first *.

>  But you can always know the _starting_ position of the _first_ `*args` argument.
> So overall, it seemed much better to at least get the first one.

Yes, that makes sense. It does feel like we are violating some basic assumptions of the points-to analysis, but given that
- no test failed
- the only existing caller is unaffected
- we do not plan to write any new points-to code

I think it is fine to change this behaviour.

yoff (approved):
I am becoming convinced that this is probably for the better :-)

PR #10388, opened 2022-09-12
JS: recognize returning an instance of a class as exporting that class

erik-krogh:
Gets a TP/FP for CVE-2020-28471

[Evaluation hints towards a performance regression](https://github.com/github/codeql-dca-main/tree/data/erik-krogh/exportNew__nightly-old__code-scanning/reports).  
Which disappeared when I tried [the same evaluation single-threaded](https://github.com/github/codeql-dca-main/tree/data/erik-krogh/exportNew-single-threaded/reports) 🤷

### DIFF
--- a/javascript/ql/lib/semmle/javascript/PackageExports.qll
+++ b/javascript/ql/lib/semmle/javascript/PackageExports.qll
@@ -52,6 +52,16 @@ private DataFlow::Node getAValueExportedByPackage() {
     not isPrivateMethodDeclaration(result)
   )
   or
+  // module.exports.foo = function () {
+  //   return new Foo(); // <- result
+  // };
+  exists(DataFlow::FunctionNode func, DataFlow::NewNode inst, DataFlow::ClassNode clz |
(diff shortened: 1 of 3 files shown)


javascript/ql/lib/semmle/javascript/PackageExports.qll
```diff
@@ -52,6 +52,16 @@ private DataFlow::Node getAValueExportedByPackage() {
...
   )
   or
+  // module.exports.foo = function () {
+  //   return new Foo(); // <- result
+  // };
+  exists(DataFlow::FunctionNode func, DataFlow::NewNode inst, DataFlow::ClassNode clz |
+    func = getAValueExportedByPackage().getALocalSource() and inst = unique( | | func.getAReturn())
+  |
+    clz.getAnInstanceReference() = inst and
+    result = clz.getAnInstanceMethod()
```
asgerf, line 62:
```suggestion
    result = clz.getAnInstanceMember(_)
```
To include getters/setters?

asgerf (approved):

PR #10390, opened 2022-09-12
QL: add unmentioned guard class query

erik-krogh:
See https://github.com/github/codeql/pull/10381#discussion_r968650892.  

[Evaluation was uneventful](https://github.com/github/codeql-dca-main/tree/data/PR-10390-0-javascript/reports).


asgerf (approved):
Thanks!

PR #10393, opened 2022-09-13
Java: Model taint flow for java.net.URI constructors in tainted path queries

zbazztian:
This PR adds missing taint steps for `java.net.URI` constructor parameters, whose absence prevented us to detect results like [this](https://github.com/OWASP-Benchmark/BenchmarkJava/blob/1b7a3a5c30a449a67bba5c7a5ee6d0a991aa3a90/src/main/java/org/owasp/benchmark/testcode/BenchmarkTest00218.java#L92-L94). With these changes the `TaintedPath.ql` query detects 12 additional results on this project, 8 of which TPs, 4 FPs. Note that the FPs are hard to avoid, because they require quite elaborate taint flow, see [here](https://github.com/OWASP-Benchmark/BenchmarkJava/blob/1b7a3a5c30a449a67bba5c7a5ee6d0a991aa3a90/src/main/java/org/owasp/benchmark/testcode/BenchmarkTest02379.java#L81-L88).

`java.net.URI` has a few more unmodeled methods such as [`normalize()`](https://docs.oracle.com/en/java/javase/18/docs/api/java.base/java/net/URI.html#normalize()). I might model those in an additional PR.


zbazztian:
> I think that tainting the fragment is not an issue in most scenarios.

I presume you are hinting at the fact that it gets url-encoded. My reasoning for including it anyway is due to:
```java
import java.net.*;
public class URITest{
	public static void main(String[] args) throws Exception{
		URI uri = new URI("a", "b", "\" \"?!'<>\n\r/..#{}_-*&%");
		System.out.println(uri.getFragment());
	}
}
```
output:
```
> javac URITest.java  && java URITest
" "?!'<>
/..#{}_-*&%
```

I haven't modeled the `get*()` methods yet, but was planning to, perhaps I could add them in this PR.
Your suggestion to add an SSRF sanitizer sounds good to me, so if you are ok with it, I would go with that approach.

zbazztian:
Perhaps, your third option is the reasonable thing to do and would give me what I originally wanted, i.e. detecting specific path injections. The example
```java
import java.net.*;
public class URITest{
	public static void main(String[] args) throws Exception{
		URI uri = new URI("../x", "../x", "../x");
		System.out.println(uri.toString());
		uri = new URI("../x", "../x", "/../x", "../x");
		System.out.println(uri.toString());
		uri = new URI("../x", "../x", "/../x", "../x", "../x");
		System.out.println(uri.toString());
		uri = new URI("../x", "../x", "../x", 0, "/../x", "../x", "../x");
		System.out.println(uri.toString());
	}
}
```
yields:
```
../x:../x#../x
../x://../x/../x#../x
../x://..%2Fx/../x?../x#../x
../x://..%2Fx@../x:0/../x?../x#../x
```
so it would seem that all arguments pass path-related taint. I am happy to add my taint steps to `TaintedPath.ql` only for now.

atorralba:
I was not referring to the URL-encoding aspect, but rather what the resulting URIs mean. Take into account that the fact that `toString()` shows the path traversal sequences doesn't immediately mean that the result is a valid path that can be used in `File`, e.g.

```
jshell> new File(new URI("file", "/var", "../../../etc/passwd"));
|  Exception java.lang.IllegalArgumentException: URI has a fragment component
|        at File.<init> (File.java:427)
|        at (#6:1)

jshell> new File(new URI("file", "../../../etc/passwd", null));
|  Exception java.lang.IllegalArgumentException: URI is not hierarchical
|        at File.<init> (File.java:420)
|        at (#10:1)

jshell> new File(new URI("file://../../../", "/etc/passwd", null));
|  Exception java.lang.IllegalArgumentException: URI has an authority component
|        at File.<init> (File.java:425)
|        at (#12:1)

```

zbazztian:
What about:
```java
import java.net.*;
import java.io.File;
public class URITest{
	public static void main(String[] args) throws Exception{
		URI u = new URI("file:", "var", "../../../../../etc/passwd");
		System.out.println(new File(u.toString()).getCanonicalPath());
	}
}
```
output:
```
> pwd
/home/sebastian/tmp

> javac URITest.java && java URITest
/etc/passwd
```

atorralba:
Crucially, your example uses `getCanonicalPath()` and `toString()` instead of directly the `URI`.

Any change in those two conditions seems to break the file path:

```
jshell> new File(new URI("file", "var", "../../../../etc/passwd").toString()).isFile()
$19 ==> false

jshell> new File("/etc/passwd").isFile()
$20 ==> true

jshell> new File(new URI("file", "var", "../../../../etc/passwd").toString())
$22 ==> file:var#../../../../etc/passwd

jshell> new File(new File(new URI("file", "var", "../../../../etc/passwd").toString()).getCanonicalPath())
$21 ==> /etc/passwd
```

This is starting to look a little adhoc and context-dependent (maybe we should require a `getCanonicalPath()` step with flow states if we go through the `fragment` step?), so we again need to find a balance between FPs and FNs with these steps. Having real-world examples of missing flow (or spurious flow) caused by this would be nice to make the decision.

Also, manipulating the scheme is similarly difficult to exploit regarding path injection (other than changing the scheme to `file://` instead of something else to trick the application into accessing the filesystem, but that again is context-dependent).

zbazztian:
The parameter that is important to me and which fixes the false negatives reference above, is the `path` parameter. Would you mind if I add on that one as a taint step to the `TaintedPath` query? Everything else seems a bit tricky.

For the `scheme` parameter I still think that it carries taint:
```java
import java.net.*;
import java.io.File;
import java.nio.file.Path;
public class URITest{
	public static void main(String[] args) throws Exception{
		// unix
		URI u = new URI("file:///etc/passwd", "/..", null);
		System.out.println(new File(new File(u).getCanonicalPath()).exists());
		System.out.println(new File(u.normalize()).exists());
		System.out.println(Path.of(u).normalize().toFile().exists());

		// windows
		u = new URI("file:///c", null, null);
		System.out.println(new File(u).exists());
	}

}
```
output:
```
true
true
true
true
```

On Linux, injection success is predicated on the second argument starting with `/..` which might not be too frequent. However, note that there is a file name limit, which for my `limits.h` (Arch Linux) is `255` characters. This let's me blow up the application by triggering an IOException:
```
Exception in thread "main" java.io.IOException: File name too long
        at java.base/java.io.UnixFileSystem.canonicalize0(Native Method)
        at java.base/java.io.UnixFileSystem.canonicalize(UnixFileSystem.java:175)
        at java.base/java.io.File.getCanonicalPath(File.java:626)
        at URITest.main(URITest.java:8)
```
Depending on the file system, there might be other issues like maximum path lengths / truncation or denial of service attacks by injecting strings such as `/doesnotexist1/../doesnotexist2/../doesnotexist3/../doesnotexist4/../doesnotexist5/..` and so forth, ideally on network drives or other slow devices. Note that in that example we do not require `toString()`.

For the windows example we neither require `toString()` nor any of the normalization methods. One can just inject arbitrary drive letters.

That being said, I only care about the `path` argument for the time being.

atorralba:
Thanks for your detailed deep dive into this. I think the `path` parameter is much more straight-forward for the tainted path query, so that should be way easier to add and approve without an extensive FP/FN evaluation. `ssp` and `host` seem less problematic as well, if you also want to include those.

I'll be happy to approve that — we can think about the other, more nuanced parameters (`scheme`, `authority`, `userInfo`, `query`, `fragment`) with more detail in the future.

zbazztian:
I have now
* moved selected taint steps into the `TaintedPath.ql` query only, specifically
* the second (`ssp`) parameter of the `(String,String,String)` constructor,
* the fifth (`path`) parameter of the `(String,String,String,int,String,String,String)` constructor. Note that I did not include the second (`host`) parameter, since in this scenario a `:<port>` will be added after the host, which makes path injection more difficult.
* the second (`host`) and third (`path`) parameters of the `(String,String,String,String)` constructor,
* the third (`path`) parameter of the `(String,String,String,String,String)` constructor.

I moved these summaries into `TaintedPathCommon.qll` so that both the remote and the local version of the query can use those.

atorralba:
Unfortunately, we shouldn't use the CSV format for summaries if they aren't intended to be globally applicable. Those should be transformed into regular CodeQL taint steps and added to the relevant configuration(s) in `isAdditionalTaintStep`.

zbazztian:
What's the reasoning behind this? Performance or readability / extensibility?
Would it be ok to just add an implementation of `TaintPreservingCallable`?

zbazztian:
I have seen it being done [here](https://github.com/github/codeql/blob/main/java/ql/src/experimental/Security/CWE/CWE-020/Log4jJndiInjection.ql#L165-L175) and thought it would be ok.

atorralba:
> What's the reasoning behind this? Performance or readability / extensibility?

Performance. Normally, when you add CSV summaries, you also add a bidirectional import in `ExternalFlow.qll`, so that the summaries are added to the global context. Those apply to the data flow graph computation, which means that if you run some queries, calculate the graph, and then another query adds more summaries that weren't in the global context, you need to recompute the data flow graph again. That's why it's usually either global or nothing with CSV models.

> Would it be ok to just add an implementation of TaintPreservingCallable or AdditionalTaintStep?

Those also apply globally. In this case, we want steps that apply specifically to the tainted path query (or queries), and the only way of doing that is adding them to the `isAdditionalTaintStep` predicate in the configuration.

> I have seen it being done [here](https://github.com/github/codeql/blob/main/java/ql/src/experimental/Security/CWE/CWE-020/Log4jJndiInjection.ql#L165-L175) and thus thought it would be ok.

Experimental queries are a different case because they're not run by default and so we are more tolerant with the performance impact. But we always move those summaries to the global context when/if we promote the experimental query.

zbazztian:
Thanks for the clarification, will make the changes.

zbazztian:
I turned the taint steps into CodeQL. I created a common taint configuration class with the additional flow steps, which is used by both `TaintedPath.ql` and `TaintedPathLocal.ql`, but if need be I can add two separate `isAdditionalTaintStep()` predicates.
I was tempted to use some of the abstractions for `java.net.URI` in `Networking.qll`, but they didn't fit very well.

atorralba:
@zbazztian I pushed a commit with a little refactoring, I hope you don't mind 🙏. I introduced an abstract class `TaintedPathAdditionalTaintStep`, and separated the two taint configurations again, for two reasons:

* Creating an abstract class that is used in `isAdditionalTaintStep` is more extensible both for us and for users. Now anyone can add other taint steps to these configurations just by extending `TaintedPathAdditionalTaintStep` (e.g. in `Customizations.qll`).
* It's better to keep instances of data flow and taint tracking configurations disjunct for performance reasons. I was concerned that the shared hierarchy between those two configurations would add unnecessary path computations.

zbazztian:
That makes a lot of sense, thanks for those insights. I'll make sure to apply that in my next PR.

atorralba (approved):

java/ql/src/change-notes/2022-09-13-flow-model-for-uri-constructors.md
```diff
@@ -0,0 +1,4 @@
+---
+category: minorAnalysis
+---
+* Added taint model for arguments of `java.net.URI` constructors.
```
aschackmull, line 4: This change note is a bit imprecise - the new models are only for TaintedPath.ql

zbazztian: I added some more detail.

PR #10396, opened 2022-09-13
JS: Fix FP in js/regexp/always-matches

asgerf:
I noticed some FPs in vscode while investigating a timeout.

This query is not currently tagged as a security query, but it's one of those queries that could be security relevant in some contexts, so it's worth maintaining a low FP rate for it.


erik-krogh (approved):
👍  

But I think you need to update the `.expected` file.

erik-krogh (approved):

erik-krogh (approved):

PR #10398, opened 2022-09-13
C++: Further work on buffer-overflow queries

MathiasVP:
Now that we've merged use-use for flow C++ into experimental the work-in-progress on https://github.com/github/codeql/pull/10340 doesn't need to live in a separate branch anymore.

This PR just switches the product dataflow library to the experimental dataflow library (which is where we've merged the use-use flow into) and cherry-picks those changes from that PR.

@rdmarsh2 the final commit accepts a bunch of test changes to the query you're working on. They also seem to be happening in https://github.com/github/codeql/pull/10340. Do you mind taking over this branch and seeing if those are changes you're expecting?


rdmarsh2:
I can take it over - it looks like we're losing more results than I expected, so I may need to do some work on the query or the product flow library.

MathiasVP:
LGTM now! Unfortunately, since this is _my_ PR I can't actually approve it 😭. So feel free to self-apply and merge if CI goes green, @rdmarsh2.

rdmarsh2 (dismissed):

geoffw0 (approved):
:+1: for experimental.

geoffw0 (approved):

rdmarsh2 (approved):

PR #10405, opened 2022-09-13
update the style guide on alert-messages

erik-krogh:
and add a QL-for-QL query detecting some violations of the guidance. 

---- 

This updates the style guide [as suggested by @felicitymay](https://github.com/github/codeql/pull/10286#discussion_r968151465).   
I've tried to write down the findings from asking about how the alert-message should look, and my own findings from revising a lot of alert-messages. 

---- 

I would like a doc review on the changes to `docs/query-metadata-style-guide.md`, and whether the guidance seems reasonable for what an alert-message should look like.


ql/ql/src/queries/style/AlertMessage.ql
```diff
@@ -0,0 +1,140 @@
+/**
+ * @name Alert messages style violation
+ * @description Alert message that doesn't follow some aspect of the style guide.
+ *              See the style guide here: https://github.com/github/codeql/blob/main/docs/query-metadata-style-guide.md#alert-messages
+ * @kind problem
+ * @problem.severity warning
+ * @id ql/alert-messages-style-violation
```
esbena, line 7:
```suggestion
 * @id ql/alert-message-style-violation
```

ql/ql/src/queries/style/AlertMessage.ql
```diff
@@ -0,0 +1,140 @@
+/**
+ * @name Alert messages style violation
```
esbena, line 2:
```suggestion
 * @name Alert message style violation
```
I think we can keep this singular.

ql/ql/src/queries/style/AlertMessage.ql
```diff
@@ -0,0 +1,140 @@
...
+ * @description Alert message that doesn't follow some aspect of the style guide.
```
esbena, line 3: An alert
```suggestion
 * @description An alert message that doesn't follow the style guide is harder for end users to digest.
```

Follows the usual description style: `if ... then ...`. Maybe that is for another query.

esbena (commented):
In this case, I think a thorough documentation on each predicate is warranted, with at least one good and bad example. Then the docteam can even confirm that the query is doing the right thing.

ql/ql/src/queries/style/AlertMessage.ql
```diff
@@ -0,0 +1,140 @@
...
+    msg = "Avoid starting a link text with an indefinite article."
+    or
+    node = dontQuoteSubstitutions(_) and
+    msg = "Don't quote substitutions in alert messages."
+    or
+    node = wrongFlowsPhrase(_, "data") and
+    msg = "Use \"flows to\" instead of \"depends on\" in taint tracking queries."
+    or
+    node = wrongFlowsPhrase(_, "taint") and
+    msg = "Use \"depends on\" instead of \"flows to\" in data flow queries."
```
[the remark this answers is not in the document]
erik-krogh: Copilot was right, I was wrong 🤦

esbena (approved):
For once, the docteam should look at the predicate documentation. See https://github.com/github/codeql/pull/10405/commits/88f1d2ae38e6e3927e715db1f32ae3865194bba7 in particular. Review it like this: does the docstring express the intent of the style guide?

ql/ql/src/queries/style/AlertMessage.ql
```diff
@@ -0,0 +1,197 @@
+/**
+ * @name Alert message style violation
+ * @description An alert message that doesn't follow the style guide is harder for end users to digest.
+ *              See the style guide here: https://github.com/github/codeql/blob/main/docs/query-metadata-style-guide.md#alert-messages
```
kaeluka, line 4: The URL doesn't render as a link in the actual alerts and I need to copy/paste the URL to go there. Can that be fixed?

docs/query-metadata-style-guide.md
```diff
@@ -179,8 +179,13 @@ The select clause of each alert query defines the alert message that is displaye
 * The message should factually describe the problem that is being highlighted–it should not contain recommendations about how to fix the problem or value judgements.
 * Program element references should be in 'single quotes' to distinguish them from ordinary words. Quotes are not needed around substitutions (`$@`).
 * Avoid constant alert message strings and include some context, if possible. For example, `The class 'Foo' is duplicated as 'Bar'.` is preferable to `This class is duplicated here.`
+* If a reference to the current location can't be avoided use "this location" instead of "here". For example, `Bad thing at this location.` is preferable to `Bad thing here.`.
```
kaeluka, line 182: It's not obvious to me why 'this location' would be better. I don't mind the rule, because consistent writing is a good thing — but can you add a half sentence to explain the reasoning?

ql/ql/src/queries/style/AlertMessage.ql
```diff
@@ -0,0 +1,197 @@
...
+    result.getValue().matches(["% flows to %", "% flow to %"])
+  )
+}
+
+from AstNode node, string msg
+where
+  not node.getLocation().getFile().getAbsolutePath().matches("%/test/%") and
+  (
+    node = shouldHaveFullStop(_) and
+    msg = "Alert message should end with a full stop."
```
kaeluka, line 172: Love it! I remember asking someone (forgot who) a long time ago about whether or not the should be one at the end of the message and I got the answer "I don't know if there's a standard".

ql/ql/src/queries/style/AlertMessage.ql
```diff
@@ -0,0 +1,197 @@
...
+      or
+      part != "here" and
+      msg = "Try to avoid using the phrase \"" + part + "\" in alert messages if possible."
+    )
+    or
+    node = avoidArticleInLinkText(_) and
+    msg = "Avoid starting a link text with an indefinite article."
+    or
+    node = dontQuoteSubstitutions(_) and
+    msg = "Don't quote substitutions in alert messages."
```
kaeluka, line 189: I'm not sure this is a good rule. There seem to be cases where it makes sense to quote, eg.:

image

docs/query-metadata-style-guide.md
```diff
@@ -179,8 +179,13 @@ The select clause of each alert query defines the alert message that is displaye
 * The message should factually describe the problem that is being highlighted–it should not contain recommendations about how to fix the problem or value judgements.
 * Program element references should be in 'single quotes' to distinguish them from ordinary words. Quotes are not needed around substitutions (`$@`).
```
kaeluka, line 180: This seems to be in conflict with your proposed "Don't quote substitutions" rule. This says they're not *needed*, while your implementation says they're not *allowed*.

kaeluka (dismissed):
Overall, looks good. Some small remarks, I'm pretty open about how you're going to address them — which is why I'm giving an approval.

One bonus comment:

I noticed that an alert message like this `"Pointer arithmetic here may be done with the wrong type because of the cast $@.", source, "here"` triggers two instances of the alert. I guess one for each string constants.

But an alert like `"$@ flows to here and is used in arithmetic, potentially causing an " + effect + ".",` (where one string contains both `flows to` and `here`) triggers only one alert that contains both messages.

This is really minor and not worth making the query harder to read and I suspect that making this more consistent will be doable without making the implementation more complex; but I'm bringing it up in case it is possible easily.

erik-krogh:
> I noticed that an alert message like this `"Pointer arithmetic here may be done with the wrong type because of the cast $@.", source, "here"` triggers two instances of the alert. I guess one for each string constants.
> 
> But an alert like `"$@ flows to here and is used in arithmetic, potentially causing an " + effect + ".",` (where one string contains both `flows to` and `here`) triggers only one alert that contains both messages.

That's because the location I'm using is the string-constant that contains the style-violation.  
In the latter case both problems are contained in the same string-constant, so the alert-location is exactly the same.  
I could put the alert-location on the select, but that would break when the string-constant is outside the from-where-select.

ql/ql/src/queries/style/AlertMessage.ql
erik-krogh: I don't know of a way to do that.

docs/query-metadata-style-guide.md
erik-krogh: I remember someone suggesting it to avoid [the "click here" anti-pattern](https://www.lamar.edu/web-communication/resources/avoid-using-click-here.html), but now I can't find that review comment.   

I've inserted a short sentence referencing the "click here" anti-pattern in the end.

ql/ql/src/queries/style/AlertMessage.ql
erik-krogh: I also didn't know until I started this work.  
Now we have it codified!

ql/ql/src/queries/style/AlertMessage.ql
erik-krogh: Substitutions are already highlighted when shown, so most of the time it makes sense to not quote substitutions.  

The example you show is a counter-example to that, but there's a reason the precision is set to `warning` instead of `error`.

docs/query-metadata-style-guide.md
erik-krogh: Not quite. 
Quotes should be used around program text, when that program text is not a substitution.  

E.g. (for a simple query telling users to use StringBuilder instead of string concat). 
```CodeQL
select concat, "Use 'new StringBuilder()' instead of concatenating a large number of strings."
```

docs/query-metadata-style-guide.md
erik-krogh: I'm wondering a bit if we should keep this line.   

@aibaars suggested in a Slack comment that the "this location" phrasing got around the general ["click-here" anti-pattern](https://www.w3.org/QA/Tips/noClickHere).  
But as @smowton points out, it might not always make the text clearer: https://github.com/github/codeql/pull/10413#issuecomment-1246527443 

Maybe the guide should be that a link-text shouldn't be "here"?

smowton: In most cases could you just linkify the immediately preceding text, so something like

`This sensitive data is written to a logfile unescaped /here/` -> `This sensitive data is /written to a logfile unescaped/`?

smowton: Then you can keep `here` when it refers to "[some statement, expression, argument or result on] the line of this alert" and replace it when it's referring to a second thing involved in the alert.

docs/query-metadata-style-guide.md
```diff
@@ -179,8 +179,13 @@ The select clause of each alert query defines the alert message that is displaye
...
+* If a reference to the current location can't be avoided use "this location" instead of "here". For example, `Bad thing at this location.` is preferable to `Bad thing here.`. This avoids the "click here" anti-pattern.
 * Where you reference another program element, link to it if possible using a substitution (`$@`). Links should be used inline in the sentence, rather than as parenthesised lists or appositions.
 * When a message contains multiple links, construct a sentence that has the most variable link (that is, the link with most targets) last. For further information, see [Defining the results of a query](https://codeql.github.com/docs/writing-codeql-queries/defining-the-results-of-a-query/).
+* Make link texts as concise and precise as possible. E.g. avoid starting a link text with an indefinite article (a, an). For example `Path construction depends on a [user-provided value]` is preferable to `Path construction depends on [a user-provided value]`. (Where the square brackets indicate a link.) See [the W3C guide on link texts](https://www.w3.org/WAI/WCAG22/Understanding/link-purpose-in-context.html) for further information.
```
hubwriter, line 185:
```suggestion
* Make link text as concise and precise as possible. For example, avoid starting a link text with an indefinite article (a, an). `Path construction depends on a [user-provided value]` is preferable to `Path construction depends on [a user-provided value]`. (Where the square brackets indicate a link.) See [the W3C guide on link texts](https://www.w3.org/WAI/WCAG22/Understanding/link-purpose-in-context.html) for further information. 
```

docs/query-metadata-style-guide.md
```diff
@@ -179,8 +179,13 @@ The select clause of each alert query defines the alert message that is displaye
...
 * When a message contains multiple links, construct a sentence that has the most variable link (that is, the link with most targets) last. For further information, see [Defining the results of a query](https://codeql.github.com/docs/writing-codeql-queries/defining-the-results-of-a-query/).
+* Make link texts as concise and precise as possible. E.g. avoid starting a link text with an indefinite article (a, an). For example `Path construction depends on a [user-provided value]` is preferable to `Path construction depends on [a user-provided value]`. (Where the square brackets indicate a link.) See [the W3C guide on link texts](https://www.w3.org/WAI/WCAG22/Understanding/link-purpose-in-context.html) for further information. 
+* For path queries, if possible, try to follow the template: `This path depends on a [user-provided value].`, or alternatively (if the first option doesn't work) `[User-provided value] flows to this location and is used in a path.`. 
+* Taint tracking queries generally have that a sink "depends on" the source, and dataflow queries generally have a source that "flows to" the sink.
```
hubwriter, line 187: I think this should be:

```suggestion
* Taint tracking queries generally have a sink that "depends on" the source, and dataflow queries generally have a source that "flows to" the sink.
```

ql/ql/src/queries/style/AlertMessage.ql
```diff
@@ -0,0 +1,198 @@
...
+      order by
+        loc.getStartLine(), loc.getStartColumn(), loc.getEndLine(), loc.getEndColumn(),
+        loc.getFile().getRelativePath()
+    )
+}
+
+/**
+ * Gets a string element that is the last part of the message, that doesn't end with a full stop.
+ *
+ * E.g.
```
hubwriter, line 45:
```suggestion
 * For example:
```

ql/ql/src/queries/style/AlertMessage.ql
```diff
@@ -0,0 +1,198 @@
...
+      str order by i
+    ) and
+  not result.getValue().matches("%.") and
+  not result.getValue().matches("%?")
+}
+
+/**
+ * Gets a string element that is the first part of the message, that starts with a lower case letter.
+ *
+ * E.g.
```
hubwriter, line 66:
```suggestion
 * For example:
```

ql/ql/src/queries/style/AlertMessage.ql
```diff
@@ -0,0 +1,198 @@
...
+    |
+      str order by i
+    ) and
+  result.getValue().regexpMatch("^[a-z].*")
+}
+
+/**
+ * Gets a string element that is used in a message that contains "here" or "this location".
+ *
+ * E.g.
```
hubwriter, line 86:
```suggestion
 * For example:
```

ql/ql/src/queries/style/AlertMessage.ql
```diff
@@ -0,0 +1,198 @@
...
+  (
+    result.getValue().regexpMatch(".*\\b" + part + "\\b.*") and
+    result = getSelectPart(_, _)
+  )
+}
+
+/**
+ * Avoid using an indefinite article ("a" or "an") in a link text.
+ *
+ * E.g.
```
hubwriter, line 104:
```suggestion
 * For example:
```

ql/ql/src/queries/style/AlertMessage.ql
```diff
@@ -0,0 +1,198 @@
...
+String avoidArticleInLinkText(Select sel) {
+  result = sel.getExpr((any(int i | i > 1))) and
+  result = getSelectPart(sel, _) and
+  result.getValue().regexpMatch("a|an .*")
+}
+
+/**
+ * Don't quote substitutions in a message.
+ *
+ * E.g.
```
hubwriter, line 122:
```suggestion
 * For example:
```

ql/ql/src/queries/style/AlertMessage.ql
```diff
@@ -0,0 +1,198 @@
...
+    |
+      n
+      order by
+        loc.getStartLine(), loc.getStartColumn(), loc.getEndLine(), loc.getEndColumn(),
+        loc.getFile().getRelativePath()
+    )
+}
+
+/**
+ * Gets a string element that is the last part of the message, that doesn't end with a full stop.
```
hubwriter, line 43: "Full stop" is British English. I'd suggest changing this throughout this PR to the American English "period". I won't mark the other instances.

```suggestion
 * Gets a string element that is the last part of the message, that doesn't end with a period.
```

hubwriter (approved):
Docs team reviewer: 
I've added a few (minor) comments.
👍

docs/query-metadata-style-guide.md
erik-krogh: I've tried to revise the text a bit, and moved the the parts about link-texts into a separate sub-section.   

>  In most cases could you just linkify the immediately preceding text.  

Yes, and the guide now includes that 👍  

------ 

I still have the guidance about "this location" when you have an alert message without links, for when you can't get around mentioning the current location in the message.

docs/query-metadata-style-guide.md
kaeluka: 👍

ql/ql/src/queries/style/AlertMessage.ql
kaeluka: ok, fair enough

docs/query-metadata-style-guide.md
kaeluka: I'm happy with the message as it is now 👍

kaeluka (approved):
lgtm, thanks for the answers

smowton (approved):
Style guide lgtm (haven't reviewed the ql-for-ql bit)

PR #10412, opened 2022-09-14
Revert "Swift: print a helpful message on macOS ARM"

redsun82:
Reverts github/codeql#10399

This breaks bumping codeql on the internal repository. Some internal change is required to bring this back.


criemen (approved):
Can you elaborate why you want to revert that commit?

PR #10414, opened 2022-09-14
Java: Tweak Member.getQualifiedName()

igfoo:
It now includes the qualified name of the declaring type, rather than just the name.

It also has some qldoc suggesting that hasQualifiedName is used instead, for efficiency (getQualifiedName has to construct new strings).


java/ql/lib/change-notes/2022-09-13-Member-getQualifier.md
```diff
@@ -0,0 +1,4 @@
+---
+category: minorAnalysis
```
intrigus-lgtm, line 2: OOI: Isn't this a breaking change?
A lot of queries could stop working, simply because they defined their classes using `Member.getQualifiedName() = foo` in the char. predicate.
As a user, I would consider this a breaking change.

igfoo: Fair point, I'll change it. Thanks!

aschackmull (approved):

Issue #9864, opened 2022-07-20
BUGS report


flowingair:
In some situations,codeql will connect unrelated nodes and believe that there is a flow between them.
for example,when using TaintTracking::Configuration or sinkNode.

Example:

Java file:
```
import java.net.MalformedURLException;
import java.net.URL;
import java.util.HashMap;

public class FakeSink extends HashMap {
    public FakeSink() {
    }

    public Object put(Object name, Object value) {
        try {
            return new URL((String) value);
        } catch (MalformedURLException e) {
            e.printStackTrace();
        } finally {
            return null;
        }
    }
}

import java.net.MalformedURLException;
import java.net.URL;

public class FakeSource {
    public FakeSource()
    {
        String data="";
        Sink sink = new Sink(data);
        try {
            new URL(sink.getData());
        } catch (MalformedURLException e) {
            e.printStackTrace();
        }
    }
}

import java.util.HashMap;
import java.util.Map;

public abstract class Path {
    private Map<String, String> initParams = new HashMap<String, String>();

    public void init(String featureName, Map<String, String> initParam) {
        this.initParams = initParam;
    }

    public Map<String, String> getInitParams() {
        return this.initParams;
    }
}

public class Sink {
    private String data;

    public Sink(String data) {
        this.data = data;
    }

    public String getData() {
        return data;
    }
}

public class Sink2 extends Path {
    public Sink2(String data) {
        getInitParams().put("name", data);
    }
}
```
ql codes
```
class Template extends TaintTracking::Configuration {
  Template() { this = "template" }

  override predicate isSource(DataFlow::Node source) {
    exists(Constructor con |
      con.getNumberOfParameters() = 1 and
      "Ljava/lang/String;" = con.getParameterType(0).getTypeDescriptor() and
      con.getAParameter() = source.asParameter()
    )
  }

  override predicate isSink(DataFlow::Node sink) {
    exists(ConstructorCall cc |
      sink.asExpr() = cc and
      cc.getConstructor().hasQualifiedName("java.net", "URL", "URL")
    )
  }
}
```

result
||||
-- | -- | --
|1 | data : String | Sink.java:4:17
| 2 | data : String | Sink.java:5:21
| 3 | this [post update] [data] : String | Sink.java:5:9
| 4 | new Sink(...) [data] : String | FakeSource.java:8:21
| 5 | sink [data] : String | FakeSource.java:10:21
| 6 | parameter this [data] : String | Sink.java:8:19
[...]

PR #10416, opened 2022-09-14
Java: Remove low confidence dispatch for which we have a manual summary.

aschackmull:
This introduces the concept of low confidence virtual dispatch. Dispatch is currently defined as low confidence if it is purely open-world type-based, and the dispatch type is either `Object`, a functional interface, or an interface from `java.util`. Furthermore, the common marker interfaces `Serializable` and `Clonable` do not count as improved type bounds for the purpose of improving confidence.
Low confidence dispatch is then ignored in data flow in those cases where we also have a manual MaD summary to dispatch to.
As a canonical example consider a call to `Map.put` - without better type bounds this is most likely just some standard map implementation and we will prefer to just use our MaD summary, and not dispatch to any and all custom map implementations that may happen to also exist in the source.

Accounting for the couple of points-to-like enhancements to virtual dispatch in the confidence level is ignored for now.


aschackmull:
dca looks good

atorralba (dismissed):
LGTM! This should help with a ton of spurious flow in certain projects.

aschackmull:
Hmm, this does probably warrant a change note - let me just add one.

aschackmull:
> Nice, this PR should fix parts of #9864 ([#9864 (comment)](https://github.com/github/codeql/issues/9864#issuecomment-1194126784)) I believe 🎉

Indeed!

atorralba (approved):

PR #10417, opened 2022-09-14
remove upper-case variable name from `HubClasses`

ginsbach:
We are phasing our upper-case variable names: https://github.com/github/codeql-core/issues/2157


jketema (approved):
This looks ok to me. What I do not quite know is whether this might have any impact on things we do not test here (I'm not really familiar with the `Architecture` queries beyond knowing hat they exist). I propose we just merge this and see what happens.

PR #10420, opened 2022-09-14
remove several upper-case `NamedExpression` variable names

ginsbach:
The previous problem (https://github.com/github/codeql/pull/10417) was apparently just hiding more.


jketema (approved):
As with the other PR: This looks ok to me. What I do not quite know is whether this might have any impact on things we do not test here (I'm not really familiar with the Architecture/Metrics queries beyond knowing hat they exist). I propose we just merge this and see what happens.

PR #10427, opened 2022-09-14
Kotlin: Compile with -Werror, and fix warnings

igfoo:
We might need to refine this if different versions give different warnings, but for now I've tried the simple approach.


tamasvajk (approved):
Looks reasonable.

PR #10430, opened 2022-09-14
Swift: Fix missing results in swift/cleartext-storage-database

geoffw0:
Fix missing results in `swift/cleartext-storage-database`.  This is something of a change in philosophy, that any write to a class derived from `RealmSwiftObject` (i.e. something that is to be use with a Realm database) is now considered a sink.  We no longer look for evidence (that we may or may not reliably find) that the `RealmSwiftObject` is truly written to the database, because why would it not be?

### DIFF
--- a/swift/ql/src/queries/Security/CWE-311/CleartextStorageDatabase.ql
+++ b/swift/ql/src/queries/Security/CWE-311/CleartextStorageDatabase.ql
@@ -20,7 +20,7 @@ import DataFlow::PathGraph
 /**
  * An `Expr` that is stored in a local database.
  */
-abstract class Stored extends Expr { }
+abstract class Stored extends DataFlow::Node { }
 
 /**
  * An `Expr` that is stored with the Core Data library.
@@ -33,7 +33,7 @@ class CoreDataStore extends Stored {
       c.getAMember() = f and
       f.getName() = ["setValue(_:forKey:)", "setPrimitiveValue(_:forKey:)"] and
       call.getStaticTarget() = f and
-      call.getArgument(0).getExpr() = this
+      call.getArgument(0).getExpr() = this.asExpr()
     )
   }
 }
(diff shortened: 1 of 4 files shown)


redsun82 (approved):
LGTM!

swift/ql/src/queries/Security/CWE-311/CleartextStorageDatabase.ql
```diff
@@ -43,22 +43,14 @@ class CoreDataStore extends Stored {
...
-      c.getAMember() = f and
-      f.getName() = "create(_:value:update:)" and
-      call.getStaticTarget() = f and
-      call.getArgument(1).getExpr() = this
+    // any write into a class derived from `RealmSwiftObject` is a sink. For
+    // example in `realmObj.data = sensitive` the post-update node corresponding
+    // with `realmObj.data` is a sink.
+    exists(ClassDecl cd, Expr e |
+      cd.getABaseTypeDecl*().getName() = "RealmSwiftObject" and
+      this.(DataFlow::PostUpdateNode).getPreUpdateNode().asExpr() = e and
```
MathiasVP, line 51: A small note: You could replace line 44 with:
```ql
class RealmStore extends Stored instanceof DataFlow::PostUpdateNode
```
and replace `this.(DataFlow::PostUpdateNode)` with `super`.

swift/ql/src/queries/Security/CWE-311/CleartextStorageDatabase.ql
```diff
@@ -20,7 +20,7 @@ import DataFlow::PathGraph
 /**
  * An `Expr` that is stored in a local database.
```
MathiasVP, line 21: This comment is now slightly wrong. `Stored` is no longer an `Expr`, but a `DataFlow::Node`.

swift/ql/src/queries/Security/CWE-311/CleartextStorageDatabase.ql
geoffw0: Yep, that's slightly more concise.  Updated.

swift/ql/src/queries/Security/CWE-311/CleartextStorageDatabase.ql
geoffw0: Good point.  Updated.

geoffw0:
Thanks for the reviews, comments addressed.

MathiasVP (approved):
LGTM!

PR #10433, opened 2022-09-15
C#: Fix join order in InterpretedCallable characteristic predicate.

michaelnebel:
The original solution joined on Callable and elementSpec based on name, which lead to a quite large intermediate relation.
The tuples counts in all the screenshots below are based on the *rappayne_webgoat* database.

The tuple counts for the original solution can be seen here:
image

To improve the join ordering and to keep the intermediate relations as small as possible, we use the following observations
- When we want to use *specific* specs (ie. a spec that directly targets a callable), we would like to join Callable and elementSpec on three columns in one go (namespace, type and name).
- When we want to use specs that are valid for all overrides (subtypes = true), we need project the elementSpec relation (on subtypes = true as this is relatively small subset) and join on Callable name. Afterwards we can inspect the types.

The new predicates and tuples counts can be seen here.
image
image
image

### DIFF
--- a/csharp/ql/lib/semmle/code/csharp/dataflow/ExternalFlow.qll
+++ b/csharp/ql/lib/semmle/code/csharp/dataflow/ExternalFlow.qll
@@ -504,15 +504,26 @@ class UnboundCallable extends Callable {
   }
 }
 
+pragma[nomagic]
+private predicate callableSpecInfo(Callable c, string namespace, string type, string name) {
+  c.getDeclaringType().hasQualifiedName(namespace, type) and
+  c.getName() = name
+}
+
+pragma[nomagic]
+private predicate subtypeSpecCandidate(Callable c, UnboundValueOrRefType t) {
+  elementSpec(_, _, true, c.getName(), _, _, t)
+}
+
 private class InterpretedCallable extends Callable {
   InterpretedCallable() {
-    exists(UnboundValueOrRefType t, boolean subtypes, string name |
-      elementSpec(_, _, subtypes, name, _, _, t) and
-      this.hasName(name)
-    |
-      this.getDeclaringType() = t
-      or
-      subtypes = true and
+    exists(string namespace, string type, string name |
+      callableSpecInfo(this, namespace, type, name) and
+      elementSpec(namespace, type, _, name, _, _)
+    )
+    or
+    exists(UnboundValueOrRefType t |
+      subtypeSpecCandidate(this, t) and
       this.getDeclaringType() = t.getASubTypeUnbound+()
     )
   }
--- a/csharp/ql/lib/semmle/code/csharp/dataflow/ExternalFlow.qll
+++ b/csharp/ql/lib/semmle/code/csharp/dataflow/ExternalFlow.qll
@@ -511,8 +511,17 @@ private predicate callableSpecInfo(Callable c, string namespace, string type, st
 }
 
 pragma[nomagic]
-private predicate subtypeSpecCandidate(Callable c, UnboundValueOrRefType t) {
-  elementSpec(_, _, true, c.getName(), _, _, t)
+private predicate subtypeSpecCandidate(string name, UnboundValueOrRefType t) {
+  exists(UnboundValueOrRefType t0 |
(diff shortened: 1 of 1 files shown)


michaelnebel:
It looks like DCA is somewhat broken ATM when it comes to reporting some of the data for join order badness (we can't see the detail of a specific join order problem, only that there might be one).
In any case, the following DCA executions indicates that the changes on this PR solves the join order problem for InterpretedCallable:

Running DCA against main has 281 issues reported in the join order badness section (33 related to InterpretedCallable):
https://github.com/github/codeql-dca-main/blob/data/michaelnebel/main__nightly__nightly/reports/summaries/any.md#join-order-badness-per-source-and-predicate-v0


Running DCA against the branch on this PR has 248 issues reported in the join order badness secion (0 related to InterpretedCallable)
https://github.com/github/codeql-dca-main/blob/data/michaelnebel/fix-joinorder-i__nightly__nightly/reports/summaries/any.md#join-order-badness-per-source-and-predicate-v0

Also note that math says: 281 - 33 = 248

csharp/ql/lib/semmle/code/csharp/dataflow/ExternalFlow.qll
```diff
@@ -504,15 +504,26 @@ class UnboundCallable extends Callable {
...
 
+pragma[nomagic]
+private predicate callableSpecInfo(Callable c, string namespace, string type, string name) {
+  c.getDeclaringType().hasQualifiedName(namespace, type) and
+  c.getName() = name
+}
+
+pragma[nomagic]
+private predicate subtypeSpecCandidate(Callable c, UnboundValueOrRefType t) {
+  elementSpec(_, _, true, c.getName(), _, _, t)
```
hvitved, line 515: This looks like a potentially bad join, as we will relate *all* calls with matching names. How about first joining on `getASubTypeUnbound+()` instead?

```ql
pragma[nomagic]
private predicate subtypeSpecCandidate(string name, UnboundValueOrRefType t) {
  exists(UnboundValueOrRefType t0 |
    elementSpec(_, _, true, name, _, _, t0) and
    t = t0.getASubTypeUnbound+()
  )
}

pragma[nomagic]
private predicate callableInfo(Callable c, string name, UnboundValueOrRefType decl) {
  name = c.getName() and
  decl = c.getDeclaringType()
}
```

and then use it in `InterpretedCallable` as

```ql
exists(string name, UnboundValueOrRefType t |
  callableInfo(this, name, t) and
  subtypeSpecCandidate(name, t)
)
```

hvitved (commented):
First join-fix LGTM, I think the second can be improved.

csharp/ql/lib/semmle/code/csharp/dataflow/ExternalFlow.qll
michaelnebel: Yes, I think you are right! However, in this particular case the join is worsened slightly.
Let us try with your solution and run DCA.

image
image
image

michaelnebel:
Weird actions error on last DCA execution. Will try again.

csharp/ql/lib/semmle/code/csharp/dataflow/ExternalFlow.qll
hvitved: > However, in this particular case the join is worsened slightly.

I think that is only because we are currently lucky that `elementSpec(_, _, true, name, _, _, t)` is rather small. But even so, if, say, `elementSpec(_, _, "ToString", _, _, t)` held, we would join with *all* `ToString` methods, which can be a large set.

hvitved (approved):

csharp/ql/lib/semmle/code/csharp/dataflow/ExternalFlow.qll
michaelnebel: Yes - I totally agree.

michaelnebel:
DCA looks good
(1) InterpretedCallable has disappeared from the join order badness list.
(2) None of the newly introduced predicates have turned up on the join order badness list

PR #10435, opened 2022-09-15
C++: remove more upper-case variable names

ginsbach:
This builds on https://github.com/github/codeql/pull/10420, apologies for not spotting all of these at once!


jketema (approved):
As with the other PRs: This looks ok to me. What I do not quite know is whether this might have any impact on things we do not test here (as these are queries of `@kind table`). I propose we just merge this and see what happens.

PR #10436, opened 2022-09-15
C#: Dynamically create type based summaries

michaelnebel:
This is an experimental branch, where we dynamically create type based summaries for all library code not already covered by a summary.


michaelnebel:
@hvitved 
Last DCA run shows that it seems feasible to do dynamic type based summary generation:
(outliers needs to be investigated, but overall there doesn't appear to be a significant slowdown)
image

PR #10438, opened 2022-09-15
C++: Add a `cpp/invalid-pointer-deref` query to experimental

MathiasVP:
This adds a new experimental query to detect issues like:
```cpp
char* p = malloc(size);
char* q = p + size;
use(*q);
```

the query correctly catches the 4 issues in https://github.com/advisories/GHSA-4q9q-728x-j7v5 (i.e., it finds 2 out-of-bounds reads and 2 out-of-bounds writes in the vulnerable database, and 0 in the fixed database).

- d981f898e403660259b1660308c9c4b3e178f380 adds flow states to the new product dataflow library. We need this to handle cases like:
```cpp
char* p = malloc(size + 1);
...
char* q = p + (size + 1);
use(*q);
```
where we use the flow state to remember that the size of the allocation is the size we're tracking + 1.
- b8a5aa5d85b0abd1c2177650d78d75fa54ed8a80 fixes a couple of range-analysis issues where the bounds we were inferring for pointers were suboptimal compared to the bounds for integral expressions. This also fixes some missing results in some of the other experimental queries.
- Finally, 4482669d7e61bf5ac6306792e9fbc4157cbd390c adds the new query.


cpp/ql/src/experimental/Security/CWE/CWE-193/InvalidPointerDeref.ql
```diff
@@ -0,0 +1,316 @@
+/**
+ * @name Invalid pointer dereference
+ * @description Dereferencing a pointer that points past it allocation is undefined behavior
+ *              and may lead to security vulnerabilities.
+ * @kind path-problem
+ * @problem.severity error
+ * @precision high
```
geoffw0, line 7: Have you done a MRVA or LGTM run to justify this?

cpp/ql/test/experimental/query-tests/Security/CWE/CWE-193/pointer-deref/test.cpp
```diff
@@ -0,0 +1,181 @@
...
+void test11(int size) {
+    char* end;
+    char* begin = mk_array(size, &end);
+
+    for (char* p = begin; p != end; ++p) {
+        *p = 0; // GOOD
+    }
+
+    for (char* p = begin; p <= end; ++p) {
+        *p = 0; // BAD
```
geoffw0, line 67: I'm impressed that we catch these!

geoffw0 (commented):
`qhelp` and tests look good.  I'll review the query itself later (and perhaps the library changes).

cpp/ql/src/experimental/Security/CWE/CWE-193/InvalidPointerDeref.ql
MathiasVP: I have not. As it's in experimental this doesn't really mean much, though. If you'd prefer I can remove the precision tag. I have not seen any FPs on the query yet, however.

Running this on MRVA is currently blocked because the current image used for MRVA has a compiler bug that the new C++ dataflow library (which the new product dataflow library is using) is hitting.

geoffw0: Good point about experimental, and good to hear you haven't seen any FPs.  We should (briefly) review this tag when it comes out of experimental to be sure.

cpp/ql/src/experimental/Security/CWE/CWE-193/InvalidPointerDeref.ql
```diff
@@ -0,0 +1,316 @@
...
+ * non-constant operand of the pointer-arithmetic operation.
+ */
+pragma[nomagic]
+predicate pointerAddInstructionHasBounds(
+  PointerAddInstruction pai, DataFlow::Node sink1, Instruction sink2, int delta
+) {
+  exists(Instruction right |
+    pointerAddInstructionHasOperands(pai, sink1.asInstruction(), right) and
+    bounded(right, sink2, delta)
+  )
```
geoffw0, line 128: Does all this work if the sink is `size + base` as well as if it is `base + size`?

cpp/ql/src/experimental/Security/CWE/CWE-193/InvalidPointerDeref.ql
```diff
@@ -0,0 +1,316 @@
...
+predicate pointerAddInstructionHasOperands(
+  PointerAddInstruction pai, Instruction left, Instruction right
+) {
+  pai.getLeft() = left and
+  pai.getRight() = right
+}
+
+/**
+ * Holds if `pai` is non-strictly upper bounded by `b + delta` and `sink1` is the
+ * non-constant operand of the pointer-arithmetic operation.
```
geoffw0, line 119: I'm confused by this.  I'm not sure what `b` is in this context, and I'm not convinced "non-constant" is what you wanted to say.  My best guess is:
```suggestion
 * Holds if `pai` is an addition of the pointer `sink1` with a value non-strictly upper bounded by `sink2 + delta`.
```

cpp/ql/src/experimental/Security/CWE/CWE-193/InvalidPointerDeref.ql
```diff
@@ -0,0 +1,316 @@
...
+) {
+  exists(Instruction right |
+    pointerAddInstructionHasOperands(pai, sink1.asInstruction(), right) and
+    bounded(right, sink2, delta)
+  )
+}
+
+/**
+ * Holds if `pai` is non-strictly upper bounded by `sink2 + delta` and `sink1` is the
+ * non-constant operand of the pointer-arithmetic operation.
```
geoffw0, line 133: Also somewhat confusing.

cpp/ql/src/experimental/Security/CWE/CWE-193/InvalidPointerDeref.ql
```diff
@@ -0,0 +1,316 @@
...
+predicate invalidPointerToDerefSource(
+  PointerArithmeticInstruction pai, DataFlow::Node source, int delta
+) {
+  exists(ProductFlow::Configuration conf, DataFlow::PathNode p, DataFlow::Node sink1 |
+    p.getNode() = sink1 and
+    conf.hasFlowPath(_, _, p, _) and
+    isSinkImpl(pai, sink1, _, _) and
+    bounded(source.asInstruction(), pai, delta) and
+    delta >= 0
+  )
```
geoffw0, line 181: I feel like either something non-trivial is going on here that should be explained, or the code should be simpler.  Is this to do with further arithmetic that happens after the initial 'end pointer' is calculated.

cpp/ql/src/experimental/Security/CWE/CWE-193/InvalidPointerDeref.ql
```diff
@@ -0,0 +1,316 @@
...
+  TPathNode3(DataFlow3::PathNode p) or
+  // The read/write that uses the invalid pointer identified by `InvalidPointerToDerefConf`.
+  // This one is needed because the sink identified by `InvalidPointerToDerefConf` is the
+  // pointer, but we want to raise an alert at the dereference.
+  TPathNodeSink(Instruction i) {
+    exists(DataFlow::Node n |
+      any(InvalidPointerToDerefConf conf).hasFlow(_, n) and
+      isInvalidPointerDerefSink(n, i, _)
+    )
+  }
```
geoffw0, line 197: So the overall structure is something like:
```
1. the allocation, e.g. p = malloc(size)
 v
2. 'end pointer' is calculated, e.g. q = p + size
(critically this is not just any pointer addition, its adding the allocation size to the allocation base)
 v
3. end pointer is accessed, e.g. *(q + 1)
```
I think this query would be a lot easier to read and understand if the above (or a corrected version of it) was established near the top, along with terminology (e.g. "allocation base", "allocation size" and "end pointer" I've used above) that can replace much of the anonymous stuff like `sink2` in the predicates that follow.

cpp/ql/src/experimental/Security/CWE/CWE-193/InvalidPointerDeref.ql
rdmarsh2: Yes - the IR translation guarantees that `PointerAddInstruction::getLeft` returns the pointer operand, even if the source is written as `base + size`

cpp/ql/src/experimental/Security/CWE/CWE-193/InvalidPointerDeref.ql
MathiasVP: I've added such a comment in 60b9a373ece9d1ec451c05422fc48f127795aa53. Let me know if you think it's more clear. I've kept the names `sink1` and `sink2` since I find it useful to know that a parameter will be used as the (first or second) sink of the product flow configuration. I've added some examples to the predicates that should hopefully explain the role of these somewhat anonymous parameters better.

cpp/ql/src/experimental/Security/CWE/CWE-193/InvalidPointerDeref.ql
MathiasVP: There is indeed something nontrivial going on here, yes. Thanks for pointing this out. I've added some QLDoc that should hopefully clear this up in 60b9a373ece9d1ec451c05422fc48f127795aa53.

cpp/ql/src/experimental/Security/CWE/CWE-193/InvalidPointerDeref.ql
MathiasVP: (Hopefully) fixed in 031f20a0ebfbcfbd6ed67334ebcd679fe5330bdf.

cpp/ql/src/experimental/Security/CWE/CWE-193/InvalidPointerDeref.ql
MathiasVP: Yeah, sorry. `b` made total sense in an earlier version of the query where I was explicitly carrying around a `SemBound` column. I had to tweak this for performance _after_ writing the QLDoc, and forgot to update it. The non-constant stuff also disappeared after writing the QLDoc >.<

I've cleared this up in 031f20a0ebfbcfbd6ed67334ebcd679fe5330bdf, I hope.

MathiasVP:
@geoffw0 I believe I've addressed all of your comments now.

cpp/ql/src/experimental/Security/CWE/CWE-193/InvalidPointerDeref.ql
rdmarsh2: That's missing a key point that the step from 2 to 3 is a relative bound, not (necessarily) a direct copy. That means it can handle things like the following:
```
int *arr = malloc(size);
int *end = arr + size;
for(int *p = arr; p <= end; p++) {
  use(*p);
}
```

cpp/ql/lib/experimental/semmle/code/cpp/dataflow/ProductFlow.qll
```diff
@@ -11,14 +11,31 @@ module ProductFlow {
      *
      * `source1` and `source2` must belong to the same callable.
      */
-    abstract predicate isSourcePair(DataFlow::Node source1, DataFlow::Node source2);
+    predicate isSourcePair(DataFlow::Node source1, DataFlow::Node source2) { none() }
+
+    predicate isSourcePair(
```
rdmarsh2, line 16: Can you add QLdoc for this (and `isSinkPair/4`)?

cpp/ql/src/experimental/Security/CWE/CWE-193/InvalidPointerDeref.ql
```diff
@@ -0,0 +1,347 @@
...
+  exists(ProductFlow::Configuration conf, DataFlow::PathNode p, DataFlow::Node sink1 |
+    p.getNode() = sink1 and
+    conf.hasFlowPath(_, _, p, _) and
+    isSinkImpl(pai, sink1, _, _) and
+    bounded(source.asInstruction(), pai, delta) and
+    delta >= 0
+  )
+}
+
+newtype TMergedPathNode =
```
rdmarsh2, line 215: Not work for this PR, and may need arbitrary flow state types, but I wonder if this should be done with a modified product flow library that allows one underlying configuration to flow farther than the other, using flow states to mark the expected sink.

rdmarsh2:
LGTM once the last couple doc points are addressed

cpp/ql/src/experimental/Security/CWE/CWE-193/InvalidPointerDeref.ql
MathiasVP: Good point. I've gone with your example (and expanded the explanation) in dc00643ad100435ca6520397f4af3342da714074.

cpp/ql/lib/experimental/semmle/code/cpp/dataflow/ProductFlow.qll
MathiasVP: Done in dc00643ad100435ca6520397f4af3342da714074 👍.

cpp/ql/src/experimental/Security/CWE/CWE-193/InvalidPointerDeref.ql
MathiasVP: Yeah, I think that would be doable.

MathiasVP:
> LGTM once the last couple doc points are addressed

Should all be fixed now. Thanks!

rdmarsh2 (approved):

cpp/ql/src/experimental/Security/CWE/CWE-193/InvalidPointerDeref.ql
```diff
@@ -0,0 +1,347 @@
...
+ */
+predicate invalidPointerToDerefSource(
+  PointerArithmeticInstruction pai, DataFlow::Node source, int delta
+) {
+  exists(ProductFlow::Configuration conf, DataFlow::PathNode p, DataFlow::Node sink1 |
+    p.getNode() = sink1 and
+    conf.hasFlowPath(_, _, p, _) and
+    isSinkImpl(pai, sink1, _, _) and
+    bounded(source.asInstruction(), pai, delta) and
+    delta >= 0
```
geoffw0, line 216: Something feels off here.  I may be wrong, there are a lot of subtleties going on.

1. we never use the state / delta value from the first dataflow config anywhere.  Without this I don't see how we can distinguish cases like:

```
char *buffer1 = malloc(size);
char *ptr1 = buffer1 + size;
*ptr1; // BAD

char *buffer2 = malloc(size + 1);
char *ptr2 = buffer2 + size;
*ptr2; // GOOD
```

2. also I think it might be OK to track a negative delta as long as the sum is positive by the end (dereference), e.g. to find this result:
```
char *p = malloc(size);
char *q = p + size - 1; // seems OK here
char a = *(q + 1); // BAD
```
(I'm not entirely sure we don't get this right, with `(q + 1)` as the midpoint between the two dataflows in this case rather than `p + size - 1`???)

cpp/ql/src/experimental/Security/CWE/CWE-193/InvalidPointerDeref.ql
```diff
@@ -0,0 +1,347 @@
...
+  InvalidPointerToDerefConf() { this = "InvalidPointerToDerefConf" }
+
+  override predicate isSource(DataFlow::Node source) { invalidPointerToDerefSource(_, source, _) }
+
+  override predicate isSink(DataFlow::Node sink) { isInvalidPointerDerefSink(sink, _, _) }
+}
+
+/**
+ * Holds if `pai` is a pointer-arithmetic operation and `source` is a dataflow node with a
+ * pointer-value that is non-strictly upper bounded by `pai + delta`.
```
geoffw0, line 202: I don't think this fully describes what's supposed to be going on in the predicate.  This description seems like it would be pretty much satisfied by just:
```
    bounded(source.asInstruction(), pai, delta)
```

geoffw0:
Thanks for the changes to comments by the way - already much easier to follow what's going on!

cpp/ql/src/experimental/Security/CWE/CWE-193/InvalidPointerDeref.ql
MathiasVP: Yeah, you're right. I didn't specify `delta >= 0` in my comment.

Other than that, the implementation you've inferred from the comment is correct. The difference between your implementation and the one I've written is that this predicate is only computed for those pointer-arithmetic instructions that are reached from the `ProductFlow` stage.

cpp/ql/src/experimental/Security/CWE/CWE-193/InvalidPointerDeref.ql
MathiasVP: > we never use the state / delta value from the first dataflow config anywhere. Without this I don't see how we can distinguish cases like:
> ```cpp
> char *buffer1 = malloc(size);
> char *ptr1 = buffer1 + size;
> *ptr1; // BAD
>
> char *buffer2 = malloc(size + 1);
> char *ptr2 = buffer2 + size;
> *ptr2; // GOOD
> ```
We do actually use the state to precisely distinguish these cases. Take a look at the `isSinkPair` (and `isSourcePair`) definitions in the `AllocToInvalidPointerConf` configuration. This is the source:
```ql
state1 = "" and
hasSize(source1.asConvertedExpr(), source2, state2)
```
and this is the sink:
```ql
state1 = "" and
exists(int delta |
  isSinkImpl(_, sink1, sink2, delta) and
  state2 = delta.toString()
)
```
So in `isSourcePair`, we're using the state to capture the fact that the size is `source2` plus a constant offset stored in `stage2`, and we're using that state in `isSinkPair` to check that the end pointer being computed has the same offset. So in this example:
```cpp
char *buffer2 = malloc(size + 1);
char *ptr2 = buffer2 + size;
*ptr2; // GOOD
```
the state of the second projection of the product dataflow configuration (i.e., `state2`) is `1` because the allocation has size `size + 1`, and then `buffer2 + size` will only be a sink when `state2` is `0`.

Does that make sense?

MathiasVP: >2. also I think it might be OK to track a negative delta as long as the sum is positive by the end (dereference), e.g. to find this result:
> ```cpp
> char *p = malloc(size);
> char *q = p + size - 1; // seems OK here
> char a = *(q + 1); // BAD
> ```
>
> (I'm not entirely sure we don't get this right, with (q + 1) as the midpoint between the two dataflows in this case rather than p + size - 1???)

Indeed, we actually correctly catch this result with `(q + 1)` as the midpoint (because we see that `q + 1` is a non-strict tupper bound for `p + size`).

We won't, however, catch an example like:
```cpp
void deref_plus_one(char* q) {
    char a = *(q + 1);
}

void test(unsigned size) {
    char *p = malloc(size);
    char *q = p + size - 1;
    deref_plus_one(q);
}
```
because the range-analysis library is intraprocedural, so we find `p + size` as the only expression that upper bounds `p + size`.

MathiasVP: 3e6576bfaf131478b4a942f27e6e36acbc82a4e1 adds the above testcase.

MathiasVP: and 02076074ffa1fd8286c3db0fd5d9b8a9e718bf0a adds more comments to make the use of flow states more clear.

geoffw0: 1. Ah yes, I can see that the delta is connected between the source and the sink in this configuration.

2. Thanks for clearing this up.

geoffw0 (approved):
I think this is ready to merge.  It should be a very valuable query!

PR #10439, opened 2022-09-15
JavaScript: remove upper-case variable names

ginsbach:
Upper-case variable names are deprecated. Warnings were not correctly displayed for `NamedExpressions`, hence we missed some occurrences.


erik-krogh (approved):
👍  

I made a QL-for-QL query that was supposed to catch this, but I see that I missed support for `AsExpr` in that query.  

I've fixed the QL-for-QL query here: https://github.com/github/codeql/pull/10443

PR #10447, opened 2022-09-16
Swift: open(2) interception

AlexDenisov:


redsun82:
We probably need a quick description for the PR, but if you prefer as I'm writing an internal document for this, I can later copy & paste the relevant part in the PR description (even after we merge it)

swift/extractor/remapping/SwiftOpenInterception.macOS.cpp
```diff
@@ -0,0 +1,79 @@
...
+#include <fishhook.h>
+#include <llvm/Support/raw_ostream.h>
+#include <llvm/Support/FileSystem.h>
+#include <llvm/Support/Path.h>
+#include <fcntl.h>
+#include <unistd.h>
+
+namespace codeql {
+
+static std::string scratchDir;
```
redsun82, line 11: I _think_ there might be a use-after-destructor problem with this, in case `open` is called after `exit` by some other destructor of an object with static storage duration. Even though this is not happening now, we cannot really guarantee this won't happen. For example, what if the swift library initializes some non-POD static global that writes down diagnostics at the end of the program?

We could either:
* undo the interception more or less at the time of calling `remapArtifacts`, putting `original_open` back to `open`
* use a `static const char*` for the `scratchDir`, making `codeql_open` auto fallback on `original_open` if `scratchDir == nullptr`, and set it back to `nullptr` in `remapArtifacts`. The `SwiftExtractorConfiguration` object owning the scratch dir string will be alive between calls to `initInterception` and `remapArtifacts`

I'm always wary of using non-const globals (which is probably unavoidable here because of having to interface with C via free functions), but the wariness doubles when using non-POD globals (const or non const) because of the subtle bugs that can happen during end of program destructor calls. So if there is a way to avoid that I would rather take it, and go for the second option above.

swift/extractor/remapping/SwiftOpenInterception.macOS.cpp
```diff
@@ -0,0 +1,79 @@
...
+      if (!hash.empty() && llvm::sys::fs::exists(hashed)) {
+        newPath = hashed;
+      }
+    }
+  }
+
+  return original_open(newPath.c_str(), oflag, mode);
+}
+
+void remapArtifacts(const std::unordered_map<std::string, std::string>& mapping) {
```
redsun82, line 54: maybe rename to `finalizeRemapping`? If there is some cleanup code going on (see my other comment) this name would cover that nicely

swift/extractor/remapping/SwiftOpenInterception.macOS.cpp
```diff
@@ -0,0 +1,79 @@
...
+    if (!hash.empty() && llvm::sys::fs::exists(patched)) {
+      if (std::error_code ec = llvm::sys::fs::create_link(/* from */ patched, /* to */ hashed)) {
+        llvm::errs() << "Cannot remap file '" << patched << "' -> '" << hashed
+                     << "': " << ec.message() << "\n";
+      }
+    }
+  }
+}
+
+void initInterception(const std::string& dir) {
```
redsun82, line 71: `initRemapping` maybe? Especially if renaming `remapArtifacts` to `finalizeRemapping`?

swift/extractor/remapping/SwiftOpenInterception.h
```diff
@@ -0,0 +1,11 @@
+#pragma once
+
+#include <string>
+#include <unordered_map>
+
+namespace codeql {
+
+void initInterception(const std::string& dir);
+void remapArtifacts(const std::unordered_map<std::string, std::string>& mapping);
```
redsun82, line 9: maybe we should consider an alternative API using RAII, something like
```cpp
class Interceptor {
  const std::unordered_map<std::string, std::string>& mapping;

 public:
  // does initInterception, stores reference to mapping
  Interceptor(const std::string& dir, const std::unordered_map<std::string, std::string>& mapping); 

  // does remapArtifacts using stored mapping, does cleanup if needed
  ~Interceptor();
};
```
The mapping does not even need to be filled at construction time.

redsun82 (changes requested):
Nice job! I think we should at least solve the possible use-after-destructor problem, the other comments are not really required to merge (especially if we later do a cross-platform thing without fishhook)

swift/extractor/remapping/SwiftOpenInterception.macOS.cpp
AlexDenisov: It's a very good point and it would've been a nightmare to catch/debug afterwards!

However, `SwiftExtractorConfiguration` doesn't really own the scratch dir as we changed `tempArtifactDir` to be a function returning a new string every time it's called 😅. We can easily workaround this limitation, but IMO having a  `char *` is way too fragile.

I'll see what's the best way to rebind the functions back.

redsun82: Ah, good point on the const char lifetime!

redsun82: If you go for a RAII class you could use the lifetime of that class to make sure the raw pointer is alive while required.

swift/extractor/remapping/SwiftOpenInterception.macOS.cpp
AlexDenisov: Good idea, renamed

misc/bazel/workspace.bzl
```diff
@@ -55,3 +56,11 @@ def codeql_workspace(repository_name = "codeql"):
             "https://github.com/bazelbuild/rules_python/archive/refs/tags/0.8.1.tar.gz",
         ],
     )
+
+    new_git_repository(
```
criemen, line 60: Using `http_archive` with GitHubs codeload service, i.e. 
https://github.com/facebook/fishhook/archive/aadc161ac3b80db07a9908851839a17ba63a9eb1.zip 
tends to be much faster than using git to pull in the code.

swift/extractor/remapping/SwiftOpenInterception.macOS.cpp
AlexDenisov: renamed

swift/extractor/remapping/SwiftOpenInterception.macOS.cpp
AlexDenisov: I played a bit with the options and the best suggestion I have is to add a boolean flag, though as I'm writing I'm wondering if it should be an atomic boolean? 🤔

misc/bazel/workspace.bzl
criemen: You probably need to set the `strip_prefix` option to (presumably) `strip_prefix = "fishhook-aadc161ac3b80db07a9908851839a17ba63a9eb1"` - see the `WORKSPACE.bazel` file in semmle-code for examples where we already do this.

AlexDenisov: It did the trick 🎉

swift/extractor/remapping/SwiftOpenInterception.macOS.cpp
redsun82: Good question, does llvm or the swift frontend fire up threads? Even if they do, are they still not joined when we execute our code?

Another option could be maybe to make `scratchDir` indestructible. Declare it as
```
static const std::string* scratchDir = nullptr;
```
and initialize it with
```
scratchDir = new std::string(dir);
```
and just fuggedaboutit... but I'm ok with this solution. We will probably need to come back to this code if we generalize the interception to Linux.

redsun82 (approved):

criemen (approved):
Bazel changes LGTM

swift/extractor/remapping/SwiftOpenInterception.macOS.cpp
AlexDenisov: Yeah, all the available options are more like "which footgun shall we pick" 😄

PR #10458, opened 2022-09-16
Update qlpack properties descriptions

aeisenberg:
Makes table easier to read. As discussed with @felicitymay.

Note that there are still a few internal properties that are used in our core packs, but I am not mentioning (eg- `groups`, `mlModels`). These are internal use only and I didn't want to clutter the descriptions with these extra properties.

### DIFF
--- a/docs/codeql/codeql-cli/about-codeql-packs.rst
+++ b/docs/codeql/codeql-cli/about-codeql-packs.rst
@@ -58,118 +58,161 @@ The following properties are supported in ``qlpack.yml`` files.
    :widths: auto
 
    * - Property
-     - Example
-     - Required by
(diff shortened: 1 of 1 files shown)


felicitymay:
Thanks @aeisenberg - much appreciated.

I'll try to check this tomorrow, but am stretched rather thin right now 😞

aeisenberg:
No worries. It's not high priority. I just didn't want this to get dropped.

felicitymay (approved):
Thank you so much for taking the time to reorganize the information in this table. The result is so much easier to read, especially on narrower screens/browser windows. 💖

aeisenberg:
Thanks for the review!

PR #10460, opened 2022-09-16
Updates the library path section of the CodeQL spec

aeisenberg:
- Remove references to `queries.xml`. It is still supported, but we don't want people using it.
- Add reference to `codeql-pack.yml`. It is just an alias for `qlpack.yml` and not being used.
- Remove reference to `libraryPathDependencies` and use `dependencies` instead.

Note that this section does not give a complete description of library paths. That will be a part of the "Developing a codeql pack" article that is forthcoming.


dbartol (approved):
LGTM

PR #10465, opened 2022-09-17
C#: Integration test(s)

michaelnebel:
PR includes integration tests for
* `dotnet publish`
* `dotnet pack`
* `dotnet test` (posix only - there seems to be an issue with nuget restore on the windows action runners).


hvitved (dismissed):
Thanks!

tamasvajk (commented):
Is `dotnet restore` required? According to https://learn.microsoft.com/en-us/dotnet/core/tools/dotnet-test#implicit-restore, it isn't.

michaelnebel:
> Is `dotnet restore` required? According to https://learn.microsoft.com/en-us/dotnet/core/tools/dotnet-test#implicit-restore, it isn't.

No - probably not - we are just reaching for straws to get the `dotnet test` to pass.
I will convert the PR back to draft.

hvitved (approved):

PR #10469, opened 2022-09-19
C#: Prepend `-p:UseSharedCompilation=false` instead of append for `dotnet run`

hvitved:
https://github.com/github/codeql/pull/10464 follow-up.

### DIFF
--- a/csharp/ql/integration-tests/all-platforms/dotnet_run/test.py
+++ b/csharp/ql/integration-tests/all-platforms/dotnet_run/test.py
@@ -1,0 +1,17 @@
+from create_database_utils import *
+
+# no arguments
+run_codeql_database_create(['dotnet run'], test_db="test-db", db=None, lang="csharp")
+
+# no arguments, but `--`
+run_codeql_database_create(['dotnet clean', 'rm -rf test-db obj bin', 'dotnet run --'], test_db="test2-db", db=None, lang="csharp")
+
+# two arguments, no `--`
+run_codeql_database_create(['dotnet clean', 'rm -rf test2-db obj bin', 'dotnet run hello world'], test_db="test3-db", db=None, lang="csharp")
+
+# two arguments, and `--`
+run_codeql_database_create(['dotnet clean', 'rm -rf test3-db obj bin', 'dotnet run -- hello world'], test_db="test4-db", db=None, lang="csharp")
+
+# shared compilation enabled; tracer should override by changing the command
+# to `dotnet run -p:UseSharedCompilation=true -p:UseSharedCompilation=false -- hello world`
+run_codeql_database_create(['dotnet clean', 'rm -rf test4-db obj bin', 'dotnet run -p:UseSharedCompilation=true -- hello world'], test_db="test5-db", db=None, lang="csharp")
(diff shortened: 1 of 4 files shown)


csharp/ql/integration-tests/all-platforms/dotnet_run/test.py
```diff
@@ -0,0 +1,17 @@
+from create_database_utils import *
+
+# no arguments
+run_codeql_database_create(['dotnet run'], test_db="test-db", db=None, lang="csharp")
+
+# no arguments, but `--`
+run_codeql_database_create(['dotnet clean', 'rm -rf test-db obj bin', 'dotnet run --'], test_db="test2-db", db=None, lang="csharp")
```
michaelnebel, line 7: It is possible to delete at least the 'obj' and 'bin' folder as a part of the clean target (if the following is added to the project folder).
```csharp
<Target Name="DeleteBinObjFolders" BeforeTargets="Clean">
   <RemoveDir Directories=".\bin" />
   <RemoveDir Directories=".\obj" />
 </Target>
```

Was it the `test-db` folder that also causes failures when running multiple tests (I suppose it is because the source code is included in the testdb folder)?

csharp/ql/integration-tests/all-platforms/dotnet_run/test.py
```diff
@@ -0,0 +1,17 @@
+from create_database_utils import *
```
michaelnebel, line 1: Would it make sense to also expect the output of running the program (alternatively, I can do that as a follow up - which I dont' mind doing now that you have done all the heavy lifting 😂 ).
One thing is that the database is created, but unless we inspect the program output we haven't verified that the arguments has been re-written as we expect.

Something along the lines of:
```python
import os
from create_database_utils import *

def run_codeql_database_create_stdout(args, dbname):
    stdout = open(dbname + "file.txt", 'w+')
    run_codeql_database_create(args, test_db=dbname, db=None, stdout=stdout, lang="csharp")
    stdout.seek(0)
    s = stdout.read()
    stdout.close()
    return s

# Compile and run the program.
# It is expected that 'hello1' is printed to stdout.
s = run_codeql_database_create_stdout(['dotnet run -- hello1 hello2'], 'without-argument-db')

// validate s
```

csharp/tools/tracing-config.lua
```diff
@@ -38,13 +39,60 @@ function RegisterExtractorPack(id)
...
+            for i, injectionArg in ipairs(injections) do
+                table.insert(argv, injectionIndex + i - 1, injectionArg)
+            end
+
+            if OperatingSystem == 'windows' then
+                return {
+                    order = ORDER_REPLACE,
+                    invocation = {
+                        path = AbsolutifyExtractorPath(id, compilerPath),
+                        arguments = {
```
michaelnebel, line 80: BuildExtractorInvocation constructs a record line type with the labels `path` and `transformedArguments`. Does that cause problems elsewhere?

csharp/ql/integration-tests/all-platforms/dotnet_run/test.py
hvitved: Yes. Can we fix that follow-up?

csharp/tools/tracing-config.lua
hvitved: No, it is only when need to pass in custom arguments on Windows, where we need to use the `commandLineString` field.

michaelnebel (approved):
Great work! 👑 of Lua.

PR #10470, opened 2022-09-19
JS: Try to parse files without using our parser extensions before enabling the extensions

erik-krogh:
All tests just pass, except the trap-test for [odasa-2593.js](https://github.com/erik-krogh/ql/blob/ccae0933c7b6047cbb22c69a259cd3b0e129ce18/javascript/extractor/tests/mozilla/input/odasa-2593.js).   
I think it found another valid way to parse that same file.  

[Evaluation looks surprisingly good](https://github.com/github/codeql-dca-main/tree/data/erik-krogh/pr-10470-ccae09__default__CustomSuite__1/reports).  
Fewer extraction errors, more LOC, more successfully extracted files, less syntax errors. 

[Another evaluation on default.yml](https://github.com/github/codeql-dca-main/tree/data/erik-krogh/pr-10470-ccae09__default__code-scanning/reports) also looks good.  

Does this need a change-note?


asgerf (commented):
Looks great! I think a change note in `src` would be in order. I suggest the following:

> Improved how the JavaScript parser handles ambiguities between plain JavaScript and dialects such as Flow and E4X that use the same file extension. The parser now prefers plain JavaScript if possible, falling back to dialects only if the source code can not be parsed as plain JavaScript. Previously, there were rare cases where parsing would fail because it would erroneously attempt to parse dialect-specific syntax in a regular JavaScript file.

asgerf (approved):

PR #10480, opened 2022-09-19
Update supported language codes

rvermeulen:
Add support for the languages Ruby and Swift based on the language codes used in the queries.

### DIFF
--- a/docs/query-metadata-style-guide.md
+++ b/docs/query-metadata-style-guide.md
@@ -76,6 +76,8 @@ You must specify an `@id` property for your query. It must be unique and should 
 * Java: `java`
 * JavaScript and TypeScript: `js`
 * Python: `py`
+* Ruby: `rb`
+* Swift: `swift`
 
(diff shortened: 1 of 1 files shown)


sj (changes requested):
I know this is documentation for query developers, and it's good to include `swift` and `rb` in that list from that point of view. I'm a little worried that the current wording *could* confuse customers or sales reps who accidentally end up in the wrong place looking for CodeQL language support.

Could we perhaps update the wording of the paragraph above the list? For example, replace this:
```
[...] forward slash. The following language codes are supported:
 - C and C++: cpp
 - ...
```

With:
```
[...] forward slash. The following language codes are currently in use for CodeQL query development:
 - C and C++: cpp
 - ...
 - Ruby: rb
 - Swift: swift

For a full list of currently supported CodeQL languages (and the level of coverage), 
please see [the CodeQL documentation](https://codeql.github.com/docs/).

sj:
Side-note: should we include Kotlin in that list, e.g. as:
` - Java and Kotlin: java`?

PR #10481, opened 2022-09-19
Update bazel to v5.3.1

chenrui333:


redsun82 (commented):
Hi @chenrui333 For starters, thank you for reminding us to use a more recent Bazel version :slightly_smiling_face: Your PR also made us spot a problem with our GitHub actions workflow configuration, that was not correctly triggering runs on such a change, so I'll just double the thanks :+1: . The latter problem was fixed by https://github.com/github/codeql/pull/10482.

The artifacts built by Bazel here are part of [github/codeql-cli-binaries](https://github.com/github/codeql-cli-binaries), which in turn is built in an internal private repository. In light of that, we would prefer to keep the Bazel version in sync between the two. We need to look whether we are ready to bump up Bazel's version in the internal repository first.

chenrui333:
> Hi @chenrui333 For starters, thank you for reminding us to use a more recent Bazel version 🙂 Your PR also made us spot a problem with our GitHub actions workflow configuration, that was not correctly triggering runs on such a change, so I'll just double the thanks 👍 . The latter problem was fixed by #10482.
> 
> The artifacts built by Bazel here are part of [github/codeql-cli-binaries](https://github.com/github/codeql-cli-binaries), which in turn is built in an internal private repository. In light of that, we would prefer to keep the Bazel version in sync between the two. We need to look whether we are ready to bump up Bazel's version in the internal repository first.

yeah, that makes total sense, thanks!

PR #10486, opened 2022-09-20
Java: Delete some unused code

erik-krogh:
- The unused predicate in `NfaUtilsSpecific.qll` was not used anywhere, and people are not expected to use predicates from a `Specific.qll` file.  
- The `guardControls_v2` predicate in `Guards.qll` was only ever called by itself, and it was documented as being internal, so I deprecated it.  
- The `initializedToField` predicate was unused. It was in the `src/` folder, and it couldn't be imported from the outside, so I deleted it.

### DIFF
--- a/java/ql/lib/semmle/code/java/controlflow/Guards.qll
+++ b/java/ql/lib/semmle/code/java/controlflow/Guards.qll
@@ -218,12 +218,12 @@ predicate guardControls_v1(Guard guard, BasicBlock controlled, boolean branch) {
 }
 
 /**
- * INTERNAL: Use `Guards.controls` instead.
+ * DEPRECATED: Use `Guards.controls` instead.
  *
  * Holds if `guard.controls(controlled, branch)`, except this doesn't rely on
  * RangeAnalysis.
  */
-predicate guardControls_v2(Guard guard, BasicBlock controlled, boolean branch) {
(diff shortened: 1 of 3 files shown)


atorralba:
Shouldn't we still deprecate `RegExpFlags::getFlags` from `NfaUtilsSpecific.qll`? The fact that no one _should_ be using it doesn't mean they couldn't, and then we'd break their code by removing it if they did.

java/ql/lib/semmle/code/java/controlflow/Guards.qll
```diff
@@ -218,12 +218,12 @@ predicate guardControls_v1(Guard guard, BasicBlock controlled, boolean branch) {
...
 
 /**
- * INTERNAL: Use `Guards.controls` instead.
+ * DEPRECATED: Use `Guards.controls` instead.
  *
  * Holds if `guard.controls(controlled, branch)`, except this doesn't rely on
  * RangeAnalysis.
  */
-predicate guardControls_v2(Guard guard, BasicBlock controlled, boolean branch) {
+deprecated predicate guardControls_v2(Guard guard, BasicBlock controlled, boolean branch) {
```
aschackmull, line 226: I'd like to keep this for completeness, I think. Sure it's not used, but that's ok.

aschackmull (approved):

PR #10490, opened 2022-09-20
JS: Remove old Portal-based flow summary implementation

asgerf:
This is not maintained. It was marked as experimental, and the docs mentioned that it didn't support the `codeql` CLI, so I don't think a change note is needed.


erik-krogh (approved):
👍 

#dead-code-society

PR #10497, opened 2022-09-20
Java: Improve `ImportStaticTypeMember` and `ImportStaticOnDemand`

Marcono1234:
- `ImportStaticTypeMember` and `ImportStaticOnDemand` are now properly considering inherited members (except for inherited member types, see #5596).
- `ImportStaticOnDemand` predicates do not have non-`static` members and initializer methods as results anymore.
- `ImportOnDemandFromPackage.getAnImport()` does not have nested types as result anymore.

### DIFF
--- a/java/ql/lib/semmle/code/java/Type.qll
+++ b/java/ql/lib/semmle/code/java/Type.qll
@@ -841,11 +841,8 @@ class LocalClass extends LocalClassOrInterface, NestedClass {
 }
 
 /** A top-level type. */
-class TopLevelType extends RefType {
-  TopLevelType() {
-    not enclInReftype(this, _) and
-    (this instanceof Class or this instanceof Interface)
-  }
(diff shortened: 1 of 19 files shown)


java/ql/lib/semmle/code/java/Type.qll
```diff
@@ -841,11 +841,8 @@ class LocalClass extends LocalClassOrInterface, NestedClass {
 }
 
 /** A top-level type. */
-class TopLevelType extends RefType {
-  TopLevelType() {
-    not enclInReftype(this, _) and
-    (this instanceof Class or this instanceof Interface)
-  }
+class TopLevelType extends ClassOrInterface {
```
Marcono1234, line 844: This should be functionally equivalent.

smowton (approved):
Looks sensible to me

PR #10498, opened 2022-09-20
Java: Add `CompilationUnit.getATypeInScope()`

### LINKED ISSUE #4119: Java: Add Import.getATypeImport
When working with Javadoc it is often necessary to find out to which type a Javadoc tag refers to. Imports are one way to find this out, however the QL `Import` class does not define a common predicate to get an imported type despite all import types being able to import types in some way.
It would therefore be useful to add a predicate `Import.getATypeImport()` which would behave like this:
- `ImportType`: `getImportedType()`
- `ImportStaticTypeMember`: `getATypeImport()`
- `ImportOnDemandFromPackage`: `getAnImport()`
- `ImportOnDemandFromType`: `getAnImport()`
- `ImportStaticOnDemand`: `getATypeImport()`

Edit: As described by #10498 instead of adding this predicate, it might make more sense to have a predicate for `CompilationUnit` which obtains a type which is available by its simple same in that compilation unit.

Marcono1234:
This predicate is mainly helpful for Javadoc queries and for queries which check whether the name of an element shadows another type.

Resolves #4119
It makes probably most sense to implement the logic requested there for `CompilationUnit`, as done in this PR because there it is also possible to account for shadowing between different kinds of imports (though this is currently not implemented by this PR).

This PR is a bit experimental and does not have a dedicated test; feedback is appreciated.

### DIFF
--- a/java/ql/lib/semmle/code/java/CompilationUnit.qll
+++ b/java/ql/lib/semmle/code/java/CompilationUnit.qll
@@ -31,5 +31,50 @@ class CompilationUnit extends Element, File {
    */
   Module getModule() { cumodule(this, result) }
 
+  /**
+   * Gets a type which is available by its simple name in this compilation unit.
+   * Reasons for this can be:
+   * - The type is declared in this compilation unit as top-level type
+   * - The type is imported
+   * - The type is declared in the same package as this compilation unit
+   * - The type is declared in the package `java.lang`
+   */
+  ClassOrInterface getATypeAvailableBySimpleName() {
+    // See "Shadowing", https://docs.oracle.com/javase/specs/jls/se17/html/jls-6.html#jls-6.4.1
+    // Note: Currently the logic below does not consider shadowing and might have multiple results
+    // with the same type name
+    result.(TopLevelType).getCompilationUnit() = this
+    or
+    exists(ImportStaticTypeMember importDecl |
+      importDecl.getCompilationUnit() = this and
+      result = importDecl.getATypeImport()
+    )
+    or
+    exists(ImportType importDecl |
+      importDecl.getCompilationUnit() = this and
+      result = importDecl.getImportedType()
+    )
+    or
+    exists(ImportStaticOnDemand importDecl |
+      importDecl.getCompilationUnit() = this and
+      result = importDecl.getATypeImport()
+    )
+    or
+    exists(ImportOnDemandFromType importDecl |
+      importDecl.getCompilationUnit() = this and
+      result = importDecl.getAnImport()
+    )
+    or
+    exists(ImportOnDemandFromPackage importDecl |
+      importDecl.getCompilationUnit() = this and
+      result = importDecl.getAnImport()
+    )
+    or
+    // From same package or java.lang, see https://docs.oracle.com/javase/specs/jls/se17/html/jls-7.html
+    result.(TopLevelType).getPackage() = this.getPackage()
+    or
+    result.(TopLevelType).getPackage().hasName("java.lang")
+  }
+
   override string getAPrimaryQlClass() { result = "CompilationUnit" }
 }
--- a/java/ql/lib/semmle/code/java/CompilationUnit.qll
+++ b/java/ql/lib/semmle/code/java/CompilationUnit.qll
@@ -32,42 +32,30 @@ class CompilationUnit extends Element, File {
   Module getModule() { cumodule(this, result) }
 
   /**
-   * Gets a type which is available by its simple name in this compilation unit.
-   * Reasons for this can be:
-   * - The type is declared in this compilation unit as top-level type
-   * - The type is imported
-   * - The type is declared in the same package as this compilation unit
-   * - The type is declared in the package `java.lang`
+   * Gets a type which is available in the top-level scope of this compilation unit.
+   * This can be a type:
+   * - declared in this compilation unit as top-level type
+   * - imported with an `import` declaration
+   * - declared in the same package as this compilation unit
+   * - declared in the package `java.lang`
+   *
+   * This predicate not consider "shadowing", it can have types as result whose simple name is
+   * shadowed by another type in scope.
    */
-  ClassOrInterface getATypeAvailableBySimpleName() {
+  ClassOrInterface getATypeInScope() {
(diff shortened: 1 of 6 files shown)


java/ql/lib/semmle/code/java/CompilationUnit.qll
```diff
@@ -31,5 +31,50 @@ class CompilationUnit extends Element, File {
...
+      importDecl.getCompilationUnit() = this and
+      result = importDecl.getAnImport()
+    )
+    or
+    exists(ImportOnDemandFromPackage importDecl |
+      importDecl.getCompilationUnit() = this and
+      result = importDecl.getAnImport()
+    )
+    or
+    // From same package or java.lang, see https://docs.oracle.com/javase/specs/jls/se17/html/jls-7.html
```
smowton, line 73: Add `CompilationUnit.getAnImport()` to make this much briefer?

java/ql/lib/semmle/code/java/CompilationUnit.qll
```diff
@@ -31,5 +31,50 @@ class CompilationUnit extends Element, File {
...
 
+  /**
+   * Gets a type which is available by its simple name in this compilation unit.
+   * Reasons for this can be:
+   * - The type is declared in this compilation unit as top-level type
+   * - The type is imported
+   * - The type is declared in the same package as this compilation unit
+   * - The type is declared in the package `java.lang`
+   */
+  ClassOrInterface getATypeAvailableBySimpleName() {
```
smowton, line 42: Maybe `AvailableBySimpleName` -> `InScope`?

java/ql/lib/semmle/code/java/CompilationUnit.qll
Marcono1234: Do you mean as separate predicate: `getAnImport(): Import`?

java/ql/lib/semmle/code/java/CompilationUnit.qll
Marcono1234: In that case should the shadowing and simple name aspects be removed from the comments? Because a shadowed type would still be in scope.

java/ql/lib/semmle/code/java/CompilationUnit.qll
smowton: yes, so you don't need all the exists quantifiers but instead can do `this.getAnImport().(ImportStaticOnDemand).getAnImport()` for example

java/ql/lib/semmle/code/java/CompilationUnit.qll
smowton: I don't know off hand if the JLS formally defines "in scope"? In either case though explicitly mentioning this tricky corner seems useful, so I'd keep it.

java/ql/lib/semmle/code/java/CompilationUnit.qll
Marcono1234: I have simplified the code but decided against a separate predicate `getAnImport()` for know because I am not sure if it would be useful in any other situations.
I hope that is fine.

java/ql/lib/semmle/code/java/CompilationUnit.qll
```diff
@@ -31,5 +31,38 @@ class CompilationUnit extends Element, File {
...
+   * - declared in the same package as this compilation unit
+   * - declared in the package `java.lang`
+   *
+   * This predicate not consider "shadowing", it can have types as result whose simple name is
+   * shadowed by another type in scope.
+   */
+  ClassOrInterface getATypeInScope() {
+    // See "Shadowing", https://docs.oracle.com/javase/specs/jls/se17/html/jls-6.html#jls-6.4.1
+    // Currently shadowing is not considered
+    result.(TopLevelType).getCompilationUnit() = this
```
smowton, line 48: Redundant w.r.t. the same-package test

java/ql/src/Advisory/Documentation/ImpossibleJavadocThrows.ql
```diff
@@ -11,22 +11,18 @@
 
 import java
 
-RefType getTaggedType(ThrowsTag tag) {
+Class getTaggedType(ThrowsTag tag) {
```
smowton, line 14: Why Class not ClassOrInterface?

java/ql/src/Advisory/Documentation/ImpossibleJavadocThrows.ql
```diff
@@ -11,22 +11,18 @@
...
+  result = tag.getFile().(CompilationUnit).getATypeInScope()
 }
 
-predicate canThrow(Callable callable, RefType exception) {
-  exists(string uncheckedException |
-    uncheckedException = "RuntimeException" or uncheckedException = "Error"
-  |
-    exception.getAnAncestor().hasQualifiedName("java.lang", uncheckedException)
-  )
+predicate canThrow(Callable callable, Class exception) {
```
smowton, line 19: Same question re: throwing things by an interface type

java/ql/lib/semmle/code/java/CompilationUnit.qll
Marcono1234: You are right. Currently this structure matches (roughly?) the precedence order of shadowing, but I could drop this nonetheless if you want.

java/ql/src/Advisory/Documentation/ImpossibleJavadocThrows.ql
Marcono1234: All exception types are classes though, so you cannot declare an interface type to be thrown, can you?
(neither in Kotlin, right?)

java/ql/src/Advisory/Documentation/ImpossibleJavadocThrows.ql
Marcono1234: Good point. Previously this also detected (unintentionally?) when an interface was erroneously documented as thrown type. I have changed this to `ClassOrInterface` now.

java/ql/src/Advisory/Documentation/ImpossibleJavadocThrows.ql
smowton: Doh right, I had `Exception` in mind as an interface. Never mind.

smowton (approved):

PR #10505, opened 2022-09-21
Data flow: Guard against `viableImplInCallContext` not being a subset of `viableCallable`

hvitved:
This PR adds a guard against, and consistency query for, cases where `callable = viableImplInCallContext(call, _)` does not imply `callable = viableCallable(call)`.

As it turned out, this revealed a bug in the C# implementation, which is also fixed on this PR.


tausbn (approved):
Python/shared changes look reasonable to me. 👍

aschackmull (approved):
Right, this was already an implicit consistency requirement, so making it more explicit LGTM.
Also, adding the filter to enforce it in `DataFlowImplCommon` looks fairly free, so that seems like a reasonable belt-and-suspenders fix.

PR #10510, opened 2022-09-21
C++: Fix FPs for cpp/unused-static-function in files that were not extracted completely

geoffw0:
Fix FPs for `cpp/unused-static-function` in files that were not extracted completely, e.g. due to a compilation error part way through the file.  As the test shows, this may hide some good results as well, but we expect the majority of results in incompletely extracted files to be false positives caused by the incomplete extraction.

### DIFF
--- a/cpp/ql/src/Best Practices/Unused Entities/UnusedStaticFunctions.ql
+++ b/cpp/ql/src/Best Practices/Unused Entities/UnusedStaticFunctions.ql
@@ -14,15 +14,28 @@
 import cpp
 
 predicate immediatelyReachableFunction(Function f) {
-  not f.isStatic() or
-  exists(BlockExpr be | be.getFunction() = f) or
-  f instanceof MemberFunction or
-  f instanceof TemplateFunction or
-  f.getFile() instanceof HeaderFile or
-  f.getAnAttribute().hasName("constructor") or
-  f.getAnAttribute().hasName("destructor") or
-  f.getAnAttribute().hasName("used") or
+  not f.isStatic()
+  or
+  exists(BlockExpr be | be.getFunction() = f)
+  or
+  f instanceof MemberFunction
+  or
+  f instanceof TemplateFunction
+  or
+  f.getFile() instanceof HeaderFile
+  or
+  f.getAnAttribute().hasName("constructor")
+  or
+  f.getAnAttribute().hasName("destructor")
+  or
+  f.getAnAttribute().hasName("used")
+  or
   f.getAnAttribute().hasName("unused")
+  or
+  // a compiler error in the same file suggests we may be missing data
+  exists(Diagnostic d | d.getFile() = f.getFile() and d.getSeverity() >= 3)
+  or
+  exists(ErrorExpr ee | ee.getFile() = f.getFile())
 }
 
 predicate immediatelyReachableVariable(Variable v) {
--- a/cpp/ql/src/change-notes/2022-09-21-unused-static-function.md
+++ b/cpp/ql/src/change-notes/2022-09-21-unused-static-function.md
@@ -1,0 +1,4 @@
+---
+category: minorAnalysis
+---
+* Fixed false positives from the "Unused static function" (`cpp/unused-static-function`) query in files that had errors during compilation.
--- a/cpp/ql/test/query-tests/Best Practices/Unused Entities/UnusedStaticFunctions/UnusedStaticFunctions.expected
+++ b/cpp/ql/test/query-tests/Best Practices/Unused Entities/UnusedStaticFunctions/UnusedStaticFunctions.expected
@@ -1,5 +1,3 @@
-| extraction_error.c:4:13:4:43 | my_function2_called_after_error | Static function my_function2_called_after_error is unreachable | extraction_error.c:4:13:4:43 | my_function2_called_after_error | my_function2_called_after_error |
-| extraction_error.c:5:13:5:35 | my_function3_not_called | Static function my_function3_not_called is unreachable | extraction_error.c:5:13:5:35 | my_function3_not_called | my_function3_not_called |
 | unused_functions.c:16:13:16:27 | unused_function | Static function unused_function is unreachable | unused_functions.c:16:13:16:27 | unused_function | unused_function |
 | unused_functions.c:20:13:20:28 | unused_function2 | Static function unused_function2 is unreachable ($@ must be removed at the same time) | unused_functions.c:24:13:24:28 | unused_function3 | unused_function3 |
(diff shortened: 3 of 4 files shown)


geoffw0:
I'm just looking at a possible performance issue on this PR...

MathiasVP (commented):
Do we want to run a DCA on this once you've fixed the performance issue?

geoffw0:
> Do we want to run a DCA on this once you've fixed the performance issue?

I'm actually fairly confident, the issue was an uncomplicated cartesian product.  I will start a DCA run to be sure.

cpp/ql/src/Best Practices/Unused Entities/UnusedStaticFunctions.ql
```diff
@@ -13,6 +13,12 @@
 
 import cpp
 
+predicate possiblyIncompleteFile(File f) {
```
MathiasVP, line 16: Should we mark this as `pragma[noinline]` to prevent the compiler undoing your join-order fix?

geoffw0: Done.  I think the optimizer would be in a pretty bad state if it did inline that, mind you!

MathiasVP (dismissed):
LGTM if DCA is happy!

MathiasVP:
Looks like this change removes > 200 results on DCA. I think we should check a subset of those to see if this looks reasonable. What do you say, @geoffw0?

geoffw0:
Yep, I'm not surprised if it removes a lot of results (and some of them will be TPs), but we should check at least a proportion of them are really FP.

geoffw0:
Do you know where to find the total number of results for `cpp/unused-static-function` in the DCA run?  Losing 236 is a lot, but I have a feeling its a noisy query in the first place.

MathiasVP:
> Do you know where to find the total number of results for cpp/unused-static-function in the DCA run? Losing 236 is a lot, but I have a feeling its a noisy query in the first place.

I don't think this is possible out-of-the-box, no. There _might_ be an option somewhere to tell it to generate the full list of alerts. We could ask in the DCA Slack channel about this if you think it'd be useful.

geoffw0:
I think it would be useful in cases like this, yes, but I'm tracking a million things today already.  Feel free to talk to them.

geoffw0:
DCA:

I'm ignoring the 5% overall increase in query run times, because this is a change to one QL file and it isn't very plausible its had an effect on so many other queries.

Looking at some of the results:
 
Kitware__CMake: Source/cmCPluginAPI.cxx:25:20:25:35 (represents 54 similar results)
 - a pointer to the function `cmGetClientData` is stored in an array `cmStaticCAPI`, but that array does not appear to be used anywhere.  I believe this is probably a TP for the query, so an unfortunate loss for this change.
 - there are two `ErrorExpr`s in the file but I believe (based on `Location`s) the entire file was extracted.

abseil-cpp-linux: absl/base/exception_safety_testing_test.cc:526:6:526:13
 - `DummyOp` is only referenced inside a `decltype` expression; if I'm understanding correctly that means the type of the function is used but the body is not, which is probably a TP for the query?
 - `ErrorExpr`s thoughout the file, but I believe the entire file was extracted.

systemd__systemd: src/analyze/analyze-security.c:148:22:148:40
 - `security_info_free` is a macro arg to `DEFINE_TRIVIAL_CLEANUP_FUNC` which creates a function calling `security_info_free`, but it isn't clear whether *that* function is called anywhere.
 - one `ErrorExpr` and one `Diagnostic`, but I believe the entire file was extracted.

Overall, many of these cases are in tests (so we wouldn't normally report them) and in macro or template heavy code (so difficult to understand manually).

On MRVA I see the total number of results in 69 projects decrease from 779 to 416 (47% reduction), which seems fine for such a noisy query.  Most cases we lose had both `ErrorExpr`s and `Diagnostic`s in the same file (unlike 2 of the 3 cases above), suggesting that something had gone more seriously wrong, though its still difficult to diagnose individual cases with any confidence.  I get most of my confidence from the QL tests.

We could remove the part of the change that looks at `ErrorExpr`s, to focus the change to the more severe cases (where we're likely to have successfully extracted less of the code)???

cpp/ql/src/Best Practices/Unused Entities/UnusedStaticFunctions.ql
```diff
@@ -13,16 +13,34 @@
 
 import cpp
 
+pragma[noinline]
+predicate possiblyIncompleteFile(File f) {
+  exists(Diagnostic d | d.getFile() = f and d.getSeverity() >= 3)
+  or
+  exists(ErrorExpr ee | ee.getFile() = f)
```
jketema, line 20: This is a very bad indicator of an incomplete file for the same reason an exception is made in the IR translation here: https://github.com/github/codeql/blob/main/cpp/ql/lib/semmle/code/cpp/ir/implementation/raw/internal/TranslatedElement.qll#L116-L119

MathiasVP: Oh, good point! I had totally forgotten about that special case. Thanks for the heads up. Now go back to your vacation 😂!

geoffw0: I think this settles that we only want to keep the `Diagnostic`s case.  My testing supports that this was generally the better indicator in practice anyway.

Though I will note there are several other queries that take the presence of an `ErrorExpr` to indicate incomplete extraction.

geoffw0:
I've just done the change to just basing this on `Diagnostic`s, not `ErrorExpr`s.

Tests still pass.  There are now 717 MRVA results on 69 repos (8% less than original).  Does anyone want to see another DCA run?

geoffw0:
DCA showed:
 - 107 fixed alerts for `cpp/unused-static-function` in `systemd__systemd`, 32 in `zeek__spicy` and 1 in `microsoft__ChakraCore`.
 - the 54 (dubious) changes to results on `Kitware__CMake` from the previous DCA run are gone; as are the 4 from `boostorg__boost`, 3 from `openjdk__jdk` and 3 from `openttd__openttd`.
 - sampling some of the fixed alerts (one from each affected project), I am pretty happy with the changes (though its difficult to see what is and isn't used by hand).
 - so I think the changes have reduced the number of "fixed alerts" greatly, but improved the ratio of them that I would deem to be "probably correct".

jketema (approved):
LGTM.

I do expect the changes in DCA results to be somewhat less dramatic in the first nightly we run after this is merged, as frontend fixes went in affect `systemd` and `zeek/spicy`

geoffw0:
Merged.  Thanks for reviewing!

PR #10511, opened 2022-09-21
Swift: fix `IfConfigDecl` in QL libraries

redsun82:
This fixes `IfConfigDecl` for both the AST printer and control flow libraries.

It turns out that the active lements of an `IfConfigDecl` are already listed in the enclosing scope (like a `BraceStmt`), so they should not be listed as children, and `IfConfigDecl` can be jsut a leaf in the control flow.


MathiasVP (approved):
LGTM! I _think_ we need some more code to handle `IfConfigDecl`s in global scope, but we can leave that as a separate PR.

MathiasVP:
There's a conflict file now :(

redsun82:
> LGTM! I _think_ we need some more code to handle `IfConfigDecl`s in global scope, but we can leave that as a separate PR.

I suspect that we do not flow from `TopLevelCodeDecl`s to their `BraceStmt` bodies

MathiasVP (approved):

PR #10507, opened 2022-09-21
CPP: Make more alert-messages follow the style guide

erik-krogh:
The tests are failing and that's fine, there is an internal PR where the tests pass (the internal PR updates some expected test outputs). 

I've been working on making the alert-messages consistent across languages, and some fan out work from that is ensuring that the alert-messages follow the style guide (which I've also revised).  

Fixes done as part of this PR: 
 - Add a full stop at the end of the alert-message. 
 - Avoid "here", but instead have something more descriptive. (Like "this allocation" or "expression").  
 - Not having single-quotes around links (replacing `'$@'` with `$@`).  
 - Make the alert-message the same as another language. 
 - Removing links that linked to the alert-location. 
 - Use `depends on` for taint-tracking queries and `flows to` for dataflow-queries.  
 - Enclose code with single quotes. 
 - Start an alert-message with an upper-case letter. 
 
Some of the messages become inconsistent with other languages, but I'm updating one language at a time, so some will get out of sync.  

Some other PRs in this series: [JS](https://github.com/github/codeql/pull/10286), [Py](https://github.com/github/codeql/pull/10252), [Rb](https://github.com/github/codeql/pull/10304), [Go](https://github.com/github/codeql/pull/10413), [Java](https://github.com/github/codeql/pull/10528) (draft).    
[Style guide update](https://github.com/github/codeql/pull/10405).

PR #10513, opened 2022-09-21
QL: improve the `ql/alert-message-style-violation` query. 

erik-krogh:
Each commit should describe itself. 

The `getASubExpression` predicate was outlined to try and improve a bad join-order 
I wasn't that succesful in improving the join-order, there is still a query specific predicate that takes ~7s on my machine.   
But there are still a few other queries that are slower, so it shouldn't affect the total runtime that much. 

These improvements finds ~~97~~ 181 more violations (see the code-scanning results).  
I'll look at fixing these in other PRs. E.g. [this upcoming one for C++](https://github.com/github/codeql/pull/10507).

### DIFF
--- a/ql/ql/src/codeql_ql/ast/Ast.qll
+++ b/ql/ql/src/codeql_ql/ast/Ast.qll
@@ -186,7 +186,9 @@ class QueryDoc extends QLDoc {
   override string getAPrimaryQlClass() { result = "QueryDoc" }
 
   /** Gets the @kind for the query */
-  string getQueryKind() { result = this.getContents().regexpCapture("(?s).*@kind (\\w+)\\s.*", 1) }
+  string getQueryKind() {
+    result = this.getContents().regexpCapture("(?s).*@kind ([\\w-]+)\\s.*", 1)
+  }
 
   /** Gets the id part (without language) of the @id */
   string getQueryId() {
@@ -241,6 +243,12 @@ class Select extends TSelect, AstNode {
    * Gets the `i`th expression in the `select` clause.
    */
   Expr getExpr(int i) { toQL(result) = sel.getChild(_).(QL::AsExprs).getChild(i) }
+
+  Expr getMessage() {
+    if this.getQueryDoc().getQueryKind() = "path-problem"
+    then result = this.getExpr(3)
+    else result = this.getExpr(1)
+  }
 
   // TODO: This gets the `i`th order-by, but some expressions might not have an order-by.
   Expr getOrderBy(int i) { toQL(result) = sel.getChild(_).(QL::OrderBys).getChild(i).getChild(0) }
--- a/ql/ql/src/codeql_ql/ast/Ast.qll
+++ b/ql/ql/src/codeql_ql/ast/Ast.qll
@@ -1403,6 +1403,14 @@ class ComparisonFormula extends TComparisonFormula, Formula {
     or
     pred = directMember("getRightOperand") and result = this.getRightOperand()
   }
+
+  /** Hplds if this comparison has the operands `a` and `b` (in any order). */
+  pragma[noinline]
+  predicate hasOperands(Expr a, Expr b) {
+    this.getLeftOperand() = a and this.getRightOperand() = b
+    or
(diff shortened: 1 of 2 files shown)


ql/ql/src/codeql_ql/ast/Ast.qll
```diff
@@ -1403,6 +1411,14 @@ class ComparisonFormula extends TComparisonFormula, Formula {
     or
     pred = directMember("getRightOperand") and result = this.getRightOperand()
   }
+
+  /** Hplds if this comparison has the operands `a` and `b` (in any order). */
```
tausbn, line 1415:
```suggestion
  /** Holds if this comparison has the operands `a` and `b` (in any order). */
```

ql/ql/src/queries/style/AlertMessage.ql
```diff
@@ -30,6 +34,14 @@ private AstNode getSelectPart(Select sel, int index) {
             )
           )
         )
+        or
+        // the string is a variable that is assigned in the `where` clause.
+        exists(VarAccess v, ComparisonFormula comp, String str |
+          v = getASubExpression(sel) and
+          loc = v.getLocation() and
+          comp.hasOperands(v.getDeclaration().getAnAccess(), str) and
```
tausbn, line 42: I wonder... Can `v.getDeclaration().getAnAccess()` ever be a `String`? If not, it would seem that
```ql
comp.getAnOperand() = v.getDeclaration().getAnAccess() and
comp.getAnOperand() = str
```
would suffice.

ql/ql/src/queries/style/AlertMessage.ql
```diff
@@ -164,6 +176,14 @@ String wrongFlowsPhrase(Select sel, string kind) {
   )
 }
 
+/**
+ * Gets a string element that contains double whitespace.
+ */
+String doubleWhitespace(Select sel) {
+  result = getSelectPart(sel, _) and
+  result.getValue().regexpMatch(".*\\s\\s.*")
```
tausbn, line 184: Will the `.*` bits match newlines as well (this usually requires a regex flag), or is this perhaps an impossibility anyway?

ql/ql/src/codeql_ql/ast/Ast.qll
```diff
@@ -186,7 +186,9 @@ class QueryDoc extends QLDoc {
   override string getAPrimaryQlClass() { result = "QueryDoc" }
 
   /** Gets the @kind for the query */
-  string getQueryKind() { result = this.getContents().regexpCapture("(?s).*@kind (\\w+)\\s.*", 1) }
+  string getQueryKind() {
+    result = this.getContents().regexpCapture("(?s).*@kind ([\\w-]+)\\s.*", 1)
```
tausbn, line 190: For maximal future-proofing, we could also just match a non-zero number of _non_-whitespace characters instead.

tausbn (dismissed):
A few minor comments here and there, but otherwise this looks good to me. :+1:

I'm slightly amused that it flags the alert messages in the query itself.

ql/ql/src/queries/style/AlertMessage.ql
erik-krogh: Yes, that would suffice, and that was also how I had it originally.   
It put it like that to try and improve a join order.  
I'm not sure it worked, but I kept it.

ql/ql/src/queries/style/AlertMessage.ql
erik-krogh: There are no message with "non-space whitespace" currently, and I'm not sure you can have an unescaped newline in a string literal in the current QL language.  

(I want multi-line string literals though 🙏 ).

ql/ql/src/codeql_ql/ast/Ast.qll
erik-krogh: I think I prefer to keep it as is.  
E.g. I'm not sure `@` is valid in an alert-message.

tausbn (approved):
Thanks for the quick replies. I'm happy to see this merged. :+1:

PR #10517, opened 2022-09-21
Ruby: Add query for debugging regexp flow

hvitved:
It is often convenient to be able to debug flow from `RegExpConfiguration`, so this PR adds a debug query for doing just that.


aibaars (approved):

nickrolfe (approved):
LGTM

PR #10518, opened 2022-09-21
Fixing wrong example

michaelpradel:
The definition of the ``ancestorOf`` predicate doesn't really describe an ancestor relation, but its inverse (i.e., something one might call ``descendantOf``). The fix replaces ``parentOf`` with ``childOf``, so that the predicate actually describes the ancestor relation. I've tested the predicate against the tutorial dataset, and it now yields the expected results.

### DIFF
--- a/docs/codeql/writing-codeql-queries/crown-the-rightful-heir.rst
+++ b/docs/codeql/writing-codeql-queries/crown-the-rightful-heir.rst
@@ -104,8 +104,8 @@ You can translate this into QL as follows:
 .. code-block:: ql
 
    Person ancestorOf(Person p) {
-     result = parentOf(p) or
-     result = parentOf(ancestorOf(p))
+     result = childOf(p) or
+     result = childOf(ancestorOf(p))
    }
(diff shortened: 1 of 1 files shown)


intrigus-lgtm:
I don't see any problem with this code.

Let's assume that we have the following facts:
The father of Alice is Bob.
The father of Eve is Bob.
Then `parentOf(Alice) = Bob` and `parentOf(Eve) = Bob`.
(`parentOf(Person p)`:  returns a parent of `p`)
So `ancestorOf(Alice) = parentOf(Alice) = Bob` which makes sense, doesn't it (ignoring the recursive case of `ancestorOf`)?

michaelpradel:
Yes, you're right. I was incorrectly understanding ``ancestorOf(x)`` as returning persons who have ``x`` as an ancestor, but it is meant to return persons who are ancestors of ``x``. In my testing, I was using the query based on my logically inverted understanding, which overall gave the correct results again. Thanks for clarifying!

PR #10523, opened 2022-09-21
Java: Disable Kotlin element of test re: database inconsistency exposed by JDK18 extractor upgrade

smowton:
This is disabled for now because Kotlin doesn't extract java.lang.CharSequence.toString, even though that interface redeclares toString, and JDK18 points a call (syntactically to Editable.toString) that previous JDKs pointed at Object.toString. This produces a database inconsistency that should be fixed Kotlin-side.


atorralba (approved):
When this gets fixed we will be able to also get rid of https://github.com/github/codeql/blob/bc2de7ed4b14b5f282a678d0d2bd39d026f3ceb5/java/ql/lib/semmle/code/java/frameworks/android/Widget.qll#L19-L37

atorralba (approved):

PR #10529, opened 2022-09-22
QL: A few more improvements to `ql/alert-message-style-violation`

erik-krogh:
`don't repeat the alert-location as a link` is not currently part of the style guide, and I don't think it needs to be.  
(It's covered under: Keep the message concise).  

I tried to use GVN when I created [don't repeat the alert-location as a link](https://github.com/github/codeql/commit/76fef28c32eef22346f3c417ba55e9f45c5c8048).  
But that seemed to run into an infinite loop (I stopped it after ~100.000 iterations).  
/cc @MathiasVP

### DIFF
--- a/ql/ql/src/codeql/GlobalValueNumbering.qll
+++ b/ql/ql/src/codeql/GlobalValueNumbering.qll
@@ -129,6 +129,7 @@ private predicate classPredicateCallValueNumber(
 
 private predicate literalValueNumber(Literal lit, string value, Type t) {
   lit.(String).getValue() = value and
+  value.length() <= 50 and
   t instanceof StringClass
   or
   lit.(Integer).getValue().toString() = value and
--- a/ql/ql/src/queries/style/AlertMessage.ql
+++ b/ql/ql/src/queries/style/AlertMessage.ql
@@ -102,11 +102,11 @@ String shouldStartCapital(Select sel) {
  * select foo(), "XSS from using a unsafe value." // <- good
  * ```
  */
-String avoidHere(string part) {
+String avoidHere(Select sel, string part) {
   part = ["here", "this location"] and
   (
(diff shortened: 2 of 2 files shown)


MathiasVP:
> I tried to use GVN when I created [don't repeat the alert-location as a link](https://github.com/github/codeql/commit/76fef28c32eef22346f3c417ba55e9f45c5c8048).
> But that seemed to run into an infinite loop (I stopped it after ~100.000 iterations).
> /cc @MathiasVP

Yeah, 100.000 iterations definitely doesn't sound right. GVN tends to go bad if it's running on a database with unfortunate inconsistencies that results in multiple results for a predicate that's only expected to have a unique result (i.e., if a variable access has multiple types, or if a field has multiple qualifiers).

erik-krogh:
> Yeah, 100.000 iterations definitely doesn't sound right. GVN tends to go bad if it's running on a database with unfortunate inconsistencies that results in multiple results for a predicate that's only expected to have a unique result (i.e., if a variable access has multiple types, or if a field has multiple qualifiers).

We definitely have some non-unique results for some predicates with how we support parameterized modules (collapse all the things into one).  
That might be it.  
It should be possible to get the GVN library to work around that 🤔

MathiasVP:
> I'm thinking that GVN just ignores those cases, and document that behavior.

That sounds good to me, indeed.

erik-krogh:
> That sounds good to me, indeed.

Removing string constants of length > 50 from the GVN computation fixed the infinite loop 😕 
I have no idea why, but it seems like a somewhat reasonable restriction.   
Why do we need GVN for literals anyway? 

Btw: It found more results, so your GVN library works 🎉

MathiasVP (approved):

PR #10531, opened 2022-09-22
Ruby: use consistent capitalization with `import ... as`

nickrolfe:


nickrolfe:
This fixes 12 QL-for-QL alerts.

hvitved (approved):

PR #10532, opened 2022-09-22
Final mergeback from `rc/3.7`

henrymercer:
There are no changes from `rc/3.7` to merge back, so this PR simply bumps the minor version of every pack that we regularly release.


erik-krogh (approved):

edoardopirovano (approved):
LGTM. I think it's fine we're not bumping the version number for the `shared` packs as these haven't been release yet. We'll want to update our instructions before the next RC mergeback, though, as they will need to be bumped then.

PR #10533, opened 2022-09-22
Java: Add support for java.util.StringJoiner

pwntester:


atorralba:
Hey @pwntester, thanks for this! I added some more CSV rows for the fluent methods of `StringJoiner`, tests, and a change note.

java/ql/lib/semmle/code/java/frameworks/Strings.qll
```diff
@@ -64,7 +64,14 @@ private class StringSummaryCsv extends SummaryModelCsv {
...
         "java.lang;CharSequence;true;subSequence;;;Argument[-1];ReturnValue;taint;manual",
-        "java.lang;CharSequence;true;toString;;;Argument[-1];ReturnValue;taint;manual"
+        "java.lang;CharSequence;true;toString;;;Argument[-1];ReturnValue;taint;manual",
+        "java.util;StringJoiner;true;StringJoiner;;;Argument[0..2];Argument[-1];taint;manual",
+        "java.util;StringJoiner;true;add;;;Argument[0];Argument[-1];taint;manual",
+        "java.util;StringJoiner;true;add;;;Argument[-1];ReturnValue;value;manual",
+        "java.util;StringJoiner;true;merge;;;Argument[0];Argument[-1];taint;manual",
+        "java.util;StringJoiner;true;merge;;;Argument[-1];ReturnValue;value;manual",
+        "java.util;StringJoiner;true;setEmptyValue;;;Argument[-1];ReturnValue;value;manual",
+        "java.util;StringJoiner;true;toString;;;Argument[-1];ReturnValue;taint;manual"
```
Marcono1234, line 74: Might be useful to add a trailing comma to avoid a diff for this line when new lines are added:
```suggestion
        "java.util;StringJoiner;true;toString;;;Argument[-1];ReturnValue;taint;manual",
```

java/ql/lib/semmle/code/java/frameworks/Strings.qll
```diff
@@ -64,7 +64,14 @@ private class StringSummaryCsv extends SummaryModelCsv {
         "java.lang;StringBuilder;true;StringBuilder;;;Argument[0];Argument[-1];taint;manual",
         "java.lang;CharSequence;true;charAt;;;Argument[-1];ReturnValue;taint;manual",
         "java.lang;CharSequence;true;subSequence;;;Argument[-1];ReturnValue;taint;manual",
-        "java.lang;CharSequence;true;toString;;;Argument[-1];ReturnValue;taint;manual"
+        "java.lang;CharSequence;true;toString;;;Argument[-1];ReturnValue;taint;manual",
+        "java.util;StringJoiner;true;StringJoiner;;;Argument[0..2];Argument[-1];taint;manual",
```
Marcono1234, line 68: Not sure if it matters, but I think these could all be `false` because `StringJoiner` is `final`.
```suggestion
        "java.util;StringJoiner;false;StringJoiner;;;Argument[0..2];Argument[-1];taint;manual",
```

java/ql/lib/semmle/code/java/frameworks/Strings.qll
```diff
@@ -64,7 +64,14 @@ private class StringSummaryCsv extends SummaryModelCsv {
...
+        "java.util;StringJoiner;true;add;;;Argument[0];Argument[-1];taint;manual",
+        "java.util;StringJoiner;true;add;;;Argument[-1];ReturnValue;value;manual",
+        "java.util;StringJoiner;true;merge;;;Argument[0];Argument[-1];taint;manual",
+        "java.util;StringJoiner;true;merge;;;Argument[-1];ReturnValue;value;manual",
+        "java.util;StringJoiner;true;setEmptyValue;;;Argument[-1];ReturnValue;value;manual",
```
Marcono1234, line 73: Should there also be taint from the `setEmptyValue` argument to the `StringJoiner` instance?

Marcono1234 (commented):
Hopefully these comments are useful.
Feel free to consider them only as suggestions since I am not a project member.

atorralba:
Thanks @Marcono1234, those are good suggestions. I had doubts about `setEmptyValue` and that's why I didn't add it in the first place, but even if taint would only get propagated in certain cases (when no other strings are joined) I guess it makes sense to have the model.

Applied in e849acb.

java/ql/lib/semmle/code/java/frameworks/Strings.qll
Marcono1234: @atorralba, not sure if this is relevant, but the tests still have comments with <code>java.util;StringJoiner;<b>true</b></code>.

atorralba: Well, it's better that we stay accurate 😄 Fixed, thanks!

PR #10535, opened 2022-09-22
Python: Model `flask.jsonify`

RasmusWL:


python/ql/lib/semmle/python/frameworks/Flask.qll
```diff
@@ -171,6 +171,24 @@ module Flask {
...
+     *
+     * See
+     * - https://flask.palletsprojects.com/en/2.2.x/api/#flask.json.jsonify
+     */
+    private class FlaskJsonifyCall extends InstanceSource, DataFlow::CallCfgNode {
+      FlaskJsonifyCall() {
+        this = API::moduleImport("flask").getMember("jsonify").getACall()
+      }
+
+      override DataFlow::Node getBody() { result = this.getArg(0) }
```
tausbn, line 185: Looking at the documentation for `flask.jsonify`, it seems like it's really _any_ argument that's put into the response, so maybe this should be `getArg(_)` (or something similar)?

tausbn (approved):
One minor question, otherwise this looks good to me. :+1:

python/ql/lib/semmle/python/frameworks/Flask.qll
RasmusWL: good catch :+1:

tausbn (approved):
Nice!

PR #10538, opened 2022-09-22
Ruby: Model flow through ActionController::Parameters

hmac:
Add flow summaries for methods on `ActionController::Parameters`,
which mostly propagate taint from receiver to return value.

[Evaluation](https://github.com/github/codeql-dca-main/issues/7317) shows that we catch a couple more TPs due to flow through parameters now reaching some sinks.

### DIFF
--- a/ruby/ql/test/query-tests/security/cwe-601/UrlRedirect.expected
+++ b/ruby/ql/test/query-tests/security/cwe-601/UrlRedirect.expected
@@ -10,9 +10,7 @@ edges
 | UrlRedirect.rb:63:38:63:43 | call to params :  | UrlRedirect.rb:63:38:63:49 | ...[...] |
 | UrlRedirect.rb:68:38:68:43 | call to params :  | UrlRedirect.rb:68:38:68:49 | ...[...] |
 | UrlRedirect.rb:73:25:73:30 | call to params :  | UrlRedirect.rb:73:25:73:36 | ...[...] |
-| UrlRedirect.rb:88:21:88:32 | input_params :  | UrlRedirect.rb:89:5:89:16 | input_params :  |
 | UrlRedirect.rb:88:21:88:32 | input_params :  | UrlRedirect.rb:89:5:89:29 | call to permit :  |
-| UrlRedirect.rb:89:5:89:16 | input_params :  | UrlRedirect.rb:89:5:89:29 | call to permit :  |
 nodes
 | UrlRedirect.rb:4:17:4:22 | call to params | semmle.label | call to params |
 | UrlRedirect.rb:9:17:9:22 | call to params :  | semmle.label | call to params :  |
@@ -35,7 +33,6 @@ nodes
 | UrlRedirect.rb:73:25:73:30 | call to params :  | semmle.label | call to params :  |
 | UrlRedirect.rb:73:25:73:36 | ...[...] | semmle.label | ...[...] |
 | UrlRedirect.rb:88:21:88:32 | input_params :  | semmle.label | input_params :  |
-| UrlRedirect.rb:89:5:89:16 | input_params :  | semmle.label | input_params :  |
 | UrlRedirect.rb:89:5:89:29 | call to permit :  | semmle.label | call to permit :  |
 subpaths
 | UrlRedirect.rb:24:31:24:36 | call to params :  | UrlRedirect.rb:88:21:88:32 | input_params :  | UrlRedirect.rb:89:5:89:29 | call to permit :  | UrlRedirect.rb:24:17:24:37 | call to filter_params |


ruby/ql/lib/codeql/ruby/frameworks/ActionController.qll
```diff
@@ -357,3 +357,123 @@ private class ActionControllerProtectFromForgeryCall extends CsrfProtectionSetti
...
+      this.asExpr().getExpr() instanceof ParamsCall
+      or
+      exists(DataFlow::CallNode call | call = this |
+        call.getReceiver() instanceof ParamsInstance and
+        call.getMethodName() = paramsMethodReturningParamsInstance()
+      )
+      or
+      exists(DataFlow::LocalSourceNode prev | prev.flowsTo(this))
+    }
+  }
```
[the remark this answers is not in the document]
asgerf: That is inevitable, I'm afraid. If `ParamsCall` depends on API graphs, shouldn't you already have an API node for it? In that case, using the API graph members will have to be good enough.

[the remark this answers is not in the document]
asgerf: Actually on second thought it seems really easy to add https://github.com/github/codeql/pull/10603. Would this work for you?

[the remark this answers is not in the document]
asgerf: OK, I've taken a look at that model and the problem arises when we're using API graphs to populate concepts that are based on AST nodes. Having to step outside API graphs should be considered tech debt, and not be a common case, and I believe Ruby's API graphs need to be developed a bit further for this to work. In particular we'll need to add `MethodDefinitionNode` (in JS this would correspond to a `MkDef` wrapping a `DataFlow::FunctionNode`) and something like the `MkClassInstance` node from JS. I'll look into that, but it will not arrive in time for this PR.

alexrford (dismissed):
Really nice, the changes make sense and the new results look legit!

ruby/ql/lib/codeql/ruby/frameworks/ActionController.qll
```diff
@@ -357,3 +357,122 @@ private class ActionControllerProtectFromForgeryCall extends CsrfProtectionSetti
...
+  private class ParamsInstance extends DataFlow::Node {
+    ParamsInstance() {
+      this.asExpr().getExpr() instanceof ParamsCall
+      or
+      exists(DataFlow::CallNode call | call = this |
+        call.getReceiver() instanceof ParamsInstance and
+        call.getMethodName() = paramsMethodReturningParamsInstance()
+      )
+      or
+      exists(DataFlow::LocalSourceNode prev | prev.flowsTo(this))
```
hvitved, line 377: Shouldn't `prev` be restricted?

ruby/ql/lib/codeql/ruby/frameworks/ActionController.qll
```diff
@@ -357,3 +357,122 @@ private class ActionControllerProtectFromForgeryCall extends CsrfProtectionSetti
...
+
+  /**
+   * An instance of `ActionController::Parameters`, including those returned
+   * from method calls on other instances.
+   */
+  private class ParamsInstance extends DataFlow::Node {
+    ParamsInstance() {
+      this.asExpr().getExpr() instanceof ParamsCall
+      or
+      exists(DataFlow::CallNode call | call = this |
```
hvitved, line 372: I tend to write `this = any(DataFlow::CallNode call` instead.

ruby/ql/lib/codeql/ruby/frameworks/ActionController.qll
```diff
@@ -357,3 +357,122 @@ private class ActionControllerProtectFromForgeryCall extends CsrfProtectionSetti
...
+
+    override MethodCall getACall() {
+      result.getMethodName() = ["merge", "reverse_merge", "with_defaults"] and
+      exists(ParamsInstance i |
+        i.asExpr().getExpr() = [result.getReceiver(), result.getArgument(0)]
+      )
+    }
+
+    override predicate propagatesFlowExt(string input, string output, boolean preservesValue) {
+      input = ["Argument[self]", "Argument[0]"] and
```
hvitved, line 449: Presumably, only `Argument[self]` when `exists(ParamsInstance i | i.asExpr().getExpr() = result.getReceiver())` (and similarly for `Argument[0]`)?

ruby/ql/lib/codeql/ruby/frameworks/ActionController.qll
```diff
@@ -357,3 +357,122 @@ private class ActionControllerProtectFromForgeryCall extends CsrfProtectionSetti
...
+
+    override MethodCall getACall() {
+      result.getMethodName() = ["merge!", "reverse_merge!", "with_defaults!"] and
+      exists(ParamsInstance i |
+        i.asExpr().getExpr() = [result.getReceiver(), result.getArgument(0)]
+      )
+    }
+
+    override predicate propagatesFlowExt(string input, string output, boolean preservesValue) {
+      input = ["Argument[self]", "Argument[0]"] and
```
hvitved, line 473: Same comment as above.

ruby/ql/lib/codeql/ruby/frameworks/ActionController.qll
hvitved: OK, just wanted to double check that there was not only supposed to be flow from the argument that is an instance of `ParamsInstance`.

hvitved (approved):
LGTM. I have started a new DCA run; let's see what it says before merging.

hvitved:
> DCA shows a 5% slowdown on `opal/opal` but I don't know why

I think that may just be noise. So go ahead and merge.

PR #10539, opened 2022-09-22
Python: add subscript to API graphs

yoff:
This is currently a draft that should allow us to discuss the following questions:

 - does the resulting user code look as nice as we hoped?
 - do we need the ability to refer to object and index? (I believe an `API::SubscriptNode` similar to `API::CallNode` could solve this.)
 - does this design interfere with plans to harmonise implementations across languages? (Currently no)
 - should the access path be `getItem` rather than `getSubscript`? Especially if we add the option to to say `getItem("cookie")`.

### DIFF
--- a/python/ql/lib/semmle/python/ApiGraphs.qll
+++ b/python/ql/lib/semmle/python/ApiGraphs.qll
@@ -570,8 +570,7 @@ module API {
      * API graph node for the prefix `foo`), in accordance with the usual semantics of Python.
      */
 
-    private import semmle.python.internal.Awaited
-
+    // private import semmle.python.internal.Awaited
     cached
     newtype TApiNode =
       /** The root of the API graph. */
@@ -762,11 +761,8 @@ module API {
         )
         or
         // awaiting
-        exists(DataFlow::Node awaitedValue |
-          lbl = Label::await() and
-          ref = awaited(awaitedValue) and
-          pred.flowsTo(awaitedValue)
-        )
+        lbl = Label::await() and
+        ref = pred.getAnAwaited()
       )
       or
       exists(DataFlow::Node def, PY::CallableExpr fn |
--- a/python/ql/lib/semmle/python/ApiGraphs.qll
+++ b/python/ql/lib/semmle/python/ApiGraphs.qll
@@ -1069,7 +1069,7 @@ module API {
 
       /** A label that gets the subscript of a sequence/mapping. */
       class LabelSubscript extends ApiLabel, MkLabelSubscript {
-        override string toString() { result = "getSubscript()" }
+        override string toString() { result = "getASubscript()" }
       }
 
       /** A label for entry points. */
--- a/python/ql/lib/semmle/python/ApiGraphs.qll
+++ b/python/ql/lib/semmle/python/ApiGraphs.qll
@@ -746,6 +746,10 @@ module API {
         lbl = Label::return() and
         ref = pred.getACall()
         or
+        // Awaiting a node that is a use of `base`
+        lbl = Label::await() and
+        ref = pred.getAnAwaited()
+        or
         // Subclassing a node
         lbl = Label::subclass() and
         exists(PY::ClassExpr clsExpr, DataFlow::Node superclass | pred.flowsTo(superclass) |
@@ -759,10 +763,6 @@ module API {
             ref.(DataFlow::ExprNode).getNode().getNode() = clsExpr.getADecoratorCall()
           )
         )
-        or
-        // awaiting
-        lbl = Label::await() and
-        ref = pred.getAnAwaited()
       )
       or
       exists(DataFlow::Node def, PY::CallableExpr fn |
--- a/python/ql/lib/semmle/python/ApiGraphs.qll
+++ b/python/ql/lib/semmle/python/ApiGraphs.qll
@@ -245,6 +245,7 @@ module API {
 
     /**
      * Gets a node representing a subscript of this node.
+     * For example `obj[x]` is a subscript of `obj`.
      */
     Node getASubscript() { result = this.getASuccessor(Label::subscript()) }
 
@@ -575,7 +576,6 @@ module API {
      * API graph node for the prefix `foo`), in accordance with the usual semantics of Python.
      */
 
-    // private import semmle.python.internal.Awaited
     cached
     newtype TApiNode =
       /** The root of the API graph. */
@@ -1067,7 +1067,7 @@ module API {
         override string toString() { result = "getAwaited()" }
       }
 
-      /** A label that gets the subscript of a sequence. */
+      /** A label that gets the subscript of a sequence/mapping. */
       class LabelSubscript extends ApiLabel, MkLabelSubscript {
         override string toString() { result = "getSubscript()" }
       }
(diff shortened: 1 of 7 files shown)


python/ql/src/experimental/semmle/python/frameworks/NoSQL.qll
```diff
@@ -141,11 +110,10 @@ private module NoSql {
    * mongo.db.user.find({'name': safe_search})
    * ```
    *
-   * `mongo.db.user.find` would be a collection method, and so the result.
+   * `mongo.db.user.find` would be a collection method.
    */
-  private DataFlow::Node mongoCollectionMethod() {
-    mongoCollection() = result.(DataFlow::AttrRead).getObject() and
-    result.(DataFlow::AttrRead).getAttributeName() instanceof MongoCollectionMethodNames
+  private API::Node mongoCollectionMethod() {
```
RasmusWL, line 115: we don't seem to need this predicate any more. Can you inline it to `MongoCollectionCall`? :blush:

python/ql/lib/semmle/python/frameworks/Flask.qll
```diff
@@ -408,21 +408,21 @@ module Flask {
...
-      exists(API::Node files | files = request().getMember("files") |
-        this.asCfgNode().(SubscriptNode).getObject() =
-          files.getAValueReachableFromSource().asCfgNode()
-        or
-        this = files.getMember("get").getACall()
-        or
-        this.asCfgNode().(SubscriptNode).getObject() =
-          files.getMember("getlist").getReturn().getAValueReachableFromSource().asCfgNode()
-      )
+      this = requestFileStorage().getAValueReachableFromSource()
```
RasmusWL, line 425: This seems wrong. For something like

```py
x = request.files['foo']
func(x)
```

we only want ONE `FlaskRequestFileStorageInstances`, but using `getAValueReachableFromSource` would make _both_ the RHS of the assignment, and the use of `x` a `FlaskRequestFileStorageInstances` right?

python/ql/lib/semmle/python/ApiGraphs.qll
```diff
@@ -243,6 +243,11 @@ module API {
      */
     Node getAwaited() { result = this.getASuccessor(Label::await()) }
 
+    /**
+     * Gets a node representing a subscript of this node.
```
RasmusWL, line 247: I think an example would be helpful. For `obj[x]`, it's not 100% clear whether you will get just the index `x` or the whole expression.

```suggestion
     * Gets a node representing a subscript of this node.
     * For example `obj[x]` is a subscript of `obj`.
```

python/ql/lib/semmle/python/ApiGraphs.qll
```diff
@@ -570,8 +575,7 @@ module API {
      * API graph node for the prefix `foo`), in accordance with the usual semantics of Python.
      */
 
-    private import semmle.python.internal.Awaited
-
+    // private import semmle.python.internal.Awaited
```
RasmusWL, line 578:
```suggestion
```

python/ql/lib/semmle/python/dataflow/new/internal/LocalSources.qll
```diff
@@ -225,4 +236,26 @@ private module Cached {
...
+  cached
+  predicate await(LocalSourceNode node, Node awaited) {
+    exists(Node awaitedValue |
+      node.flowsTo(awaitedValue) and
+      awaited = awaited(awaitedValue)
+    )
+  }
+
+  /**
+   * Holds if `node` flows to a sequence of which `subscript` is a subscript.
```
RasmusWL, line 252: NIT: "Sequence" is a slightly wrong term here. If we look at the [abstract base classes of Python](https://docs.python.org/3/library/collections.abc.html), a `dict` will not be considered a sequence, but can certainly be subscripted.

According to [the docs of `__getitem__`](https://docs.python.org/3/reference/datamodel.html#object.__getitem__), it seems like saying sequence/mapping will be good enough.

```suggestion
   * Holds if `node` flows to a sequence/mapping of which `subscript` is a subscript.
```

python/ql/lib/semmle/python/ApiGraphs.qll
```diff
@@ -1061,6 +1067,11 @@ module API {
         override string toString() { result = "getAwaited()" }
       }
 
+      /** A label that gets the subscript of a sequence. */
```
RasmusWL, line 1070:
```suggestion
      /** A label that gets the subscript of a sequence/mapping. */
```

RasmusWL (changes requested):
Some minor things, but overall it looks good :+1: 

---

> * does the resulting user code look as nice as we hoped?

Yes :muscle: 

> * do we need the ability to refer to object and index? (I believe an `API::SubscriptNode` similar to `API::CallNode` could solve this.)

:shrug: Agreed that we could do something similar to `API::CallNode` if we need it.

> * does this design interfere with plans to harmonise implementations across languages? (Currently no)

Did you clear this with `@asgerf`?

> * should the access path be `getItem` rather than `getSubscript`? Especially if we add the option to to say `getItem("cookie")`.

Can you give a python code example of what `getItem("cookie")` would find? I don't understand.

yoff:
> > * does the resulting user code look as nice as we hoped?
> 
> Yes 💪

Excellent, then I think this PR has merit as it is now :-)

> > * do we need the ability to refer to object and index? (I believe an `API::SubscriptNode` similar to `API::CallNode` could solve this.)
> 
> 🤷 Agreed that we could do something similar to `API::CallNode` if we need it.

Ok, I will not extend this PR then, but wait for a use case.

> > * does this design interfere with plans to harmonise implementations across languages? (Currently no)
> 
> Did you clear this with `@asgerf`?

Yes :-)

> Can you give a python code example of what `getItem("cookie")` would find? I don't understand.

I was thinking of something like (from `aiohttp/response_test.py`)
```python
resp.headers["Set-Cookie"] = "key2=value2"
```
which should be findable by
```codeql
response().getMember("headers").getItem("Set-Cookie")
```

Also, ideally, "key2=value2" would be findable as
```codeql
response().getMember("headers").getItem("Set-Cookie").getAValueReachingSink
```
but that requires me to fix up def-nodes which I plan to do in a separate PR.

python/ql/lib/semmle/python/frameworks/Flask.qll
yoff: The equality means that they _are_ both instances. However, seeing as this is extending `InstanceSource` and will likely be the starting point of type tracking, it is better to use `asSource` here.

python/ql/src/experimental/semmle/python/frameworks/NoSQL.qll
yoff: Done :-)

tausbn (changes requested):
Overall I think this looks good. I'm a bit ambivalent about introducing `getItem` (specifically the "item" part of the name), though I guess that _would_ be a more language-neutral term than "subscript" (much like how we use "member" instead of "attribute", etc.).

python/ql/lib/semmle/python/frameworks/Flask.qll
RasmusWL: exactly my point :blush:

RasmusWL (dismissed):
LGTM

tausbn (approved):
I'm happy with the PR as it is right now. It already improves the modelling quite a bit. 👍

PR #10548, opened 2022-09-23
Ruby: Add call graph tests for unsupported constructs

hvitved:
This PR documents some (corner) cases that our call graph construction currently doesn't handle.

### DIFF
--- a/ruby/ql/test/library-tests/modules/calls.rb
+++ b/ruby/ql/test/library-tests/modules/calls.rb
@@ -410,3 +410,76 @@ end
 
 SingletonOverride2.call_singleton1
 SingletonOverride2.call_singleton2
+
+class ConditionalInstanceMethods
+    if rand() > 0 then
+        def m1
+            puts "ConditionalInstanceMethods#m1"
+        end
+    end
+
+    def m2
+        puts "ConditionalInstanceMethods#m2"
+        
+        def m3
+            puts "ConditionalInstanceMethods#m3"
+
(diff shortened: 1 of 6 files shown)


ruby/ql/test/library-tests/modules/calls.rb
```diff
@@ -410,3 +410,76 @@ def self.singleton2
...
+EsotericInstanceMethods.new.baz_0 # currently unable to resolve
+EsotericInstanceMethods.new.baz_1 # currently unable to resolve
+EsotericInstanceMethods.new.baz_2 # currently unable to resolve
+
+module ExtendSingletonMethod
+    def singleton
+        puts "ExtendSingletonMethod#singleton"
+    end
+
+    extend self
```
aibaars, line 482: `extend` is also commonly used with module names instead of `self`, for example `extend My::Module`. A bit more rare is `M.extend Other`.

I've also see cases like
```
class Bundler::Thor
  module Invocation
    def self.included(base) #:nodoc:
      base.extend ClassMethods
    end

    module ClassMethods
    ...
   end
end
```

The `extend` method can also be used to add singleton methods to non-module objects. I doubt that is used much in practice though.
```
"hello".extend FancyGreetings
```

hvitved: Will you perhaps add tests for that follow-up?

aibaars (approved):

PR #10550, opened 2022-09-23
C++: New Query `cpp/comma-before-misleading-indentation`

d10c:
- C++: Initial commit of `cpp/comma-before-missing-indentation`
- C++: Initial `cpp/comma-before-misleading-indentation`


aibaars:
@d10c Could you test this query against `git/git` to see if there are any false positives. They use the comma operator quite a bit, but mostly in `for` loop headers.

MathiasVP:
> @d10c Could you test this query against `git/git` to see if there are any false positives. They use the comma operator quite a bit, but mostly in `for` loop headers.

`git/git` is part of the C/C++ DCA suite, so I definitely expect that we'll investigate results on that project before we merge this PR.

cpp/ql/src/Best Practices/Likely Errors/CommaBeforeMisleadingIndentation.ql
```diff
@@ -0,0 +1,19 @@
...
+ * @tags maintainability
+ *       readability
+ */
+
+import cpp
+
+from CommaExpr ce
+where
+  ce.fromSource() and
+  not exists(MacroInvocation me | ce = me.getAnAffectedElement()) and
```
MathiasVP, line 16: It's indeed a good idea to filter out these results! We actually have a [library predicate](https://github.com/github/codeql/blob/main/cpp/ql/lib/semmle/code/cpp/commons/Exclusions.qll#L126) that handles this for you. I think we should make use of that here.

cpp/ql/test/query-tests/Best Practices/Likely Errors/CommaBeforeMisleadingIndentation/test.cpp
```diff
@@ -0,0 +1,19 @@
...
+		(void)i,	// BAD
+	(void)j;
+
+	foo((i++, j++)); // GOOD
+	foo((i++,        // GOOD
+	     j++));
+	foo((i++
+	   , j++));      // GOOD
+	foo((i++,
+	j++));           // BAD (?)
```
dscho, line 18: Maybe throw in something like this, for good measure?

```c
for (int i = 0,
     j = 1; i < 10; i += 2,
    j++)
        (void)i;
```

cpp/ql/test/query-tests/Best Practices/Likely Errors/CommaBeforeMisleadingIndentation/test.cpp
```diff
@@ -9,11 +25,123 @@ void test(int i, int j, int (*foo)(int))
...
+    // Mixed tabs and spaces (ugly case):
+
+    for (i = 0,         // GOOD if tab >= 4 spaces else BAD
+		 j = 0;
+		 i + j < 10;
+         i++,           // GOOD if tab >= 4 spaces else BAD
+		 j++);
+
+	if (i)
+	    (void)i,	    // GOOD if tab >= 4 spaces else BAD
```
geoffw0, line 98: Great tests / real world testing!

With a query like this I would be happy to simply not report dubious cases such as these three, and in general I wouldn't worry about missed results of any kind too much.

d10c: Yes, but how would you distinguish between this case on line 97 and the one on [line 24](https://github.com/github/codeql/pull/10550/files/0e9b77e7c314509beea26479cb10b47a4044d5e0..cacf78838c61bb855e7bf212a77606552b2fe7a7#diff-5c2f6639161ed558bb0ba560dc0224ec85b97a24b1b17ed1f31af90b9d81c6e4R24-R26)? Without access to source code text, we cannot filter out the "dubious" cases (mixed tabs and spaces). So we have to alert for both cases, or neither.

cpp/ql/src/Best Practices/Likely Errors/CommaBeforeMisleadingIndentation.ql
```diff
@@ -0,0 +1,33 @@
...
+ * @id cpp/comma-before-misleading-indentation
+ * @problem.severity recommendation
+ * @tags maintainability
+ *       readability
+ */
+
+import cpp
+import semmle.code.cpp.commons.Exclusions
+
+Expr normalizeExpr(Expr e) {
```
MathiasVP, line 14: Can you add a QLDoc to this predicate?

cpp/ql/test/query-tests/Best Practices/Likely Errors/CommaBeforeMisleadingIndentation/test.cpp
```diff
@@ -77,6 +77,10 @@ int test(int i, int j, int (*foo)(int), int (*bar)(int, int))
 	                   j++);
 	(void)sizeof(x.foo(i++), // GOOD
 	             j++);
+	using U = decltype(x.foo(i++), // GOOD? Unlikely to be misread
+				j++);
+	(void)sizeof(x.foo(i++), // GOOD? Unlikely to be misread
+				j++);
```
MathiasVP, line 85: Ouch. I feel bad for whoever ends up reading code like this. This seems like an obvious case of https://wiki.sei.cmu.edu/confluence/display/cplusplus/EXP52-CPP.+Do+not+rely+on+side+effects+in+unevaluated+operands.

cpp/ql/src/Best Practices/Likely Errors/CommaBeforeMisleadingIndentation.ql
```diff
@@ -26,6 +27,12 @@ predicate isInLoopHead(CommaExpr ce) {
   ce.getEnclosingStmt() = any(ForStmt f).getInitialization()
 }
 
+predicate isInDecltypeOrSizeof(CommaExpr ce) {
```
MathiasVP, line 33: Could this predicate just be defined as `ce.isUnevaluated()`? That certainly includes the two cases you've given here, along with operands of `noexcept`, `alignof` and `typeid` (which are all surrounded by parentheses).

cpp/ql/src/Best Practices/Likely Errors/CommaBeforeMisleadingIndentation.ql
MathiasVP: Ah! The QLDoc makes it a lot more clear what's going on. Thank you!

Could this predicate be written more declaratively as:
```ql
Expr normalizeExpr(Expr e) {
  result =
    min(Expr child, int column |
      child.getParentWithConversions*() = e and column = child.getLocation().getStartColumn()
    |
      child order by column
    )
}
```
?

cpp/ql/test/query-tests/Best Practices/Likely Errors/CommaBeforeMisleadingIndentation/test.cpp
d10c: Ouch indeed. There is a certain popular header-only JSON library that does [something like that](https://github.com/syoyo/tinygltf/blob/9bb5806df4055ac973b970ba5b3e27ce27d98148/json.hpp#L4188-L4188). Kept popping up in the MRVA top 1000.

cpp/ql/src/Best Practices/Likely Errors/CommaBeforeMisleadingIndentation.ql
MathiasVP: cc @d10c on ^.

cpp/ql/src/Best Practices/Likely Errors/CommaBeforeMisleadingIndentation.ql
```diff
@@ -0,0 +1,66 @@
...
+ *       readability
+ */
+
+import cpp
+import semmle.code.cpp.commons.Exclusions
+
+/**
+ * Holds if this is an implicit `this`.
+ *
+ * ThisExpr.isCompilerGenerated() is currently not being extracted, so use a heuristic.
```
MathiasVP, line 18: We should create an internal issue for this. This is most likely a simple oversight.

cpp/ql/src/Best Practices/Likely Errors/CommaBeforeMisleadingIndentation.ql
```diff
@@ -0,0 +1,66 @@
...
+ * @precision medium
+ * @tags maintainability
+ *       readability
+ */
+
+import cpp
+import semmle.code.cpp.commons.Exclusions
+
+/**
+ * Holds if this is an implicit `this`.
```
MathiasVP, line 16: I'd write this if this was a member-predicate, but since this is a nonmember-predicate I'd call out the parameter explicitly.
```suggestion
 * Holds if `te` is an implicit `this`.
```

cpp/ql/src/Best Practices/Likely Errors/CommaBeforeMisleadingIndentation.ql
```diff
@@ -0,0 +1,66 @@
...
+      then result = normalizeExpr(e.getFullyConverted())
+      else result = e
+}
+
+predicate isParenthesized(CommaExpr ce) {
+  ce.getParent*().(Expr).isParenthesised()
+  or
+  ce.isUnevaluated() // sizeof(), decltype(), alignof(), noexcept(), typeid()
+  or
+  ce.getParent*() = any(IfStmt i).getCondition()
```
MathiasVP, line 48: Should this predicate also include a `SwitchStmt`'s expression?

cpp/ql/src/Best Practices/Likely Errors/CommaBeforeMisleadingIndentation.qhelp
```diff
@@ -0,0 +1,31 @@
...
+<qhelp>
+
+<overview>
+<p>
+If the expression to the right of a comma operator starts at an earlier column than the expression to the left, then
+this suspicious indentation possibly indicates a logic error, caused by a typo that may escape visual inspection.
+</p>
+<warning>
+This query has medium precision because CodeQL currently does not distinguish between tabs and spaces in whitespace.
+Alerts may therefore flag code that appears readable for one value of tab size but not another.
```
MathiasVP, line 13: This is very useful information for anyone looking up why they're getting an alert on this new query (and so it should definitely be in here). But I don't think we have any precedent for how to phrase this in a qhelp file, so I hope the Docs team will be able to advise on this.

MathiasVP (commented):
This looks super nice! I really appreciate the amount of testing that has gone into this PR. We should be sure to start a DCA run for this before we merge it (and we might as well do that now while we're waiting for a Docs review).

cpp/ql/src/Best Practices/Likely Errors/CommaBeforeMisleadingIndentation.ql
d10c: Mmm, yes, probably.

cpp/ql/src/Best Practices/Likely Errors/CommaBeforeMisleadingIndentation.ql
d10c: Trying it out, looks promising.

kornbik (commented):
Finish your review

MathiasVP (dismissed):
The code LGTM! Let's merge this once the Docs review is in 🎉.

cpp/ql/src/Best Practices/Likely Errors/CommaBeforeMisleadingIndentation.qhelp
```diff
@@ -0,0 +1,31 @@
+<!DOCTYPE qhelp PUBLIC
+  "-//Semmle//qhelp//EN"
+  "qhelp.dtd">
+<qhelp>
+
+<overview>
+<p>
+If the expression to the right of a comma operator starts at an earlier column than the expression to the left, then
+this suspicious indentation possibly indicates a logic error, caused by a typo that may escape visual inspection.
```
felicitymay, line 9: I'm not entirely sure what we're trying to say here. Is this only a problem when we have a comma and bad indentation together? Or, is a comma always a problem, but we're only flagging code where the indentation suggests that it is the comma that's wrong, rather than the indentation?

cpp/ql/src/Best Practices/Likely Errors/CommaBeforeMisleadingIndentation.qhelp
felicitymay: A search of the repository revealed that `<warning>` is mostly used in older code, but that we are still using it. Judging by: https://codeql.github.com/codeql-query-help/cpp/cpp-missing-check-scanf/, we don't currently have any styling set up for this warning, so we might want to open an issue to add styling for warnings.

Suggest maybe:

```suggestion
This query has medium precision because CodeQL currently does not distinguish between tabs and spaces in whitespace.
If a file contains mixed tabs and spaces, alerts may highlight code that is correctly indented for one value of tab size but not for other tab sizes.
```

cpp/ql/src/Best Practices/Likely Errors/CommaBeforeMisleadingIndentation.qhelp
```diff
@@ -0,0 +1,31 @@
...
+Use standard indentation around the comma operator: begin the right-hand-side operand at the same level of
+indentation (column number) as the left-hand-side operand.
+</p>
+<p>
+When it comes to whitespace, either do not mix tabs and spaces, or mix them consistently.
+</p>
+</recommendation>
+
+<example>
+<sample src="CommaBeforeMisleadingIndentation.cpp" />
```
felicitymay, line 28: We usually expect there to be some introduction or explanation of the example. One possibility is to convert the example into three different example files. Alternatively, something like this might work:

```suggestion
<example>
<p>
This example shows three different ways of writing the same code. The first example contains a comma instead of a semicolon which means that the final line is part of the <code>if</code> statement, even though the indentation suggests that it is intended to be separate. The second example looks different put is functionally the same as the first example. It is more likely that the developer intended to write the third example.
</p>
<sample src="CommaBeforeMisleadingIndentation.cpp" />
```

cpp/ql/src/Best Practices/Likely Errors/CommaBeforeMisleadingIndentation.qhelp
```diff
@@ -0,0 +1,31 @@
...
+<warning>
+This query has medium precision because CodeQL currently does not distinguish between tabs and spaces in whitespace.
+Alerts may therefore flag code that appears readable for one value of tab size but not another.
+</warning>
+</overview>
+
+<recommendation>
+<p>
+Use standard indentation around the comma operator: begin the right-hand-side operand at the same level of
+indentation (column number) as the left-hand-side operand.
```
felicitymay, line 20: Would this work?

```suggestion
To ensure that your code is easy to read and review, use standard indentation around the comma operator. Always begin the right-hand-side operand at the same level of
indentation (column number) as the left-hand-side operand. This makes it easier for other developers to see the intended behavior of your code.
```

cpp/ql/src/Best Practices/Likely Errors/CommaBeforeMisleadingIndentation.qhelp
```diff
@@ -0,0 +1,31 @@
...
+</p>
+<p>
+When it comes to whitespace, either do not mix tabs and spaces, or mix them consistently.
```
felicitymay, line 23: My understanding is that most developers prefer to avoid mixed tabs and spaces. Should this be more like:

```suggestion
Use whitespace consistently to communicate your coding intentions. Where possible, avoid mixing tabs and spaces within a file. If you need to mix them, use them consistently.
```

cpp/ql/src/Best Practices/Likely Errors/CommaBeforeMisleadingIndentation.qhelp
```diff
@@ -0,0 +1,31 @@
...
+</p>
+</recommendation>
+
+<example>
+<sample src="CommaBeforeMisleadingIndentation.cpp" />
+</example>
+
```
felicitymay, line 30: Are there any articles or guidelines that we could refer to? I had a quick look but couldn't find anything.

cpp/ql/src/Best Practices/Likely Errors/CommaBeforeMisleadingIndentation.ql
```diff
@@ -0,0 +1,53 @@
+/**
+ * @name Comma before misleading indentation
+ * @description The expressions before and after the comma operator can be misread because of an unusual difference in start columns.
```
felicitymay, line 3: We usually aim to structure the description as "syntax X causes problem Y". It's more difficult with non-security queries. Would this work?

```suggestion
 * @description If expressions before and after a comma operator use different indentation, it is easy to misread the purpose of the code.
```

cpp/ql/src/Best Practices/Likely Errors/CommaBeforeMisleadingIndentation.ql
```diff
@@ -0,0 +1,53 @@
...
+  ce.fromSource() and
+  not isFromMacroDefinition(ce) and
+  left = normalizeExpr(ce.getLeftOperand()) and
+  right = normalizeExpr(ce.getRightOperand()) and
+  leftLoc = left.getLocation() and
+  rightLoc = right.getLocation() and
+  not isParenthesized(ce) and
+  leftLoc.getEndLine() < rightLoc.getStartLine() and
+  leftLoc.getStartColumn() > rightLoc.getStartColumn()
+select right, "The indentation level may be misleading (for some tab sizes)."
```
felicitymay, line 53: Was there any discussion over putting "for some tab sizes)" in parentheses? I'm not sure that they are needed.

cpp/ql/src/Best Practices/Likely Errors/CommaBeforeMisleadingIndentation.cpp
```diff
@@ -0,0 +1,32 @@
+/*
+ * Here, the comma should have been a semicolon:
```
felicitymay, line 2: Suggest:

```suggestion
 * In this example, the developer intended to use a semicolon but accidentally used a comma:
```

cpp/ql/src/Best Practices/Likely Errors/CommaBeforeMisleadingIndentation.cpp
```diff
@@ -0,0 +1,32 @@
...
+
+enum privileges entitlements = NONE;
+
+if (is_admin)
+    entitlements = FULL, // BAD
+
+restrict_privileges(entitlements);
+
+/*
+ * This is misleading, because the code is unexpectedly equivalent to:
```
felicitymay, line 13: Maybe:

```suggestion
 * The use of a comma means that the first example is equivalent to this second example:
```

cpp/ql/src/Best Practices/Likely Errors/CommaBeforeMisleadingIndentation.cpp
```diff
@@ -0,0 +1,32 @@
...
+
+enum privileges entitlements = NONE;
+
+if (is_admin) {
+    entitlements = FULL;
+    restrict_privileges(entitlements);
+}
+
+/*
+ * Whereas the following code was probably intended:
```
felicitymay, line 24:
```suggestion
 * The indentation of the first example suggests that the developer probably intended the following code:
```

cpp/ql/src/Best Practices/Likely Errors/CommaBeforeMisleadingIndentation.qhelp
geoffw0: The issue is about bad indentation following a comma, so it requires both of those elements.  Perhaps the confusing part here is actually the use of "left" and "right" to describe relative positions when at least one of these expressions occurs on a different line (with indentation).
```suggestion
If the expression after the comma operator starts at an earlier column than the expression before the comma, then
this suspicious indentation possibly indicates a logic error, caused by a typo that may escape visual inspection.
```

cpp/ql/src/Best Practices/Likely Errors/CommaBeforeMisleadingIndentation.qhelp
geoffw0: `put` -> `but` in your suggestion :)

cpp/ql/src/Best Practices/Likely Errors/CommaBeforeMisleadingIndentation.ql
d10c: Removed the parentheses in a later commit.

d10c:
Thanks for the Doc review @felicitymay! I've made changes based on your suggestions. This PR is now ready for a final review/approval.

cpp/ql/src/Best Practices/Likely Errors/CommaBeforeMisleadingIndentation.qhelp
felicitymay: Thanks Geoffrey 🤦🏻‍♀️

cpp/ql/src/Best Practices/Likely Errors/CommaBeforeMisleadingIndentation.qhelp
d10c: Not very targeted, but I added references to Wikipedia.

PR #10558, opened 2022-09-23
Java: Improve performance of StaticInitializationVector.

aschackmull:
As the comment suggests, it looks like `ArrayUpdateConfig` was meant to be value flow rather than taint flow, and this makes a big performance difference. Also, since we don't car about specific sinks, we might as well add an out-barrier on sinks for a little bit of added performance.


atorralba (approved):
I like the trick about the out-barrier, we probably can apply this pattern to other configurations that similarly don't care about the sink but only that it exists at least one.

PR #10559, opened 2022-09-23
Ruby: some improvements

aibaars:
I investigated a case where we didn't flag up a [result](https://github.com/rubygems/bundler/blob/674966e21cc7257dd52797ef56ee9628c3ce1bc6/lib/bundler.rb#L187) for the rb/overly-permissive-file query .
This pull request implements the improvements needed to flag up that result, and makes the following changes:

* track Pathname instances using global flow instead of local flow
* model Object.dup and Kernel.tap
* add flow steps for logical operators
* add support for extend in the call graph

The dataflow improvements caused a fairly large number of false positives on our test suite for the missing regex anchor and incomplete hostname regex queries. This was caused by misclassifying some string literals as regular expression due to confusing `Regexp#match` and `String#match`. Commit f93bc900b90c264c935b25d5eb669aa9e353b3c4 improves this and remove the false positive results.


ruby/ql/lib/codeql/ruby/dataflow/internal/DataFlowPrivate.qll
```diff
@@ -181,6 +181,12 @@ module LocalFlow {
         ) and
         nodeFrom.asExpr() = for.getValue()
       )
+    or
+    nodeTo.asExpr() =
```
hvitved, line 185: This is already included in taint tracking, and I don't think it is safe to include in data flow, as the value may not be preserved.

aibaars: Why wouldn't it be value preserving?

aibaars: As far as I know:
* `a || b`  means `if (tmp = a) then tmp else b`
* `a && b`  means `if (tmp = a) then b else tmp`

hvitved: You are right!

aibaars: I guess I should remove the equivalent code from the taint steps.

ruby/ql/lib/codeql/ruby/dataflow/internal/DataFlowDispatch.qll
```diff
@@ -674,6 +721,11 @@ private predicate singletonMethodOnModule(MethodBase method, string name, Module
   )
   or
   flowsToSingletonMethodObject(trackModuleAccess(m), method, name)
+  or
+  exists(Module other |
+    extendCallModule(m, other) and
+    method = lookupMethod(other, name)
```
hvitved, line 727: Should this be limited to public methods?

ruby/ql/lib/codeql/ruby/dataflow/internal/DataFlowDispatch.qll
```diff
@@ -708,6 +765,12 @@ predicate singletonMethodOnInstance(MethodBase method, string name, Expr object)
   not selfInModule(object.(SelfVariableReadAccess).getVariable(), _) and
   // cannot use `trackModuleAccess` because of negative recursion
   not exists(resolveConstantReadAccess(object))
+  or
+  exists(DataFlow::ExprNode receiver, Module other |
+    extendCall(receiver, other) and
+    object = receiver.getExprNode().getExpr() and
+    method = lookupMethod(other, name)
```
hvitved, line 772: Should this be limited to public methods?

ruby/ql/lib/codeql/ruby/dataflow/internal/DataFlowDispatch.qll
aibaars: Good point. `private` can be called, but only if the receiver is `self` and `protected` methods can be called if the receiver is a subclass of the enclosing class or something.

ruby/ql/lib/codeql/ruby/dataflow/internal/DataFlowPrivate.qll
```diff
@@ -181,6 +181,12 @@ module LocalFlow {
...
+      any(CfgNodes::ExprNodes::BinaryOperationCfgNode op |
+        op.getExpr() instanceof BinaryLogicalOperation and
+        nodeFrom.asExpr() = op.getAPredecessor()
```
aibaars, line 188:
```suggestion
        nodeFrom.asExpr() = op.getAnOperand()
```

ruby/ql/lib/codeql/ruby/dataflow/internal/DataFlowDispatch.qll
```diff
@@ -294,6 +294,30 @@ private DataFlowCallable viableLibraryCallable(DataFlowCall call) {
   )
 }
 
+/** Holds if there is a call like `receiver.extend(M)`. */
+pragma[nomagic]
+private predicate extendCall(DataFlow::ExprNode receiver, Module m) {
+  exists(DataFlow::CallNode extendCall |
+    extendCall.getMethodName() = "extend" and
+    exists(DataFlow::LocalSourceNode sourceNode | sourceNode.flowsTo(extendCall.getArgument(0)) |
```
aibaars, line 302:
```suggestion
    exists(DataFlow::LocalSourceNode sourceNode | sourceNode.flowsTo(extendCall.getArgument(_)) |
```

ruby/ql/lib/codeql/ruby/dataflow/internal/DataFlowDispatch.qll
aibaars: Same as above.

hvitved (approved):

PR #10560, opened 2022-09-23
Ruby: treat `Psych` and `YAML` as aliases for rb/unsafe-deserialization

alexrford:
In semi-recent versions of ruby the `YAML` module is an alias of `Psych`.
https://ruby-doc.com/stdlib-3.0.1/libdoc/yaml/rdoc/YAML.html

### DIFF
--- a/ruby/ql/lib/change-notes/2022-09-23-yaml-load-file.md
+++ b/ruby/ql/lib/change-notes/2022-09-23-yaml-load-file.md
@@ -1,4 +1,0 @@
----
-category: minorAnalysis
----
-* Added `YAML.load_file` as a potential sink for unsafe deserialization.
--- a/ruby/ql/lib/change-notes/2022-09-23-yaml-load-file.md
+++ b/ruby/ql/lib/change-notes/2022-09-23-yaml-load-file.md
@@ -1,0 +1,4 @@
+---
+category: minorAnalysis
+---
+* Added `YAML.load_file` as a potential sink for unsafe deserialization.
--- a/ruby/ql/lib/codeql/ruby/security/UnsafeDeserializationCustomizations.qll
+++ b/ruby/ql/lib/codeql/ruby/security/UnsafeDeserializationCustomizations.qll
@@ -48,12 +48,15 @@ module UnsafeDeserialization {
   }
 
   /**
-   * An argument in a call to `YAML.load`, considered a sink for unsafe
-   * deserialization.
+   * An argument in a call to `YAML.load` or `YAML.load_file`, considered a sink
+   * for unsafe deserialization. As the `YAML` module is an alias of `Psych` in
    */
   class YamlLoadArgument extends Sink {
     YamlLoadArgument() {
-      this = API::getTopLevelMember("YAML").getAMethodCall("load").getArgument(0)
+      this =
+        API::getTopLevelMember(["YAML", "Psych"])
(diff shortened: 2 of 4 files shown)


ruby/ql/lib/codeql/ruby/security/UnsafeDeserializationCustomizations.qll
```diff
@@ -48,12 +48,15 @@ module UnsafeDeserialization {
   }
 
   /**
-   * An argument in a call to `YAML.load`, considered a sink for unsafe
-   * deserialization.
+   * An argument in a call to `YAML.load` or `YAML.load_file`, considered a sink
+   * for unsafe deserialization. As the `YAML` module is an alias of `Psych` in
```
nickrolfe, line 52: Looks like you forgot to finish the sentence here.

ruby/ql/lib/codeql/ruby/security/UnsafeDeserializationCustomizations.qll
```diff
@@ -48,12 +48,15 @@ module UnsafeDeserialization {
...
    */
   class YamlLoadArgument extends Sink {
     YamlLoadArgument() {
-      this = API::getTopLevelMember("YAML").getAMethodCall("load").getArgument(0)
+      this =
+        API::getTopLevelMember(["YAML", "Psych"])
+            .getAMethodCall(["load", "load_file"])
```
nickrolfe, line 58: I think I'm ok with treating `load_file` as a sink, but it strikes me as far less serious than `load`, since it's only an unsafe-deserialization problem if there's some existing dangerous YAML file on the server. I vaguely recall that that's why I didn't originally include it.

ruby/ql/lib/codeql/ruby/security/UnsafeDeserializationCustomizations.qll
alexrford: Oops, thanks

ruby/ql/lib/codeql/ruby/security/UnsafeDeserializationCustomizations.qll
alexrford: Yeah, I agree that this case seems far less severe in most cases. I had a quick look at the JS and Python versions for precedent. In those cases the `load` like methods seem to be able to take plain data or some file stream. However, I'm not sure if the argument would be consider to be tainted if the file input stream path was user-controlled?

alexrford: As discussed in the weekly meeting, I've removed the `load_file` arg as a sink and just kept the `Psych` change.

ruby/ql/lib/change-notes/2022-09-23-yaml-load-file.md
```diff
@@ -0,0 +1,4 @@
+---
+category: minorAnalysis
+---
+* Added `YAML.load_file` as a potential sink for unsafe deserialization.
```
nickrolfe, line 4: This needs to be changed now.

alexrford: Thanks, this slipped my mind. I've dropped the change note entirely as I don't think this change warrants one.

nickrolfe (approved):
👍

PR #10564, opened 2022-09-23
Swift: fix regression.

geoffw0:
Fix the recent regression in `swift/string-length-conflation`.  It appears to have been caused by the 2nd parameter of `index(_:offsetBy:)` being renamed from `n` to `distance`.  *Hopefully* this kind of thing won't be too common, or we might have to change our strategy a little.


MathiasVP (approved):
What an odd change, indeed!

PR #10572, opened 2022-09-26
C++: Add FP for `CWE-193`

MathiasVP:
This PR adds testcase with an FP that appears quite common on some codebases for the new `cpp/invalid-pointer-deref` query.


geoffw0 (approved):
Good to have a test for this.

PR #10576, opened 2022-09-26
SSA: Turn consistency predicates into `query` predicates

hvitved:
Possible now that class signatures are supported.


aibaars (approved):
Nice! Do we need to update the version of codeql/ssa in the qlpack.yml files. At the very least we'd need to update the lock files, don't we?

aibaars:
> Nice! Do we need to update the version of codeql/ssa in the qlpack.yml files. At the very least we'd need to update the lock files, don't we?

Although, perhaps we don't; see also https://github.com/github/codeql/pull/10668#pullrequestreview-1162986575 .

michaelnebel (approved):
Looks good to me!

hvitved:
> Nice! Do we need to update the version of codeql/ssa in the qlpack.yml files. At the very least we'd need to update the lock files, don't we?

All versions bumps should be part of the release process, AFAIK.

PR #10580, opened 2022-09-27
Java: remove `stubs/android` directory

jcogs33:
**Summary:**
This PR removes the old `stubs/android` directory and replaces anywhere it is used with `stubs/google-android-9.0.0` instead.

**Description:**
- Searched for usages of `stubs/android` and found that it was only used in two `options` files. Updated these files to use `stubs/google-android-9.0.0` instead.
- Ran the tests related to these `options` files to confirm that they still work when using `stubs/google-android-9.0.0`. Since they worked after the switch, there did not appear to be any need to generate additional stubs in `stubs/google-android-9.0.0` for them.
- Deleted the `stubs/android` directory.

**Consideration:**
- I noticed that a `com/android/internal/R.java` file existed in `stubs/android` but not in `stubs/google-android-9.0.0`. I did not move this file to `stubs/google-android-9.0.0` since it does not seem that any tests are dependent on it (based on the Java Language Tests check passing without it). Let me know if this file should be moved to `stubs/google-android-9.0.0` instead of being removed.
- Let me know if there is anything else that needs to be updated/considered when removing this directory.


atorralba (approved):
This is great, thank you! The `R.java` file is auto-generated in real Android applications, so it makes sense we don't need it in our tests.

PR #10593, opened 2022-09-27
C++: Fix FPs on `cpp/invalid-pointer-deref`

MathiasVP:
It's super difficult to properly fix the FP I added in https://github.com/github/codeql/pull/10572. So for now I've decided to just block all flow through back-edges of phi nodes (as suggested by `@aschackmull`) since this one _one_ of the things that had to be present in order to get an FP here.

### DIFF
--- a/cpp/ql/src/experimental/Security/CWE/CWE-193/InvalidPointerDeref.ql
+++ b/cpp/ql/src/experimental/Security/CWE/CWE-193/InvalidPointerDeref.ql
@@ -128,6 +128,10 @@ class AllocToInvalidPointerConf extends ProductFlow::Configuration {
       state2 = delta.toString()
     )
   }
+
+  override predicate isBarrierOut2(DataFlow::Node node) {
+    node = any(DataFlow::SsaPhiNode phi).getAnInput(true)
+  }
 }
 
 pragma[nomagic]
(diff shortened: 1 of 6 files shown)


cpp/ql/src/experimental/Security/CWE/CWE-193/InvalidPointerDeref.ql
```diff
@@ -154,11 +158,11 @@ predicate pointerAddInstructionHasOperands(
  */
 pragma[nomagic]
 predicate pointerAddInstructionHasBounds(
-  PointerAddInstruction pai, DataFlow::Node sink1, Instruction sink2, int delta
+  PointerAddInstruction pai, DataFlow::Node sink1, DataFlow::Node sink2, int delta
```
rdmarsh2, line 161: Why this change? Just a performance thing or is there some semantic difference I'm missing?

MathiasVP: Oh, right. Sorry. That was part of a solution that I ended up discarding. But I kept it in the PR since I liked the symmetry of having both sinks be nodes instead of one being passed as a node and another being passed as an instruction.

So no, there is no behavior change from this at all.

rdmarsh2 (approved):

PR #10595, opened 2022-09-27
Swift: Add `ClassOrStructDecl` class

MathiasVP:
This PR adds a class that's defined as the union of `ClassDecl and `StructDecl`. This is motivated by us having used the wrong one of these on several occasions.

Hopefully, by instead relying on this union class, we'll be less likely to run into these issues.


geoffw0 (approved):
LGTM, thanks for adding this!

PR #10596, opened 2022-09-27
Swift: URL is a struct, not a class

geoffw0:
See: https://developer.apple.com/documentation/foundation/url

Updates the tests and code to match reality.

We hope this kind of error will be less common when we have complete Models-As-Data support, and helper methods for matching class members by class and member name, so that use of the `ClassDecl` and `StructDecl` QL classes directly becomes uncommon.


geoffw0:
I think there are a couple of affected tests I missed, fix to come...

geoffw0:
Fixed the other affected test.

MathiasVP (commented):
LGTM!

MathiasVP (approved):

PR #10597, opened 2022-09-27
Swift: URL taint sources

geoffw0:
Add `URL.resourceBytes`, `URL.lines` and `URL.resourceBytes.lines` as taint sources.

I haven't used Models-As-Data because it doesn't currently support member variables in Swift.  When that changes we should be able to simplify the model itself to three lines of CSV.


MathiasVP (approved):
We should probably convert this to InlineExpectationsTests, but let's do that in a separate PR.

PR #10598, opened 2022-09-27
Ruby: Identify ActionController::Metal controllers

hmac:
Subclasses of `ActionController::Metal` are stripped-down controllers.
We want to recognise them as ActionController controllers.
There are some common ActionController methods that are not available in
Metal, but these are not likely to be used anyway as they would throw an
exception, so I don't think there's much harm in including them in the
modelling.


alexrford (approved):
👍

PR #10599, opened 2022-09-27
Ruby: Model send_file

hmac:
Recognise send_file as a FileSystemAccess

This method is available in ActionController actions, and sends the file
at the given path to the client.


nickrolfe (approved):
LGTM, and adding the sanitisers as a followup (hopefully soon!) makes sense.

PR #10605, opened 2022-09-28
QL: allow getURL as an acronym

erik-krogh:
`getURL` is a language feature, so it should be allowed as an upper-case acronym.

### DIFF
--- a/ql/ql/src/codeql_ql/style/AcronymsShouldBeCamelCaseQuery.qll
+++ b/ql/ql/src/codeql_ql/style/AcronymsShouldBeCamelCaseQuery.qll
@@ -37,6 +37,7 @@ predicate shouldBePascalCased(string name, AstNode node, string kind) {
   not node.hasAnnotation("deprecated") and
   // allowed upper-case acronyms.
   not name.regexpMatch(".*(PEP|AES|DES|EOF).*") and
+  not name = "getURL" and // getURL is a language feature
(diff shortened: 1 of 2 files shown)


nickrolfe (approved):
Makes sense.

kaeluka (approved):
Strictly speaking, that change means that users can write their own `getURL` methods (right?), but this implementation still seems to strike a good balance between precision and complexity.

MathiasVP:
> Strictly speaking, that change means that users can write their own `getURL` methods (right?), but this implementation still seems to strike a good balance between precision and complexity.

People _can_ write their own getURL predicate, yes. See https://codeql.github.com/docs/writing-codeql-queries/providing-locations-in-codeql-queries/#providing-urls.

erik-krogh:
> Strictly speaking, that change means that users can write their own `getURL` methods (right?), but this implementation still seems to strike a good balance between precision and complexity.

Yes, and that was always the case, now they just don't get a warning when they do it.  

[The `getURL` predicate is special at the language level](https://codeql.github.com/docs/writing-codeql-queries/providing-locations-in-codeql-queries/#providing-urls).  
So sometimes it's the right name to use, even if the name doesn't follow our style-guide.  

I got some FPs due to that in [another PR](https://github.com/github/codeql/pull/10592), so that's why I'm doing this PR.

PR #10606, opened 2022-09-28
C#: Add meta query for reporting calls to unsupported library methods

hvitved:


michaelnebel (approved):
Excellent.
Can't wait to try it out!

PR #10608, opened 2022-09-28
Python: subscript def nodes

yoff:
Follow on to #10539. Fix the def-nodes for subscripts.

I also experimented with recovering the index of a subscript (see [the commit](https://github.com/github/codeql/pull/10608/commits/2435ccf06e6f3640d5c091d06b7ad4a1c29e85a2)). I did it for the case where the subscript is being written to. This is the complicated case, because the actual `SubscriptNode` is _not_ an API node:
For
```python
  header["key"] = "value"
```
`header` would be an API use node and `"value"` would be an API def node with path `[path to header].getASubscript()`. (To get the data flow node, we can then use `asSink()` or `getAValueReachingSink()` as appropriate.)

Update: There is now a `SubscriptReadNode` and a `SubscriptWriteNode` to help with recovering the index and the written value.

### DIFF
--- a/python/ql/lib/change-notes/2022-10-04-api-subscript-nodes.md
+++ b/python/ql/lib/change-notes/2022-10-04-api-subscript-nodes.md
@@ -3,4 +3,4 @@ category: minorAnalysis
 ---
 * Fixed labels in the API graph pertaining to definitions of subscripts. Previously, these were labelled by `getMember` rather than `getASubscript`.
 * Added `API::SubscriptReadNode` and `API::SubscriptWriteNode` with convenience predicates to obtain the index and, in the case of a write, the written value.
-* Added convenience predicates to obtain a subscript at a specific index, when the index happens to be a statically known string.
+* Added convenience predicates `getMember("key")` to obtain a subscript at a specific index, when the index happens to be a statically known string.
--- a/python/ql/lib/change-notes/2022-10-04-api-subscript-nodes.md
+++ b/python/ql/lib/change-notes/2022-10-04-api-subscript-nodes.md
@@ -3,4 +3,4 @@ category: minorAnalysis
 ---
 * Fixed labels in the API graph pertaining to definitions of subscripts. Previously, these were labelled by `getMember` rather than `getASubscript`.
 * Added `API::SubscriptReadNode` and `API::SubscriptWriteNode` with convenience predicates to obtain the index and, in the case of a write, the written value.
-* Added convenience predicates `getMember("key")` to obtain a subscript at a specific index, when the index happens to be a statically known string.
+* Added convenience predicates `getSubscript("key")` to obtain a subscript at a specific index, when the index happens to be a statically known string.
--- a/python/ql/lib/change-notes/2022-10-04-api-subscript-nodes.md
+++ b/python/ql/lib/change-notes/2022-10-04-api-subscript-nodes.md
@@ -2,5 +2,6 @@
 category: minorAnalysis
 ---
 * Fixed labels in the API graph pertaining to definitions of subscripts. Previously, these were labelled by `getMember` rather than `getASubscript`.
-* Added `API::SubscriptReadNode` and `API::SubscriptWriteNode` with convenience predicates to obtain the index and, in the case of a write, the written value.
+* Added edges for indices of subscripts to the API graph. Now a subscripted API node will have an edge to the API node for the index expression. So if `foo` is matched by API node `A`, then `"key"` in `foo["key"]` will be matched by the API node `A.getIndex()`, which can be used to track the origin of the index.
+* Added `API::SubscriptReadNode` and `API::SubscriptWriteNode` with convenience predicates to obtain the index and, in the case of a write, the written value. This is useful to tie indices to matching subscript expressions.
 * Added convenience predicates `getSubscript("key")` to obtain a subscript at a specific index, when the index happens to be a statically known string.
--- a/python/ql/lib/change-notes/2022-10-04-api-subscript-nodes.md
+++ b/python/ql/lib/change-notes/2022-10-04-api-subscript-nodes.md
@@ -1,0 +1,6 @@
+---
+category: minorAnalysis
+---
+* Fixed labels in the API graph pertaining to definitions of subscripts. Previously, these were labelled by `getMember` rather than `getASubscript`.
+* Added `API::SubscriptReadNode` and `API::SubscriptWriteNode` with convenience predicates to obtain the index and, in the case of a write, the written value.
+* Added convenience predicates to obtain a subscript at a specific index, when the index happens to be a statically known string.
--- a/python/ql/lib/semmle/python/ApiGraphs.qll
+++ b/python/ql/lib/semmle/python/ApiGraphs.qll
@@ -694,11 +694,11 @@ module API {
           rhs = aw.getValue()
         )
         or
-        // TODO: I had expected `DataFlow::AttrWrite` to contain the attribute writes from a dict, that's how JS works.
+        // for dictionary literals, from `x` to `{ "key": x }`
         exists(PY::Dict dict, PY::KeyValuePair item |
           dict = pred.(DataFlow::ExprNode).getNode().getNode() and
           dict.getItem(_) = item and
-          lbl = Label::member(item.getKey().(PY::StrConst).getS()) and
+          lbl = Label::subscript() and
           rhs.(DataFlow::ExprNode).getNode().getNode() = item.getValue()
         )
         or
@@ -718,6 +718,15 @@ module API {
         use(base, src) and aw = trackUseNode(src).getAnAttributeWrite() and rhs = aw.getValue()
       |
         lbl = Label::memberFromRef(aw)
+      )
+      or
+      // from `x` to a definition of `x[...]`
+      exists(DataFlow::LocalSourceNode src, DataFlow::CfgNode subscript |
+        use(base, src) and
+        subscript = trackUseNode(src).getASubscript() and
+        rhs.asCfgNode() = subscript.asCfgNode().(PY::DefinitionNode).getValue()
+      |
+        lbl = Label::subscript()
       )
       or
       exists(EntryPoint entry |
@@ -757,7 +766,8 @@ module API {
         or
         // Subscripting a node that is a use of `base`
         lbl = Label::subscript() and
-        ref = pred.getASubscript()
+        ref = pred.getASubscript() and
+        ref.asCfgNode().isLoad()
         or
         // Subclassing a node
         lbl = Label::subclass() and
--- a/python/ql/lib/semmle/python/ApiGraphs.qll
+++ b/python/ql/lib/semmle/python/ApiGraphs.qll
@@ -506,6 +506,18 @@ module API {
 
   /**
    * A read of a subscript that is connected to the API graph.
+   *
+   * This can be used to reason about the index of subscripts.
+   * For the code
+   * ```python
+   * x = foo[bar]
+   * y = foo[baz]
+   * ```
+   * there wil be subscript edges from `foo` to both `foo[bar]` and `foo[baz]`,
+   * and there will be index edges from `foo` to both `bar` and `baz`.
+   * There will be a separate `SubscriptReadNode` for `foo[bar]` (and one for `foo[baz]`),
+   * and the member predicates `getObject` and `getIndex` will return nodes
+   * consistent with that particular read.
    */
   class SubscriptReadNode extends DataFlow::Node {
     API::Node subscripted;
@@ -525,7 +537,17 @@ module API {
   /**
    * A write to a subscript that is connected to the API graph.
    *
-   * The member predicate `getValue` is guaranteed to exist and be unique to this write.
+   * This can be used to reason about the index and written values of subscripts.
+   * For the code
+   * ```python
+   * foo[bar] = x
+   * foo[baz] = y
+   * ```
+   * there wil be subscript edges from `foo` to both `x` and `y`,
+   * and there will be index edges from `foo` to both `bar` and `baz`.
+   * There will be a separate `SubscriptWriteNode` for `x` (and one for `y`),
+   * and the member predicates `getObject`, `getIndex`, and `getValue` will return nodes
+   * consistent with that particular write.
    */
   class SubscriptWriteNode extends DataFlow::Node {
     API::Node subscripted;
--- a/python/ql/lib/semmle/python/ApiGraphs.qll
+++ b/python/ql/lib/semmle/python/ApiGraphs.qll
@@ -520,16 +520,16 @@ module API {
    * consistent with that particular read.
    */
   class SubscriptReadNode extends DataFlow::Node {
-    API::Node subscripted;
+    API::Node object;
 
-    SubscriptReadNode() { this = subscripted.getASubscript().asSource() }
+    SubscriptReadNode() { this = object.getASubscript().asSource() }
 
-    /** Gets an API node for the object being subscripted. */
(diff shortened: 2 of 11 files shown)


python/ql/lib/semmle/python/frameworks/Aiohttp.qll
```diff
@@ -621,14 +621,12 @@ module AiohttpWebModel {
...
-        this.asCfgNode() = subscript
-      |
-        subscript.getObject() =
-          aiohttpResponseInstance().getMember("cookies").getAValueReachableFromSource().asCfgNode() and
-        value.asCfgNode() = subscript.(DefinitionNode).getValue() and
+      // `value` defines a subscript
+      value = aiohttpResponseInstance().getMember("cookies").getASubscript().asSink() and
+      // recover the `SubscriptNode` to get the index (and to define `this`)
+      // this part assumes a specific syntax for defining the subscript
+      exists(SubscriptNode subscript | value.asCfgNode() = subscript.(DefinitionNode).getValue() |
```
tausbn, line 628: I don't like having to drop down to the control-flow layer like this. We should add the appropriate classes to the API-graph and data-flow layers.

tausbn (commented):
I'm still mulling over some of the consequences of the changes in this PR, but here at least are some interim comments that should be discussed.

python/ql/lib/semmle/python/ApiGraphs.qll
```diff
@@ -694,11 +694,11 @@ module API {
           rhs = aw.getValue()
         )
         or
-        // TODO: I had expected `DataFlow::AttrWrite` to contain the attribute writes from a dict, that's how JS works.
+        // for dictionary literals, from `x` to `{ "key": x }`
         exists(PY::Dict dict, PY::KeyValuePair item |
           dict = pred.(DataFlow::ExprNode).getNode().getNode() and
           dict.getItem(_) = item and
-          lbl = Label::member(item.getKey().(PY::StrConst).getS()) and
+          lbl = Label::subscript() and
```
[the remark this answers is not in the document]
tausbn: I would say it depends. If we find an API that exposes some components as numerically-indexed bits of some object, then maybe it'll be relevant.

(For that matter, I don't know for certain that _string_ indices are necessary, but it feels like it would be more likely for an API to have been constructed in this way.)

At any rate, we would still have an escape hatch where we assign an API node to the index, thus permitting more fine-grained control over the exact form of the index.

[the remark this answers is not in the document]
yoff: I added a convenience predicate for `getSubscript("bar")`. It is not part of the API graph, so I wrote a separate test to show how it works. Is this the right approach, or did you wan to add labels with strings on them for the known cases?

python/ql/lib/semmle/python/frameworks/Aiohttp.qll
yoff: I added subscript nodes, so now it is hidden in those. I could still also add a `DataFlow::SubscriptNode` to wrap `SubscriptNode` from the CFG. Then the hidden code would also be nice...

python/ql/lib/semmle/python/ApiGraphs.qll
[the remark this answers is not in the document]
yoff: Ah, I did not refresh and see your comment. Maybe you want a new label `getIndex` that points to the index of a subscript?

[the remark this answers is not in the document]
tausbn: I think we should probably have a wider conversation about this, before committing to a specific API. I think adding `getIndex` labels definitely makes sense, though I'm not entirely sure if they source node of such a label should be the `foo` bit of `foo[bar]` or the entire subscript expression. By comparison with how arguments/parameters are modelled, I'm inclined to say it should be the former.

python/ql/lib/semmle/python/ApiGraphs.qll
```diff
@@ -249,6 +249,24 @@ module API {
      */
     Node getASubscript() { result = this.getASuccessor(Label::subscript()) }
 
+    /**
+     * Gets a node representing a subscript of this node at index `key`.
+     * This requires that the index can be statically determined.
```
RasmusWL, line 254:
````suggestion
     * Gets a node representing a subscript of this node at (string) index `key`.
     * This requires that the index can be statically determined.
     * 
     * For example, the subscripts of `a` and `b` below would be found using
     * the index `foo`:
     * ```py
     * a["foo"]
     * x = "foo" if cond else "bar"
     * b[x]
     * ```
````

File: python/ql/lib/semmle/python/ApiGraphs.qll [hunk 4]
```diff
@@ -474,6 +492,49 @@ module API {
...
 
+  /**
+   * A read of a subscript that is connected to the API graph.
+   */
+  class SubscriptReadNode extends DataFlow::Node {
+    API::Node subscripted;
+
+    SubscriptReadNode() { this = subscripted.getASubscript().asSource() }
+
+    /** Gets ans API node for the object being subscripted */
```
RasmusWL, line 503:
```suggestion
    /** Gets an API node for the object being subscripted. */
```

python/ql/lib/semmle/python/ApiGraphs.qll
```diff
@@ -474,6 +492,49 @@ module API {
...
+    API::Node getObject() { result = subscripted }
+
+    /** Gets the data flow node representing the index of this read, if any. */
```
RasmusWL, line 506: I'm surprised by the `if any` part. Will it ever be the case that there is no result for this predicate? :open_mouth: -- that is, a subscript without an index? :open_mouth:

python/ql/lib/semmle/python/ApiGraphs.qll
```diff
@@ -474,6 +492,49 @@ module API {
...
+   * A write to a subscript that is connected to the API graph.
+   *
+   * The member predicate `getValue` is guaranteed to exist and be unique to this write.
+   */
+  class SubscriptWriteNode extends DataFlow::Node {
+    API::Node subscripted;
+
+    SubscriptWriteNode() { this = subscripted.getASubscript().asSink() }
+
+    /** Gets ans API node for the object being subscripted */
```
RasmusWL, line 522:
```suggestion
    /** Gets an API node for the object being subscripted. */
```

python/ql/lib/semmle/python/ApiGraphs.qll
```diff
@@ -474,6 +492,49 @@ module API {
...
+    API::Node getObject() { result = subscripted }
+
+    /** Gets the data flow node representing the index of this write, if any. */
```
RasmusWL, line 525: same comment about the `if any` part

RasmusWL (changes requested):
From principle I'm a little disturbed by the fact that we only add nice support for string based dictionary-like lookups, but I totally agree that this will be the most common use-case **by far**!

python/ql/lib/semmle/python/ApiGraphs.qll
yoff: Hm, I probably mean "if we can find it by assuming a `SubscriptNode` is here" but that is not what is expressed. I agree the current text is confusing. Perhaps "if it can be determined"?

yoff: I have just removed the "if any". At the moment there will always be a result. If we add more ways of reading or writing subscripts, we will probably do it in a `DataFlow::SubscriptNode` and it will become apparent if the index can be determined or not. If not, we can update these comments.

yoff:
> From principle I'm a little disturbed by the fact that we only add nice support for string based dictionary-like lookups, but I totally agree that this will be the most common use-case **by far**!

Note that we have just added convenience methods for this case, we did not design the API graph around it. So we could equally add convenience methods for other situations in the future.

python/ql/lib/change-notes/2022-10-04-api-subscript-nodes.md
```diff
@@ -0,0 +1,6 @@
+---
+category: minorAnalysis
+---
+* Fixed labels in the API graph pertaining to definitions of subscripts. Previously, these were labelled by `getMember` rather than `getASubscript`.
+* Added `API::SubscriptReadNode` and `API::SubscriptWriteNode` with convenience predicates to obtain the index and, in the case of a write, the written value.
+* Added convenience predicates to obtain a subscript at a specific index, when the index happens to be a statically known string.
```
RasmusWL, line 6: NIT: maybe we could mention what these `convenience predicates`, otherwise it'll be a bit of a treasure hunt for people that read this change-note without also viewing the diff :sweat_smile:

RasmusWL (approved):
> > From principle I'm a little disturbed by the fact that we only add nice support for string based dictionary-like lookups, but I totally agree that this will be the most common use-case **by far**!
> 
> Note that we have just added convenience methods for this case, we did not design the API graph around it. So we could equally add convenience methods for other situations in the future.

agreed :+1: from my POV, I'm happy (besides the small NIT I just found)

python/ql/lib/change-notes/2022-10-04-api-subscript-nodes.md
yoff: Heh, yes!

python/ql/lib/change-notes/2022-10-04-api-subscript-nodes.md
```diff
@@ -3,4 +3,4 @@ category: minorAnalysis
 ---
 * Fixed labels in the API graph pertaining to definitions of subscripts. Previously, these were labelled by `getMember` rather than `getASubscript`.
 * Added `API::SubscriptReadNode` and `API::SubscriptWriteNode` with convenience predicates to obtain the index and, in the case of a write, the written value.
-* Added convenience predicates to obtain a subscript at a specific index, when the index happens to be a statically known string.
+* Added convenience predicates `getMember("key")` to obtain a subscript at a specific index, when the index happens to be a statically known string.
```
RasmusWL, line 6: I think it should be this? :confused: 

```suggestion
* Added convenience predicates `getSubscript("key")` to obtain a subscript at a specific index, when the index happens to be a statically known string.
```

yoff: Wow, yes, now I understand your confusion...maybe lack of refreshes from me drove this concurrency issue..

python/ql/lib/semmle/python/ApiGraphs.qll
```diff
@@ -694,12 +791,18 @@ module API {
...
-          rhs.(DataFlow::ExprNode).getNode().getNode() = item.getValue()
+          dict.getItem(_) = item
+        |
+          // from `x` to `{ "key": x }`
+          rhs.(DataFlow::ExprNode).getNode().getNode() = item.getValue() and
+          lbl = Label::subscript()
+          or
+          // from `"key"` to `{ "key": x }`
+          rhs.(DataFlow::ExprNode).getNode().getNode() = item.getKey() and
+          lbl = Label::index()
```
tausbn, line 805: I'm beginning to wonder if either/both of these should include the string index (if present). There's something here that I can't quite put my finger on...

tausbn (commented):
I feel like I've managed to confuse myself a bit while thinking about (the consequences of) this PR. Perhaps we should have a meeting to go over some of my questions/suggestions below?

python/ql/lib/semmle/python/ApiGraphs.qll
```diff
@@ -474,6 +504,73 @@ module API {
...
+    SubscriptWriteNode() { this = subscripted.getASubscript().asSink() }
+
+    /** Gets an API node for the object being subscripted. */
+    API::Node getObject() { result = subscripted }
+
+    /** Gets an API node representing the index of this write. */
+    API::Node getIndex() {
+      result = subscripted.getIndex() and
+      exists(PY::SubscriptNode subscriptNode |
+        subscriptNode.(PY::DefinitionNode).getValue() = this.asCfgNode() and
```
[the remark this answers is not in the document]
yoff: Indeed, the idea is that other syntax could also constitute a subscript access. In the example you give, the read and the write will have different indices.

python/ql/lib/semmle/python/ApiGraphs.qll
```diff
@@ -474,6 +504,73 @@ module API {
...
+   * x = foo[bar]
+   * y = foo[baz]
+   * ```
+   * there wil be subscript edges from `foo` to both `foo[bar]` and `foo[baz]`,
+   * and there will be index edges from `foo` to both `bar` and `baz`.
+   * There will be a separate `SubscriptReadNode` for `foo[bar]` (and one for `foo[baz]`),
+   * and the member predicates `getObject` and `getIndex` will return nodes
+   * consistent with that particular read.
+   */
+  class SubscriptReadNode extends DataFlow::Node {
```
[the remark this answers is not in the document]
yoff: Would you prefer this? For attributes we have them split up. And I would assume that models would mostly need a specific one, matching the modelled API..

python/ql/lib/semmle/python/ApiGraphs.qll
yoff: This case should be folded into the general DataFlow::SubscriptWriteNode I think..

python/ql/lib/semmle/python/ApiGraphs.qll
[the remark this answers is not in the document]
tausbn: My point wasn't that it should only cover the `a[b] = c` syntax, but rather that it doesn't make much sense to me to tie it to `c`.

Perhaps a comparison with [`AttrWrite`](https://github.com/github/codeql/blob/main/python/ql/lib/semmle/python/dataflow/new/internal/Attributes.qll#L79) is useful. Here, for the statement `a.b = c`, the `AttrWrite` node is associated with `a.b`, not `c`. Of course, `c` is still available via `getValue`.

Similarly, if and when we add support for interpreting `a.__setitem__(b, c)` as a subscript write, I would expect the entire expression to be the `SubscriptWriteNode`, not just the `c` part.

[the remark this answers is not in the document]
tausbn: The underlying `DefinitionNode` is associated with the _target_ of the assignment, not the value or the whole assignment: https://github.com/github/codeql/blob/main/python/ql/lib/semmle/python/Flow.qll#L625

(Indeed, it must, otherwise your inline cast would fail.)

To put a finer point on it, say I want to _only_ select those `SubscriptWriteNode`s that correspond to the form `a[b]`, and not the `__setitem__` form. It would be very intuitive for me to write, say, `this.asCfgNode() instanceof SubscriptNode`, but that wouldn't do what I expected!

python/ql/lib/semmle/python/ApiGraphs.qll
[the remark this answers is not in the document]
tausbn: I'm mostly just wondering if it's _useful_ to have this distinction. It seems to me that the expression `a[b]`, regardless of whether it appears on the left hand side of an assignment or not, could still be thought of as an API source. 

Having things split up leads to a lot of duplication, e.g. the definition of `getSubscript/1` above. And even worse, by having two distinct classes, the return value of `getSubscript/1` can't be more specific than `API::Node`, which means you'll have to explicitly cast into the appropriate `Read`/`Write` class to use the convenience methods. (Contrast with `getACall() which returns an `API::CallNode`.)

To fix this, we could introduce a third `SubscriptRefNode` class (superclass of both `Read` and `Write`), but this would only help if you wanted to extract the object or index. For the value being written, you would still need to cast to `API::SubscriptWriteNode`.

yoff:
Closed, superseded by #10783.

PR #10609, opened 2022-09-28
C++: Make `OverrunWriteProductFlow` raise alerts on overflows

MathiasVP:
We merged https://github.com/github/codeql/pull/10398 before we really finished off the `OverrunWriteProductFlow.ql` query. In particular:
- Its tests weren't all annotated.
- It was highlighting any flow from an `(allocation, size)` pair to an `ArrayFunction` operation regardless of whether the operation was dangerous or not.
- It was missing metadata and qhelp.

This PR fixed all of those rings. Commit-by-commit review is encouraged:

- https://github.com/github/codeql/commit/51758aa928113d810acf72269b07f2d1c0cf3770 adds more tests and adds annotations.
- https://github.com/github/codeql/commit/ccbbb5754e78b809133b69d6f18657ad696d9d08 does the majority of the work of this PR: it turns the query into a query that's finding vulnerabilities instead of being a testing query for the product-dataflow library.
- https://github.com/github/codeql/commit/769ff5c6f3fea89635daa98a8e376ef075b65207 adds an `isAdditionalFlowStep` predicate to the product-dataflow library to fix some missing flow results.
- https://github.com/github/codeql/commit/4ab676774e3b634368ede30e5cb2dc1a443ee84a adds qhelp to the new query. This is mostly a copy/paste from https://github.com/github/codeql/blob/main/cpp/ql/src/Critical/OverflowStatic.qhelp, except that I've edited the parts that specifically mentions `static` buffers (since this new query deals with any kind of buffers).

### DIFF
--- a/cpp/ql/src/experimental/Likely Bugs/OverrunWriteProductFlow.ql
+++ b/cpp/ql/src/experimental/Likely Bugs/OverrunWriteProductFlow.ql
@@ -123,6 +123,21 @@ class StringSizeConfiguration extends ProductFlow::Configuration {
       delta > state2.toInt()
     )
   }
+
+  override predicate isAdditionalFlowStep2(
+    DataFlow::Node node1, DataFlow::FlowState state1, DataFlow::Node node2,
+    DataFlow::FlowState state2
+  ) {
+    exists(AddInstruction add, Operand op, int delta, int s1, int s2 |
+      s1 = [0 .. 32] and // An arbitrary bound because we need to bound `state`
+      state1 = s1.toString() and
+      state2 = s2.toString() and
+      add.hasOperands(node1.asOperand(), op) and
+      semBounded(op.getDef(), any(SemZeroBound zero), delta, true, _) and
+      node2.asInstruction() = add and
+      s1 = s2 + delta
+    )
+  }
 }
 
 from
--- a/cpp/ql/src/experimental/Likely Bugs/OverrunWriteProductFlow.ql
+++ b/cpp/ql/src/experimental/Likely Bugs/OverrunWriteProductFlow.ql
@@ -39,49 +39,27 @@ predicate bounded(Instruction i, Instruction b, int delta) {
   )
 }
 
-/**
- * Holds if the combination of `n` and `state` represents an appropriate
- * source for the expression `e` suitable for use-use flow.
- */
-private predicate hasSizeImpl(Expr e, DataFlow::Node n, string state) {
-  // The simple case: If the size is a variable access with no qualifier we can just use the
-  // dataflow node for that expression and no state.
-  exists(VariableAccess va |
-    va = e and
-    not va instanceof FieldAccess and
-    n.asConvertedExpr() = va.getFullyConverted() and
-    state = "0"
-  )
-  or
-  // If the size is a choice between two expressions we allow both to be nodes representing the size.
-  exists(ConditionalExpr cond | cond = e | hasSizeImpl([cond.getThen(), cond.getElse()], n, state))
-  or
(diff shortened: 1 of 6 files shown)


MathiasVP:
@rdmarsh2 I think the `isAdditionalFlowStep` I added for this query in https://github.com/github/codeql/commit/769ff5c6f3fea89635daa98a8e376ef075b65207 might also be useful for the `cpp/invalid-pointer-deref` query.

cpp/ql/src/experimental/Likely Bugs/OverrunWriteProductFlow.ql
```diff
@@ -1,42 +1,139 @@
...
+    e.(AddExpr).hasOperands(const, nonconst) and
+    state = const.getValue() and
+    hasSizeImpl(nonconst, n, _)
+  )
+  or
+  exists(Expr const, Expr nonconst |
+    e.(SubExpr).hasOperands(const, nonconst) and
+    state = "-" + const.getValue() and
+    hasSizeImpl(nonconst, n, _)
+  )
```
rdmarsh2, line 71: This case seems like it could break in a few different ways - const minus nonconst or a negative constant would both fail here, I think?

cpp/ql/src/experimental/Likely Bugs/OverrunWriteProductFlow.ql
```diff
@@ -1,42 +1,139 @@
...
-    exists(CallInstruction c, int bufIndex, int sizeIndex |
-      c.getStaticCallTarget().(ArrayFunction).hasArrayWithVariableSize(bufIndex, sizeIndex) and
-      c.getArgument(bufIndex) = bufSink.asInstruction() and
-      c.getArgument(sizeIndex) = sizeSink.asInstruction()
+  override predicate isSinkPair(
+    DataFlow::Node bufSink, DataFlow::FlowState state1, DataFlow::Node sizeSink,
+    DataFlow::FlowState state2
+  ) {
+    state1 instanceof DataFlow::FlowStateEmpty and
+    state2 = [0 .. 32].toString() and // An arbitrary bound because we need to bound `state2`
```
rdmarsh2, line 120: Will we still need this bound once we have arbitrary flow state types?

cpp/ql/src/experimental/Likely Bugs/OverrunWriteProductFlow.ql
MathiasVP: Oh, yeah you might be right. This might also affect the `cpp/invalid-pointer-deref`. I'll add some tests to both queries and fix this.

MathiasVP: Yeah, you were totally right: https://github.com/github/codeql/pull/10609/commits/6537c817ef1b14e0f18b1f4ed3687f190e66da21 adds a bunch more tests that showed that this didn't work at all for subtraction 🤦. https://github.com/github/codeql/pull/10609/commits/70837dbd934e151ed1c6d9ddacf417182d7c9a9c fixes all those issues, and https://github.com/github/codeql/pull/10609/commits/4e3b445515ecfad47e3773f93c51011c397d1d76 demonstrates that it works on the newly added tests!

rdmarsh2: Very elegant solution, I like this a lot better.

rdmarsh2 (approved):

cpp/ql/src/experimental/Likely Bugs/OverrunWriteProductFlow.ql
MathiasVP: Thank you 😄!

PR #10610, opened 2022-09-28
Ruby: tweak ActionController::getAnAction

nickrolfe:


nickrolfe:
I thought maybe this would result in slightly better compiler output, but I now realise that we never actually call `getAnAction`…

PR #10613, opened 2022-09-28
ATM: Update expected test output

henrymercer:
Update the expected test output given some changes to the JavaScript libraries which reclassified one of the test endpoints.


TomBolton (approved):
Thanks for doing this @henrymercer 

:shipit:

PR #10616, opened 2022-09-29
QL: adjust the consistency query to not be noisy on parameterised modules

erik-krogh:
The query was starting to get noisy on e.g. https://github.com/github/codeql/pull/10592 

With these changes the query reports nothing on `main` (and nothing on the PR above).


MathiasVP (approved):
LGTM!

PR #10617, opened 2022-09-29
Kotlin: extract operator expression when operator is in method call form

tamasvajk:


igfoo:
My rationale for extracting them differently is that people may wish to have queries that check coding style, for example, where the syntax used is useful information. I guess you want to unify them to make higher-level dataflow etc simpler. We'll need to make a decision on which way we want to go.

tamasvajk:
> My rationale for extracting them differently is that people may wish to have queries that check coding style, for example, where the syntax used is useful information. I guess you want to unify them to make higher-level dataflow etc simpler. We'll need to make a decision on which way we want to go.

Yes, that's the reason. I don't want to add summaries or modify the dataflow library in any way for these to be handled. Also, we're getting rid of at least one unneeded warnings: https://github.com/github/codeql/pull/10617/commits/de9714de84c097c838c0a340ed0ea46334daa6bc#diff-ef330ff9b3a3a796ae2c0b8656c4106a08394947b68a6d0de716764ea42b2d2bR2.

igfoo:
OK, how about we go with this PR, then, and possibly (either now or in the future) add a table so that we can do
```
if (c.origin == IrStatementOrigin.PLUS) {
    tw.writeExpr_using_operator_syntax(id)
}
```
?

tamasvajk:
Sounds good to me.

tamasvajk:
I've rebase this PR to fix merge conflicts.

java/ql/test/kotlin/library-tests/exprs/binop.expected
```diff
@@ -73,13 +73,12 @@
...
 | exprs.kt:114:14:114:21 | ... \|\| ... | exprs.kt:114:14:114:15 | b1 | exprs.kt:114:20:114:21 | b2 |
 | exprs.kt:127:31:127:41 | ... + ... | exprs.kt:127:31:127:34 | str1 | exprs.kt:127:38:127:41 | str2 |
-| exprs.kt:127:50:127:60 | ... + ... | exprs.kt:127:50:127:53 | str2 | exprs.kt:127:57:127:60 | str1 |
 | exprs.kt:128:16:128:26 | ... + ... | exprs.kt:128:16:128:19 | str1 | exprs.kt:128:23:128:26 | str2 |
 | exprs.kt:131:12:131:23 | ... > ... | exprs.kt:131:12:131:19 | variable | exprs.kt:131:23:131:23 | 0 |
 | exprs.kt:135:12:135:20 | ... + ... | exprs.kt:135:12:135:14 | 123 | exprs.kt:135:18:135:20 | 456 |
 | exprs.kt:161:8:161:16 | ... (value not-equals) ... | exprs.kt:161:8:161:8 | r | exprs.kt:161:13:161:16 | null |
 | exprs.kt:190:31:190:37 | ... + ... | exprs.kt:190:31:190:32 | getA1(...) | exprs.kt:190:36:190:37 | a2 |
-| exprs.kt:204:19:204:23 | ... + ... | exprs.kt:204:19:204:19 | s | exprs.kt:204:23:204:23 | 5 |
+| exprs.kt:205:23:205:29 | ... + ... | exprs.kt:205:20:205:21 | ...!! | exprs.kt:205:28:205:28 | 5 |
```
igfoo, line 81: Why did one of the 127s and 204 disappear?

tamasvajk: Because those were incorrect.
- 204 is `s + 5`, where `s: String?`, so it's actually calling `public operator fun String?.plus(other: Any?): String`, which is `Intrinsics.stringPlus()` in JVM.
- 127 is similar: 
```
val str1: String = "string lit"
val str2: String? = "string lit"
val str5: String = "foo ${str1 + str2} bar ${str2 + str1} baz"
```
So again, `str2 + str1` should be a call to `Intrinsics.stringPlus()`. (But because it's in an interpolated string, in reality it is optimized to a `StringBuilder.append`, which we don't do in the DB)

igfoo (approved):

PR #10622, opened 2022-09-29
Ruby: Postupdate notes for assignment expressions.

michaelnebel:
This is a follow up of https://github.com/github/codeql/pull/10444
There is flow from "taint" to the field `field` on post-update node for `x` in the call to `taint_field`.

```ruby
def taint_field x
  x.field = "taint"
end

taint_field(y = x)
sink(x.field)
```

### DIFF
--- a/ruby/ql/lib/codeql/ruby/dataflow/internal/DataFlowPrivate.qll
+++ b/ruby/ql/lib/codeql/ruby/dataflow/internal/DataFlowPrivate.qll
@@ -52,6 +52,8 @@ private CfgNodes::ExprCfgNode getALastEvalNode(CfgNodes::ExprCfgNode n) {
   or
   result = n.(CfgNodes::ExprNodes::ConditionalExprCfgNode).getBranch(_)
   or
+  result = n.(CfgNodes::ExprNodes::AssignExprCfgNode).getRhs()
+  or
   exists(CfgNodes::AstCfgNode branch |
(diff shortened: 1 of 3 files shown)


ruby/ql/lib/codeql/ruby/dataflow/internal/DataFlowPrivate.qll
```diff
@@ -52,6 +52,8 @@ private CfgNodes::ExprCfgNode getALastEvalNode(CfgNodes::ExprCfgNode n) {
   or
   result = n.(CfgNodes::ExprNodes::ConditionalExprCfgNode).getBranch(_)
   or
+  result = n.(CfgNodes::ExprNodes::AssignExprCfgNode).getRhs()
```
hvitved, line 55: This case should now be removed in `localFlowStepCommon`.

hvitved (changes requested):

ruby/ql/lib/codeql/ruby/dataflow/internal/DataFlowPrivate.qll
michaelnebel: Oh - right!

hvitved (approved):
Thanks, LGTM. Let's see what DCA says.

michaelnebel:
DCA doesn't show any change in alerts or any signs of performance regressions.

PR #10625, opened 2022-09-29
Added job.getRunsOn

henti:
Added https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idruns-on


erik-krogh (dismissed):
👍

erik-krogh (approved):

PR #10626, opened 2022-09-29
WIP: Kotlin: Adjust java-kotlin function lookup

tamasvajk:


java/ql/integration-tests/posix-only/kotlin/gradle_kotlinx_serialization/diagnostics.expected
```diff
@@ -1 +1,5 @@
+| CodeQL Kotlin extractor | 2 |  | Couldn't find a Java equivalent function to kotlin.Int.and in java.lang.Integer | file://:0:0:0:0 | file://:0:0:0:0 |
+| CodeQL Kotlin extractor | 2 |  | Couldn't find a Java equivalent function to kotlin.Int.or in java.lang.Integer | file://:0:0:0:0 | file://:0:0:0:0 |
```
tamasvajk, line 2: We're not extracting these at the moment as operator calls, but as method calls on `kotlin.Int`, which is probably not what we want.

PR #10628, opened 2022-09-29
Java: Type based summary models.

michaelnebel:
In this PR we introduce the a query for generating *type based* models for Java.
This is the Java version of https://github.com/github/codeql/pull/10238.

It is worth mentioning the following differences.
- Java does not have a `Func` or `delegate` like type, but instead relies on *functional interfaces* (ie. interfaces with a single abstract method) as these can be perceived as functions, which complicated the implementation compared to C#.
- Java has different way of providing accesspaths into collections depending on the collection type.
- As a reference example `java.util.stream.Stream` is used (an approximate equivalent to the `System.Linq.Enumerable` extension methods for C#).


java/ql/test/utils/model-generator/typebasedflow/p/Stream.java
```diff
@@ -0,0 +1,166 @@
+package p;
+
+import java.util.*;
+import java.util.function.*;
+import java.util.stream.*;
+
+public class Stream<T> {
```
aschackmull, line 7: Something's weird here. We're importing `java.util.stream.*` such that the standard `Stream` class is in scope, but we're also defining a class with the same name, which I guess shadows the imported name?? Are we attempting to stub-implement `java.util.stream.Stream` or import it?

java/ql/test/utils/model-generator/typebasedflow/p/TypeBasedComplex.java
```diff
@@ -0,0 +1,57 @@
...
+
+    public List<T> getMany() {
+        return null;
+    }
+
+    public Integer apply(Function<T, Integer> f) {
+        return null;
+    }
+
+    public <T1, T2> T2 apply2(T1 x, Function<T1, T2> f) {
```
aschackmull, line 19: This type signature doesn't refer to `T`. Was that intentional or was the method perhaps meant to be static?

aschackmull: If it's meant as-is, then it deserves an explanatory comment.

java/ql/test/utils/model-generator/typebasedflow/CaptureTypeBasedSummaryModels.expected
```diff
@@ -0,0 +1,124 @@
+| p;MyFunction;true;apply;(Object,Object);;Argument[-1].SyntheticField[ArgType2];ReturnValue;value;generated |
```
aschackmull, line 1: The naming of the synthetic fields could be a bit more precise. In particular they ought to refer to the type (when they aren't derived from a super-class/-interface), or reuse the field that the supertype indicates. I.e. a class `MyClass<T>` can use `SyntheticField[package.MyClass.T]`, a class `MySubclass<T2> extends MyClass<T2>` should use the same field as the superclass `SyntheticField[package.MyClass.T]`, and an extension of known collection-types should reuse those, such that `MyMultiMap<K,V> implements Map<K,List<V>>` uses `MapKey` and `MapValue.Element`. (The test `TypeBasedCollection<T> extends ArrayList<T>` already checks that to some degree).
Note that this will also ensure that methods on `A` and `B` will be compatible when we have `class A<T1,T2> extends B<T2>`, whereas the current approach fails to do so (equating the wrong fields and misses the equality that ought to be present).

aschackmull: Given that a class can implement multiple interfaces, then we could have a conflict, so in that case the preference order should probably be something like "known collection-like super-interface" > "super-class" > "super-interface".

aschackmull: Also, could we make this into some sort of inline test expectation? We don't necessarily need to use the inline expectation library - a more ad-hoc approach could be fine: I'm thinking perhaps collect all comments of the form `// MaD = ...` and then compare the query output with the set of those strings. We don't necessarily need to make use of the comment locations, which means that we could just write several expectation lines above the method that generates them.

java/ql/test/utils/model-generator/typebasedflow/p/Stream.java
aschackmull: I wonder then why the generated models use `Element`? I mean they certainly ought to for `java.util.stream.Stream` but this stub class isn't that. If the stub class was implementing, say, `Iterable<T>` then it would make more sense, as that ought to be enough to make it use `Element` as opposed to generating a synthetic field.

aschackmull: When we're stubbing like this, then I think we should use explicit imports from `java.util.stream` such that we aren't shadowing anything. That would make it more clear that the references to `Stream` are in fact this stub class.

java/ql/test/utils/model-generator/typebasedflow/p/TypeBasedComplex.java
michaelnebel: Yes, that as intentional. The test is there to check that no summary that accesses the *implicit* field corresponding to the field of type `T` will be generated (even for instance methods).

java/ql/test/utils/model-generator/typebasedflow/CaptureTypeBasedSummaryModels.expected
michaelnebel: > The naming of the synthetic fields could be a bit more precise. In particular they ought to refer to the type (when they aren't derived from a super-class/-interface), or reuse the field that the supertype indicates. I.e. a class `MyClass<T>` can use `SyntheticField[package.MyClass.T]`, a class `MySubclass<T2> extends MyClass<T2>` should use the same field as the superclass `SyntheticField[package.MyClass.T]`, and an extension of known collection-types should reuse those, such that `MyMultiMap<K,V> implements Map<K,List<V>>` uses `MapKey` and `MapValue.Element`. (The test `TypeBasedCollection<T> extends ArrayList<T>` already checks that to some degree). Note that this will also ensure that methods on `A` and `B` will be compatible when we have `class A<T1,T2> extends B<T2>`, whereas the current approach fails to do so (equating the wrong fields and misses the equality that ought to be present).

You are absolutely right that the implementation could be improved and that the precision for more complicated inheritance is not ideal. The same is the case for C#, which also lacks exactly what you are describing here.
Do you prefer that we get this sorted as a first attempt before merging this?

aschackmull: > Do you prefer that we get this sorted out as a first attempt before merging this?

I'll leave that to your preference. But I would like a refactor of the test, such that the models could be seen directly in the Java source next to the method.

aschackmull: Something like
```
string expects() {
  exists(JavaDoc doc |
    doc.getChild(0).toString().regexpCature(" *MaD *= *\\(.*\\)", 1) = result
  )
}

...
```

michaelnebel: Yes, this is a great idea. I have included in it in this PR for the type based summaries, but maybe we should consider to do the same thing for both Java and C# for all model generator like tests.

java/ql/test/utils/model-generator/typebasedflow/p/TypeBasedComplex.java
michaelnebel: Comment added.

java/ql/test/utils/model-generator/typebasedflow/p/Stream.java
michaelnebel: (1) As far as can tell from the documentation of the `Stream` (java) interface and `ContainerType` (QL) implementation, then `Stream` will not be identified as a `ContainerType` (note that `Stream<T>` does not implement `Iterable<T>`). The implementation on this PR will also perceive classes or interfaces having any method that returns an `Iterator` as being a container like type (but this purely has effect the type based model generation implementation. 
(2) Imports has been changed to avoid confusion.

java/ql/test/utils/model-generator/typebasedflow/CaptureTypeBasedSummaryModels.ql
```diff
@@ -0,0 +1,24 @@
...
+private string expects() {
+  exists(Javadoc doc | doc.getChild(0).toString().regexpCapture(" *MaD=(.*)", 1) = result)
+}
+
+private string flows() { exists(TypeBasedFlowTargetApi api | result = captureFlow(api)) }
+
+query predicate unexpectedSummary(string msg) {
+  exists(string flow |
+    flow = flows() and
+    not exists(string e | e = expects() and e = flow) and
```
aschackmull, line 13: Simplify:
```suggestion
    not flow = expects() and
```

java/ql/test/utils/model-generator/typebasedflow/CaptureTypeBasedSummaryModels.ql
```diff
@@ -0,0 +1,24 @@
...
+    msg = "Unexpected summary found: " + flow
+  )
+}
+
+query predicate expectedSummary(string msg) {
+  exists(string e |
+    e = expects() and
+    not exists(string flow | flow = flows() and e = flow) and
```
aschackmull, line 21:
```suggestion
    not e = flows() and
```

java/ql/test/utils/model-generator/typebasedflow/p/Stream.java
```diff
@@ -0,0 +1,249 @@
...
+    public Optional<T> findAny() {
+        return null;
+    }
+
+    // MaD=p;Stream;true;findFirst;();;Argument[-1].Element;ReturnValue.SyntheticField[ArgType0];value;generated
+    public Optional<T> findFirst() {
+        return null;
+    }
+
+    // public <R> Stream<R> flatMap(Function<? super T, ? extends Stream<? extends
```
aschackmull, line 70: Why is this test case excluded?

aschackmull: If it doesn't work, then we might as well keep it with a comment about why and what doesn't work.

java/ql/test/utils/model-generator/typebasedflow/p/Stream.java
```diff
@@ -0,0 +1,249 @@
...
+    // MaD=p;Stream;true;forEachOrdered;(Consumer);;Argument[-1].Element;Argument[0].Parameter[0];value;generated
+    public void forEachOrdered(Consumer<? super T> action) {
+    }
+
+    // MaD=p;Stream;true;generate;(Supplier);;Argument[0].ReturnValue;ReturnValue.Element;value;generated
+    public static <T> Stream<T> generate(Supplier<T> s) {
+        return null;
+    }
+
+    // Model generator adds a couple of extra models, which can't be
```
aschackmull, line 103: All four of these models look correct to me?

java/ql/test/utils/model-generator/typebasedflow/p/Stream.java
```diff
@@ -0,0 +1,249 @@
...
+    public Optional<T> min(Comparator<? super T> comparator) {
+        return null;
+    }
+
+    // MaD=p;Stream;true;noneMatch;(Predicate);;Argument[-1].Element;Argument[0].Parameter[0];value;generated
+    public boolean noneMatch(Predicate<? super T> predicate) {
+        return false;
+    }
+
+    // Issue with model generator. ... is not supported.
```
aschackmull, line 158: That's fine, but no reason to comment the test case.

java/ql/test/utils/model-generator/typebasedflow/p/Stream.java
```diff
@@ -0,0 +1,249 @@
...
+    public Stream<T> peek(Consumer<? super T> action) {
+        return null;
+    }
+
+    // Model generator yields a couple of extra results as models are generated for
+    // writing to the stream.
+    // MaD=p;Stream;true;reduce;(BinaryOperator);;Argument[-1].Element;Argument[0].Parameter[0];value;generated
+    // MaD=p;Stream;true;reduce;(BinaryOperator);;Argument[-1].Element;Argument[0].Parameter[1];value;generated
+    // MaD=p;Stream;true;reduce;(BinaryOperator);;Argument[-1].Element;ReturnValue.SyntheticField[ArgType0];value;generated
+    // MaD=p;Stream;true;reduce;(BinaryOperator);;Argument[0].ReturnValue;Argument[-1].Element;value;generated
```
aschackmull, line 179: Yeah, this one is spurious but mostly harmless. Maybe it would be nice if the regex for matching `MaD=` would allow a `SPURIOUS` prefix so we could write
```suggestion
    // SPURIOUS MaD=p;Stream;true;reduce;(BinaryOperator);;Argument[0].ReturnValue;Argument[-1].Element;value;generated
```

java/ql/test/utils/model-generator/typebasedflow/p/Stream.java
```diff
@@ -0,0 +1,249 @@
...
+        return null;
+    }
+
+    // MaD=p;Stream;true;peek;(Consumer);;Argument[-1].Element;Argument[0].Parameter[0];value;generated
+    // MaD=p;Stream;true;peek;(Consumer);;Argument[-1].Element;ReturnValue.Element;value;generated
+    public Stream<T> peek(Consumer<? super T> action) {
+        return null;
+    }
+
+    // Model generator yields a couple of extra results as models are generated for
```
aschackmull, line 174: s/a couple of/one/

java/ql/src/utils/modelgenerator/internal/CaptureTypeBasedSummaryModels.qll
```diff
@@ -0,0 +1,333 @@
...
+ * A type representing instantiations of class types
+ * that has a method which returns an iterator.
+ */
+private class IterableType extends Class {
+  private Type elementType;
+
+  IterableType() {
+    exists(Method m, Type return, GenericType t, Type et, int position | m.getDeclaringType() = t |
+      return = m.getReturnType() and
+      return.getName().matches("Iterator%") and
```
aschackmull, line 17:
```suggestion
      return.getSourceDeclaration().hasQualifiedName("java.util", "Iterator") and
```

java/ql/src/utils/modelgenerator/internal/CaptureTypeBasedSummaryModels.qll
```diff
@@ -0,0 +1,333 @@
...
+
+/**
+ * A type representing instantiations of class types
+ * that has a method which returns an iterator.
+ */
+private class IterableType extends Class {
+  private Type elementType;
+
+  IterableType() {
+    exists(Method m, Type return, GenericType t, Type et, int position | m.getDeclaringType() = t |
```
aschackmull, line 15:
```suggestion
    exists(Method m, RefType return, GenericType t, Type et, int position | m.getDeclaringType() = t |
```

java/ql/src/utils/modelgenerator/internal/CaptureTypeBasedSummaryModels.qll
```diff
@@ -0,0 +1,333 @@
...
+ * Holds if type `bound` is a lower bound for type `t` or equal to `t`.
+ */
+private predicate isEffectivelyLowerBound(Type t, Type bound) {
+  t = bound or
+  t.(Wildcard).getLowerBound().getType() = bound
+}
+
+/**
+ * Holds if `t` is a container like type of `tv` (eg. `List<T>`).
+ * Note that collections are covariant in their element type.
```
aschackmull, line 48: No? In Java they aren't. Or is this something that you assume for some heuristic purpose?

java/ql/src/utils/modelgenerator/internal/CaptureTypeBasedSummaryModels.qll
```diff
@@ -0,0 +1,333 @@
...
+      genericContainerType(pt, tv)
+    )
+  )
+}
+
+/**
+ * Gets the string representation of a synthetic field corresponding to `tv`.
+ */
+private string getSyntheticField(TypeVariable tv) {
+  result = ".SyntheticField[ArgType" + tv.getIndex() + "]"
```
aschackmull, line 120: This will need some refinement at some point, but it's probably good enough as a starting point.

java/ql/src/utils/modelgenerator/internal/CaptureTypeBasedSummaryModels.qll
```diff
@@ -0,0 +1,333 @@
...
+private import semmle.code.java.Collections
+private import semmle.code.java.dataflow.internal.ContainerFlow
+private import CaptureModelsSpecific as Specific
+private import CaptureModels
+
+/**
+ * A type representing instantiations of class types
+ * that has a method which returns an iterator.
+ */
+private class IterableType extends Class {
```
aschackmull, line 11: There are some implicit assumptions throughout this file that an `IterableType` has exactly one type variable. This is probably fine as a starting point, as that's likely to generally be the case, but might deserve a comment.

java/ql/src/utils/modelgenerator/internal/CaptureTypeBasedSummaryModels.qll
```diff
@@ -0,0 +1,333 @@
...
+  private TypeVariable getTypeReplacement(Type t) {
+    exists(int position |
+      instantiates(this, fi, position, t) and
+      result = fi.getTypeParameter(position)
+    )
+  }
+
+  /**
+   * Gets the parameter type of `this` function at position `position`.
+   * Note that functions are contravariant in their parameter types.
```
aschackmull, line 165:
```suggestion
   * Note that functions are often contravariant in their parameter types.
```

java/ql/src/utils/modelgenerator/internal/CaptureTypeBasedSummaryModels.qll
```diff
@@ -0,0 +1,333 @@
...
+  Type getParameterType(int position) {
+    exists(Type t |
+      fi.getRunMethod().getParameterType(position) = this.getTypeReplacement(t) and
+      isEffectivelyLowerBound(t, result)
+    )
+  }
+
+  /**
+   * Gets the return type of `this` function.
+   * Note that functions are covariant in their return type.
```
aschackmull, line 176:
```suggestion
   * Note that functions are often covariant in their return type.
```

java/ql/src/utils/modelgenerator/internal/CaptureTypeBasedSummaryModels.qll
```diff
@@ -0,0 +1,333 @@
...
+
+/**
+ * Holds if `callable` has a function parameter `f` at parameter position `position`.
+ */
+private predicate functional(Callable callable, Function f, int position) {
+  exists(Parameter p |
+    p = callable.getAParameter() and
+    f = p.getType() and
+    position = p.getPosition()
+  )
```
aschackmull, line 194: Simplify:
```suggestion
  callable.getParameterType(position) = f
```

aschackmull (commented):
Mostly minor comments, otherwise LGTM!

java/ql/src/utils/modelgenerator/internal/CaptureTypeBasedSummaryModels.qll
michaelnebel: Ok, that claim is to strong.
At least the idea is that, if we have "relevant" type variable that serves as an upper bound for a given type, then we use that to eg. decide if the method reads from an implicit field or the implicit collection.
I will delete the comment.

java/ql/src/utils/modelgenerator/internal/CaptureTypeBasedSummaryModels.qll
michaelnebel: Yes, we do not dig into the inheritance hierarchy at the moment, but that needs to be improved.

java/ql/src/utils/modelgenerator/internal/CaptureTypeBasedSummaryModels.qll
michaelnebel: An IterableType can be parameterized over multiple type variables. As an example
```java
public class Stream2<T1, T2> {

    public Iterator<T1> iterator() {
        return null;
    }

    public T2 getField() {
        return null;
    }

    public T1 getElement() {
        return null;
    }
}
```
produce the models
```
p;Stream2;true;getElement;();;Argument[-1].Element;ReturnValue;value;generated
p;Stream2;true;getField;();;Argument[-1].SyntheticField[ArgType1];ReturnValue;value;generated p;Stream2;true;iterator;();;Argument[-1].Element;ReturnValue.Element;value;generated
```

I am not sure where you discovered the implicit assumptions.
If you have seen something, then it might be worth narrowing the InterableType definition, such we only accept types that are parameterized over a single type variable and where there exists an iterator method over that exact type.

java/ql/test/utils/model-generator/typebasedflow/p/Stream.java
michaelnebel: That is an oversight on my part - it does work and I will comment it back in.

java/ql/test/utils/model-generator/typebasedflow/p/Stream.java
michaelnebel: Yes, you are right! I will remove the comment (I had overlooked that the "seed" is the first element of the sequence, which is why I thought that Argument[0] -> ReturnValue.Element was incorrect - but that is not the case).

java/ql/test/utils/model-generator/typebasedflow/p/Stream.java
michaelnebel: I will comment it in.

java/ql/test/utils/model-generator/typebasedflow/p/Stream.java
michaelnebel: Also: Argument[0].ReturnValue -> Argument[0].Parameter[1] is spurious. In the documentation it is stated that only required is that the operator is associative.

michaelnebel: Hmm - looks like that one is included in the manual models.

java/ql/src/utils/modelgenerator/internal/CaptureTypeBasedSummaryModels.qll
aschackmull: Consider this (probably an obscure corner-case, but still):
```
public class Stream2<T1, T2> {
    public Iterator<T1> iterator1() {
        return null;
    }

    public Iterator<T2> iterator2() {
        return null;
    }

    public T1 getElement1() {
        return null;
    }

    public T2 getElement2() {
        return null;
    }
}
```

java/ql/src/utils/modelgenerator/internal/CaptureTypeBasedSummaryModels.qll
aschackmull: Not yet deleted.

michaelnebel: haha - no - you are right! I will try once more :-)

java/ql/src/utils/modelgenerator/internal/CaptureTypeBasedSummaryModels.qll
michaelnebel: You are right. In that case, we probably shouldn't think of `Stream2` as being an interable type.
It is an easy fix.

aschackmull (approved):

Saidyyh (commented):
2011685e89656c624fb2fe8e4ae3e16715727186

PR #10629, opened 2022-09-29
Python: Fix flask request modeling

RasmusWL:
This takes us part of the way. We still get multiple paths for the same
alert, but that will be fixed in a different PR.


tausbn (dismissed):
It's slightly strange to have the path start at the import, but I understand the reason for doing so.

I guess an alternative would be to _not_ have `request` be a remote flow source in its own right, but only attributes thereof.

In the short term, I think this is a good fix, however. 👍

RasmusWL:
> I guess an alternative would be to _not_ have `request` be a remote flow source in its own right, but only attributes thereof.

I didn't consider this, but I agree it could be a solution :thinking:

yoff:
Another alternative, which may amount to the same thing, is to say that importing the request object is not a case of remote input, but reading from it is. I am not sure if attribute reads is the only way to access it.

RasmusWL:
this PR turned into quite a bit of a mixed bag now :flushed: but everything was connected :shrug:

PR #10631, opened 2022-09-29
Python: Remove last `-p ../lib/` in `options` files

RasmusWL:
These were only needed for points-to.

If they only contained `--max-import-depth`, I've removed the `options` file entirely.


tausbn (approved):
Nice! 👍

PR #10634, opened 2022-09-29
Python: Rewrite type trackers

yoff:
While looking through our existing custom type trackers, I found these two that we already have the means to rewrite.


tausbn (approved):
Looks good to me. 👍

PR #10635, opened 2022-09-29
C++: Recognize allocation functions heuristically

MathiasVP:
This PR does two things:
- It adds a couple of new interface classes, `HeuristicAllocationExpr` and `HeuristicAllocationFunction`, that complement the already existing `AllocationExpr` and `HeuristicAllocation` classes with functions that we think may perform allocations.
- It switched the new `cpp/invalid-pointer-deref` query to use `HeuristicAllocationExpr` instead of `AllocationExpr` as a source of flow.

Currently, the heuristics used for detecting an `HeuristicAllocationFunction` is:
1. The function name matches `%alloc%`
2. It returns a pointer type
3. We can find a unique parameter of `unsigned` integral type

In addition, a `HeuristicAllocationFunction` is recognized as a `realloc`-like function if, in addition to the above, it:
1. Matches the name `%realloc%`, and
2. We can find a unique parameter of a pointer type

This finds a bunch of new allocation functions on the databases I've tried. For example, on `php` we recognize the following set of new allocation functions:

Screenshot 2022-09-29 at 20 04 07

I've manually verified that these are all implementations that we want to recognize as allocation functions.

Commit-by-commit review encouraged.
- https://github.com/github/codeql/commit/a9710453f4366ebb999b6e383e4cbc71ec364d51 adds the new classes by wrapping a parameterized module around the logic for generating an `AllocationExpr` from a call to an `AllocationFunction`.
- https://github.com/github/codeql/commit/d12a76559a44708274446cb4527da862ebaf5dcb uses the new class in `cpp/invalid-pointer-deref`.
- https://github.com/github/codeql/commit/2a514d60d448b3ef1ef0d7810c66cf158b3973a9 adds a `isBarrierIn` predicate to the product dataflow library. This is necessary since many of the heuristic allocators are just simple wrappers around allocation expressions that we already know, and we'd otherwise get duplicated paths from starting flow both inside those functions, _and_ from the return value of the function we now recognize as a heuristic allocation expression.

Finally, I had to add `nomagic` to a predicate in the product dataflow library because I was observing a bad join due to bad magic.

With this PR, we now recognize https://nvd.nist.gov/vuln/detail/CVE-2016-10160 out of the box 🎉.

### DIFF
--- a/cpp/ql/lib/semmle/code/cpp/models/implementations/Allocation.qll
+++ b/cpp/ql/lib/semmle/code/cpp/models/implementations/Allocation.qll
@@ -205,57 +205,104 @@ private predicate deconstructSizeExpr(Expr sizeExpr, Expr lengthExpr, int sizeof
   sizeof = 1
 }
 
-/**
- * An allocation expression that is a function call, such as call to `malloc`.
- */
-private class CallAllocationExpr extends AllocationExpr, FunctionCall {
-  AllocationFunction target;
+private signature class CallAllocationExprTarget extends Function;
 
-  CallAllocationExpr() {
-    target = this.getTarget() and
-    // realloc(ptr, 0) only frees the pointer
-    not (
-      exists(target.getReallocPtrArg()) and
-      this.getArgument(target.getSizeArg()).getValue().toInt() = 0
-    ) and
-    // these are modeled directly (and more accurately), avoid duplication
-    not exists(NewOrNewArrayExpr new | new.getAllocatorCall() = this)
-  }
+private module CallAllocationExprBase<CallAllocationExprTarget Target> {
+  signature int getReallocPtrArgSig(Target target);
 
-  override Expr getSizeExpr() {
-    exists(Expr sizeExpr | sizeExpr = this.getArgument(target.getSizeArg()) |
-      if exists(target.getSizeMult())
-      then result = sizeExpr
-      else
-        exists(Expr lengthExpr |
-          deconstructSizeExpr(sizeExpr, lengthExpr, _) and
-          result = lengthExpr
+  signature int getSizeArgSig(Target target);
+
+  signature int getSizeMultSig(Target target);
+
+  signature predicate requiresDeallocSig(Target target);
+
+  module With<
+  getReallocPtrArgSig/1 getReallocPtrArg, getSizeArgSig/1 getSizeArg, getSizeMultSig/1 getSizeMult,
+  requiresDeallocSig/1 requiresDealloc> {
+    /**
+     * An allocation expression that is a function call, such as call to `malloc`.
+     */
+    class CallAllocationExprImpl instanceof FunctionCall {
+      Target target;
+
+      CallAllocationExprImpl() {
+        target = this.getTarget() and
+        // realloc(ptr, 0) only frees the pointer
+        not (
+          exists(getReallocPtrArg(target)) and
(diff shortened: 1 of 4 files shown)


MathiasVP:
> I like the idea but I'm not following the parameterized module tricks you're doing.

Thanks! I'll add some comments to explain what's going on. The code looks slightly uglier than I wanted to because we cannot yet have type signature member predicates (although it's coming very soon!). So instead of writing the code like:
```ql
private signature class CallAllocationExprTarget extends Function {
  int getReallocPtrArgSig();

  int getSizeArgSig();

  int getSizeMultSig();

  predicate requiresDeallocSig();
}

private module CallAllocationExprBase<CallAllocationExprTarget Target> {
  class CallAllocationExprImpl instanceof FunctionCall {
    ...
  }
}
```
we have to write it as:
```ql
private signature class CallAllocationExprTarget extends Function;

private module CallAllocationExprBase<CallAllocationExprTarget Target> {
  signature int getReallocPtrArgSig(Target target);

  signature int getSizeArgSig(Target target);

  signature int getSizeMultSig(Target target);

  signature predicate requiresDeallocSig(Target target);

  module With<
  getReallocPtrArgSig/1 getReallocPtrArg, getSizeArgSig/1 getSizeArg, getSizeMultSig/1 getSizeMult,
  requiresDeallocSig/1 requiresDealloc> {
    class CallAllocationExprImpl instanceof FunctionCall {
      ...
    }
  }
}
```

MathiasVP:
cc @ginsbach on parameterized module usage (you can restrict your 👀 to [the first commit only](https://github.com/github/codeql/pull/10635/commits/a9710453f4366ebb999b6e383e4cbc71ec364d51))

ginsbach:
> cc @ginsbach on parameterized module usage (you can restrict your 👀 to [the first commit only](https://github.com/github/codeql/pull/10635/commits/a9710453f4366ebb999b6e383e4cbc71ec364d51))

Always happy to see parameterised module PRs!
Wouldn't it be easier to use a module signature, rather than 4 predicate signature arguments for `With<>`?

MathiasVP:
> Wouldn't it be easier to use a module signature, rather than 4 predicate signature arguments for With<>?

I ... I need to lookup my parameterized module syntax, then. I very much think you're right!

MathiasVP:
> Wouldn't it be easier to use a module signature, rather than 4 predicate signature arguments for `With<>`?

Fixed in https://github.com/github/codeql/pull/10635/commits/483ff58c39ec8704862705bffab7aef63950883f.

cpp/ql/lib/semmle/code/cpp/models/implementations/Allocation.qll
```diff
@@ -294,3 +386,99 @@ private class NewArrayAllocationExpr extends AllocationExpr, NewArrayExpr {
 
   override predicate requiresDealloc() { not exists(this.getPlacementPointer()) }
 }
+
+private module HeuristicAllocation {
+  /** A class that maps an `AllocationExpr` to an `HeuristicAllocationExpr`. */
+  private class HeuristicAllocationModeled extends HeuristicAllocationExpr instanceof AllocationExpr {
```
rdmarsh2, line 392: Do we actually need this? I'd expect it to be handled by `HeuristicAllocationFunctionModeled`

rdmarsh2 (approved):

PR #10639, opened 2022-09-30
Ruby: Do not attempt to track precise hash indices for floats and complex numbers

hvitved:
As pointed out by @asgerf, the inherent imprecision when comparing floats (and complex numbers) makes them unsuitable for tracking precise hash index information. We therefore simply treat reads and stores with float/complex indices as unknown indices.


asgerf:
I think we should also restrict it in `TKnownPairValueContent`

hvitved:
> I think we should also restrict it in `TKnownPairValueContent`

Right, I'll fix.

asgerf (dismissed):
LGTM assuming evaluation works out

asgerf (approved):

PR #10640, opened 2022-09-30
Java: Fix cartesian product

atorralba:
Fixes a cartesian product in the `canThrow` predicate, where `callable` wasn't bound in one side of the disjunct.


aschackmull:
The fix only affects one query, so you could have just selected that in dca.

aschackmull (approved):
LGTM.

PR #10641, opened 2022-09-30
JS/Python/Ruby: s/a HTML/an HTML/

nickrolfe:
HTML is universally pronounced beginning with a vowel-sound, so "a HTML" sounds wrong and weird.

This started with fixing the alert message for `IncompleteMultiCharacterSanitizationQuery.qll` in JavaScript and Ruby, but then I also fixed the places that showed up in a search (restricted to the dynamic language directories) for "a HTML".


aibaars (approved):

erik-krogh (approved):

tausbn (approved):
Wait... Are you saying I _shouldn't_ be pronouncing "HTML" as "Hut-mill"?

nickrolfe:
How about… HoTMaiL?

PR #10642, opened 2022-09-30
Run QLHelp preview for all languages

aibaars:
[Here’s an example](https://github.com/github/codeql/pull/10008#issuecomment-1210861535) of a qhelp preview that the workflow generates. 

If you don't want this for your language then please speak up before this is merged. You can always opt-out later with a paths-ignore property if needed.


nickrolfe (approved):
LGTM. Let's see if anyone complains :)

PR #10644, opened 2022-09-30
Ruby: Prevent reevaluation of expensive predicates

hvitved:
The effect of these changes was not quite what I had hoped for, presumably because the evaluator automatically avoids recomputing some of the expensive predicates during whole-suite evaluation. But it does make up for some of the slow-down on `gitlabhq` reported on https://github.com/github/codeql/pull/10090.

I also think the changes are worthwhile, because we are guaranteed to not reevaluate the predicates during interactive execution in VS Code.


erik-krogh (approved):
JS 👍

tausbn (approved):
Seems reasonable to me (as far as the shared/Python changes are concerned). 👍

PR #10645, opened 2022-09-30
C++: Port SimpleRangeAnalysis tests to the new range-analysis

MathiasVP:
This PR copy/pastes almost all the tests we had for `SimpleRangeAnalysis` into a testfile that's exercised by the new range-analysis library, and then moves the file over to use InlineExpectationsTest.

- https://github.com/github/codeql/commit/c4c7c95db2889bbc19c4d46c6a6bf91054b8077d does the copy/paste. Nothing else is going on in this commit.
- https://github.com/github/codeql/commit/d14b2c2880c1a53dc6c5fd96c0b1b848d0c0cc84 makes it possible for the expected values to have whitespace in them. This is needed to print expected values containing bounds with a `toString` that contains spaces (for example if a bound is `call to foo` + 42).
- Finally, https://github.com/github/codeql/commit/56b5010f6b505ad03af1b36d2a20fd6d2921d4b6 converts the file to InlineExpectationsTest by adding a bunch of InlineExpectations comments.

### DIFF
--- a/cpp/ql/test/library-tests/ir/range-analysis/RangeAnalysis.ql
+++ b/cpp/ql/test/library-tests/ir/range-analysis/RangeAnalysis.ql
@@ -17,7 +17,7 @@ class RangeAnalysisTest extends InlineExpectationsTest {
       tag = "range" and
       element = e.toString() and
       location = e.getLocation() and
-      value = getARangeString(e)
+      value = quote(getARangeString(e))
     )
   }
 }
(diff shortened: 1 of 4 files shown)


rdmarsh2:
Needs a .expected file

MathiasVP:
> Needs a .expected file

Oops. You're right! Will add it right away.

MathiasVP:
There's already an `RangeAnalysis.expected` in the directory. So I don't think that should be needed? That said, I see that tests are failing, but I'm getting a 500 when trying to see the error :(

rdmarsh2 (approved):

PR #10646, opened 2022-09-30
Kotlin: Simplify `kotlinFunctionToJavaEquivalent`

tamasvajk:
This PR 
- removes the approximation from `kotlinFunctionToJavaEquivalent` and instead looks up only exact member match;
- removes the list of `expectedMissingEquivalents`, which was hiding the fact that some of the mappings were missing;
- extracts missing operator calls: `inv`, `not`, `and`, `or`, `xor`, `shl`, `shr` and `ushr`;
- adds consistency query to report missing mappings
- fixes `kotlinFunctionToJavaEquivalent` to ignore nullability;
- ~only logs each missing mapping once.~


java/kotlin-extractor/src/main/kotlin/KotlinExtractorExtension.kt
```diff
@@ -201,6 +201,8 @@ class KotlinExtractorGlobalState {
     val syntheticToRealClassMap = HashMap<IrClass, IrClass?>()
     val syntheticToRealFunctionMap = HashMap<IrFunction, IrFunction?>()
     val syntheticToRealFieldMap = HashMap<IrField, IrField?>()
+
+    val deduplicatedWarnings = mutableSetOf<String>()
```
tamasvajk, line 205: Shouldn't these be some variants of concurrent maps? Is there any guarantee that we run single threaded? Could `KotlinExtractorExtension::generate` be invoked by kotlinc from multiple threads?

java/kotlin-extractor/src/main/kotlin/KotlinUsesExtractor.kt
```diff
@@ -1296,8 +1290,10 @@ open class KotlinUsesExtractor(
                             else null
                         } ?: run {
                             val parentFqName = parentClass.fqNameWhenAvailable?.asString()
-                            if (!expectedMissingEquivalents.contains(parentFqName)) {
-                                logger.warn("Couldn't find a Java equivalent function to $parentFqName.${f.name} in ${javaClass.fqNameWhenAvailable}")
+                            val msg = "Couldn't find a Java equivalent function to $parentFqName.${f.name.asString()} in ${javaClass.fqNameWhenAvailable?.asString()}"
+                            if (!globalExtensionState.deduplicatedWarnings.contains(msg)) {
+                                logger.warn(msg)
+                                globalExtensionState.deduplicatedWarnings.add(msg)
```
igfoo, line 1296: Note that https://github.com/github/codeql/blob/main/java/kotlin-extractor/src/main/kotlin/utils/Logger.kt#L117 is already going to limit the number of any kind of warning to (by default) 100, so although this is more specific, I'm not sure that it's worth the extra code here.

tamasvajk: Good point, let me revert this.

igfoo (approved):

PR #10650, opened 2022-09-30
Ruby: more type-tracking steps

asgerf:
Expands on the work in https://github.com/github/codeql/pull/10375 to add more type-tracking steps.

In particular, we are now able to generate type-tracking steps in more cases:
- The input and/or output can now be mapped to a parameter or return of a callback/block.
- `WithContent` and `WithoutContent` are now supported with precision that's more appropriate for type tracking.
- When both input and output end with a content, we can now map it to a load-store step.

There are a couple of performance tweaks along the way, resulting in mostly neutral performance overall.

I found the easiest way to leverage the existing test suite was to add inline test that rely on type-tracking, and simply let the test output document the different between DataFlow::Configuration flow and type-tracking flow.

[Evaluation](https://github.com/github/codeql-dca-main/tree/data/asgerf/summarize-more__nightly__code-scanning__8/reports) shows mostly neutral performance and around 21,000 new call edges.

I triaged some of the changes in an earlier evaluation, and the only spurious call edge I could find was the [call to delete!](https://github.com/github/codeql-dca-main/blob/data/asgerf/summarize-more__nightly__code-scanning__6/reports/alert-meta-comparison.md#-rbmetacall-graph-for-fastlane__fastlane-7) which I believe is ultimately because a `yield` call resolved to a callback coming from a different caller. (In JS, plain callbacks are not part of the static call graph and are handled specially in type-tracking - we could do the same in Ruby but it doesn't seem urgent.)

There’s still some work to be done, but given the performance-related commits in here, I’d prefer to merge sooner rather than later, so we don’t end up redoing or undoing each other’s performance work.

_Edit: updated text to reflect final state of PR_


asgerf:
Rebased to resolve conflict with https://github.com/github/codeql/pull/10664. Started another evaluation.

asgerf:
[New evaluation](https://github.com/github/codeql-dca-main/issues/7545)

ruby/ql/lib/codeql/ruby/typetracking/TypeTracker.qll
```diff
@@ -14,6 +14,11 @@ private module Cached {
...
     StoreStep(TypeTrackerContent content) { basicStoreStep(_, _, content) } or
     LoadStep(TypeTrackerContent content) { basicLoadStep(_, _, content) } or
+    LoadStoreStep(TypeTrackerContent load, TypeTrackerContent store) {
+      basicLoadStoreStep(_, _, load, store)
+    } or
+    WithContent(ContentFilter filter) { basicWithContentStep(_, _, filter) } or
+    WithoutContent(ContentFilter filter) { basicWithoutContentStep(_, _, filter) } or
     JumpStep()
 
   cached
```
hvitved, line 24: Add
```
or
basicLoadStoreStep(_, _, loadContents, _)
```
to `MkTypeTracker`, and
```
or
basicLoadStoreStep(_, _, _, storeContent)
```
to `MkTypeBackTracker`.

hvitved (commented):
Initial review comments. We should be able to revert the Hash changes once https://github.com/github/codeql/pull/10686 is in.

ruby/ql/lib/codeql/ruby/frameworks/core/Hash.qll
```diff
@@ -40,7 +42,7 @@ module Hash {
    * Gets a call to the method `name` invoked on the `Hash` object
    * (not on a hash instance).
    */
-  private MethodCall getAStaticHashCall(string name) {
+  MethodCall getAStaticHashCall(string name) {
```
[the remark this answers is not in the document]
asgerf: This was a left-over from [Ruby: move special treatment of Hash.[] into Hash.qll](https://github.com/github/codeql/pull/10650/commits/3ccc3a205808c2d206af7772d5795bd2ec94a92a). Will make it private again.

asgerf:
Thanks for the review! I'll work on getting type-tracking to rely on the `Argument[hash-splat]`-based summary instead, to replace the hand-written steps for `Hash.[]`.

ruby/ql/lib/codeql/ruby/dataflow/internal/DataFlowPrivate.qll
```diff
@@ -381,7 +381,7 @@ private module Cached {
     n instanceof SynthReturnNode
     or
     // Needed for stores in type tracking
-    TypeTrackerSpecific::postUpdateStoreStep(_, n, _)
+    TypeTrackerSpecific::storeStepIntoSourceNode(_, n, _)
```
hvitved, line 384: I belive we should remove
```
not n instanceof SynthHashSplatParameterNode
```
above as well.

ruby/ql/lib/codeql/ruby/frameworks/core/Hash.qll
```diff
@@ -1,6 +1,7 @@
 /** Provides flow summaries for the `Hash` class. */
 
 private import codeql.ruby.AST
+private import codeql.ruby.CFG as Cfg
```
hvitved, line 4: Remove this import.

asgerf:
[New evaluation](https://github.com/github/codeql-dca-main/tree/data/asgerf/summarize-more__nightly__code-scanning__8/reports) looks... unexpectedly good. We now gain 21,000 call edges at a very modest cost. Previously we only gained 2k call edges. I'll take some time to investigate exactly how the recent changes caused this change.

hvitved:
> I had to specialise handling of the `Hash.[]` summary as it would otherwise generate hundreds of useless edges for each element in a hash literal (the number depends on the number of constants in the program). For example, on one project a literal `{ foo: bar }` would map to 648 load-store steps from the `foo: bar` pair into the hash literal itself. The specialised version simply adds a store step directly from `bar` to the hash literal. I’ve added hooks so the specialisation happens in the model itself, not as a special case in the type-tracking library.

You may want to remove this from the description now.

ruby/ql/lib/codeql/ruby/typetracking/TypeTrackerSpecific.qll
```diff
@@ -264,3 +465,195 @@ private DataFlowPublic::Node evaluateSummaryComponentLocal(
...
+/**
+ * A data flow edge that should be followed by type tracking.
+ *
+ * This type of edge does not affect the local data flow graph, and is not used by data-flow configurations.
+ *
+ * Note: For performance reasons, all subclasses of this class should be part
+ * of the standard library, and their implementations may not depend on API graphs.
+ * For query-specific steps, consider including the custom steps in the type-tracking predicate itself.
+ */
+class TypeTrackingStep extends TUnit {
```
hvitved, line 559: I am not convinced that we should add this, especially as it is not used presently.

ruby/ql/lib/codeql/ruby/typetracking/TypeTrackerSpecific.qll
```diff
@@ -168,24 +209,27 @@ predicate returnStep(Node nodeFrom, Node nodeTo) {
...
+    SummaryComponentStack output
   |
-    hasStoreSummary(callable, contents, input, output) and
+    hasStoreSummary(callable, contents, pragma[only_bind_into](input),
+      pragma[only_bind_into](output)) and
     call.asExpr().getExpr() = callable.getACallSimple() and
-    nodeFrom = evaluateSummaryComponentLocal(call, input) and
-    nodeTo = evaluateSummaryComponentLocal(call, output)
+    nodeFrom = evaluateSummaryComponentStackLocal(callable, call, input) and
+    nodeTo = evaluateSummaryComponentStackLocal(callable, call, output)
```
hvitved, line 222: For a store step induced from a flow summary
```ql
input = Argument[0] and
output = Argument[1].Element[?]
```
and a call to the relevant method
```rb
foo(x, y)
```
doesn't this yield a store-step `x -store-> y`, while it should have been `x -store-> [post] y`?

The same for read-store steps.

ruby/ql/lib/codeql/ruby/typetracking/TypeTrackerSpecific.qll
```diff
@@ -230,31 +334,128 @@ class Boolean extends boolean {
...
+    or
+    content.isElementLowerBoundOrUnknown(_)
+    or
+    content.isSingleton(any(DataFlow::Content::UnknownElementContent c))
+  ) and
+  result = MkElementFilter()
+}
+
+pragma[nomagic]
+private predicate hasWithoutContentSummary(
```
hvitved, line 392: Does this work with compound `WithoutContent`s (see https://github.com/github/codeql/pull/10691)?

ruby/ql/lib/codeql/ruby/typetracking/TypeTrackerSpecific.qll
```diff
@@ -230,31 +334,128 @@ class Boolean extends boolean {
...
+    not TypeTrackingStep::suppressSummary(callable) and
+    callable.propagatesFlow(push(SummaryComponent::withoutContent(content), input), output, true) and
+    filter = getFilterFromWithoutContentStep(content) and
+    not isNonLocal(input.head()) and
+    not isNonLocal(output.head()) and
+    input != output
+  )
+}
+
+/**
```
hvitved, line 406: Copy-paste error (mentions `WithoutContent`.

ruby/ql/lib/codeql/ruby/typetracking/TypeTrackerSpecific.qll
```diff
@@ -230,31 +334,128 @@ class Boolean extends boolean {
...
+    or
+    content.isElementLowerBoundOrUnknown(_)
+    or
+    content.isSingleton(any(DataFlow::Content::ElementContent c))
+  ) and
+  result = MkElementFilter()
+}
+
+pragma[nomagic]
+private predicate hasWithContentSummary(
```
hvitved, line 428: Same question as for `hasWithoutContentSummary`.

ruby/ql/lib/codeql/ruby/typetracking/TypeTrackerSpecific.qll
```diff
@@ -230,31 +334,128 @@ class Boolean extends boolean {
...
+  SummarizedCallable callable, ContentFilter filter, SummaryComponentStack input,
+  SummaryComponentStack output
+) {
+  exists(DataFlow::ContentSet content |
+    not TypeTrackingStep::suppressSummary(callable) and
+    callable.propagatesFlow(push(SummaryComponent::withoutContent(content), input), output, true) and
+    filter = getFilterFromWithoutContentStep(content) and
+    not isNonLocal(input.head()) and
+    not isNonLocal(output.head()) and
+    input != output
```
hvitved, line 402: This restriction is super important, as otherwise use-use flow will ignore it (that's what `prohibitsUseUseFlow` is there for in data flow). Perhaps highlight that with a comment.

hvitved (commented):
Overall this LGTM. Some comments.

ruby/ql/lib/codeql/ruby/typetracking/TypeTrackerSpecific.qll
asgerf: If at most one of those `WithoutContent` has an associated `ContentFilter` then yes, because the rest are ignored and treated as ordinary flow (in `evaluateSummaryComponentStackLocal`).

ruby/ql/lib/codeql/ruby/typetracking/TypeTrackerSpecific.qll
asgerf: No, doesn't handle compound `WithContent`.

ruby/ql/lib/codeql/ruby/typetracking/TypeTrackerSpecific.qll
asgerf: Type-tracking doesn't depend on `prohibitsUseUseFlow`. It's simply here to clean up some steps that would have no effect for type-tracking.

hvitved (approved):

asgerf:
It seems the large number of extra call edges is a result of merging with @aibaars's PR https://github.com/github/codeql/pull/10559. The added summaries can now be used by type-tracking to great effect. 💪

PR #10663, opened 2022-10-03
Javascript: Improve Restify support and add new Spife support

pwntester:
This PR improves Restify support and adds brand new support for Spife framework


javascript/ql/lib/semmle/javascript/frameworks/Restify.qll
```diff
@@ -157,13 +166,182 @@ module Restify {
...
+          // server.use([fun1, fun2])
+          this.getMethodName() = ["use", "pre"] and
+          arg = this.getAnArgument()
+        ) and
+        (
+          not arg.asExpr() instanceof ArrayExpr and
+          result.flowsTo(arg)
+          or
+          arg.asExpr() instanceof ArrayExpr and
+          result.flowsTo(DataFlow::exprNode(arg.asExpr().(ArrayExpr).getAnElement()))
```
erik-krogh, line 198:
```suggestion
          not arg.getALocalSource() instanceof DataFlow::ArrayCreationNode and
          result.flowsTo(arg)
          or
          result = arg.getALocalSource().(DataFlow::ArrayCreationNode).getAnElement()
```

javascript/ql/lib/semmle/javascript/frameworks/Restify.qll
```diff
@@ -157,13 +166,182 @@ module Restify {
...
+   */
+  class FormatterSetup extends DataFlow::CallNode {
+    DataFlow::ObjectLiteralNode formatters;
+
+    FormatterSetup() {
+      // `server = restify.createServer()`
+      this = DataFlow::moduleMember("restify", "createServer").getACall() and
+      exists(DataFlow::ObjectLiteralNode options |
+        options.hasPropertyWrite("formatters", formatters)
+      )
```
erik-krogh, line 217: I think you might be missing an `options = this.getArgument(..)` or similar. 
You are building a cartesian product between all "formatters" and all calls to `createServer`.

erik-krogh (commented):
two quick comments from a quick read of the code.

javascript/ql/lib/semmle/javascript/frameworks/Restify.qll
pwntester: Thanks, I had to change `result = arg.getALocalSource().(DataFlow::ArrayCreationNode).getAnElement()` to `result.flowsTo(arg.getALocalSource().(DataFlow::ArrayCreationNode).getAnElement())` to make it work, but that account for cases where the array is created in a different statement. Thanks a lot! Didnt know about `getLocalSource`

javascript/ql/lib/semmle/javascript/frameworks/Restify.qll
pwntester: You are right, thanks! Also added `getALocalSource` approach that you suggested above

javascript/ql/lib/semmle/javascript/frameworks/Restify.qll
```diff
@@ -19,23 +19,24 @@ module Restify {
...
      * Gets the parameter of the route handler that contains the response object.
      */
-    Parameter getResponseParameter() { result = function.getParameter(1) }
+    Parameter getResponseParameter() { result = this.asExpr().(Function).getParameter(1) }
+  }
+
+  /**
+   * A standard Restify route handler.
+   */
+  abstract class StandardRouteHandler extends Http::Servers::StandardRouteHandler,
```
pwntester, line 37: Thanks! Just out of curiosity, why is it necessary to split the routeHandler into these two?

erik-krogh:
The qldoc check is failing because you don't have a toplevel QLDoc in `Spife.qll`.  

I think you should add a change-note to `javascript/ql/lib/change-notes`.

asgerf (commented):
Looks great, just a few things I noticed from a first read through. I haven't done a detailed review and I'm not very familiar with these libraries so will need more time to digest this.

erik-krogh:
You have a few failing tests:  
`javascript/ql/test/library-tests/frameworks/restify/tests.ql`, `javascript/ql/test/library-tests/frameworks/HTTP/RemoteRequestInput.ql`, `javascript/ql/test/library-tests/frameworks/NodeJSLib/tests.ql`.

javascript/ql/lib/semmle/javascript/frameworks/Spife.qll
```diff
@@ -0,0 +1,440 @@
...
+
+  /**
+   * A call to `reply.template('template', { ... })`, seen as a template instantiation.
+   */
+  private class TemplateCall extends Templating::TemplateInstantiation::Range, DataFlow::CallNode {
+    TemplateCall() {
+      exists(ReplySource reply |
+        reply.ref().(DataFlow::MethodCallNode).getMethodName() = "template" and
+        this = reply.ref()
+      )
```
erik-krogh, line 431: Same as above.

erik-krogh (commented):
A lot of minor comments after a readthrough.  

My suggestions will definitely cause the autoformatter to fail, so you need to run that.

javascript/ql/test/library-tests/frameworks/Spife/tests.ql
```diff
@@ -33,6 +33,15 @@ query predicate passingPositiveTests(string res, string expectation, InlineTest
   res = "PASSED" and
   t.hasPositiveTest(expectation) and
   (
+    expectation = "headerDefinition" and
```
pwntester, line 36: @erik-krogh Im curious, what would be the best way to combine these 4 query predicates so I dont need to repeat each test item?

erik-krogh: Not sure what the best is.  
You could have a predicate `hasResult(File f, int line, string expectation)`.   
Then you could have a single `query predicate()` where you checked that they match up with the `InlineTest`.  

Something like (untested):  
```CodeQL
query predicate test(string expectation, string msg) {
  exists(InlineTest t | 
    t.hasPositiveTest(expectation) and 
    hasResult(t.getFile(), t.getLocation().getStartLine(), expectation) and
    result = "PASS. TP"
  )
  or
  exists(InlineTest t | 
    t.hasNegativeTest(expectation) and 
    hasResult(t.getFile(), t.getLocation().getStartLine(), expectation) and
    result = "FAIL FP"
  )
  // .... 
}
```

pwntester: Thanks, is there a way to check the subtype of a `DataFlow::Node` given its type name? eg:

```
test("handler", "Http::RouteHandler") or
test("source", "RemoteFlowSource") or
...
```

So that in `test` we do something like:

```
predicate test(expectation, nodeType) {
     ...
     exists(DataFlow::Node n | n instanceof nodeType
     ...
}
```

I tried `exists(DataFlow::Node n | n.getAQlClass() = nodeType ...)` but that returns specfic types (eg `StandardRouteHandler`) instead of the abstract class (eg: `Http::RouteHandler`)?

erik-krogh: > Thanks, is there a way to check the subtype of a `DataFlow::Node` given its type name?

No, unfortunately not. 
And `getAQlClass()` is only suitable for debugging, it's not suitable for a test like this (the results might change for unrelated reasons).

pwntester: Thanks! I will keep the current version then

erik-krogh:
The qldoc check is complaining about missing documentation on `Restify::Restify::FormatterSetup::getAFormatterHandler`.

javascript/ql/lib/semmle/javascript/frameworks/Spife.qll
```diff
@@ -0,0 +1,416 @@
...
+      exists(DataFlow::MethodCallNode call |
+        this.ref() = call and
+        call.getMethodName() = "header" and
+        call.getNumArgument() = 3
+      )
+    }
+
+    override predicate definesHeaderValue(string headerName, DataFlow::Node headerValue) {
+      // reply.header(RESPONSE, 'Cache-Control', 'no-cache')
+      headerName = this.getNameNode().getStringValue() and
```
erik-krogh, line 264:
```suggestion
      this.getNameNode().mayHaveStringValue(headerName) and
```

javascript/ql/lib/semmle/javascript/frameworks/Spife.qll
```diff
@@ -0,0 +1,416 @@
...
+  private class MultipleHeaderDefinitions extends Http::ExplicitHeaderDefinition, DataFlow::CallNode {
+    ReplySource reply;
+
+    MultipleHeaderDefinitions() {
+      // reply.header(RESPONSE, {'Cache-Control': 'no-cache'})
+      // reply(RESPONSE, {'Cache-Control': 'no-cache'})
+      reply.ref().(DataFlow::CallNode).getCalleeName() = ["header", "reply"] and
+      reply.ref().(DataFlow::CallNode).getAnArgument().getALocalSource() instanceof
+        DataFlow::ObjectLiteralNode and
+      this = reply
```
erik-krogh, line 285: Multiple calls to `.ref()`, each call might not have the same value.  
Also, you won't flag `reply.header(..)` with this code.  

Also, it's the call that defines a header, not the reply.  

```suggestion
      exists(DataFlow::CallNode call | call = [reply.ref(), reply.ref().getAMethodCall("header")] | 
        call.getAnArgument().getALocalSource() instanceof DataFlow::ObjectLiteralNode and
        this = call
      )
```

(This will definitely cause the autoformatter to fail).  

Additionally. You could "save" the object-literal in a field, and then use that field in the member predicates in this class.  (Specifically `getAHeaderSource()`).

javascript/ql/lib/semmle/javascript/frameworks/Spife.qll
```diff
@@ -0,0 +1,416 @@
...
+    override Http::RouteHandler getRouteHandler() { result = this }
+  }
+
+  /**
+   * An HTTP cookie defined in a Spife HTTP response.
+   */
+  private class CookieDefinition extends Http::CookieDefinition, DataFlow::MethodCallNode instanceof ReplySource {
+    CookieDefinition() {
+      // reply.cookie(RESPONSE, 'TEST', 'FOO', {"maxAge": 1000, "httpOnly": true, "secure": true})
+      this.ref().(DataFlow::MethodCallNode).getMethodName() = "cookie"
```
erik-krogh, line 327: That won't work.  

You want a method called on a reply, not a reply that is a method-call. 
Also, the reply isn't a definition of a cookie, the call to `cookie` is.  

```suggestion
  private class CookieDefinition extends Http::CookieDefinition, DataFlow::MethodCallNode {
    CookieDefinition() {
      // reply.cookie(RESPONSE, 'TEST', 'FOO', {"maxAge": 1000, "httpOnly": true, "secure": true})
      this = any(ReplySource r).ref().getAMethodCall("cookie")
```

javascript/ql/lib/semmle/javascript/frameworks/Spife.qll
```diff
@@ -0,0 +1,416 @@
...
+   * A response argument passed to the `reply` method.
+   */
+  private class ReplyArgument extends Http::ResponseSendArgument, DataFlow::Node {
+    RouteHandler rh;
+
+    ReplyArgument() {
+      exists(ReplySource reply, DataFlow::CallNode call |
+        reply.ref() = call and
+        call.getCalleeName() =
+          ["reply", "cookie", "link", "header", "headers", "raw", "status", "toStream", "vary"] and
```
erik-krogh, line 347: I think this is the same problem as some of the above.  

Here you are asking for a reference to `reply` that is a call to `cookie`, `link`, etc.   
What you want is a method called on a reference to `reply` named `cookie`, `link`, etc.

erik-krogh: You can try to delete `, "cookie", "link", "header", "headers", "raw", "status", "toStream", "vary"`, and see that all tests still pass (I think that should happen).

javascript/ql/lib/semmle/javascript/frameworks/Spife.qll
```diff
@@ -0,0 +1,416 @@
...
+  /**
+   * An object passed to the `template` method of the reply object.
+   */
+  private class TemplateObjectInput extends DataFlow::Node {
+    ReplySource reply;
+
+    TemplateObjectInput() {
+      exists(DataFlow::MethodCallNode call |
+        reply.ref() = call and
+        call.getMethodName() = "template" and
```
erik-krogh, line 381: The same as above. Use `getAMethodCall()`.  

You are not testing this class, otherwise you would have found this.  
I haven't looked into why that happens.

javascript/ql/lib/semmle/javascript/frameworks/Spife.qll
```diff
@@ -0,0 +1,416 @@
...
+     * Gets the route handler that uses this object.
+     */
+    RouteHandler getRouteHandler() { result = reply.getRouteHandler() }
+  }
+
+  /**
+   * An invocation of the `redirect` method of an HTTP response object.
+   */
+  private class RedirectInvocation extends Http::RedirectInvocation, DataFlow::MethodCallNode instanceof ReplySource {
+    RedirectInvocation() { this.ref().(DataFlow::MethodCallNode).getMethodName() = "redirect" }
```
erik-krogh, line 396: Same as above.

javascript/ql/lib/semmle/javascript/frameworks/Spife.qll
```diff
@@ -0,0 +1,416 @@
...
+
+    override RouteHandler getRouteHandler() { result = this.getRouteHandler() }
+  }
+
+  /**
+   * A call to `reply.template('template', { ... })`, seen as a template instantiation.
+   */
+  private class TemplateCall extends Templating::TemplateInstantiation::Range,
+    DataFlow::MethodCallNode instanceof ReplySource {
+    TemplateCall() { this.getMethodName() = "template" }
```
erik-krogh, line 408: Same as above. But I think you are also missing a `.ref()` call.

erik-krogh (commented):
next round of comments, we are getting closer.

javascript/ql/lib/semmle/javascript/frameworks/Spife.qll
```diff
@@ -0,0 +1,416 @@
...
+module Spife {
+  /**
+   * An API graph entry point ensuring all tagged template exprs are part of the API graph
+   */
+  private class TaggedTemplateEntryPoint extends API::EntryPoint {
+    TaggedTemplateEntryPoint() { this = "TaggedTemplateEntryPoint" }
+
+    override DataFlow::SourceNode getASource() { result.asExpr() instanceof TaggedTemplateExpr }
+  }
+
```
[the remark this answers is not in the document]
pwntester: This was suggested by @asgerf so CCing him to double check

javascript/ql/lib/semmle/javascript/frameworks/Spife.qll
pwntester: The problem is that you can use the reply variable directly or a method on it:

```
const reply = require("...")

reply(RESPONSE, {'Cache-Control': 'no-cache'})
reply.header(RESPONSE, {'Cache-Control': 'no-cache'})
```

So I modelled `ReplySource` as a `DataFlow::CallNode` and `reply.ref()` should be a `CallNode` too. Not sure if this is the best way to model it but it worked.

> What you want is a method called on a reference to reply named cookie, link, etc.

That is true for the `reply.header(RESPONSE, {'Cache-Control': 'no-cache'})` case, but not for the `reply(RESPONSE, {'Cache-Control': 'no-cache'})` one, right?

pwntester: I ended up using the same approach that you proposed for the `MultipleHeaderDefinitions` predicate:

```ql
    ReplyArgument() {
      exists(ReplySource reply, DataFlow::CallNode call |
        call =
          [
            // reply(RESPONSE, {'Cache-Control': 'no-cache'})
            reply.ref(),
            // reply.header(RESPONSE, {'Cache-Control': 'no-cache'})
            reply
                .ref()
                .getAMethodCall([
                    "cookie", "link", "header", "headers", "raw", "status", "toStream", "vary"
                  ])
          ]
      |
        this = call.getArgument(0) and
        rh = reply.getRouteHandler()
      )
      or
      this = rh.getAReturn()
    }
```

javascript/ql/lib/semmle/javascript/frameworks/Spife.qll
pwntester: I think I can simplify this one by using the `TemplateCall` defined below and just get the arg 0

javascript/ql/lib/semmle/javascript/frameworks/Spife.qll
```diff
@@ -218,29 +208,25 @@ module Spife {
...
       this = API::moduleImport(["@npm/spife/reply", "spife/reply"]).getAMember().getACall()
     }
 
-    private DataFlow::SourceNode reachesHandlerReturn(DataFlow::TypeTracker t) {
-      result = this and
-      t.start()
-      or
-      exists(DataFlow::TypeTracker t2 | result = this.reachesHandlerReturn(t2).track(t2, t))
+    predicate isDirectReplyCall() {
+      this = API::moduleImport(["@npm/spife/reply", "spife/reply"]).getACall()
```
erik-krogh, line 221: This is duplication from the charpred, you could save whether it's a direct call as a field?

javascript/ql/lib/semmle/javascript/frameworks/Spife.qll
```diff
@@ -218,29 +208,25 @@ module Spife {
...
 
     /**
      * Gets the route handler that provides this response.
      */
-    override RouteHandler getRouteHandler() {
+    RouteHandler getRouteHandler() {
       exists(RouteHandler handler |
-        handler.(DataFlow::FunctionNode).getAReturn().getALocalSource() =
-          this.reachesHandlerReturn(DataFlow::TypeTracker::end()) and
+        handler.getAReturn() = this.getReturn().getAValueReachableFromSource() and
```
erik-krogh, line 229: You don't need the `exists(RouteHandler..`, you can just directly do: 
```CodeQL
result.getAReturn() = this.getReturn().getAValueReachableFromSource()
```

javascript/ql/lib/semmle/javascript/frameworks/Spife.qll
```diff
@@ -249,47 +235,46 @@ module Spife {
...
+      this.getCalleeName() = "header" and
+      this.getNumArgument() = 3
     }
 
     override predicate definesHeaderValue(string headerName, DataFlow::Node headerValue) {
       // reply.header(RESPONSE, 'Cache-Control', 'no-cache')
-      headerName = this.getNameNode().getStringValue() and
-      headerValue = this.getArgument(2)
+      this.getNameNode().mayHaveStringValue(headerName) and
+      headerValue = this.(DataFlow::MethodCallNode).getArgument(2)
```
erik-krogh, line 248:
```suggestion
      headerValue = super.getArgument(2)
```

javascript/ql/lib/semmle/javascript/frameworks/Spife.qll
```diff
@@ -249,47 +235,46 @@ module Spife {
...
     }
 
-    override DataFlow::Node getNameNode() { result = this.getArgument(1) }
+    override DataFlow::Node getNameNode() {
+      result = this.(DataFlow::MethodCallNode).getArgument(1)
```
erik-krogh, line 252:
```suggestion
      result = super.getArgument(1)
```

javascript/ql/lib/semmle/javascript/frameworks/Spife.qll
```diff
@@ -249,47 +235,46 @@ module Spife {
...
+    }
 
-    override RouteHandler getRouteHandler() { result = this.getRouteHandler() }
+    override RouteHandler getRouteHandler() { result = this.(ReplyCall).getRouteHandler() }
```
erik-krogh, line 255:
```suggestion
    override RouteHandler getRouteHandler() { result = ReplyCall.super.getRouteHandler() }
```

javascript/ql/lib/semmle/javascript/frameworks/Spife.qll
```diff
@@ -249,47 +235,46 @@ module Spife {
...
+      this.getAnArgument().getALocalSource() instanceof DataFlow::ObjectLiteralNode
     }
 
     /**
      * Gets a reference to the multiple headers object that is to be set.
      */
-    private DataFlow::ObjectLiteralNode getAHeaderSource() {
-      result = this.getAnArgument().getALocalSource()
+    DataFlow::ObjectLiteralNode getAHeaderSource() {
+      result = this.(DataFlow::CallNode).getAnArgument().getALocalSource()
```
erik-krogh, line 277:
```suggestion
      result = super.getAnArgument().getALocalSource()
```

javascript/ql/lib/semmle/javascript/frameworks/Spife.qll
```diff
@@ -303,114 +288,128 @@ module Spife {
       result = this.getAHeaderSource().getAPropertyWrite().getPropertyNameExpr().flow()
     }
 
-    override RouteHandler getRouteHandler() { result = reply.getRouteHandler() }
+    override RouteHandler getRouteHandler() { result = this.(ReplyCall).getRouteHandler() }
```
erik-krogh, line 291:
```suggestion
    override RouteHandler getRouteHandler() { result = super.getRouteHandler() }
```

javascript/ql/lib/semmle/javascript/frameworks/Spife.qll
```diff
@@ -303,114 +288,128 @@ module Spife {
...
+  private class CookieDefinition extends Http::CookieDefinition instanceof ReplyCall {
     CookieDefinition() {
       // reply.cookie(RESPONSE, 'TEST', 'FOO', {"maxAge": 1000, "httpOnly": true, "secure": true})
-      this.ref().(DataFlow::MethodCallNode).getMethodName() = "cookie"
+      this.getCalleeName() = "cookie"
     }
 
-    override DataFlow::Node getNameArgument() { result = this.getArgument(1) }
+    //  this = any(ReplyCall r).ref().getAMethodCall("cookie")
+    override DataFlow::Node getNameArgument() { result = this.(ReplyCall).getArgument(1) }
```
erik-krogh, line 315:
```suggestion
    override DataFlow::Node getNameArgument() { result = super.getArgument(1) }
```

javascript/ql/lib/semmle/javascript/frameworks/Spife.qll
```diff
@@ -303,114 +288,128 @@ module Spife {
...
 
-    override DataFlow::Node getValueArgument() { result = this.getArgument(2) }
+    override DataFlow::Node getValueArgument() { result = this.(ReplyCall).getArgument(2) }
```
erik-krogh, line 317:
```suggestion
    override DataFlow::Node getValueArgument() { result = super.getArgument(2) }
```

javascript/ql/lib/semmle/javascript/frameworks/Spife.qll
```diff
@@ -303,114 +288,128 @@ module Spife {
...
 
-    override RouteHandler getRouteHandler() { result = this.getRouteHandler() }
+    override RouteHandler getRouteHandler() { result = this.(ReplyCall).getRouteHandler() }
```
erik-krogh, line 319:
```suggestion
    override RouteHandler getRouteHandler() { result = super.getRouteHandler() }
```

javascript/ql/lib/semmle/javascript/frameworks/Spife.qll
```diff
@@ -303,114 +288,128 @@ module Spife {
...
+  private class TemplateCall extends Templating::TemplateInstantiation::Range instanceof ReplyCall {
+    TemplateCall() { this.getCalleeName() = "template" }
 
-    TemplateInput() {
-      obj.getALocalSource().(DataFlow::ObjectLiteralNode).hasPropertyWrite(_, this)
-    }
+    override DataFlow::SourceNode getOutput() { result = this }
 
-    override RouteHandler getRouteHandler() { result = obj.getRouteHandler() }
+    override DataFlow::Node getTemplateFileNode() { result = this.(ReplyCall).getArgument(0) }
```
erik-krogh, line 374:
```suggestion
    override DataFlow::Node getTemplateFileNode() { result = super.getArgument(0) }
```

javascript/ql/lib/semmle/javascript/frameworks/Spife.qll
```diff
@@ -303,114 +288,128 @@ module Spife {
...
+
+    override DataFlow::Node getTemplateParamsNode() { result = this.(ReplyCall).getArgument(1) }
```
erik-krogh, line 376:
```suggestion
    override DataFlow::Node getTemplateParamsNode() { result = super.getArgument(1) }
```

javascript/ql/lib/semmle/javascript/frameworks/Spife.qll
```diff
@@ -303,114 +288,128 @@ module Spife {
...
   }
 
   /**
    * An object passed to the `template` method of the reply object.
    */
   private class TemplateObjectInput extends DataFlow::Node {
-    ReplySource reply;
+    TemplateCall call;
```
erik-krogh, line 383: If you change the type to `ReplyCall`, then you can make the char-pred (and the member predicates) simpler.  

```suggestion
    ReplyCall call;
```

javascript/ql/lib/semmle/javascript/frameworks/Spife.qll
```diff
@@ -303,114 +288,128 @@ module Spife {
...
 
-    TemplateObjectInput() {
-      exists(DataFlow::MethodCallNode call |
-        reply.ref() = call and
-        call.getMethodName() = "template" and
-        this = call.getArgument(1)
-      )
-    }
+    TemplateObjectInput() { this = call.(ReplyCall).getArgument(1) }
```
erik-krogh, line 385: (Works with the above change)

```suggestion
    TemplateObjectInput() { this = call.getArgument(1) }
```

javascript/ql/lib/semmle/javascript/frameworks/Spife.qll
```diff
@@ -303,114 +288,128 @@ module Spife {
...
 
     /**
      * Gets the route handler that uses this object.
      */
-    RouteHandler getRouteHandler() { result = reply.getRouteHandler() }
+    RouteHandler getRouteHandler() { result = call.(ReplyCall).getRouteHandler() }
```
erik-krogh, line 390:
```suggestion
    RouteHandler getRouteHandler() { result = call.getRouteHandler() }
```

javascript/ql/lib/semmle/javascript/frameworks/Spife.qll
```diff
@@ -303,114 +288,128 @@ module Spife {
...
+  private class RedirectInvocation extends Http::RedirectInvocation instanceof ReplyCall {
+    RedirectInvocation() { this.getCalleeName() = "redirect" }
 
-    override DataFlow::SourceNode getOutput() { result = this }
-
-    override DataFlow::Node getTemplateFileNode() { result = this.getArgument(0) }
+    override DataFlow::Node getUrlArgument() { result = this.getAnArgument() }
 
-    override DataFlow::Node getTemplateParamsNode() { result = this.getArgument(1) }
+    override RouteHandler getRouteHandler() { result = this.getRouteHandler() }
```
erik-krogh, line 413: You are defining `getRouteHandler` as `getRouteHandler`? 

I think you are missing a super.  
```suggestion
    override RouteHandler getRouteHandler() { result = super.getRouteHandler() }
```

erik-krogh (commented):
Lots of simplications (rewriting to use `super.`) from a quick read-through.

erik-krogh:
@pwntester did you see my above suggestions?

pwntester:
> @pwntester did you see my above suggestions?

Sorry I missed the notifications about the review comments. Will take a look ASAP

javascript/ql/lib/semmle/javascript/frameworks/Spife.qll
pwntester: I had to restore this cast to avoid this error `Predicate getRouteHandler/0 is ambiguous between: predicate CookieDefinition.HTTP::Http::CookieDefinition::getRouteHandler() from HTTP.qll:153, predicate ReplyCall.Spife::Spife::ReplyCall::getRouteHandler() from Spife.qll:229. You can resolve the ambiguity by qualifing super with the desired type.`

pwntester:
Thanks @erik-krogh I applied all suggestions, but had to revert 3 of them since there raised some ambiguity errors between parent types

javascript/ql/lib/semmle/javascript/frameworks/Spife.qll
erik-krogh: You can do `ReplyCall.super.getRouteHandler()` in that case.

erik-krogh: (And similarly for the others).

pwntester: Thanks @erik-krogh, I was not aware of that syntax. What is the difference between the two?

erik-krogh: When a class has multiple super-classes there might be multiple incompatible `getRouteHandler()` predicates in the class hierarchy.  
With `ReplyCall.super` you specify which super-classes to look for the predicate in.

javascript/ql/lib/semmle/javascript/frameworks/Spife.qll
```diff
@@ -0,0 +1,412 @@
...
+      this.getCalleeNode().asExpr() = template and
+      API::moduleImport(["@npm/spife/routing", "spife/routing"])
+          .asSource()
+          .flowsToExpr(template.getTag())
+    }
+
+    private string getRoutePattern() {
+      // Concatenate the constant parts of the expression
+      result =
+        concat(Expr e, int i |
```
erik-krogh, line 28:
```suggestion
        strictconcat(Expr e, int i |
```

If you just use `concat`, then you risk that `result = ""`, and that doesn't seem to be the thing you want.  

It makes no difference in the end (`hasLine` has the same result regardless), it's just a little nice-to-have.

javascript/ql/lib/semmle/javascript/frameworks/Spife.qll
```diff
@@ -0,0 +1,412 @@
...
+import semmle.javascript.frameworks.HTTP
+
+/**
+ * Provides classes for working with [Spife](https://github.com/npm/spife) applications.
+ */
+module Spife {
+  /**
+   * A call to a Spife method that sets up a route.
+   */
+  private class RouteSetup extends API::CallNode, Http::Servers::StandardRouteSetup {
```
erik-krogh, line 15: This doesn't work, as there is no API node for the relevant call.  

In general there is not API nodes for everything, only the things you can reach using a chain of calls to the `API-graph` predicates.   

```suggestion
  private class RouteSetup extends DataFlow::CallNode, Http::Servers::StandardRouteSetup {
```

Some things will stop compiling, and you should add some inline tests to the `javascript/ql/test/library-tests/frameworks/Spife/lib/routes/index.js` file.  

To be clear, the use of `API::Node` in `ReplyCall` is fine.

javascript/ql/lib/semmle/javascript/frameworks/Spife.qll
```diff
@@ -0,0 +1,412 @@
...
+
+    private predicate hasLine(string method, string path, string handlerName) {
+      exists(string line | line = this.getARouteLine() |
+        line.splitAt(" ", 0) = method and
+        line.splitAt(" ", 1) = path and
+        line.splitAt(" ", 2) = handlerName
+      )
+    }
+
+    API::Node getHandlerByName(string name) { result = this.getParameter(0).getMember(name) }
```
erik-krogh, line 47: You also have to remove the rest of the uses of `API::Node` in this class.

javascript/ql/lib/semmle/javascript/frameworks/Spife.qll
```diff
@@ -0,0 +1,412 @@
...
+   * A standard Spife route handler.
+   */
+  private class StandardRouteHandler extends RouteHandler, DataFlow::FunctionNode {
+    StandardRouteHandler() { any(RouteSetup setup).getARouteHandler() = this }
+  }
+
+  /**
+   * A function that looks like a Spife route handler.
+   *
+   * For example, this could be the function `function(req, res, next){...}`.
```
erik-krogh, line 93: Documentation doesn't match the implementation.   

Your example has 3 parameters, but the charpred looks for exactly two parameters.  
Looks like the qldoc wasn't changed after copy-pasting.

javascript/ql/lib/semmle/javascript/frameworks/Spife.qll
```diff
@@ -0,0 +1,412 @@
...
+      this.(DataFlow::MethodCallNode).getMethodName() = ["cookie", "cookies"] and
+      kind = "cookie"
+      or
+      exists(DataFlow::PropRead validated, DataFlow::MethodCallNode get |
+        rh.getARequestSource().ref().getAPropertyRead() = validated and
+        validated.getPropertyName().matches("validated%") and
+        get.getReceiver() = validated and
+        this = get and
+        kind = "body"
+      )
```
erik-krogh, line 164: It's hard to read what this does, and I suspect there might be an issue here.  
But without some example code, that's hard to tell.   

Could you add a comment with an example of what this will match?

erik-krogh (commented):
Another round of comments.  

Could you also do a merge with `main`? your base-commit is getting old.

javascript/ql/lib/semmle/javascript/frameworks/Spife.qll
pwntester: thanks

javascript/ql/lib/semmle/javascript/frameworks/Spife.qll
pwntester: @erik-krogh how would you rewrite this predicate? I thought about this `result = this.getACallee().getParameter(0).(DataFlow::ObjectLiteralNode).getAPropertyRead(name)` but ObjectLiteralNode is not compatible with Parameter

javascript/ql/lib/semmle/javascript/frameworks/Spife.qll
pwntester: sure, will do

javascript/ql/lib/semmle/javascript/frameworks/Spife.qll
erik-krogh: Just remove `.(DataFlow::ObjectLiteralNode)` from your attempt and you should be good.  
`getAPropertyRead()` is defined on `SourceNode`, and both `ObjectLiteralNode` and `ParameterNode` are `SourceNode`s.

pwntester: Ok, I was not getting that option from the autocomplete menu, but I was missing a wrapping call to `DataFlow::parameterNode()`

erik-krogh (commented):
One more comment, and then I think we're there.  

I have done an evaluation, and that looks fine.

erik-krogh:
You fixed the implementation in https://github.com/github/codeql/pull/10663/commits/4ba3190d29f3b5273c38e8f93b59c42ee36ba480, but you didn't add any tests to match.  
I don't think you have any tests of `Spife::RouteSetup` (and the classes that depend on `Spife::RouteSetup`).   

Those tests can probably be added as inline test comments in `javascript/ql/test/library-tests/frameworks/Spife/lib/routes/index.js`, and that file should probably also be expanded.

pwntester:
Thought about adding some tests as you proposed but I thought it was already tested indirectly and we have plenty of tests for `handlers` and those can only be correctly identified if the RouteSetup was matched correctly. Will add some explicit tests though

javascript/ql/test/library-tests/frameworks/Spife/tests.ql
```diff
@@ -0,0 +1,254 @@
+import javascript
+import semmle.javascript.security.dataflow.CleartextStorageCustomizations
+import semmle.javascript.security.dataflow.CorsMisconfigurationForCredentialsCustomizations
+import semmle.javascript.security.dataflow.StackTraceExposureCustomizations
+import semmle.javascript.security.dataflow.ServerSideUrlRedirectCustomizations
+import semmle.javascript.security.dataflow.RequestForgeryCustomizations
+import semmle.javascript.security.dataflow.ReflectedXssCustomizations
+import semmle.javascript.security.dataflow.ReflectedXssQuery as XssConfig
+import semmle.javascript.heuristics.AdditionalRouteHandlers
+
```
erik-krogh, line 10:
```suggestion

```

This was the reason that your tests "worked" before [the fix](https://github.com/github/codeql/commit/4ba3190d29f3b5273c38e8f93b59c42ee36ba480).  

Your `Spife::RouteSetup` and `Spife::StandardRouteHandler` were effectively unused in your tests before [the fix](https://github.com/github/codeql/commit/4ba3190d29f3b5273c38e8f93b59c42ee36ba480), because your tests don't distinguish between heuristic and non-heuristic route-handlers.  
You have to distinguish between heuristic and non-heuristic, otherwise you're only testing the former. 

You can try to delete `Spife::StandardRouteHandler`, your tests still pass as they are right now.  

You have a bunch of test failures when you delete this line, and I see three possible solution to that:   
- Accept the new test output (including failures), to document that some things can only be detected using the heuristic route-handlers.  
- Delete the tests lines are only detected by the heuristic route-handlers.   
   This is fine, our other heuristic route-handlers don't really have tests.  
- Add another test that tests the heuristic route-handlers. 

(The above assumes that no other bug surfaces from this).  

Sorry for figuring this out so late, but I didn't look that closely at your test implementations.

javascript/ql/test/library-tests/frameworks/Restify2/tests.ql
```diff
@@ -0,0 +1,230 @@
+import javascript
+import semmle.javascript.security.dataflow.CleartextStorageCustomizations
+import semmle.javascript.security.dataflow.CorsMisconfigurationForCredentialsCustomizations
+import semmle.javascript.security.dataflow.StackTraceExposureCustomizations
+import semmle.javascript.security.dataflow.ServerSideUrlRedirectCustomizations
+import semmle.javascript.security.dataflow.RequestForgeryCustomizations
+import semmle.javascript.security.dataflow.ReflectedXssCustomizations
+import semmle.javascript.security.dataflow.ReflectedXssQuery as XssConfig
+import semmle.javascript.heuristics.AdditionalRouteHandlers
+
```
erik-krogh, line 10:
```suggestion

```

Same here.

javascript/ql/lib/semmle/javascript/frameworks/Spife.qll
```diff
@@ -53,20 +45,33 @@ module Spife {
       )
     }
 
-    API::Node getHandlerByName(string name) { result = this.getParameter(0).getMember(name) }
+    DataFlow::SourceNode getHandlerDefinitions(TypeBackTracker t) {
+      t.start() and
+      result = this.getArgument(0).getALocalSource()
+      or
+      exists(TypeBackTracker t2 | result = getHandlerDefinitions(t2).backtrack(t2, t))
```
erik-krogh, line 52:
```suggestion
      exists(TypeBackTracker t2 | result = this.getHandlerDefinitions(t2).backtrack(t2, t))
```

javascript/ql/lib/semmle/javascript/frameworks/Spife.qll
```diff
@@ -53,20 +45,33 @@ module Spife {
...
+    }
+
+    DataFlow::SourceNode getHandlerDefinitions() {
+      result = getHandlerDefinitions(TypeBackTracker::end())
```
erik-krogh, line 56:
```suggestion
      result = this.getHandlerDefinitions(TypeBackTracker::end())
```

javascript/ql/lib/semmle/javascript/frameworks/Spife.qll
```diff
@@ -53,20 +45,33 @@ module Spife {
...
+    }
+
+    DataFlow::SourceNode getHandlerByName(string name) {
+      result = getHandlerDefinitions().getAPropertySource(name)
```
erik-krogh, line 60:
```suggestion
      result = this.getHandlerDefinitions().getAPropertySource(name)
```

erik-krogh (approved):
Lets get this merged.   

I got a final confirmation evaluation running, but I'll do something if that doesn't turn out OK.

PR #10664, opened 2022-10-03
Ruby/Python: Cache more type tracking predicates

hvitved:
https://github.com/github/codeql/pull/10375 follow-up.

Prevents reevaluation of flow summaries for Ruby, since those are now taken into account in `basicLoadStep` and `basicStoreStep`.

### DIFF
--- a/python/ql/lib/semmle/python/dataflow/new/internal/TypeTracker.qll
+++ b/python/ql/lib/semmle/python/dataflow/new/internal/TypeTracker.qll
@@ -16,6 +16,32 @@ private module Cached {
     LoadStep(TypeTrackerContent content) { basicLoadStep(_, _, content) } or
     JumpStep()
 
+  cached
+  newtype TTypeTracker =
+    MkTypeTracker(Boolean hasCall, OptionalTypeTrackerContent content) {
+      content = noContent()
+      or
+      // Restrict `content` to those that might eventually match a load.
+      // We can't rely on `basicStoreStep` since `startInContent` might be used with
+      // a content that has no corresponding store.
+      exists(TypeTrackerContent loadContents |
+        basicLoadStep(_, _, loadContents) and
+        compatibleContents(content, loadContents)
+      )
+    }
+
+  cached
+  newtype TTypeBackTracker =
+    MkTypeBackTracker(Boolean hasReturn, OptionalTypeTrackerContent content) {
+      content = noContent()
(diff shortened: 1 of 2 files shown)


asgerf (approved):
👍

python/ql/lib/semmle/python/dataflow/new/internal/TypeTracker.qll
```diff
@@ -188,49 +245,9 @@ module StepSummary {
...
-   * ```
-   * for the attribute write `x.attr = y`, we will have `content` being the literal string `"attr"`,
-   * `nodeFrom` will be `y`, and `nodeTo` will be the object `Foo()` created on the first line of the
-   * function. This means we will track the fact that `x.attr` can have the type of `y` into the
-   * assignment to `z` inside `bar`, even though this attribute write happens _after_ `bar` is called.
-   */
-  predicate localSourceStoreStep(Node nodeFrom, TypeTrackingNode nodeTo, TypeTrackerContent content) {
-    exists(Node obj | nodeTo.flowsTo(obj) and basicStoreStep(nodeFrom, obj, content))
-  }
+  deprecated predicate localSourceStoreStep = flowsToStoreStep/3;
```
yoff, line 248: I am curious as to why you felt this name change necessary?

yoff (approved):
The minor regression for Python seems greatly outweighed by the gain for Ruby (even if Python probably represents a bigger corpus?). Especially as we hope to use flow summaries in Python in the future..

python/ql/lib/semmle/python/dataflow/new/internal/TypeTracker.qll
hvitved: `localSourceStoreStep` is only used in `smallstepNoCall`, which is now cached, so there is no need to cache this predicate as well. However, it was not a private predicate, so we need to keep it around in case others use it (unlikely).

yoff: Yes, I was more wondering why you changed the name to `flowsToStoreStep` :-)

hvitved: I had to come up with _some_ name, so I didn't really rename it. `localSourceStoreStep` still exists as a public deprecated predicate.

PR #10667, opened 2022-10-04
C#: Recognize options to `dotnet run` in tracer when injecting `-p:UseSharedCompilation=false`

hvitved:
Invocations like
```
dotnet run --no-build arg
```
would incorrectly be rewritten to
```
dotnet run -p:UseSharedCompilation=false -- --no-build arg
```

With this PR, they instead get rewritten to
```
dotnet run -p:UseSharedCompilation=false --no-build arg
```


michaelnebel (approved):
Oh right!
Command line options are complicated 😹 
LGTM!

PR #10668, opened 2022-10-04
Ruby: update dependencies

aibaars:
Update the lock files to include the `codeql/ssa` library and replace ad-hoc pack create command with a proper `codeql pack install` command now the ssa library has been published.

### DIFF
--- a/.github/workflows/ruby-build.yml
+++ b/.github/workflows/ruby-build.yml
@@ -96,8 +96,8 @@ jobs:
       - name: Build Query Pack
         run: |
           codeql pack create ../shared/ssa --output target/packs
+          codeql pack create ../misc/suite-helpers --output target/packs
           codeql pack create ql/lib --output target/packs
-          codeql pack install ql/src
           codeql pack create ql/src --output target/packs
           PACK_FOLDER=$(readlink -f target/packs/codeql/ruby-queries/*)
           codeql generate query-help --format=sarifv2.1.0 --output="${PACK_FOLDER}/rules.sarif" ql/src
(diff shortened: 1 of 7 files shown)


.github/workflows/ruby-build.yml
```diff
@@ -95,7 +95,7 @@ jobs:
         uses: ./.github/actions/fetch-codeql
       - name: Build Query Pack
         run: |
-          codeql pack create ../shared/ssa --output target/packs
+          codeql pack install ql/lib
```
hvitved, line 98: Will this still build the pack from the repo, or fetch it from the registry?

aibaars: I guess it fetches it from the registry.

hvitved: That doesn't work then. We do not guarantee, that the code in `main` is compatible with any released shared pack; only that it is compatible with the version of the pack at `main` as well.

aibaars: Ah ok. In that case we should remove the `codeql pack install ql/src` call too and use the suite-helpers from `../misc`

erik-krogh (dismissed):
LGTM - except for the lock files.  
There is no need for explicit dependency versions in the lock files, as all the dependencies are present locally.  
So I removed those and pushed that to this PR.  

Ruby and Go were the only languages that had any explicit dependencies in their lockfiles, which I suppose are leftovers from the days of living in a separate repo.  

/cc @aeisenberg: Am I correct in assuming that these lock files don't need anything in `dependencies:`?

.github/workflows/ruby-build.yml
```diff
@@ -202,13 +202,14 @@ jobs:
...
           dependencies:
-            codeql/ruby-all: 0.0.1
+            codeql/ruby-all: "*"
           extractor: ruby
           tests: .
           ' > qlpack.yml
       - name: Run QL test
         shell: bash
         run: |
+          codeql pack install .
```
aibaars, line 212: Let's see if we still need this:
```suggestion
```

erik-krogh (approved):

aeisenberg:
> /cc @aeisenberg: Am I correct in assuming that these lock files don't need anything in `dependencies:`?

That's right!

PR #10670, opened 2022-10-04
JS: Improve detection of XSS when JSON.stringify()

tyage:
`JSON.stringify` is sometimes used to inject the object information into HTML.
In these cases, many products disable HTML escape so that JSON string will not corrupted. Although it makes the apps potentially vulnerable to XSS...
https://cs.github.com/?scopeName=All+repos&scope=&q=dangerouslySetInnerHTML+JSON.stringify

CodeQL cannot detect those XSS because `JsonStringifyTaintStep` only support when the argument of `JSON.stringify()` is equal to another source.

I made the changes in `JsonStringifyTaintStep` so that we can detect XSS when the malicious input is located in the argument object property.

```javascript
  <script
    type="application/ld+json"
    dangerouslySetInnerHTML={{ __html:
      JSON.stringify({
        foo: req.query.bar
      })
    }}
  />;
```


asgerf:
Hi @tyage 👋 

Thanks for the contribution! I've taken a look and I generally like the idea, just need a bit more time to investigate performance. I'm hoping we can merge sometime next time.

asgerf:
Since this changes files not in the `experimental` folder, I've gone ahead and pushed some commits to the PR:
- Now uses the data-flow API instead of AST
- The recursive relation is now restricted to relevant tuples. Computing it for all `Expr`s creates unnecessary work.
- Edited the change note to focus more on the user-facing changes (for example, not mentioning the private class `JsonStringifyTaintStep`).

I've run an internal evaluation and it performs well and the results look reasonable.

erik-krogh (approved):

tyage:
thanks for the rewrite. I learned a lot!

PR #10671, opened 2022-10-04
Java: Revert #10489 and #10640

atorralba:
Reverts #10498, but keeps #10640.

The performance impact of `getATypeInScope` is too high (it produces a cartesian product of all `CompilationUnit`s with all types in `java.lang`), we will need to reconsider this feature. @Marcono1234 let us know if you can think of an alternative.

### DIFF
--- a/java/ql/lib/change-notes/2022-09-20-CompilationUnit-getATypeInScope.md
+++ b/java/ql/lib/change-notes/2022-09-20-CompilationUnit-getATypeInScope.md
@@ -1,4 +1,0 @@
----
-category: feature
----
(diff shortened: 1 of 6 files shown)


aschackmull (dismissed):

aschackmull (approved):

aschackmull:
> it produces a cartesian product of all `CompilationUnit`s with all types in `java.lang`

And even without that disjunct, the predicate is still too big.

Marcono1234:
Should #4119 be reopened again?
(I can't do that because I did not close it)

aschackmull:
> Should #4119 be reopened again?

done

PR #10673, opened 2022-10-04
Ruby: remove public abstract classes for Action{View,Controller}

nickrolfe:
Most of these were abstract because we provide separate implementations for calls in the context of ActionView and ActionController. In those cases, I've moved the public-facing class to `Rails.qll`, introduced `Impl` (range) classes for them in `internal/Rails.qll`, and kept the `Impl` subclasses as private classes in the appropriate `ActionView.qll` and `ActionController.qll` file.

I've also moved the `this.getMethodName() = ...` restrictions out of the base class charpreds and into the individual implementation charpreds. This adds a little more duplication, but I have a reason for doing it: when I add ActionMailer support, I'll want a call to `mail` to be considered a `RenderCall`, so `this.getMethodName() = "render"` in the base class would not work.

### DIFF
--- a/ruby/ql/lib/codeql/ruby/frameworks/ActionController.qll
+++ b/ruby/ql/lib/codeql/ruby/frameworks/ActionController.qll
@@ -8,8 +8,10 @@ private import codeql.ruby.controlflow.CfgNodes
 private import codeql.ruby.DataFlow
 private import codeql.ruby.dataflow.RemoteFlowSources
 private import codeql.ruby.ApiGraphs
-private import codeql.ruby.frameworks.ActionView
 private import codeql.ruby.frameworks.ActionDispatch
+private import codeql.ruby.frameworks.ActionView
+private import codeql.ruby.frameworks.Rails
+private import codeql.ruby.frameworks.internal.Rails
 
 /**
  * A `ClassDeclaration` for a class that extends `ActionController::Base`.
@@ -120,13 +122,6 @@ private class ActionControllerContextCall extends MethodCall {
 }
 
 /**
- * A call to the `params` method to fetch the request parameters.
- */
-abstract class ParamsCall extends MethodCall {
-  ParamsCall() { this.getMethodName() = "params" }
-}
-
-/**
  * A `RemoteFlowSource::Range` to represent accessing the
  * ActionController parameters available via the `params` method.
  */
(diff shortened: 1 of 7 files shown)


ruby/ql/lib/codeql/ruby/frameworks/ActionController.qll
```diff
@@ -119,13 +121,6 @@ private class ActionControllerContextCall extends MethodCall {
   ActionControllerControllerClass getControllerClass() { result = controllerClass }
 }
 
-/**
- * A call to the `params` method to fetch the request parameters.
- */
-abstract class ParamsCall extends MethodCall {
```
alexrford, line 125: Does moving this (and `CookiesCall`) technically break the deprecation policy as these are no longer exported from this library? Wondering if we should re-export these to avoid potential problems if a query happens to rely on `ActionController.qll` exporting these classes.

alexrford (commented):
Thanks for fixing this - just a question around the deprecation policy.

ruby/ql/lib/codeql/ruby/frameworks/ActionController.qll
nickrolfe: That's a good point. I confirmed it won't cause any problems with the internal query repo, but let me see if I can ex-export them.

nickrolfe: It seems that you can't just export a class as an alias under the same name without getting errors about conflicting definitions in files that import both libraries. So I think we have three options:

1. Leave it as is and break the deprecation policy.
2. Rename the classes that got moved to `Rails.qll` and export the old names as aliases in `ActionView.qll` and `ActionController.qll`.
3. Move them back to `ActionView.qll` and `ActionController.qll`.

nickrolfe: OK, I think I've found a nicer solution than any of those. The classes in `Rails.qll` can go inside a module named `Rails`, and then we can introduce deprecated aliases in `ActionView.qll` and `ActionController.qll` without causing conflicting definitions anywhere.

alexrford: Nice, I find it a bit neater having these enclosed in a separate module anyway, so this seems like a good solution.

alexrford (approved):
👍

PR #10676, opened 2022-10-04
RB: add a link to the source in the alert-message for `rb/kernel-open`

erik-krogh:
Just something I encountered while looking into the query.

### DIFF
--- a/ruby/ql/src/queries/security/cwe-078/KernelOpen.ql
+++ b/ruby/ql/src/queries/security/cwe-078/KernelOpen.ql
@@ -73,5 +73,5 @@ where
   sourceNode = source.getNode() and
   call.asExpr().getExpr().(MethodCall).getArgument(0) = sink.getNode().asExpr().getExpr()
 select sink.getNode(), source, sink,
-  "This call to " + call.(Replacement).getFrom() +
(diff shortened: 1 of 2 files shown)


ruby/ql/src/queries/security/cwe-078/KernelOpen.ql
```diff
@@ -73,5 +73,5 @@ where
   sourceNode = source.getNode() and
   call.asExpr().getExpr().(MethodCall).getArgument(0) = sink.getNode().asExpr().getExpr()
 select sink.getNode(), source, sink,
-  "This call to " + call.(Replacement).getFrom() +
-    " depends on a user-provided value. Replace it with " + call.(Replacement).getTo() + "."
+  "This call to " + call.(Replacement).getFrom() + " depends on a . Replace it with " +
```
aibaars, line 76:
```suggestion
  "This call to " + call.(Replacement).getFrom() + " depends on a $@. Replace it with " +
```

aibaars (commented):
Looks good to me, but the alert message is missing a `$@`

erik-krogh:
I also added a drive-by simplification.

aibaars (dismissed):

aibaars (approved):

PR #10677, opened 2022-10-04
Kotlin: Extract parameter modifiers (`noinline`, `crossinline`)

tamasvajk:


smowton (approved):
Looks plausible -- though I now need to google 'crossinline'

PR #10679, opened 2022-10-04
C#/Java: Limit telemetry results.

michaelnebel:
It turns out that there are some minor issues with the number of reported results by the telemetry queries.
The queries are supposed to *limit* the results to the top 1k most relevant (most frequent) results.
However, the ordering (ranking) was based on the callables themselves and not the string representation used for the result. Unfortunately, the string representation is not a precise description of a callable, which means that multiple callables can map to the same description causes slight inconsistencies in the *ranking* (ie. we can risk getting less than 1k results even though there result set we try to limit is bigger than 1k).

For C# some examples cases are callables of the same name in subclasses of the same name, but nested in classes with different names.
To solve this the string representation are now used in the *ranking*. However, this might give a slight inconsistency in the usage count as callables that have the same string representation gets their usage count collapsed.

Furthermore, it was also discovered that for C#, calls to testing frameworks were not excluded correctly. This has also been fixed.

### DIFF
--- a/csharp/ql/src/Telemetry/ExternalApi.qll
+++ b/csharp/ql/src/Telemetry/ExternalApi.qll
@@ -17,8 +17,10 @@ private import semmle.code.csharp.security.dataflow.flowsources.Remote
 class TestLibrary extends RefType {
   TestLibrary() {
     this.getNamespace()
-        .getName()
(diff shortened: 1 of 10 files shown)


michaelnebel:
DCA: No performance regressions are shown.

PR #10681, opened 2022-10-04
Swift: Use ClassOrStructDecl

geoffw0:
Use the new `ClassOrStructDecl` (from https://github.com/github/codeql/pull/10595 by @MathiasVP), in place of `ClassDecl` and `StructDecl` in all Swift queries.  I consider this more robust in several ways:
 - when the query + test were written with the wrong one (out of `class` / `struct`).  Such a mistake is hard to spot but will mean we find no results in real world usage.  I've made this mistake myself, twice, and I've seen someone else make it as well.
   - an alternative solution would be to check all current QL for Swift diligently.
 - if something in a standard library changes between `struct` and `class` in future.
 - if the QL is reused (copy-pasted) to match a different thing in the standard library in future.

The downside is minimal - if there is a thing matching the standard library one in all ways except being the opposite between `struct` / `class`, we could match it in addition to the correct thing.

### DIFF
--- a/swift/ql/src/queries/Security/CWE-079/UnsafeWebViewFetch.ql
+++ b/swift/ql/src/queries/Security/CWE-079/UnsafeWebViewFetch.ql
@@ -31,7 +31,7 @@ class Sink extends DataFlow::Node {
       int baseUrlArg
     |
       // arguments to method calls...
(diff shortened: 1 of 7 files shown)


MathiasVP (approved):
LGTM!

PR #10682, opened 2022-10-04
C++: Fix potentially bad join

MathiasVP:
The joins in this predicate turned out to be quite bad after I tried to switch to IR use-use dataflow instead of AST dataflow. This PR should hopefully prevent the optimizer from trying to do a join on the base type.

### DIFF
--- a/cpp/ql/src/Likely Bugs/Conversion/CastArrayPointerArithmetic.ql
+++ b/cpp/ql/src/Likely Bugs/Conversion/CastArrayPointerArithmetic.ql
@@ -46,7 +46,7 @@ class CastToPointerArithFlow extends DataFlow::Configuration {
  */
 pragma[inline]
 predicate hasBaseType(Expr e, Type base) {
-  pragma[only_bind_into](base) = pragma[only_bind_out](e.getType().(DerivedType).getBaseType())
+  pragma[only_bind_into](base) = e.getType().(DerivedType).getBaseType()
 }
 
 /**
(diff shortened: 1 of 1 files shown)


cpp/ql/src/Likely Bugs/Conversion/CastArrayPointerArithmetic.ql
```diff
@@ -35,6 +38,17 @@ class CastToPointerArithFlow extends DataFlow::Configuration {
...
 
+/**
+ * Holds if the type of `e` is a `DerivedType` with `base` as its base type.
+ *
+ * This predicate ensures that joins go from `e` to `base` instead
+ * of the other way around.
+ */
+pragma[inline]
+predicate hasBaseType(Expr e, Type base) {
+  pragma[only_bind_into](base) = pragma[only_bind_out](e.getType().(DerivedType).getBaseType())
```
aschackmull, line 49: `pragma[only_bind_into](x) = pragma[only_bind_out](y)` is completely equivalent to both
`x = pragma[only_bind_out](y)` and `pragma[only_bind_into](x) = y`, so you can just choose one. They both express "if you have `y` then you also have `x`". I.e. one of your two pragmas are superfluous.

MathiasVP: Oh, right. I had blindly copy-pasted the join-order trick from `getNodeEnclosingCallable` in `DataFlowImplCommon` without paying attention to the fact that this was an equality operation. Thanks for spotting this!

MathiasVP: Fixed in fcd69a005f4e533013843b3ffd283049c2a49f1b.

geoffw0 (approved):
LGTM.

Looks like the DCA run was a bit slower, but it doesn't appear to relate to any changes here.  Just bad luck I guess.  Tuple sums for the query seem as good or better than before.

PR #10683, opened 2022-10-04
Kotlin: implement $default function synthesis

smowton:
These functions substitute in default parameter values where they aren't supplied. The main trickiness arises when generics are in play, including generic functions, but we need to extract perhaps deeply-nested types that refer to those type variables in erased form. This is realised by recording on the declaration stack that we're extracting a `$default` method, and erasing references to such type variables each time one is referenced.


java/kotlin-extractor/src/main/kotlin/KotlinFileExtractor.kt
```diff
@@ -1514,6 +1636,147 @@ open class KotlinFileExtractor(
...
+    private fun getDefaultsMethodLastArgType(f: IrFunction) =
+        (
+            if (f is IrConstructor)
+                defaultConstructorMarkerType
+            else
+                null
+        ) ?: pluginContext.irBuiltIns.anyType
+
+    private fun getDefaultsMethodArgTypes(f: IrFunction) =
+        // The $default method has type ([extensionReceiver], [dispatchReciever], paramTypes..., int, Object)
```
tamasvajk, line 1656: There's a typo in `Reciever`.

java/kotlin-extractor/src/main/kotlin/KotlinFileExtractor.kt
```diff
@@ -826,10 +827,117 @@ open class KotlinFileExtractor(
...
+                                        dispatchReceiver?.let { Triple(it.type, realFnIdxOffset, -1) },
+                                        extReceiver?.let { Triple(it.type, 0, 0) }
+                                    )
+                                paramMappings.forEach { (type, fromIdx, toIdx) ->
+                                    extractVariableAccess(tw.getLabelFor<DbParam>(getValueParameterLabel(id, fromIdx)), type, locId, thisCallId, toIdx, id, returnId)
+                                }
+                                if (f.shouldExtractAsStatic)
+                                    extractStaticTypeAccessQualifier(f, thisCallId, locId, id, returnId)
+                                else if (f.isLocalFunction()) {
+                                    extractNewExprForLocalFunction(getLocallyVisibleFunctionLabels(f), thisCallId, locId, id, returnId)
```
tamasvajk, line 913: I don't think we need to extract a `new` expression for local functions. Shouldn't this be either param 0 or a static type access instead?

java/ql/test/kotlin/library-tests/parameter-defaults/PrintAst.expected
```diff
@@ -0,0 +1,849 @@
...
+#   86|               5: [BlockStmt] { ... }
+#   87|                 0: [ExprStmt] <Expr>;
+#   87|                   0: [MethodAccess] sink(...)
+#   87|                     -1: [TypeAccess] TestKt
+#   87|                     0: [VarAccess] y
+#   86|             3: [Method] f$default
+#   86|               3: [TypeAccess] Unit
+#-----|               4: (Parameters)
+#   86|                 0: [Parameter] p0
+#   86|                   0: [TypeAccess] String
```
tamasvajk, line 524: Shouldn't the first parameter be the dispatch receiver?

java/ql/test/kotlin/library-tests/parameter-defaults/PrintAst.expected
```diff
@@ -0,0 +1,849 @@
...
+#   86|                       1: [VarAccess] p3
+#   86|                     1: [IntegerLiteral] 0
+#   86|                   1: [ExprStmt] <Expr>;
+#   86|                     0: [AssignExpr] ...=...
+#   86|                       0: [VarAccess] p2
+#   86|                       1: [StringLiteral] hello world
+#   86|                 2: [ReturnStmt] return ...
+#   86|                   0: [MethodAccess] f(...)
+#   86|                     -1: [ClassInstanceExpr] new (...)
+#   86|                       -3: [TypeAccess] Object
```
tamasvajk, line 557: I don't think we should create an instance here, the instance is created at [the call site](https://github.com/github/codeql/pull/10683/commits/7e1a6035449e99cedea7fcb0f3e861a89ac4852c#diff-3e5daacdf0970b183314af74d106e324d2530bdb765f4aecf5c524518efd4757R571-R572), and should be propagated here as the qualifier.

java/ql/test/kotlin/library-tests/parameter-defaults/PrintAst.expected
```diff
@@ -0,0 +1,849 @@
...
+#   90|         1: [LocalTypeDeclStmt] class ...
+#   90|           0: [LocalClass] 
+#   90|             1: [Constructor] 
+#   90|               5: [BlockStmt] { ... }
+#   90|                 0: [SuperConstructorInvocationStmt] super(...)
+#   90|             2: [Method] user
+#   90|               3: [TypeAccess] Unit
+#   90|               5: [BlockStmt] { ... }
+#   91|                 0: [ExprStmt] <Expr>;
+#   91|                   0: [MethodAccess] f$default(...)
```
tamasvajk, line 570: `f$default` is `static`, so it should have a type access as qualifier, instead of the below expression. Shouldn't this result in some missing expected flows?

tamasvajk (changes requested):
Added my findings below. Also, you're adding an `Object` parameter to the `$default` methods to be consistent with the generated bytecode. You could add a `.java` test file to call one of these methods and see if the two extractors share a consistent view.

java/kotlin-extractor/src/main/kotlin/KotlinFileExtractor.kt
```diff
@@ -826,10 +827,117 @@ open class KotlinFileExtractor(
...
+
+        val id = getDefaultsMethodLabel(f)
+        val locId = getLocation(f, null)
+        val extReceiver = f.extensionReceiverParameter
+        val dispatchReceiver = if (f.shouldExtractAsStatic) null else f.dispatchReceiverParameter
+        val parameterTypes = listOfNotNull(extReceiver?.let { erase(it.type) }) + getDefaultsMethodArgTypes(f)
+        val allParamTypeResults = parameterTypes.mapIndexed { i, paramType ->
+            val paramId = tw.getLabelFor<DbParam>(getValueParameterLabel(id, i))
+            extractValueParameter(paramId, paramType, "p$i", locId, id, i, paramId, isVararg = false, syntheticParameterNames = true).also {
+                extractTypeAccess(useType(paramType), locId, paramId, -1)
```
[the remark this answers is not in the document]
smowton: Yes, it depends on whether we're extracting the body or not (following the Java extractor's convention) -- this behaviour is added in a later commit.

java/ql/test/kotlin/library-tests/parameter-defaults/PrintAst.expected
smowton: Not for a local function it seems:

```
public class Test {
  fun f() {
    fun g(x: Int = 1) { }
  }
}
```

Leads to

```
  private static final void f$g(int);
...
  static void f$g$default(int, int, java.lang.Object);
```

java/kotlin-extractor/src/main/kotlin/KotlinFileExtractor.kt
```diff
@@ -826,10 +827,117 @@ open class KotlinFileExtractor(
...
+        if (f.symbol is IrConstructorSymbol) {
+            val constrId = id.cast<DbConstructor>()
+            extractConstructor(constrId, shortName, paramsSignature, parentId, constrId)
+        } else {
+            val methodId = id.cast<DbMethod>()
+            extractMethod(methodId, locId, shortName, erase(f.returnType), paramsSignature, parentId, methodId, origin = null, extractTypeAccess = true)
+            addModifiers(id, "static")
+        }
+        tw.writeHasLocation(id, locId)
+        addModifiers(id, "public")
```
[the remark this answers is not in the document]
smowton: 👍 will fix

java/kotlin-extractor/src/main/kotlin/KotlinFileExtractor.kt
```diff
@@ -1514,6 +1636,147 @@ open class KotlinFileExtractor(
...
+    private fun getDefaultsMethodName(f: IrFunction) =
+        if (f is IrConstructor) {
+            f.returnType.let {
+                when {
+                    it.isAnonymous -> ""
+                    else -> it.classFqName?.shortName()?.asString() ?: f.name.asString()
+                }
+            }
+        } else {
+            f.name.asString() + "\$default"
```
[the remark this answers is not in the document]
smowton: 👍 will check what happens

java/ql/test/kotlin/library-tests/parameter-defaults/PrintAst.expected
smowton: Surprisingly not -- I guess the rogue parameter -1 just gets ignored. I'll fix this though; thinking it should be a case of calling the static function without a `new` expression here, then calling the 'real' local function using the normal `(new WrapperClass()).f(...)` pattern in `f$default`.

java/ql/test/kotlin/library-tests/parameter-defaults/PrintAst.expected
tamasvajk: Hmm, it might be somehow compiler or target version dependent. The byte code viewer in IntelliJ shows me the following. (I stripped the noise for brevity.)
```
final class Test$f$1 extends kotlin/jvm/internal/Lambda implements kotlin/jvm/functions/Function1 {
  public final invoke(I)V
  public static synthetic invoke$default(LTest$f$1;IILjava/lang/Object;)V
}
```

tamasvajk: @smowton I wonder if this difference can only happen for constructs that are definitely not visible from outside. Otherwise it could occur that we're not compatible with the java extractor, couldn't it?

smowton: Hmm, true. I'm inclined to punt on the details for now on the grounds that function-local declarations getting direct calls from Java is surely very niche.

java/kotlin-extractor/src/main/kotlin/KotlinFileExtractor.kt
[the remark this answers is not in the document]
smowton: Fixed in https://github.com/github/codeql/pull/10683/commits/5e182755a53086b5e629bebfea19583d441636d5

java/kotlin-extractor/src/main/kotlin/KotlinFileExtractor.kt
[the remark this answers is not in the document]
smowton: Fixed in https://github.com/github/codeql/pull/10683/commits/34b83f01d013ee8db522748c9fb33b631609cee2

java/kotlin-extractor/src/main/kotlin/KotlinFileExtractor.kt
smowton: Fixed in https://github.com/github/codeql/pull/10683/commits/bec948682d01f4b87179d4b1a056b6c608ea953a

smowton:
The internal query failure will be fixed by https://github.com/github/semmle-code/pull/43920

tamasvajk (approved):
LGTM

PR #10684, opened 2022-10-04
Java: Add query for Sensitive Keyboard Cache 

joefarebrother:
Query that detects instances of sensitive information that could be saved into the keyboard cache. 

Considers cases in which an input field declared is declared via XML with a name that indicates it may contain sensitive information, but its input type, as declared either in XML or in code, indicates that it is saved to the keyboard cache.


java/ql/src/Security/CWE/CWE-524/SensitiveKeyboardCache.qhelp
```diff
@@ -0,0 +1,36 @@
...
+<p>In the following example, the field labeled BAD could allow the password to be saved to the keyboard cache;
+ whereas the field labeled GOOD uses the <code>"textPassword"</code> input type, which ensures that it is not.</p>
+
+<sample src="Example.xml" />
+
+</example>
+<references>
+
+<li>
+  OWASP Mobile Application Security Testing Guie: <a href="https://github.com/OWASP/owasp-mastg/blob/b7a93a2e5e0557cc9a12e55fc3f6675f6986bb86/Document/0x05d-Testing-Data-Storage.md#determining-whether-the-keyboard-cache-is-disabled-for-text-input-fields-mstg-storage-5">Determining Whether the Keyboard Cache Is Disabled for Text Input Fields</a>.
```
Marcono1234, line 29:
```suggestion
  OWASP Mobile Application Security Testing Guide: <a href="https://github.com/OWASP/owasp-mastg/blob/b7a93a2e5e0557cc9a12e55fc3f6675f6986bb86/Document/0x05d-Testing-Data-Storage.md#determining-whether-the-keyboard-cache-is-disabled-for-text-input-fields-mstg-storage-5">Determining Whether the Keyboard Cache Is Disabled for Text Input Fields</a>.
```

java/ql/src/Security/CWE/CWE-524/SensitiveKeyboardCache.ql
```diff
@@ -0,0 +1,17 @@
+/**
+ * @name Android sensetive keyboard cache
```
Marcono1234, line 2:
```suggestion
 * @name Android sensitive keyboard cache
```

java/ql/src/Security/CWE/CWE-524/SensitiveKeyboardCache.qhelp
```diff
@@ -0,0 +1,36 @@
...
+</li>
+<li>
+  Android Developers: <a href="https://developer.android.com/reference/android/widget/TextView#attr_android:inputType"> <code>android:inputType</code> attribute documentation.
```
Marcono1234, line 32: Looks like this is missing a closing `</a>`:
```suggestion
  Android Developers: <a href="https://developer.android.com/reference/android/widget/TextView#attr_android:inputType"> <code>android:inputType</code> attribute documentation</a>.
```

java/ql/src/Security/CWE/CWE-524/SensitiveKeyboardCache.qhelp
```diff
@@ -0,0 +1,36 @@
...
+
+<overview>
+<p>When a user enters information to a text input field on an Android application, then by default their input is saved to a keyboard cache, 
+which provides autocomplete suggestions and predictions. If the input field is expected to contain sensitive information, such as a password or banking details, 
+this sensitive data may be leaked to other applications via the keyboard cache.</p>
+
+</overview>
+<recommendation>
+
+<p>For input fields expected to accept sensitive information, an input type such as <code>"textNoSuggestions"</code> (or <code>"textPassword"</code> for a password)
```
Marcono1234, line 14: Out of curiosity, is there any guarantee that `textNoSuggestions` will for sure not cache or process the input in any way? The documentation for the corresponding [`TYPE_TEXT_FLAG_NO_SUGGESTIONS`](https://developer.android.com/reference/android/text/InputType#TYPE_TEXT_FLAG_NO_SUGGESTIONS) says (emphasis mine):
>  Note the contrast with `TYPE_TEXT_FLAG_AUTO_CORRECT` and `TYPE_TEXT_FLAG_AUTO_COMPLETE`: `TYPE_TEXT_FLAG_NO_SUGGESTIONS` means the IME does not need to show an interface to display suggestions. **Most IMEs** will also take this to mean they do not need to try to auto-correct what the user is typing.

This sounds as if some IMEs might process the input in some way.

Is `textVisiblePassword` maybe better fitting?

atorralba: The source for this is https://github.com/OWASP/owasp-mastg/blob/8cb9ed4a5db21acd40d87212042d98ef7042fbf0/Document/0x05d-Testing-Data-Storage.md#static-analysis-4, which we expect to be accurate in principle.

Also, IMEs are assumed to honor these attributes, but in the end it's up to [the actual implementation](https://developer.android.com/develop/ui/views/touch-and-input/creating-input-method#GeneralDesign). I would expect at least the Android IME to properly not cache the input when suggestions are disabled.

java/ql/lib/semmle/code/java/security/SensitiveKeyboardCacheQuery.qll
```diff
@@ -0,0 +1,53 @@
...
+
+/** An Android Layout XML file. */
+class AndroidLayoutXmlFile extends XmlFile {
+  AndroidLayoutXmlFile() { this.getAbsolutePath().matches("%/res/layout/%.xml") }
+}
+
+/** An XML element that represents an editable text field. */
+class AndroidEditableXmlElement extends XmlElement {
+  XmlAttribute inputType;
+  XmlAttribute id;
```
atorralba, line 14: Suggest using the `AndroidXmlAttribute` type so that you can remove the namespace prefix checks below:

```suggestion
  AndroidXmlAttribute inputType;
  AndroidXmlAttribute id;
```

java/ql/lib/semmle/code/java/security/SensitiveKeyboardCacheQuery.qll
```diff
@@ -0,0 +1,53 @@
+/** Definitions for the keyboard cache query */
+
+import java
+import semmle.code.java.security.SensitiveActions
+
+/** An Android Layout XML file. */
+class AndroidLayoutXmlFile extends XmlFile {
```
atorralba, line 7: Can't this be `private`?

java/ql/lib/semmle/code/java/security/SensitiveKeyboardCacheQuery.qll
```diff
@@ -0,0 +1,53 @@
...
+  AndroidLayoutXmlFile() { this.getAbsolutePath().matches("%/res/layout/%.xml") }
```
atorralba, line 8: I think it'd be better to use `getRelativePath()` here instead. The chances are low, but let's just not risk `/res/layout/` appearing outside the project directory in the filesystem and skewing the results.

java/ql/src/Security/CWE/CWE-524/SensitiveKeyboardCache.ql
```diff
@@ -0,0 +1,18 @@
+/**
+ * @name Android sensitive keyboard cache
+ * @description Sensitive information should not be saved to the keyboard cache.
+ * @kind problem
+ * @problem.severity warning
+ * @security-severity 8.1
+ * @id java/android/sensitive-keyboard-cache
+ * @tags security
+ *       external/cwe/cwe-524
+ * @precision high
```
atorralba, line 10: Given the heuristic nature of the query (the main logic being regex), I wonder if this should be `medium` instead. In your experiments, how does the FP ratio look like?

atorralba: It being `medium` would be consistent with the `CleartextStorage*` queries, which also handle sensitive data in a similar way.

atorralba (commented):
This mostly looks good, I added some minor comments.

Other than that, have you considered that the input type could be added programmatically via [`setInputType`](https://developer.android.com/reference/android/widget/TextView#setInputType(int))? I'm curious about how many alerts raised by this query could be discarded because of this. If there are enough, it could might sense to handle it as well.

WDYT?

java/ql/lib/semmle/code/java/security/SensitiveKeyboardCacheQuery.qll
```diff
@@ -0,0 +1,143 @@
...
+  }
+}
+
+/** An XML element that represents an editable text field. */
+class AndroidEditableXmlElement extends AndroidLayoutXmlElement {
+  AndroidEditableXmlElement() {
+    exists(Class editText |
+      editText.hasQualifiedName("android.widget", "EditText") and
+      editText = this.getClass().getASourceSupertype*()
+    )
```
atorralba, line 42: Could be simplified:

```suggestion
    this.getClass().getASourceSupertype*().hasQualifiedName("android.widget", "EditText")
```

java/ql/lib/semmle/code/java/security/SensitiveKeyboardCacheQuery.qll
```diff
@@ -9,24 +10,106 @@ private class AndroidLayoutXmlFile extends XmlFile {
...
+  (
+    not f.getName().matches("%TEXT%")
+    or
+    f.getName().matches("%PASSWORD%")
+    or
+    f.getName() = "TYPE_TEXT_FLAG_NO_SUGGESTIONS"
+  )
+}
+
+/** Configuration that finds uses of `setInputType` that for non cached fields. */
```
atorralba, line 97:
```suggestion
/** Configuration that finds uses of `setInputType` for non cached fields. */
```

java/ql/lib/semmle/code/java/security/SensitiveKeyboardCacheQuery.qll
```diff
@@ -9,24 +10,106 @@ private class AndroidLayoutXmlFile extends XmlFile {
...
+    or
+    exists(Method m |
+      m.hasQualifiedName("android.app", "Activity", ["findViewById", "requireViewById"]) and
+      this = m.getAnOverride*()
+    )
+  }
+}
+
+/** Gets a use of the view that has the given id. */
+private Expr getAUseOfId(string id) {
```
atorralba, line 62: Result is always `findView`, so the return type can be restricted to `MethodAccess`, right? Also, maybe a slight rename would help clarify what the predicate does:


```suggestion
private MethodAccess getAUseOfViewWithId(string id) {
```

java/ql/lib/semmle/code/java/security/SensitiveKeyboardCacheQuery.qll
```diff
@@ -9,24 +10,106 @@ private class AndroidLayoutXmlFile extends XmlFile {
...
+    id_field.getDeclaringType() = r_id and
+    id_field.hasName(name)
+  |
+    DataFlow::localExprFlow(id_field.getAnAccess(), findView.getArgument(0)) and
+    result = findView
+  )
+}
+
+/** Gets the argument of a use of `setInputType` called on the view with the given id. */
+private Expr setInputTypeForId(string id) {
```
atorralba, line 77:
```suggestion
private Argument setInputTypeForId(string id) {
```

java/ql/lib/semmle/code/java/security/SensitiveKeyboardCacheQuery.qll
```diff
@@ -9,24 +10,106 @@ private class AndroidLayoutXmlFile extends XmlFile {
...
+
+/** Holds if the given field is a constant flag indicating that an input with this type will not be cached. */
+private predicate inputTypeFieldNotCached(Field f) {
+  f.getDeclaringType().hasQualifiedName("android.text", "InputType") and
+  (
+    not f.getName().matches("%TEXT%")
+    or
+    f.getName().matches("%PASSWORD%")
+    or
+    f.getName() = "TYPE_TEXT_FLAG_NO_SUGGESTIONS"
```
atorralba, line 93: Use `hasName` for better performance?

```suggestion
    f.hasName("TYPE_TEXT_FLAG_NO_SUGGESTIONS")
```

java/ql/lib/semmle/code/java/security/SensitiveKeyboardCacheQuery.qll
```diff
@@ -48,5 +131,13 @@ private predicate inputTypeCached(string ty) {
...
 AndroidEditableXmlElement getASensitiveCachedInput() {
   result.getId().regexpMatch(getInputSensitiveInfoRegex()) and
-  inputTypeCached(result.getInputType())
+  (
+    inputTypeCached(result.getInputType())
+    or
+    not exists(result.getInputType()) and
+    not exists(GoodInputTypeConf conf, DataFlow::Node src, DataFlow::Node sink |
+      conf.hasFlow(src, sink) and
+      sink.asExpr() = setInputTypeForId(result.getId())
```
atorralba, line 140: Does setting the input type in source code only work if it hasn't been set in XML? (i.e. setting it in XML has preference over source code)

atorralba (commented):
Hey @joefarebrother thanks for working on this, I think it'll be a valuable addition. I added some more minor comments and a conceptual question, but otherwise this LGTM. If DCA and MRVA look good, we should ask for a docs review.

java/ql/src/Security/CWE/CWE-524/SensitiveKeyboardCache.qhelp
```diff
@@ -0,0 +1,36 @@
+<!DOCTYPE qhelp PUBLIC
+  "-//Semmle//qhelp//EN"
+  "qhelp.dtd">
+<qhelp>
+
+<overview>
+<p>When a user enters information to a text input field on an Android application, then by default their input is saved to a keyboard cache, 
+which provides autocomplete suggestions and predictions. If the input field is expected to contain sensitive information, such as a password or banking details, 
+this sensitive data may be leaked to other applications via the keyboard cache.</p>
```
sabrowning1, line 9:
```suggestion
<p>When a user enters information in a text input field on an Android application, their input is saved to a keyboard cache which provides autocomplete suggestions and predictions. There is a risk that sensitive user data, such as passwords or banking information, may be leaked to other applications via the keyboard cache.</p>
```

java/ql/src/Security/CWE/CWE-524/SensitiveKeyboardCache.qhelp
```diff
@@ -0,0 +1,36 @@
...
+
+</overview>
+<recommendation>
+
+<p>For input fields expected to accept sensitive information, an input type such as <code>"textNoSuggestions"</code> (or <code>"textPassword"</code> for a password) 
+should be used to ensure that the input does not get stored in the keyboard cache.</p>
```
sabrowning1, line 15:
```suggestion
<p>For input fields expected to accept sensitive information, use input types such as <code>"textNoSuggestions"</code> (or <code>"textPassword"</code> for a password) to ensure the input does not get stored in the keyboard cache.</p>
```

java/ql/src/Security/CWE/CWE-524/SensitiveKeyboardCache.qhelp
```diff
@@ -0,0 +1,36 @@
...
+<p>The input type can also be set in code through <code>TextView.setInputType()</code> rather than declared through XML.</p>
```
sabrowning1, line 16:
```suggestion
<p>In addition to declaring an input type through XML, you can also set the input type through code using <code>TextView.setInputType()</code>.</p>
```

java/ql/src/Security/CWE/CWE-524/SensitiveKeyboardCache.qhelp
```diff
@@ -0,0 +1,36 @@
...
+</recommendation>
+<example>
+
+<p>In the following example, the field labeled BAD could allow the password to be saved to the keyboard cache;
```
sabrowning1, line 20:
```suggestion
<p>In the following example, the field labeled BAD allows the password to be saved to the keyboard cache,
```

java/ql/src/Security/CWE/CWE-524/SensitiveKeyboardCache.qhelp
```diff
@@ -0,0 +1,36 @@
...
+ whereas the field labeled GOOD uses the <code>"textPassword"</code> input type, which ensures that it is not.</p>
```
sabrowning1, line 21:
```suggestion
 whereas the field labeled GOOD uses the <code>"textPassword"</code> input type to ensure that it is not.</p>
```

sabrowning1 (commented):
👋🏼 @joefarebrother, thanks for your work on this! I've suggested a few small changes to the `.qhelp` file to align with our style guide. Let me know if you have any questions!

java/ql/src/Security/CWE/CWE-524/SensitiveKeyboardCache.qhelp
joefarebrother: How accurately does this convey that setting the input type in code is an alternative to setting it in XML? To me this wording suggests that it's something you'd do as well as declare it in XML, which isn't necassary.

sabrowning1: Good point, I've updated my suggestion above to make that more clear 👍🏼

sabrowning1 (approved):
Looks good for Docs, thanks @joefarebrother! 👍🏼

atorralba (approved):
🎉

PR #10685, opened 2022-10-04
Ruby: summarize unary splat operators and add local field step

asgerf:
Adds the local field step to type-tracking, such that any write to an instance variable can flow to a read.
```rb
class Foo
  def foo(x)
    @field = Bar.new
  end
  def bar
    @field # 'Bar.new' flows here
  end
end
```
For the time being it doesn't take inheritance into account.

This initially caused a regression in `opal`. Investigating it I noticed some dubious API graph edges, and concluded that we should not propagate use/def nodes through self-argument passing.

I wonder if we should remove self-argument passing from type-tracking altogether, since some parts of the call-graph construction also block this. But tracking of singleton methods still relies on self-argument passing so I didn't go that far in this PR, and just blocked it directly in API graphs instead.

Also marks the unary `*` operator as a simple callable.

[Evaluation](https://github.com/github/codeql-dca-main/issues/7706) shows
- 39k new call edges
- Lost some taint sinks and sources inside the source code of libraries for which we have a model. AFAICT a model would find a use of a class which then flows into `self` in one of its methods, and then starts flagging up sinks inside the class. If we _want_ to flag these sinks, the better solution is to directly treat `self` in the class as uses of it, and not rely on finding a call site. I don't believe we currently lose any sources/sinks in clients of these libraries.
- Gained a few taint sinks and sources which seem legit.
- Gained a new cleartext-storage alert due to the new call edges. It seems to be a FP due to `HashLiteralPasswordSource` treating an entire hash literal as a source, but where the value being read in the end isn't the key containing the password

### DIFF
--- a/ruby/ql/lib/codeql/ruby/ApiGraphs.qll
+++ b/ruby/ql/lib/codeql/ruby/ApiGraphs.qll
@@ -522,7 +522,8 @@ module API {
       or
       exists(TypeTrackerSpecific::TypeTrackerContent c |
         TypeTrackerSpecific::basicLoadStep(node, ref, c) and
-        lbl = Label::content(c.getAStoreContent())
+        lbl = Label::content(c.getAStoreContent()) and
+        not c.isSingleton(any(DataFlow::Content::AttributeNameContent k))
       )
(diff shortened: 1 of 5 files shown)


hvitved (approved):
LGTM

PR #10686, opened 2022-10-04
Ruby: Remove `PairValueContent`

hvitved:
As pointed out by @asgerf, we no longer need `PairValueContent`, after `SynthHashSplatArgumentNode` was introduced (it was originally not there at the time when `PairValueContent` got introduced).


asgerf (approved):
Looks great! 👍

aibaars (approved):
Looks good to me.

PR #10689, opened 2022-10-04
Swift: Query for CWE-312: Exposure of sensitive information using NSUserDefaults

d10c:
This PR introduces a new query `swift/cleartext-storage-preferences`, largely modelled after `swift/cleartext-storage-database`.


swift/ql/src/queries/Security/CWE-312/CleartextStoragePreferences.qhelp
```diff
@@ -0,0 +1,28 @@
+<!DOCTYPE qhelp PUBLIC "-//Semmle//qhelp//EN" "qhelp.dtd">
+<qhelp>
+<overview>
+
+<p>Sensitive information that is stored unencrypted in the defaults database is accessible to an attacker who gains access to that database. For example, the information could be accessed by any process or user in a rooted device, or exposed through another vulnerability.</p>
```
geoffw0, line 5: According to https://developer.apple.com/documentation/foundation/userdefaults, preferences can also be accessed by "App extensions on macOS and iOS".  I'm not sure but it seems like this (and perhaps other aspects described in that link) widens the attack surface as well.

swift/ql/src/queries/Security/CWE-312/CleartextStoragePreferences.qhelp
```diff
@@ -0,0 +1,28 @@
...
+<p>The following example shows three cases of storing information using NSUserDefaults. In the 'BAD' case, the data that is stored is sensitive (a credit card number) and is not encrypted.  In the 'GOOD' cases, the data is either not sensitive, or is protected with encryption.</p>
+
+<sample src="CleartextStoragePreferences.swift" />
+
+</example>
+<references>
+
+<li>
+  OWASP Top 10:2021:
+  <a href="https://owasp.org/Top10/A02_2021-Cryptographic_Failures/">A02:2021 � Cryptographic Failures</a>.
```
geoffw0, line 24: https://developer.apple.com/documentation/foundation/userdefaults might make a good reference as well?

swift/ql/src/queries/Security/CWE-312/CleartextStoragePreferences.ql
```diff
@@ -0,0 +1,105 @@
...
+      f.getName() = ["set(_:forKey:)"] and
+      call.getStaticTarget() = f and
+      call.getArgument(0).getExpr() = this.asExpr()
+    )
+  }
+
+  override string getStoreName() { result = "the user defaults database" }
+}
+
+/** The `DataFlow::Node` of an expression that gets written to iCloud */
```
geoffw0, line 41: This is a great addition.  Do you think calling this "application preferences" in the query `@name` and the "defaults database" in the `.qhelp` is accurate enough when we're also including iCloud ... or should we look for a more generic term?

geoffw0 (commented):
Looks good so far, I've left a few thoughts but nothing critical.

swift/ql/src/queries/Security/CWE-312/CleartextStoragePreferences.ql
d10c: I was trying to be generic, but it's kinda hard. Maybe: "an application preference store"? "Local and/or {shared,distributed} preferences"?

geoffw0: Yeah I can see its difficult to be generic, while also being clear what we're talking about.  I think I prefer "an application preference store", because its readable but does imply more than one type of store is supported.  Whatever we pick we should try to be consistent.

swift/ql/test/query-tests/Security/CWE-312/testUserDefaults.swift
```diff
@@ -0,0 +1,70 @@
...
+{
+}
+
+class UserDefaults : NSObject
+{
+    class var standard: UserDefaults {
+        return UserDefaults()
+    }
+
+	func set(_ value: Any?, forKey key: String) {}
```
geoffw0, line 14: And this:
```
func set(_ value: Any?, forKey defaultName: String) {}
```
(https://developer.apple.com/documentation/foundation/userdefaults/1414067-set)

geoffw0 (commented):
Otherwise LGTM. :+1:

swift/ql/test/query-tests/Security/CWE-312/testNSUbiquitousKeyValueStore.swift
```diff
@@ -0,0 +1,70 @@
...
+{
+}
+
+class NSUbiquitousKeyValueStore : NSObject
+{
+    class var `default`: NSUbiquitousKeyValueStore {
+        return NSUbiquitousKeyValueStore()
+    }
+
+	func set(_ value: Any?, forKey key: String) {}
```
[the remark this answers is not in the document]
d10c: Yeah, except I assumed the naming of parameters doesn't really matter in stubs. The selector is the same, `set(_:forKey:)`, in both cases.

swift/tools/qltest.sh
```diff
@@ -10,5 +10,7 @@ for src in *.swift; do
   opts=(-sdk "$CODEQL_EXTRACTOR_SWIFT_ROOT/qltest/$CODEQL_PLATFORM/sdk" -c -primary-file $src)
   opts+=($(sed -n '1 s=//codeql-extractor-options:==p' $src))
   echo -e "calling extractor with flags: ${opts[@]}\n" >> $QLTEST_LOG
-  "$CODEQL_EXTRACTOR_SWIFT_ROOT/tools/$CODEQL_PLATFORM/extractor" "${opts[@]}" >> $QLTEST_LOG 2>&1
+  "$CODEQL_EXTRACTOR_SWIFT_ROOT/tools/$CODEQL_PLATFORM/extractor" "${opts[@]}" >> $QLTEST_LOG 2>&1 || FAILED=1
 done
+
+[ -z "$FAILED" ] || cat "$QLTEST_LOG" # Show compiler errors on extraction failure
```
geoffw0, line 16: This should be a nice quality of life improvement!

swift/ql/test/query-tests/Security/CWE-312/testUserDefaults.swift
d10c: Ditto.

swift/ql/test/query-tests/Security/CWE-312/testNSUbiquitousKeyValueStore.swift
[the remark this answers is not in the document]
MathiasVP: As long as the query isn't looking at the name of the parameter, then yes, it shouldn't matter. But in case we ever decide to do that (for whatever reason that I cannot currently think of), I think we should stick to the same parameter names as used in real codebases.

geoffw0:
I think this is just waiting on the docs review and either a good DCA run, or for us to give up on getting a good DCA run at this time.  The only 'good' run I've seen for Swift was https://github.com/github/codeql-dca-main/issues/7605 - perhaps some of the settings from that run might be worth trying???

stevecat (approved):
Reviewing on behalf of the docs team. These changes look good! :+1: ✨

PR #10692, opened 2022-10-05
Ruby: fix CFG and toString for anonymous '*' and '**'

aibaars:
The CFG and toString were broken for `*` and `**` without identifiers.


hvitved (approved):
LGTM

PR #10693, opened 2022-10-05
Java: Fixes bad magic in `Guard::guardControls_v3`

atorralba:
Fixes bad magic in `Guard::guardControls_v3`, which was causing mutual recursion with `Guard`s that used `Guard::controls` in their definition.

Credits to @aschackmull for the solution.


atorralba:
DCA seems uneventful.

aschackmull (approved):

PR #10694, opened 2022-10-05
C#: Count unsupported external library generics.

michaelnebel:


michaelnebel:
@hvitved : What do you think about this. Would this be a query for reasonably measuring the changes to summaries for generic like methods.

csharp/ql/src/meta/frameworks/UnsupportedExternalGenericAPIs.ql
```diff
@@ -0,0 +1,22 @@
+/**
+ * @name Usage count of unsupported external library generic APIs
+ * @description A call to an unsuppported external library generic API.
+ * @kind metric
+ * @tags summary
```
hvitved, line 5: Needs a `meta` tag.

hvitved (commented):
What is it exactly we want to measure; is it the number of calls to methods that use type-based summaries? As-is, I don't see the exact value of this query compared to `UnsupportedExternalAPIs.ql`, except being restricted to generics.

PR #10700, opened 2022-10-06
Ruby: Model some ActiveSupport methods

hmac:
Model many ActiveSupport extensions to `String`, `Hash` and `Object`.

### DIFF
--- a/ruby/ql/lib/codeql/ruby/frameworks/ActiveSupport.qll
+++ b/ruby/ql/lib/codeql/ruby/frameworks/ActiveSupport.qll
@@ -119,16 +119,16 @@ module ActiveSupport {
       }
 
       /**
-       *A flow summary for `Hash#extract!`. This method removes the key/value pairs
-       *matching the given keys from the receiver and returns them (as a Hash).
+       * A flow summary for `Hash#extract!`. This method removes the key/value pairs
+       * matching the given keys from the receiver and returns them (as a Hash).
        *
-       *Example:
+       * Example:
        *
-       *```rb
-       * hash = { a: 1, b: 2, c: 3, d: 4 }
-       * hash.extract!(:a, :b) # => {:a=>1, :b=>2}
-       * hash                  # => {:c=>3, :d=>4}
-       *```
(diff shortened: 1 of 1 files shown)


ruby/ql/lib/codeql/ruby/frameworks/ActiveSupport.qll
```diff
@@ -60,6 +65,90 @@ module ActiveSupport {
...
+          preservesValue = true
+        }
+      }
+
+      private string getExtractComponent(MethodCall mc, int i) {
+        mc.getMethodName() = "extract!" and
+        result = DataFlow::Content::getKnownElementIndex(mc.getArgument(i)).serialize()
+      }
+
+      private class ExtractSummary extends SummarizedCallable {
```
nickrolfe, line 119: This summary is sufficiently complicated that I can't quite follow what it's doing. Could you add a comment, perhaps with an example call to `extract!` and then an explanation of the flow step we're adding?

ruby/ql/lib/codeql/ruby/frameworks/ActiveSupport.qll
```diff
@@ -60,6 +65,90 @@ module ActiveSupport {
...
+            "extract!(" +
+              concat(int i, string s | s = getExtractComponent(mc, i) | s, "," order by i) + ")"
+        }
+
+        final override MethodCall getACall() { result = mc }
+
+        override predicate propagatesFlowExt(string input, string output, boolean preservesValue) {
+          (
+            exists(string s | s = getExtractComponent(mc, _) |
+              input = "Argument[self].Element[" + s + "!]" and
```
nickrolfe, line 134: I"m also unsure what the exclamation marks are for.

hvitved: See https://github.com/github/codeql/pull/10574 for an explanation.

alexrford (dismissed):
Looks great 👍

alexrford (approved):
👍

PR #10704, opened 2022-10-06
RB: add some more meta queries for Ruby evaluations

erik-krogh:
[The evaluation](https://github.com/github/codeql-dca-main/tree/data/erik-krogh/pr-10704-c1fae9__nightly__code-scanning/reports) doesn't make much sense because the queries are introduced.  
I've found these meta-queries nice when adding new taint-steps in JS.


ruby/ql/src/queries/meta/TaintSteps.ql
```diff
@@ -0,0 +1,17 @@
...
+ * @id rb/meta/taint-steps
+ */
+
+import ruby
+import internal.TaintMetrics
+import codeql.ruby.dataflow.internal.TaintTrackingPublic
+
+predicate relevantStep(DataFlow::Node pred, DataFlow::Node succ) { localTaintStep(pred, succ) }
+
+select projectRoot(), count(DataFlow::Node pred, DataFlow::Node succ | relevantStep(pred, succ))
```
asgerf, line 17: This query kind worked in `dist-compare`, but I don't think DCA ever had support for it. Meta queries have to be `@kind problem`.

erik-krogh: And that had too many results and gave OOMs. 
So now I'm counting taint-steps pr. file.

asgerf: I think it's (unfortunately) normal for meta-queries to OOM the report-generator if run as a standalone query, the question is whether the _diff_ is large enough to cause an OOM in practice.

erik-krogh: The baseline is empty in [this new evaluation](https://github.com/github/codeql-dca-main/tree/data/erik-krogh/pr-10704-c1fae9__nightly__code-scanning/reports), and that didn't OOM the report-generator on `nightly.yml`, so that should be good enough.

ruby/ql/src/queries/meta/TaintedNodes.ql
```diff
@@ -0,0 +1,28 @@
...
+import codeql.ruby.TaintTracking
+
+class BasicTaintConfiguration extends TaintTracking::Configuration {
+  BasicTaintConfiguration() { this = "BasicTaintConfiguration" }
+
+  override predicate isSource(DataFlow::Node node) { node = relevantTaintSource(_) }
+
+  override predicate isSink(DataFlow::Node node) {
+    // To reduce noise from synthetic nodes, only count nodes that have an associated expression.
+    exists(node.asExpr().getExpr())
```
hvitved, line 22: So you will only report nodes that can be reached with an empty access path (i.e., where data is not stored into a field). Is that OK? If not, you would have to use partial flow, but I don't know how well that scales.

erik-krogh: > So you will only report nodes that can be reached with an empty access path (i.e., where data is not stored into a field). Is that OK?

Yes, that is on purpose.  
It's marking the nodes that would be flagged in a query, if the node was a sink.  

Sometimes adding a new step can make a node accessible with an empty access-path, where that same node was previously available with a non-empty access-path.  

And with partial flow I would have to set the exploration limit to (basically) infinite, and include the access-path in the alert-message. And I'm pretty sure that won't scale (although I haven't tested).

ruby/ql/src/queries/meta/TaintSteps.ql
asgerf: My concern wasn't that this particular query would OOM, but rather that it isn't working when used on a diff, and it's better to have one that only works on a diff than one that doesn't.

When selecting the steps directly, like in `CallGraph.ql`
- **Can not** generate report when run without a baseline (sadly)
- **Can** see source locations for individual changes
- **Can** see aggregate number of changes at the top of the report
- **Can** see changes when an equal number of results were gained and lost

When selecting the number as part of the alert message:
- **Can** generate report when run without a baseline (although in the [linked evaluation](https://github.com/github/codeql-dca-main/tree/data/erik-krogh/pr-10704-c1fae9__nightly__code-scanning/reports) it doesn't seem to render the alert message for some reason)
- **Can not** see source locations for individual changes
- **Can not** see aggregate number of changes at the top of the report (have to manually subtract numbers from the alert messages)
- **Can not** see changes when an equal number of results were gained and lost

erik-krogh: You're right, the `taint-steps` meta query will probably be confusing to interpret results from as it currently is.  
I don't see how we could fix that, so I'll remove it for now.

asgerf (approved):

PR #10705, opened 2022-10-06
Ruby: Take overrides into account for singleton methods defined on modules

hvitved:
We were previously not handling cases like this:

```rb
class A
    def self.singleton; end
end

class B < A
end
B.singleton # should resolve to B#singleton

class C < A
    def self.singleton; end
end
C.singleton # should resolve to C#singleton
```

DCA shows 20k new call edges, and those I sampled looked legit.

### DIFF
--- a/ruby/ql/lib/codeql/ruby/dataflow/internal/DataFlowDispatch.qll
+++ b/ruby/ql/lib/codeql/ruby/dataflow/internal/DataFlowDispatch.qll
@@ -736,6 +736,10 @@ private predicate singletonMethodOnModule(MethodBase method, string name, Module
   )
 }
 
+/**
+ * Holds if `method` is a singleton method named `name`, defined on module
+ * `m`, or any transitive base class of `m`.
+ */
 pragma[nomagic]
 private MethodBase lookupSingletonMethod(Module m, string name) {
   singletonMethodOnModule(result, name, m)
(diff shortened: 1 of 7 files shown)


ruby/ql/lib/codeql/ruby/dataflow/internal/DataFlowDispatch.qll
```diff
@@ -725,14 +725,33 @@ private predicate singletonMethodOnModule(MethodBase method, string name, Module
...
+  )
   or
   exists(Module other |
     extendCallModule(m, other) and
     method = lookupMethod(other, name)
   )
 }
 
+pragma[nomagic]
+private MethodBase lookupSingletonMethod(Module m, string name) {
```
asgerf, line 740: Could you add some QLDoc indicating that this takes inheritance into account, and likewise on `singletonMethodOnModule`.

I'd probably have named these predicates in a way that clarifies their relationship, such as:
- `lookupOwnSingletonMethod`
- `lookupSingletonMethod`

asgerf (commented):
LGTM with a minor comment

Related to our previous discussion, I noticed some of the new call edges are calls to user-defined `new` methods.

ruby/ql/lib/codeql/ruby/dataflow/internal/DataFlowDispatch.qll
hvitved: I simply used the naming after the existing `lookupMethod` predicate for instance methods. I'll add ql doc.

asgerf (approved):

PR #10706, opened 2022-10-06
C++: Tune cpp/unterminated-variadic-call

geoffw0:
`cpp/unterminated-variadic-call` is a heuristic query - it looks for cases where a variadic call looks like its missing a sentinel to end the variadic arguments (such as a `0` at the end) by comparing with other calls to the same variadic function.  The threshold of calls required to establish such a pattern is currently just 50%, with many real-world false positive results being backed by only 50-70% of calls as evidence - while the query has `@precision medium`.

This PR increases the threshold from 50% to 80%, greatly improving precision, though there will still be some false positive results even after the change.  The query `@precision` is left at `medium`, but I feel it's a more honest `medium` now whereas before this change it should perhaps have been `low`.  The alert message has been slightly expanded to help users understand why the result is being generated.

### DIFF
--- a/cpp/ql/src/Security/CWE/CWE-121/UnterminatedVarargsCall.ql
+++ b/cpp/ql/src/Security/CWE/CWE-121/UnterminatedVarargsCall.ql
@@ -67,11 +67,7 @@ class VarargsFunction extends Function {
     not exists(FunctionCall fc, int index | this.nonTrailingVarArgValue(fc, index) = result)
   }
 
-  predicate isWhitelisted() {
-    this.hasGlobalName("open") or
-    this.hasGlobalName("fcntl") or
-    this.hasGlobalName("ptrace")
-  }
+  predicate isWhitelisted() { this.hasGlobalName(["open", "fcntl", "ptrace", "mremap"]) }
 }
 
 from VarargsFunction f, FunctionCall fc, string terminator, int cnt, int totalCount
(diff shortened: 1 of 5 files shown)


MathiasVP:
The code LGTM! We should probably run a DCA for this before we merge it, though.

geoffw0:
QL-for-QL alerts fixed and DCA run in progress...

geoffw0:
DCA:
 - some performance wobbles, not including `cpp/unterminated-variadic-call` though.
 - we lose 2 alerts for `cpp/unterminated-variadic-call` on `boostorg__boost`.
	- `CALL_MREMAP` is a macro for [mremap](https://man7.org/linux/man-pages/man2/mremap.2.html), which only uses the varargs if a particular flag is specified in one of the fixed parameters - so this is a FP.  I think I'll add this function to the query whitelist actually as this pattern is likely to cause false positives wherever it is used.
	- [detail::convertible<Target*>::check(x)](https://www.boost.org/doc/libs/1_68_0/boost/python/detail/convertible.hpp) appears to only be varargs as some kind of EDG workaround.  It therefore can't use a sentinel and the result is FP.
 - we lose 56 alerts for `cpp/unterminated-variadic-call` on `systemd__systemd`.
	- these also appear to be false positives.

geoffw0:
Added `mremap` to the whitelist / exclusions for this query.  Happy with the DCA results.

MathiasVP (approved):

PR #10712, opened 2022-10-06
Kotlin: allow building a single embeddable plugin version

smowton:


java/kotlin-extractor/build.py
```diff
@@ -233,12 +235,13 @@ def compile_standalone(version):
...
 if args.single_version:
-    compile_standalone(args.single_version)
+    compile_single_version(args.single_version)
 elif args.many:
     for version in kotlin_plugin_versions.many_versions:
         compile_standalone(version)
         compile_embeddable(version)
 else:
-    compile_standalone(kotlin_plugin_versions.get_single_version())
+    compile_single_version(kotlin_plugin_versions.get_single_version())
```
igfoo, line 247: I'm not sure it makes sense to support `--single-version-embeddable` without `--single-version`, in which case `compile_single_version` could be inlined into the `args.single_version` case.

smowton: Done

igfoo (dismissed):

PR #10713, opened 2022-10-06
C++: Fix `getType` for experimental IR dataflow

MathiasVP:
This PR replaces `IRType` with `Type` in the IR-based use-use dataflow library. This change is needed to remain compatible with the `getType` in the old AST-based dataflow library. And in fact, I think using Type here makes more sense since this is often the type people want to use for dataflow queries.

The main point of this PR is done in [the last commit](e147a6032ef2ff1886bf152ed5c2a7592afae686). The first commit is a refactoring PR that should be behavior-preserving, and the second commit is a slight performance change that reduces the number of dataflow nodes generated from IR operands.

### DIFF
--- a/cpp/ql/lib/experimental/semmle/code/cpp/ir/dataflow/internal/DataFlowUtil.qll
+++ b/cpp/ql/lib/experimental/semmle/code/cpp/ir/dataflow/internal/DataFlowUtil.qll
@@ -112,7 +112,7 @@ class Node extends TIRDataFlowNode {
   Declaration getFunction() { none() } // overridden in subclasses
 
   /** Gets the type of this node. */
-  IRType getType() { none() } // overridden in subclasses
+  DataFlowType getType() { none() } // overridden in subclasses
 
   /** Gets the instruction corresponding to this node, if any. */
   Instruction asInstruction() { result = this.(InstructionNode).getInstruction() }
@@ -272,7 +272,7 @@ class Node extends TIRDataFlowNode {
   /**
    * Gets an upper bound on the type of this node.
    */
-  IRType getTypeBound() { result = this.getType() }
+  DataFlowType getTypeBound() { result = this.getType() }
 
   /** Gets the location of this element. */
   cached
@@ -321,7 +321,7 @@ class InstructionNode extends Node, TInstructionNode {
 
   override Declaration getFunction() { result = instr.getEnclosingFunction() }
 
-  override IRType getType() { result = instr.getResultIRType() }
+  override DataFlowType getType() { result = instr.getResultType() }
 
   final override Location getLocationImpl() { result = instr.getLocation() }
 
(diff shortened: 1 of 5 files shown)


cpp/ql/lib/experimental/semmle/code/cpp/ir/dataflow/internal/DataFlowUtil.qll
```diff
@@ -406,7 +425,7 @@ class SsaPhiNode extends Node, TSsaPhiNode {
 
   override Declaration getFunction() { result = phi.getBasicBlock().getEnclosingFunction() }
 
-  override IRType getType() { result instanceof IRVoidType }
+  override DataFlowType getType() { result = this.getAnInput().getType() }
```
jketema, line 428: Was the IRType here wrong? Or has `getType` a different meaning now?

jketema (commented):
The first two commits LGTM. Related to the third commit I've one question below. I also have trouble wrapping my head around the `GLValue` stuff and `Type0`, so I think I need a voiceover for those.

cpp/ql/lib/experimental/semmle/code/cpp/ir/dataflow/internal/DataFlowUtil.qll
MathiasVP: The `IRType` here was very wrong, yes. We want the type of a phi node to properly reflect the values that can flow through here, and in that case, that's any of its possible inputs.

The reason this was `IRVoidType` is that I actually thought that our own queries never looked at the type of a dataflow node (but instead did something like `node.asExpr().getType()`), and thus I didn't really bother implementing it properly. However, I then spotted a few queries (for example [this one](https://github.com/github/codeql/blob/main/cpp/ql/src/Likely%20Bugs/Format/NonConstantFormat.ql#L113)) that actually use it. So it should have a proper implementation.

cpp/ql/lib/experimental/semmle/code/cpp/ir/dataflow/internal/DataFlowUtil.qll
```diff
@@ -859,6 +890,8 @@ abstract class PostUpdateNode extends Node {
    * Gets the node before the state update.
    */
   abstract Node getPreUpdateNode();
+
+  final override Type getType() { result = this.getPreUpdateNode().getType() }
```
jketema, line 894: Is there a reason this returns a `Type` and not a `DataFlowType`?

MathiasVP: Oh. Good point. It should really return a `DataFlowType` (it doesn't really matter since the types are equal, but we should really be consistent).

MathiasVP: Fixed in 5cfc3fe8df5b37daae06755e8f1acae664e15f41.

MathiasVP:
@jketema https://github.com/github/codeql/pull/10713/commits/95e798565b66c651e81a92804f9c9f54818841ae adds a more detailed comment about missing types in the database. I also took the liberty of renaming `getType0` to `getTypeImpl` to limit (but probably not eliminate 😂) confusion about the predicate.

jketema (approved):
LGTM

PR #10714, opened 2022-10-06
Ruby: Model flow through `initialize` constructors

hvitved:
As discussed with @asgerf offline, we have decided to model flow through `initialize` constructors in the following way:

- When we see a call to `C.new(x1, ..., xn)`, pretend that it is actually a call to `C#initialize(x1, ..., xn)`, but only if `C` does not define it's own `self.new` method (in which case we dispatch to that instead).
- Introduce a new return kind, `NewReturnKind`, and
  - let all calls to `new` accept only `NewReturnKind` (and not `NormalReturnKind`),
  - let user-defined `self.new` methods have return kind `NewReturnKind` instead of `NormalReturnKind`, and
  - let `initialize` methods have `NewReturnKind` for all post-update accesses to `self` (and still have `NormalReturnKind` for normal returns, in case someone were to invoke `initialize` directly).

This means that if we have
```rb
class C
  def initialize(x)
    @field = x
  end
end

C.new(y)
```

then `y` will flow into the parameter `x` of `initalize`, and from there to `(post-update) self [field]`. And since `(post-update) self` has `NewReturnKind`, we will get flow back out to `C.new(y) [field]`.

### DIFF
--- a/ruby/ql/lib/codeql/ruby/dataflow/internal/DataFlowDispatch.qll
+++ b/ruby/ql/lib/codeql/ruby/dataflow/internal/DataFlowDispatch.qll
@@ -11,7 +11,8 @@ private import codeql.ruby.dataflow.SSA
 
 newtype TReturnKind =
   TNormalReturnKind() or
-  TBreakReturnKind()
+  TBreakReturnKind() or
+  TNewReturnKind()
 
 /**
  * Gets a node that can read the value returned from `call` with return kind
@@ -41,6 +42,15 @@ class NormalReturnKind extends ReturnKind, TNormalReturnKind {
  */
 class BreakReturnKind extends ReturnKind, TBreakReturnKind {
   override string toString() { result = "break" }
+}
+
+/**
+ * A special return kind that is used to represent the value returned
+ * from user-defined `new` methods as well as the effect on `self` in
+ * `initialize` methods.
+ */
+class NewReturnKind extends ReturnKind, TNewReturnKind {
+  override string toString() { result = "new" }
 }
 
 /** A callable defined in library code, identified by a unique string. */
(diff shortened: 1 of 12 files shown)


ruby/ql/lib/codeql/ruby/dataflow/internal/DataFlowDispatch.qll
```diff
@@ -280,14 +290,28 @@ private predicate hasAdjacentTypeCheckedReads(
   )
 }
 
+/** Holds if `new` is a user-defined `self.new` method. */
+predicate isUserDefinedNew(SingletonMethod new) {
+  new.getName() = "new" and
+  new.getObject().(SelfVariableAccess).getVariable().getDeclaringScope() instanceof ClassDeclaration
```
aibaars, line 296: What about things like
```
def Foo.new
   ...
end
```

aibaars (commented):
LGTM. We might want to add a case for `Class.allocate()` too. The `allocate` method creates a new object, but does not call initialize on it.

ruby/ql/lib/codeql/ruby/dataflow/internal/DataFlowDispatch.qll
```diff
@@ -547,6 +472,12 @@ private DataFlow::LocalSourceNode trackModuleAccess(Module m) {
   result = trackModuleAccess(m, TypeTracker::end())
 }
 
+pragma[nomagic]
+private predicate hasUserDefinedSelf(Module m) {
+  // cannot use `lookupSingletonMethod` due to negative recursion
+  singletonMethodOnModule(_, "new", m.getSuperClass*()) // not `getAnAncestor` because singleton methods cannot be included
```
asgerf, line 478: I suggest renaming this predicate to `hasUserDefinedNew`.

Could it be implemented in terms of `isUserDefinedNew` defined earlier in this file?

----

Also, we seem to lose some call edges due to `C.new` not being seen as an instance of `C` when `C` has an override of `self.new`, but where the implementation of `self.new` always returns `self.allocate`.

I think a reasonably simple fix would be to redefine this predicate to look for overriden `new` that _doesn't_ obviously return `self.allocate`:
```suggestion
  exists(DataFlow::MethodNode method |
    // not `getAnAncestor` because singleton methods cannot be included
    singletonMethodOnModule(method.asCallableAstNode(), "new", m.getSuperClass*()) and
    not method.getSelfParameter().getAMethodCall("allocate").flowsTo(method.getAReturningNode())
  )
```

But I'm OK with postponing this fix so as not to delay the PR any further.

asgerf (approved):
I have some inline suggestions, which are probably best left for future PRs.

Looking at the evaluation results, we lose some call edges and a few sinks, but on the whole we gain much more than we lose. So I'd say it's still better to merge now than to leave the PR hanging.

ruby/ql/lib/codeql/ruby/dataflow/internal/DataFlowDispatch.qll
hvitved: Good spot; thanks for checking. Let's fix follow-up.

ruby/ql/lib/codeql/ruby/dataflow/internal/DataFlowDispatch.qll
```diff
@@ -1083,30 +1172,39 @@ DataFlowCallable viableImplInCallContext(DataFlowCall call, DataFlowCall ctx) {
     exists(RelevantCall call0, Callable res |
       call0 = call.asCall() and
       res = result.asCallable() and
-      result = viableSourceCallable(call) and // make sure to not include e.g. private methods
```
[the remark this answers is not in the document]
hvitved: It is replaced by the `res = getTargetInstance(call0, name)` calls.

asgerf (approved):

PR #10723, opened 2022-10-07
Kotlin: Recognize generated files

tamasvajk:


atorralba (dismissed):

java/ql/lib/semmle/code/java/GeneratedFiles.qll
```diff
@@ -65,3 +65,10 @@ private class GeneratedFileMarker extends Top instanceof JavadocElement {
     )
   }
 }
+
+/**
+ * A file detected as generated by the Kotlin Annotation Processing Tool (kapt). Detection is based on file path.
+ */
+private class KaptFile extends GeneratedFile {
+  KaptFile() { this.getRelativePath().matches("%/generated/source/kapt%") }
+}
```
tamasvajk, line 74: This shouldn't be added. Code scanning already identifies generated files by path, and this case is already covered.

atorralba: Did you find that out by experimenting, or is it documented somewhere?

tamasvajk: I talked to the codescanning team.

atorralba (approved):

PR #10725, opened 2022-10-07
C: fix some more style-guide violations in the alert-messages

erik-krogh:


MathiasVP (approved):
LGTM!

PR #10729, opened 2022-10-07
Py: fix some more style-guide violations in the alert-messages

erik-krogh:

Ignore the QL-for-QL alerts, I'm also changing the alert-messages in those other languages.


python/ql/src/Statements/NonIteratorInForLoop.ql
```diff
@@ -22,4 +22,4 @@ where
   not t.failedInference(_) and
   not v = Value::named("None") and
   not t.isDescriptorType()
-select loop, "$@ of class '$@' may be used in for-loop.", origin, "Non-iterable", t, t.getName()
+select loop, "$@ of class $@ may be used in for-loop.", origin, "Non-iterable", t, t.getName()
```
tausbn, line 25: For consistency, this formulation is probably better.
```suggestion
select loop, "This for-loop may attempt to iterate over a $@ of class $@.", origin, "non-iterable instance", t, t.getName()
```

python/ql/src/Variables/MultiplyDefined.ql
```diff
@@ -67,5 +67,5 @@ where
   forall(Name el | el = asgn1.getParentNode().(Tuple).getAnElt() | multiply_defined(el, _, _)) and
   not uninteresting_definition(asgn1)
 select asgn1,
-  "This assignment to '" + v.getId() +
-    "' is unnecessary as it is redefined $@ before this value is used.", asgn2 as t, "here"
+  "This assignment to '" + v.getId() + "' is unnecessary as $@ before this value is used.", asgn2,
+  "it is redefined"
```
tausbn, line 71:
```suggestion
  "This assignment to '" + v.getId() + "' is unnecessary as it is $@ before this value is used.", asgn2,
  "redefined"
```
This seems more consistent with a/an being pushed out of the interpolated part in many of the other messages.

python/ql/src/experimental/Security/CWE-022/ZipSlip.ql
```diff
@@ -18,5 +18,6 @@ import DataFlow::PathGraph
 
 from ZipSlipConfig config, DataFlow::PathNode source, DataFlow::PathNode sink
 where config.hasFlowPath(source, sink)
-select sink.getNode(), source, sink, "Extraction of zipfile from $@", source.getNode(),
-  "a potentially untrusted source"
+select source.getNode(), source, sink,
+  "Unsanitized archive entry, which may contain '..', is used in a $@.", sink.getNode(),
+  "file system operation"
```
tausbn, line 23: This, to me, seems in want of an article at the beginning ("An unsanitized archive entry ... is used ..." or perhaps even "This ...".). If this is the way these messages are usually formulated, however, then I have no objection.

python/ql/src/experimental/Security/CWE-522/LDAPInsecureAuth.ql
```diff
@@ -16,5 +16,4 @@ import experimental.semmle.python.security.LDAPInsecureAuth
 
 from LdapInsecureAuthConfig config, DataFlow::PathNode source, DataFlow::PathNode sink
 where config.hasFlowPath(source, sink)
-select sink.getNode(), source, sink, "$@ is authenticated insecurely.", sink.getNode(),
-  "This LDAP host"
+select sink.getNode(), source, sink, "This LDAP host is authenticated insecurely."
```
tausbn, line 19: We don't even need `.getNode()` here, do we?

(Maybe we need a QL-for-QL query for this, as it seems to be a bit hit-and-miss throughout our queries.)

tausbn (changes requested):
A bunch of suggestions for minor fixes, but other than that I think this looks good! Thanks!

python/ql/src/experimental/Security/CWE-522/LDAPInsecureAuth.ql
erik-krogh: > We don't even need .getNode() here, do we? 

`.getNode()` or not, it produces the same result. So it doesn't matter.  

We might get a performance diff depending on how much magic is pushed down `getLocation()` in `PathNode`/`DataFlow::Node`, but that's it.

python/ql/src/Security/CWE-798/HardcodedCredentials.ql
```diff
@@ -129,4 +129,5 @@ from HardcodedCredentialsConfiguration config, TaintedPathSource src, TaintedPat
 where
   config.hasFlowPath(src, sink) and
   not any(TestScope test).contains(src.getAstNode())
-select sink.getSink(), src, sink, "Use of $@.", src.getSource(), "hardcoded credentials"
+select sink.getSink(), src, sink, "This hardcoded value is $@.", sink.getNode(),
+  "used as credentials"
```
RasmusWL, line 133: the alert location is still at the sink -- should it be at the source instead?

```suggestion
select src.getNode(), src, sink, "This hardcoded value is $@.", sink.getNode(),
  "used as credentials"
```

python/ql/src/Security/CWE-312/CleartextLogging.ql
```diff
@@ -22,5 +22,5 @@ from Configuration config, DataFlow::PathNode source, DataFlow::PathNode sink, s
 where
   config.hasFlowPath(source, sink) and
   classification = source.getNode().(Source).getClassification()
-select sink.getNode(), source, sink, "$@ is logged here.", source.getNode(),
-  "Sensitive data (" + classification + ")"
+select sink.getNode(), source, sink, "This logs sensitive data returned by $@ as clear text.",
```
RasmusWL, line 25: The final message would contain `sensitive data` twice.. and there's no reason to believe that the source is a call. So my suggestion would be

```diff
-select sink.getNode(), source, sink, "This logs sensitive data returned by $@ as clear text.",
+select sink.getNode(), source, sink, "This logs $@ as clear text.",
```

combining with @tausbn's suggestion:

```suggestion
select sink.getNode(), source, sink, "This expression logs $@ as clear text.",
```

python/ql/src/Security/CWE-295/RequestWithoutValidation.ql
```diff
@@ -22,6 +22,6 @@ where
   // certificate validation, for example in `requests.get(..., verify=arg)`, `arg` would
   // be the `disablingNode`, and the `origin` would be the place were `arg` got its
   // value from.
-  if disablingNode = origin then ending = "." else ending = " by the value from $@."
-select request, "This request may run without certificate validation because it is $@" + ending,
-  disablingNode, "disabled here", origin, "here"
+  if disablingNode = origin then ending = "." else ending = " by $@."
+select request, "This request may run without certificate validation because $@" + ending,
+  disablingNode, "the request is disabled", origin, "this value"
```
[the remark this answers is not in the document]
RasmusWL: Why was this changed? What is the problem with saying that `This request may run without certificate validation because it is disabled`? -- If we want to follow the style guide, you can add `because it is disabled at [this location]` and make `this location` the link to the `disablingNode`

python/ql/src/Security/CWE-209/StackTraceExposure.ql
```diff
@@ -19,6 +19,5 @@ import DataFlow::PathGraph
 
 from Configuration config, DataFlow::PathNode source, DataFlow::PathNode sink
 where config.hasFlowPath(source, sink)
-select sink.getNode(), source, sink,
-  "$@ flows to this location and may be exposed to an external user.", source.getNode(),
-  "Stack trace information"
+select sink.getNode(), source, sink, "This information exposed to the user depends on $@.",
+  source.getNode(), "stack trace information"
```
RasmusWL, line 23: The message sounds strange. Personally, I feel this fits the ``or alternatively (if the first option doesn't work) `[User-provided value] flows to this location and is used in a path`.`` part of the [style guide](https://github.com/github/codeql/blob/main/docs/query-metadata-style-guide.md#alert-messages).

Can we please keep the old message?

RasmusWL (changes requested):

python/ql/src/Security/CWE-295/RequestWithoutValidation.ql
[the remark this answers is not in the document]
erik-krogh: I'm trying to avoid all kinds of "here" links. 

I'll try to go more towards the previous text, but adjusting where the links are. 
How about: `This request may run without certificate validation because [it is disabled].`
And the long version: `This request may run without certificate validation because [it is disabled] by [this value]`.

python/ql/src/experimental/Security/CWE-022/ZipSlip.ql
erik-krogh: I'll add a `This `.

python/ql/src/Security/CWE-209/StackTraceExposure.ql
```diff
@@ -21,4 +21,4 @@ from Configuration config, DataFlow::PathNode source, DataFlow::PathNode sink
 where config.hasFlowPath(source, sink)
 select sink.getNode(), source, sink,
   "$@ flows to this location and may be exposed to an external user.", source.getNode(),
-  "Stack trace information"
+  "stack trace information"
```
RasmusWL, line 24:
```suggestion
  "Stack trace information"
```

RasmusWL (commented):
LGTM besides a small capitalization problem :+1:

tausbn (approved):

PR #10730, opened 2022-10-07
QL: fix some more style-guide violations in the alert-messages

erik-krogh:

And I removed some benign results in e.g. meta queries.


ql/ql/src/queries/performance/TransitiveStep.ql
```diff
@@ -161,5 +161,5 @@ predicate transitiveDelta(Call rec, TransitivelyClosedCall tc) {
 
 from Call rec, TransitivelyClosedCall tc, AstNode reason
 where transitiveDelta(rec, tc) and reason = tc.getReason()
-select tc, "This recursive delta is transively closed $@, which may be a performance problem.",
-  reason, "here"
+select tc, "This recursive delta $@, which may be a performance problem.", reason,
+  "is transively closed"
```
tausbn, line 165:
```suggestion
select tc, "This recursive delta is $@, which may be a performance problem.", reason,
  "transitively closed"
```
Typo (and moved the "is" outside of the link).

ql/ql/src/queries/performance/UnusedField.ql
```diff
@@ -31,5 +31,5 @@ where
     )
   ) and
   (if clz = implClz then extraMsg = "." else extraMsg = " of any class between it and $@.")
-select clz, "The field $@ declared in $@ is not used in the characteristic predicate" + extraMsg,
-  field, field.getName(), clz, clz.getName(), implClz, implClz.getName()
+select clz, "This class declares $@ that is not used in the characteristic predicate" + extraMsg,
+  field, "field " + field.getName(), implClz, implClz.getName()
```
tausbn, line 35:
```suggestion
select clz, "This class declares the $@ but does not bind it in the characteristic predicate" + extraMsg,
  field, "field " + field.getName(), implClz, implClz.getName()
```
Reformulated to read a bit better (to me at least).

ql/ql/src/queries/style/AlertMessage.ql
```diff
@@ -126,7 +126,7 @@ String avoidHere(Select sel, string part) {
 String avoidArticleInLinkText(Select sel) {
   result = sel.getExpr((any(int i | i > 1))) and
   result = getSelectPart(sel, _) and
-  result.getValue().regexpMatch("a|an .*")
+  result.getValue().regexpMatch("(a|an) .*")
```
tausbn, line 129: 😱

ql/ql/src/queries/style/DeadCode.ql
```diff
@@ -12,4 +12,4 @@ import codeql_ql.style.DeadCodeQuery
 
 from AstNode node
 where isDead(node)
-select node, "Code is dead"
+select node, "Code is dead."
```
tausbn, line 15: While I appreciate the nihilism expressed by this statement, perhaps "This code is never used." would be more in line with our other alert messages.

tausbn (changes requested):
A few comments here and there, but otherwise this looks good to me. 👍

ql/ql/src/queries/style/AlertMessage.ql
erik-krogh: Agreed.

tausbn (approved):

PR #10732, opened 2022-10-07
Koltin: Extract the corresponding classes of enum entries

smowton:


java/ql/test/kotlin/library-tests/methods/enumClass.kt
```diff
@@ -2,3 +2,15 @@ enum class EnumClass(val v: Int) {
     enum1(1),
     enum2(1)
 }
+
+enum class EnumWithFunctions {
+
+  VAL {
```
tamasvajk, line 8: Is this becoming a nested class? If so, it might be worth checking if the `enclInReftype` is correctly extracted.

smowton: I checked, it indeed gets enclosed in EnumWithFunctions

tamasvajk (approved):

PR #10737, opened 2022-10-07
Kotlin: fix type variable erasure inside default function values

smowton:
Previously because extractClassInstance didn't use the declaration stack, we wouldn't notice that it was legal to refer to its type variable in the context of extracting a specialised method <-> method source-decl edge. This led to erasing the types of the source-decl, so that e.g. Map.put(...) would have signature (Object, Object) not (K, V) as it should.

### DIFF
--- a/java/ql/test/kotlin/library-tests/parameter-defaults/test.kt
+++ b/java/ql/test/kotlin/library-tests/parameter-defaults/test.kt
@@ -161,3 +161,15 @@ class VisibilityTests {
   private fun i(x: Int, y: Int = 0) = x + y
 
 }
+
+class TestGenericUsedWithinDefaultValue<T> {
+
+  // This tests parameter erasure works properly: we should notice that here the type variable T
+  // isn't used in the specialisation TestGenericUsedWithinDefaultValue<String>, but it can be
+  // cited in contexts like "the signature of the source declaration of 'TestGenericUsedWithinDefaultValue<String>.f(String)' is 'f(T)'",
+  // not 'f(Object)' as we might mistakenly conclude if we're inappropriately erasing 'T'.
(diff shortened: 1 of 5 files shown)


java/ql/test/kotlin/library-tests/parameter-defaults/test.kt
```diff
@@ -161,3 +161,15 @@ class VisibilityTests {
   private fun i(x: Int, y: Int = 0) = x + y
 
 }
+
+class TestGenericUsedWithinDefaultValue<T> {
+
+  // This tests parameter erasure works properly: we should notice that here the type variable T
+  // isn't used in the specialisation TestGenericUsedWithinDefaultValue<String>, but it can be
+  // cited in contexts like "the signature of the source declaration of 'TestGenericUsedWithinDefaultValue<String>.f(String)' is 'f(T)'",
+  // not 'f(Object)' as we might mistakenly conclude if we're inappropriately erasing 'T'.
```
tamasvajk, line 170: How would I see this in QL? Would this mean that `Callable.getSourceDeclaration().getSignature()` would return `ident(T)` on the target of `.ident("Hello world")`?

tamasvajk (commented):
Looking at the PR, I'm unsure if the extractor changes are actually covered by the tests, are they?

java/ql/test/kotlin/library-tests/parameter-defaults/test.kt
smowton: Yes. Without the extractor change we'd produce an inconsistency, trying to target a source-decl ident(Object) that doesn't exist -- you should be able to see it by reverting the .kt change and running the test.

tamasvajk (approved):
Looks plausible to me.

PR #10741, opened 2022-10-09
Ruby: Avoid computing full `fastTC` for `AstNode::getParent`

hvitved:
Rewritten to use explicit recursion instead of (implicit) `fastTC`. See each commit for details.


nickrolfe (approved):
LGTM

PR #10745, opened 2022-10-09
Ruby: Cache use of `DataFlowImplFor(Pathname|HttpClientLibraries)`

hvitved:
As these libraries are used by multiple queries, it makes sense to cache them.


erik-krogh (approved):
[Minimal increase in DB size](https://github.com/github/codeql-dca-main/blob/data/hvitved/pr-10745-efa6b3__nightly__nightly/reports/summaries/any.md#database-size-including-cache-per-source), and [neutral (trending towards better) performance](https://github.com/github/codeql-dca-main/tree/data/hvitved/pr-10745-efa6b3__nightly__nightly/reports#summaries-2).  

LGTM 👍

PR #10746, opened 2022-10-09
Ruby: Add `ActiveJob::Serializers.deserialize` as a code execution sink

alexrford:


ruby/ql/lib/codeql/ruby/frameworks/ActiveJob.qll
```diff
@@ -0,0 +1,30 @@
...
+module ActiveJob {
+  /**
+   * `ActiveJob::Serializers`
+   */
+  module Serializers {
+    /**
+     * A call to `ActiveJob::Serializers.deserialize`, which interprets part of
+     * its argument as a Ruby constant.
+     */
+    class DeserializeCall extends DataFlow::CallNode, CodeExecution::Range {
```
aibaars, line 21: Should this also be a UnsafeDeserialization Sink?

aibaars (dismissed):
Looks good to me. Just wondering if this should also be a sink for the unsafe deserialization query

alexrford:
@nickrolfe I think you originally wrote `rb/unsafe-deserialization` - do you have any thoughts on if these sinks make sense for that query? It's not exactly direct deserialization to an object, but more like an object's type being determined by user-input in the cases that I've seen.

aibaars:
@alexrford Shall we re-run a DCA experiment just in case?

alexrford:
DCA looks uninteresting, merging.

PR #10747, opened 2022-10-10
Ruby: also treat included/prepended modules as subclasses

aibaars:
When resolving a call we should also consider any included/prepended modules as ancestors


ruby/ql/lib/codeql/ruby/dataflow/internal/DataFlowDispatch.qll
```diff
@@ -197,7 +197,7 @@ private predicate instanceMethodCall(CfgNodes::ExprNodes::CallCfgNode call, Modu
     or
     // When we don't know the exact type, it could be any sub class
     exact = false and
-    tp.getSuperClass+() = m
+    [tp.getSuperClass+(), tp.getAnIncludedModule+(), tp.getAPrependedModule+()] = m
```
hvitved, line 200: I guess it should really be
```
tp.getACludedModule+()
```
where `getACludedModule` is the union of `getAnIncludedModule` and `getAPrependedModule`?

ruby/ql/lib/codeql/ruby/frameworks/http_clients/Faraday.qll
```diff
@@ -37,7 +37,8 @@ class FaradayHttpRequest extends Http::Client::Request::Range, DataFlow::CallNod
         API::getTopLevelMember("Faraday").getInstance()
       ] and
     requestNode =
-      connectionNode.getReturn(["get", "head", "delete", "post", "put", "patch", "trace"]) and
+      connectionNode
+          .getReturn(["get", "head", "delete", "post", "put", "patch", "trace", "run_request"]) and
```
hvitved, line 41: Perhaps this change belongs to another PR?

aibaars: I could have made a separate PR for it indeed. It was such a minor change that I left it in .

ruby/ql/lib/codeql/ruby/dataflow/internal/DataFlowDispatch.qll
aibaars: Perhaps even `getASuperCludedModule+` , a perhaps better named `getAnAncestor+`()?

ruby/ql/lib/codeql/ruby/ast/Module.qll
```diff
@@ -23,6 +23,14 @@ class Module extends TModule {
   /** Gets an `include`d module. */
   Module getAnIncludedModule() { result = getAnIncludedModule(this) }
 
+  /** Gets a direct ancestor of this module. */
+  Module getADirectAncestor() {
```
hvitved, line 27: I think we tend to use "immediate" instead of "direct"

ruby/ql/lib/codeql/ruby/ast/Module.qll
```diff
@@ -23,6 +23,14 @@ class Module extends TModule {
...
+    result = [getAnIncludedModule(this), getAPrependedModule(this), getSuperClass(this)]
+  }
+
+  /** Gets an ancestor of this module. */
+  Module getAnAncestor() { result = this.getADirectAncestor+() }
```
hvitved, line 32: I would `pragma[inline]` this.

ruby/ql/lib/codeql/ruby/dataflow/internal/DataFlowDispatch.qll
```diff
@@ -464,9 +461,9 @@ private module Cached {
       )
     )
     or
-    exists(Module superClass, string method |
-      superCall(call, superClass, method) and
-      result = lookupMethod(superClass, method)
+    exists(Module cls, string method |
+      superCall(call, cls, method) and
+      result = lookupMethod(cls.getAnImmediateAncestor+(), method)
```
asgerf, line 466: Shouldn't this just be `getAnImmediateAncestor()` without the `+`? `lookupMethod` looks up in ancestors already.

aibaars: Good point!

asgerf (approved):

PR #10748, opened 2022-10-10
Ruby: TypeTracker: model instance variables as attributes

aibaars:


hvitved:
I think this may clash with @asgerf 's work on local field steps; but I'll let him look into it.

PR #10752, opened 2022-10-10
Python: DB Modeling: Add `pymssql` and `executemany` in general

RasmusWL:
This PR is actually 2 major features:

* 4ee71ae Python: Add support for `pymssql` package
* dba42d6 Python: Model `executemany` on PEP-249 DB APIs

I just stuck them together, since it was only the modeling of `pymssql` that made me realize we were missing the `executemany` calls :flushed:

### DIFF
--- a/python/ql/src/change-notes/2022-10-10-pep249-executemany-modeling.md
+++ b/python/ql/src/change-notes/2022-10-10-pep249-executemany-modeling.md
@@ -1,4 +1,4 @@
 ---
 category: minorAnalysis
 ---
-* Added model of `executemany` calls on PEP-249 compliant Database APIs, resulting in additional sinks for `py/sql-injection`.
+* Added model of `executemany` calls on PEP-249 compliant database APIs, resulting in additional sinks for `py/sql-injection`.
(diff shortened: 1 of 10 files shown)


python/ql/src/change-notes/2022-10-10-pep249-executemany-modeling.md
```diff
@@ -0,0 +1,4 @@
+---
+category: minorAnalysis
+---
+* Added model of `executemany` calls on PEP-249 compliant Database APIs, resulting in additional sinks for `py/sql-injection`.
```
tausbn, line 4: Is the uppercase D in "Database" significant?

tausbn (approved):
One minor comment, otherwise this looks good to me. 👍

python/ql/src/change-notes/2022-10-10-pep249-executemany-modeling.md
RasmusWL: no.

```suggestion
* Added model of `executemany` calls on PEP-249 compliant database APIs, resulting in additional sinks for `py/sql-injection`.
```

tausbn (approved):

PR #10753, opened 2022-10-10
CI: fix qhelp preview

aibaars:
The command to gather the changed files uses NULL character terminated "lines", therefore we should supply the `-z` flag to `basename` as well. Otherwise we end up calling `git grep -l "\n"` which would list all files containing a newline.

This should fix errors on `qhelp.inc` files like in https://github.com/github/codeql/actions/runs/3214105642/jobs/5261655998 .


nickrolfe (approved):
That seems reasonable.

PR #10755, opened 2022-10-10
C#: Draft implementation of using extensible predicates for CSV rows.

michaelnebel:


michaelnebel:
I am struggling a bit to get the extensible predicates to work correctly in packs depending on packs, where the extensible predicates are defined. This comes up for tests.
The following works as expected, when executed from the `ql` root folder
```bash
codeql database run-queries <any_db> csharp/ql/src/codeql-suites/csharp-security-and-quality.qls
```
where as
```bash
codeql test run csharp/ql/test/library-tests/dataflow/external-models/ExternalFlow.ql
```
doesn't appear to be working and yields an error
```
+The query depends on an extensional predicate flowsTo which has not been defined.
```
@aeisenberg : Do you by any chance have an idea for what the problem could be? :-)

aeisenberg:
I'm guessing that the CLI query path is not finding the codeql workspace in the second case. Can you add the root of the codeql repo to `--additional-packs`?

michaelnebel:
> I'm guessing that the CLI query path is not finding the codeql workspace in the second case. Can you add the root of the codeql repo to `--additional-packs`?

I think the cause might be that the data extensions are not loaded as a part of the test execution.
`CodeQLTestCase.java` does not use `ExecuteQueriesCommand.java`.

aeisenberg:
Ah...yes. @asgerf mentioned this as well. I'll take a look this week.

aeisenberg:
Any idea on the performance impact of this change?

michaelnebel:
> Any idea on the performance impact of this change?

Not yet :-)
I will keep you in the loop, when I know more! :-)

PR #10757, opened 2022-10-10
Swift: Query for SQL injection

geoffw0:
Query for SQL injection in Swift.

~Draft: currently there are just test cases, covering the [C API](https://www.sqlite.org/c3ref/funclist.html) that can be [used in Swift](https://developer.apple.com/documentation/swift/using-imported-c-functions-in-swift), and the [SQLite.swift](https://github.com/stephencelis/SQLite.swift) wrapper that I believe is commonly used as well.  I'd like someone to review these test cases for realism / misconceptions / any important details I've missed before going too much further.~

@atorralba would you mind taking a look?


swift/ql/test/query-tests/Security/CWE-089/sqlite3_c_api.swift
```diff
@@ -0,0 +1,163 @@
...
+			let result = sqlite3_step(stmt3)
+			// ...
+		}
+	}
+	sqlite3_finalize(stmt3)
+
+	var stmt4: OpaquePointer?
+
+	if (sqlite3_prepare_v2(db, varQuery, -1, &stmt4, nil) == SQLITE_OK) { // GOOD
+		if (sqlite3_bind_text(stmt4, 1, remoteString, -1, SQLITE_TRANSIENT) == SQLITE_OK) { // GOOD???
```
atorralba, line 155: This is safe indeed. That's the proper way of building SQL statements with untrusted data: using prepared statements and binding the untrusted data to the statement with the appropriate type.

swift/ql/test/query-tests/Security/CWE-089/sqlite3_c_api.swift
```diff
@@ -0,0 +1,163 @@
...
+
+func sqlite3_step(
+		_ _: OpaquePointer?
+	) -> Int32 { return SQLITE_OK }
+
+func sqlite3_finalize(
+		_ pStmt: OpaquePointer?
+	) -> Int32 { return SQLITE_OK }
+
+func sanitize(_ string: String) -> String { return string }
```
atorralba, line 98: I'm not sure how `sanitize` should work, though. Is it a test-only function to test sanitizers in general, or do we expect the query to recognize any function called `sanitize` or similar as a valid sanitizer?

atorralba (commented):
The tests make sense to me. `sqlite3_exec` and `sqlite3_prepare*` seem to be the appropriate sinks, because they handle raw SQL statements. `sqlite3_bind*` handle the parameters of prepared statements, which is safe to do even with untrusted data.

This should cover the raw `sqlite3` library. Are we planning on covering wrapper libraries on top of it? e.g. https://github.com/stephencelis/SQLite.swift or https://github.com/groue/GRDB.swift.

geoffw0:
> Are we planning on covering wrapper libraries on top of it? e.g. https://github.com/stephencelis/SQLite.swift

The tests in `SQLite.swift` should cover this library - assuming I've correctly identified the right / most important interfaces.

> or https://github.com/groue/GRDB.swift.

Yeah it would be good to cover this library as well.  However I'm a bit behind on this task as it is, so I'll create a follow-up ticket rather than addressing this now.

swift/ql/test/query-tests/Security/CWE-089/sqlite3_c_api.swift
geoffw0: Thanks, good to confirm this is the right way.

swift/ql/test/query-tests/Security/CWE-089/sqlite3_c_api.swift
geoffw0: The query is likely to recognize functions using a heuristic such as the function name, so matching functions called `sanitize`, `filter` etc (`escape`, perhaps).  The tricky part is that's likely to include some bad sanitizers (like just adding quotes to the ends of the string, or escaping some but not all important characters) that we would rather flag somehow.

geoffw0: I see that in Java we have a different approach - `PrimitiveType`, `BoxedType`, `NumberType` are sanitizers.  I assume this is because they can't possibly contain or expand to free text as would be required for an injection attack.

geoffw0: I'm unclear whether it's *possible* to write a safe sanitizer for SQL, or whether in fact the prepared statement approach is the only 'right' way?

atorralba: You _technically_ could, but it's not recommended because it's easy to do it wrong. In general, the suggested approach always is using prepared statements (or wrapper libraries that do not handle raw SQL) when possible, or otherwise using a well-established security library to do the sanitization/validation for you.

swift/ql/test/query-tests/Security/CWE-089/SQLite.swift
```diff
@@ -0,0 +1,79 @@
...
+	let stmt1 = try db.prepare(unsafeQuery3) // BAD
+	try stmt1.run()
+
+	let stmt2 = try db.prepare(varQuery, localString) // GOOD
+	try stmt2.run()
+
+	let stmt3 = try db.prepare(varQuery, sanitizedString) // GOOD
+	try stmt3.run()
+
+	let stmt4 = try db.prepare(varQuery, remoteString) // GOOD???
```
atorralba, line 75: Same reason as in `sqlite3_c_api.swift`, this is safe.

swift/ql/test/query-tests/Security/CWE-089/SQLite.swift
```diff
@@ -0,0 +1,79 @@
...
+	try stmt4.run()
+
+	// TODO: test all versions of prepare, run, scalar on Connection and Statement
```
atorralba, line 78: This should include one of `unsafeQuery*` being used as first parameter of the multi-parameter versions of `prepare()`, which would be `BAD`.

I guess you are already aware, but `Connection.(run|bind|scalar)`'s first argument is a sink, whereas `Statement.(bind|run|scalar)`'s arguments are all safe. The string parameter of `Statement`'s constructor is also a sink.

atorralba (commented):
> The tests in SQLite.swift should cover this library - assuming I've correctly identified the right / most important interfaces.

D'oh, true, sorry.

geoffw0:
I've removed draft status from this PR, it's ready to review.

I'd like to acknowledge a few issues though.  IMO none of these prevent us from merging what we have:
 - we're getting no results on the C API test due to an issue with taint flow, possibly itself caused by an extractor issue.  I'm discussing this with other members of the team and I'm not sure if the fix will be something simple I can do here, or something that will need to be done as follow-up.
 - there are no sanitizers for remote data in numerical form yet (considered to be safe).  This is because in my test case taint doesn't flow into `remoteNumber` anyway, so there's no way to test such a sanitizer (and no strict need for it yet).
 - we may also want sanitizers along the lines of Ruby's `StringConstCompareBarrier` and `StringConstArrayInclusionCallBarrier`.  I might write a follow-up issue for this but I'm inclined not to worry too much until we see patterns like those in real Swift code (they are talked about in https://cheatsheetseries.owasp.org/cheatsheets/SQL_Injection_Prevention_Cheat_Sheet.html).

geoffw0:
I've just merged main into this PR - with the extractor fix https://github.com/github/codeql/pull/10854 that gives us the C API results for this query.  We still don't get results through the UTF-16 interface, but that's much less concerning than missing all C API results.

I'd be happy to see this reviewed and merged, and deal with the remaining known issues as follow-up.

swift/ql/src/queries/Security/CWE-089/SqlInjection.ql
```diff
@@ -0,0 +1,83 @@
...
+      f.getName() = ["execute(_:)", "prepare(_:_:)", "run(_:_:)", "scalar(_:_:)"] and
+      call.getStaticTarget() = f and
+      call.getArgument(0).getExpr() = this.asExpr()
+    )
+    or
+    // String argument to the `Statement` constructor.
+    exists(ClassDecl c, AbstractFunctionDecl f, CallExpr call |
+      c.getName() = "Statement" and
+      c.getAMember() = f and
+      f.getName() = "init(_:_:)" and
```
jketema, line 62: Just for my understanding: `init` is the name assigned to the constructor by the extractor?

swift/ql/test/query-tests/Security/CWE-089/SQLite.swift
```diff
@@ -0,0 +1,133 @@
...
+	let stmt9 = try db.prepare(varQuery) // GOOD
+	stmt9.bind(remoteString)
+	stmt9.bind([remoteString])
+	stmt9.bind(["username": remoteString])
+	try stmt9.run(remoteString)
+	try stmt9.run([remoteString])
+	try stmt9.run(["username": remoteString])
+	try stmt9.scalar(remoteString)
+	try stmt9.scalar([remoteString])
+	try stmt9.scalar(["username": remoteString])
```
jketema, line 130: These I do not quite understand, even after looking at the SQLite.swift docs. Shouldn't these be tagged with `// GOOD`, or am I missing something?

jketema (approved):
Two small questions. Otherwise this looks ok to me.

jketema (commented):
I'm happy with the current version, so just waiting for the doc review.

swift/ql/src/queries/Security/CWE-089/SqlInjection.qhelp
```diff
@@ -0,0 +1,38 @@
+<!DOCTYPE qhelp PUBLIC
+  "-//Semmle//qhelp//EN"
+  "qhelp.dtd">
+<qhelp>
+<overview>
+
+<p>
+If a database query (such as an SQL query) is built from user-provided data without sufficient sanitization, a user may be able to run malicious database queries.
```
subatoi, line 8:
```suggestion
If a database query (such as a SQL query) is built from user-provided data without sufficient sanitization, a user may be able to run malicious database queries.
```

[MS Style guide treats as an acronym](https://learn.microsoft.com/en-us/style-guide/a-z-word-list-term-collections/s/sql-sql-server) ("a SQL...")

swift/ql/src/queries/Security/CWE-089/SqlInjection.qhelp
```diff
@@ -0,0 +1,38 @@
...
+<recommendation>
+
+<p>
+Most database connector libraries offer a way of safely embedding untrusted data into a query by means of query parameters or prepared statements. Use these features rather than building queries by string concatenation.
+</p>
+
+</recommendation>
+<example>
+
+<p>In the following example, an SQL query is prepared using string interpolation to directly include a user-controlled value <code>userControlledString</code> in the query. An attacker could craft the part they control to change the overall meaning of the SQL query.
```
subatoi, line 21:
```suggestion
<p>In the following example, a SQL query is prepared using string interpolation to directly include a user-controlled value <code>userControlledString</code> in the query. An attacker could craft the part they control to change the overall meaning of the SQL query.
```

(as above)

swift/ql/src/queries/Security/CWE-089/SqlInjection.qhelp
```diff
@@ -0,0 +1,38 @@
...
+</p>
+
+<sample src="SqlInjectionBad.swift" />
+
+<p>A better way to do this is with a prepared statement, binding <code>userControlledString</code> to a parameter of that statement. An attacker who controls the contents of that parameter cannot 'break out' and change the overall meaning of the SQL query.
```
subatoi, line 26: Is there a different way of expressing the "break out" part? Unless this is a technical term related to SQL then it'd be preferable to avoid metaphors and expressions (basically anything in inverted commas) so as to help comprehension, particularly for people for whom English is a second language 🙏

swift/ql/src/queries/Security/CWE-089/SqlInjection.qhelp
subatoi, line 21: I hope this isn't adding too much overhead but could you kindly explain what's meant by "the part they control" (the attacker)? The reason I ask is because this paragraph feels like the central part of the whole piece and some feedback I have might or might not be relevant.

swift/ql/src/queries/Security/CWE-089/SqlInjection.qhelp
```diff
@@ -0,0 +1,38 @@
...
+
+<p>
+If a database query (such as an SQL query) is built from user-provided data without sufficient sanitization, a user may be able to run malicious database queries.
+</p>
+
+</overview>
+<recommendation>
+
+<p>
+Most database connector libraries offer a way of safely embedding untrusted data into a query by means of query parameters or prepared statements. Use these features rather than building queries by string concatenation.
```
subatoi, line 15: In the context, is "user-provided data without sufficient sanitization" equivalent to "building queries by string concatenation"?

subatoi (commented):
Hi there @geoffw0, I'm the first responder for the Docs team this week so I've taken a look at this and left some small comments and asked a couple of minor questions. I hope this isn't adding too much overhead 🙏 

I'll just tag my colleague @felicitymay as well just in case (might not be necessary but I'm still quite new to reviewing CodeQL! 😁 )

swift/ql/src/queries/Security/CWE-089/SqlInjection.qhelp
geoffw0: Ah, I've seen that elsewhere and thought it was a mistake!  I've accepted your changes.

swift/ql/src/queries/Security/CWE-089/SqlInjection.qhelp
geoffw0: Yes this is the same idea, though they're not quite formally equivalent.  Concatenation will be by far the most common case.  I'll think about whether we can be more consistent describing this.

swift/ql/src/queries/Security/CWE-089/SqlInjection.qhelp
geoffw0: It means the part of the concatenated string that comes from the user.  I agree it may be worth moving this bit to sooner in the document.

swift/ql/src/queries/Security/CWE-089/SqlInjection.qhelp
geoffw0: Yeah, I can see that this might be confusing for someone who doesn't speak English as a first language.  I've removed the metaphor.

swift/ql/src/queries/Security/CWE-089/SqlInjection.qhelp
geoffw0: I've added something very similar to this point to the opening paragraph of the doc.  There's a bit of repetition but that may not be a bad thing.  I've also rephrased "the part they control" to be a bit more explicit.

geoffw0:
Thanks for the review @subatoi .  I believe I've addressed all your points, but please don't hesitate to say if you think it needs more work.

swift/ql/src/queries/Security/CWE-089/SqlInjection.qhelp
```diff
@@ -0,0 +1,38 @@
...
+
+<p>
+If a database query (such as a SQL query) is built from user-provided data without sufficient sanitization, a user may be able to run malicious database queries. An attacker can craft the part of the query they control to change the overall meaning of the query.
+</p>
+
+</overview>
+<recommendation>
+
+<p>
+Most database connector libraries offer a way of safely embedding untrusted data into a query by means of query parameters or prepared statements. Use these features rather than building queries by string concatenation or similar methods without sufficient sanitization.
```
subatoi, line 15:
```suggestion
Most database connector libraries offer a way to safely embed untrusted data into a query using query parameters or prepared statements. You should use these features to build queries, rather than string concatenation or similar methods without sufficient sanitization.
```

swift/ql/src/queries/Security/CWE-089/SqlInjection.qhelp
```diff
@@ -0,0 +1,38 @@
...
+</p>
+
+</recommendation>
+<example>
+
+<p>In the following example, a SQL query is prepared using string interpolation to directly include a user-controlled value <code>userControlledString</code> in the query. An attacker could craft <code>userControlledString</code> in such a way as to change the overall meaning of the SQL query.
```
subatoi, line 21:
```suggestion
<p>In the following example, a SQL query is prepared using string interpolation to directly include a user-controlled value <code>userControlledString</code> in the query. An attacker could craft <code>userControlledString</code> to change the overall meaning of the SQL query.
```

subatoi (approved):
Just added a couple of minor last points to make the language a bit simpler. There's nothing wrong with the existing language; these suggestions are just minor and style-related. Otherwise approving and LGTM. Thank you!

swift/ql/src/queries/Security/CWE-089/SqlInjection.qhelp
geoffw0: Sometimes no words are the best words. :)

geoffw0:
Thanks - those suggestions definitely help make things read a bit more easily.  Please approve the PR if you're happy with the docs now (the code has already been reviewed and approved).

subatoi (approved):
Thanks @geoffw0, apologies, didn't realise my review would be dismissed. LGTM!

geoffw0:
Thanks again for the review, suggestions and approval.  On to the next query!

PR #10758, opened 2022-10-10
Type tracking: Split up `levelStep` into `levelStepCall` and `levelStepNoCall`

hvitved:
To reduce non-linear recursion during call graph construction, like we already have `flowStepCall` and `flowStepNoCall`.

### DIFF
--- a/python/ql/lib/semmle/python/dataflow/new/internal/TypeTrackerSpecific.qll
+++ b/python/ql/lib/semmle/python/dataflow/new/internal/TypeTrackerSpecific.qll
@@ -46,7 +46,7 @@ predicate simpleLocalFlowStep = DataFlowPrivate::simpleLocalFlowStepForTypetrack
 predicate jumpStep = DataFlowPrivate::jumpStepSharedWithTypeTracker/2;
 
 /** Holds if there is a level step from `nodeFrom` to `nodeTo`, which may depend on the call graph. */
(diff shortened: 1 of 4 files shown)


python/ql/lib/semmle/python/dataflow/new/internal/TypeTrackerSpecific.qll
```diff
@@ -45,8 +45,11 @@ predicate simpleLocalFlowStep = DataFlowPrivate::simpleLocalFlowStepForTypetrack
 
 predicate jumpStep = DataFlowPrivate::jumpStepSharedWithTypeTracker/2;
 
-/** Holds if there is a level step from `pred` to `succ`. */
-predicate levelStep(Node pred, Node succ) { none() }
+/** Holds if there is a level step from `nodeFrom` to `nodeTo`, which may depend on the call graph. */
+predicate levelStepCall(Node pred, Node succ) { none() }
```
tausbn, line 49:
```suggestion
predicate levelStepCall(Node nodeFrom, Node nodeTo) { none() }
```
For consistency.

tausbn (changes requested):
Looks good to me. One minor suggestion.

hvitved:
Rebased to resolve merge conflict.

asgerf (approved):
Also LGTM

PR #10760, opened 2022-10-10
Ruby: Restrict regexp taint flow to `String` summaries

hvitved:
Allowing arbitrary taint flow when figuring out which string literals may flow to a `Regexp.compile` (et al) call seems a bit too general, so this PR restricts taint flow to only include flow summaries from `core/String.qll`. The missing `rb/incomplete-hostname-regexp` results were false positives, as they had flow that went through `Regexp.escape`.


aibaars (dismissed):
This looks good to me. However, wouldn't it be clearer to switch to value flow and add string summaries and string append (`+`) as additional flow steps?

hvitved:
> This looks good to me. However, wouldn't it be clearer to switch to value flow and add string summaries and string append (`+`) as additional flow steps?

The trouble with that is that there is currently no easy way to add ad-hoc flow through `SummarizedCallable`s; we only have the possibility to add ad-hoc flow steps, but for flow summaries we actually need a bunch of steps (flow into, flow inside, and flow out).

hvitved:
> The trouble with that is that there is currently no easy way to add ad-hoc flow through `SummarizedCallable`s; we only have the possibility to add ad-hoc flow steps, but for flow summaries we actually need a bunch of steps (flow into, flow inside, and flow out).

Actually, on second thought this may be doable.

aibaars (approved):

PR #10764, opened 2022-10-11
JS: Consider other XSS unsafe content-types when reasoning about XSS vulnerabilities

pwntester:
This PRs includes new content-types that may lead to XSS vulnerabilities as per https://portswigger.net/web-security/cross-site-scripting/cheat-sheet#content-types


javascript/ql/lib/semmle/javascript/security/dataflow/ReflectedXssCustomizations.qll
```diff
@@ -25,28 +25,43 @@ module ReflectedXss {
...
-    HttpResponseSink() { not exists(getANonHtmlHeaderDefinition(this)) }
+    HttpResponseSink() { not exists(getAXssSafeHeaderDefinition(this)) }
   }
 
   /**
-   * Gets a HeaderDefinition that defines a non-html content-type for `send`.
+   * Gets a HeaderDefinition that defines a XSS safe content-type for `send`.
    */
-  Http::HeaderDefinition getANonHtmlHeaderDefinition(Http::ResponseSendArgument send) {
+  Http::HeaderDefinition getAXssSafeHeaderDefinition(Http::ResponseSendArgument send) {
```
erik-krogh, line 34: Could you keep the old versions around as `deprecated` predicates?  
`Customizations.qll` files should follow our deprecation policy.

pwntester: Sure, let me know if the changes are correct

erik-krogh: They look good 👍

erik-krogh (commented):
LGTM. 

But could you add a test that showcases a new result that is found by this change?  

I've already done [an evaluation](https://github.com/github/codeql-dca-main/tree/data/erik-krogh/pr-10764-default-reflectedXss/reports) of your change, and it seems fine.   
(No performance regression, but also no new results in the projects I tried).

pwntester:
@erik-krogh Just added a test, but Im not sure is in the expected form. Also, how should I recreate the expected file?

pwntester:
> It's an OK test, but to me it looks like results that were also flagged with the previous version of the query?
> You should add something that wasn't flagged by the old version of the query.

the `application/xml` endpoint should not be generating a XSS alert before this change. So now both `text/html` and `application/xml` should get a XSS alert and `text/plain` should not

pwntester:
@erik-krogh re: the last commit, I found a XSS FPs in a handler similar to this:
```javascript
function test4(req, context) { // test: handler
  const body = req.body // test: source
  const newPackument = body['package-json']
  const message = `INFO: User invited to package ${newPackument._id} successfully.`
  return reply(message, 200, { 'npm-notice': message }) // test: stackTraceExposureSink, !xssSink, !xss
}
```

The reason is that the `reply` call (similar to express' send) is setting the `npm-notice` header, and this was considered as the dominating header event if it was not a `content-type` header. This was preventing another implicit header definition that I added for all handlers (`application/json`)

PR #10768, opened 2022-10-11
JS: fix that js/file-system-race could have FPs related to loops

erik-krogh:
[Evaluation shows that we loose a bunch of results](https://github.com/github/codeql-dca-main/tree/data/erik-krogh/pr-10768-7500a3__nightly-old__CustomSuite/reports), but that was the point.  
In the lost results the file-check doesn't dominate the file-access.  
In most cases that's from the file-check being inside a conditional.  

The point of the PR was originally just to remove FPs from the check/access being inside a loop.   
But I like removing these other results too, so I want to keep this.


asgerf (approved):
I think it's worth adding a change note for this, otherwise LGTM

asgerf (approved):

PR #10769, opened 2022-10-11
C#: Include CIL SSA definitions in `DataFlow::Node`

hvitved:
This PR adds CIL SSA definitions to `DataFlow::Node`, just like we already have SSA definitions for source code.

The motivation for adding SSA definitions into `DataFlow::Node` in the first place is to avoid having to compute all phi-node input/output combinations up front. For example, if we have

```mermaid
  graph TD;
      A-->phi;
      B-->phi;
      C-->phi;
      phi-->D;
      phi-->E;
      phi-->F;
```

then we would have to generate 9 flow steps (quadratic), as opposed to 6 (linear).

### DIFF
--- a/csharp/ql/lib/semmle/code/cil/Stubs.qll
+++ b/csharp/ql/lib/semmle/code/cil/Stubs.qll
@@ -29,14 +29,17 @@ private module Cached {
 
   cached
   predicate bestImplementation(MethodImplementation mi) {
-    not assemblyIsStubImpl(mi.getLocation()) and
-    not exists(MethodImplementation better | mi.getMethod() = better.getMethod() |
-      mi.getNumberOfInstructions() < better.getNumberOfInstructions()
-      or
-      mi.getNumberOfInstructions() = better.getNumberOfInstructions() and
-      mi.getLocation().getFile().toString() > better.getLocation().getFile().toString()
-    ) and
-    exists(mi.getAnInstruction())
+    exists(Assembly asm |
+      asm = mi.getLocation() and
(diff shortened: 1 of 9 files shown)


csharp/ql/lib/semmle/code/cil/Stubs.qll
```diff
@@ -29,14 +29,17 @@ private module Cached {
...
-    not exists(MethodImplementation better | mi.getMethod() = better.getMethod() |
-      mi.getNumberOfInstructions() < better.getNumberOfInstructions()
-      or
-      mi.getNumberOfInstructions() = better.getNumberOfInstructions() and
-      mi.getLocation().getFile().toString() > better.getLocation().getFile().toString()
-    ) and
-    exists(mi.getAnInstruction())
+    exists(Assembly asm |
+      asm = mi.getLocation() and
+      (assemblyIsStubImpl(asm) implies asm.getFile().extractedQlTest()) and
```
michaelnebel, line 34: Why are assembly stub implementations taken into considerations now (looks like it is relevant for test)?

michaelnebel (approved):
Looks plausible to me! :-)

csharp/ql/lib/semmle/code/cil/Stubs.qll
hvitved: Yeah, this is only relevant for tests. One of our test assemblies would otherwise incorrectly be considered a stub, and therefore not be included in `TCilExprNode`.

PR #10773, opened 2022-10-11
Ruby: bugfix in type-tracking singleton class resolution

asgerf:
Fixes a bug in how the type-tracking associates a singleton class with its target module.

[Evaluation](https://github.com/github/codeql-dca-main/issues/7754) looks good, about 4k new call edges, and neutral performance. There are 4 new alerts from `rb/weak-cryptographic-algorithm` which correctly point out that DES is used, although it seems intentional in that codebase.


hvitved (approved):
Good spot!

PR #10774, opened 2022-10-11
Swift: Add summaries for tainted URL fields

atorralba:
If a URL is tainted, so are its fields.


swift/ql/lib/codeql/swift/frameworks/StandardLibrary/Url.qll
```diff
@@ -22,7 +22,41 @@ private class UrlSummaries extends SummaryModelCsv {
...
+        ";URL;true;init(string:relativeTo:);(String,URL?);;Argument[0,1];ReturnValue.Field[pathExtension];taint",
+        ";URL;true;init(string:relativeTo:);(String,URL?);;Argument[0,1];ReturnValue.Field[port];taint",
+        ";URL;true;init(string:relativeTo:);(String,URL?);;Argument[0,1];ReturnValue.Field[query];taint",
+        ";URL;true;init(string:relativeTo:);(String,URL?);;Argument[0,1];ReturnValue.Field[relativePath];taint",
+        ";URL;true;init(string:relativeTo:);(String,URL?);;Argument[0,1];ReturnValue.Field[relativeString];taint",
+        ";URL;true;init(string:relativeTo:);(String,URL?);;Argument[0,1];ReturnValue.Field[scheme];taint",
+        ";URL;true;init(string:relativeTo:);(String,URL?);;Argument[0,1];ReturnValue.Field[standardized];taint",
+        ";URL;true;init(string:relativeTo:);(String,URL?);;Argument[0,1];ReturnValue.Field[standardizedFileURL];taint",
+        ";URL;true;init(string:relativeTo:);(String,URL?);;Argument[0,1];ReturnValue.Field[user];taint",
+        ";URL;true;init(string:relativeTo:);(String,URL?);;Argument[0,1];ReturnValue.Field[password];taint",
```
geoffw0, line 59: It would be good to have a couple of comments explaining what blocks of these rows do. 
 Something like your PR description:
```
If a URL is tainted, so are its fields.
```

swift/ql/test/library-tests/dataflow/taint/url.swift
```diff
@@ -41,19 +59,36 @@ func taintThroughURL() {
...
+	sink(arg: urlTainted.standardizedFileURL) // $ tainted=57
+	sink(string: urlTainted.user!) // $ tainted=57
+	sink(string: urlTainted.password!) // $ tainted=57
 
 	sink(arg: URL(string: clean, relativeTo: nil)!)
-	sink(arg: URL(string: tainted, relativeTo: nil)!) // $ tainted=39
+	sink(arg: URL(string: tainted, relativeTo: nil)!) // $ tainted=57
 	sink(arg: URL(string: clean, relativeTo: urlClean)!)
-	sink(arg: URL(string: clean, relativeTo: urlTainted)!) // $ tainted=39
+	sink(arg: URL(string: clean, relativeTo: urlTainted)!) // $ tainted=57
```
geoffw0, line 84: I think we could do with at least one test for a field after `init(string:relativeTo:)`.

geoffw0 (approved):
LGTM, two very minor suggestions.

swift/ql/lib/codeql/swift/frameworks/StandardLibrary/Url.qll
atorralba: Good suggestion, applied in 1e48500, thanks!

swift/ql/test/library-tests/dataflow/taint/url.swift
atorralba: This actually made me think more detailedly about this constructor, and after testing it, the conclusion is:

* if the `string:` argument is an absolute URL, basically it acts as the 1-argument constructor and `relativeTo:` is ignored (but `baseURL` is tainted by it).
* if the `string:` argument is a relative path:
  * the path, query, and fragment related fields are tainted by it.
  * all the other fields are tainted by `relativeTo:`.

I adjusted the CSV rows to model this behavior as closely as possible, but of course the conditional logic is not handled. If we wanted more precise models we would need to switch to regular CodeQL instead, and apply some heuristics to the first argument to determine whether it's an absolute URL or a relative path.

What's your opinion about that?

atorralba:
@geoffw0 I decided to handle this with `TaintInheritingContent` in the end: it introduces some spurious flow (marked as such in the tests), but simplifies the implementation _and_ handles a generally tainted `URL` correctly (i.e. taints its fields too). In case we want more precision in the future (if we start seeing FPs because of this), I think we should apply the heuristics I mentioned [in my previous comment](https://github.com/github/codeql/pull/10774#discussion_r996833954). But I think this should be good enough for the moment.

swift/ql/test/library-tests/dataflow/taint/url.swift
```diff
@@ -61,7 +61,8 @@ func taintThroughURL() {
 	sink(arg: urlClean)
 	sink(arg: urlTainted) // $ tainted=57
 	sink(arg: urlTainted.absoluteURL) // $ tainted=57
-	sink(arg: urlTainted.baseURL) // $ tainted=57
+	sink(arg: urlTainted.baseURL) // $ Safe
```
geoffw0: Ah that's interesting, so if I'm understanding correctly `URL("http://example.com/abc")` and `URL("abc", URL("http://example.com/"))` are not two ways of constructing exactly the same thing - because the latter has a `baseURL` (amongst other visible differences).

geoffw0 (approved):
> I decided to handle this with TaintInheritingContent in the end: it introduces some spurious flow (marked as such in the tests)

This seems fine to me.  I can't really imagine someone constructing a part-tainted, part-clean URL and then depending on that for correct safe behaviour afterwards.

swift/ql/test/library-tests/dataflow/taint/url.swift
atorralba: Correct. It seems that if the first argument is a relative path it only overrides the path, query, and fragment parts of the `relativeTo` URL, but then the original is saved as `baseURL`. With the one-argument constructor `baseURL` is apparently never populated.

PR #10777, opened 2022-10-11
C#: Generate data extension files

michaelnebel:
The purpose of this PR is to use *data extensions* for MaD models instead of inlining them in the Code QL source code for C#.

To that end we
- Introduce a script that produces data extension (YAML) files for the existing hand-written models. The scripts can be used in the future, if we want to change the generated YAML for our existing hand-written models.
- Introduce a new model generator scripts that produces a data extension YAML file instead of a Code QL source code file.
- Introduce the relevant extensible predicates in `ExternalFlow.qll`.
- Add hand-written models as data extensions and cleanup the Code QL source code for hand-written MaD models.
- Add generated models as data extensions for .NET runtime (including negative models) and cleanup the Code QL source for generated and negative models.

Before making any changes the generated .NET runtime models were refreshed as a part of this PR.


aeisenberg:
cc: @cklin, In previous runs, the C# Language Tests job failed with OOME almost certainly due to 50k tuples from data extensions. We need to think of some ways of being more memory efficient when we read in the tuples.

aeisenberg:
@michaelnebel just for fun...could you try adding `--ram=2048` (or higher) to [`integration-tests-runner.py`](https://github.com/github/semmle-code/blob/6a8c4e91b09433abfcc7869693d322763931d3bc/buildutils-internal/sembuild/integration-tests-runner.py#L504). This _might_ work, but since the test runs things in slices, I don't know if this will require too much memory on the underpowered actions vms.

csharp/ql/lib/semmle/code/csharp/dataflow/internal/DataFlowPrivate.qll
```diff
@@ -2212,6 +2212,15 @@ predicate allowParameterReturnInSelf(ParameterNode p) {
   FlowSummaryImpl::Private::summaryAllowParameterReturnInSelf(p)
 }
 
+/**
+ * A module importing the modules that provide synthetic field declarations,
+ * ensuring that they are visible to the taint tracking / data flow library.
+ */
+private module SyntheticFields {
+  private import semmle.code.csharp.frameworks.system.threading.Tasks
+  private import semmle.code.csharp.frameworks.system.runtime.CompilerServices
```
hvitved, line 2221: This files only had code deleted, so why is this needed now and not before (perhaps it was)?

csharp/ql/src/utils/modelconverter/project/Program.cs
```diff
@@ -0,0 +1,2 @@
+﻿// See https://aka.ms/new-console-template for more information
+Console.WriteLine("Hello, World!");
```
hvitved, line 2: Perhaps let your script do `dotnet new` instead, to avoid having to check in this file.

hvitved (commented):
Approach LGTM, and great work on converting all existing flow summaries from `.qll` to `.yml`.

csharp/ql/lib/semmle/code/csharp/dataflow/ExternalFlow.qll
```diff
@@ -209,15 +219,43 @@ predicate sourceModel(
     row.splitAt(";", 7) = kind and
     row.splitAt(";", 8) = provenance
   )
+  or
+  extSourceModel(namespace, type, subtypes, name, signature, ext, output, kind, provenance)
+}
+
+/** Holds if `row` is a source model. */
+predicate sourceModel(string row) {
```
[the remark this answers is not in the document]
michaelnebel: I introduced it as a part of inverting the dependencies between sourceModel/1 and sourceModel/9 and to not remove anything that is exposed (forgot to deprecate them).
In any case you are right - let is just delete these right now. AFAIK the field team doesn't rely on these either (they only extend the classes that are now deprecated).

csharp/ql/lib/semmle/code/csharp/dataflow/ExternalFlow.qll
```diff
@@ -229,15 +267,43 @@ predicate sinkModel(
     row.splitAt(";", 7) = kind and
     row.splitAt(";", 8) = provenance
   )
+  or
+  extSinkModel(namespace, type, subtypes, name, signature, ext, input, kind, provenance)
 }
 
+/** Holds if `row` is a sink model. */
+predicate sinkModel(string row) {
```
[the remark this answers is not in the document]
michaelnebel: Will delete.

csharp/ql/lib/semmle/code/csharp/dataflow/ExternalFlow.qll
```diff
@@ -250,20 +316,63 @@ predicate summaryModel(
...
     row.splitAt(";", 2) = name and
     row.splitAt(";", 3) = signature and
     row.splitAt(";", 4) = provenance
   )
+  or
+  extNegativeSummaryModel(namespace, type, name, signature, provenance)
+}
+
+/** Holds if `row` is a negative summary model. */
+predicate negativeSummaryModel(string row) {
```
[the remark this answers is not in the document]
michaelnebel: Will delete.

csharp/ql/lib/semmle/code/csharp/dataflow/internal/DataFlowPrivate.qll
michaelnebel: This is to ensure the bi-directional import for synthetic fields - this is implicitly handled by the bi-directional import in the *Frameworks* module in `ExternalFlow.qll`.
It turns out that the massive gordian knot that is the *Frameworks* module in ExternalFlow can't *just* be deleted (this needs to be a follow up, if we want to do that). I have document the this on: https://github.com/github/codeql-csharp-team/issues/298

michaelnebel: Added an import for a similar reason in `TaintTrackingPrivate.qll`

csharp/ql/lib/ext/Dapper.model.yml
```diff
@@ -0,0 +1,65 @@
...
+      - ["Dapper", "SqlMapper", False, "QuerySingleOrDefault<>", "(System.Data.IDbConnection,System.String,System.Object,System.Data.IDbTransaction,System.Nullable<System.Int32>,System.Nullable<System.Data.CommandType>)", "", "Argument[1]", "sql", "manual"]
+      - ["Dapper", "SqlMapper", False, "QuerySingleOrDefaultAsync", "(System.Data.IDbConnection,System.String,System.Object,System.Data.IDbTransaction,System.Nullable<System.Int32>,System.Nullable<System.Data.CommandType>)", "", "Argument[1]", "sql", "manual"]
+      - ["Dapper", "SqlMapper", False, "QuerySingleOrDefaultAsync", "(System.Data.IDbConnection,System.Type,System.String,System.Object,System.Data.IDbTransaction,System.Nullable<System.Int32>,System.Nullable<System.Data.CommandType>)", "", "Argument[2]", "sql", "manual"]
+      - ["Dapper", "SqlMapper", False, "QuerySingleOrDefaultAsync<>", "(System.Data.IDbConnection,System.String,System.Object,System.Data.IDbTransaction,System.Nullable<System.Int32>,System.Nullable<System.Data.CommandType>)", "", "Argument[1]", "sql", "manual"]
+
+    
+
```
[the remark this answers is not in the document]
michaelnebel: I will remove all the blank lines and re-generate the files.

csharp/ql/src/utils/modelconverter/project/Program.cs
michaelnebel: That is a good idea!

hvitved (approved):

PR #10779, opened 2022-10-11
C++: Add `UninitializedNode` to experimental IR dataflow

MathiasVP:
This is the final piece of syntactic compatibility needed to do the AST to IR dataflow rewrite.

Note that `uninitializedNode` is already present in the non-experimental IR dataflow library, and is implemented as `none()` with a deprecated predicate. However, we have [1 query](https://github.com/github/codeql/blob/af79139c3008e533e6583c91585bccffe02039ae/cpp/ql/src/Security/CWE/CWE-311/CleartextTransmission.ql#L46) that currently uses uninitialized nodes, so to ease the transition I've added a proper implementation of this concept in the experimental IR dataflow library.

### DIFF
--- a/cpp/ql/lib/experimental/semmle/code/cpp/ir/dataflow/internal/DataFlowUtil.qll
+++ b/cpp/ql/lib/experimental/semmle/code/cpp/ir/dataflow/internal/DataFlowUtil.qll
@@ -232,6 +232,12 @@ class Node extends TIRDataFlowNode {
   Parameter asParameter() { result = asParameter(0) }
 
   /**
+   * Gets the uninitialized local variable corresponding to this node, if
+   * any.
+   */
+  LocalVariable asUninitialized() { result = this.(UninitializedNode).getLocalVariable() }
+
+  /**
    * Gets the positional parameter corresponding to the node that represents
    * the value of the parameter after `index` number of loads, if any. For
    * example, in:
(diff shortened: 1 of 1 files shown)


jketema:
Since we only used this in one query, I wonder if we could rewrite the query? I'm probably missing some details, but I wonder if
```codeql
this.asUninitialized() instanceof SourceVariable
```
could be replaced by something like:
```codeql
exists(SourceVariable sv | sv = this.asVariable() | not sv.hasInitializer())
```
Or are there other good reasons to have `asUninitialized`?

If there are good reasons or the replacement is more difficult, then this PR LGTM.

MathiasVP:
> Since we only used this in one query, I wonder if we could rewrite the query? I'm probably missing some details, but I wonder if
> 
> ```ql
> this.asUninitialized() instanceof SourceVariable
> ```
> 
> could be replaced by something like:
> 
> ```ql
> exists(SourceVariable sv | sv = this.asVariable() | not sv.hasInitializer())
> ```
> 
> Or are there other good reasons to have `asUninitialized`?
> 
> If there are good reasons or the replacement is more difficult, then this PR LGTM.

Good point. I think that idea would also work.

That query in particular is actually one of the queries that I have a PR ready for once we can swap out the dataflow library. This query is a great example of use-use flow being a lot more user-friendly: https://github.com/MathiasVP/ql/commit/59389af2c0c8b93c929a9aa843309716478b7825#diff-2131961afc9674bdfcd7ed8b59cfffbb826d395634bc90e103c9827dd5e12de1R242 (since, instead of trying to chase down the definition of a "sensitive" variable in order to start dataflow, we can just start dataflow at the _use_ of the variable).

So if it was just a question of that one query, I think we could work around it by breaking that query temporarily, do the swap, and then merge the repair-query PR.

But unfortunately, the [AST dataflow](https://github.com/github/codeql/blob/main/cpp/ql/lib/semmle/code/cpp/dataflow/internal/DataFlowUtil.qll#L91) library hasn't deprecated these `uninitialized` dataflow nodes, and there might very well be external code out there that rely on this predicate working.

So I suggest we keep the concept fo an `uninitialized` dataflow node around for the time being, and then we can start the deprecation period once we've done the transition to IR dataflow.

jketema (approved):

PR #10780, opened 2022-10-11
Swift: detect hardcoded encryption keys

karimhamdanali:
Hardcoded keys should not be used for creating encryption ciphers. Data encrypted using hardcoded keys are more vulnerable to the possibility of recovering them.

The rule currently supports all ciphers that the CryptoSwift API provides, but we can always extend it further if more cophers are added.

I'd appreciate a review of the query itself, the accompanying tests, and the associated documentation.

### DIFF
--- a/swift/ql/src/queries/Security/CWE-321/HardcodedEncryptionKey.qhelp
+++ b/swift/ql/src/queries/Security/CWE-321/HardcodedEncryptionKey.qhelp
@@ -1,0 +1,17 @@
+<!DOCTYPE qhelp PUBLIC
+  "-//Semmle//qhelp//EN"
+  "qhelp.dtd">
+<qhelp>
+  <overview>
+    <p>Hardcoded keys should not be used for creating encryption ciphers. Data encrypted using hardcoded keys are more vulnerable to the possiblity of recovering them.</p>
+  </overview>
+
+  <recommendation>
+    <p>Use a randomly generated key material to initialize the encryption cipher.</p>
+  </recommendation>
+
+  <example>
+    <p>The following example shows a few cases of instantiating a cipher with various encryption keys. In the 'BAD' cases, the key material is hardcoded, making the encrypted data vulnerable to recovery.  In the 'GOOD' cases, the key material is randomly generated and not hardcoded, which protects the encrypted data against recovery.</p>
+    <sample src="HardcodedEncryptionKey.swift" />
+  </example>
+</qhelp>
--- a/swift/ql/src/queries/Security/CWE-321/HardcodedEncryptionKey.qhelp
+++ b/swift/ql/src/queries/Security/CWE-321/HardcodedEncryptionKey.qhelp
@@ -3,7 +3,7 @@
   "qhelp.dtd">
 <qhelp>
   <overview>
-    <p>Hardcoded keys should not be used for creating encryption ciphers. Data encrypted using hardcoded keys are more vulnerable to the possiblity of recovering them.</p>
+    <p>Hardcoded keys should not be used for creating encryption ciphers. Data encrypted using hardcoded keys are more vulnerable to the possibility of recovering them.</p>
   </overview>
 
   <recommendation>
--- a/swift/ql/src/queries/Security/CWE-321/HardcodedEncryptionKey.qhelp
+++ b/swift/ql/src/queries/Security/CWE-321/HardcodedEncryptionKey.qhelp
@@ -7,7 +7,7 @@
   </overview>
 
   <recommendation>
-    <p>Use a randomly generated key material to initialize the encryption cipher.</p>
+    <p>Use randomly generated key material to initialize the encryption cipher.</p>
   </recommendation>
 
   <example>
--- a/swift/ql/src/queries/Security/CWE-321/HardcodedEncryptionKey.ql
+++ b/swift/ql/src/queries/Security/CWE-321/HardcodedEncryptionKey.ql
@@ -1,6 +1,6 @@
 /**
  * @name Hard-coded encryption key
- * @description Using hardcoded keys for encryption is not secure, because potential attacker can easiy guess them.
+ * @description Using hardcoded keys for encryption is not secure, because potential attackers can easily guess them.
  * @kind path-problem
  * @problem.severity error
  * @security-severity 8.1
@@ -29,9 +29,7 @@ class ByteArrayLiteralSource extends KeySource {
 /**
  * A string literal is a key source.
  */
-class StringLiteralSource extends KeySource {
-  StringLiteralSource() { this instanceof StringLiteralExpr }
-}
+class StringLiteralSource extends KeySource instanceof StringLiteralExpr { }
 
 /**
  * A class for all ways to set a key.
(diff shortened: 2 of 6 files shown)


swift/ql/src/queries/Security/CWE-321/HardcodedEncryptionKey.ql
```diff
@@ -0,0 +1,70 @@
...
+/**
+ * An `Expr` that is used to initialize a key.
+ */
+abstract class KeySource extends Expr { }
+
+/**
+ * The creation of a literal byte array.
+ */
+class ByteArrayLiteralSource extends KeySource {
+  ByteArrayLiteralSource() { this = any(ArrayExpr arr | arr.getType().toString() = "Array<UInt8>") }
```
MathiasVP, line 26: I _think_ `toString` does the same as `getName` here, right? In that case, I'd prefer using `getName` here (since we like to be able to mess with the output of `toString` without fear of breaking query logic).

swift/ql/src/queries/Security/CWE-321/HardcodedEncryptionKey.ql
```diff
@@ -0,0 +1,70 @@
...
+}
+
+/**
+ * Any string literal as a key source.
+ */
+class StringLiteralSource extends KeySource {
+  StringLiteralSource() { this instanceof StringLiteralExpr }
+}
```
MathiasVP, line 34: If you want to be fancy:
```suggestion
class StringLiteralSource extends KeySource instanceof StringLiteralExpr { }
```
You can read about this use of `instanceof` [here](https://codeql.github.com/docs/ql-language-reference/types/#non-extending-subtypes).

swift/ql/src/queries/Security/CWE-321/HardcodedEncryptionKey.ql
```diff
@@ -0,0 +1,70 @@
...
+
+/**
+ * A class for all ways to set a key.
+ */
+class EncryptionKeySink extends Expr {
+  EncryptionKeySink() {
+    // `key` arg in `init` is a sink
+    exists(ClassDecl c, AbstractFunctionDecl f, CallExpr call |
```
MathiasVP, line 42: Should this `ClassDecl` be a `ClassOrStructDecl`? Or do we know that they're all actually `class`es?

swift/ql/src/queries/Security/CWE-321/HardcodedEncryptionKey.ql
```diff
@@ -0,0 +1,70 @@
+/**
+ * @name Hard-coded encryption key
+ * @description Using hardcoded keys for encryption is not secure, because potential attacker can easiy guess them.
```
MathiasVP, line 3: Did you intend to write `because a potential attacker`? Or maybe `because potential attackers`?

MathiasVP (commented):
This looks great so far!

swift/ql/src/queries/Security/CWE-321/HardcodedEncryptionKey.ql
karimhamdanali: Agreed, and it does behave the same way. Confirmed by re-running the tests, so I've changed it to `getName` instead.

swift/ql/src/queries/Security/CWE-321/HardcodedEncryptionKey.ql
karimhamdanali: Neat! Changed it to the less verbose syntax that you suggested.

swift/ql/src/queries/Security/CWE-321/HardcodedEncryptionKey.ql
karimhamdanali: They are indeed all classes.

swift/ql/src/queries/Security/CWE-321/HardcodedEncryptionKey.ql
karimhamdanali: I intended the latter. Fixed it now.

swift/ql/src/queries/Security/CWE-321/HardcodedEncryptionKey.qhelp
```diff
@@ -0,0 +1,17 @@
+<!DOCTYPE qhelp PUBLIC
+  "-//Semmle//qhelp//EN"
+  "qhelp.dtd">
+<qhelp>
+  <overview>
+    <p>Hardcoded keys should not be used for creating encryption ciphers. Data encrypted using hardcoded keys are more vulnerable to the possibility of recovering them.</p>
+  </overview>
+
+  <recommendation>
+    <p>Use a randomly generated key material to initialize the encryption cipher.</p>
```
geoffw0, line 10:
```suggestion
    <p>Use a randomly generated key to initialize the encryption cipher.</p>
```
or
```
    <p>Use randomly generated key material to initialize the encryption cipher.</p>
```

swift/ql/src/queries/Security/CWE-321/HardcodedEncryptionKey.swift
```diff
@@ -0,0 +1,22 @@
...
+	_ = try AES(key: keyString, iv: ivString)
+
+
+
+	// GOOD: Using randomly generated keys for encryption
+	let key = (0..<10).map({ _ in UInt8.random(in: 0...UInt8.max) })
+	let keyString = String(cString: key)
+	let ivString = getRandomIV()
+	_ = try Blowfish(key: key, blockMode: CBC(), padding: padding)
+	_ = try Blowfish(key: keyString, iv: ivString)
```
geoffw0, line 19: I think using a different encryption algorithm in the `GOOD` cases distracts from the point, which is that you're using a randomly generated key in this version.  Perhaps use AES everywhere, or AES and Blowfish once each equally in the `GOOD` and `BAD` block?

swift/ql/test/query-tests/Security/CWE-321/test.swift
```diff
@@ -0,0 +1,169 @@
...
+}
+
+// Helper functions
+func getConstantString() -> String {
+  "this string is constant"
+}
+
+func getConstantArray() -> Array<UInt8> {
+	[UInt8](getConstantString().utf8)
+}
```
geoffw0, line 81: `getConstantArray` is currently unused in the test.

geoffw0 (commented):
Looks great, a few minor suggestions.

swift/ql/src/queries/Security/CWE-321/HardcodedEncryptionKey.swift
karimhamdanali: You're right. I've added both algorithms to show that this vulnerability is not specific to just AES/Blowfish.

swift/ql/src/queries/Security/CWE-321/HardcodedEncryptionKey.qhelp
karimhamdanali: Fixed.

swift/ql/test/query-tests/Security/CWE-321/test.swift
karimhamdanali: Right. I was planning on using this function for a corner case that deals with constant propagation. It's a true negative though, so I'd have to properly document it in the test.

geoffw0: Its fine to have test cases we don't catch or don't yet catch.  Just mark them with `// [NOT DETECTED]` or similar.

stevecat (approved):
:wave: Reviewing on behalf of the docs team. This looks great! ✨

geoffw0 (approved):
LGTM. 🎉

PR #4926, opened 2021-01-07
Java: Query to detect weak encryption: insufficient key size

luchua-bc:
Strength of cryptographic algorithms depends on a sufficient key size to be robust against brute-force attacks. A theoretically sound encryption scheme with an insufficient key size can still be subjected to brute force attacks that have a reasonable chance of succeeding using current attack methods and resources.

This type of issues is categorized as [CWE-326: Inadequate Encryption Strength](https://cwe.mitre.org/data/definitions/326.html).

This query finds uses of strong encryption algorithms (AES, DSA, RSA, and EC) with too small a key size. 

Please consider to merge the PR. Thanks.

PR #10785, opened 2022-10-11
Java: Promote insufficient key size query from experimental

jcogs33:
This PR promotes https://github.com/github/codeql/pull/4926 from experimental.


java/ql/lib/semmle/code/java/security/InsufficientKeySize.qll
```diff
@@ -0,0 +1,191 @@
...
+      this.asExpr() = ma.getKeySizeArg()
+    )
+  }
+
+  override predicate hasState(DataFlow::FlowState state) { state = getMinSymKeySize().toString() }
+}
+
+// ********************** SINKS HELPER CLASSES & PREDICATES **********************
+/** A call to a method that initializes a key generator. */
+abstract private class KeyGenInitMethodAccess extends MethodAccess {
```
atorralba, line 127: This class doesn't seem too useful. Its only value is to reuse the `getKeySizeArg` predicate, but the class itself isn't used anywhere. Besides, if in the future another 
class extending `KeyGenInitMethodAccess` was added, it may not have its keysize argument at position 0, so this would need refactoring anyway.

Not a strong opinion (your call), but I'd remove this and just add the predicate to `AsymmetricInitMethodAccess` and `SymmetricInitMethodAccess`.

java/ql/lib/semmle/code/java/security/InsufficientKeySize.qll
```diff
@@ -0,0 +1,191 @@
...
+    this instanceof JavaSecurityKeyPairGenerator or
+    this instanceof JavaSecurityAlgoParamGenerator
+  }
+
+  override Expr getAlgoSpec() { result = this.getAlgoSpec() }
+}
+
+/** An instance of a `javax.crypto.KeyGenerator`. */
+private class SymmetricKeyGenerator extends AlgoGeneratorObject {
+  SymmetricKeyGenerator() { this instanceof JavaxCryptoKeyGenerator }
```
atorralba, line 169: Can be simplified:

```suggestion
private class SymmetricKeyGenerator extends AlgoGeneratorObject instanceof JavaxCryptoKeyGenerator {
```

java/ql/lib/semmle/code/java/security/InsufficientKeySize.qll
```diff
@@ -0,0 +1,191 @@
...
+
+  override Expr getAlgoSpec() { result = this.getAlgoSpec() }
```
atorralba, line 171: Same as above, but we can be even more explicit in this case if we apply the previous suggestion:

```suggestion
  override Expr getAlgoSpec() { result = JavaxCryptoKeyGenerator.super.getAlgoSpec() }
```

java/ql/lib/semmle/code/java/security/InsufficientKeySize.qll
```diff
@@ -0,0 +1,191 @@
...
+ * An instance of a `java.security.KeyPairGenerator`
+ * or of a `java.security.AlgorithmParameterGenerator`.
+ */
+private class AsymmetricKeyGenerator extends AlgoGeneratorObject {
+  AsymmetricKeyGenerator() {
+    this instanceof JavaSecurityKeyPairGenerator or
+    this instanceof JavaSecurityAlgoParamGenerator
+  }
+
+  override Expr getAlgoSpec() { result = this.getAlgoSpec() }
```
atorralba, line 164: Maybe a nitpick, but this apparent recursion makes me slightly nervous. I'd prefer being explicit here:

```suggestion
  override Expr getAlgoSpec() {
    result =
      [
        this.(JavaSecurityAlgoParamGenerator).getAlgoSpec(),
        this.(JavaSecurityAlgoParamGenerator).getAlgoSpec()
      ]
  }
```

java/ql/lib/semmle/code/java/security/InsufficientKeySize.qll
```diff
@@ -0,0 +1,191 @@
...
+
+/** An instance of a `javax.crypto.KeyGenerator`. */
+private class SymmetricKeyGenerator extends AlgoGeneratorObject {
+  SymmetricKeyGenerator() { this instanceof JavaxCryptoKeyGenerator }
+
+  override Expr getAlgoSpec() { result = this.getAlgoSpec() }
+}
+
+/** An instance of an algorithm specification. */
+abstract private class AlgoSpec extends ClassInstanceExpr {
```
atorralba, line 175: Same as before. If this type only exists to reuse the `getKeySizeArg` predicate I think we can pass without it, even if we repeat that one line of code in the two other types.

atorralba (commented):
Great use of flow states @jcogs33! 😄 

I added some comments, but I'll probably give it another review if you decide to apply the suggested refactor (should be easier to review the details then).

java/ql/lib/semmle/code/java/security/Encryption.qll
```diff
@@ -367,8 +383,8 @@ class JavaSecuritySignature extends JavaSecurityAlgoSpec {
   override Expr getAlgoSpec() { result = this.(ConstructorCall).getArgument(0) }
 }
 
-/** A method call to the Java class `java.security.KeyPairGenerator`. */
-class JavaSecurityKeyPairGenerator extends JavaxCryptoAlgoSpec {
+/** An instance of a `java.security.KeyPairGenerator`. */
```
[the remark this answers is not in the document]
jcogs33: Updated in commit https://github.com/github/codeql/pull/10785/commits/4c8e0a7648a6cb2ffbce812a65ae92887442d98e.

java/ql/lib/semmle/code/java/security/Encryption.qll
```diff
@@ -378,3 +394,53 @@ class JavaSecurityKeyPairGenerator extends JavaxCryptoAlgoSpec {
...
+      m.getDeclaringType() instanceof AlgorithmParameterGenerator and
+      m.getName() = "getInstance"
+    )
+  }
+
+  override Expr getAlgoSpec() { result = this.(MethodAccess).getArgument(0) }
+}
+
+/** The Java interface `java.security.spec.AlgorithmParameterSpec` */
+abstract class AlgorithmParameterSpec extends RefType { }
```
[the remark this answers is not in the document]
jcogs33: Updated QLDoc in commit https://github.com/github/codeql/pull/10785/commits/2ee23f004e0ee0c7c962e799dc4f548866b4eb64.

java/ql/lib/semmle/code/java/security/InsufficientKeySize.qll
```diff
@@ -0,0 +1,191 @@
...
+}
+
+/** A source for an insufficient key size used in AES algorithms. */
+private class SymmetricSource extends InsufficientKeySizeSource {
+  SymmetricSource() { getNodeIntValue(this) < getMinSymKeySize() }
+
+  override predicate hasState(DataFlow::FlowState state) { state = getMinSymKeySize().toString() }
+}
+
+// ************************** SOURCES HELPER PREDICATES **************************
```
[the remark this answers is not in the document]
jcogs33: Removed in commit https://github.com/github/codeql/pull/10785/commits/eb69b98dffd95802ef451ff3e93fb4ce9bc50a20.

java/ql/lib/semmle/code/java/security/InsufficientKeySize.qll
```diff
@@ -0,0 +1,191 @@
...
+private int getMinAsymNonEcKeySize() { result = 2048 }
+
+/** Returns the minimum recommended key size for elliptic curve (EC) algorithms. */
+private int getMinAsymEcKeySize() { result = 256 }
+
+/** Returns the minimum recommended key size for AES algorithms. */
+private int getMinSymKeySize() { result = 128 }
+
+/** Returns the integer value of a given DataFlow::Node. */
+private int getNodeIntValue(DataFlow::Node node) {
```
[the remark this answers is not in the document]
jcogs33: @atorralba Would it be better to simply remove this predicate? It's only being used to prevent repetition of one small line of code.

java/ql/lib/semmle/code/java/security/InsufficientKeySize.qll
jcogs33: Updated in commit https://github.com/github/codeql/pull/10785/commits/8bc0a648637a91bd3c9a0e09516fb95eeba8a18c.

java/ql/lib/semmle/code/java/security/InsufficientKeySize.qll
jcogs33: Updated in commit https://github.com/github/codeql/pull/10785/commits/09829d7f7ae1070dcce6dae992de7b17997e5862.

java/ql/lib/semmle/code/java/security/InsufficientKeySize.qll
jcogs33: Updated in commit https://github.com/github/codeql/pull/10785/commits/d569f93e7846d3b21caae6f71463971dfb8236ca.

java/ql/lib/semmle/code/java/security/InsufficientKeySize.qll
jcogs33: Updated in commit https://github.com/github/codeql/pull/10785/commits/c742a09defcab8ef7bc4fee078ff11451a9777ed.

java/ql/lib/semmle/code/java/security/InsufficientKeySize.qll
```diff
@@ -0,0 +1,191 @@
...
+  predicate hasState(DataFlow::FlowState state) { state instanceof DataFlow::FlowStateEmpty }
+}
+
+/** A sink for an insufficient key size. */
+abstract class InsufficientKeySizeSink extends DataFlow::Node {
+  /** Holds if this sink has the specified `state`. */
+  predicate hasState(DataFlow::FlowState state) { state instanceof DataFlow::FlowStateEmpty }
+}
+
+// *********************************** SOURCES ***********************************
```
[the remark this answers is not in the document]
jcogs33: @atorralba Would using [modules](https://codeql.github.com/docs/ql-language-reference/modules/#kinds-of-modules) to organize the code within the same file work instead? Creating separate files feels somewhat messy to me as well.

java/ql/lib/semmle/code/java/security/InsufficientKeySize.qll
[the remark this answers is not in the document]
jcogs33: Removed in https://github.com/github/codeql/pull/10785/commits/1a1245343deb643219b8914c590e78a57d1fdf36.

java/ql/lib/semmle/code/java/security/InsufficientKeySize.qll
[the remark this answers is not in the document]
atorralba: That would work too I guess, although we wouldn't be taking advantage of the [alleged main benefit of using modules](https://codeql.github.com/docs/ql-language-reference/modules/#importing-modules). The syntax is a little more verbose when you use modules, too.

Having said that, I think there's precedent for both approaches (split into files or into modules) in the codebase, so if using modules feels better to you, please go ahead.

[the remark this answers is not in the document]
jcogs33: Updated in https://github.com/github/codeql/pull/10785/commits/1e80fa118c08df3a13469354d154be8be7bccba9.

java/ql/lib/semmle/code/java/security/InsufficientKeySize.qll
```diff
@@ -0,0 +1,197 @@
...
+/** A sink for an insufficient key size. */
+abstract class InsufficientKeySizeSink extends DataFlow::Node {
+  /** Holds if this sink has the specified `state`. */
+  predicate hasState(DataFlow::FlowState state) { state instanceof DataFlow::FlowStateEmpty }
+}
+
+private module Asymmetric {
+  private module NonEllipticCurve {
+    /** A source for an insufficient key size used in RSA, DSA, and DH algorithms. */
+    private class AsymmetricNonEcSource extends InsufficientKeySizeSource {
```
atorralba, line 21: Now that we're using modules, do you think these classes and predicates need the `[Asymmetric[Non]Ec|Symmetric]-` prefixes? I'd slightly prefer shorter versions that'd make sense based on the modules they're in (`Source`, `Sink`, `Spec`, `getMinKeySize`, etc...). They're private anyway, but if they were referenced externally, the module syntax would provide the context IMHO (e.g. `Asymmetric::NonEllipticCurve::Source`).

But again, no strong opinion, so feel free to disregard this suggestion.

jcogs33: Updated in https://github.com/github/codeql/pull/10785/commits/1bfdfc954bb00c6b6c160dbfdf35472556d49614.

java/ql/lib/semmle/code/java/security/InsufficientKeySize.qll
```diff
@@ -0,0 +1,191 @@
...
+    /** Returns the key size from an EC algorithm's curve name string */
+    bindingset[algorithm]
+    private int getKeySize(string algorithm) {
+      algorithm.matches("sec%") and // specification such as "secp256r1"
+      result = algorithm.regexpCapture("sec[p|t](\\d+)[a-zA-Z].*", 1).toInt()
+      or
+      algorithm.matches("X9.62%") and //specification such as "X9.62 prime192v2"
+      result = algorithm.regexpCapture("X9\\.62 .*[a-zA-Z](\\d+)[a-zA-Z].*", 1).toInt()
+      or
+      (algorithm.matches("prime%") or algorithm.matches("c2tnb%")) and //specification such as "prime192v2"
```
atorralba, line 101: Simplification:

```suggestion
      algorithm.matches(["prime%", "c2tnb%"]) and //specification such as "prime192v2"
```

java/ql/lib/semmle/code/java/security/InsufficientKeySize.qll
```diff
@@ -0,0 +1,191 @@
...
+  /** An instance of a `javax.crypto.KeyGenerator`. */
+  private class KeyGen extends GeneratorAlgoSpec instanceof JavaxCryptoKeyGenerator {
+    override Expr getAlgoSpec() { result = JavaxCryptoKeyGenerator.super.getAlgoSpec() }
+  }
+}
+
+/** An instance of a generator that specifies an encryption algorithm. */
+abstract private class GeneratorAlgoSpec extends CryptoAlgoSpec {
+  /** Returns an uppercase string representing the algorithm name specified by this generator object. */
+  string getAlgoName() { result = this.getAlgoSpec().(StringLiteral).getValue().toUpperCase() }
```
atorralba, line 190: Would it make sense to use `CompileTimeConstantExpr` here instead?

```suggestion
  string getAlgoName() { result = this.getAlgoSpec().(CompileTimeConstantExpr).getStringValue().toUpperCase() }
```

atorralba (commented):
Two more minor comments, but otherwise this LGTM. If DCA and MRVA are happy, let's ask for a docs review and then merge.

java/ql/lib/semmle/code/java/security/InsufficientKeySize.qll
jcogs33: Updated in https://github.com/github/codeql/pull/10785/commits/65f74741103383fdd8e16ebb2c5e4d248e0e7c6c.

java/ql/lib/semmle/code/java/security/InsufficientKeySize.qll
jcogs33: Updated in https://github.com/github/codeql/pull/10785/commits/f40eefce57fda15bdf03953a7d6fcda89070b040. I'm rerunning MRVA to see if this affects the results.

mchammer01 (approved):
@jcogs33 👋🏻 - this looks good from a Docs point of view ✨ 
Added a couple of very minor comments (feel free to ignore them if you don't agree 🙂 )

java/ql/src/change-notes/2022-10-19-insufficient-key-size.md
```diff
@@ -0,0 +1,4 @@
+---
+category: newQuery
+---
+* Added a new query, `java/insufficient-key-size`, to detect the use of cryptographic algorithms with insufficient key sizes.
```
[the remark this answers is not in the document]
atorralba: True! See https://github.com/joefarebrother/codeql/blob/7c188a6b962db77f780aba33dcf220fb85dfe096/java/ql/src/change-notes/2022-07-19-static-initialization-vector.md.for an example.

java/ql/src/Security/CWE/CWE-326/InsufficientKeySize.ql
```diff
@@ -0,0 +1,22 @@
+/**
+ * @name Use of a cryptographic algorithm with insufficient key size
+ * @description Using cryptographic algorithms with too small a key size can
+ *              allow an attacker to compromise security.
```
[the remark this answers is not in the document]
jcogs33: Thanks for the review! I think I'd like to leave the description as is to keep it consistent with the description for the [C++ version](https://github.com/github/codeql/blob/main/cpp/ql/src/Security/CWE/CWE-326/InsufficientKeySize.ql) of this query. Unless you feel strongly about the simplification?

atorralba (approved):
🎉

PR #10791, opened 2022-10-12
Ruby: treat render 'file:' argument as a file system access

asgerf:
I came across this sink and noticed it wasn't modelled yet. Treats the `file:` keyword argument of the `render` method as a file system access, and thus a sink for path injection.

I ran [evaluation](https://github.com/github/codeql-dca-main/issues/7779) on some Rails projects, and it finds a couple of new sinks and some new path-injections alerts (in what seems to be intentionally-insecure code).


alexrford (approved):
Makes sense, new results are in test code but look plausible.

aibaars (approved):

PR #10792, opened 2022-10-12
C++: Generate better names for `InstructionNode`s and `OperandNode`s

MathiasVP:
This PR improves the `toString` on `OperandNode`s and `InstructionNode`s in experimental IR dataflow. For now I've kept it simple and only modified the strings for instructions and operands that are generated from a variable in the source code. I'm sure we can make other changes, but this is definitely the easiest one to do, and it will dramatically reduce the number of changes in internal tests (and be a great usability improvement!)


jketema:
Do you have some examples of what this looks like before and after?

PR #10795, opened 2022-10-12
Java: Add support for synthetic callables with flow summaries and model Stream.collect

aschackmull:
This covers basic support `Stream.collect`. There are more patterns that could be covered (see e.g. the qltest - currently the collector is required to be directly in the argument position), but adding support for more collectors would require some additional plumbing to connect the collector arguments to the `collect` call.


java/ql/lib/semmle/code/java/frameworks/Stream.qll
```diff
@@ -1,6 +1,101 @@
...
+
+  predicate hasName(string name) { this.getMethod().hasName(name) }
+}
+
+private class CollectToContainer extends SyntheticCallable {
+  CollectToContainer() { this = "java.util.stream.collect()+Collectors.[toList,...]" }
+
+  override Call getACall() {
+    result
+        .(CollectCall)
```
hvitved, line 27: Could change return type to `CollectCall` instead.

java/ql/lib/semmle/code/java/frameworks/Stream.qll
```diff
@@ -1,6 +1,101 @@
...
+    input = SummaryComponentStack::elementOf(SummaryComponentStack::qualifier()) and
+    output = SummaryComponentStack::elementOf(SummaryComponentStack::return()) and
+    preservesValue = true
+  }
+}
+
+private class CollectToJoining extends SyntheticCallable {
+  CollectToJoining() { this = "java.util.stream.collect()+Collectors.joining" }
+
+  override Call getACall() { result.(CollectCall).getArgument(0).(Collector).hasName("joining") }
```
hvitved, line 48: Same about return type.

java/ql/lib/semmle/code/java/frameworks/Stream.qll
```diff
@@ -1,6 +1,101 @@
...
+}
+
+private class CollectToGroupingBy extends SyntheticCallable {
+  CollectToGroupingBy() {
+    this = "java.util.stream.collect()+Collectors.[groupingBy(Function),...]"
+  }
+
+  override Call getACall() {
+    exists(Method m |
+      m = result.(CollectCall).getArgument(0).(Collector).getMethod() and
```
hvitved, line 68: same

hvitved (commented):
Approach LGTM. Some minor comments.

java/ql/lib/semmle/code/java/frameworks/Stream.qll
aschackmull: I'm not sure I like that particular asymmetric QL language feature, so I think I'll keep it as-is.

aschackmull:
dca looks good.

hvitved (approved):

PR #10796, opened 2022-10-12
C++: use explicit `this`

nickrolfe:
I realise this is in experimental code, but it fixes a QL-for-QL alert that's showing up in other PRs under "Unchanged files with check annotations".


jketema (approved):

jketema:
Thanks for the fix. Experimental is pretty relative here, as we're actively working on this code.

PR #10797, opened 2022-10-12
Kotlin: fix bit-inversion operator for Byte and Short types

smowton:


java/kotlin-extractor/src/main/kotlin/KotlinFileExtractor.kt
```diff
@@ -2442,7 +2442,10 @@ open class KotlinFileExtractor(
                         }
                     }
                     tw.writeExprsKotlinType(id, type.kotlinResult.id)
-                    unaryopDisp(id)
+                    if (target.name.asString() == "inv" && (c.type.isByte() || c.type.isShort()))
```
tamasvajk, line 2445: Does `isFunction` work on extensions? Would `isFunction(target, "kotlin.experimental", "Byte", "inv")` match on this target? This way we would also check the package and not just the function name.

smowton: OK, done (though it checks for the package of Byte / Short, not the package where the ext method is declared)

tamasvajk (approved):

PR #10798, opened 2022-10-12
Rb: add case-when expressions as a sink to rb/polynomial-redos

erik-krogh:
CVE-2021-22880: Recognizes the sink (TP with https://github.com/github/codeql/pull/10782). 
CVE-2022-30122: One step closer to the sink. 

[Evaluation looks OK](https://github.com/github/codeql-dca-main/tree/data/erik-krogh/PR-10798-0-ruby/reports).  
There is a hint towards a performance regression, but I couldn't replicate locally.  
And [a repeat](https://github.com/github/codeql-dca-main/tree/data/erik-krogh/PR-10798-0-ruby__1/reports) shows a smaller regression, and in other projects 🤷

### DIFF
--- a/ruby/ql/lib/codeql/ruby/security/regexp/PolynomialReDoSCustomizations.qll
+++ b/ruby/ql/lib/codeql/ruby/security/regexp/PolynomialReDoSCustomizations.qll
@@ -106,6 +106,14 @@ module PolynomialReDoS {
             regexp.asExpr() = call.getReceiver() and
             this.asExpr() = call.getArgument(0)
           )
+          or
+          // a case-when statement
+          exists(CfgNodes::ExprNodes::CaseExprCfgNode caseWhen |
+            matchNode.asExpr() = caseWhen and
+            this.asExpr() = caseWhen.getValue() and
+            regexp.asExpr() =
+              caseWhen.getBranch(_).(CfgNodes::ExprNodes::WhenClauseCfgNode).getPattern(_)
+          )
         )
       )
     }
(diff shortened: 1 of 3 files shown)


ruby/ql/lib/codeql/ruby/security/regexp/PolynomialReDoSCustomizations.qll
```diff
@@ -106,6 +106,14 @@ module PolynomialReDoS {
             regexp.asExpr() = call.getReceiver() and
             this.asExpr() = call.getArgument(0)
           )
+          or
+          // a case-when statement
+          exists(CfgNodes::ExprNodes::CaseExprCfgNode caseWhen |
+            matchNode.asExpr() = caseWhen and
+            this.asExpr() = caseWhen.getValue() and
+            regexp.asExpr() =
+              caseWhen.getBranch(_).(CfgNodes::ExprNodes::WhenClauseCfgNode).getPattern(_)
```
aibaars, line 115: Shall we also add regular expressions in `InClause`s?

aibaars (dismissed):
Looks good to me. We may want to do the same for `in` clauses like
```
foo = "foo"
case foo
  in /f.+/ then ...
end

aibaars (approved):

erik-krogh:
[A new evaluation came in](https://github.com/github/codeql-dca-main/tree/data/erik-krogh/pr-10798-3a3a5a__nightly__code-scanning/reports), it looks the same as the previous evaluation.  

Merging.

PR #10743, opened 2022-10-09
Spelling

jsoref:
This is not a full run of check-spelling. It's _mostly_ limited to `.qhelp` files -- The action's sarif support is very new; codeql doesn't like >5000 item reports, and I hadn't written the code to automatically trim (I've drafted some thoughts but won't be able to implement them this week).

This PR corrects misspellings identified by the [check-spelling action](https://github.com/marketplace/actions/check-spelling).

The misspellings have been reported at https://github.com/jsoref/codeql/actions/runs/3214090382#summary-8787786982

The action reports that the changes in this PR would make it happy: https://github.com/jsoref/codeql/commit/30e5caf5d61dda5a8d180f829ba9b83097056234
 
Note: this PR does not include the action. If you're interested in running a spell check on every PR and push, that can be offered separately.

PR #10801, opened 2022-10-12
Spelling ruby

jsoref:
follow-up to #10743 

per @aibaars, I'm including the changes for the handful of shared items in this PR.


ruby/ql/lib/codeql/ruby/security/BadTagFilterQuery.qll
```diff
@@ -65,7 +65,7 @@ predicate isBadRegexpFilter(HtmlMatchingRegExp regexp, string msg) {
   regexp.matches("<!-- foo --!>") and
   exists(int a, int b | a != b |
     regexp.fillsCaptureGroup("<!-- foo -->", a) and
-    // <!-- foo --> might be ambigously parsed (matching both capture groups), and that is ok here.
+    // <!-- foo --> might be ambiguous parsed (matching both capture groups), and that is ok here.
```
aibaars, line 68: This change would be grammatically incorrect. 

```suggestion
    // <!-- foo --> might be ambiguously parsed (matching both capture groups), and that is ok here.
```
or perhaps
```suggestion
    // <!-- foo --> might be parsed ambiguously (matching both capture groups), and that is ok here.
```

aibaars (commented):
Thanks a lot for the changes. I think there is one grammatical mistake, but otherwise fine.

aibaars:
@jsoref to fix the errors in the `synchronized files`  job you can run:
```
python3 config/sync-files.py --latest
```
or if that doesn't do the right thing for some reason, you can run
```
python3 config/sync-files.py ./ruby/ql/lib/codeql/ruby/controlflow/internal/ControlFlowGraphImplShared.qll
python3 config/sync-files.py ./ruby/ql/lib/codeql/ruby/security/BadTagFilterQuery.qll
python3 config/sync-files.py ./ruby/ql/lib/codeql/ruby/security/regexp/ExponentialBackTracking.qll
python3 config/sync-files.py ./ruby/ql/lib/codeql/ruby/security/regexp/RegexpMatching.qll
python3 config/sync-files.py ./ruby/ql/lib/codeql/ruby/security/regexp/SuperlinearBackTracking.qll
```

aibaars:
The QL test error can be fixed by updating the expected output in `ruby/ql/test/query-tests/experimental/manually-check-http-verb/ManuallyCheckHttpVerb.expected` and replace `mappting => mapping`

ruby/ql/lib/codeql/ruby/security/BadTagFilterQuery.qll
jsoref: Sigh. I saw that in one place and I know I switched it to the `ly` form.

aibaars: This file is shared with javascript and python, so you might have seen it there.

geoffw0 (approved):
Swift :+1:

aibaars (approved):

PR #10803, opened 2022-10-13
Ruby: Model ActionDispatch::Response

hmac:
Model the various methods on `ActionDispatch::Response` which set parts of the HTTP response. I've added a new concept `Http::Server::HeaderWriteAccess` which represents a write to a header in a HTTP response. The reason for this is that the existing concept, `HttpResponse`, describes an entire HTTP response, whereas these methods just set _part_ of the response.

There are currently no queries that use this concept, but that's just because I can't think off the top of my head which queries it would apply to.

Besides headers, we also model `response.body = ...` and `response.send_file ...`, which are regular `HttpResponse` instances.

### DIFF
--- a/ruby/ql/lib/codeql/ruby/Concepts.qll
+++ b/ruby/ql/lib/codeql/ruby/Concepts.qll
@@ -474,8 +474,8 @@ module Http {
      * extend `HeaderWriteAccess::Range` instead.
      */
     class HeaderWriteAccess extends DataFlow::Node instanceof HeaderWriteAccess::Range {
-      /** Gets the name of the header that is written to. */
-      string getName() { result = super.getName() }
+      /** Gets the (lower case) name of the header that is written to. */
+      string getName() { result = super.getName().toLowerCase() }
 
       /** Gets the value that is written to the header. */
       DataFlow::Node getValue() { result = super.getValue() }
--- a/ruby/ql/lib/codeql/ruby/Concepts.qll
+++ b/ruby/ql/lib/codeql/ruby/Concepts.qll
@@ -468,7 +468,7 @@ module Http {
     }
 
     /**
-     * A data flow node that writes data to a header in a HTTP response.
+     * A data flow node that writes data to a header in an HTTP response.
      *
      * Extend this class to refine existing API models. If you want to model new APIs,
      * extend `HeaderWriteAccess::Range` instead.
@@ -484,7 +484,7 @@ module Http {
     /** Provides a class for modeling new HTTP header writes. */
     module HeaderWriteAccess {
       /**
-       * A data flow node that writes data to the a header in a HTTP response.
+       * A data flow node that writes data to the header in an HTTP response.
        *
        * Extend this class to model new APIs. If you want to refine existing API models,
        * extend `HeaderWriteAccess` instead.
--- a/ruby/ql/lib/codeql/ruby/Concepts.qll
+++ b/ruby/ql/lib/codeql/ruby/Concepts.qll
@@ -468,6 +468,37 @@ module Http {
     }
 
     /**
+     * A data flow node that writes data to a header in a HTTP response.
+     *
+     * Extend this class to refine existing API models. If you want to model new APIs,
+     * extend `HeaderWriteAccess::Range` instead.
+     */
+    class HeaderWriteAccess extends DataFlow::Node instanceof HeaderWriteAccess::Range {
+      /** Gets the name of the header that is written to. */
+      string getName() { result = super.getName() }
+
+      /** Gets the value that is written to the header. */
+      DataFlow::Node getValue() { result = super.getValue() }
+    }
+
+    /** Provides a class for modeling new HTTP header writes. */
+    module HeaderWriteAccess {
+      /**
+       * A data flow node that writes data to the a header in a HTTP response.
+       *
+       * Extend this class to model new APIs. If you want to refine existing API models,
+       * extend `HeaderWriteAccess` instead.
+       */
+      abstract class Range extends DataFlow::Node {
+        /** Gets the name of the header that is written to. */
+        abstract string getName();
+
+        /** Gets the value that is written to the header. */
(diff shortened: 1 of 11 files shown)


aibaars:
> There are currently no queries that use this concept, but that's just because I can't think off the top of my head which queries it would apply to.

What about the XSS queries? We could treat header and body writes as sinks. User controlled data in headers can be quite dangerous.

ruby/ql/lib/codeql/ruby/frameworks/ActionController.qll
```diff
@@ -161,6 +165,124 @@ private class ActionControllerParamsCall extends ActionControllerContextCall, Pa
...
+      // Request headers are prefixed with `HTTP_` to distinguish them from
+      // "headers" supplied by Rack middleware.
+      this.getMethodName() = ["get_header", "fetch_header"] and
+      this.getArgument(0).asExpr().getExpr().getConstantValue().getString().regexpMatch("^HTTP_.+")
+    }
+
+    override Http::Server::RequestInputKind getKind() { result = Http::Server::headerInputKind() }
+  }
+
+  // TODO: each_header
```
aibaars, line 238: There's a TODO here, would you like to implement this in the current PR or a follow-up?

ruby/ql/lib/codeql/ruby/security/UrlRedirectCustomizations.qll
```diff
@@ -50,7 +50,9 @@ module UrlRedirect {
   /**
    * A source of remote user input, considered as a flow source.
    */
-  class RemoteFlowSourceAsSource extends Source, RemoteFlowSource { }
+  class HttpRequestInputAccessAsSource extends Source, Http::Server::RequestInputAccess {
```
aibaars, line 53: Could you explain this change? I would expect `Http::Server::RequestInputAccess` to be included in `RemoteFlowSource` and having URL redirects with input from other remote sources that are not `Http::Server::RequestInputAccess` would still be bad.

ruby/ql/lib/codeql/ruby/security/XSS.qll
```diff
@@ -312,9 +312,11 @@ module ReflectedXss {
   deprecated predicate isAdditionalXSSTaintStep = isAdditionalXssTaintStep/2;
 
   /**
-   * A source of remote user input, considered as a flow source.
+   * A HTTP request input, considered as a flow source.
    */
-  class RemoteFlowSourceAsSource extends Source, RemoteFlowSource { }
+  class HttpRequestInputAccessAsSource extends Source, Http::Server::RequestInputAccess {
+    HttpRequestInputAccessAsSource() { this.isThirdPartyControllable() }
```
aibaars, line 318: Same question here.

ruby/ql/test/query-tests/security/cwe-601/UrlRedirect.rb
```diff
@@ -83,6 +83,11 @@ def route14
     redirect_back_or_to params[:key], allow_other_host: false
   end
 
+  # GOOD
+  def route15
+    redirect_to cookies[:foo]
```
aibaars, line 88: I'm not entirely sure cookies should be considered safe. It is indeed harder for an attacker to influence, however, https://cwe.mitre.org/data/definitions/601.html states:
> Many open redirect problems occur because the programmer assumed that certain inputs could not be modified, such as **cookies** and hidden form fields.
--
 

<br class="Apple-interchange-newline">

aibaars (commented):
Looks generally good; I have some questions and comments .

ruby/ql/test/query-tests/security/cwe-601/UrlRedirect.rb
aibaars: I don't know, perhaps some funny `javascript:` URLs, but mitre mentions it in the CWE for some reason. Best ask the security lab folk.

aibaars: Sounds good to me.

ruby/ql/lib/codeql/ruby/Concepts.qll
```diff
@@ -443,6 +443,37 @@ module Http {
       override RequestInputKind getKind() { result = parameterInputKind() }
     }
 
+    /**
+     * A data flow node that writes data to a header in a HTTP response.
```
nickrolfe, line 447:
```suggestion
     * A data flow node that writes data to a header in an HTTP response.
```

ruby/ql/lib/codeql/ruby/Concepts.qll
```diff
@@ -443,6 +443,37 @@ module Http {
...
+      string getName() { result = super.getName() }
+
+      /** Gets the value that is written to the header. */
+      DataFlow::Node getValue() { result = super.getValue() }
+    }
+
+    /** Provides a class for modeling new HTTP header writes. */
+    module HeaderWriteAccess {
+      /**
+       * A data flow node that writes data to the a header in a HTTP response.
```
nickrolfe, line 463:
```suggestion
       * A data flow node that writes data to the header in an HTTP response.
```

ruby/ql/src/change-notes/2022-10-14-actiondispatch-response.md
```diff
@@ -0,0 +1,5 @@
+---
+category: minorAnalysis
+---
+* HTTP response header and body writes via `ActionDispatch::Response` are now
+  recognised.
```
nickrolfe, line 5:
```suggestion
  recognized.
```

nickrolfe (commented):
Looks great, nice one. I've suggested some really pedantic (i.e. unimportant) spelling/grammar fixes.

aibaars (dismissed):

asgerf (changes requested):

aibaars (approved):

asgerf (approved):

PR #10806, opened 2022-10-13
Dataflow:  Add additional annotation.

aschackmull:
This change currently has no effect, but will enable a new compiler warning in the future.


ginsbach (approved):
To confirm that you did not miss any predicates, I compiled all cpp queries with my warning-enabled draft branch. Your changes make all warnings disappear.

LGTM

PR #10810, opened 2022-10-13
Ruby: Update supported languages and frameworks

calumgrant:
This was just copied from the other languages, with links updated and checked.

A minor question is whether we could specify the RoR version supported (7.0). None of the other languages do this so I decided to leave this out @hmac.


aibaars:
Looks good to me. Do we want to mention that "one-line pattern matches" are not supported?

calumgrant:
I would not mention that - it's a small limitation that does not block analysis.

aibaars (approved):
Looks good to me. Let's wait for @hmac to answer your question about Rails version.

calumgrant:
Will merge this now. The idea listing these additional libraries is good and I can do this in a follow-up if we agree to do this.

PR #10814, opened 2022-10-13
Dataflow: Add support for synthetic global fields in MaD.

aschackmull:
This allows MaD summaries that resp. reads or writes from some global state, e.g. a database, such that jump steps are inserted between the synthesized function bodies.

This allows us to add models for `storeInDb` and `readFromDb` such that we have flow in the following case even if there's no path between `foo` and `bar`:
```
foo() {
  storeInDb(tainted);
}

bar() {
  sink(readFromDb());
}
```

### DIFF
--- a/csharp/ql/lib/semmle/code/csharp/dataflow/internal/FlowSummaryImpl.qll
+++ b/csharp/ql/lib/semmle/code/csharp/dataflow/internal/FlowSummaryImpl.qll
@@ -61,6 +61,20 @@ module Public {
 
     /** Gets a summary component for a return of kind `rk`. */
     SummaryComponent return(ReturnKind rk) { result = TReturnSummaryComponent(rk) }
+
+    /** Gets a summary component for synthetic global `sg`. */
+    SummaryComponent syntheticGlobal(SyntheticGlobal sg) {
+      result = TSyntheticGlobalSummaryComponent(sg)
+    }
+
+    /**
+     * A synthetic global. This represents some form of global state, which
+     * summaries can read and write individually.
+     */
+    abstract class SyntheticGlobal extends string {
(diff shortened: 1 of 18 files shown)


csharp/ql/lib/semmle/code/csharp/dataflow/internal/FlowSummaryImpl.qll
```diff
@@ -871,18 +908,28 @@ module Private {
...
+    /** Holds if specification component `token` parses as argument `pos`. */
     predicate parseArg(AccessPathToken token, ParameterPosition pos) {
       token.getName() = "Argument" and
       pos = parseArgBody(token.getAnArgument())
     }
 
+    /** Holds if specification component `token` parses as synthetic global `sg`. */
+    predicate parseSynthGlobal(AccessPathToken token, string sg) {
+      token.getName() = "SyntheticGlobal" and
+      sg = token.getAnArgument()
```
hvitved, line 926:
```suggestion
      sg = token.getAnArgument("SyntheticGlobal")
```

hvitved (commented):
Looks great; only one small suggestion. Also, we should run DCA for all languages.

csharp/ql/lib/semmle/code/csharp/dataflow/internal/FlowSummaryImpl.qll
aschackmull: I find the `.getAnArgument(string name)` slightly more difficult to understand than simply using `.getAnArgument()` plus `.getName()`, so I'd like to keep it as-is, since there shouldn't be any join-order issues with that.

hvitved (approved):

PR #10816, opened 2022-10-13
Kotlin: extract function references using compiler-generated adapters

smowton:


java/kotlin-extractor/src/main/kotlin/KotlinFileExtractor.kt
```diff
@@ -4487,7 +4493,10 @@ open class KotlinFileExtractor(
                 val baseClass = pluginContext.referenceClass(FqName("kotlin.jvm.internal.FunctionReference"))?.owner?.typeWith()
                     ?: pluginContext.irBuiltIns.anyType
 
-                val classId = extractGeneratedClass(ids, listOf(baseClass, fnInterfaceType), locId, functionReferenceExpr, declarationParent)
+                val classId = extractGeneratedClass(ids, listOf(baseClass, fnInterfaceType), locId, functionReferenceExpr, declarationParent, { it.valueParameters.size == 1 }) {
+                    // The argument to FunctionReference's constructor is the function arity.
```
tamasvajk, line 4497: Is this also the case for big arity functions?

tamasvajk (dismissed):
Looks reasonable to me. We could change to use the `reference.symbol` if all our tests are passing with that. (Slightly odd that `reference.reflectionTarget` is nullable, while `symbol` is not.)

java/kotlin-extractor/src/main/kotlin/KotlinFileExtractor.kt
smowton: The base class is `kotlin.jvm.internal.FunctionReference` regardless of the arity

tamasvajk (approved):

smowton:
@tamasvajk rebased, please re-approve?

tamasvajk (approved):

PR #10822, opened 2022-10-13
Koltin: support collection literals

smowton:
These are reserved for use in annotation initialisers at the moment. This would also replace the array-copying use of `IrVararg` outside the context of a call with `.clone()` call, similar to the treatment of `arrayOf(*otherArray)` -- this likely the intent of the synthetic code that uses this construct, in common with e.g. Python `x = [*y]`, which would similarly amount to a clone op.


tamasvajk (commented):
Some tests are failing, possibly with type access expressions on `int[]` -> `int` vs `Integer`.

smowton:
Yep, I had accidentally knocked out the distinction between primitive and non-primitive arrays. It appears annotations cannot carry a non-primitive int array or the like, so that path always translates `Integer` -> `int` and so on.

PR #10829, opened 2022-10-14
Ruby: Call graph performance improvements

hvitved:
Various performance improvements, most notable reducing input relations for non-linear recursion in `getTarget`.

These performance improvements are a prerequisite for getting acceptable slow-downs when adding [flow through constructors](https://github.com/github/codeql/pull/10714), but I thought it would be better to pull out these changes into a separate PR.

### DIFF
--- a/ruby/ql/lib/codeql/ruby/dataflow/internal/DataFlowDispatch.qll
+++ b/ruby/ql/lib/codeql/ruby/dataflow/internal/DataFlowDispatch.qll
@@ -151,17 +151,24 @@ private class NormalCall extends DataFlowCall, TNormalCall {
   override Location getLocation() { result = c.getLocation() }
 }
 
+/** A call for which we want to compute call targets. */
+private class RelevantCall extends CfgNodes::ExprNodes::CallCfgNode {
+  pragma[nomagic]
+  RelevantCall() {
+    // Temporarily disable operation resolution (due to bad performance)
+    not this.getExpr() instanceof Operation
+  }
+}
+
 pragma[nomagic]
-private predicate methodCall(
-  CfgNodes::ExprNodes::CallCfgNode call, DataFlow::Node receiver, string method
-) {
+private predicate methodCall(RelevantCall call, DataFlow::Node receiver, string method) {
   method = call.getExpr().(MethodCall).getMethodName() and
   receiver.asExpr() = call.getReceiver()
 }
 
 pragma[nomagic]
-private predicate flowsToMethodCall(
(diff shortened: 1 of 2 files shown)


ruby/ql/lib/codeql/ruby/dataflow/internal/DataFlowDispatch.qll
```diff
@@ -318,6 +316,19 @@ private predicate extendCallModule(Module m, Module n) {
...
+/**
+ * Gets a method available in module `m`, or in one of `m`'s transitive
+ * sub classes when `exact = false`.
+ */
+pragma[nomagic]
+private Method lookupMethod(Module m, string name, boolean exact) {
+  result = lookupMethod(m, name) and
+  exact in [false, true]
+  or
+  result = lookupMethodInSubClasses(m, name) and
```
aibaars, line 328: This change looks like it would make https://github.com/github/codeql/pull/10747 mostly obsolete, right ?

ruby/ql/lib/codeql/ruby/dataflow/internal/DataFlowDispatch.qll
```diff
@@ -332,98 +343,127 @@ private module Cached {
...
-        superCall(call, superClass, method) and
-        result = lookupMethod(superClass, method)
-      )
-      or
-      result = yieldCall(call)
     )
+    or
+    exists(Module superClass, string method |
+      superCall(call, superClass, method) and
+      result = lookupMethod(superClass, method)
```
aibaars, line 463: To make https://github.com/github/codeql/pull/10747 fully obsolete:

```suggestion
      result = lookupMethod(superClass, method, _)
```

ruby/ql/lib/codeql/ruby/dataflow/internal/DataFlowDispatch.qll
hvitved: It still only considers sub classes, not all ancestors.

ruby/ql/lib/codeql/ruby/dataflow/internal/DataFlowDispatch.qll
hvitved: I would rather not change the semantics in this PR; only performance.

aibaars (commented):
This looks good to me. Let's run a performance test and check that there are no unexpected changes to the callgraph query.

aibaars (approved):
LGTM, @hvitved feel free to merge this after confirming the callgraph looks good.

PR #10832, opened 2022-10-14
RB: add model for the `Digest` and `OpenSSL::Digest` modules

erik-krogh:
CVE-2019-5420: TP/TN

[An evaluation](https://github.com/github/codeql-dca-main/tree/data/erik-krogh/PR-10832-0-ruby/reports) shows good performance.  
The `rb/weak-cryptographic-algorithm` query has many new results with this change, but the results seem reasonable given the query.

### DIFF
--- a/ruby/ql/lib/codeql/ruby/frameworks/core/Digest.qll
+++ b/ruby/ql/lib/codeql/ruby/frameworks/core/Digest.qll
@@ -1,0 +1,34 @@
+/**
+ * Provides modeling for the `Digest` module.
+ */
+
+private import codeql.ruby.ApiGraphs
+private import codeql.ruby.Concepts
+private import codeql.ruby.DataFlow
+
+/** Gets an API node for a Digest class that hashes using `algo`. */
+private API::Node digest(Cryptography::HashingAlgorithm algo) {
+  exists(string name | result = API::getTopLevelMember("Digest").getMember(name) |
+    name = ["MD5", "SHA1", "SHA2", "RMD160"] and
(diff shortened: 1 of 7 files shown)


ruby/ql/lib/codeql/ruby/frameworks/core/Digest.qll
```diff
@@ -0,0 +1,34 @@
...
+/** A call that hashes some input using a hashing algorithm from the `Digest` module. */
+private class DigestCall extends Cryptography::CryptographicOperation::Range instanceof DataFlow::CallNode {
+  Cryptography::HashingAlgorithm algo;
+
+  DigestCall() {
+    this = digest(algo).getAMethodCall(["hexdigest", "base64digest", "bubblebabble"])
+    or
+    this = digest(algo).getAMethodCall("file") // it's directly hashing the contents of a file, but that's close enough for us.
+    or
+    this = digest(algo).getMethod("new").getReturn().getAMethodCall(["digest", "update", "<<"])
```
aibaars, line 26: Isn't `getMethod("new").getReturn()` equivalent to `getAnInstance()` ?

aibaars (approved):
Looks good to me.

ruby/ql/lib/codeql/ruby/frameworks/core/Digest.qll
erik-krogh: It is 👍

aibaars (approved):
Looks good to me.

PR #10833, opened 2022-10-14
C++: Prepare `Buffer.qll` for IR-based use-use dataflow

MathiasVP:
This PR prepares the `Buffer.qll` library for the IR-based use-use dataflow-library. There are a couple of new FPs that we didn't get prior to the dataflow change, but this is because we now have better dataflow.

### DIFF
--- a/cpp/ql/lib/semmle/code/cpp/commons/Buffer.qll
+++ b/cpp/ql/lib/semmle/code/cpp/commons/Buffer.qll
@@ -25,10 +25,22 @@ predicate memberMayBeVarSize(Class c, MemberVariable v) {
 }
 
 /**
- * Get the size in bytes of the buffer pointed to by an expression (if this can be determined).
+ * Gets the expression associated with `n`. Unlike `n.asExpr()` this also gets the
+ * expression underlying an indirect dataflow node.
  */
-language[monotonicAggregates]
-int getBufferSize(Expr bufferExpr, Element why) {
+private Expr getExpr(DataFlow::Node n) {
+  result = [n.asExpr(), n.asIndirectArgument(), n.asIndirectArgument()]
+}
+
+private DataFlow::Node exprNode(Expr e) { e = getExpr(result) }
+
+/**
+ * Holds if `bufferExpr` is an allocation-like expression.
+ *
+ * This includes both actual allocations, as well as various operations that return a pointer to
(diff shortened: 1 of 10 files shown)


cpp/ql/lib/semmle/code/cpp/commons/Buffer.qll
```diff
@@ -25,10 +25,22 @@ predicate memberMayBeVarSize(Class c, MemberVariable v) {
...
 
 /**
- * Get the size in bytes of the buffer pointed to by an expression (if this can be determined).
+ * Gets the expression associated with `n`. Unlike `n.asExpr()` this also gets the
+ * expression underlying an indirect dataflow node.
  */
-language[monotonicAggregates]
-int getBufferSize(Expr bufferExpr, Element why) {
+private Expr getExpr(DataFlow::Node n) {
+  result = [n.asExpr(), n.asIndirectArgument(), n.asIndirectArgument()]
```
intrigus-lgtm, line 32:
```suggestion
  result = [n.asExpr(), n.asIndirectArgument()]
```
or is there perhaps another indirect dataflow node besides indirect arguments?

rdmarsh2 (approved):

cpp/ql/lib/semmle/code/cpp/commons/Buffer.qll
intrigus-lgtm: @rdmarsh2 ^?

rdmarsh2: Good point, I noticed that after approving and thought I dismissed my review.

MathiasVP: Thanks for noticing this @intrigus-lgtm! I've been away on vacation last week, but I'll take a look at this. I _think_ the `asIndirectArgument` should actually have been `asIndirectExpr`, but I need to page back in all the context later today 😅.

MathiasVP: Fixed in a7d6f4ff19e92c814d83d40f906752a9ed591ad8. Indeed, `asIndirectExpr` was the right predicate to use.

rdmarsh2 (approved):

PR #10834, opened 2022-10-14
ATM/refactor endpoint filters to labels 1

kaeluka:
This is the first of several refactoring PRs.

It introduces the standard endpoint LABELS module that still relies on the standard endpoint FILTERS module and updates the nosql query to rely on labels.

At the end of the refactoring, 
* ... the filters module will have been merged into the labels module.
* ... the choice which query selects/deselects which points will be made in one location and there will be consistency tests to make sure the choice is complete.
* ... I will no longer be using hard-coded constants for selecting endpoints, but rather something type-checkable for easier maintenance.

# Performance

Performance-Experiments for this PR:

| SHA | Experiment | Result | Comment |
|-----|-------|-----|-----|
| [6208071](https://github.com/github/codeql/commit/620807157540dc75d2f701b30a29fadfc94ee37f) | https://github.com/github/codeql-dca-main/issues/7969 | KR: 👍*) | this is a current `main` head |
| [5fa3242](https://github.com/github/codeql/pull/10834/commits/5fa32428aa2c75c3a181108baa10849d71b23adc) | https://github.com/github/codeql-dca-main/issues/7970 | KR: 👍*) | comparable to the current `main` |
| [99fa66c](https://github.com/github/codeql/pull/10834/commits/99fa66c213c4d43cab893412e7d222b291d1f214) | https://github.com/github/codeql-dca-main/issues/7980 | KR: 👍*) | comparable to the current `main`; this is the current `atm/refactor-endpoint-filters-to-labels-1` head |
| [b41daa5](https://github.com/github/codeql/commit/b41daa5146b8aa53d5457a6739655bf5730ca85b) | https://github.com/github/codeql-dca-main/issues/8026 | :+1: | |

*) these weren't run with the four minutes wiggle room that the KR uses.

# To do:

 - [ ] Make a unit test to test we're not changing the behaviour of `getAReasonSinkExcluded` by accident.
 - [ ] Document the new module.

### DIFF
--- a/javascript/ql/experimental/adaptivethreatmodeling/lib/experimental/adaptivethreatmodeling/StandardEndpointLabels.qll
+++ b/javascript/ql/experimental/adaptivethreatmodeling/lib/experimental/adaptivethreatmodeling/StandardEndpointLabels.qll
@@ -10,6 +10,7 @@ private import javascript
 private import semmle.javascript.filters.ClassifyFiles as ClassifyFiles
 private import semmle.javascript.heuristics.SyntacticHeuristics
 private import CoreKnowledge as CoreKnowledge
+private import StandardEndpointFilters as StandardEndpointFilters
 
 module Labels {
   private newtype TEndpointLabel =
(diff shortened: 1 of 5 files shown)


kaeluka:
Hey, @henrymercer! I've added you in case you want to take an early peek; but I'll only remove the draft label once the performance experiment looks good.

javascript/ql/experimental/adaptivethreatmodeling/lib/experimental/adaptivethreatmodeling/StandardEndpointLabels.qll
```diff
@@ -0,0 +1,125 @@
...
+  }
+
+  class LegacyEndpointLabel extends EndpointLabel, TLegacyEndpointLabel {
+    override string getLabel(DataFlow::Node n) {
+      n = StandardEndpointFilters::getALikelyExternalLibraryCall() and
+      result = "legacy/likely-external-library-call"
+      or
+      StandardEndpointFilters::flowsToArgumentOfLikelyExternalLibraryCall(n) and
+      result = "legacy/flows-to-argument-of-likely-external-library-call"
+    }
```
henrymercer, line 48: This reads as more of a hierarchy of labels to me, and calling `getLabel` on something that is already a label doesn't entirely make sense to me.

Another way of organizing this would be to split up `LegacyEndpointLabel` into `LegacyExternalLibraryCallLabel` and `LegacyFlowsToArgumentOfLikelyExternalLibraryCallLabel` where the last two are subclasses of the first.  Then we could have a `Label#matchesEndpoint(DataFlow::Node n)` predicate.  Why use strings?  Is there a performance reason?  If we're leaving it to later, why not do it now?

henrymercer (commented):
Thanks for splitting out a small first PR to review!  This looks like it preserves behaviour, but I have a question about the structure.

javascript/ql/experimental/adaptivethreatmodeling/lib/experimental/adaptivethreatmodeling/StandardEndpointLabels.qll
kaeluka: > calling getLabel on something that is already a label doesn't entirely make sense to me.

This is great feedback. I think the solution will be to rename the `EndpointLabel` to `EndpointLabeller`!

> Why use strings?

The reason I'm returning a string is that there's one instance per label-range (a class will only ever produce labels in a certain range like `legacy/reason-sink-excluded/**`, but that `**` may be a set that is open to extension. I want consuming code to be able to pick the labeller by matching on the *range* of the labeller by default as this will very likely be faster than filtering labels; but filtering on the label should be possible, because this is more fine-grained.

Example: consider a label that measures how large the function that surrounds an endpoint is. That'd have a range of, eg., `context/function/lines-of-code/**`, but specific labels returned by `getLabel` would be `context/function/lines-of-code/32`, `context/function/lines-of-code/4`, etc..

With the `matchesEndpoint` design, you'd not be able to select "all endpoints in functions that have 15 lines or less" in a natural manner (natural means: relying on labels alone, not by adding extra methods to that labeller and using them in the .ql files) because that labeller would need to have an infinite number of instances (which isn't possible).

I'll add a documentation commit that will make this design more clear 👍

kaeluka:
Odd, I had formatted this file from vs code, but when I run `codeql query format -i path/to/file.qll` it changes the formatting yet again 😕

kaeluka:
@henrymercer, can I get a review on this? I think my comment above addresses your concerns

PR #10836, opened 2022-10-14
Ruby: fix spurious singleton calls

asgerf:
Fixes a bug and enforces a new pragmatic restriction on how calls between singleton methods are handled:
- Checks that the caller is indeed a singleton method (the comment clearly established this expectation but it wasn't enforced)
- Do not allow singleton methods on `Object` to call singleton methods on subclasses via `self`. Singleton methods are inherited by subclasses, which means `self` may refer to any subclass of the enclosing class. However, for singleton methods defined on `Object` this means all classes can flow into `self`. Sometimes a singleton method is spuriously seen as being on `Object` due to things like `Struct.new` not being modelled. Hence this restriction. See comment and test case.

[Evaluation](https://github.com/github/codeql-dca-main/issues/7850) looks good:
- About 4k fewer call edges. I triaged what I could and everything seems to be fixes of spurious edges. In some cases these are calls to a singleton method in a class that also has an instance method of the same name. I've confirmed that we do indeed call the instance method as expected in those cases, and the lost edge was a spurious additional edge.
- We also get 41 new taint sinks, which seem legit. The `KernelMethodCall` class ignores calls with a known call target, and some of the spurious call edges thus prevented us from recognising these kernel calls.


aibaars (approved):
Looks good to me

asgerf:
Merged with main to resolve a conflict, started [another evaluation](https://github.com/github/codeql-dca-main/issues/7867).

aibaars (approved):

asgerf:
[New evaluation](https://github.com/github/codeql-dca-main/issues/7870) looks the same as the last one, except a few more alerts from `rb/non-constant-kernel-open` showed up, since https://github.com/github/codeql/pull/10708 was merged in the meantime.

PR #10837, opened 2022-10-14
Ruby: Add more frameworks to the list of supported frameworks

calumgrant:
I've added the suggested frameworks to this list, however these frameworks have *partial* support so it's not clear to me which if any of these should be added. Putting this PR up here for discussion, and I don't feel too strongly about whether this should be included in the GA release.

### DIFF
--- a/docs/codeql/support/reusables/frameworks.rst
+++ b/docs/codeql/support/reusables/frameworks.rst
@@ -271,5 +271,5 @@ and the CodeQL library pack ``codeql/ruby-all`` (`changelog <https://github.com/
    rest-client, HTTP client
    Ruby on Rails, Web framework
    rubyzip, Compression library
-   typohoeus, HTTP client
+   typhoeus, HTTP client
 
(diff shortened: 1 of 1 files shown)


aibaars:
Looks good to me, @nickrolfe is there anything missing from the list?

docs/codeql/support/reusables/frameworks.rst
```diff
@@ -260,4 +260,16 @@ and the CodeQL library pack ``codeql/ruby-all`` (`changelog <https://github.com/
...
+   http_client, HTTP client
+   httparty, HTTP client
+   libxml-ruby, XML processing library
+   nokogiri, XML processing library
+   open-uri, HTTP client
+   posix-spawn, Utility library
+   rest-client, HTTP client
    Ruby on Rails, Web framework
+   rubyzip, Compression library
+   typohoeus, HTTP client
```
nickrolfe, line 274:
```suggestion
   typhoeus, HTTP client
```

nickrolfe (commented):
We also model [Net::HTTP](https://ruby-doc.org/stdlib-3.1.0/libdoc/net/http/rdoc/Net/HTTP.html), the HTTP client in the Ruby standard library. Not sure if that warrants a separate entry.

aibaars (approved):

PR #10839, opened 2022-10-14
C++: Prepare `FlowSources.qll` for IR-based use-use dataflow

MathiasVP:
This PR repairs the `FlowSources.qll` library for IR-based use-use dataflow. In fact, this PR is mostly just a cleanup of the library. Prior to this PR, the library basically re-implemented the mapping from `FunctionInput`/`FunctionOutput` to `DataFlow` nodes. This PR deletes all this re-implemented logic and forwards it to the implementation in `ModelUtil`.


rdmarsh2 (approved):
I like how much simpler this gets

rdmarsh2:
Failing internal tests will be resolved when the destination branch is merged to `main`

PR #10841, opened 2022-10-14
C++: Prepare `cpp/command-line-injection` for IR-based use-use dataflow

MathiasVP:
NOTE: This PR also contains the commits from https://github.com/github/codeql/pull/10839. Only the very last commit is new.

----------------------------

This is one of the queries that gets significantly prettier after the dataflow change 🎉.


rdmarsh2 (approved):

rdmarsh2:
I like the code but the DCA run has a 1.76x slowdown - can you take a look at it when you're back?

rdmarsh2 (approved):

PR #10844, opened 2022-10-14
Ruby: Content flow through captured variables

hvitved:
We were previously supporting direct flow through captured variables:
```rb
x = taint1
[1, 2, 3].each do |i|
    sink x # `taint1` flows in
end

[1, 2, 3].each do |i|
    y = taint2
end

sink y # `taint2` flows out
```

However, we were not supporting flow through contents (e.g. fields) on captured variables:
```rb
class C
    def set_field x
        @field = x
    end
    def get_field
        return @field
    end
end

x = C.new
x.set_field(taint1)
[1, 2, 3].each do |i|
    sink(x.get_field) # `taint1` flows in
end

y = C.new
[1, 2, 3].each do |i|
    y.set_field(taint2)
end

sink(y.get_field) # `taint2` flows out
```

Luckily, the existing SSA library provides the bits that we need to also cover the cases above:
- The first call to `each` already gives rise to an implicit read of `c`, so instead of limiting flow in to be the SSA definition for the implicit read, we do like with normal SSA flow and consider all adjacent accesses (reads or the definition itself) to flow in (that is, it is no longer `x = C.new` that flows in, but instead the post-update node for `x` in `x.set_field(taint1)`).
- Again, the second call to `each` already gives rise to an implicit read of `c`, so we simply need to add flow from the post-update of `y` in `y.set_field(taint2)` to all adjacent reads/phi nodes that immediately follow the implicit read at the call to `each`.

### DIFF
--- a/ruby/ql/lib/codeql/ruby/controlflow/CfgNodes.qll
+++ b/ruby/ql/lib/codeql/ruby/controlflow/CfgNodes.qll
@@ -727,16 +727,30 @@ module ExprNodes {
 
     override VariableAccess e;
 
-    final override VariableAccess getExpr() { result = ExprCfgNode.super.getExpr() }
+    override VariableAccess getExpr() { result = ExprCfgNode.super.getExpr() }
+
+    /** Gets the variable that is being accessed. */
+    Variable getVariable() { result = this.getExpr().getVariable() }
   }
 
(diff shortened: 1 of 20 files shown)


hvitved:
Superseded by https://github.com/github/codeql/pull/11725

PR #10845, opened 2022-10-14
ATM: Remove worsening-based queries

henrymercer:
This PR removes the queries that powered the worsening-based end-to-end and classifier evaluations.  These evaluations are now deleted or deprecated, so we no longer need to maintain the queries they used.  This speeds up running the ATM CodeQL tests, as well as removing a large number of lines of code.


kaeluka (commented):
Have you tested that this is not removing anything that still is being used later in the pipeline?

henrymercer:
> Have you tested that this is not removing anything that still is being used later in the pipeline?

Yes — see the discussion on the internal issue.

kaeluka (approved):
👍

PR #10851, opened 2022-10-16
TarSlip vulnerability improvements

Sim4n6:
Hi,

- add two new sources, as such: 

```python
tarfile.TarFile()
MKtarfile.Tarfile.open()
```
- add one more new source, which is the use of a context manager for auto-closing, as such: 
```python
contextlib.closing(XXXX)
```
- add a new sink for the method `_extract_member()` 
- distinguish two cases for the method `extractall()`
- combine the use of the previous sources with the sinks.

In the folder `/examples/`, you would find the bad samples named as: `TarSlip_xxx.py` and the relatively good samples as `NoHIT_TarSlip_xxx.py`. Plus to that, a little bash script `script.sh` that is supposed to execute the good and the bad samples using the malicious tarball `archive_malign.tar`.

Thank you


python/ql/lib/semmle/python/security/dataflow/TarSlipCustomizations.qll
```diff
@@ -58,20 +92,36 @@ module TarSlip {
...
+        not exists(Node arg | arg = call.getArgByName("members")) and
+        this = call.getObject()
+      )
+    }
+  }
+
+    /**
+   * A sink capturing method calls to `extractall`.
+   *
+   * For a call to `file.extractall` with `members` argument, `file` is considered a sink.
```
yoff, line 110: This does not seem to be what the code does.

python/ql/lib/semmle/python/security/dataflow/TarSlipCustomizations.qll
```diff
@@ -58,20 +92,36 @@ module TarSlip {
...
-        call =
-          API::moduleImport("tarfile")
-              .getMember("open")
-              .getReturn()
-              .getMember("extractall")
-              .getACall() and
-        not exists(call.getArg(_)) and
-        not exists(call.getArgByName(_)) and
-        this = call.(DataFlow::MethodCallNode).getObject()
+  class ExtractAllwoMembersSink extends Sink {
```
yoff, line 97: This name should probably be written out: `ExtractAllWithoutMembersSink`

yoff (changes requested):
Good start. It would be nice if the examples were turned into query tests.

At the moment, you are making changes in the shipped query, so we will require fairly high quality before merging. Alternatively, you copy the whole query to experimental, and it could basically be merged as is.

I have given some pointers to things to improve. Further to that would be the comments referring to relevant APIs so other developers can convince themselves that the model is reasonable.

python/ql/lib/semmle/python/security/dataflow/TarSlipCustomizations.qll
Sim4n6: renamed accordingly

python/ql/lib/semmle/python/security/dataflow/TarSlipCustomizations.qll
Sim4n6: The comment is updated to: 
```
    /**
   * A sink capturing method calls to `extractall` with `members` arguments.
   *
   * For a call to `file.extractall` with `members` argument, `file` is considered a sink if not
   * a the `members` argument contains a NameConstant as None, a List or call to the method `getmembers`.
   * Otherwise, the argument of `members` is considered a sink.  
   */
```

Sim4n6:
>  Good start. It would be nice if the examples were turned into query tests.
I did not get this one "query tests", please.

> At the moment, you are making changes in the shipped query, so we will require fairly high quality before merging. Alternatively, you copy the whole query to experimental, and it could basically be merged as is.

> I have given some pointers to things to improve. Further to that would be the comments referring to relevant APIs so other developers can convince themselves that the model is reasonable.

Ok, I will close this PR, take into consideration the previously suggested changes and make a new PR with a separate query within this [folder](https://github.com/github/codeql/tree/main/python/ql/src/experimental/Security) as `/CWE-022bis/`.

yoff:
That is fine, but it would also be fine to do it in this PR..

PR #10853, opened 2022-10-17
Kotlin: extract called private methods of specialised types, and specialised instances of anonymous types

smowton:


tamasvajk (approved):
Looks plausible to me.

java/kotlin-extractor/src/main/kotlin/KotlinFileExtractor.kt
```diff
@@ -1066,6 +1071,9 @@ open class KotlinFileExtractor(
         }
     }
 
+    private fun signatureOrWarn(t: TypeResult<*>, associatedElement: IrElement) =
+        t.signature ?: "<signature unavailable>".also { logger.warnElement("Needed a signature for a type that doesn't have one", associatedElement) }
```
igfoo, line 1075: Is this going to unify things that should be distinct? (because they both map to the same `<signature unavailable>` string?

If so, would we be better off just not extracting `associatedElement` instead?

smowton: This should only happen if an anonymous type occurs in a function's parameter list, which ought to be impossible. The magic string is also what we used to do in that circumstance; I've just had to move the code to where the signature is used rather than where it is produced.

smowton: Groan, turns out Kotlin can refer to anonymous types in parameter context: for example, within an `object : SuperType { ... }` declaration I can take a function reference `this::f`, which will capture `this`, synthesising a class implementing `FunctionX` with a constructor that takes an anonymous parameter type. Such refs will currently generate a warning that we can't produce a proper signature for that method, but we always got this wrong silently, so I think it's probably better to have it produce the warning and try to figure out what if anything to do about this later rather than re-silence the warning.

igfoo (approved):

PR #10854, opened 2022-10-17
Swift: extract all `ImplicitConversionExpr`

redsun82:
In order to do so, `VisitorBase` was changed to allow writing one `translate` function for an abstract class like `ImplicitConversionExpr`.

As there is one point in code extracting all `ImplicitConversionExpr`, there is also now only one generated test for those (using the pragma `qltest.collapse_hierarchy`). The test however does not cover yet all implicit conversions (but it is still better coverage than before).


AlexDenisov (approved):

geoffw0:
This fixes the issue I was hitting on https://github.com/github/codeql/pull/10757/ , giving us many more results for the query.  Thanks!

randydwhitephotography (commented):
i don't know

PR #10855, opened 2022-10-17
Ruby: taint-steps for printf calls - and add a `AdditionalTaintStep` class

erik-krogh:
Inspired by CVE-2020-16254, and lays some foundations to write a query that flags that CVE.  

[Evaluation looks good](https://github.com/github/codeql-dca-main/tree/data/erik-krogh/formatTaint/reports).  
Neutral performance, and a little extra taint-propagation in a single project (see the meta-alerts).


ruby/ql/lib/codeql/ruby/frameworks/StringFormatters.qll
```diff
@@ -78,6 +78,28 @@ class IOPrintfCall extends PrintfStyleCall {
...
 
+/**
+ * A call to `String#%`.
+ */
+class StringPercentCall extends PrintfStyleCall {
+  StringPercentCall() { this.getMethodName() = "%" }
+
+  override DataFlow::Node getFormatString() { result = this.getReceiver() }
+
+  override DataFlow::Node getFormatArgument(int n) {
```
alexrford, line 89: I'm kind of confused about this implementation. It seems like we want to cover 2 main cases here:
"positional" args:
```rb
"%s %d %s" % ["hello", 7, "world"]
# "hello 7 world"
```
"keyword" args:
```rb
"%{foo} %{bar}" % {foo: "hello", bar: "world"}
# "hello world"
```
I'd have expected to see an `ArrayLiteral` as `this.getArgument(0).asExpr().getExpr()` in the first case and a `HashLiteral` as that value in the second case, so something like:
```ql
exists(ArrayLiteral arrLit |
  arrLit = this.getArgument(0).asExpr().getExpr() |
  result.asExpr().getExpr() = arrLit.getElement(n)
)
or
exists(HashLiteral hashLit |
  hashLit = this.getArgument(0).asExpr().getExpr() |
  n = -2 and
  result.asExpr().getExpr() = hashLit.getAnElement()
)
```
(sorry about the ugly AST/dataflow layer mapping here)

Does this make sense? As it is, the test cases don't seem to have flow from arguments to `String#%` to the result of the call.

alexrford (commented):
Very happy with the shape of this and the `AdditionalTaintStep` class. One question/comment around `String#%`.

ruby/ql/lib/codeql/ruby/frameworks/StringFormatters.qll
erik-krogh: > Does this make sense?  

Yes it does, but a bug I'm encountering while doing so does not. 

Your example code should work but doesn't.  
The `this.getArgument(0).asExpr().getExpr()` expression is not an `Ast::ArrayLiteral`, it's an `Ast::Call` (to `[]`).  
But, if drop down to the AST layer from the start, then it does work:  
```CodeQL
exists(Ast::Call call | call = this.asExpr().getExpr() |
  exists(Ast::ArrayLiteral arrLit | arrLit = call.getArgument(0) |
    result.asExpr().getExpr() = arrLit.getElement(n)
  )
  or
  exists(Ast::HashLiteral hashLit | hashLit = call.getArgument(0) |
    n = -2 and // -2 is indicates that the index does not make sense in this context
    result.asExpr().getExpr() = hashLit.getAnElement()
  )
)
```

I suspect that desugaring is at play, but I haven't looked into the details (and I don't plan to do so).   

I've commited the code above. 

/cc @hvitved

hvitved: You should be using `ArrayLiteralCfgNode` and `HashLiteralCfgNode` instead.

erik-krogh:
A [new evaluation](https://github.com/github/codeql-dca-main/tree/data/erik-krogh/pr-10855-33cca2__nightly__code-scanning/reports) still looks good, and I've merged with `main` to resolve a compilation error.

erik-krogh:
> A general question: can we not use flow summaries for these, rather than the `AdditionalTaintStep` concept?

We could, yes.   
But then we would have to effectively reimplement the existing models (which we also need for other purposes) as flow summaries. 
So I prefer it like this.  

Also, the security lab would like an `AdditionalTaintStep` class, so we should add that class in any case.

PR #10859, opened 2022-10-17
Kotlin: Exclude fields of live literals from `java/field-masks-super-field`

tamasvajk:


smowton:
Looks good once the ql alert is fixed

tamasvajk:
> Looks good once the ql alert is fixed

@smowton Do you know if it suggests to use `not masking.getInitializer() instanceof LiveLiteral`? Is one more performant than the other? What's the benefit of the latter vs the original? (I find the original more readable, but that's probably just me)

smowton:
Yes, that's the suggestion. Mildly prefer it myself, but if we want to go against it we'd have to take it up with the style guide :)

smowton (approved):

PR #10861, opened 2022-10-17
Python: Clean up import resolution

tausbn:
A fairly complicated bit of modelling, mostly due to the quirks of how imports are handled in Python.

A few notes:

- The handling of `__all__` is not actually needed (and perhaps not desirable, as it only pertains to `import *`, though it does match the current behaviour), but it might become useful at a later date, so I left it in.
- Ideally, we would represent `foo as bar` in an `import` as a `DefinitionNode` in the CFG. I opted _not_ to do this, as it would also affect points-to, and I did not want to deal with any fallout arising from that.

--- 
I'm not sure if this merits a change note -- in principle nearly all of the old behaviour is preserved. On the other hand, if the evaluation shows a big improvement in results, then this is probably worth mentioning. For now I've left the https://github.com/github/codeql/labels/no-change-note-required label on.


python/ql/lib/semmle/python/dataflow/new/internal/ImportResolution.qll
```diff
@@ -1,14 +1,72 @@
...
+ *
+ * First, let's consider the humble `import` statement:
+ * ```python
+ * import foo
+ * import bar.baz
+ * import ham.eggs as spam
+ * ```
+ *
+ * In the AST, all imports are aliased, as in the last import above. That is, `import foo` becomes
+ * `import foo as foo`, and `import bar.baz` becomes `import bar as bar`. Note that `import` is
```
yoff, line 25: `import bar.baz` must expand to more than just `import bar as bar`, perhaps also to `from bar import baz as baz`?

python/ql/lib/semmle/python/dataflow/new/internal/ImportResolution.qll
```diff
@@ -1,14 +1,72 @@
...
+ *
+ * In general, `from foo import bar` can mean two different things:
+ *
+ * 1. If `foo` is a module, and `bar` is an attribute of `foo`, then `from foo import bar` imports
+ *    the attribute `bar` into the current module (binding it to the name `bar`).
+ * 2. If `foo` is a package, and `bar` is a submodule of `foo`, then `from foo import bar` first imports
+ *    `foo.bar`, and then attempts to locate the `bar` attribute again. In most cases, that attribute
+ *    will then point to the `bar` submodule.
+ *
+ * Now, when in comes to how these imports are represented in the AST, things get a bit complicated.
```
yoff, line 41:
```suggestion
 * Now, when it comes to how these imports are represented in the AST, things get a bit complicated.
```

python/ql/lib/semmle/python/dataflow/new/internal/ImportResolution.qll
tausbn: No. The guarantee that `import bar.baz` gives is that subsequent uses of `bar.baz` seen during execution will resolve to that module. (Technically, it also has an effect on the order of initialisation, but this is irrelevant to our modelling of module resolution.)

Yes, this is confusing. 🙂

RasmusWL:
> * Ideally, we would represent `foo as bar` in an `import` as a `DefinitionNode` in the CFG. I opted _not_ to do this, as it would also affect points-to, and I did not want to deal with any fallout arising from that.

understandable :+1: (do we have an internal issue for remembering we should do this?)

> I'm not sure if this merits a change note -- in principle nearly all of the old behaviour is preserved. On the other hand, if the evaluation shows a big improvement in results, then this is probably worth mentioning. For now I've left the [ no-change-note-required](https://github.com/github/codeql/labels/no-change-note-required) This PR does not need a change notelabel on.

:+1:

python/ql/lib/semmle/python/dataflow/new/internal/ImportResolution.qll
```diff
@@ -1,14 +1,72 @@
...
+ * aliasing desugaring as above applied to it. Thus, `from foo import bar` becomes
+ * `from foo import bar as bar`.
+ *
+ * In general, `from foo import bar` can mean two different things:
+ *
+ * 1. If `foo` is a module, and `bar` is an attribute of `foo`, then `from foo import bar` imports
+ *    the attribute `bar` into the current module (binding it to the name `bar`).
+ * 2. If `foo` is a package, and `bar` is a submodule of `foo`, then `from foo import bar` first imports
+ *    `foo.bar`, and then attempts to locate the `bar` attribute again. In most cases, that attribute
+ *    will then point to the `bar` submodule.
```
RasmusWL, line 39: I'm confused about the "attempt" part, could we use the following instead?

```suggestion
 * 2. If `foo` is a package, and `bar` is a submodule of `foo`, then `from foo import bar` first imports
 *    `foo.bar`, and then reads the `bar` attribute on `foo`. In most cases, that attribute
 *    will then point to the `bar` submodule.
```

_I guess you could actually use `del bar` in `foo/__init__.py` to remove the attribute :thinking:_

AHA, I remembered about some import tests I added in https://github.com/github/codeql/commit/0265f263018f58305966d37e690a991d04960939 (but required me looking over all files I had ever changed in this repo, since I couldn't remember the new location, and couldn't find it from just searching :shrug:)

test2 is the interesting one here (I did not annotate the original testfile, but checked out the old commit to run it, and do so here).

```py
from mypkg import foo, bar
print(foo) # 42
print(bar) #<module 'mypkg.bar' ...
```

these results means that it doesn't look like the semantics are to do an `import mypkg.foo` first, and then look up the attribute, but to see if an attribute is already available on the package, if it is, use that, if it isn't, see if there is a submodule with the right name, import that as an attribute of the parent package, and then do the attribute read. After writing that, I went and looked up [the Python documentation on `from`](https://docs.python.org/3.10/reference/simple_stmts.html#index-34), and that states the same.

Could we maybe expand the tests you have created to also have such a setup? and also update our QLDocs with the right semantics?

(personally, my guess is it will not matter that we have this slightly wrong behavior, but I think it's good to have a test to show we know about it)


Here is my take on what the docs could say:

```suggestion
 * 2. If `foo` is a package, and `bar` is already defined in `foo/__init__.py`,
 *    that value will be imported. If it is not defined, and `bar` is a submodule of `foo`, then `bar` is imported to `foo`, and the `bar` submodule imported. 
 *    Note: We don't currently model if the attribute is already defined in `__init__.py`
 *    and always assume that the submodule will be used.
```

### Where are the old import tests now?

I note that since these were ported to ApiGraph in https://github.com/github/codeql/commit/e4c3544a3f2706625f4c961a38c5d9c8700486e1
the original copy was removed in https://github.com/github/codeql/commit/79c3ccd56e0ac80cbc14bb887dc3b78aafe58d01, but sadly the current copy in the repo has been made non-executable by https://github.com/github/codeql/commit/3801a158a84a89583a1f574b1f3fe4fffaaaec5a

I honestly wasn't aware this had happened, and it would be nice to get these import tests to work again, but I'm a little in over my head in regards to other tasks, so I'm not going to work on this now.

python/ql/lib/semmle/python/dataflow/new/internal/ImportResolution.qll
```diff
@@ -18,12 +76,213 @@ module ImportResolution {
...
+      )
+    )
+  }
+
+  Module getModuleImportedByImportStar(ImportStar i) {
+    isPreferredModuleForName(result.getFile(), i.getImportedModuleName())
   }
+
+  /** Gets a data-flow node that may be a reference to a module with the name `module_name`. */
+  DataFlow::Node getReferenceToModuleName(string module_name) {
```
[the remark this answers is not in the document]
tausbn: Hm.. I don't think that's quite in line with how module naming works right now. Basically, the name of a module is the name of the corresponding file (minus extension) with the name of the package it's in prepended. And in turn, the package bit may have the name of _its_ parent package prepended, and so on. This bottoms out when we encounter something that cannot be a package.

This means that, say, if you have the following setup:
```
not-a-package/
    bar/
        __init__.py
        foo.py
also-not-a-package/
    bar/
        __init__.py
        foo.py
```
then both `foo.py` files would get the same module name `bar.foo`.

---
I think the best way to think of this predicate is as a helper predicate for `getImmediateModuleReference`. Once we have the "name" of a reference, we can go about trying to locate the module that that name may refer to. That can be either a module with that exact name, or one with `.__init__` appended (if the module is in fact a package) or whatever was added to `sys.modules` elsewhere in the code. (In this way, for instance, `os.path` refers to either the `posixpath.py` file or the `ntpath.py` file, both of which are stored _next to_ `os.py`.)

I think I'll make this predicate `private`. I don't see any immediate use for it, other than as a helper predicate.

python/ql/lib/semmle/python/dataflow/new/internal/ImportResolution.qll
```diff
@@ -18,12 +76,216 @@ module ImportResolution {
...
+
+  /** Join-order helper for `getModuleReference`. */
+  pragma[nomagic]
+  private predicate module_reference_in_scope(DataFlow::Node node, Scope s, string name) {
+    node.getScope() = s and
+    exists(Name n | n = node.asExpr() |
+      n.getId() = name and
+      pragma[only_bind_into](n).isUse()
+    )
+  }
```
yoff, line 259: I get the feeling these two should swap names.. 😅

yoff (changes requested):
Great effort modelling all this, and I love the testing part!

I was a bit confused by the names of `module_export`, which is stated to be an over-approximation, and `potential_module_export`, which sounds like an over-approximation. Why are both needed, and what are their respective roles?

It also took me a bit to discover the following structure:

- `getModuleReference`, the new top-level predicate. Closing `getImmediateModuleReference` under data flow and sub-scopes.
- `getImmediateModuleReference`, extends `getReferenceToModuleName` with additions to `sys.modules`, attribute reads, and implicit definitions. Recursive with `getModuleReference` through the disjuncts dealing with `sys.modules` and attribute reads. (And with itself via the logic for reexports.)
- `getReferenceToModuleName`, deals with the various ways of naming modules in import statements.

Could this perhaps be guided a bit in comments?

python/ql/lib/semmle/python/dataflow/new/internal/ImportResolution.qll
tausbn: I think you're right. Not entirely sure what happened there.

python/ql/lib/semmle/python/dataflow/new/internal/ImportResolution.qll
```diff
@@ -18,12 +76,213 @@ module ImportResolution {
...
+    // For parity with the points-to based solution, the `ImportExpr` and `ImportMember` bits of the
+    // above cases should _also_ point to the right modules.
+    result.asExpr() = any(ImportExpr i | i.getImportedModuleName() = module_name)
+    or
+    result.asExpr() =
+      any(ImportMember i |
+        i.getModule().(ImportExpr).getImportedModuleName() = module_name
+        or
+        i.getModule().(ImportExpr).getImportedModuleName() + "." + i.getName() = module_name and
+        none()
```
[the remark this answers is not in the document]
tausbn: Ah, so this is not something we test directly, since it's part of the "parity with point-to" aspect of the implementation. I just checked, and this does indeed add some results (that points-to finds) that would otherwise be missing.

python/ql/lib/semmle/python/dataflow/new/internal/ImportResolution.qll
```diff
@@ -18,12 +76,213 @@ module ImportResolution {
...
+    // from ..foo.bar import baz    # ..foo.bar might point to, say, package.subpackage.foo.bar
+    exists(ImportMember i | result.asExpr() = i.getModule() |
+      module_name = i.getModule().(ImportExpr).getImportedModuleName()
+    )
+    or
+    // Modules (not attributes) imported via `from ... import ... statements`, e.g.
+    // from foo.bar import baz    # imports foo.bar.baz as baz
+    // from foo.bar import baz as baz_alias    # imports foo.bar.baz as baz_alias
+    exists(Import i, Alias a, ImportMember im | a = i.getAName() and im = a.getValue() |
+      i.isFromImport() and
```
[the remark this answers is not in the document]
tausbn: Indeed it is!

tausbn:
I added a test for the case where an attribute name overlaps with the name of a submodule. However, you may notice the `MISSING` annotation that popped up on one of the earlier tests. There appears to be something missing from my modelling which did not become apparent until the fix to `getARelevantTag` was added. (This feels like a major footgun to me, incidentally.)

python/ql/lib/semmle/python/dataflow/new/internal/ImportResolution.qll
```diff
@@ -18,12 +76,213 @@ module ImportResolution {
       or
       defn.getNode() = v.getDefinition().(ArgumentRefinement).getArgument()
     )
+    or
+    exists(Alias a |
+      defn.asExpr() = [a.getValue(), a.getValue().(ImportMember).getModule()] and
```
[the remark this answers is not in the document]
yoff: So in the end, we do want more than `ImportExpr`s here?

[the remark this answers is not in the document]
tausbn: We want both the `ImportMember`s and the `ImportExpr`s that can appear on the left hand side of an `Alias`, yes.

RasmusWL (approved):
Really easy to follow the ql changes when I could see the changes in the tests as well :muscle: :heart:

I can see that performance looks good now, and I looked at the first two result changes which also looks good (couldn't hold myself back, since I was excited about this work) :tada: 

So from my POV ready to merge once we're confident about the rest of the result changes :+1:

-- actually, the fact that this gives more results makes me wonder if we should have a change note as well?

python/ql/test/experimental/import-resolution/importflow.ql
```diff
@@ -0,0 +1,122 @@
...
+    not mod.getName() in ["__future__", "trace"]
+  }
+
+  string getName() { result = mod.getName() }
+}
+
+/**
+ * A data-flow node that is guarded by a version check. Only supports checks of the form `if
+ *sys.version_info[0] == ...` where the right hand side is either `2` or `3`.
+ */
```
yoff, line 39: This is very specific. But its use seems even more specific, namely only those where `version` is `3`. Are you keeping the option open for a `ResolutionTest2`? Otherwise this node should perhaps be `Python3Node`...unless the point is to exclude code guarded by version 2 from all tests?

yoff (approved):
Only one issue, which I do not think a blocker, so I have marked as approved.

yoff:
> -- actually, the fact that this gives more results makes me wonder if we should have a change note as well? (I'm not feeling strongly about it)

We probably should..

python/ql/test/experimental/import-resolution/importflow.ql
tausbn: Yeah, I figured there was no harm in leaving in the possibility of extending it to also handle Python 2-only tests. However, as I was not aware of any Python 2 behaviour that needed testing, I decided not to implement it now. (In part also because I had only just gotten this to work reliably, after trying a whole bunch of variations that failed for one reason or another.)

If we end up needing Python 2 stuff, I think the better approach would be to simply rename `ResolutionTest3` to `VersionSpecificResolutionTest` and use `major_version()` to decide which tag is active and which tag to look for in the actual results. That way, we don't have to (basically) duplicate the body of `hasActualResult`.

tausbn: Yeah, I figured there was no harm in leaving in the possibility of extending it to also handle Python 2-only tests. However, as I was not aware of any Python 2 behaviour that needed testing, I decided not to implement it now. (In part also because I had only just gotten this to work reliably, after trying a whole bunch of variations that failed for one reason or another.)

If we end up needing Python 2 stuff, I think the better approach would be to simply rename `ResolutionTest3` to `VersionSpecificResolutionTest` and use `major_version()` to decide which tag is active and which tag to look for in the actual results. That way, we don't have to (basically) duplicate the body of `hasActualResult`.

tausbn:
With two approvals in the bag, I'll merge this now, and create a follow-up PR with change note (and a bit of version-handling cleanup in the tests). Thank you both for the excellent reviews! ❤️

PR #10862, opened 2022-10-17
Rb: Add an `unsafe-code-construction` query

erik-krogh:
TODO:  
 - [ ] depends on https://github.com/github/codeql/pull/10855 
 - [ ] depends on https://github.com/github/codeql/pull/10680


erik-krogh:
> how would you rewrite the following to not use module_eval? 

The example is incomplete so thats hard to say, but I don't see anything that couldn't be rewritten to not use `module_eval` with a constructed string.   

To be clear, using `module_eval` is not the issue, using `module_eval` with a constructed string is the issue.   
The documentation should be clearer about that, I'll look into it.  

Below is an example I made of how to rewrite from the bad pattern to the good pattern.   
I'm going to include that in the `.qhelp`, I'll push that when I've done it.   

```Ruby
# Makes a class with a method named `getter_name` that returns `val` in a nasty way
module BadMakeGetter
  def self.define_getter_class(getter_name, val)
    new_class = Class.new
    new_class.module_eval <<-END
      def #{getter_name}
        #{val}
      end
    END
    new_class
  end
end

# Same as above, but uses `define_method` instead of constructing a string
module GoodMakeGetter
  def self.define_getter_class(getter_name, val)
    new_class = Class.new
    new_class.module_eval do
      define_method(getter_name) { val }
    end
    new_class
  end
end

one = BadMakeGetter.define_getter_class(:one, 1)
puts "One is #{one.new.one}"

two = GoodMakeGetter.define_getter_class(:two, 2)
puts "Two is #{two.new.two}"
```

erik-krogh:
@hmac are you up for writing some examples I can put in the `.qhelp`?  
You seem to be more into the finer details of e.g. `class_variable_get` than I am at this point.  

I think the only methods missing from your list are `{class,instance}_variable_{get,set}` and `send`.  

A note about `send`:  
The code-injection query only flags unsafe uses of `send` if an attacker can control the entire string, so if an attacker can only control e.g. a prefix/suffix, then we don't produce an alert.  
That was the change I made using flow-labels recently.   
And this query does not treat `send()` as a sink. This query only uses code-execution sinks where the `runsArbitraryCode()` predicate hold, which is not the case for e.g. `send`.

ruby/ql/lib/codeql/ruby/security/UnsafeCodeConstructionCustomizations.qll
```diff
@@ -0,0 +1,119 @@
...
+    override DataFlow::Node getCodeSink() { result = s }
+
+    override string getSinkType() { result = "array" }
+  }
+
+  /**
+   * A string constructed from a string-literal (e.g. `"foo #{sink}"`),
+   * where the resulting string ends up being executed as a code.
+   */
+  class StringFormatAsSink extends Sink {
```
alexrford, line 84: This is very nitpicky, but `StringInterpolationAsSink` seems like a more accurate name for this class.
```suggestion
  class StringInterpolationAsSink extends Sink {
```

ruby/ql/lib/codeql/ruby/security/UnsafeCodeConstructionCustomizations.qll
```diff
@@ -0,0 +1,119 @@
...
+    Concepts::CodeExecution s;
+
+    StringFormatAsSink() {
```
alexrford, line 87: As above
```suggestion
    StringInterpolationAsSink() {
```

ruby/ql/lib/codeql/ruby/security/UnsafeCodeConstructionQuery.qll
```diff
@@ -0,0 +1,49 @@
...
+  override predicate isSink(DataFlow::Node sink) { sink instanceof Sink }
+
+  override predicate isSanitizer(DataFlow::Node node) {
+    node instanceof StringConstCompareBarrier or
+    node instanceof StringConstArrayInclusionCallBarrier
+  }
+
+  // override to require the path doesn't have unmatched return steps
+  override DataFlow::FlowFeature getAFeature() {
+    result instanceof DataFlow::FeatureHasSourceCallContext
```
alexrford, line 31: Just a question - I'm not familiar with `FlowFeature`s - why is predicate necessary for this particular flow configuration?

alexrford (changes requested):
Happy with the code, and the qhelp is extremely informative. Just some minor nitpicks/questions.

ruby/ql/lib/codeql/ruby/security/UnsafeCodeConstructionQuery.qll
erik-krogh: It prevents flow out of functions that weren't called as part of the path.   
I have a test-case for it here: https://github.com/github/codeql/blob/44213f0144fdd54bb679ca48d68b28dcf820f7a8/ruby/ql/test/query-tests/security/cwe-078/UnsafeShellCommandConstruction/impl/unsafeShell.rb#L37-L44 

In that example the `x` parameter in `id()` is correctly identified as a source.  
However, if that value flows out of the `id()` function (into e.g. `thisIsSafe()`), then it is no longer a valid source.  

This flow-feature blocks that kind of flow for returns that doesn't have a matching call.

alexrford (approved):
👍

mchammer01:
I'll review this on behalf of Docs 😃

ruby/ql/src/queries/security/cwe-094/UnsafeCodeConstruction.qhelp
```diff
@@ -0,0 +1,111 @@
+<!DOCTYPE qhelp PUBLIC
+  "-//Semmle//qhelp//EN"
+  "qhelp.dtd">
+<qhelp>
+
+<overview>
+<p>
+When a library function dynamically constructs code in a potentially unsafe way, then
```
mchammer01, line 8: I don't think we need this word here.
```suggestion
When a library function dynamically constructs code in a potentially unsafe way, 
```

ruby/ql/src/queries/security/cwe-094/UnsafeCodeConstruction.qhelp
```diff
@@ -0,0 +1,111 @@
...
+<p>
+Properly document library functions that construct code from unsanitized 
+inputs, or avoid constructing code in the first place.
+</p>
+</recommendation>
+
+<example>
+<p>
+The following example shows two methods implemented using <code>eval</code>: a simple
+deserialization routine and a getter method.
```
mchammer01, line 29: Do we need some formatting here, like this? 🤔 
```suggestion
deserialization routine and a `getter` method.
```

ruby/ql/src/queries/security/cwe-094/UnsafeCodeConstruction.qhelp
```diff
@@ -0,0 +1,111 @@
...
+If untrusted inputs are used with these methods,
+then an attacker might be able to execute arbitrary code on the system. 
+</p>
+
+<sample src="examples/UnsafeCodeConstruction.rb" />
+
+<p>
+To avoid this problem, either properly document that the function is potentially
+unsafe, or use an alternative solution such as <code>JSON.parse</code> or another library, like in the examples below,
```
mchammer01, line 38: Simplifying as the removed clause was cutting the flow of what you were trying to say.
```suggestion
unsafe, or use an alternative solution such as <code>JSON.parse</code> or another library 
```

mchammer01 (dismissed):
@erik-krogh - LGTM ✨ 
A few minor nits for your consideration.

ruby/ql/src/queries/security/cwe-094/UnsafeCodeConstruction.qhelp
erik-krogh: I don't think we need any formatting here. 
There is not a method named "getter". "getter" describes a kind of method in this context.

alexrford (approved):
LGTM

PR #10863, opened 2022-10-17
Ruby: add XML tables to dbscheme

aibaars:


ruby/ql/lib/upgrades/3595c826de6db850f16b9da265a54dbf24dd3126/upgrade.properties
```diff
@@ -0,0 +1,2 @@
+description: <INSERT DESCRIPTION HERE>
+compatibility: full|backwards|partial|breaking
```
nickrolfe, line 2: You forgot to change this.

ruby/ql/lib/codeql/xml/XML.qll
```diff
@@ -0,0 +1,307 @@
+/**
+ * Provides classes and predicates for working with XML files and their content.
```
nickrolfe, line 2: This seems like another candidate for a shared library pack, but that can be done separately.

nickrolfe (changes requested):

.github/workflows/ruby-dataset-measure.yml
```diff
@@ -40,11 +47,13 @@ jobs:
...
       - name: Create database
         run: |
-          codeql database create \
+          codeql database init \
             --search-path "${{ github.workspace }}/ruby/extractor-pack" \
-            --threads 4 \
             --language ruby --source-root "${{ github.workspace }}/repo" \
             "${{ runner.temp }}/database"
+          codeql database trace-command --threads 4 --index-traceless-dbs "${{ runner.temp }}/database"
+          codeql database index-files -l xml --include "**/*.xml.erb" "${{ runner.temp }}/database"
```
calumgrant, line 55: I am a bit surprised that our xml parser handles `xml.erb` files in the general case. Do you think there could be spurious syntax errors, or will it be guaranteed to work?

ruby/ql/lib/codeql/xml/XML.qll
aibaars: I actually copied the XML.qll files that the other languages share via `identical-files.json` . Unfortunately, the shared file contains an invalid (for us) `import semmle.files.FileSystem` statement. In addition the shared file has very many deprecated, renamed classes. So I cleaned it up.

Ideally we make this a proper shared library pack based on @erik-krogh's work on a shared Locations/Files library.

erik-krogh: > Ideally we make this a proper shared library pack based on @erik-krogh's work on a shared Locations/Files library. 

I think it will take a while before I got the CLI features I need to make that library. So don't wait on that.

ruby/ql/lib/upgrades/3595c826de6db850f16b9da265a54dbf24dd3126/upgrade.properties
aibaars: Ah indeed. I changed it, but I regenerated the files because I had forgotten about the `numlines` table.

aibaars: > Why did the PR checks pass?

CI doesn't check the description or compatibility fields. It only checks the dbscheme and whether the upgrade queries can be compiled (there are none in this case). The "upgrade" is simply adding empty tables which can be achieved by doing nothing at all ;-)

nickrolfe (approved):
👍

.github/workflows/ruby-dataset-measure.yml
aibaars: > That's a good point. I don't think it would work in the general case, but it might be alright for all the .xml.erb files in the five projects above.

There are plenty of XML syntax errors in the `xml.erb` files in the 5 projects we currently use. However, these errors are non-fatal. The purpose is just to gather some stats about XML-like data. For that purpose it works fine.

calumgrant (approved):
Feel free to hold this back from the GA if you think it's too risky.

Should we also update the autobuilder to extract XML files by default (and common extensions), or will this always be a manual build step?

aibaars:
> Feel free to hold this back from the GA if you think it's too risky.
> 
> Should we also update the autobuilder to extract XML files by default (and common extensions), or will this always be a manual build step?

We could, but I don't think it is very useful.  Let's wait for a use-case first.  It took me a while to find some repo with a non-trivial amount of xml-like data to gather stats, so I suspect that XML data is not very popular in Ruby code bases..

PR #10865, opened 2022-10-17
Java: Add library support for activity-alias elements in AndroidManifest.qll

egregius313:
Android supports `<activity-alias>` elements which alias `<activity>` elements. This PR adds the `AndroidActivityAliasXmlElement` class to add support for activity aliases.

### DIFF
--- a/java/ql/lib/change-notes/2022-10-17-android-activity-alias.md
+++ b/java/ql/lib/change-notes/2022-10-17-android-activity-alias.md
@@ -1,8 +1,4 @@
 ---
 category: minorAnalysis
 ---
-* Added the `AndroidActivityAliasXmlElement` class to represent `<activity-alias>` components in Android manifests.
-* Added the `AndroidActivityXmlElement.getAnAlias` method to get the aliases of an `<activity>` element in an Android manifest.
-* Added the `AndroidIdentifierXmlAttribute` class to represent XML attributes representing component identifiers, such as `android:name` and `android:targetActivity`
-* Added the `AndroidComponentXmlElement.getResolvedIdentifier` method, for finding the fully qualified identifier from an identifier.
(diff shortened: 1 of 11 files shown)


java/ql/lib/change-notes/2022-10-17-android-activity-alias.md
```diff
@@ -0,0 +1,8 @@
+---
+category: minorAnalysis
+---
+* Added the `AndroidActivityAliasXmlElement` class to represent `<activity-alias>` components in Android manifests.
```
atorralba, line 4: I would avoid being so detailed in the change note (I know it’s consistent with previous PRs, but it’s getting out of hand) and instead focus on the added functionality. In this case, I would mention something along the lines of “Data flow sources are now properly recognized in exported Android activity aliases when they target a not exported activity”. Similarly with permissions if we actually implement that as well.

java/ql/lib/semmle/code/java/frameworks/android/Android.qll
```diff
@@ -27,7 +27,12 @@ class AndroidComponent extends Class {
 
   /** The XML element corresponding to this Android component. */
   AndroidComponentXmlElement getAndroidComponentXmlElement() {
-    result.getResolvedComponentName() = this.getQualifiedName()
+    // Find an element with an identifier matching the qualified name of the component.
+    // Aliases have two identifiers (name and target), so check both identifiers (if present).
```
atorralba, line 31: Isn't the `target` attribute the one which actually points to the activity being aliased? AFAIK the alias' `name` can be arbitrarily set and doesn't affect its actual target.

atorralba (approved):
Now that I think of it, this isn't a new query, just library changes, so a docs review isn't required. Let's not add more overhead to our docs teams, and just merge.

@github/docs-content-codeql sorry for the unnecessary ping 🙏

PR #10869, opened 2022-10-18
Swift: replace `Base` suffix with `Generated::` module

redsun82:
It makes more sense and aids readability.

To make sense of the huge commits:
* 307c885c1f79f21cd4c1b317b3ad1229be57d1f6 updates `qlgen.py` and the template, and all generated classes. There was a hidden issue where a generated qll file would import the stub qll file of the same class in some cases, which is not possible any more. This was fixed by skipping self in `get_classes_used_by`.
* 581939d139adeb84830deabf25244136392be386 updates automatically all modified stubs, simply running `sed -ri 's/\b([A-Z]\w+)Base\b/Generated::\1/g'` on the `swift/ql/lib/codeql/swift/elements` directory.
* af3f782ad554f952d05824a20c537df4d1bb0226 is the only manual change to QL code, fixing an import that does not work any more.

After the first push of the PR the following fixes were pushed:
* f4f5e3e38267638b38c4fb50411a4ecde7e486db removes a redudant module namespace
* e29fe54b3c4ed1eeb5fa9a3cf1b3c805df4ec523 removes a redundant import
* 9c7eec5e440fa139764c4518da33b4e1fdda83cc removes a debug print from `qlgen.py`


geoffw0 (approved):
LGTM.  I think its a good change.  Most of my confidence its actually correct comes from the tests passing.  :)

PR #10873, opened 2022-10-18
Ruby: drop beta notice

aibaars:
This PR should be merged once our documentation is in a more complete state. Just leaving this pull request as a remiinder.


felicitymay (approved):
Thanks for getting this update to the docs ready for when we want to publish this change ✨

PR #10875, opened 2022-10-18
Swift: add infrastructure for documenting generated code

redsun82:
This introduces the infrastructure to add documentation comments to generated QL code.

Classes use the standard Python docstrings in `schema.py` verbatim.

Property getters have a standard generated header  `Gets the <property>` or ``Gets the `index`th <property> (0-based)`` (adding `if it exists` for optional properties), optionally followed by a more in-depth description that can be specified adding `| desc("""<description>""")` to the property in `schema.py`. The `<property>` part is automatically derived from the property and class names, expanding any common abbreviations (for example `params` in `Callable`  becomes `parameters of this callable`), but it can be overridden with `| doc("<property>")`. In case the default is used, the class name used in it can still be tweaked at the class level by applying the pragma `@ql.default_doc_name("<class name>")` (example provided for `AnyFunctionType`).

A special case is overriding the doc of repeated properties that require both singular and plural forms. When overriding with `doc`, by default the first provided word is considered the noun (example is given for `ApplyExpr` and `arguments`). If however the noun is not the first word, one is required to wrap the noun in curly braces (example is given for `GenericFunctionType` and `generic_params` which has `type {parameters} of this generic type`). This can also be used to mark multiple nouns, if ever the need arises (for example in case of a hypothetical `{cats} or {dogs} of this household`. This special formatting is unneeded for non repeated properties, and the curly braces will be removed in that case.

See changes to `swift/schema.py` for the included examples.

@geoffw0 I would appreciate comments on how the included docs look like (including the hand-written ones in `Element.qll`) and how the specifying of docs in `schema.py` looks like to you.


swift/ql/lib/codeql/swift/generated/Callable.qll
```diff
@@ -7,37 +7,70 @@ import codeql.swift.elements.decl.ParamDecl
 
 module Generated {
   class Callable extends Synth::TCallable, Element {
+    /**
+     * Gets the self param, if it exists.
+     * This is taken from the "hidden" AST and should only be used to be overridden by classes.
```
geoffw0: I'm not especially fond of the spacing here.  I think I'd prefer it to be all on the same line, or with a blank line between.

swift/ql/lib/codeql/swift/generated/Comment.qll
```diff
@@ -7,6 +7,9 @@ module Generated {
   class Comment extends Synth::TComment, Locatable {
     override string getAPrimaryQlClass() { result = "Comment" }
 
+    /**
+     * Gets the text.
```
geoffw0, line 11: If we can do it, I'd prefer `Gets the name of this comment.`, `Gets the name of this file.` etc?

swift/ql/lib/codeql/swift/generated/Locatable.qll
```diff
@@ -6,15 +6,25 @@ import codeql.swift.elements.Location
 
 module Generated {
   class Locatable extends Synth::TLocatable, Element {
+    /**
+     * Gets the location, if it exists.
```
geoffw0, line 10: How easy is it to tweak these individually, e.g. if I want this to say `Gets the location associated with this element.` or `Gets the location of this element in the source.` or some such?  Which file would I need to edit in future if I wanted to accomplish this?

swift/ql/lib/codeql/swift/generated/decl/ConcreteVarDecl.qll
```diff
@@ -7,6 +7,9 @@ module Generated {
   class ConcreteVarDecl extends Synth::TConcreteVarDecl, VarDecl {
     override string getAPrimaryQlClass() { result = "ConcreteVarDecl" }
 
+    /**
+     * Gets the introducer int.
```
geoffw0, line 11: This one is not very self explanatory and could do with being expanded.

swift/ql/lib/codeql/swift/generated/Callable.qll
```diff
@@ -9,7 +9,7 @@ module Generated {
   class Callable extends Synth::TCallable, Element {
     /**
      * Gets the self param, if it exists.
-     * This is taken from the "hidden" AST and should only be used to be overridden by classes.
+     * This includes nodes from the "hidden" AST and can be used to be overridden by subclasses.
```
geoffw0, line 12: I'm not entirely clear what the second part of this sentence is saying (before or after the edit).

geoffw0 (commented):
Looks good to me.  Some suggestions and questions follow - but I think we should mostly focus on the infrastructure in this PR and not on getting every comment perfect.

swift/ql/lib/codeql/swift/generated/Callable.qll
redsun82: Ok, I was hesitating on the blank line as well

swift/ql/lib/codeql/swift/generated/Comment.qll
redsun82: I think it is doable, but I have to think about implications. For example
``Gets the `index`th argument of this apply expression`` from `ApplyExpr` sounds good, but
``Gets the name of this abstract function declaration`` from `AbstractFunctionDecl` should probably drop the `abstract` part. Or also `declaration`, but then `argument of this apply` or `member of this brace` from `BraceStmt` would sound weird.

Probably just requires the `docname` modifier to be extended to classes as well, so we can fix the cases that end up not to our liking.

swift/ql/lib/codeql/swift/generated/decl/ConcreteVarDecl.qll
redsun82: definitely, it was my first candidate for an example, but actually this might be something we might want to take out of the extraction altogether. This is telling whether syntactically the variable was introduced with `var` or `let`. That affects settability, but there is other information that we should extract that tells us that, we should probably not really care whether syntaically the programmer has written `var` or `let`.

swift/ql/lib/codeql/swift/generated/Locatable.qll
redsun82: Actually, you can achieve what you propose in the current version by editing `schema.py` like this
```diff
 @qltest.skip
 class Locatable(Element):
-    location: optional[Location] | cpp.skip
+    location: optional[Location] | cpp.skip | docname("location of this element in the code")
```
however that does not work very well currently with repeated properties, where one needs to switch between singular (normal getter) and plural (documenting `getA...` and `getNumberOf...`), because the generator would not necessarily know where to put the `s`.

redsun82:
Thanks for the feedback @geoffw0 , I'm now thinking about these tweaks and enhancements:
* the default doc name for properties should include `of this <class name>`
* the doc name should optionally allow in case of repeated fields to separate what should be pluralized and what not, and should override also the `of this <class name>` part. Sounds complicated, but I will present an example that should make it clear.
* probably the `| docname(...)` modifier looks better just as `| doc(...)`

I will show these tomorrow, and address other comments.

swift/ql/lib/codeql/swift/generated/decl/ConcreteVarDecl.qll
geoffw0: I can imagine stylistic queries that care whether the programmer wrote `var` or `let`.

swift/ql/lib/codeql/swift/generated/Callable.qll
geoffw0: I'm not sure if this is right but I've come to:
```
This includes nodes from the "hidden" AST. It can be overridden in subclasses to change the behaviour of both the Immediate and non-Immediate versions.
```

swift/ql/lib/codeql/swift/generated/Locatable.qll
redsun82: please look at the updated `location` doc now :slightly_smiling_face:

swift/ql/lib/codeql/swift/generated/Callable.qll
redsun82: nice suggestion, committed as 7b181a2de0f9afe1d1d8e865a9c91ba18fe1425d

swift/ql/lib/codeql/swift/generated/decl/ConcreteVarDecl.qll
redsun82: added by b65f49bd50eed1828f81eab3d9bb2432c7bc4d83

geoffw0:
Much improved. :+1:

I'll finish reviewing this tomorrow.

swift/ql/lib/codeql/swift/generated/decl/ConcreteVarDecl.qll
```diff
@@ -8,7 +8,8 @@ module Generated {
     override string getAPrimaryQlClass() { result = "ConcreteVarDecl" }
 
     /**
-     * Gets the introducer int of this concrete var declaration.
+     * Gets the introducer enumeration value.
+     * This is 0 if the variable was introduced with `let` and 1 if it was introduced with `var`.
```
geoffw0, line 13: :+1:

swift/ql/lib/codeql/swift/generated/decl/ModuleDecl.qll
```diff
@@ -7,10 +7,16 @@ module Generated {
   class ModuleDecl extends Synth::TModuleDecl, TypeDecl {
     override string getAPrimaryQlClass() { result = "ModuleDecl" }
 
+    /**
+     * Holds if this module is the built-in one.
```
geoffw0, line 11: Is there only one built-in module?

geoffw0 (approved):
Looks great, thanks for all the attention to detail here!

swift/ql/lib/codeql/swift/generated/decl/ModuleDecl.qll
redsun82: yes :slightly_smiling_face: 

By the way, this is a module you cannot import normally, only if you pass `-parse-std-lib` to the compiler. As the flag suggests, the module should not be used outside the Swift standard library.

PR #10876, opened 2022-10-18
Java models-as-data: infer Kotlin $default models from that of its parent function

smowton:
For example, if we have `A.f arg1 -> retval`, infer `A.f$default arg2 -> retval`, the arg position being adjusted because the dispatch receiver is included at the beginning of the regular parameter list. I'm using regex hacking rather than our nice `AccessPathToken` API to allow me to accomplish everything in impl-specific land without introducing mutual recursion between the the language-specific and -independent parts of the flow summary implementation.


java/ql/integration-tests/posix-only/kotlin/default-parameter-mad-flow/test.expected
```diff
@@ -0,0 +1,66 @@
...
+| user.kt:38:29:38:30 | st [post update] : SomeToken | semmle.label | st [post update] : SomeToken |
+| user.kt:39:10:39:11 | st | semmle.label | st |
+| user.kt:42:13:42:20 | source(...) | semmle.label | source(...) |
+| user.kt:43:16:43:23 | source(...) | semmle.label | source(...) |
+| user.kt:44:31:44:38 | source(...) | semmle.label | source(...) |
+| user.kt:45:20:45:27 | source(...) | semmle.label | source(...) |
+| user.kt:47:39:47:46 | source(...) | semmle.label | source(...) |
+subpaths
+#select
+| user.kt:7:32:7:39 | source(...) : Number | user.kt:7:32:7:39 | source(...) : Number | user.kt:7:8:7:43 | new ConstructorWithDefaults(...) | flow path |
```
aschackmull, line 49: Could we make this an inline expectation test?

java/ql/lib/semmle/code/java/Member.qll
```diff
@@ -306,7 +306,9 @@ class Callable extends StmtParent, Member, @callable {
    */
   Callable getKotlinParameterDefaultsProxy() {
```
aschackmull, line 307: I hope this predicate is well-tested. I'll tentatively say it looks plausible, but it sure is involved.

java/ql/lib/semmle/code/java/dataflow/internal/FlowSummaryImplSpecific.qll
```diff
@@ -65,6 +65,61 @@ private boolean isGenerated(string provenance) {
...
+    exists(string regex |
+      // Note I use a regex and not AccessPathToken because this feeds summaryElement et al,
+      // which would introduce mutual recursion with the definition of AccessPathToken.
+      regex = "Argument\\[([0-9]+)\\](.*)" and
+      (
+        exists(string oldArgNumber, string rest, int paramOffset |
+          oldArgNumber = originalArgSpec.regexpCapture(regex, 1) and
+          rest = originalArgSpec.regexpCapture(regex, 2) and
+          paramOffset =
+            (
```
aschackmull, line 107: superfluous parenthesis

java/ql/lib/semmle/code/java/dataflow/internal/FlowSummaryImplSpecific.qll
```diff
@@ -65,6 +65,61 @@ private boolean isGenerated(string provenance) {
...
+  relatedArgSpec(originalCallable, originalArgSpec) and
+  defaultsCallable = originalCallable.getKotlinParameterDefaultsProxy() and
+  (
+    originalCallable instanceof Constructor and originalArgSpec = defaultsArgSpec
+    or
+    originalCallable instanceof Method and
+    exists(string regex |
+      // Note I use a regex and not AccessPathToken because this feeds summaryElement et al,
+      // which would introduce mutual recursion with the definition of AccessPathToken.
+      regex = "Argument\\[([0-9]+)\\](.*)" and
```
aschackmull, line 101: I believe the syntax allows for a richer specification of the position than just `[0-9]+`, I think you can also list multiple with ranges or comma-separation. I'm not sure whether we want to handle that level of complexity.

aschackmull (commented):
A couple of comments, otherwise looks plausible to me.

java/ql/integration-tests/posix-only/kotlin/default-parameter-mad-flow/test.expected
smowton: Done in https://github.com/github/codeql/pull/10876/commits/8d10b1b77bac85d3323b139cd76c4881f2007d9b

java/ql/lib/semmle/code/java/dataflow/internal/FlowSummaryImplSpecific.qll
smowton: Done in https://github.com/github/codeql/pull/10876/commits/7a0bded2aced4d08498435d5be84cfcaac7d6483

atorralba (approved):
Proxying @aschackmull approval after suggestions have been addressed.

PR #10883, opened 2022-10-19
RB: don't flag code-injection for dynamic loading where an attacker only controls a substring

erik-krogh:
I wasn't sure what to name the new predicate on `CodeExecution`. So I wrote the QLDoc and let Copilot pick a name. 

It's my first use of `DataFlow::FlowState`, but I think I figured it out.  

[Evaluation looks good](https://github.com/github/codeql-dca-main/tree/data/erik-krogh/pr-10883-b7d634__nightly__code-scanning__1/reports).  
Neutral performance, and removal of a few results that I think look like FPs.


ruby/ql/src/queries/security/cwe-094/CodeInjection.ql
```diff
@@ -21,6 +21,13 @@ import DataFlow::PathGraph
...
   config.hasFlowPath(source, sink) and
-  sourceNode = source.getNode()
+  sourceNode = source.getNode() and
+  // removing duplications of the same path, but different flow-labels.
+  sink =
+    min(DataFlow::PathNode otherSink |
+      config.hasFlowPath(any(DataFlow::PathNode s | s.getNode() = source.getNode()), otherSink)
+    |
+      otherSink order by otherSink.getState()
+    )
```
asgerf, line 31: Seems like this should be fixed in the shared libraries, the same way we did in JS. I've opened an internal issue for it.

asgerf (approved):

PR #10885, opened 2022-10-19
Kotlin: Add test case for false positive with modified captured variable

tamasvajk:


java/ql/test/kotlin/query-tests/ConstantLoopCondition/A.kt
```diff
@@ -0,0 +1,24 @@
+fun fn0(f: Function0<Unit>) = f()
+
+fun fn1() {
+    var c = true
+    while (c) {
+        fn0 {
+            c = false
```
aschackmull, line 7: So this is legal Kotlin? That's definitely not legal in Java. How are captured local variables compiled in Kotlin? This seems to indicate that our current handling of captured variables is completely wrong when applied to Kotlin.

aschackmull: We may need two different concepts to describe the two different kinds of variable capture, and we should be careful with adding general purpose predicates that falsely indicates that there's only one such concept.

java/ql/src/Likely Bugs/Termination/ConstantLoopCondition.ql
```diff
@@ -84,6 +84,9 @@ where
     not fa.getField().isFinal() and
     fa.getParent*() = cond
   ) and
-  not exists(ArrayAccess aa | aa.getParent*() = cond)
+  not exists(ArrayAccess aa | aa.getParent*() = cond) and
+  not exists(RValue use |
+    use.getParent*() = cond and use.getVariable().(LocalVariableDecl).isCaptured()
+  )
```
aschackmull, line 90: I think we should revert these changes to the query and revert the changes in `Variable.qll`. The only thing we should keep in this PR is a test indicating the FPs. The query as written ought to resolve those FPs once we switch our SSA implementation to the shared library and add proper handling of Kotlin-captured variables.

aschackmull (changes requested):

aschackmull (approved):

PR #10886, opened 2022-10-19
Dataflow: Fix a couple of join-orders.

aschackmull:
Easy wins.

### DIFF
--- a/cpp/ql/lib/experimental/semmle/code/cpp/ir/dataflow/internal/DataFlowImpl.qll
+++ b/cpp/ql/lib/experimental/semmle/code/cpp/ir/dataflow/internal/DataFlowImpl.qll
@@ -2629,6 +2629,7 @@ private predicate evalUnfold(AccessPathApprox apa, boolean unfold, Configuration
 /**
  * Gets the number of `AccessPath`s that correspond to `apa`.
  */
+pragma[assume_small_delta]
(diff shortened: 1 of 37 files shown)


geoffw0 (commented):
Changes seem reasonable to me.  Do you have any performance data / can you run DCA on this?

tausbn (dismissed):
Looks reasonable to me. 

Is `assume_small_delta` available in a currently-released version of the CodeQL CLI? If not, we might want to hold off on merging this until that is the case, so we don't suddenly break things for our users.

aschackmull:
> can you run DCA on this

Will do.



> Is `assume_small_delta` available in a currently-released version of the CodeQL CLI

It's currently bleeding edge, so you have to compile your own CLI to try it out, but it'll be in the next release. So yes, we should hold the merge until then.

aschackmull:
This should btw. be a noop in most common case (because the standard order won't be chosen), but it should measurably improve performance for queries with very large data flow graphs, i.e. on large databases or other otherwise poorly performing cases.

hvitved (dismissed):
Let's see what DCA says.

aschackmull:
Can't get dca for swift to work, but all other languages look uneventful, so I'd say LGTM from dca.

MathiasVP (approved):

PR #10887, opened 2022-10-19
Python: Add TarSlip Improv query

Sim4n6:
- Re-adapted the PR https://github.com/github/codeql/pull/10851 into this one. 
 - added an additional taint step.

### DIFF
--- a/python/ql/src/experimental/Security/CWE-022bis/TarSlipImprov.ql
+++ b/python/ql/src/experimental/Security/CWE-022bis/TarSlipImprov.ql
@@ -49,6 +49,7 @@ class AllTarfileOpens extends API::CallNode {
     )
   }
 }
+
 /**
  * A taint-tracking configuration for detecting more "TarSlip" vulnerabilities.
  */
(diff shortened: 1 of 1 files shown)


python/ql/src/experimental/Security/CWE-022bis/TarSlipImprov.ql
```diff
@@ -0,0 +1,108 @@
...
+    this = tarfileOpen().getACall()
+    or
+    exists(API::Node closing, Node arg |
+      closing = API::moduleImport("contextlib").getMember("closing") and
+      this = closing.getACall() and
+      arg = this.getArg(0) and
+      arg = tarfileOpen().getACall()
+    )
+  }
+}
```
yoff, line 51: I would write this as a predicate like `fileOpen`.

I would also create a taint-step for the `closing`-call. For code like
```python
with closing(tarfile.open('example.tar', 'r')) as t:
    t.extractall('outdir')
```
we want a step from the argument of `closing()` to the call itself, whenever the argument is a value reachable from `tarfileOpen`. Then taint will automatically end up at `t`.

If you add such a taint-step, you only need `tarfileOpen()` as sources, and the paths will be nicer (but you still need `allTarfileOpen` to define sinks).

Sim4n6: So, if I understood well, you suggest that I simplify the Sources with an additional taint while keeping the AllTarfileOpens for the sinks, right?

python/ql/src/experimental/Security/CWE-022bis/TarSlipImprov.ql
```diff
@@ -0,0 +1,108 @@
...
+    )
+  }
+
+  override predicate isAdditionalTaintStep(DataFlow::Node nodeFrom, DataFlow::Node nodeTo) {
+    exists(AttrRead attr, MethodCallNode call |
+      attr.accesses(nodeFrom, "getmembers") and
+      nodeFrom = call.getObject() and
+      nodeTo = call
+    )
+  }
```
yoff, line 109: This now adds steps to any call to `getmembers` from any object. It would be good to restrict it to relevant objects.

Sim4n6: done in [9163cbe](https://github.com/github/codeql/pull/10887/commits/9163cbec094a376504091b3591186e6aa01f14c2)

python/ql/src/experimental/Security/CWE-022bis/TarSlipImprov.ql
```diff
@@ -0,0 +1,116 @@
...
+      nodeFrom instanceof AllTarfileOpens and
+      nodeTo = call
+    )
+    or
+    exists(DataFlow::CallCfgNode closing |
+      closing = API::moduleImport("contextlib").getMember("closing").getACall() and
+      nodeFrom = closing.getArg(0) and
+      nodeFrom = tarfileOpen().getACall() and
+      nodeTo = closing
+    )
```
yoff, line 109: Good, this is exactly what I meant. There is one opportunity for improvement here, if we use the `API::CallNode` (which extends `DataFlow::CallCfgNode`):
```suggestion
    exists(API::CallNode closing |
      closing = API::moduleImport("contextlib").getMember("closing").getACall() and
      nodeFrom = closing.getArg(0) and
      nodeFrom = tarfileOpen().getReturn().getAValueReachingSink() and
      nodeTo = closing
    )
```
This way, we would also handle code like
```python
f = tarfile.open(...)
...
with closing(f) as t:
   ...
```

yoff (changes requested):
You have done all that I asked, but I could not help showing you one final little trick :-)
To answer your question about query tests (asked on the previous PR), the existing one for tar slip is [here](https://github.com/github/codeql/tree/main/python/ql/test/query-tests/Security/CWE-022-TarSlip). So one approach would be to simply move your examples into that directory (or rather, the equivalent one for `experimental`) along with an `.expected`-file and a `.qlref`-file. Ideally, we would use inline expectations like in our concept tests ([example](https://github.com/github/codeql/blob/main/python/ql/test/library-tests/frameworks/stdlib/Logging.py)), but we do not really have a convenient way for doing that yet.

I am also thinking that, while I sympathise deeply with the idea of executable tests, we might not want to check in a malicious tarball..

python/ql/src/experimental/Security/CWE-022bis/TarSlipImprov.ql
yoff, line 109: Good, this is the step I wanted. However, now it is only added if the argument looks like an `open`-call, and we would like to also handle cases like
```python
f = tar file.open(...)
...
with closing(f) as t:
   ...
```
The data flow will be able to find this path, except the crucial step from `f` to `t` will not be there.
To fix this, while still restricting the argument, we can use the `API::CallNode` (which extends `DataFlowCallCfgNode`):
```suggestion
    exists(API::CallNode closing |
      closing = API::moduleImport("contextlib").getMember("closing").getACall() and
      nodeFrom = closing.getArg(0) and
      nodeFrom = tarfileOpen().getReturn().getAValueReachingSink() and
      nodeTo = closing
    )
```

Sim4n6:
I'm not gonna lie to you. The expected file for tests is a bit modified since it got me files like /usr/../python3.10/tarfile.py.

I've run the following command to generate the expected file:
```shell
Codeql run test --learn ./TarSlip.qlref .
```

And then deleted some entries. It's messy, but I don't know how to proceed, please.

Sim4n6:
By the way, I really appreciate the effort you put to well describe each step.

Sim4n6:
Now, that updated the query to omit sinks within std lib files, the tests are getting better.

yoff:
> I'm not gonna lie to you. The expected file for tests is a bit modified since it got me files like /usr/../python3.10/tarfile.py.

Does that mean that the tests do not pass?
You may need to filter that out explicitly [like we do here](https://github.com/github/codeql/blob/b9f1cc5c6f129ca7238e4096c45636da53696f90/python/ql/lib/semmle/python/security/dataflow/TarSlipCustomizations.qll#L44). I guess `inStdLib` is flaky as it makes some assumptions about directory layouts... (but I appreciate that you found that, it is a nicer, more semantic way of excluding files. Perhaps we could fix that up at some point so it can be used reliably).

If the tests can be made to pass, this looks ready :-) 👍

yoff:
> Now, that updated the query to omit sinks within std lib files, the tests are getting better.

If your tests pass locally, let me know and I will start the run here..

Sim4n6:
All tests are PASSing using the filtering out with `InStdLib`.

yoff:
Cool, then let us see what CI says :-)

yoff (approved):
LGTM - Thanks :-)

Sim4n6:
Thank you for this wonderful learning experience.

Sim4n6:
Hi @yoff ,one last detail please. Regarding https://github.com/github/securitylab/issues/710 do I have to write a new issue there. It seems to remain within **results analysis**...

yoff:
No need, we decided to run the final version of the query to check if the scope has been extended :-)

Sim4n6:
perfect

PR #10889, opened 2022-10-19
Kotlin: Don't warn on extracting an enum-entry class

smowton:
I don't think these should get `writeIsEnumType`, since they aren't the enum itself, but rather a specialisation used to realise one of its arms. That can be reconstructed QL-side by looking for inheritance of an enum type if need be.


java/ql/test/kotlin/library-tests/methods/diagnostics.expected
```diff
@@ -0,0 +1 @@
+| enumClass.kt:8:3:11:4 | Unrecognised class kind ENUM_ENTRY |
```
tamasvajk, line 1: Shouldn't this diagnostic be removed instead of added?

smowton: Doh, yes, committed the wrong test output; fixed now.

tamasvajk (approved):

PR #10890, opened 2022-10-19
Java: Add flow summaries for startActivities

atorralba:
Uses `SyntheticCallable`s and `SyntheticGlobal`s to pair each `startActivities` call to `getIntent` calls in the components targeted by the intent(s).

Props to @aschackmull for this, which is the proper solution to the problem we discussed in https://github.com/github/codeql/pull/10660#discussion_r986814913.

### DIFF
--- a/java/ql/lib/semmle/code/java/frameworks/android/Intent.qll
+++ b/java/ql/lib/semmle/code/java/frameworks/android/Intent.qll
@@ -1,7 +1,10 @@
 import java
+private import semmle.code.java.frameworks.android.Android
 private import semmle.code.java.dataflow.DataFlow
 private import semmle.code.java.dataflow.ExternalFlow
 private import semmle.code.java.dataflow.FlowSteps
+private import semmle.code.java.dataflow.FlowSummary
+private import semmle.code.java.dataflow.internal.BaseSSA as BaseSSA
 
 /** The class `android.content.Intent`. */
 class TypeIntent extends Class {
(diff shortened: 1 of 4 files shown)


java/ql/lib/semmle/code/java/frameworks/android/Intent.qll
```diff
@@ -271,6 +307,87 @@ private class StartActivityIntentStep extends AdditionalValueStep {
...
+      isTargetableType(targetType, id)
+    )
+  }
+
+  AndroidComponent getTargetType() { result = targetType }
+}
+
+private class RequiredComponentStackForStartActivities extends RequiredSummaryComponentStack {
+  override predicate required(SummaryComponent head, SummaryComponentStack tail) {
+    head = SummaryComponent::element() and
```
aschackmull, line 386: This actually needs to be arrayElement now that you fixed the summary above.

aschackmull: It accidentally works because there are existing `Argument[0].ArrayElement` specs in some MaD rows, so this `RequiredComponentStackForStartActivities` is somewhat superfluous, but best to have it to complete the pattern. We can simplify the `propagatesFlow` interface to use MaD spec strings like `Argument[0].ArrayElement` instead in a future PR - that's probably nice to do to avoid needing a `RequiredSummaryComponentStack` here.

atorralba: True, I forgot about that, thanks! Fixed in 0678b06.

> We can simplify the propagatesFlow interface to use MaD spec strings like Argument[0].ArrayElement instead in a future PR - that's probably nice to do to avoid needing a RequiredSummaryComponentStack here.

That'd be nice to have indeed.

atorralba:
@aschackmull do you think this is ready to merge?

aschackmull (approved):
> @aschackmull do you think this is ready to merge?

Looks reasonable to me.

PR #10892, opened 2022-10-19
Swift: Add a new Custom URL Scheme source

atorralba:
The `launchOption` parameter in [`application(_:,didFinishLaunchingWithOptions:)`](https://developer.apple.com/documentation/uikit/uiapplicationdelegate/1623032-application) and [`application(_:willFinishLaunchingWithOptions:)`](https://developer.apple.com/documentation/uikit/uiapplicationdelegate/1622921-application) contains a `url` value that can also contain a remotely-provided URL through custom URL schemes.

Note that this had to be modeled with plain CodeQL because we currently don't have support for `MapValue` contents in Swift. We should come back to this once we do and switch to MaD.

This PR also adds a couple of data flow steps to model flow through `?` expressions.


geoffw0 (commented):
LGTM.  Couple of minor comments.

geoffw0 (approved):

swift/ql/lib/codeql/swift/dataflow/internal/DataFlowPrivate.qll
```diff
@@ -148,6 +148,11 @@ private module Cached {
     // flow through `!`
     nodeFrom.asExpr() = nodeTo.asExpr().(ForceValueExpr).getSubExpr()
     or
+    // flow through `?`
+    nodeFrom.asExpr() = nodeTo.asExpr().(BindOptionalExpr).getSubExpr()
+    or
+    nodeFrom.asExpr() = nodeTo.asExpr().(OptionalEvaluationExpr).getSubExpr()
+    or
```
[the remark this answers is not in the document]
atorralba: Sorry for the wait @geoffw0, I intended to quickly add tests for this but fell into a rabbit hole where a MaD row I needed to add so that the optional access could be tested didn't work as expected. To not waste more time, I've added the test cases with `MISSING` tags — we should be able to remove those once we fix the root cause of the MaD row not working.

swift/ql/test/library-tests/dataflow/dataflow/FlowConfig.qll
```diff
@@ -14,10 +15,17 @@ class TestConfiguration extends DataFlow::Configuration {
...
       sinkCall.getAnArgument().getExpr() = sink.asExpr()
     )
   }
 
   override int explorationLimit() { result = 100 }
 }
+
+private class TestSummaries extends SummaryModelCsv {
+  override predicate row(string row) {
+    // dumb model to allow testing flow through optional chaining (`x?.signum()`)
```
geoffw0, line 28:
```suggestion
    // model to allow data flow through `signum()` as though it were an identity function, for the benefit of testing flow through optional chaining (`x?.`).
```
I'm not sure `signum` is the right thing to use in this test as this would really be taint flow, not data flow.  I can see that this is difficult to test any other way though.

geoffw0 (commented):
Otherwise LGTM.

swift/ql/test/library-tests/dataflow/dataflow/FlowConfig.qll
atorralba: Yes, I struggled to find an `Int` method that was value-preserving. I could've used a different type too, but it started to look like too many changes for testing such a little thing.

geoffw0 (dismissed):
:+1:

geoffw0 (approved):

PR #10894, opened 2022-10-19
Kotlin: Refactor PSI handling

igfoo:
We were giving warnings about comments, when we were actually trying to populate numlines.

### DIFF
--- a/java/kotlin-extractor/src/main/kotlin/LinesOfCode.kt
+++ b/java/kotlin-extractor/src/main/kotlin/LinesOfCode.kt
@@ -16,9 +16,8 @@ class LinesOfCode(
     val tw: FileTrapWriter,
     val file: IrFile
 ) {
-    val psi2Ir = getPsi2Ir(logger)
-    init {
-        if (psi2Ir == null) {
+    val psi2Ir = getPsi2Ir(logger).also {
+        if (it == null) {
             logger.warn("Lines of code will not be populated as Kotlin version is too old (${KotlinCompilerVersion.getVersion()})")
         }
     }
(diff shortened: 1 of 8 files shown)


java/kotlin-extractor/src/main/kotlin/LinesOfCode.kt
```diff
@@ -15,9 +16,17 @@ class LinesOfCode(
     val tw: FileTrapWriter,
     val file: IrFile
 ) {
-    val psi2Ir = Psi2Ir(logger)
+    val psi2Ir = getPsi2Ir(logger)
+    init {
+        if (psi2Ir == null) {
+            logger.warn("Lines of code will not be populated as Kotlin version is too old (${KotlinCompilerVersion.getVersion()})")
+        }
+    }
```
smowton, line 24:
```suggestion
    val psi2Ir = getPsi2Ir(logger) ?: run { 
      logger.warn("Lines of code will not be populated as Kotlin version is too old (${KotlinCompilerVersion.getVersion()})")
    }
```

smowton (approved):
Looks sensible; one suggestion for brevity

java/kotlin-extractor/src/main/kotlin/LinesOfCode.kt
igfoo: I think that this would also need a `; null` in the `run` to get the right result. I've switched it to use `.also` instead, which I think is nicer than a separate `init` block. It doesn't get us the concise `?:` null check, but I think that it's more natural to "also" do something that doesn't have a result.

PR #10899, opened 2022-10-20
Ruby: Document flow summary syntax

hmac:
This PR adds some internal documentation for the Ruby flow summary syntax. There's a few TODOs left where I'm not sure of some behaviour - if you know the answer, let me know and I'll fill them in.

### DIFF
--- a/ruby/ql/docs/flow_summaries.md
+++ b/ruby/ql/docs/flow_summaries.md
@@ -42,8 +42,9 @@ Any call to `chomp` in the database will be translated, in the dataflow graph,
 to a call to this fake definition. 
 
 `input` and `output` define the "from" and "to" locations in the flow summary.
-They use a custom string-based syntax which is similar but not identical to
-Models as Data. These strings are often referred to as access paths.
+They use a custom string-based syntax which is similar to that used in `path`
+column in the Models as Data format. These strings are often referred to as
+access paths.
 
 Note: The behaviour documented below is tested in
 `dataflow/flow-summaries/behaviour.ql`. Where specific quirks exist, we may


aibaars:
@hmac Could you split this PR? Documentation changes for the GA should go into `codeql-cli-2.11.2` , but QL changes should go to `main`.

aibaars:
> I don't know why this has a GA label - these are internal docs (and also this PR is unfinished). I've removed it.

Ok great! I mistakenly labelled it as GA, sorry for that.

ruby/ql/docs/flow_summaries.md
```diff
@@ -0,0 +1,205 @@
...
+
+If `preservesValue = true` then value flow is propagated. If it is `false` then
+only taint flow is propagated.
+
+Any call to `chomp` in the database will be translated, in the dataflow graph,
+to a call to this fake definition. 
+
+`input` and `output` define the "from" and "to" locations in the flow summary.
+They use a custom string-based syntax which is similar but not identical to
+Models as Data. These strings are often referred to as access paths.
```
hvitved, line 46: These strings should be identical to the `input` and `output` string in MaD.

ruby/ql/docs/flow_summaries.md
```diff
@@ -0,0 +1,205 @@
...
+The block argument passed to the call, if any.
+
+#### `any`
+Any argument to the call, except `self` or `block` arguments.
+
+#### `any-named`
+Any keyword argument to the call.
+
+#### `hash-splat`
+The special "hash splat" argument/parameter, which is written as `**args`.
```
hvitved, line 102: Note that this will also include all explicit keyword arguments, wrapped in an implicit hash splat argument.

ruby/ql/docs/flow_summaries.md
```diff
@@ -0,0 +1,205 @@
...
+
+## `ReturnValue`
+`ReturnValue` refers to the return value of the element identified in the
+preceding access path. For example, `Argument[0].ReturnValue` refers to the
+return value of the first argument. Of course this only makes sense if the first
+argument is a callable.
```
hvitved, line 108: Maybe `callback` instead of `callable`.

ruby/ql/docs/flow_summaries.md
```diff
@@ -0,0 +1,205 @@
...
+for `Hash#to_h` is
+
+```ql
+input = "Argument[self].WithElement[any]" and
+output = "ReturnValue" and
+preservesValue = true
+```
+
+TODO: I've not seen this component used in an output path; I don't know if it makes
+sense to do so, or what meaning it would have.
```
hvitved, line 183: Doesn't make sense to use in ouput.

ruby/ql/docs/flow_summaries.md
```diff
@@ -0,0 +1,205 @@
...
+```ql
+input = "Argument[0].WithoutElement[0]" and
+output = "ReturnValue"
+```
+
+any data in any index of the first argument will be copied to the return value,
+with the exception of data at index 0.
+
+TODO: I've not seen this component used in an output path; I don't know if it makes
+sense to do so, or what meaning it would have.
```
hvitved, line 203: No, doesn't make sense.

ruby/ql/docs/flow_summaries.md
```diff
@@ -0,0 +1,205 @@
...
+relevant elements when copying from input to output. For example in the
+following summary:
+
+```ql
+input = "Argument[0].WithoutElement[0]" and
+output = "ReturnValue"
+```
+
+any data in any index of the first argument will be copied to the return value,
+with the exception of data at index 0.
```
hvitved, line 200: A prominent use case of `WithoutElement` is when you want to model a method that removes (a set of) element(s) from a collection. For example, if we want to model a method that removes the first element of the receiver collection, we can do so via

```ql
input = "Argument[self].WithoutElement[0]" and
output = "Argument[self]"
```

where both the input and the output refer to the same argument. The data-flow library will then prevent use-use flow from any corresponding receivers:

```rb
a[0] = taint
a[1] = taint
a.remove_first # use-use flow from `a` to the `a` below will be blocked, but there will still be flow from `[post-update] a` to `a` below
sink(a[0]) # good
sink(a[1]) # bad
```

hvitved (commented):
Thanks for the excellent write-up.

calumgrant:
Ping @hmac - ok to merge (after resolving conflicts)?

ruby/ql/docs/flow_summaries.md
hvitved: Ok, I see what you mean. However, I think of MaD collectively as all the components that make up a summary; but you are right that the access path component (which identifies the relevant calls) maps to API graphs, where as the input/output components have their own syntax.

ruby/ql/docs/flow_summaries.md
hvitved: 👍

ruby/ql/docs/flow_summaries.md
hvitved: A good example of this is something like `Hash#except`. If we see a call like `hash.except(:foo)`, we would like that to map to a flow summary with
```
input = Argument[self].WithoutElement[":foo"].WithElement[any]
output = ReturnValue
```
I.e. data can flow out of the call, but only when it is stored in _some_ element that is not known to be at index `:foo`.

hvitved: This would also incorrectly include flow for `sink(source("d").s53())`.

hvitved: Yes, that looks good. `WithoutElement[x]` just says that data must _not_ be stored inside a collection at index `x`, not that data _must_ be stored inside a collection.

This is also the reason why something like a `reverse` method needs a summary like
```ql
input = "Argument[self].WithElement[any]" and
output = "ReturnValue"
```
and not
```ql
input = "Argument[self]" and
output = "ReturnValue"
```

ruby/ql/docs/flow_summaries.md
asgerf: I don't think we should call the difference with the MaD `path` column just now (called the "identifying access path"). We can elaborate on this when documenting the MaD format.

aibaars (approved):

PR #10903, opened 2022-10-20
Swift: Add some summary queries.

geoffw0:
I wrote these queries to assess the progress we're making identifying flow sources and sensitive expressions in real world Swift snapshots.  I intend to track these things fairly often in the near-mid future, so it would be convenient to have the queries in the public repo.  I don't think there's any harm in letting curious users see them as well.

It might be nice to have `LinesOfCode.ql` and `LinesOfUserCode.ql` summary queries as well as all (?) other languages do, but I don't think we extract the data for them yet.

### DIFF
--- a/swift/ql/src/queries/Summary/SummaryStats.ql
+++ b/swift/ql/src/queries/Summary/SummaryStats.ql
@@ -5,6 +5,7 @@
  *              features interesting to analysis that have been found.
  * @kind table
  * @id swift/summary/summary-statistics
+ * @tags summary
  */
 
 import swift
--- a/swift/ql/src/queries/Summary/SummaryStats.ql
+++ b/swift/ql/src/queries/Summary/SummaryStats.ql
@@ -13,7 +13,7 @@ import codeql.swift.security.SensitiveExprs
 predicate statistic(string what, int value) {
   what = "Files" and value = count(File f)
   or
-  what = "Expressions" and value = count(Expr e | e.getFile().getName() != "")
+  what = "Expressions" and value = count(Expr e | not e.getFile() instanceof UnknownFile)
   or
   what = "Remote flow sources" and value = count(RemoteFlowSource s)
   or
--- a/swift/ql/src/queries/Summary/SummaryStats.ql
+++ b/swift/ql/src/queries/Summary/SummaryStats.ql
@@ -1,0 +1,26 @@
+/**
+ * @name Summary statistics
+ * @description A table of summary statistics about a database. Includes
+ *              values that measure its size, and the numbers of certain
+ *              features interesting to analysis that have been found.
+ * @kind table
+ * @id swift/summary/summary-statistics
+ */
+
+import swift
+import codeql.swift.dataflow.FlowSources
+import codeql.swift.security.SensitiveExprs
+
+predicate statistic(string what, int value) {
+  what = "Files" and value = count(File f)
+  or
+  what = "Expressions" and value = count(Expr e)
+  or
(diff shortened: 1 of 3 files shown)


swift/ql/src/queries/Summary/SummaryStats.ql
```diff
@@ -0,0 +1,27 @@
+/**
+ * @name Summary statistics
+ * @description A table of summary statistics about a database. Includes
+ *              values that measure its size, and the numbers of certain
+ *              features interesting to analysis that have been found.
```
jketema, line 5: I have a very hard time parsing "and the numbers of certain features interesting to analysis that have been found". What does "that have been found" refer to? "features" or "analysis"? Maybe just drop that part? I don't think that's quite sufficient for me to understand what you're trying to say, but it's a step.

jketema (commented):
Mostly LGTM. Just one description I do not fully understand.

swift/ql/src/queries/Summary/SummaryStats.ql
geoffw0: You're right.  I've removed it.  Keep it simple.

geoffw0:
In addition to your suggested change I've restricted the count of "Expressions" to ones with locations in code now.  It turns out there's quite a lot of locationless stuff that was exaggerating the size of smaller projects (and in turn making me wonder why we weren't finding many results, taint sources etc in them).

atorralba (dismissed):
This LGTM. I guess summary queries don't need a docs review.

swift/ql/src/queries/Summary/SummaryStats.ql
```diff
@@ -0,0 +1,25 @@
...
+ */
+
+import swift
+import codeql.swift.dataflow.FlowSources
+import codeql.swift.security.SensitiveExprs
+
+predicate statistic(string what, int value) {
+  what = "Files" and value = count(File f)
+  or
+  what = "Expressions" and value = count(Expr e | e.getFile().getName() != "")
```
jketema, line 16: For C++ we have:
```
class UnknownLocation extends Location {
  UnknownLocation() { this.getFile().getAbsolutePath() = "" }
}
```
Doesn't that exist for Swift?

MathiasVP: Indeed, Swift also has an `UnknownLocation` class: https://github.com/github/codeql/blob/30fc6acb19b5ddf8a23dd667d41b8750931bc643/swift/ql/lib/codeql/swift/elements/UnknownLocation.qll#L2.

geoffw0: If it does, I should use it.  If it doesn't, I should add it...

geoffw0:
> I guess summary queries don't need a docs review.

As far as I can tell summary queries don't usually get full `.qhelp` documentation, just an explanatory `@description`.  They're not really as visible as our main 'problem' queries so this seems appropriate to me.  Happy to add full docs, or get the `@description`s reviewed by the doc team if you would prefer.

swift/ql/src/queries/Summary/SummaryStats.ql
geoffw0: Thanks @MathiasVP !

atorralba:
> Happy to add full docs, or get the @descriptions reviewed by the doc team if you would prefer.

No, let's keep it consistent with the rest of summary queries. Looks good as it is.

jketema (approved):

PR #10907, opened 2022-10-20
QL language spec: pragma[assume_small_delta]

cklin:


ginsbach (approved):
This seems like a very minimal documentation - but I see no problem with building it up incrementally. Therefore, LGTM.

PR #10909, opened 2022-10-20
Kotlin: Fix varargs dataflow, and varargs default handling

smowton:
Dataflow requires accounting for the fact that the varargs parameter isn't necessarily last in the parameter list in a couple more places. Default handling just requires that if the only null parameter is the varargs argument, and it has no default value, then no $default method is required-- the caller is expected to simply pass nothing (at QL / source level) or an empty array (at JVM level).


java/kotlin-extractor/src/main/kotlin/KotlinFileExtractor.kt
```diff
@@ -15,6 +15,7 @@ import org.jetbrains.kotlin.descriptors.java.JavaVisibilities
 import org.jetbrains.kotlin.ir.IrElement
 import org.jetbrains.kotlin.ir.IrStatement
 import org.jetbrains.kotlin.ir.ObsoleteDescriptorBasedAPI
+import org.jetbrains.kotlin.ir.backend.js.transformers.irToJs.varargParameterIndex
```
tamasvajk, line 18: This import seems suspicious.

tamasvajk (commented):
Found one minor issue, otherwise looks good to me.

java/kotlin-extractor/src/main/kotlin/KotlinFileExtractor.kt
smowton: Removed

tamasvajk (approved):

aschackmull (approved):
QL changes LGTM.

PR #10910, opened 2022-10-20
Ruby: Refactor call graph logic for singleton methods

hvitved:
Refactored to follow the same pattern as for singleton methods. This allows for subsequent similar call-context-sensitive singleton resolution (follow-up PR), and may as well reduce tuple duplication (the same method may otherwise be looked up in many sub classes).

DCA confirms no changes to the call graph.


asgerf (approved):
👍

PR #10911, opened 2022-10-20
Python: add `debug based on location` snippet

RasmusWL:


tausbn (approved):
I feel like there's a somewhat painful story behind this. 😛

RasmusWL:
> I feel like there's a somewhat painful story behind this.

I usually end up doing this for selecting a single sink/source/thing when debugging

PR #10913, opened 2022-10-20
Add Faraday::Connection.new as sink for SSRF query

thiggy1342:
This PR aims to expand the sinks for the Ruby SSRF query to catch cases where `Faraday::Connection.new` is used to instantiate new Faraday Connections instead of `Faraday.new`. This PR accomplishes that by adding this instantiation to the Faraday model as part of the `FaradayHttpRequests` function. I've also updated the framework tests and added two new code snippets to the query test to demonstrate what we're looking to capture with this change.

### DIFF
--- a/ruby/ql/lib/change-notes/2022-10-20-expand-faraday-model-for-ssrf-sink
+++ b/ruby/ql/lib/change-notes/2022-10-20-expand-faraday-model-for-ssrf-sink
@@ -1,0 +1,5 @@
+---
+category: minorAnalysis
+---
+
+- Instatiations using `Faraday::Connection.new` are now recognized as part of `FaradayHttpRequest`s, meaning they will be considered as sinks for queries such as rb/request-forgery.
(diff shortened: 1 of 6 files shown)


ruby/ql/lib/change-notes/2022-10-20-expand-faraday-model-for-ssrf-sink
```diff
@@ -0,0 +1,5 @@
+---
+category: minorAnalysis
+---
+
+- Instatiations using `Faraday::Connection.new` are now recognized as part of `FaradayHttpRequest`s, meaning they will be considered as sinks for queries such as rb/request-forgery.
```
rzhade3, line 5:
```suggestion
- Instantiations using `Faraday::Connection.new` are now recognized as part of `FaradayHttpRequest`s, meaning they will be considered as sinks for queries such as `rb/request-forgery`.
```

thiggy1342:
Thanks @hmac that did the trick!

PR #10919, opened 2022-10-21
QL language spec: Document built-in equivalence relation module

kaspersv:


docs/codeql/ql-language-reference/modules.rst
```diff
@@ -273,4 +273,63 @@ The ``<module_expression>`` itself can be a module name, a selection, or a quali
...
+
+.. code-block:: ql
+
+    newtype Node = MkNode(int x) {  x in [1..6] }
+    predicate base(Node x, Node y) {(x = MkNode(1) and y = MkNode(2)) or (x = MkNode(3) and y = MkNode(4)) }
+    module Equiv = QlBuiltins::EquivalenceRelation<Node, base/2>;
+
+    from int x, int y
+    where Equiv::getEquivalenceClass(MkNode(x)) = Equiv::getEquivalenceClass(MkNode(y))
+    select x, y
```
aschackmull, line 309: This should be auto-formatted:
```suggestion
    newtype Node = MkNode(int x) { x in [1 .. 6] }

    predicate base(Node x, Node y) {
      x = MkNode(1) and y = MkNode(2)
      or
      x = MkNode(3) and y = MkNode(4)
    }

    module Equiv = QlBuiltins::EquivalenceRelation<Node, base/2>;

    from int x, int y
    where Equiv::getEquivalenceClass(MkNode(x)) = Equiv::getEquivalenceClass(MkNode(y))
    select x, y
```

docs/codeql/ql-language-reference/modules.rst
```diff
@@ -273,4 +273,63 @@ The ``<module_expression>`` itself can be a module name, a selection, or a quali
...
 For information about how import statements are looked up, see "`Module resolution <https://codeql.github.com/docs/ql-language-reference/ql-language-specification/#module-resolution>`__"
-in the QL language specification. 
\ No newline at end of file
+in the QL language specification. 
+
+Built-in modules
+****************
+
+QL defines a `QlBuiltins` module, which is a built-in module that all databases
+include. Currently, it defines a single parameterized sub-module
```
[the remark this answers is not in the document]
kaspersv: I agree that it reads weird. It was written like this to keep it consistent with the existing [Built-ins](https://codeql.github.com/docs/ql-language-reference/ql-language-specification/#built-ins) page. I am happy to change it.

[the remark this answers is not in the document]
aschackmull: Hmm, right, I guess that's because databases actually contain predicates, so it makes a bit more sense there even though that also reads a bit weird.

docs/codeql/ql-language-reference/modules.rst
```diff
@@ -273,4 +273,69 @@ The ``<module_expression>`` itself can be a module name, a selection, or a quali
...
 
 For information about how import statements are looked up, see "`Module resolution <https://codeql.github.com/docs/ql-language-reference/ql-language-specification/#module-resolution>`__"
-in the QL language specification. 
\ No newline at end of file
+in the QL language specification. 
+
+Built-in modules
+****************
+
+QL defines a `QlBuiltins` module, which is a built-in module that is always in scope.
```
jbj, line 281:
```suggestion
QL defines a `QlBuiltins` module that is always in scope.
```
This is shorter, and I think it communicates the same thing.

docs/codeql/ql-language-reference/modules.rst
```diff
@@ -273,4 +273,69 @@ The ``<module_expression>`` itself can be a module name, a selection, or a quali
...
+Currently, it defines a single parameterized sub-module
+`EquivalenceRelation`, that provides an efficient abstraction for working with
```
jbj, line 283:
```suggestion
``EquivalenceRelation``, that provides an efficient abstraction for working with
```

Looking at the preview and at the rest of this file, it seems that identifiers should be in double backticks. This also applies to everything below here.

docs/codeql/ql-language-reference/modules.rst
```diff
@@ -273,4 +273,69 @@ The ``<module_expression>`` itself can be a module name, a selection, or a quali
...
+(partial) equivalence relations in QL.
+
+Equivalence relations
+=====================
+
+The built-in `EquivalenceRelation` module is parameterized by a type `T` and a
+binary base relation `eq` on `T`. The symmetric and transitive closure of `eq`
```
jbj, line 290: I don't think `eq` is a good name for the relation that's _not_ an equivalence relation. Other possible names: `base`, `baseRelation`, `rel`. I like `base`, and that's also what you're using in the example below.

docs/codeql/ql-language-reference/modules.rst
```diff
@@ -273,4 +273,69 @@ The ``<module_expression>`` itself can be a module name, a selection, or a quali
...
+induces a partial equivalence relation on `T`. If `eq` is reflexive, then the
+induced relation is an equivalence relation on `T`.
```
jbj, line 292:
```suggestion
induces a partial equivalence relation on `T`. If every value of `T` appears in `eq`, then the
induced relation is an equivalence relation on `T`.
```

I know this is equivalent, but that may not be apparent to everyone.

docs/codeql/ql-language-reference/modules.rst
```diff
@@ -273,4 +273,69 @@ The ``<module_expression>`` itself can be a module name, a selection, or a quali
...
+The `EquivalenceRelation` module exports a `getEquivalenceClass` predicate that
+gets the equivalence class, if any, associated with a given `T` element by the
+(partial) equivalence relation induced by `eq`.
+
+The following example illustrates an application of the `EquivalenceRelation`
+module to generate a custom equivalence relation on an :ref:`algebraic datatype <algebraic-datatypes>`:
+
+.. code-block:: ql
+
+    newtype Node = MkNode(int x) { x in [1 .. 6] }
```
jbj, line 303: `newtype` is a somewhat under-documented feature, and users might want to learn about `EquivalenceRelation` without first having to learn about `newtype`. Can you change this into

```
class Node extends int {
  Node() { this in [1 .. 6] }
}
```
I forget whether a `toString` override is mandatory here. Hopefully not if you have `int`s in the `from` clause.

jbj (commented):
Other than these comments, the new docs LGTM!

docs/codeql/ql-language-reference/modules.rst
kaspersv: Good call. Yes, it works without the `toString` because we are selecting `int`s.

jbj (approved):

PR #10923, opened 2022-10-21
Download GitHub database: fix `gh` invocation

dscho:
At least as of GitHub CLI v2.13.0, the leading slash in `gh api /repos/...` leads to a 404. In all GitHub CLI versions, the trailing slash of `gh api .../databases/` seems to not be acceptable, either.

### DIFF
--- a/docs/codeql/reusables/download-github-database.rst
+++ b/docs/codeql/reusables/download-github-database.rst
@@ -3,12 +3,12 @@ GitHub stores CodeQL databases for over 200,000 repos on GitHub.com, which you c
 You can check if a repository has any CodeQL databases available for download using the ``/repos/<owner>/<repo>/code-scanning/codeql/databases`` endpoint.
 For example, to check for CodeQL databases using the `GitHub CLI <https://cli.github.com/manual/gh_api>`__ you would run::
 
-   gh api /repos/<owner>/<repo>/code-scanning/codeql/databases
+   gh api repos/<owner>/<repo>/code-scanning/codeql/databases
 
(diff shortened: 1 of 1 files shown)


dscho:
I should clarify that this happens on Windows, in Git Bash.

smowton (approved):
This appears to be accurate

smowton (approved):

dscho:
@jketema so how is this PR supposed to proceed from here?

jketema:
@dscho  With gh version 2.74.2 it looks like that it works both with and without the `/` on both macOS and Ubuntu. Is this still broken with Git bash on Windows (this is not something I can easily check)?

dscho:
@jketema thank you for your response!

> With gh version 2.74.2 it looks like that it works both with and without the `/` on both macOS and Ubuntu.

The crucial part is "on both macOS and Ubuntu", leaving out the still most prevalent platform: Windows.

> Is this still broken with Git bash on Windows

Yes.

The reason is described in https://www.msys2.org/docs/filesystem-paths/: Git Bash needs to automatically convert paths on the command-line which look like Unix paths to look like Windows paths instead, when calling a Win32 program (which `gh.exe` is). And therefore it mistakes the `/repos/...` argument for a path in need of this conversion, leading to the 404.

I've updated the commit message to reflect this.

> (this is not something I can easily check)?

I guess you could use the `mxschmitt/action-tmate` Action to check ;-)

jketema:
@dscho thanks for your answer. Since either works on macOS and Ubuntu, I'm happy with this change.

PR #10924, opened 2022-10-21
Swift: add qltest tests and fix its failure reporting

redsun82:
`qltest.sh` was not failing on extractor errors since 946e1f498a772c3778d30324d36c66984ac01c88 when it was generalized to extract multiple source files, as it was ignoring the return code of those multiple extractor calls.

This fixes that issue and adds some tests of `qltest.sh` to avoid similar problems in the future.

After the fix it turned out there were some tests with swift compile errors. This was fixed in 408968a41737a8f5d22104c0593e1c456adebd7c.


swift/ql/test/library-tests/dataflow/flowsources/customurlschemes.swift
```diff
@@ -1,26 +1,35 @@
...
 
 struct URL {}
 
 protocol UIApplicationDelegate {
-    optional func application(_ app: UIApplication, open url: URL, options: [UIApplication.OpenURLOptionsKey : Any]) -> Bool
-    optional func application(_ application: UIApplication, handleOpen url: URL) -> Bool
-    optional func application(_ application: UIApplication, open url: URL, sourceApplication: String?, annotation: Any) -> Bool
+    func application(_ app: UIApplication, open url: URL, options: [UIApplication.OpenURLOptionsKey : Any]) -> Bool
+    func application(_ application: UIApplication, handleOpen url: URL) -> Bool
+    func application(_ application: UIApplication, open url: URL, sourceApplication: String?, annotation: Any) -> Bool
```
redsun82, line 17: `optional` is only allowed for `@objc` protocols, which in turn are only allowed with `-enable-objc-interop`, and in any case only by importing `Foundation` that we do not have available in QL tests...

geoffw0: I don't think it will matter to what we're testing here.

PR #10927, opened 2022-10-21
C#: Include "phi reads" in `DataFlow::Node`

hvitved:
In this PR, we add [SSA phi-read](https://github.com/github/codeql/pull/11198) nodes to the `DataFlow::Node` type, both for C# and CIL.

The benefit of doing this (copied from the other PR) is to be able to avoid the same potential combinatorial explosion that we also avoid by including normal phi nodes in the data flow graph:

```mermaid
  graph TD;
      A-->read_phi;
      B-->read_phi;
      C-->read_phi;
      read_phi-->D;
      read_phi-->E;
      read_phi-->F;
```

Assuming that `A,...,F` read the same underlying variable as `read_phi`, one can reduce the number of data flow edges from 9 (quadratic) to 6 (linear) by including the `read_phi` join-point.

The work of this PR basically amounts to shifting from `Ssa::Definition` to `SsaImpl::DefinitionExt`.

On a local database, making this change I was able to reduce the size of `localFlowStep` from 89,121,074 to 8,349,054 tuples.


csharp/ql/lib/semmle/code/cil/internal/SsaImpl.qll
```diff
@@ -68,9 +68,35 @@ private module Cached {
...
-  predicate lastRefBeforeRedef(Definition def, BasicBlock bb, int i, Definition next) {
-    lastRefRedef(def, bb, i, next)
+  predicate lastRefBeforeRedefExt(DefinitionExt def, BasicBlock bb, int i, DefinitionExt next) {
+    lastRefRedefExt(def, _, bb, i, next)
   }
 }
 
 import Cached
+
+private module Deprecated {
```
michaelnebel, line 78: Is this to avoid deprecation warnings or something like that?

michaelnebel (approved):
Looks plausible to me!

csharp/ql/lib/semmle/code/cil/internal/SsaImpl.qll
hvitved: It was simply to group the deprecated predicates.

PR #10928, opened 2022-10-21
Ruby: assume some global constants are defined

asgerf:
When classes are missing a canonical name, we end up not resolving its outgoing subclass edges. For example, we missed that `Unkn::B` is a subclass of `Unkn::A` in this example:
```rb
class Unkn::A
end

class Unkn::B < Unkn::A
end
```
[Here](https://github.com/discourse/discourse/blob/c1887d95ffd749df86d55acd6630e8ece85614d1/app/controllers/admin/api_controller.rb#L3) is an example of a class that was affected by this issue.

This PR doesn't try to fix the problem in general, but rather aims to pick some low-hanging fruit, enough to unblock my work on migrating some models from AST to DataFlow.

We simply treat a constant `C` as being defined at the top-level if we've seen `C` referenced in a context where it can only refer to a top-level constant.

Interestingly, the API graph does contain a subclassing edge in the above example, as this isn't dependent on the classes having canonical names. API graphs of course has the luxury of not being in the same SCC as constant resolution itself, but it would be nice if we could make the two systems agree on the superclass/ancestor relationship. But that's for another PR.

I noticed some call graph changes due to the `String` class now being considered resolved, and this affects flow through pattern-matching in a few cases. I decided to add `String` to `builtins()` to ensure this is consistently the case, though I'm not sure if it's technically a built-in or just a class in the standard library (does it matter?).

[Evaluation](https://github.com/github/codeql-dca-main/issues/8027) looks good
- About 5000 new call edges, due to more subclassing relationships. Everything I've looked at appear to be true edges.
- Fixes 3 spurious call edges, due to pattern-matching now properly blocking flow.
- The lost taint sink in `opal` seems to be due to the call to `const_get` now resolved to a definition so no longer an `UnknownMethodCall`.

### DIFF
--- a/ruby/ql/lib/codeql/ruby/ast/internal/Module.qll
+++ b/ruby/ql/lib/codeql/ruby/ast/internal/Module.qll
@@ -408,12 +408,8 @@ private module ResolveImpl {
   string getAnAssumedGlobalConst() {
     exists(ConstantAccess access |
       not exists(access.getScopeExpr()) and
-      result = access.getName()
-    |
-      access.hasGlobalScope()
-      or
-      // At the top-level but not inside a block
-      enclosingModuleNoBlock(access) instanceof Toplevel
+      result = access.getName() and
+      isToplevel(access)
     )
   }
 


ruby/ql/lib/codeql/ruby/ast/internal/Module.qll
```diff
@@ -495,23 +524,31 @@ private module ResolveImpl {
 private import ResolveImpl
 
 /**
- * A variant of AstNode::getEnclosingModule that excludes
+ * Gets an enclosing scope of `scope`, stopping at the first module or block.
+ *
+ * Includes `scope` itself and the final module/block.
+ */
+private Scope enclosingScopesNoBlock(Scope scope) {
```
aibaars, line 531: Why is it important to stop at block scopes? I'm not sure this change is right. For example:

```
A = "top"
module M
  A = "M"
end

M.module_eval { puts A }
```
prints `"top"`

asgerf: You're right. I'm not sure what I was thinking there. The block thing is specific to `include` calls and such since they rely on `self`.

Technically there was a bug in old implementation of `enclosingModule` since it relies on `getParent()`, although it would only affect really obscure cases like,
```rb
module (include A)::B
end
```
where the `include` call would be seen as being part of the `B` module, not the top-level. Still, it's a footgun that might reappear later so I think it's best to keep the new formulation.

I've pushed a commit to fix the place where we incorrectly relies on this predicate.

aibaars (approved):

PR #10931, opened 2022-10-21
Ruby: Fix flow steps into phi nodes

hvitved:
- Add missing flow from post-update nodes into phi nodes.
- Prevent flow from reads into phi nodes when use-use flow is prohibited.


asgerf:
LGTM but probably get a second opinion and some tests

nickrolfe:
Agreed, it'd be good to have a test that would have failed without this change.

asgerf (approved):

PR #10933, opened 2022-10-21
C#: Fix flow steps into phi/uncertain def nodes

hvitved:
- Add missing flow from post-update nodes into phi nodes.
- Prevent flow from reads into phi nodes when use-use flow is prohibited.


michaelnebel (approved):
Looks plausible to me!

PR #10934, opened 2022-10-21
ATM: add hello world version of workflow checking queries run

jhelie:


.github/workflows/atm-check-queries-run.yml
```diff
@@ -0,0 +1,13 @@
+name: ATM Check Queries Run
+
+# This check is required, therefore we must run it on all PRs, even if only Markdown has changed.
```
henrymercer, line 3:
```suggestion
```

henrymercer (approved):
Approving to allow testing of this workflow via workflow dispatch on a branch as it is implemented.

.github/workflows/atm-check-queries-run.yml
jhelie: 👍 copy paste, ignore the file content as long as there's a `workflow_dispatch:`

PR #10938, opened 2022-10-21
C++: Fix printf.qll bug

geoffw0:
Fix a bug in `printf.qll` where `%f` and similar format specifiers with a non-default value for precision (e.g. `%.2f`) weren't given a value for `getMaxConvertedLength`.

The original bug was a classic bracketing issue, but was quite subtle:
```
(if this.hasExplicitPrecision(n) then
  afterdot = this.getPrecision(n)
 else
  not this.hasImplicitPrecision(n) and afterdot = 6)
```
it became more obvious a few years ago when the file was autoformatted:
```
(
   if this.hasExplicitPrecision(n)
   then afterdot = this.getPrecision(n)
  else not this.hasImplicitPrecision(n)
) and
afterdot = 6
```
the fix is to put `afterdot = 6` where it should have been all along:
```
if this.hasExplicitPrecision(n)
then afterdot = this.getPrecision(n)
else (
  not this.hasImplicitPrecision(n) and
  afterdot = 6
)
```
I've also added some related tests (that previously existed within a private repo) and extended them to cover the problem.


rdmarsh2 (approved):

rdmarsh2:
Looks like there's an internal test that also needs to be updated.

geoffw0:
> Looks like there's an internal test that also needs to be updated.

There is, and there's already a PR to do that. :)

MathiasVP:
I'm very confused about the test change to `ql/cpp/ql/test/library-tests/templates/CPP-223/decls.ql`. Once that's fixed I'd be happy to merge this.

MathiasVP:
@geoffw0 I took the liberty of merging in the latest `main` to make CI happy.

geoffw0:
> I took the liberty of merging in the latest main to make CI happy.

Thanks, I had been meaning to do that.  I was expecting test failures on the old internal copy of the tests I moved, at least until the change to remove that gets through to here.  I've no idea why `ql/cpp/ql/test/library-tests/templates/CPP-223/decls.ql` failed.

jketema:
> I've no idea why `ql/cpp/ql/test/library-tests/templates/CPP-223/decls.ql` failed.

Somehow this was tested with an un-bumped submodule, while the extractor change that caused this test change was used while running the tests. It's not clear how this set of circumstances could occur, but hopefully merging in `main` here fixes this.

PR #10943, opened 2022-10-22
Javascript/Python: Tokens built from predictable UUIDs

bananabr:
A couple of queries based on Daniel Thatcher's work  https://www.intruder.io/research/in-guid-we-trust to look for "tokens" and "codes" generated from predictable UUIDs.


tausbn:
@bananabr Are you planning on applying for a GitHub Security Lab [bounty](https://securitylab.github.com/bounties/) for this PR? 
I wasn't able to find a bounty issue for this PR.

tausbn:
@bananabr Are you planning on applying for a GitHub Security Lab [bounty](https://securitylab.github.com/bounties/) for this PR? 
I wasn't able to find a bounty issue for this PR.

bananabr:
> @bananabr Are you planning on applying for a GitHub Security Lab [bounty](https://securitylab.github.com/bounties/) for this PR? I wasn't able to find a bounty issue for this PR.

Not at this time @tausbn.

python/ql/src/experimental/Security/CWE-340/TokenBuiltFromUUID.ql
```diff
@@ -0,0 +1,69 @@
...
+import semmle.python.ApiGraphs
+import semmle.python.dataflow.new.TaintTracking
+import DataFlow::PathGraph
+
+class PredictableResultSource extends DataFlow::Node {
+  PredictableResultSource() {
+    exists(API::Node uuidCallRet |
+      uuidCallRet =
+        API::moduleImport("uuid")
+            .getMember(["uuid1", "uuid2", "uuid3", "uuid5"])
```
tausbn, line 25:
```suggestion
            .getMember(["uuid1", "uuid3", "uuid5"])
```
As far as I can tell, there is no `uuid.uuid2` in the `uuid` library.

python/ql/src/experimental/Security/CWE-340/TokenBuiltFromUUID.ql
```diff
@@ -0,0 +1,69 @@
...
+            .getACall()
+            .getReturn()
```
tausbn, line 27: You should be able to replace this with just `getReturn`.

tausbn, line 27:
```suggestion
            .getReturn()
```
You should be able to replace this with just `getReturn`.

python/ql/src/experimental/Security/CWE-340/TokenBuiltFromUUID.ql
```diff
@@ -0,0 +1,69 @@
...
+
+class TokenBuiltFromUuidConfig extends TaintTracking::Configuration {
+  TokenBuiltFromUuidConfig() { this = "TokenBuiltFromUuidConfig" }
+
+  override predicate isSource(DataFlow::Node source) { source instanceof PredictableResultSource }
+
+  override predicate isSink(DataFlow::Node sink) { sink instanceof TokenAssignmentValueSink }
+
+  override predicate isAdditionalTaintStep(DataFlow::Node nodeFrom, DataFlow::Node nodeTo) {
+    exists(Call call, Name name |
```
tausbn, line 58: Here, I would recommend using `DataFlow::CallCfgNode` instead, to avoid dropping down to the AST layer for specifying the function.

That way, you can also use `API::builtin("str").getACall()` to select the appropriate functions which is slightly more general, and saves a bit of typing. 🙂

python/ql/src/experimental/Security/CWE-340/TokenBuiltFromUUID.ql
```diff
@@ -0,0 +1,69 @@
...
+      this = uuidCallRet.asSource()
+      or
+      this = uuidCallRet.getMember(["hex", "bytes", "bytes_le"]).asSource()
+    )
+  }
+}
+
+class TokenAssignmentValueSink extends DataFlow::Node {
+  TokenAssignmentValueSink() {
+    exists(Assign a, Expr target | this = DataFlow::exprNode(a.getValue()) |
```
tausbn, line 38: For more generality, it would be better to use `DefinitionNode` here, and handle things at the control-flow layer. (This would make the code cover not only simple assignments, but also things like `:=` and `as`, which serve the same function, but are different at the level of syntax.)

Also, it would be better to use `AttrWrite` rather than `Attribute`. The former is defined at the data-flow layer, and again covers more constructs that all write to attributes.

tausbn (changes requested):
Thank you for your submission!

Overall, the Python bits look really solid to me. I have made a bunch of suggestions for improvements, mostly to help you learn how to use the Python QL libraries even more effectively. 

Let me know if I need to elaborate on anything. 🙂

python/ql/src/experimental/Security/CWE-340/TokenBuiltFromUUID.ql
bananabr: Thanks for the suggestion. This was my first try with the python CodeQL API, so there was some trial and error involved. =)

python/ql/src/experimental/Security/CWE-340/TokenBuiltFromUUID.ql
bananabr: Could you elaborate on this a bit more? I understand the concepts of what you said, and it makes totally sense. However, I am not familiar with the CodeQL python API enough to re-write this according to your suggestions.

Thank you!

bananabr: Does this reflect your suggestions?

```codeql
class TokenAssignmentValueSink extends DataFlow::Node {
  TokenAssignmentValueSink() {
    exists(DefinitionNode dn, DataFlow::AttrWrite aw | this = aw.getValue() |
      aw.getValue().asCfgNode() = dn.getValue() and
      aw.getAttributeName().toLowerCase().matches(["%token", "%code"])
    )
  }
}
```

bananabr: Uow! Thanks for the class! I was always bothered by the string repetition but never thought about using a string variable the way you did.  Moreover, I was missing the cast to NameNode idea to make your suggestion. I went with the one with the outer and inner exists cause I do agree readability is better.

python/ql/src/experimental/Security/CWE-340/TokenBuiltFromUUID.ql
bananabr: Done. Thanks for your suggestion. I learned a lot with this PR =)

javascript/ql/src/experimental/Security/CWE-340/TokenBuiltFromUUID.ql
```diff
@@ -0,0 +1,51 @@
...
+
+import javascript
+import DataFlow
+import DataFlow::PathGraph
+
+class PredictableResultSource extends DataFlow::Node {
+  PredictableResultSource() {
+    exists(API::Node uuidCallRet |
+      uuidCallRet =
+        API::moduleImport("uuid").getMember(["v1", "v2", "v3", "v5"]).getACall().getReturn()
```
erik-krogh, line 22:
```suggestion
        API::moduleImport("uuid").getMember(["v1", "v3", "v5"]).getACall().getReturn()
```
Same fix as the Python query.  

Otherwise this looks fine for an experimental query.

tausbn: This change caused the autoformatter to become unhappy (probably because the line is now shorter, and so it can format things in a different way). 

@bananabr Could you reformat this file and push the formatted version? 🙂 
(I think that's all that's needed in order for us to merge it.)

bananabr: Done. I also took the chance to remove the unnecessary getACAll call from the JS query.

erik-krogh (approved):
JS parts look OK for an experimental query.

tausbn (approved):
Thanks again!

PR #10947, opened 2022-10-23
Swift: detect hash functions with low # of iterations

karimhamdanali:
Using hash functions with less than 1,000 iterations is not secure. That scheme is vulnerable to password cracking attacks due to having an insufficient level of computational effort.

The rule currently supports all ciphers that the CryptoSwift API provides, but we can always extend it further if more cophers are added.

I'd appreciate a review of the query itself, the accompanying tests, and the associated documentation.

### DIFF
--- a/swift/ql/src/queries/Security/CWE-916/InsufficientHashIterations.qhelp
+++ b/swift/ql/src/queries/Security/CWE-916/InsufficientHashIterations.qhelp
@@ -7,11 +7,11 @@
   </overview>
 
   <recommendation>
-    <p>Use sufficient number of iterations (that is, greater than or equal 1000) for generating password-based keys.</p>
+    <p>Use sufficient number of iterations (that is, greater than or equal 120000) for generating password-based keys.</p>
   </recommendation>
 
   <example>
-    <p>The following example shows a few cases of instantiating a password-based key. In the 'BAD' cases, the key is initialized with insufficient iterations, making it susceptible to password cracking attacks. In the 'GOOD' cases, the key is initialized with at least 1000 iterations, which protects the encrypted data against recovery.</p>
+    <p>The following example shows a few cases of instantiating a password-based key. In the 'BAD' cases, the key is initialized with insufficient iterations, making it susceptible to password cracking attacks. In the 'GOOD' cases, the key is initialized with at least 120000 iterations, which protects the encrypted data against recovery.</p>
     <sample src="InsufficientHashIterations.swift" />
   </example>
 
--- a/swift/ql/src/queries/Security/CWE-916/InsufficientHashIterations.qhelp
+++ b/swift/ql/src/queries/Security/CWE-916/InsufficientHashIterations.qhelp
@@ -7,11 +7,16 @@
   </overview>
 
   <recommendation>
-    <p>Use sufficient number of iterations (i.e., greater than or equal 1000) for generating password-based keys.</p>
+    <p>Use sufficient number of iterations (that is, greater than or equal 1000) for generating password-based keys.</p>
   </recommendation>
 
   <example>
     <p>The following example shows a few cases of instantiating a password-based key. In the 'BAD' cases, the key is initialized with insufficient iterations, making it susceptible to password cracking attacks. In the 'GOOD' cases, the key is initialized with at least 1000 iterations, which protects the encrypted data against recovery.</p>
     <sample src="InsufficientHashIterations.swift" />
   </example>
+
+  <references>
+    <li>Password-Based Cryptography Specification Version 2.0. 2000.<a href="https://www.rfc-editor.org/rfc/rfc2898">RFC2898</a>.</li>
+    <li>OWASP <a href="https://cheatsheetseries.owasp.org/cheatsheets/Password_Storage_Cheat_Sheet.html">Password Storage Cheat Sheet.</a></li>
+  </references>
 </qhelp>
--- a/swift/ql/src/queries/Security/CWE-916/InsufficientHashIterations.qhelp
+++ b/swift/ql/src/queries/Security/CWE-916/InsufficientHashIterations.qhelp
@@ -3,15 +3,16 @@
   "qhelp.dtd">
 <qhelp>
   <overview>
-    <p>Using hash functions with less than 1,000 iterations is not secure. That scheme is vulnerable to password cracking attacks due to having an insufficient level of computational effort.</p>
+    <p>Storing cryptographic hashes of passwords is standard security practice, but it is equally important to select the right hashing scheme. If an attacker obtains the hashed passwords of an application, the password hashing scheme should still prevent the attacker from easily obtaining the original cleartext passwords.</p>
+    <p>A good password hashing scheme requires a computation that cannot be done efficiently. Hashing schemes with low number of iterations are efficiently computable, and are therefore not suitable for password hashing.</p>
   </overview>
 
   <recommendation>
-    <p>Use sufficient number of iterations (that is, greater than or equal 120000) for generating password-based keys.</p>
+    <p>Use the OWASP recommendation for sufficient number of iterations (currently, that is greater than or equal to 120,000) for password hashing schemes.</p>
   </recommendation>
 
   <example>
-    <p>The following example shows a few cases of instantiating a password-based key. In the 'BAD' cases, the key is initialized with insufficient iterations, making it susceptible to password cracking attacks. In the 'GOOD' cases, the key is initialized with at least 120000 iterations, which protects the encrypted data against recovery.</p>
+    <p>The following example shows a few cases where a password hashing scheme is instantiated. In the 'BAD' cases, the scheme is initialized with insufficient iterations, making it susceptible to password cracking attacks. In the 'GOOD' cases, the scheme is initialized with at least 120,000 iterations, which protects the hashed data against recovery.</p>
     <sample src="InsufficientHashIterations.swift" />
   </example>
 
--- a/swift/ql/src/queries/Security/CWE-916/InsufficientHashIterations.qhelp
+++ b/swift/ql/src/queries/Security/CWE-916/InsufficientHashIterations.qhelp
@@ -1,0 +1,17 @@
+<!DOCTYPE qhelp PUBLIC
+  "-//Semmle//qhelp//EN"
+  "qhelp.dtd">
+<qhelp>
+  <overview>
+    <p>Using hash functions with less than 1,000 iterations is not secure. That scheme is vulnerable to password cracking attacks due to having an insufficient level of computational effort.</p>
+  </overview>
+
+  <recommendation>
+    <p>Use sufficient number of iterations (i.e., greater than or equal 1000) for generating password-based keys.</p>
+  </recommendation>
+
+  <example>
+    <p>The following example shows a few cases of instantiating a password-based key. In the 'BAD' cases, the key is initialized with insufficient iterations, making it susceptible to password cracking attacks. In the 'GOOD' cases, the key is initialized with at least 1000 iterations, which protects the encrypted data against recovery.</p>
+    <sample src="InsufficientHashIterations.swift" />
+  </example>
+</qhelp>
--- a/swift/ql/src/queries/Security/CWE-916/InsufficientHashIterations.ql
+++ b/swift/ql/src/queries/Security/CWE-916/InsufficientHashIterations.ql
@@ -1,0 +1,76 @@
+/**
+ * @name Insufficient hash iterations
+ * @description Using hash functions with < 1000 iterations is not secure, because that scheme leads to password cracking attacks due to having an insufficient level of computational effort.
+ * @kind path-problem
+ * @problem.severity error
+ * @security-severity 7.8
+ * @precision high
+ * @id swift/insufficient-hash-iterations
+ * @tags security
+ *       external/cwe/cwe-916
+ */
+
+import swift
+import codeql.swift.dataflow.DataFlow
+import codeql.swift.dataflow.TaintTracking
+import DataFlow::PathGraph
+
+/**
+ * An `Expr` that is used to initialize a password-based encryption key.
+ */
+abstract class IterationsSource extends Expr { }
+
+/**
+ * A literal integer that is 1000 or less is a source of taint for iterations.
+ */
+class IntLiteralSource extends IterationsSource instanceof IntegerLiteralExpr {
+  IntLiteralSource() { this.getStringValue().toInt() >= 1000 }
+}
+
+/**
+ * A class for all ways to set the iterations of hash function.
+ */
+class InsufficientHashIterationsSink extends Expr {
+  InsufficientHashIterationsSink() {
+    // `iterations` arg in `init` is a sink
+    exists(ClassOrStructDecl c, AbstractFunctionDecl f, CallExpr call |
+      c.getFullName() = "PKCS5.PBKDF1" and
+      c.getAMember() = f and
+      f.getName().matches("init(%iterations:%") and
+      call.getStaticTarget() = f and
+      call.getArgument(2).getExpr() = this
+    )
+    or
+    exists(ClassOrStructDecl c, AbstractFunctionDecl f, CallExpr call |
+      c.getFullName() = "PKCS5.PBKDF2" and
+      c.getAMember() = f and
+      f.getName().matches("init(%iterations:%") and
+      call.getStaticTarget() = f and
+      call.getArgument(3).getExpr() = this
+    )
+  }
+}
+
+/**
+ * A dataflow configuration from the hash iterations source to expressions that use
+ * it to initialize hash functions.
+ */
+class InsufficientHashIterationsConfig extends TaintTracking::Configuration {
+  InsufficientHashIterationsConfig() { this = "InsufficientHashIterationsConfig" }
+
+  override predicate isSource(DataFlow::Node node) { node.asExpr() instanceof IterationsSource }
+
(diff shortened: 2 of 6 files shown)


martin389:
👋 Hi from docs, I've added this PR to the team board for a writer to review.

felicitymay:
Hi @karimhamdanali - thanks for asking for a docs team review of your pull request. 👋🏻 

Martin's followed our first responder notes and added this to our review board. Normally we expect a new query to have a technical review either complete or nearly complete before we start a docs review. This avoids duplicated effort because a technical review may change the focus of a query, so the user-facing content will change as a result of the review.

I'm going to remove the label and remove this pull request from our review board for now. Please do re-add the label when your confident that the technical review is nearly complete.

karimhamdanali:
Thanks @felicitymay for the clarification. I actually didn't know that! Will re-add the label once the PR is through the technical review.

swift/ql/src/queries/Security/CWE-916/InsufficientHashIterations.qhelp
```diff
@@ -0,0 +1,17 @@
+<!DOCTYPE qhelp PUBLIC
+  "-//Semmle//qhelp//EN"
+  "qhelp.dtd">
+<qhelp>
+  <overview>
+    <p>Using hash functions with less than 1,000 iterations is not secure. That scheme is vulnerable to password cracking attacks due to having an insufficient level of computational effort.</p>
+  </overview>
+
+  <recommendation>
+    <p>Use sufficient number of iterations (i.e., greater than or equal 1000) for generating password-based keys.</p>
```
geoffw0, line 10:
```suggestion
    <p>Use sufficient number of iterations (that is, greater than or equal 1000) for generating password-based keys.</p>
```
I believe "that is," is preferred over "i.e." (for readers with less experience in English).

Out of interest, I understand this comes from previous research but I'm interested to hear why the threshold we chose is 1,000, rather than say 100 or 10,000?  Is this recommendation reasonably future proof or is it something we might need to increase in future?

swift/ql/src/queries/Security/CWE-916/InsufficientHashIterations.qhelp
```diff
@@ -0,0 +1,17 @@
...
+  </recommendation>
+
+  <example>
+    <p>The following example shows a few cases of instantiating a password-based key. In the 'BAD' cases, the key is initialized with insufficient iterations, making it susceptible to password cracking attacks. In the 'GOOD' cases, the key is initialized with at least 1000 iterations, which protects the encrypted data against recovery.</p>
+    <sample src="InsufficientHashIterations.swift" />
+  </example>
+</qhelp>
```
geoffw0, line 17: I feel that this `.qhelp` badly needs a `<references>` section, with a link to something that explains why deliberately increasing the compute cost of these functions through iterations is considered a good use of resources (my understanding of the issue here).

swift/ql/src/queries/Security/CWE-916/InsufficientHashIterations.ql
```diff
@@ -0,0 +1,69 @@
...
+/**
+ * A class for all ways to set the iterations of hash function.
+ */
+class InsufficientHashIterationsSink extends Expr {
+  InsufficientHashIterationsSink() {
+    // `iterations` arg in `init` is a sink
+    exists(ClassOrStructDecl c, AbstractFunctionDecl f, CallExpr call, int arg |
+      c.getFullName() = ["PBKDF1", "PBKDF2"] and
+      c.getAMember() = f and
+      f.getName().matches("init(%iterations:%") and
```
geoffw0, line 39: This is slightly different from how we've usually matched function names before (listing all the variations explicitly).  I think I'm happy with it, it will be more robust if new versions of `init` with slightly different parameters are added in future.

swift/ql/src/queries/Security/CWE-916/InsufficientHashIterations.ql
```diff
@@ -0,0 +1,69 @@
...
+  }
+}
+
+// The query itself
+from
+  InsufficientHashIterationsConfig config, DataFlow::PathNode sourceNode,
+  DataFlow::PathNode sinkNode
+where config.hasFlowPath(sourceNode, sinkNode)
+select sinkNode.getNode(), sourceNode, sinkNode,
+  "The variable '" + sinkNode.getNode().toString() +
```
geoffw0, line 67: I think I would rephrase this as the number of iterations wasn't necessarily stored in a variable.  We could pull the number itself out from the source using `.getStringValue()` and describe it as "The value 123" (or whatever value it was).  Or we could simplify the message to something static like:
```
This value is an insufficient number of iterations for secure password hashing."
```

swift/ql/src/queries/Security/CWE-916/InsufficientHashIterations.ql
```diff
@@ -0,0 +1,69 @@
...
+    "' is an insufficient number of iterations, which is not secure for hash functions.",
```
geoffw0, line 68:
```suggestion
    "' is an insufficient number of iterations, which is not secure for password hash functions.",
```
Many users will only read the query message despite our efforts with metadata and `.qhelp` and so forth - so I think its worth being as clear as possible here (while also being concise).

geoffw0 (commented):
Looks great, a few minor suggestions...

swift/ql/src/queries/Security/CWE-916/InsufficientHashIterations.qhelp
karimhamdanali: Good question, the initial standard is 1000, but that number should increase with more CPU power (https://en.wikipedia.org/wiki/PBKDF2). The latest recommendation from OWASP is actually at least 120,000 iterations (https://cheatsheetseries.owasp.org/cheatsheets/Password_Storage_Cheat_Sheet.html). Given that this number should increase over time, how do we best handle this?

swift/ql/src/queries/Security/CWE-916/InsufficientHashIterations.qhelp
karimhamdanali: You're right! I forgot to add them to the initial PR. Added them now.

swift/ql/src/queries/Security/CWE-916/InsufficientHashIterations.ql
karimhamdanali: sgtm!

swift/ql/src/queries/Security/CWE-916/InsufficientHashIterations.ql
karimhamdanali: I like your former suggestion, so I've changed the output message to something like 
`The value '999' is an insufficient number of iterations for secure password hashing.`

swift/ql/src/queries/Security/CWE-916/InsufficientHashIterations.ql
karimhamdanali: I addressed that comment in my my latest commit to this PR.

swift/ql/src/queries/Security/CWE-916/InsufficientHashIterations.qhelp
geoffw0: If we want to increase it over time, I guess we should create some kind of issue to track this and preferably attach it to something (like an iteration or quarter) far in the future.

But it sounds like there's already a big gap between the current threshold of 1,000 and the OWASP recommendation of 120,000 iterations.  I guess we don't want to be too fussy - telling someone that 100,000 iterations isn't enough when the recommendation is 120,000 seems picky - but we do want to call out when code is risky e.g. because it follows advice from 20 years ago.  Is there any reason we shouldn't increase our threshold to somewhere in the 10,000 - 100,000 range?

swift/ql/src/queries/Security/CWE-916/InsufficientHashIterations.ql
geoffw0: LGTM.

swift/ql/src/queries/Security/CWE-916/InsufficientHashIterations.ql
geoffw0: Yep, thanks!

swift/ql/src/queries/Security/CWE-916/InsufficientHashIterations.qhelp
geoffw0: ... in fact reading the OWASP page, 120,000 appears to be the lowest of the recommendations they make.  So perhaps we _should_ set the threshold to 120,000???

karimhamdanali:
Commit is now ready for docs review.

swift/ql/src/queries/Security/CWE-916/InsufficientHashIterations.qhelp
karimhamdanali: I'd say 120k should be the lowest threshold that we should generate warnings for. I'll update the queries and the associated documentation, examples, etc.

geoffw0 (commented):
I'm happy for this to be merged after the docs review. :+1:

martin389:
Hi again from docs -- Adding this PR to the team board for a writer to review ✨

swift/ql/src/queries/Security/CWE-916/InsufficientHashIterations.qhelp
```diff
@@ -0,0 +1,22 @@
+<!DOCTYPE qhelp PUBLIC
+  "-//Semmle//qhelp//EN"
+  "qhelp.dtd">
+<qhelp>
+  <overview>
+    <p>Using hash functions with less than 1,000 iterations is not secure. That scheme is vulnerable to password cracking attacks due to having an insufficient level of computational effort.</p>
```
felicitymay, line 6: This query seems as if it's fairly closely related to the JavaScript query [Use of password hash with insufficient computational effort](https://codeql.github.com/codeql-query-help/javascript/js-insufficient-password-hash/).

I think that it might be worth reusing some of the explanation of why this is important.

swift/ql/src/queries/Security/CWE-916/InsufficientHashIterations.qhelp
```diff
@@ -0,0 +1,22 @@
...
+  </overview>
+
+  <recommendation>
+    <p>Use sufficient number of iterations (that is, greater than or equal 120000) for generating password-based keys.</p>
```
felicitymay, line 10: We don't say how we arrived at this number. Is it taken from an OWASP recommendation, or somewhere else that we could refer to? I'm assuming that recommended minimums will gradually rise, so referring to a source where the current best practice is detailed will help this information stay up to date.

swift/ql/src/queries/Security/CWE-916/InsufficientHashIterations.qhelp
```diff
@@ -0,0 +1,22 @@
...
+  </recommendation>
+
+  <example>
+    <p>The following example shows a few cases of instantiating a password-based key. In the 'BAD' cases, the key is initialized with insufficient iterations, making it susceptible to password cracking attacks. In the 'GOOD' cases, the key is initialized with at least 120000 iterations, which protects the encrypted data against recovery.</p>
```
felicitymay, line 14: Minor text suggestion:

```suggestion
    <p>The following example shows a few cases where a password-based key is instantiated. In the 'BAD' cases, the key is initialized with insufficient iterations, making it susceptible to password cracking attacks. In the 'GOOD' cases, the key is initialized with at least 120,000 iterations, which protects the encrypted data against recovery.</p>
```

swift/ql/src/queries/Security/CWE-916/InsufficientHashIterations.swift
```diff
@@ -0,0 +1,14 @@
+
+func encrypt() {
+	// ...
+
+	// BAD: Using insufficient (i.e., < 120,000) hash iterations keys for encryption
```
felicitymay, line 5: We write out "i.e." in full English for clarity:

```suggestion
	// BAD: Using insufficient (that is, < 120,000) hash iterations keys for encryption
```

swift/ql/src/queries/Security/CWE-916/InsufficientHashIterations.swift
```diff
@@ -0,0 +1,14 @@
...
+	_ = try PKCS5.PBKDF1(password: getRandomArray(), salt: getRandomArray(), iterations: 90000, keyLength: 0)
+	_ = try PKCS5.PBKDF2(password: getRandomArray(), salt: getRandomArray(), iterations: 90000, keyLength: 0)
+
+	// GOOD: Using sufficient (i.e., >= 120,000) hash iterations keys for encryption
```
felicitymay, line 9:
```suggestion
	// GOOD: Using sufficient (that is, >= 120,000) hash iterations keys for encryption
```

swift/ql/src/queries/Security/CWE-916/InsufficientHashIterations.swift
```diff
@@ -0,0 +1,14 @@
...
+	_ = try PKCS5.PBKDF1(password: getRandomArray(), salt: getRandomArray(), iterations: 120120, keyLength: 0)
+	_ = try PKCS5.PBKDF2(password: getRandomArray(), salt: getRandomArray(), iterations: 120120, keyLength: 0)
```
felicitymay, line 11: Judging by the OWASP reference, what is an adequate number of iterations seems to vary according to the algorithm you're using. It's not entirely clear to me looking at this example, whether or not this is adequate for the algorithm because I don't see which hashing algorithm is used.

This may be entirely unimportant - I'm a writer, not a developer 🙂

swift/ql/src/queries/Security/CWE-916/InsufficientHashIterations.ql
```diff
@@ -0,0 +1,68 @@
+/**
+ * @name Insufficient hash iterations
+ * @description Using hash functions with less than 120,000 iterations is not secure, because that scheme leads to password cracking attacks due to having an insufficient level of computational effort.
```
felicitymay, line 3: Suggest:

```suggestion
 * @description Using hash functions with fewer than 120,000 iterations is insufficient to protect passwords because a cracking attack will require a low level of computational effort.
```

felicitymay (commented):
Hi @karimhamdanali 👋🏻 

Many thanks for including query help, references and an example for this new query ✨ 

I've added a few suggestions and comments with the intention of making the information clearer for users. I'm hopeful that my suggestions don't change the intended meaning, but am aware that I have little technical context. Very happy to discuss any suggestions or comments.

swift/ql/src/queries/Security/CWE-916/InsufficientHashIterations.swift
geoffw0: I think this is a good point.  We don't _implement_ the different recommended values for different algorithms, rather we just took the lowest recommendation and check the value used is at least that.  In practice this design is likely to catch the vast majority of problems and keeps the query simple and reliable - but there's no reason our _examples_ shouldn't set a good example by being fully OWASP compliant.

swift/ql/src/queries/Security/CWE-916/InsufficientHashIterations.qhelp
karimhamdanali: Agreed. I've updated the help text here with the relevant information from the JavaScript query. Thanks for the pointer to that query @felicitymay !

swift/ql/src/queries/Security/CWE-916/InsufficientHashIterations.qhelp
karimhamdanali: Agreed. I've updated the text to state that we should follow the OWASP recommendation, with a note to the current recommendation of at least 120,000 iterations.

swift/ql/src/queries/Security/CWE-916/InsufficientHashIterations.qhelp
karimhamdanali: fixed as suggested

swift/ql/src/queries/Security/CWE-916/InsufficientHashIterations.swift
karimhamdanali: fixed as suggested

swift/ql/src/queries/Security/CWE-916/InsufficientHashIterations.swift
karimhamdanali: fixed as suggested

swift/ql/src/queries/Security/CWE-916/InsufficientHashIterations.swift
karimhamdanali: Agreed. I've updated the test to be fully OWASP compliant :)

swift/ql/src/queries/Security/CWE-916/InsufficientHashIterations.ql
karimhamdanali: fixed as suggested

karimhamdanali:
Thanks @felicitymay for the docs review! I think I've addressed all your comments so far. Please let me know if you have any further comments or suggestions before we can merge this PR. Thanks :)

felicitymay (approved):
Many thanks for the updates to the query help 💖

geoffw0 (approved):

karimhamdanali:
Awesome! Thanks @felicitymay. @geoffw0, I think this one is ready to get merged then :) Thanks all for reviewing this PR.

PR #10950, opened 2022-10-24
Swift: extract `RegexLiteralExpr`

redsun82:
This waits on fixes on the swift prebuilt package.


redsun82:
> Do we need a downgrade and upgrade script in this case?

yes, but it should be a standard upgrade/downgrade pair due to added columns, so nothing too fancy.

The other issue, is that current github macOS runners are actually unable to compile regexes (their macOS version is too old...). So probably for the time being the QL test should be turned into a linux-only integration test.

AlexDenisov (approved):

PR #10951, opened 2022-10-24
Kotlin: Fix location (start position) of method calls

tamasvajk:


java/kotlin-extractor/src/main/kotlin/TrapWriter.kt
```diff
@@ -269,11 +270,27 @@ open class FileTrapWriter (
      */
     val fileId = mkFileId(filePath, populateFileTables)
 
+    private fun getStartOffset(e: IrElement): Int {
+        return when (e) {
+            is IrCall -> {
+                // Calls have incorrect startOffset, so we adjust them:
+                val dr = e.dispatchReceiver?.let { getStartOffset(it) } ?: run { Int.MAX_VALUE }
+                val er = e.extensionReceiver?.let { getStartOffset(it) } ?: run { Int.MAX_VALUE }
+                minOf(e.startOffset, dr, er)
```
igfoo, line 279: I think that this could be something like
```
val dr = e.dispatchReceiver?.let { getStartOffset(it) }
val er = e.extensionReceiver?.let { getStartOffset(it) }
dr ?: er ?: e.startOffset
```
which is sadly not as symmetric, but I assume you can't have both receivers having non-null start offsets.

igfoo: Hmm, it probably ought to actually be something equivalent to:
```
val s = e.startOffset
val dr = e.dispatchReceiver?.let { getStartOffset(it) }
val er = e.extensionReceiver?.let { getStartOffset(it) }
if (s == UNDEFINED_OFFSET || s == SYNTHETIC_OFFSET) {
    s
} else if (dr != null && dr != UNDEFINED_OFFSET && dr != SYNTHETIC_OFFSET) {
    dr
} else if (er != null && er != UNDEFINED_OFFSET && er != SYNTHETIC_OFFSET) {
    er
} else {
   s
}
```

tamasvajk: > but I assume you can't have both receivers having non-null start offsets.

Unfortunately this is not the case:
```kotlin
class C {
    val x = 5
    fun Int.fn0(): Int {
        return this@C.x + this
    }

    fun fn1() {
        5.fn0() // <--- Both dispatchReceiver and extensionReceiver are non-null, and their start offset is >= 0
    }
}
```

But you're right, that it could change in the future, and kotlinc could decide to return -1 as the start offset for the dispatch receiver in the above case, and then the original implementation would choose that as the minimum value.

At the same time, I think we're better off with a `minOf` call than hardcoding the priority order. Let me adjust the implementation...

tamasvajk: How about https://github.com/github/codeql/pull/10951/commits/aa80f9358400d4db770cad036946a3c540a5be59?

igfoo (dismissed):
`offsetMinOf` might be a better name, but LGTM

igfoo (approved):

igfoo (dismissed):

tamasvajk:
I rebased this PR again to fix some merge conflicts.

smowton (approved):
Proxying Ian's previous review without re-reviewing

igfoo (dismissed):

smowton (approved):

PR #10952, opened 2022-10-24
Kotlin: extract interface redeclarations of `Object` methods

smowton:
Due to a probable compiler bug (?) the redeclaration looks like a fake symbol, leading to Java dispatching against a declaration that Kotlin doesn't believe exists.


tamasvajk (approved):
Looks plausible.

PR #10953, opened 2022-10-24
Bazel/CMake: make cmake runnable from outside the workspace

redsun82:
Also added a small `README.md` file.

### DIFF
--- a/misc/bazel/cmake/README.md
+++ b/misc/bazel/cmake/README.md
@@ -1,0 +1,12 @@
+This directory contains a Bazel to CMake generator intended mainly for IDE integration.
+
+[`cmake.bzl`](./cmake.bzl) contains the Bazel side, with the `cmake_aspect` rule gathering the necessary information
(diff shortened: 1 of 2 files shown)


AlexDenisov (approved):

misc/bazel/cmake/README.md
```diff
@@ -0,0 +1,12 @@
+This directory contains a Bazel to CMake generator intended mainly for IDE integration.
+
+[`cmake.bzl`](./cmake.bzl) contains the Bazel side, with the `cmake_aspect` rule gathering the necessary information
+from `cc_*` rules and `generate_cmake` translating that information into CMake commands. `generate_cmake` targets also
+depend on all files that are either generated or fetched from external repositories, so that Bazel will fill in those
+dependencies before letting CMake do the C/C++ compilation.
```
jketema, line 6: Do you mean:
```suggestion
depend on all files that are either generated or fetched from external repositories, Bazel will generate and fetch those
dependencies before letting CMake do the C/C++ compilation.
```

PR #10956, opened 2022-10-24
Swift: disambuigate entities using linkage awareness on modules

redsun82:
This PR introduces Swift module disambuigation via linkage awareness, and uses this to disambuigate named entities, by stamping their trap key with the module key.

Commit by commit review is strongly encouraged.

The core of the PR is https://github.com/github/codeql/pull/10956/commits/f383fd1dc111974d2f707ce99851ad8b19639477, where linkage awareness related files are emitted:
* for each swift module created by the build, a special trap file is created with a specific implementation id assigned to the trap label, together with a `.link` and `.odep` file, the latter listing both sources and encountered modules as dependencies
* for each source we note the generated traps in a related `.odep` file
* a step to be careful about is a `-merge-modules` invocation, as this will be used to merge previously created `.swiftmodule` files into a new one. The delicate part, is that when we create a module we don't know whether it will be later merged or not, and this changes what implementation ID we must use. We face this issue by replacing the previously emitted trap files with module implemenation IDs of the merged modules with the trap file of the merge target, effectively rewriting the implementation IDs of the merged modules
* in order to reference a module throughout this, we use the hash of the original module

This machinery uses data collected during the run. This data is being stored into a new, mutable `SwiftExtractorState` structure. `SwiftExtractorConfiguration` was made a const field of it, as there were many functions that required both and passing just one made for more concise signatures. The four `collect` commits introduce this mechanism, but the collected data is only read in the main commit.

Consider that this does not yet solve the module and symbol ambiguity of package manifests, as that compilation does not involve a creation of a swiftmodule. That is left for future work.


swift/extractor/infra/TargetDomains.cpp
```diff
@@ -42,4 +46,33 @@ std::optional<TrapDomain> createTargetTrapDomain(SwiftExtractorState& state,
...
+    return std::nullopt;
+  }
+  auto file = getRelativeTrapPath(target, TrapType::linkage, ".odep");
+  auto ret =
+      TargetFile::create(file, state.configuration.trapDir, state.configuration.getTempTrapDir());
+  if (ret) {
+    return ObjectDomain{*std::move(ret)};
+  }
+  return std::nullopt;
+}
```
AlexDenisov, line 82: Nit: it seems these two methods can be squashed into one with some (template) arguments?

swift/extractor/invocation/SwiftInvocationExtractor.cpp
```diff
@@ -1,7 +1,166 @@
...
+                                        const std::string& name,
+                                        const fs::path& mergeTarget,
+                                        const std::string& mergedPartHash) {
+  std::error_code ec;
+  auto mergedPartTarget = getModuleTarget(name, mergedPartHash);
+  fs::copy(getTrapPath(state, mergeTarget, TrapType::linkage),
+           getTrapPath(state, mergedPartTarget, TrapType::linkage),
+           fs::copy_options::overwrite_existing, ec);
+  if (ec) {
+    std::cerr << "unable to replace merged module trap implementation id (" << ec.message() << ")";
```
AlexDenisov, line 109: Nit: maybe add the name as well? Otherwise it would be hard to find what exactly went wrong.

AlexDenisov (approved):
LGTM as soon as DCA is (relatively) happy 😄

swift/extractor/infra/TargetDomains.cpp
redsun82: I'm surprised at being suggested to put _more_ templates in my code :rofl:

AlexDenisov (approved):

PR #10957, opened 2022-10-24
Ruby: document API graphs

aibaars:


docs/codeql/codeql-language-guides/using-api-graphs-in-ruby.rst
```diff
@@ -0,0 +1,183 @@
...
+
+.. code-block:: ql
+
+    import codeql.ruby.ApiGraphs
+
+    select API::getTopLevelMember("File").getMethod("write").getParameter(1).asSink()
+
+A more complex example is a call to ``File.open`` with a block argument. This function creates a ```File`` instance
+and passes it to the supplied block. In this case the first parameter of the block is the place where an
+externally created value enters the code base, i.e. the ``|file|`` in the example below:
```
felicitymay, line 140: Fix rst syntax and minor changes:

```suggestion
A more complex example is a call to ``File.open`` with a block argument. This function creates a ``File`` instance
and passes it to the supplied block. In this case, we are interested in the first parameter of the block because this is where an
externally created value enters the code base, that is, the ``|file|`` in the Ruby example below:
```

docs/codeql/codeql-language-guides/using-api-graphs-in-ruby.rst
```diff
@@ -0,0 +1,183 @@
...
+
+.. code-block:: ruby
+
+    File.open("/my/file.txt", "w") { |file| file << "Hello world" }
+
+The following snippet finds parameters of blocks of ``File.open`` method calls:
```
felicitymay, line 146:
```suggestion
The following snippet of CodeQL finds parameters of blocks of ``File.open`` method calls:
```

docs/codeql/codeql-language-guides/using-api-graphs-in-ruby.rst
```diff
@@ -0,0 +1,183 @@
...
+
+.. code-block:: ql
+
+    import codeql.ruby.ApiGraphs
+
+    select API::getTopLevelMember("File").getMethod("open").getBlock().getParameter(0).asSource()
+
+The following example is a dataflow query that that uses API graphs to find cases where data that 
+is read flows into a call to ```File.write``.
```
felicitymay, line 155:
```suggestion
is read flows into a call to ``File.write``.
```

docs/codeql/codeql-language-guides/using-api-graphs-in-ruby.rst
```diff
@@ -0,0 +1,183 @@
...
+
+.. code-block:: ql
```
felicitymay, line 157: Is there a reason why we used `File.open` above as an example and then use `File.read` here? Could we use `File.read` above so that the example follows through?

felicitymay (commented):
@aibaars - thanks for writing this new article for people who want to write queries for Ruby analysis. It generally looks clear. Many of my comments are to fix backtick errors, the remainder are suggestions that I hope will make the content easier to read.

docs/codeql/codeql-language-guides/using-api-graphs-in-ruby.rst
```diff
@@ -0,0 +1,183 @@
...
+
+    import codeql.ruby.ApiGraphs
+
+    select API::getTopLevelMember("Net").getMember("HTTP")
+
+Note that the given module name *must not* contain any ```::`` symbols. Thus, something like
+`API::getTopLevelMember("Net::HTTP")`` will not do what you expect. Instead, this should be decomposed
+into an access of the ``HTTP`` member of the API graph node for ``Net``, as in the example above.
```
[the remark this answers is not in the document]
aibaars:
```suggestion
Note that you should specify module names without ``::`` symbols. If you write ``API::getTopLevelMember("Net::HTTP")``, it will not do what you expect. Instead, you need to decompose this name
into an access of the ``HTTP`` member of the API graph node for ``Net``, as shown in the example above.
```

docs/codeql/codeql-language-guides/using-api-graphs-in-ruby.rst
```diff
@@ -0,0 +1,183 @@
...
+Subclasses
+----------
+
+For many libraries, the main mode of usage is to extend one or more library classes. To track this
+in the API graph, you can use the ``getASubclass`` method to get the API graph node corresponding to
+all the immediate subclasses of this node. To find *all* subclasses, use ``*`` or ``+`` to apply the
+method repeatedly, as in ``getASubclass*``.
+
+Note that ``getASubclass`` does not account for any subclassing that takes place in library code
+that has not been extracted. Thus, it may be necessary to account for this in the models you write.
```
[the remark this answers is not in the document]
aibaars: That's much nicer, thanks!

docs/codeql/codeql-language-guides/using-api-graphs-in-ruby.rst
```diff
@@ -0,0 +1,183 @@
+.. _using-api-graphs-in-ruby:
+
+Using API graphs in Ruby
```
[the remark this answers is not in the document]
felicitymay: Thanks - somehow I overlooked that article 🤦🏻‍♀️

docs/codeql/codeql-language-guides/using-api-graphs-in-ruby.rst
aibaars: `File.read` and `File.write` are simpler functions to read/write files. They were mentioned a little earlier in the text. The `File.open` method is more versatile and complex. I used it as an example to illustrate how to refer to  places where data flows into the parameter of a block (the `|file|` in the example). 

The final large example of a dataflow query is meant to show how API graphs and dataflow fit together in a more or less realistic query. I chose to use `File.read` and `File.write` because these are simpler sources/sinks . 

The `|file|` argument in the `File.open` example is not the file contents but a file handle from/to which we can read/write data. We could extend the example query with that, to make it even more realistic, however, it would also get a bit more complex as we need to track calls to `read` and `write` on the `|file|` parameter to identify additional sources and sinks for the query.

docs/codeql/codeql-language-guides/using-api-graphs-in-ruby.rst
```diff
@@ -0,0 +1,183 @@
...
+
+Module and class references
+---------------------------
+
+The most common entry point into the API graph will be the point where a toplevel module or class is
+accessed. For example, you can access the API graph node corresponding to the ``::Regexp`` class
+by using the ``API::getTopLevelMember`` method defined in the ``codeql.ruby.ApiGraphs`` module, as the
+following snippet demonstrates.
+
+.. code-block:: ql
```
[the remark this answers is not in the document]
aibaars: Good point. I think most of the cases are description first, followed by the code snippet. Shall we do it like that consistently, or do you prefer a different format?

aibaars:
@felicitymay Many thanks for your review. Those are great improvements.

docs/codeql/codeql-language-guides/using-api-graphs-in-ruby.rst
felicitymay: Thanks for explaining. I'm guessing that your target audience will be aware of all of the above.

docs/codeql/codeql-language-guides/using-api-graphs-in-ruby.rst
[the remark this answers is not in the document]
felicitymay: Description followed by code snippet works for me. Thanks.

felicitymay (commented):
Thanks for the updates @aibaars ✨

docs/codeql/codeql-language-guides/using-api-graphs-in-ruby.rst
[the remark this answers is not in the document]
aibaars: @felicitymay I think the text is already in description first, followed by a `ql` snippet. I think the confusion with the example you commented on is the use of the phrase `This query` to refer to the example above while show another query right below. That's indeed not great. I changed the text in between the two QL snippets to:

> QL snippet 1
>
> The example above finds references to a top-level class. For nested
modules and classes ....
>
> QL snippet 2

Does that address your concern?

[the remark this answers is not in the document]
felicitymay: Thanks. Your changes make things much clearer 💖

docs/codeql/codeql-language-guides/using-api-graphs-in-ruby.rst
```diff
@@ -0,0 +1,185 @@
...
+
+Using the API graph in dataflow queries
+---------------------------------------
+
+Dataflow queries often search for points where data from external sources enters the code base
+as well as places where data leaves the code base. API graphs provide a convenient way to refer 
+to external API components such as library functions and their inputs and outputs. 
+However, you do not use API graph nodes directly in dataflow queries. 
+
+- API graph nodes model entities that are defined outside your code base.
```
erik-krogh, line 113: Not technically true, but close enough - especially for the common usage of API-graphs - and it aids in understanding.  
(To be clear, I don't think anything should change here).

docs/codeql/codeql-language-guides/using-api-graphs-in-ruby.rst
```diff
@@ -0,0 +1,185 @@
...
+
+The ``asSource()`` method is used to select dataflow nodes where a value from an external source
+enters the current code base. A typical example is the return value of a library function such as
+``File.read(path)``:
+
+.. code-block:: ql
+
+    import codeql.ruby.ApiGraphs
+
+    select API::getTopLevelMember("File").getMethod("read").getParameter(1).asSource()
```
erik-krogh, line 127:
```suggestion
    select API::getTopLevelMember("File").getMethod("read").getReturn().asSource()
```
Looks like you copy-pasted a bit much from the below example.

aibaars: Thank you, well spotted!

erik-krogh (approved):

PR #10962, opened 2022-10-25
Kotlin: Exclude .kt files from `java/unreachable-catch-clause`

tamasvajk:


smowton (commented):
Can you say a bit about why we're excluding Kotlin?

tamasvajk:
> Can you say a bit about why we're excluding Kotlin?

There's some detail in the linked internal issue. The main reason is that we're not extracting the `exceptions` relation in the Kotlin extractor, and this check is heavily based on that.

smowton (approved):

PR #10964, opened 2022-10-25
qlpacks: libraryPathDependencies -> dependencies

smowton:
This is an initial stab at just removing all references to the deprecated key. I note though that a bunch of these files are named `legacy-xyz` -- I don't know what these are doing or if they have a reason to use the deprecated key (compatibility with an old `codeql`?)

### DIFF
--- a/csharp/ql/integration-tests/qlpack.yml
+++ b/csharp/ql/integration-tests/qlpack.yml
@@ -1,2 +1,2 @@
-libraryPathDependencies:
-  - codeql-csharp
+dependencies:
+  codeql/csharp-all: '*'
--- a/go/ql/config/legacy-support/qlpack.yml
+++ b/go/ql/config/legacy-support/qlpack.yml
@@ -1,4 +1,4 @@
 name: legacy-libraries-go
 version: 0.0.0
-dependencies:
-  codeql/go-all: '*'
+# Note libraryPathDependencies is obsolete and should not be used in new qlpacks.
+libraryPathDependencies: codeql-go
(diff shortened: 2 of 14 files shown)


RasmusWL (approved):
awesome, thanks :muscle:

erik-krogh:
QL-for-QL fails to build because the version of the shared `typos` pack was bumped, but the version specified in `ql/ql/src/qlpack.yml` wasn't bumped.   
We should somehow ensure that those are kept in sync, but for now you can manually bump the version.

aeisenberg (commented):
Certainly all new packs and tests should be using the new syntax. 

I'm hesitant to change any of the `legacy-support` packs (until I learn more about what they are). I would also like to retain at least one test for legacy syntax.

dbartol (changes requested):
Please revert all of the changes to the `legacy-support` packs as discussed in Slack.

Also, I'm assuming that the rest of these dependencies are all to packs in the same repo? If so, then all of the `'*'` version ranges should be replaced with `${workspace}`, which means "whatever version is currently in the same repo". It's not particularly important for query packs, but for library packs that we might eventually publish, it ensures that other consumers of that library pack actually get a compatible version of that dependency.

smowton:
Legacy packs reverted; left `*` alone for now since this is also in use all over the repo, so I'd suggest following up with a different PR rather than broaden this one's scope

dbartol (approved):
LGTM

aeisenberg (approved):
Nice.

PR #10967, opened 2022-10-25
Swift: Fix flow out of summarized callables

MathiasVP:
We had forgotten to generalize the code in a couple of places when we added flow through MaD summaries. This PR fixes these issues.

cc @atorralba this PR should make https://github.com/github/codeql/pull/10954 unnecessary.

### DIFF
--- a/swift/ql/lib/codeql/swift/dataflow/internal/DataFlowDispatch.qll
+++ b/swift/ql/lib/codeql/swift/dataflow/internal/DataFlowDispatch.qll
@@ -53,17 +53,19 @@ class ParamReturnKind extends ReturnKind, TParamReturnKind {
  * defined in library code.
  */
 class DataFlowCallable extends TDataFlowCallable {
-  CfgScope scope;
-
-  DataFlowCallable() { this = TDataFlowFunc(scope) }
+  /** Gets the location of this callable. */
+  Location getLocation() { none() } // overridden in subclasses
 
   /** Gets a textual representation of this callable. */
-  string toString() { result = scope.toString() }
+  string toString() { none() } // overridden in subclasses
 
-  /** Gets the location of this callable. */
-  Location getLocation() { result = scope.getLocation() }
+  CfgScope asSourceCallable() { this = TDataFlowFunc(result) }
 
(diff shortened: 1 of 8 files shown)


MathiasVP:
Hmmm... For some reason this messes up something with call contexts so that we now get flow in:
```swift
func source() -> String { return "" }
func sink(arg: URL) {}

func taintThroughURL() {
  let clean = "http://example.com/"
  let tainted = source()
  let urlClean = URL(string: clean)!
  let urlTainted = URL(string: tainted)! // <--- For some reason flow out of `init:` jumps to the line above.
  sink(arg: urlClean)
}
```
I'll mark this PR as a draft until I've investigated what's going on here.

MathiasVP:
> Hmmm... For some reason this messes up something with call contexts so that we now get flow in:
> 
> ```swift
> func source() -> String { return "" }
> func sink(arg: URL) {}
> 
> func taintThroughURL() {
>   let clean = "http://example.com/"
>   let tainted = source()
>   let urlClean = URL(string: clean)!
>   let urlTainted = URL(string: tainted)! // <--- For some reason flow out of `init:` jumps to the line above.
>   sink(arg: urlClean)
> }
> ```
> 
> I'll mark this PR as a draft until I've investigated what's going on here.

I've force-pushed the fix for this and everything appears to be working as it should now: We do get the right flow through `signum` now without any call-context conflation in other tests 🎉.

However, it looks like this fix uncovered an extractor issue (I've provided a backlink from that issue). The extractor issue means that we're losing the results from the [SQL injection query](https://github.com/github/codeql/pull/10967/commits/0b0771040ad56a944630e40f05269fb8245c8736#diff-a11068810d4b20e1b94401eb8a00cac96d93b99dc8313225669f67b29b6de6bcL17). I've verified that we get the results back when I change the test to work around the extractor bug (basically: replace all the `UnsafeFoo<...>` types with `String`).

I don't think the extractor bug should block this PR as @atorralba is waiting for this PR for https://github.com/github/codeql/pull/11027.

atorralba (approved):
Looks plausible to me! Thanks for getting this fixed.

PR #10969, opened 2022-10-25
CI: Add Internal CI Checks workflow

fossjunkie:
This PR adds a workflow that checks if the author of a PR is a member of CodeQL team and adds `ready-for-internal-ci` label to trigger internal CI checks automatically.


.github/workflows/internal-ci-checks.yml
```diff
@@ -0,0 +1,28 @@
...
+jobs:
+  set-label:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Set a label to trigger internal CI checks
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: |
+          USERNAME="${{ github.event.pull_request.user.login }}"
+          LABEL="ready-for-internal-ci"
```
aibaars, line 17: It's unlikely that `pull_request.user.login` contains dangerous symbols like `"` and `$()` but better safe than sorry:
```suggestion
        env:
          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
          USERNAME: ${{ github.event.pull_request.user.login }}
          PR_NUMBER: ${{ github.event.pull_request.number }}
          LABEL: "ready-for-internal-ci"
        run: |
```

.github/workflows/internal-ci-checks.yml
```diff
@@ -0,0 +1,28 @@
...
+          set +eo pipefail
+          echo "Checking if user $USERNAME is a member of the CodeQL team"
+          gh api -H "Accept: application/vnd.github+json" /orgs/github/teams/codeql/memberships/$USERNAME > /dev/null 2>&1
+          if [ "$?" == 0 ]; then
```
aibaars, line 21: We can drop `-H "Accept: application/vnd.github+json"` . It works fine without, I suppose the `gh api` command sets that by default. I'd recommend adding quotes around the URL argument even though the `USERNAME` variable is likely free of spaces and other unwanted symbols.

The API returns a bit of JSON like  `{"role":"member","state":"active","url":...}` . I think we should check the `state` property and accept only `state: "active"`. The state property can also have value `pending` and perhaps even more values. A `--jq`  flag of `gh api` is very handy.

```suggestion          
          if [ "$(gh api "/orgs/github/teams/codeql/memberships/$USERNAME" --jq .state)" == "active" ]; then
```

.github/workflows/internal-ci-checks.yml
```diff
@@ -0,0 +1,28 @@
...
+              echo "User $USERNAME is a member of the CodeQL team"
+              echo "Adding 'ready-for-internal-ci' label"
+              gh pr edit "${{ github.event.pull_request.number }}" --repo $GITHUB_REPOSITORY --add-label $LABEL
```
aibaars, line 24:
```suggestion
              echo "Adding '${LABEL}' label"
              gh pr edit "${PR_NUMBER}" --repo "$GITHUB_REPOSITORY" --add-label "$LABEL"
```

.github/workflows/internal-ci-checks.yml
```diff
@@ -0,0 +1,28 @@
...
+          else
+            echo "User $USERNAME is not a member of the CodeQL team"
+            echo "To trigger the internal CI, a maintainer needs to add the ready-for-internal-ci label"
```
aibaars, line 27:
```suggestion
            echo "To trigger the internal CI, a maintainer needs to add the '${LABEL}' label"
```

aibaars (commented):
Looks good overall, some suggested improvements.

.github/workflows/internal-ci-checks.yml
```diff
@@ -0,0 +1,28 @@
+name: Internal CI Checks
+# This workflows checks if the author of a PR is a member of the CodeQL team
+# and adds `ready-for-internal-ci` label to trigger the internal CI
+
+on:
+  pull_request_target:
+
```
aibaars, line 7: I recommend adding a permissions block to request only the permission that are needed for label posting.

fossjunkie: Thank you! Done: https://github.com/github/codeql/pull/10969/commits/e557e518f4f315ddcbc116967a2596e103238f33

.github/workflows/internal-ci-checks.yml
```diff
@@ -0,0 +1,31 @@
...
+permissions:
+  pull-requests: write
```
aibaars, line 9: I think we need `read:org` permissions for the team membership check.
```suggestion
  pull-requests: write
  org: read
```

aibaars: Although that permission is not mentioned in the Actions docs: https://docs.github.com/en/actions/using-jobs/assigning-permissions-to-jobs

.github/workflows/internal-ci-checks.yml
```diff
@@ -0,0 +1,28 @@
...
+      - name: Set a label to trigger internal CI checks
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          USERNAME: ${{ github.event.pull_request.user.login }}
+          PR_NUMBER: ${{ github.event.pull_request.number }}
+          LABEL: "ready-for-internal-ci"
+        run: |
+          set +eo pipefail
+          echo "Checking if user $USERNAME is a member of the CodeQL team"
+          if [ "$(gh api "/orgs/github/teams/codeql/memberships/$USERNAME" --jq .state)" == "active" ]; then
```
aibaars, line 21: I think `secrets.GITHUB_TOKEN` does not have permissions to query team membership information.

fossjunkie:
This workflow will be implemented in the internal repo.

PR #10975, opened 2022-10-25
C++: repair InconsistentLoopDirection

rdmarsh2:


MathiasVP (approved):
LGTM!

Indeed, this change makes a lot of sense. IR dataflow steps are "smaller" than AST dataflow steps, so we're less likely to get from one `ExprNode` to another in a single step.

MathiasVP:
Oops. It looks like the file needs to be autoformatted. Feel free to self-merge after doing the formatting.

MathiasVP (approved):

PR #10983, opened 2022-10-26
Swift: move libraries from `tools` to `third_party`

redsun82:
I did this as part of something I was trying out that ended up not really requiring this cleanup, but I guessed it still made sense.


jketema (approved):
LGTM and seems more consistent with what we do internally.

PR #10984, opened 2022-10-26
JS: Add Next.js parameters as source

tyage:
I added following params as remote flow sources, in addition to the params in getStaticProps function.

- context parameter `params` and `query` in getServerSideProps function
  - https://nextjs.org/docs/api-reference/data-fetching/get-server-side-props#context-parameter
- query property of Next.js router
  - https://nextjs.org/docs/routing/dynamic-routes


erik-krogh (commented):
Looks pretty good 👍  

But the `javascript/ql/test/query-tests/Security/CWE-079/DomBasedXss/XssWithAdditionalSources.ql` test is failing.   
You can use e.g. `codeql test run --learn <path-to-test>` to update the expected output of a test.

tyage:
Thanks!
I forgot to apply the test `XssWithAdditionalSources`.
I'll fix it soon.

erik-krogh (approved):
Thanks 👍

Issue #10937, opened 2022-10-21
Analysis of module-ified TypeScript compiler repo takes >6 hours, normally 6 minutes


jakebailey:
**Description of the issue**

The TypeScript repo uses the CodeQL action, with the default `javascript-queries` set. For TypeScript 5.0, we're going to be changing the codebase from namespaces to modules. However, in my testing of that new compiler, I noticed that the CodeQL workflow would always time out at 6 hours. See: https://github.com/microsoft/TypeScript/actions/runs/3229429756/jobs/5286722214

I ran CodeQL locally, and it looks like it gets stuck on these queries:

```
TaintedPath.ql                 : iteration 2 of Configuration#e7756e4f::appendStep#5#fffff etc
CommandInjection.ql            : iteration 26 of Configuration#e7756e4f::reachesReturn#4#ffff#reorder_3_0_1_2 etc
CodeInjection.ql               : iteration 26 of Configuration#e7756e4f::reachesReturn#4#ffff#reorder_3_0_1_2 etc
ImproperCodeSanitization.ql    : iteration 25 of Configuration#e7756e4f::reachesReturn#4#ffff#reorder_3_0_1_2 etc
UnsafeDynamicMethodAccess.ql   : iteration 20 of Configuration#e7756e4f::reachesReturn#4#ffff#reorder_3_0_1_2 etc
CleartextLogging.ql            : iteration 24 of Configuration#e7756e4f::reachesReturn#4#ffff#reorder_3_0_1_2 etc
RegExpInjection.ql             : iteration 26 of Configuration#e7756e4f::reachesReturn#4#ffff#reorder_3_0_1_2 etc
UnvalidatedDynamicMethodCall.ql: iteration 12 of Configuration#e7756e4f::appendStep#5#fffff etc
InsecureDownload.ql            : iteration 34 of Configuration#e7756e4f::reachesReturn#4#ffff#reorder_3_0_1_2 etc
PrototypePollutingAssignment.ql: iteration 21 of Configuration#e7756e4f::reachesReturn#4#ffff#reorder_3_0_1_2 etc
RequestForgery.ql              : iteration 26 of Configuration#e7756e4f::reachesReturn#4#ffff#reorder_3_0_1_2 etc
```

Letting these run to completion overnight took upwards of 9 hours on my beefy machine.

[...]

PR #10985, opened 2022-10-26
JS: Do not track returned values out of the enclosing function

asgerf:
`reachesReturn` would in some cases track the returned value out to call sites, which doesn't lead anywhere. It could also track into callees via return edges, but that's not useful either, since summaries take care of flow through callees.

This fixes a performance issue raised by https://github.com/github/codeql/issues/10937 (but it's possible there are more underlying issues to be fixed)

[Evaluation](https://github.com/github/codeql-dca-main/tree/data/asgerf/reaches-return-__default__code-scanning__1/reports) looks neutral


erik-krogh (approved):
👍

PR #10986, opened 2022-10-26
JS: push more context into load/store steps from the exploratory flow-analysis

erik-krogh:
Saves an additional 15s (out of ~6 minutes) when running the code-scanning suite on the database from https://github.com/github/codeql/issues/10937 

[Evaluation looks somewhat uneventful](https://github.com/github/codeql-dca-main/tree/data/erik-krogh/pr-10986-21e7e2__default__code-scanning/reports).   
I'm pretty sure the cache size is just noise.

### DIFF
--- a/javascript/ql/lib/semmle/javascript/dataflow/Configuration.qll
+++ b/javascript/ql/lib/semmle/javascript/dataflow/Configuration.qll
@@ -1175,6 +1175,7 @@ private predicate parameterPropReadStep(
     invk = getAwaitOperand(succ)
   ) and
   callInputStep(f, invk, arg, parm, cfg) and
+  prop = pragma[only_bind_into](getARelevantProp(cfg)) and
   (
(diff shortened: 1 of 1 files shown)


asgerf (commented):
Implementation LGTM but the evaluation looks a bit underwhelming now that the other PR has been merged. It's leaning towards a slodown, with a 17s increase in vscode being the worst offender. Maybe that was just an unlucky run, but if it's really just trading a speedup in one project for a slowdown in another, I'm not sure we should merge it.

erik-krogh:
> It's leaning towards a slowdown

I did a repeat evaluation: https://github.com/github/codeql-dca-main/tree/data/erik-krogh/pr-10986-21e7e2__default__code-scanning__1/reports 
It shows much more neutral performance.

asgerf (approved):
Great, that looks more reasonable 👍

PR #10987, opened 2022-10-26
Swift: use `std::filesystem` and `picoSHA2`

redsun82:
This replaces usages of `llvm::fs` and string manipulation with `std::filesystem`, also replacing `std::string` with `std::filesystem::path` where it made sense.

Moreover MD5 hashing used in macOS file remapping was replaced by SHA256 hashing using a small header-only SHA256 C++ library with an MIT license, https://github.com/okdshin/PicoSHA2.

File contents hashing was relocated to the newly created `file` library for later planned reuse.

In order to review changes to `TargetFile.cpp` it's best to exclude e422a4eef9dda4917df1e658ecd31202d22287f5 from the diff view, so that actual changes are not hidden by the file move.


AlexDenisov (approved):
:shipit:

PR #10994, opened 2022-10-26
C++: repair the ReturnCstr query

rdmarsh2:


MathiasVP:
Note: This PR probably conflicts with https://github.com/github/codeql/pull/10995, but I think we should get your PR in first. Then we can fixup https://github.com/github/codeql/pull/10995 afterwards.

rdmarsh2:
There's still some internal test failures - should we handle those in this PR or handle them when we merge `replace-ast-with-ir-use-usedataflow` into main?

MathiasVP:
> There's still some internal test failures - should we handle those in this PR or handle them when we merge `replace-ast-with-ir-use-usedataflow` into main?

Feel free to ignore the test failures in the internal repo. I think we should fixup the test failures in `library-tests/dataflow/taint-tests/taint.ql` by changing the inline expectation comments, however.

cpp/ql/lib/semmle/code/cpp/ir/dataflow/internal/DataFlowUtil.qll
```diff
@@ -738,11 +738,19 @@ private predicate exprNodeShouldBeIndirectOperand(IndirectOperand node, Expr e,
   not convertedExprMustBeOperand(e)
 }
 
+private predicate exprNodeShouldBeIndirectOutNode(IndirectArgumentOutNode node, Expr e) {
+  exists(CallInstruction call |
```
MathiasVP, line 742: Thinking about this some more: Should this only hold when call targets a constructor?

rdmarsh2: I think that should be true already because `ConstructorCall` will always target a constructor, but I can add the constraint.

rdmarsh2:
> I think we should fixup the test failures in library-tests/dataflow/taint-tests/taint.ql by changing the inline expectation comments, however.

Missed these in testing... They look messier than I expected, like something is off about our source/sink/flow models?

cpp/ql/lib/semmle/code/cpp/ir/dataflow/internal/DataFlowUtil.qll
MathiasVP: I might just be blind, but where is the ConstructorCall constraint currently?

MathiasVP:
> Missed these in testing... They look messier than I expected, like something is off about our source/sink/flow models?

Yeah, we still have an [issue](https://github.com/github/codeql-c-team/issues/1257) for repairing the internal tests. I think it's pretty simple to fix by slightly changing the source and/or sink.

So I suggest that you just update the inline expectation comments so that the test pass, and then we'll leave fixing the tests to https://github.com/github/codeql-c-team/issues/1257. It might also be that some of the tests are failing because of https://github.com/github/codeql/pull/10994#discussion_r1006141865?

cpp/ql/lib/semmle/code/cpp/ir/dataflow/internal/DataFlowUtil.qll
rdmarsh2: It's in `ConstructorCall`'s charpred

rdmarsh2:
The root cause of the `taint.ql` failures was the missing constructor constraint.

PR #10995, opened 2022-10-26
C++: Fix `asExpr` and `asIndirectExpr` in IR dataflow

MathiasVP:
This PR contains my ongoing progress on fixing some of the issues I encountered with @jketema while we were trying to fixup the `cpp/non-constant-format` query in preparation for use-use dataflow.

<s>I'm still not super happy with the PR. I think there are some nodes that we want to have a `asExpr` result for, but which currently don't have a result because `.getConvertedResultExpression()` doesn't have a result for some of the instruction nodes we care about (for instance, `VariableAddressInstruction`s).</s> This has been fixed in e2a0d62adfb2fcfb866e949dbcde160adcc934bd.


cpp/ql/test/library-tests/syntax-zoo/dataflow-ir-consistency.expected
```diff
@@ -564,11 +564,17 @@ uniqueNodeToString
 | break_labels.c:2:11:2:11 | x | Node should have one toString but has 2. |
 | break_labels.c:2:11:2:11 | x | Node should have one toString but has 2. |
 | break_labels.c:4:9:4:9 | i | Node should have one toString but has 2. |
+| break_labels.c:4:9:4:9 | i indirection | Node should have one toString but has 2. |
```
jketema, line 567: Should I be worried about the changes to this file?

jketema: Or is this basically a consequence of the direct node having the same problem and the strings being constructed differently now?

MathiasVP: Yeah, I think so.

jketema: You think I should be worried, or you think it's a consequence of? 😄

MathiasVP:
Merge conflicts fixed. They were as simple as accepting both changes 🎉.

PR #10996, opened 2022-10-26
Swift: Add MethodDecl.hasQualifiedName

geoffw0:
Add `MethodDecl.hasQualifiedName` helper predicates (two variants) plus `AbstractFunctionDecl.hasName` (not so useful but feels like its missing).

There's a test to prove we do the right thing in some tricky cases like nested classes, protocols and extensions.  I have only _used_ the new predicates in one query so far (as a proof of concept), but I expect it to be useful in many more places and I intend to follow up with a PR updating existing queries.

### DIFF
--- a/swift/ql/lib/codeql/swift/elements/decl/MethodDecl.qll
+++ b/swift/ql/lib/codeql/swift/elements/decl/MethodDecl.qll
@@ -26,7 +26,8 @@ class MethodDecl extends AbstractFunctionDecl {
    * Holds if this function is called `funcName` and its a member of a
    * class, struct, extension, enum or protocol call `typeName`.
    */
-  cached predicate hasQualifiedName(string typeName, string funcName) {
+  cached
+  predicate hasQualifiedName(string typeName, string funcName) {
     this.getName() = funcName and
     (
       exists(NominalTypeDecl c |
--- a/swift/ql/lib/codeql/swift/elements/decl/MethodDecl.qll
+++ b/swift/ql/lib/codeql/swift/elements/decl/MethodDecl.qll
@@ -26,7 +26,7 @@ class MethodDecl extends AbstractFunctionDecl {
    * Holds if this function is called `funcName` and its a member of a
    * class, struct, extension, enum or protocol call `typeName`.
    */
-  predicate hasQualifiedName(string typeName, string funcName) {
+  cached predicate hasQualifiedName(string typeName, string funcName) {
     this.getName() = funcName and
     (
       exists(NominalTypeDecl c |
--- a/swift/ql/lib/codeql/swift/elements/decl/MethodDecl.qll
+++ b/swift/ql/lib/codeql/swift/elements/decl/MethodDecl.qll
@@ -21,4 +21,33 @@ class MethodDecl extends AbstractFunctionDecl {
     or
     this = getAMember(any(ProtocolDecl c))
   }
+
+  /**
+   * Holds if this function is called `funcName` and its a member of a
+   * class, struct, extension, enum or protocol call `typeName`.
+   */
+  predicate hasQualifiedName(string typeName, string funcName) {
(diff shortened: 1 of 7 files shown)


swift/ql/lib/codeql/swift/elements/decl/MethodDecl.qll
```diff
@@ -21,4 +21,33 @@ class MethodDecl extends AbstractFunctionDecl {
...
+
+  /**
+   * Holds if this function is called `funcName` and its a member of a
+   * class, struct, extension, enum or protocol call `typeName`.
+   */
+  predicate hasQualifiedName(string typeName, string funcName) {
+    this.getName() = funcName and
+    (
+      exists(NominalTypeDecl c |
+        c.getFullName() = typeName and
```
redsun82, line 33: this might incur the string concatenation performance penalty, but only in case of nested types. We might maybe want to revisit this if this proves to be bothersome. Doesn't `Type::getName` already give the dotted name of nested types straight out of the DB?

swift/ql/lib/codeql/swift/elements/decl/MethodDecl.qll
```diff
@@ -21,4 +21,33 @@ class MethodDecl extends AbstractFunctionDecl {
     or
     this = getAMember(any(ProtocolDecl c))
   }
+
+  /**
+   * Holds if this function is called `funcName` and its a member of a
+   * class, struct, extension, enum or protocol call `typeName`.
+   */
+  predicate hasQualifiedName(string typeName, string funcName) {
```
MathiasVP, line 29: This should probably be `cached`.

swift/ql/lib/codeql/swift/elements/decl/MethodDecl.qll
```diff
@@ -21,4 +21,33 @@ class MethodDecl extends AbstractFunctionDecl {
...
+  }
+
+  /**
+   * Holds if this function is called `funcName` and its a member of a
+   * class, struct, extension, enum or protocol call `typeName` in a module
+   * called `moduleName`.
+   */
+  predicate hasQualifiedName(string moduleName, string typeName, string funcName) {
+    this.hasQualifiedName(typeName, funcName) and
+    this.getModule().getFullName() = moduleName
```
MathiasVP, line 51: I think the `getFullName` on `Module` should probably also be cached, but we can worry about that later 👍.

swift/ql/lib/codeql/swift/elements/decl/MethodDecl.qll
MathiasVP: I don't think it does. It certainly didn't appear to be the case when I added it for MaD 🤔.

swift/ql/lib/codeql/swift/elements/decl/MethodDecl.qll
geoffw0: Done.  I wasn't sure whether to do the three argument version as its only a small bit of logic + a call to the two argument version.  So I left it uncached for now.

swift/ql/lib/codeql/swift/elements/decl/MethodDecl.qll
geoffw0: Seems reasonable.  Done as well.

swift/ql/lib/codeql/swift/elements/decl/MethodDecl.qll
MathiasVP: Yeah, I don't think the three-argument version should be cached (because, as you say, it's only a small wrapper around cached stuff at this point).

MathiasVP (commented):
LGTM once the formatting is fixed.

geoffw0:
Ah, sorry.  Fixed.

MathiasVP (approved):

PR #10997, opened 2022-10-26
Go: Extract locations of successfully extracted files

henrymercer:
Switch the successfully extracted files query to the `location, message` results format so that we get rich location information when exporting the results of this query to SARIF.  Previously the query used the `message` results format, which meant the interpreted results lacked a location.


smowton:
Looks good once test changes are accepted

smowton (dismissed):

smowton (approved):

PR #10998, opened 2022-10-26
Python: Add failing ESSA use-use test

RasmusWL:
I initially created this as a dataflow test, but then realized it could just be an ESSA test. I couldn't find any existing ESSA tests though :| so created a new dir for it.

@yoff let me know if you think this should actually be under our dataflow tests, since the adjacent use-use was only added to serve the new dataflow lib.

### DIFF
--- a/python/ql/test/library-tests/essa/ssa-compute/UseUse.ql
+++ b/python/ql/test/library-tests/essa/ssa-compute/UseUse.ql
@@ -1,0 +1,41 @@
+import python
+import semmle.python.essa.SsaCompute
+import TestUtilities.InlineExpectationsTest
+
+class UseTest extends InlineExpectationsTest {
+  UseTest() { this = "UseTest" }
+
+  override string getARelevantTag() { result in ["use-use", "def-use", "def"] }
+
+  override predicate hasActualResult(Location location, string element, string tag, string value) {
+    exists(location.getFile().getRelativePath()) and
+    exists(string name | name in ["x", "y"] |
+      exists(NameNode nodeTo, Location prevLoc |
+        (
+          exists(NameNode nodeFrom | AdjacentUses::adjacentUseUse(nodeFrom, nodeTo) |
+            prevLoc = nodeFrom.getLocation() and
+            name = nodeFrom.getId() and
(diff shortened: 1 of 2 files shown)


tausbn (dismissed):
I think this looks reasonable (but I'll hold off on merging so yoff can have a gander). 👍

python/ql/test/library-tests/essa/ssa-compute/UseUse.ql
```diff
@@ -0,0 +1,41 @@
...
+      exists(NameNode nodeTo, Location prevLoc |
+        (
+          exists(NameNode nodeFrom | AdjacentUses::adjacentUseUse(nodeFrom, nodeTo) |
+            prevLoc = nodeFrom.getLocation() and
+            name = nodeFrom.getId() and
+            tag = "use-use"
+          )
+          or
+          exists(EssaVariable var | AdjacentUses::firstUse(var, nodeTo) |
+            prevLoc = var.getDefinition().getLocation() and
```
yoff, line 22: There is a convenience predicate for this (and also: `getDefinition()` simply returns `this`)
```suggestion
            prevLoc = var.getLocation() and
```

python/ql/test/library-tests/essa/ssa-compute/UseUse.ql
```diff
@@ -0,0 +1,41 @@
...
+          )
+        ) and
+        value = name + ":" + prevLoc.getStartLine() and
+        location = nodeTo.getLocation() and
+        element = nodeTo.toString()
+      )
+      or
+      exists(EssaVariable var | AdjacentUses::firstUse(var, _) |
+        value = var.getName() and
+        location = var.getDefinition().getLocation() and
```
yoff, line 34:
```suggestion
        location = var.getLocation() and
```
same as above

yoff (changes requested):
Super nice to have this test! 👍 
Only reason I mention changes is that it looks different from the code in data flow and I had to remind myself that it does the same thing.

I think it is fine to keep the test here by the predicates they test.

yoff (approved):
LGTM

PR #10999, opened 2022-10-26
InlineExpectationsTest: Fail if missing `getARelevantTag`

RasmusWL:
This PR changes InlineExpectationsTest so it explicitly fails if you provided an actual result with a tag that is not part of `getARelevantTag`.

To illustrate why this is useful, previously if you had the annotated code below

```py
foo # $ foo=val

# with wrong value
foo # $ foo=bad-value

# there is a typo here, so this result is actually missing!
fooo # $ foo=val
```

and got the following in your .expected file, you might assume that the remainder of the expectations with `foo` tags were valid, which they are not!

```
| test.py:4:1:4:3 | foo | Unexpected result: foo=val |
```


smowton (dismissed):

erik-krogh:
A fixed a version bump in the QL-for-QL `qlpack.yml` directly to this branch, that should get it to compile again.

RasmusWL:
Java (kotlin) tests might still fail until https://github.com/github/codeql/pull/11006 is merged

michaelnebel (dismissed):
C# LGTM

erik-krogh (dismissed):

MathiasVP:
I might be misunderstanding what the goal of this PR is, but doesn't this change go against the QLDoc on [getARelevantTag](https://github.com/github/codeql/blob/main/cpp/ql/test/TestUtilities/InlineExpectationsTest.qll#L111):
```ql
/**
 * [...] Any expected result comments for a tag that is not returned by the `getARelevantTag()`
 * predicate for an active test will be ignored. This makes it possible to write multiple tests in
 * different `.ql` files that all query the same source code.
 */
```
?

MathiasVP (dismissed):

RasmusWL:
Java was failing CI. I fixed up the annotations for the `JaxWs` modeling, but the integrations test will probably keep on failing.

java/ql/test/library-tests/frameworks/JaxWs/JakartaRs1.java
```diff
@@ -41,7 +41,7 @@ void Post() { // $ ResourceMethod ResourceMethodOnResourceClass
   @Produces("text/plain") // $ ProducesAnnotation=text/plain
   @DELETE
   double Delete() { // $ ResourceMethod=text/plain ResourceMethodOnResourceClass
-    return 0.0; // $ XssSink
+    return 0.0; // $ MISSING: XssSink
```
atorralba, line 44: This seems to be an invalid sink due to the `text/plain` content type (see https://github.com/github/codeql/blob/f237360d814afe7eb8a9da963344537d7d3bd31f/java/ql/lib/semmle/code/java/security/XSS.qll#L119-L120).

So we should actually remove the expectations.

```suggestion
    return 0.0;
```

java/ql/test/library-tests/frameworks/JaxWs/JakartaRs1.java
```diff
@@ -77,7 +77,7 @@ boolean Post() { // $ ResourceMethod=text/html ResourceMethodOnResourceClass
     @Produces(MediaType.TEXT_PLAIN) // $ ProducesAnnotation=text/plain
     @DELETE
     double Delete() { // $ ResourceMethod=text/plain ResourceMethodOnResourceClass
-      return 0.0; // $ XssSink
+      return 0.0; // $ MISSING: XssSink
```
atorralba, line 80: Same as above:

```suggestion
      return 0.0;
```

RasmusWL:
> This seems to be an invalid sink due to the `text/plain` content type

Seems reasonable to me. Have removed the annotations from both files.

RasmusWL:
https://github.com/github/codeql/pull/11025 should fix the Java integration tests, I will merge this PR regardless, since it doesn't touch these!

PR #11001, opened 2022-10-26
Swift: Unsafe JS Eval Query

d10c:


swift/ql/src/queries/Security/CWE-094/UnsafeJsEval.ql
```diff
@@ -0,0 +1,119 @@
...
+    ).getArgument(1).getExpr() = this.asExpr()
+  }
+}
+
+// TODO: Consider moving the following to the library, e.g.
+// - Decl.hasQualifiedName(moduleName?, declaringDeclName?, declName)
+// - parentDecl = memberDecl.getDeclaringDecl() <=> parentDecl.getAMember() = memberDecl
+IterableDeclContext getDeclaringDeclOf(Decl member) { result.getAMember() = member }
+
+MethodDecl getMethodWithQualifiedName(string className, string methodName) {
```
geoffw0, line 89: As of https://github.com/github/codeql/pull/10996 (which was opened after you made this PR) we have `MethodDecl.getQualifiedName` for this, and https://github.com/github/codeql/pull/11034 adds `AbstractFunctionDecl.hasGlobalName` that can be used instead of the predicate below.  Feel free to either update your QL, or leave it to me as part of the follow-up to my work.

.vscode/settings.json
```diff
@@ -1,3 +1,10 @@
 {
-    "omnisharp.autoStart": false
+  "omnisharp.autoStart": false,
+  "cmake.sourceDirectory": "${workspaceFolder}/swift",
+  "cmake.buildDirectory": "${workspaceFolder}/bazel-cmake-build",
+  "clangd.arguments": [
+    "--header-insertion=never",
+    "--compile-commands-dir=${workspaceFolder}/swift",
+    "--query-driver=**"
+  ]
```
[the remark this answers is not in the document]
d10c: Yeah, I got it from [here](https://github.com/hedronvision/bazel-compile-commands-extractor#editor-setup--for-autocomplete-based-on-compile_commandsjson) back when I was trying to get bazel to produce a clangd compilation database, but there's a good chance it's not necessary, so I'll remove this change and see if it still works.

swift/ql/test/query-tests/Security/CWE-094/UnsafeJsEval.swift
```diff
@@ -0,0 +1,299 @@
...
+}
+
+func testUsage(_ sink: @escaping (String) async throws -> ()) {
+	Task {
+		let localString = "console.log('localString')"
+		let localStringFragment = "'localStringFragment'"
+		let remoteString = getRemoteData()
+
+		try! await sink(localString) // GOOD: the HTML data is local
+		try! await sink(getRemoteData()) // BAD: HTML contains remote input, may access local secrets
```
geoffw0, line 177: How will we tell which sink types each case is working for?  Might we be better off with thorough tests for just one sink type, then an additional simple test of each other sink type to verify coverage?

swift/ql/test/query-tests/Security/CWE-094/UnsafeJsEval.swift
```diff
@@ -0,0 +1,299 @@
...
+		try! await sink(remoteString) // BAD
+
+		try! await sink("console.log(" + localStringFragment + ")") // GOOD: the HTML data is local
+		try! await sink("console.log(" + remoteString + ")") // BAD
+
+		let localData = Data(localString.utf8)
+		let remoteData = Data(remoteString.utf8)
+
+		try! await sink(String(decoding: localData, as: UTF8.self)) // GOOD: the data is local
+		try! await sink(String(decoding: remoteData, as: UTF8.self)) // BAD: the data is remote
```
geoffw0, line 187: We may well need to plan follow-up work to catch these ones, but I see a lot of code like this in the wild so it should be worth the extra effort.

swift/ql/test/query-tests/Security/CWE-094/UnsafeJsEval.swift
d10c: True, it would be simpler if I inlined (most of) the tests. But we can eye-ball the correctness of the test by running the query in VSCode and checking that each of the sinks has the same number of paths leading to it.

swift/ql/src/queries/Security/CWE-094/UnsafeJsEval.ql
```diff
@@ -94,8 +94,57 @@ class UnsafeJsEvalConfig extends TaintTracking::Configuration {
 
   override predicate isSink(DataFlow::Node node) { node instanceof Sink }
 
-  override predicate isSanitizer(DataFlow::Node node) {
-    none() // TODO: A conversion to a primitive type or an enum
+  // TODO: convert to new taint flow models
```
geoffw0, line 97: I think we should aim to create follow-up tasks for most of these things rather than addressing them all now.  Many are planned already.  And I don't want this PR just growing and growing.

swift/ql/lib/codeql/swift/controlflow/internal/ControlFlowGraphImpl.qll
```diff
@@ -1015,6 +1015,14 @@ module Decls {
 
 module Exprs {
   module AssignExprs {
+    /**
+     * The control-flow of a `DiscardAssignmentExpr`, which represents the
+     * `_` leaf expression that may appear on the left-hand side of an `AssignExpr`.
+     */
+    private class DiscardAssignmentExprTree extends AstLeafTree {
+      override DiscardAssignmentExpr ast;
+    }
```
geoffw0, line 1024: We decided this was the right way to fix this in the end, didn't we - so no follow-up work required here?

swift/ql/src/queries/Security/CWE-094/UnsafeJsEval.ql
```diff
@@ -94,8 +94,57 @@ class UnsafeJsEvalConfig extends TaintTracking::Configuration {
...
+        ).getArgument(0)
+    |
+      nodeFrom.asExpr() = arg.getExpr() and
+      nodeTo.asExpr() = arg.getApplyExpr()
+    )
+    or
+    exists(CallExpr ce, Expr self, AbstractClosureExpr closure |
+      ce.getStaticTarget()
+          .getName()
+          .matches(["withContiguousStorageIfAvailable(%)", "withUnsafeBufferPointer(%)"]) and
```
geoffw0, line 137: This is currently planned (but perhaps not yet prioritized highly enough) as https://github.com/github/codeql-c-team/issues/1333 .  Please do add comments to that issue about any relevant details you've discovered.

swift/ql/src/queries/Security/CWE-094/UnsafeJsEval.ql
```diff
@@ -94,8 +94,57 @@ class UnsafeJsEvalConfig extends TaintTracking::Configuration {
...
+      self = ce.getQualifier() and
+      ce.getArgument(0).getExpr() = closure
+    |
+      nodeFrom.asExpr() = self and
+      nodeTo.(DataFlow::ParameterNode).getParameter() = closure.getParam(0)
+    )
+    or
+    exists(MemberRefExpr e, Expr self, VarDecl member |
+      self.getType().getName() = "String" and
+      member.getName() = ["utf16", "utf8CString"]
```
geoffw0: This is being worked on (https://github.com/github/codeql-c-team/issues/1301).

swift/ql/test/query-tests/Security/CWE-094/UnsafeJsEval.swift
```diff
@@ -184,7 +184,7 @@ func testUsage(_ sink: @escaping (String) async throws -> ()) {
 		let remoteData = Data(remoteString.utf8)
 
 		try! await sink(String(decoding: localData, as: UTF8.self)) // GOOD: the data is local
-		try! await sink(String(decoding: remoteData, as: UTF8.self)) // BAD: the data is remote
+		try! await sink(String(decoding: remoteData, as: UTF8.self)) // BAD [NOT DETECTED]: the data is remote (TODO: model Data taint sources)
```
geoffw0: A model for `Data` is planned as well https://github.com/github/codeql-c-team/issues/1334 .  I think its another important one.

geoffw0 (commented):
Looks great.  We have a lot of follow-up work to do on models!

Also thanks for updating the QL to use the new predicates `(MethodDecl).hasQualifiedName` and `FreeFunctionDecl`.

swift/ql/src/queries/Security/CWE-094/UnsafeJsEval.ql
d10c: Makes sense. Though it's a tradeoff between a growing PR and shipping more tech debt. An alternative is that I move these additional taint steps into the library, and create issues to flesh the library out to handle other cases that I didn't handle here.

swift/ql/lib/codeql/swift/controlflow/internal/ControlFlowGraphImpl.qll
d10c: Correct.

swift/ql/src/queries/Security/CWE-094/UnsafeJsEval.ql
d10c: This particular taint step is an overly broad approximation (matches name with % and doesn't take self type into account), so probably shouldn't be moved to the library as-is. A thorough taint step library for this these kinds of methods would be welcome.

swift/ql/src/queries/Security/CWE-094/UnsafeJsEval.ql
geoffw0: That sounds fine as long as its straightforward.

I have a preference for smaller, faster PRs where possible as they're easier for everyone to review and manage.

swift/ql/src/queries/Security/CWE-094/UnsafeJsEval.ql
geoffw0: Good point.  I've made a note on the above issue to remove this approximation when we have a proper solution.  It will do until then.

swift/ql/test/query-tests/Security/CWE-094/UnsafeJsEval.swift
geoffw0: ... and now scheduled for next week.  Perfectly fine for this to be merged with a few missing results though.

swift/ql/src/queries/Security/CWE-094/UnsafeJsEvalGood.swift
```diff
@@ -0,0 +1,10 @@
+let webview: WKWebView
+let remoteData = try String(contentsOf: URL(string: "http://example.com/evil.json")!)
+
+...
+
+_ = try await webview.callAsyncJavaScript(
+    "alert(JSON.parse(data))",
```
geoffw0, line 7: Is there a good reason to have `JSON.parse(data)` here rather than just `data`?

Other than that question I really like this example.  Its simple but plausible.

d10c: Presumably the intent of the example is to read a remote JSON string, either by inlining into Js (bad) or parsing as JSON (good). Though perhaps alert() is not the best function to wrap this in.

geoffw0: I'm happy with `alert`, its simple and the bad example is clearly vulnerable.  On reflection I think its the presence of `JSON.parse` is one example but not the other that bothers me just a little.  A reader could misunderstand that to be relevant to the safety of the good example compared to the bad one.

swift/ql/test/query-tests/Security/CWE-094/UnsafeJsEval.swift
```diff
@@ -0,0 +1,336 @@
...
+}
+
+func testAsync(_ sink: @escaping (String) async throws -> ()) {
+	Task {
+		let localString = "console.log('localString')"
+		let localStringFragment = "'localStringFragment'"
+		let remoteString = getRemoteData()
+
+		try! await sink(localString) // GOOD: the HTML data is local
+		try! await sink(getRemoteData()) // BAD [NOT DETECTED - TODO: extract Callables of @MainActor method calls]: HTML contains remote input, may access local secrets
```
geoffw0, line 178: I wonder if the results of this test might be easier to interpret if we replace the calls to `getRemoteData` with an in-place taint source such as:
```
try String(contentsOf: URL(string: "http://example.com/")!)
```
Then the sources + sinks on the result lines should show us all of the combinations are working?

swift/ql/src/queries/Security/CWE-094/UnsafeJsEval.qhelp
```diff
@@ -0,0 +1,32 @@
...
+
+<p>To fix the problem, we sanitize the remote data by passing it using the <code>arguments</code> dictionary of <code>WKWebView.callAsyncJavaScript</code>. This ensures that untrusted data cannot be evaluated as JavaScript source code.</p>
+
+<sample src="UnsafeJsEvalGood.swift" />
+
+</example>
+<references>
+
+<li>
+  <a href="https://developer.apple.com/documentation/webkit/wkwebview/3824703-callasyncjavascript">Apple Developer Documentation - WKWebView.callAsyncJavaScript(_:arguments:in:contentWorld:)</a>
```
sabrowning1, line 28:
```suggestion
  Apple Developer Documentation: <a href="https://developer.apple.com/documentation/webkit/wkwebview/3824703-callasyncjavascript">WKWebView.callAsyncJavaScript(_:arguments:in:contentWorld:)</a>
```

Small change to align with the [query help style guide](https://github.com/github/codeql/blob/main/docs/query-help-style-guide.md#websites)

swift/ql/src/queries/Security/CWE-094/UnsafeJsEval.qhelp
```diff
@@ -0,0 +1,32 @@
...
+  "-//Semmle//qhelp//EN"
+  "qhelp.dtd">
+<qhelp>
+<overview>
+<p>Evaluating JavaScript that contains a substring from a remote origin may lead to remote code execution. Code under control of an attacker can execute arbitrary unauthorized actions, including exfiltration of local data by sending it to a third party web service.</p>
+
+</overview>
+<recommendation>
+
+<p>When loading JavaScript into a web view, evaluate only known, locally-defined source code. If a part of the input does come from a remote source, instead of injecting it into the JavaScript code to be evaluated, prefer sending it as data to the web view using an API such as <code>WKWebView.callAsyncJavaScript</code> with the <code>arguments</code> dictionary to pass remote data objects.</p>
```
sabrowning1, line 11:
```suggestion
<p>When loading JavaScript into a web view, evaluate only known, locally-defined source code. If part of the input comes from a remote source, do not inject it into the JavaScript code to be evaluated. Instead, send it to the web view as data using an API such as <code>WKWebView.callAsyncJavaScript</code> with the <code>arguments</code> dictionary to pass remote data objects.</p>
```

Small edit to break sentence up for readability

swift/ql/src/queries/Security/CWE-094/UnsafeJsEval.qhelp
```diff
@@ -0,0 +1,32 @@
+<!DOCTYPE qhelp PUBLIC
+  "-//Semmle//qhelp//EN"
+  "qhelp.dtd">
+<qhelp>
+<overview>
+<p>Evaluating JavaScript that contains a substring from a remote origin may lead to remote code execution. Code under control of an attacker can execute arbitrary unauthorized actions, including exfiltration of local data by sending it to a third party web service.</p>
```
sabrowning1, line 6:
```suggestion
<p>Evaluating JavaScript that contains a substring from a remote origin may lead to remote code execution. Code written by an attacker can execute unauthorized actions, including exfiltration of local data through a third party web service.</p>
```

Small edits for concision

swift/ql/src/queries/Security/CWE-094/UnsafeJsEval.qhelp
```diff
@@ -0,0 +1,32 @@
...
+
+</overview>
+<recommendation>
+
+<p>When loading JavaScript into a web view, evaluate only known, locally-defined source code. If a part of the input does come from a remote source, instead of injecting it into the JavaScript code to be evaluated, prefer sending it as data to the web view using an API such as <code>WKWebView.callAsyncJavaScript</code> with the <code>arguments</code> dictionary to pass remote data objects.</p>
+
+</recommendation>
+<example>
+
+<p>In the following example, a call to <code>WKWebView.evaluateJavaScript</code> evaluates JavaScript source code that is tainted with remote data, potentially introducing a code injection vulnerability.</p>
```
sabrowning1, line 16:
```suggestion
<p>In the following (bad) example, a call to <code>WKWebView.evaluateJavaScript</code> evaluates JavaScript source code that is tainted with remote data, potentially introducing a code injection vulnerability.</p>
```

Small addition to make it abundantly clear that this is the bad example

swift/ql/src/queries/Security/CWE-094/UnsafeJsEval.qhelp
```diff
@@ -0,0 +1,32 @@
...
+
+<sample src="UnsafeJsEvalBad.swift" />
+
+<p>To fix the problem, we sanitize the remote data by passing it using the <code>arguments</code> dictionary of <code>WKWebView.callAsyncJavaScript</code>. This ensures that untrusted data cannot be evaluated as JavaScript source code.</p>
```
sabrowning1, line 20:
```suggestion
<p>In the following (good) example, we sanitize the remote data by passing it using the <code>arguments</code> dictionary of <code>WKWebView.callAsyncJavaScript</code>. This ensures that untrusted data cannot be evaluated as JavaScript source code.</p>
```

Small edit to make it abundantly clear that this is the good example

sabrowning1 (commented):
@d10c hello from Docs! 👋🏼 Thanks for your work on the documentation; I made some small suggestions to bring things in line with our style guide. Let me know if you have any questions!

swift/ql/test/query-tests/Security/CWE-094/UnsafeJsEval.swift
d10c: Weirdly enough, they are harder to interpret (see the latest commit). Now there is a separate result for this new in-place source (as expected). But unexpectedly, for the JSEvaluateScript sinks, there are two separate paths for this result: one passing through the JSString constructor function, one omitting this taint step.

geoffw0: > for the JSEvaluateScript sinks, there are two separate paths for this result: one passing through the JSString constructor function, one omitting this taint step.

I think I see what you mean, there's an unnecessary edge:
```
| UnsafeJsEval.swift:287:31:287:97 | call to JSStringCreateWithCharacters(_:_:) : | UnsafeJsEval.swift:291:17:291:17 | jsstr |
```
and also:
```
| UnsafeJsEval.swift:286:51:286:51 | stringBytes : | UnsafeJsEval.swift:291:17:291:17 | jsstr |
```
But they don't appear to be new in this commit.

The new results under `#select` look good to me.  You will need to replace the other calls to `getRemoteData` with inline sinks to see the complete grid of results there.

geoffw0: Notably both unnecessary edges are the transitive closure of more than one `isAdditionalTaintStep` step, ending at the sink.  I wonder if this is a bug in taint flow itself?

MathiasVP: The dataflow library will ["compress"](https://github.com/github/codeql/blob/main/swift/ql/lib/codeql/swift/dataflow/internal/DataFlowImpl.qll#L2060) both dataflow and taint-flow steps before reporting them as edges in the `edges` relation. So we can't really conclude that there's a bug based on the `edges` data in the `.expected` file.

If you want to disable this compression, you can replace `none()` with `any()` in the charpred for `CastNode` [here](https://github.com/github/codeql/blob/main/swift/ql/lib/codeql/swift/dataflow/internal/DataFlowPrivate.qll#L617). Then each local step will be visible in the `edges` relation.

geoffw0: Are you saying the additional edges are expected / harmless?

geoffw0: I checked over the rules in `isAdditionalTaintStep` for this query again and they all seem reasonable.  I'm not sure what else could be generating edges through this - `JSString` isn't modelled yet so it should be nothing.

MathiasVP: Feel free to create an issue for it. I don't think this should block this PR.

sabrowning1 (approved):
Looks good for Docs, thanks for making those changes @d10c! 🎉

swift/ql/test/query-tests/Security/CWE-094/UnsafeJsEval.swift
```diff
@@ -0,0 +1,336 @@
...
+	}
+}
+
+func testSync(_ sink: @escaping (String) -> ()) {
+	let localString = "console.log('localString')"
+	let localStringFragment = "'localStringFragment'"
+	let remoteString = getRemoteData()
+
+	sink(localString) // GOOD: the HTML data is local
+	sink(try! String(contentsOf: URL(string: "http://example.com/")!)) // BAD: HTML contains remote input, may access local secrets
```
geoffw0, line 204: Do you plan to change all of the cases in `testSync` / `testAsync` work this way?  I still think it will be easier to read the results like that - though if you feel otherwise I don't want us to get stuck on this.

d10c (commented):
Last commit LGTM 👍🏻

geoffw0 (approved):

PR #11003, opened 2022-10-26
Kotlin: reintroduce pointless wildcards when a Java declaration explicitly uses them

smowton:
For example, Java code might use `HasOutVariance<? extends String>`, or `HasInVariance<? super Object>`, both of which are needless wildcards and which the Kotlin extractor would previously have refused to reintroduce due to their not specifying a larger type than their bound. However this led to inconsistency with Java extraction, which extracts the type as it appears in source.

This seems to particularly happen with generated code, e.g. the output of the Kotlin protobuf compiler.


igfoo (approved):

smowton:
Merging across an irrelevant test failure caused by crossing PRs and resolved in https://github.com/github/codeql/pull/11006

PR #11004, opened 2022-10-26
Use `${workspace}` for intra-workspace dependencies

dbartol:
Now that the released CLI supports replacement variables in dependency version ranges, we can now mark our published library packs as depending on whatever version of their dependency is in our workspace, without having to manually bump the dependency version every release.

Note that when the packs are published, the dependencies in the published pack file are rewritten to have the correct specific version.


aibaars (approved):
LGTM

aeisenberg (commented):
This looks fine, but this is a breaking change. Older versions of the CLI will no longer be able to read the qlpack.yml files from the repo. I think there needs to be some change notes somewhere, but I'm not sure of the most appropriate place.

I don't think it's necessary in the change notes for individual qlpacks, since compiled and published packs are fine. 

Or, are we fine with implicitly requiring the latest released CLI version in order to handle the latest from main?

dbartol:
> are we fine with implicitly requiring the latest released CLI version in order to handle the latest from main?

This has always been our policy.

erik-krogh (approved):
JS / QL-for-QL 👍

aeisenberg (approved):

PR #11008, opened 2022-10-27
Swift: rework workflows

redsun82:
* A unique workflow file has been created merging all `swift-*.yml`
  workflows, which also means all the checks are visible in one summary view
* Change filtering at job level was added using [dorny/paths-filter][1]
* only one build of the extractor is made, and then shared via cache
  (not as an artifact because of [this longstading issue][2])
* the above also means that if the build is broken, the QL and integration tests are not even started
  which saves us runner resources
* integration tests are now run on on macOS
* qltests on macOS are split in 2 slices to cut on feedback time
* all non QL tests (including autobuilder tests) were moved to the build jobs to avoid loading
  bazel twice, which takes quite some time on macOS runners especially
* the build and QL tests jobs were split in separate Linux and macOS jobs instead of a using a matrix.
  This was done because job dependencies (`needs`) cannot be established between specific matrix sub-jobs,
  which meant that linux jobs depending on the build needed to wait for the macOS build to finish as well
* while the macOS runners where changed to `macos-12-xl`, this is more for future-readiness,
  as the workflow is still picking up normal runners. When and if we move to `xl` runners, we can
  reconsider the check architecture, maybe moving the build jobs back into a matrix
* There was a bit of further job optimization by using cache actions to store a repository and disk cache for Bazel

[1]: https://github.com/dorny/paths-filter#examples
[2]: https://github.com/actions/upload-artifact/issues/38

Here's a peek at the full summary view:
image

Last run totalled at 20 minutes total, which is an improvement over current feedback time (consider this is with the newly added macOS integration test job). macOS runners are still a bottleneck, we would probably get blazing fast feedback if we managed to get `macos-12-xl` runners. In any case most of the time we get negative feedback before that time because of the Linux tests.


.github/workflows/swift.yml
```diff
@@ -0,0 +1,121 @@
...
+      - main
+
+jobs:
+  changes:
+    runs-on: ubuntu-latest
+    outputs:
+      codegen: ${{ steps.filter.outputs.codegen }}
+      ql: ${{ steps.filter.outputs.ql }}
+    steps:
+      - uses: dorny/paths-filter@v2
```
adityasharad, line 25: Since you're using a third-party Action, I recommend pinning to a specific commit SHA.

redsun82: done

AlexDenisov (approved):

PR #11009, opened 2022-10-27
Misc: Add automatic `DataFlow Library` label

RasmusWL:
Would just be nice for filtering PRs

### DIFF
--- a/.github/labeler.yml
+++ b/.github/labeler.yml
@@ -43,3 +43,11 @@ documentation:
 "QL-for-QL":
   - ql/**/*
   - .github/workflows/ql-for-ql*
+
+# since these are all shared files that needs to be synced, I just picked _one_ version
+"DataFlow Library":
+  - "java/ql/lib/semmle/code/java/dataflow/internal/DataFlowImpl.qll"
+  - "java/ql/lib/semmle/code/java/dataflow/internal/DataFlowImplCommon.qll"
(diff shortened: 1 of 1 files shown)


.github/labeler.yml
```diff
@@ -43,3 +43,11 @@ documentation:
 "QL-for-QL":
   - ql/**/*
   - .github/workflows/ql-for-ql*
+
+# since these are all shared files that needs to be synced, I just picked _one_ version
```
jketema, line 47: Grammar s/needs/need/ and getting rid of the "I", which is unclear unless you look at the commit history.
```suggestion
# Since these are all shared files that need to be synced, just pick _one_ copy of each.
```

jketema (commented):
LGTM and makes sense to me. One suggestion. If that suggestion is ok with you and since there are no code owners for this file, I'm guessing me approving it is as good as anyone else.

.github/labeler.yml
RasmusWL: thanks

jketema (approved):

RasmusWL:
> LGTM and makes sense to me. One suggestion. If that suggestion is ok with you and since there are no code owners for this file, I'm guessing me approving it is as good as anyone else.

agreed :+1: (and if someone is mad/sad about the effect of this PR, we can always change things)

calumgrant:
👍 I have no objections

PR #11012, opened 2022-10-27
Kotlin: Fix integration tests on Mac

igfoo:


smowton (approved):

smowton:
Looks like java-interface-redeclares-tostring has the same problem. This isn't mac-specific fwiw since I develop on a Mac, but it clearly is particular to some build of javac that does or doesn't create a non-existent output directory.

smowton (approved):

PR #11015, opened 2022-10-27
Go: exclude protobuf read steps from cleartext-logging query

smowton:
This query already treats structs differently to usual: it includes field -> whole struct taint steps, but explicitly excludes struct -> field steps. This means that a logging framework sinking an entire struct with a tainted field yields an alert, but we don't get FPs caused by writing field `x` but then reading field `y`.

However, protobuf messages have a special treatment, with taint usually associated with the whole struct and getter methods propagating that taint out. Suppressing these getter method steps specifically for the cleartext-logging query mirrors its treatment of structs in general and avoids this sort of field-mismatch FP.

On the downside we will miss same-field propagation like `m.field = password; Log(m.GetField())` if we don't have source code for the implementation of `m`. However this is hopefully unusual since the typical use of protobufs is to serialize and deserialize, rather than using the struct as a general-purpose datastructure.


owen-mc (approved):
Why has `ProtoReflectMethod` been changed to a `TaintTracking::FunctionModel`? I'm guessing we don't want those methods to be considered for this particular query. Are there any side effects of making these two classes `TaintTracking::FunctionModel` instead of `DataFlow::FunctionModel`

smowton:
Just noticed while I was there that since these export taint from a whole struct to a member, they're definitely not identity-preserving steps. It doesn't have an effect on this query.

PR #11017, opened 2022-10-27
Kotlin: fix wildcard suppression where the annotation applies to a parent type/argument.

smowton:
In the process I also fix the missed case where suppression can be switched off using a parameterized annotation.


java/kotlin-extractor/src/main/kotlin/KotlinUsesExtractor.kt
```diff
@@ -1008,12 +1015,13 @@ open class KotlinUsesExtractor(
 
     fun addJavaLoweringWildcards(t: IrType, addByDefault: Boolean, javaType: JavaType?): IrType =
         (t as? IrSimpleType)?.let {
+            val newAddByDefault = getWildcardSuppressionDirective(t)?.let { b -> !b } ?: addByDefault
```
tamasvajk, line 1018: Is `?.let { b -> !b}` the same as `?.not()`? The latter feels easier to read.

smowton: Done

igfoo (dismissed):

igfoo (approved):

PR #11018, opened 2022-10-27
Kotlin: specialise extension receivers the same as other function parameters

smowton:
This arises when a generic class extends one of its parameters; for example, `class G<T> { val T.v; get() = 1 }`, where specialisation `G<List>` should generate a method specialisation `getV(List)`.

### DIFF
--- a/java/kotlin-extractor/src/main/kotlin/KotlinFileExtractor.kt
+++ b/java/kotlin-extractor/src/main/kotlin/KotlinFileExtractor.kt
@@ -1188,7 +1188,7 @@ open class KotlinFileExtractor(
                             id
 
                 val extReceiver = f.extensionReceiverParameter
-                val fParameters = (overriddenAttributes?.valueParameters ?: f.valueParameters) + listOfNotNull(extReceiver)
+                val fParameters = listOfNotNull(extReceiver) + (overriddenAttributes?.valueParameters ?: f.valueParameters)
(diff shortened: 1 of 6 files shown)


java/kotlin-extractor/src/main/kotlin/KotlinFileExtractor.kt
```diff
@@ -1188,22 +1188,16 @@ open class KotlinFileExtractor(
                             id
 
                 val extReceiver = f.extensionReceiverParameter
-                val idxOffset = if (extReceiver != null) 1 else 0
-                val fParameters = overriddenAttributes?.valueParameters ?: f.valueParameters
+                val fParameters = (overriddenAttributes?.valueParameters ?: f.valueParameters) + listOfNotNull(extReceiver)
```
tamasvajk, line 1191: Is this correct? The extension receiver is added to the end of the list and below we get the first item in `extendedType = paramTypes[0]`. I think some of the failing tests might also be due to this mixup in the order..

tamasvajk (changes requested):

java/kotlin-extractor/src/main/kotlin/KotlinFileExtractor.kt
smowton: Doh, fixed.

igfoo (approved):

tamasvajk (approved):

PR #11020, opened 2022-10-27
C++: repair Adding365DaysPerYear.ql

rdmarsh2:


MathiasVP (approved):

MathiasVP:
I think the CI failure was due to an old base branch, but just to make sure I've merged in the latest `main`. I'll go ahead and merge it if the test failures are fixed.

cpp/ql/src/Likely Bugs/Leap Year/LeapYear.qll
```diff
@@ -248,24 +248,27 @@ class FiletimeYearArithmeticOperationCheckConfiguration extends DataFlow::Config
 /**
  * `DataFlow::Configuration` for finding an operation with hardcoded 365 that will flow into any known date/time field.
```
geoffw0, line 249:
```suggestion
 * Taint configuration for finding an operation with hardcoded 365 that will flow into any known date/time field.
```

MathiasVP: Oh, good catch!

PR #11022, opened 2022-10-27
Ruby: try/try! as code execution

hmac:
Recognise `Object#try(!)` as a code execution. These methods are ActiveSupport extensions. ~~[Evaluation](https://github.com/github/codeql-dca-main/issues/8160) shows a slow-down for `canvas-lms`, but I think that's because it contains a lot of calls to `try` (848 vs 389 for `gitlabhq`).~~  [New evaluation](https://github.com/github/codeql-dca-main/issues/8160) seems fine.

### DIFF
--- a/ruby/ql/lib/change-notes/2022-10-28-try-code-execution.md
+++ b/ruby/ql/lib/change-notes/2022-10-28-try-code-execution.md
@@ -1,0 +1,4 @@
+---
+category: minorAnalysis
+---
+* The `ActiveSupport` extensions `Object#try` and `Object#try!` are now recognised as code executions.


ruby/ql/lib/codeql/ruby/frameworks/ActiveSupport.qll
```diff
@@ -81,6 +81,29 @@ module ActiveSupport {
...
+       * ```rb
+       * x = "abc"
+       * x.try(:upcase) # => "ABC"
+       * y = nil
+       * y.try(:upcase) # => nil
+       * ```
+       * `Object#try!` behaves similarly but raises `NoMethodError` if the
+       * receiver is not `nil` and does not respond to the method.
+       */
+      class TryCallCodeExecution extends CodeExecution::Range, DataFlow::CallNode {
```
erik-krogh, line 97:
```suggestion
      class TryCallCodeExecution extends CodeExecution::Range instanceof DataFlow::CallNode {
```

erik-krogh (approved):
LGTM 👍  

Optional cosmetic suggestion.

PR #10918, opened 2022-10-21
Ruby: handle compound constant-assignment

asgerf:
Kudos to @aibaars for the implementation (pushed directly to this PR).

Constants can be used in compound assignments but were not previously desugared, creating holes in the CFG:
```rb
MyConst ||= 123
```

PR #11024, opened 2022-10-28
Ruby: expand DataFlow API

asgerf:
Adds some classes to do basic stuff with `DataFlow` nodes without having to map down to CFG and AST nodes, and a few other tidbits I found myself missing.

To ensure some degree of battle-hardening I (mostly) ported three library models to use the new nodes where it makes sense. The affected models are `Rails.qll`, `Railties.qll`, and `ActionController.qll`.

Migrating from AST nodes to DataFlow nodes revealed a couple of pre-existing issues, some of which have now been fixed or fixed enough to avoid a regression:
- https://github.com/github/codeql/pull/10928
- https://github.com/github/codeql/pull/10918

Some other issues which were too hard to fix prior to opening this PR:
- We're missing flows into captured variables in some cases, in particular when returning a lambda. I attempted a fix in [this draft PR](https://github.com/github/codeql/pull/11010) which would align closer with how we do things in JS, but after a whole day of debugging strange call graph changes, I've put that aside for now. Instead, I've resorted to a rather unambitious workaround in https://github.com/github/codeql/pull/11024/commits/ab91b5772bcfb72084dd83523de1abac43f24942.
- Blocks are assumed to capture `self`, but of course that's not always the case. There is currently no way to tell the analysis what the value of `self` is, so we have to work around it for now. This means using things like `getParent+()` and treating the module `self` as if it's an instance of the module.

### getConst() and what about API graphs?

One of the changes that may look odd is the introduction of `DataFlow::getConst()` and `ConstRef`. It provides convenient access to constants that may resolve to a given qualified name. For example, the incantation
```codeql
DataFlow::getConst("ActionController").getConst("Base").getADescendentModule()
```
gives us all descendents of `ActionController::Base`.

This was probably the biggest use case for API graphs in Ruby, as you can get to (almost) the same thing with `API::getTopLevelMember("ActionController").getMember("Base")`. So why add `getConst()`?

- `getConst()` doesn't rely on the call graph, which means it can be used for things like `getACallSimple()` and other pre-call graph customizations (we have quite a few of those in JS and I expect we'll find the need in Ruby as well).
- I intend to share more of the API graphs implementation between languages (stay tuned!), but the rules for constant lookup in Ruby are pretty language-specific, so it would be nice to have some of this factored out anyway.
- `getConst()` handles the rules of constant lookup more precisely than API graphs, in particular when related to `include`, nested modules, and modules with more than one declaration. I think API graphs could be extended to make this work, but it's not really a natural fit for its use/def tracking, it would need to be a bunch of new predicates specific to constant lookup, which might as well exist earlier in the pipeline.

I replaced some uses of API graphs with `getConst`, not because it matters in practice, I just had to ensure `getConst()` is being used enough to test it in the wild.

I hope the name `getConst()` makes more sense in the context of Ruby. I think the name in API graphs, `getMember`, was used mostly because that's what JS called it.

### Evaluation

[Evaluation](https://github.com/github/codeql-dca-main/issues/8214) shows
- 100 new taint sinks
- 491 new taint sources, resulting in 1890 new tainted expressions
- 3 new call edges due to the change in `flowsTo`
- There seems to be a slight performance hit, which I might investigate a bit further, but I wanted to get the PR up for now.

The new sources and sinks are mainly due to resolving methods upwards in the hierarchy in `ActionController`. The combination `getADescendentModule().getAnInstanceMethod()` takes "sideways" method resolution into account without having to really think about it, so I like that.

### DIFF
--- a/ruby/ql/lib/codeql/ruby/ast/Constant.qll
+++ b/ruby/ql/lib/codeql/ruby/ast/Constant.qll
@@ -170,6 +170,24 @@ module ConstantValue {
 
   /** A constant `nil` value. */
   class ConstantNilValue extends ConstantValue, TNil { }
+
+  /** Gets the integer constant `x`. */
+  ConstantValue getInt(int x) { result.getInt() = x }
+
+  /** Gets the float constant `x`. */
+  ConstantValue getFloat(float x) { result.getFloat() = x }
+
+  /** Gets the string constant `x`. */
+  ConstantValue getString(string x) { result.getString() = x }
+
+  /** Gets the symbol constant `x`. */
+  ConstantValue getSymbol(string x) { result.getSymbol() = x }
+
+  /** Gets the regexp constant `x`. */
+  ConstantValue getRegExp(string x) { result.getRegExp() = x }
+
+  /** Gets the string, symbol, or regexp constant `x`. */
+  ConstantValue getStringlikeValue(string x) { result.getStringlikeValue() = x }
 }
 
 /** An access to a constant. */
--- a/ruby/ql/lib/codeql/ruby/ast/Constant.qll
+++ b/ruby/ql/lib/codeql/ruby/ast/Constant.qll
@@ -172,22 +172,22 @@ module ConstantValue {
   class ConstantNilValue extends ConstantValue, TNil { }
 
   /** Gets the integer constant `x`. */
-  ConstantValue getInt(int x) { result.getInt() = x }
+  ConstantValue fromInt(int x) { result.getInt() = x }
 
   /** Gets the float constant `x`. */
-  ConstantValue getFloat(float x) { result.getFloat() = x }
+  ConstantValue fromFloat(float x) { result.getFloat() = x }
 
   /** Gets the string constant `x`. */
-  ConstantValue getString(string x) { result.getString() = x }
+  ConstantValue fromString(string x) { result.getString() = x }
 
   /** Gets the symbol constant `x`. */
-  ConstantValue getSymbol(string x) { result.getSymbol() = x }
+  ConstantValue fromSymbol(string x) { result.getSymbol() = x }
 
   /** Gets the regexp constant `x`. */
-  ConstantValue getRegExp(string x) { result.getRegExp() = x }
+  ConstantValue fromRegExp(string x) { result.getRegExp() = x }
 
   /** Gets the string, symbol, or regexp constant `x`. */
-  ConstantValue getStringlikeValue(string x) { result.getStringlikeValue() = x }
+  ConstantValue fromStringlikeValue(string x) { result.getStringlikeValue() = x }
 }
 
 /** An access to a constant. */
--- a/ruby/ql/lib/codeql/ruby/ast/Module.qll
+++ b/ruby/ql/lib/codeql/ruby/ast/Module.qll
@@ -23,6 +23,20 @@ class Module extends TModule {
   /** Gets an `include`d module. */
   Module getAnIncludedModule() { result = getAnIncludedModule(this) }
 
+  /** Gets the super class or an included or prepended module. */
+  Module getADirectAncestor() {
+    result = [this.getSuperClass(), this.getAPrependedModule(), this.getAnIncludedModule()]
+  }
+
+  /** Gets a direct subclass or module including or prepending this one. */
+  Module getADirectDescendent() { this = result.getADirectAncestor() }
+
+  /** Gets a module that is transitively subclassed, included, or prepended by this module. */
+  Module getAnAncestor() { result = this.getADirectAncestor*() }
+
+  /** Gets a module that transitively subclasses, includes, or prepends this module. */
+  Module getADescendent() { result = this.getADirectDescendent*() }
+
   /** Holds if this module is a class. */
   pragma[noinline]
   predicate isClass() {
--- a/ruby/ql/lib/codeql/ruby/ast/Module.qll
+++ b/ruby/ql/lib/codeql/ruby/ast/Module.qll
@@ -33,9 +33,11 @@ class Module extends TModule {
   Module getAnImmediateDescendent() { this = result.getAnImmediateAncestor() }
 
   /** Gets a module that is transitively subclassed, included, or prepended by this module. */
+  pragma[inline]
   Module getAnAncestor() { result = this.getAnImmediateAncestor*() }
 
   /** Gets a module that transitively subclasses, includes, or prepends this module. */
+  pragma[inline]
   Module getADescendent() { result = this.getAnImmediateDescendent*() }
 
   /** Holds if this module is a class. */
--- a/ruby/ql/lib/codeql/ruby/ast/Module.qll
+++ b/ruby/ql/lib/codeql/ruby/ast/Module.qll
@@ -3,6 +3,7 @@ private import codeql.ruby.CFG
 private import internal.AST
 private import internal.Module
 private import internal.TreeSitter
+private import internal.Scope
 
 /**
  * A representation of a run-time `module` or `class` value.
@@ -262,6 +263,31 @@ class ModuleBase extends BodyStmt, Scope, TModuleBase {
     or
     not this instanceof Namespace and
     result = this.getEnclosingModule().getNamespaceOrToplevel()
+  }
+
+  /**
+   * Gets an expression denoting the super class or an included or prepended module.
+   *
+   * For example, `C` is an ancestor expression of `M` in each of the following examples:
+   * ```rb
+   * class M < C
+   * end
+   *
+   * module M
+   *   include C
+   *   prepend C
+   * end
+   * ```
+   */
+  Expr getAnAncestorExpr() {
+    exists(MethodCall call |
+      call.getReceiver().(SelfVariableAccess).getVariable() = this.getModuleSelfVariable() and
+      call.getMethodName() = ["include", "prepend"] and
+      result = call.getArgument(0) and
+      scopeOfInclSynth(call) = this // only permit calls directly in the module scope, not in a block
+    )
+    or
+    result = this.(ClassDeclaration).getSuperclassExpr()
   }
 }
 
--- a/ruby/ql/lib/codeql/ruby/ast/Module.qll
+++ b/ruby/ql/lib/codeql/ruby/ast/Module.qll
@@ -241,6 +241,14 @@ class ModuleBase extends BodyStmt, Scope, TModuleBase {
    * Does not include the `self` variable from any of the methods in the module.
    */
   SelfVariable getModuleSelfVariable() { result.getDeclaringScope() = this }
+
+  /** Gets the nearest enclosing `Namespace` or `Toplevel`, possibly this module itself. */
+  Namespace getNamespaceOrToplevel() {
+    result = this
+    or
+    not this instanceof Namespace and
+    result = this.getEnclosingModule().getNamespaceOrToplevel()
+  }
 }
 
 /**
--- a/ruby/ql/lib/codeql/ruby/ast/Module.qll
+++ b/ruby/ql/lib/codeql/ruby/ast/Module.qll
@@ -118,6 +118,44 @@ class Module extends TModule {
     or
     result.(SelfVariableAccess).getVariable().getDeclaringScope() = this.getAnOwnSingletonMethod()
   }
+
+  pragma[nomagic]
+  private string getEnclosingModuleName() {
+    exists(string qname |
+      qname = this.getQualifiedName() and
+      result = qname.regexpReplaceAll("::[^:]*$", "") and
+      qname != result
+    )
+  }
+
+  pragma[nomagic]
+  private string getOwnModuleName() {
+    result = this.getQualifiedName().regexpReplaceAll("^.*::", "")
+  }
+
+  /**
+   * Gets the enclosing module, as it appears in the qualified name of this module.
+   *
+   * For example, the canonical enclosing module of `A::B` is `A`, and `A` itself has no canonical enclosing module.
+   */
+  pragma[nomagic]
+  Module getCanonicalEnclosingModule() { result.getQualifiedName() = this.getEnclosingModuleName() }
+
+  /**
+   * Gets a module named `name` declared inside this one (not aliased), provided
+   * that such a module is defined or reopened in the current codebase.
+   *
+   * For example, for `A::B` the canonical nested module named `C` would be `A::B::C`.
+   *
+   * Note that this is not the same as constant lookup. If `A::B::C` would resolve to a
+   * module whose qualified name is not `A::B::C`, then it will not be found by
+   * this predicate.
+   */
+  pragma[nomagic]
+  Module getCanonicalNestedModule(string name) {
+    result.getCanonicalEnclosingModule() = this and
+    result.getOwnModuleName() = name
+  }
 }
 
 /**
--- a/ruby/ql/lib/codeql/ruby/ast/Module.qll
+++ b/ruby/ql/lib/codeql/ruby/ast/Module.qll
@@ -25,18 +25,18 @@ class Module extends TModule {
   Module getAnIncludedModule() { result = getAnIncludedModule(this) }
 
   /** Gets the super class or an included or prepended module. */
-  Module getADirectAncestor() {
+  Module getAnImmediateAncestor() {
     result = [this.getSuperClass(), this.getAPrependedModule(), this.getAnIncludedModule()]
   }
 
   /** Gets a direct subclass or module including or prepending this one. */
-  Module getADirectDescendent() { this = result.getADirectAncestor() }
+  Module getAnImmediateDescendent() { this = result.getAnImmediateAncestor() }
 
   /** Gets a module that is transitively subclassed, included, or prepended by this module. */
-  Module getAnAncestor() { result = this.getADirectAncestor*() }
+  Module getAnAncestor() { result = this.getAnImmediateAncestor*() }
 
   /** Gets a module that transitively subclasses, includes, or prepends this module. */
-  Module getADescendent() { result = this.getADirectDescendent*() }
+  Module getADescendent() { result = this.getAnImmediateDescendent*() }
 
   /** Holds if this module is a class. */
   pragma[noinline]
--- a/ruby/ql/lib/codeql/ruby/ast/Module.qll
+++ b/ruby/ql/lib/codeql/ruby/ast/Module.qll
@@ -63,6 +63,33 @@ class Module extends TModule {
           loc.getStartColumn()
       )
   }
+
+  /** Gets a constant or `self` access that refers to this module. */
+  private Expr getAnImmediateReferenceBase() {
+    resolveConstantReadAccess(result) = this
+    or
+    result.(SelfVariableAccess).getVariable() = this.getADeclaration().getModuleSelfVariable()
+  }
+
+  /** Gets a singleton class that augments this module object. */
+  SingletonClass getASingletonClass() { result.getValue() = this.getAnImmediateReferenceBase() }
+
+  /**
+   * Gets a singleton method on this module, either declared as a singleton method
+   * or an instance method on a singleton class.
+   */
+  MethodBase getASingletonMethod() {
+    result.(SingletonMethod).getObject() = this.getAnImmediateReferenceBase()
+    or
+    result = this.getASingletonClass().getAMethod().(Method)
+  }
+
+  /** Gets a constant or `self` access that refers to this module. */
+  Expr getAnImmediateReference() {
+    result = this.getAnImmediateReferenceBase()
+    or
+    result.(SelfVariableAccess).getVariable().getDeclaringScope() = this.getASingletonMethod()
+  }
 }
 
 /**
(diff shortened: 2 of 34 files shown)


ruby/ql/lib/codeql/ruby/ast/Constant.qll
```diff
@@ -170,6 +170,24 @@ module ConstantValue {
 
   /** A constant `nil` value. */
   class ConstantNilValue extends ConstantValue, TNil { }
+
+  /** Gets the integer constant `x`. */
+  ConstantValue getInt(int x) { result.getInt() = x }
```
aibaars, line 175: It feels a bit strange that the "inverse" of `getInt` is named `getInt` . Wouldn't `ConstantValue::fromInt` be a better name?

ruby/ql/lib/codeql/ruby/ast/Module.qll
```diff
@@ -23,6 +24,20 @@ class Module extends TModule {
   /** Gets an `include`d module. */
   Module getAnIncludedModule() { result = getAnIncludedModule(this) }
 
+  /** Gets the super class or an included or prepended module. */
```
aibaars, line 27: This change is very similar to https://github.com/github/codeql/pull/10747/files#diff-9333de60c5c1b8283b484ae3b7e3cbef24b7bf52aee695df420d0e3f4d55403b 

The main difference is that you use `*` instead of `+` which would make all modules ancestors of themselves, which is odd (even though it is handy in some cases).

@hvitved suggested a `pragma[inline]` annotation on `getAnAncestor` , perhaps you should do that as well.

ruby/ql/lib/codeql/ruby/ast/Module.qll
```diff
@@ -63,6 +78,99 @@ class Module extends TModule {
...
+    result = this.getQualifiedName().regexpReplaceAll("^.*::", "")
+  }
+
+  /**
+   * Gets the enclosing module, as it appears in the qualified name of this module.
+   *
+   * For example, the canonical enclosing module of `A::B` is `A`, and `A` itself has no canonical enclosing module.
+   */
+  pragma[nomagic]
+  Module getCanonicalEnclosingModule() { result.getQualifiedName() = this.getEnclosingModuleName() }
```
aibaars, line 157: I'm not sure I understand what a "canonical" enclosing module means. Could you give an example of a non-canonical enclosing module?

ruby/ql/lib/codeql/ruby/dataflow/internal/DataFlowPublic.qll
```diff
@@ -174,6 +225,82 @@ class LocalSourceNode extends Node {
...
+
+  /**
+   * Gets a value stored as an element on this node, such as the `x` in `obj[key] = x`.
+   *
+   * Concretely, this gets the second argument from any call to `[]=` where this node flows to the receiver.
+   */
+  Node getAnElementWriteValue() {
+    exists(CallNode call |
+      call = this.getAMethodCall("[]=") and
+      call.getNumberOfArguments() = 2 and
```
aibaars, line 285: You can have assignments like
```
obj[a, b, c] = x
```
I'm not sure we want to capture such things in this predicate though. Just wanted to make sure you ignored such calls on purpose.

aibaars (commented):
Partial review.

ruby/ql/lib/codeql/ruby/ast/Constant.qll
asgerf: `fromInt` sounds good to me 👍

ruby/ql/lib/codeql/ruby/ast/Module.qll
asgerf: > which would make all modules ancestors of themselves, which is odd (even though it is handy in some cases)

I think it's quite common to define things like this, for example, a type is usually considered a subtype of itself. Also a module is technically on its own ancestor chain.
```rb
module A
end
A.ancestors # [A]
```

I also tend to prefer `*` because you can easily get the `+` variant by doing `getADirectAncestor().getAnAncestor()`. 

However, if the predicate only exposes the `+` variant, you have to do something like `[x, x.getAnAncestor()]` to get the `*` variant, and even that doesn't work if `x` has type `Class`. Either way it can't appear in the middle of an expression chain, which is important iMO.

ruby/ql/lib/codeql/ruby/ast/Module.qll
asgerf: A module can aliased, i.e. reachable via more than one chain of constants. One of these chains of constants is considered to be the "canonical" name, the one used in the `TResolved` constructor, and the rest are just aliases. This predicate gets the enclosing module _as it appears in the canonical name_.

For the module `A::X` below, the module `A` would be its canonical enclosing module.
```rb
module A
end
module B
  module A::X
  end
end
```

ruby/ql/lib/codeql/ruby/dataflow/internal/DataFlowPublic.qll
asgerf: It's not intended to match anything that wouldn't work for `Array` or `Hash`.

ruby/ql/lib/codeql/ruby/ast/Module.qll
aibaars: You're right. I somehow thought that `A.ancestors` would not be reflexive.

ruby/ql/lib/codeql/ruby/ast/Module.qll
aibaars: The problem is that the "canonical" name in `TResolved`  is rather arbitrary and does not always reflect what the Ruby interpreter would report as the "name" . So it may be a bit misleading. In the interpreter a module's name appears to be determined by the name of the first constant to which it is assigned.

ruby/ql/lib/codeql/ruby/dataflow/internal/DataFlowPublic.qll
```diff
@@ -559,3 +710,618 @@ abstract deprecated class BarrierGuard extends CfgNodes::ExprCfgNode {
     result.asExpr() = this.getAMaybeGuardedCapturedDef().getARead()
   }
 }
+
+/**
+ * A representation of a run-time module or class.
+ *
+ * This is equivalent to the type `Ast::Module` but provides data-flow specific methods.
+ */
+class ModuleNode instanceof Module {
```
nickrolfe, line 719: Can you explain your thinking behind making this an AST node and not a `DataFlow::Node`? I note that JS's `DataFlow::ClassNode` extends `DataFlow::Node`.

nickrolfe (commented):
Not a thorough review, but a few things I noticed. In general, I think this is a great change that should let us simplify a lot of awkward code.

Before this is merged, please make sure to add a release note describing not only the new classes, but the deprecation of `ActionControllerControllerClass`.

ruby/ql/lib/codeql/ruby/ast/Module.qll
asgerf: Is there another name we could use for this concept that you'd find more intuitive?

aibaars: That is a very good question. I couldn't find a name that is used in the Ruby documentation or implementation for the "parent" or "container" of a module/class.

There does not seem to be any reflection method to get that information in the standard library (apart from parsing the value of `Module.name`). However, there is a method named `module_parent` in Rails: https://apidock.com/rails/v6.0.0/Module/module_parent

So perhaps that is the best name for this concept?

ruby/ql/lib/codeql/ruby/ast/Constant.qll
aibaars: Same of the other getters, of course.

asgerf: Renamed to `fromInt` etc

ruby/ql/lib/codeql/ruby/ast/Module.qll
asgerf: Pushed a commit to rename:
- `getCanonicalEnclosingModule` -> `getParentModule`
- `getCanonicalNestedModule` -> `getNestedModule`

asgerf:
I've pushed 0f1b348, which [fixed the regression in metasploit](https://github.com/github/codeql-dca-main/issues/8229).

nickrolfe:
I've been doing some Rails modelling based on this branch, and I'm wondering about this scenario:

```rb
class Base
end

class Foo < Base
end

class Bar < Foo
end
```

`DataFlow::getConst("Base").getADescendentModule()` gives me `Base`, `Foo`, and `Bar`, but I couldn't see an obvious way of excluding `Base` to get only `Foo` and `Bar`.

(I'm trying to find methods on classes derived from a Rails base class, but I don't want to find methods on that base class itself when we're analysing `rails/rails`).

asgerf:
@nickrolfe on that snippet it only gives me `Foo` and `Bar`. I'm guessing there must be something else going on in `rails/rails` that's causing this behaviour?

nickrolfe:
Ah. Thanks, I'll do some digging into Rails.

ruby/ql/lib/codeql/ruby/dataflow/internal/DataFlowPublic.qll
```diff
@@ -559,3 +710,609 @@ abstract deprecated class BarrierGuard extends CfgNodes::ExprCfgNode {
...
+   * this predicate.
+   */
+  ModuleNode getCanonicalNestedModule(string name) { result = super.getNestedModule(name) }
+}
+
+/**
+ * A representation of a run-time class.
+ */
+class ClassNode extends ModuleNode {
+  ClassNode() { isClass() }
```
hvitved, line 949: `this.`

ruby/ql/lib/codeql/ruby/dataflow/internal/DataFlowPublic.qll
```diff
@@ -559,3 +710,609 @@ abstract deprecated class BarrierGuard extends CfgNodes::ExprCfgNode {
...
+  /** Gets the value expression of this pair, such as the `V` in `K => V` or `K: V`. */
+  Node getValue() { this.hasKeyAndValue(_, result) }
+}
+
+/**
+ * A data-flow node that corresponds to a hash literal. Hash literals are desugared
+ * into calls to `Hash.[]`, so this includes both desugared calls as well as
+ * explicit calls.
+ */
+class HashLiteralNode extends LocalSourceNode, ExprNode {
```
hvitved, line 1069: `CallNode` instead of `ExprNode`?

ruby/ql/lib/codeql/ruby/dataflow/internal/DataFlowPublic.qll
```diff
@@ -559,3 +710,609 @@ abstract deprecated class BarrierGuard extends CfgNodes::ExprCfgNode {
...
+  /**
+   * Gets an element of the array.
+   */
+  Node getAnElement() { result = this.(CallNode).getPositionalArgument(_) }
+}
+
+/**
+ * A place in which a named constant can be found up during constant lookup.
+ */
+private newtype TConstLookupScope =
```
hvitved, line 1104: cache?

ruby/ql/lib/codeql/ruby/dataflow/internal/DataFlowPublic.qll
```diff
@@ -559,3 +710,609 @@ abstract deprecated class BarrierGuard extends CfgNodes::ExprCfgNode {
...
+    not access.hasGlobalScope() and
+    (
+      result = MkAncestorLookup(access.getEnclosingModule().getNamespaceOrToplevel().getModule())
+      or
+      result = MkNestedLookup(access.getEnclosingModule().getEnclosingModule*().getModule())
+    )
+  }
+
+  /**
+   * Holds if this can reference a constant named `name` from `scope` using a lookup of `kind`.
```
hvitved, line 1223: ql doc is out of sync

hvitved (changes requested):
Initial review; overall LGTM.

ruby/ql/lib/codeql/ruby/dataflow/internal/DataFlowPrivate.qll
```diff
@@ -109,11 +109,27 @@ module LocalFlow {
...
     nodeTo = getSelfParameterDefNode(nodeFrom.(SelfParameterNode).getMethod())
   }
 
+  /**
+   * Holds if `nodeFrom -> nodeTo` is a step from a parameter to a capture entry node for
+   * that parameter.
+   *
+   * This is intended to recover from flow not currently recognised by ordinary capture flow.
+   */
+  predicate localFlowSsaParamCaptureInput(Node nodeFrom, Node nodeTo) {
```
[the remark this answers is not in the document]
asgerf: The underlying bug is that SSA doesn't consider a lambda expression to escape when returned, as highlighted by the added test case. As mentioned in the PR description, the code here is a workaround to avoid regression in the short-term, and I'm not really interested in generalizing the workaround. I initially tried landing such fixes prior to opening this PR, but after a while I didn't want to block the PR any longer.

Obviously the bug in SSA should be fixed, but at the architectural level I don't think this layer ought care about it. I'd like to post-pone flow-sensitive handling of captured variables until after the call graph, so the layers would look something like:
- Materialize data-flow nodes.
- Compute flow-insensitive step relation, treating captured variables flow-insensitively (basically SSA only used for true locals)
- Compute call graph, using the flow-insensitive relation for propagating type information
- Compute flow-sensitive step relation (for use by DataFlow::Configuration etc). This stage can then in principle benefit from seeing the call graph.

Flow sensitivity doesn't matter very much for propagating type information, whereas it matters a lot more to taint tracking and the like. Using a single handling of captured variables to cover both isn't optimal.

I tried to make a start on this [here](https://github.com/github/codeql/pull/11010) but as mentioned, I had to shelf it to make some progress on this PR.

ruby/ql/lib/codeql/ruby/dataflow/internal/DataFlowPrivate.qll
```diff
@@ -502,6 +528,38 @@ class SsaSelfDefinitionNode extends LocalSourceNode, SsaDefinitionNode {
...
+  override Location getLocationImpl() { result = namespace.getLocation() }
+
+  override string toStringImpl() { result = namespace.toString() }
+}
+
+/**
+ * Gets a representative for the created or re-opened class/module in a `Namespace`.
+ *
+ * This is usually the SSA definition for `self`, but if this node does not exist due to lack of liveness,
+ * it is the `RawNamespaceSelf` node.
```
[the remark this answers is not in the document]
asgerf: I haven't been able to figure out how get the last CFG node in a `Namespace`, but if you can write up the relevant changes to the SSA library I'm happy to include it in the PR and then remove the new data-flow node.

ruby/ql/test/library-tests/dataflow/global/captured_variables.rb
```diff
@@ -0,0 +1,38 @@
+def capture_local_call x
```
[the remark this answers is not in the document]
asgerf: The test includes calls to unknown methods, which makes it a bit more tricky to make the test runnable.

ruby/ql/test/library-tests/dataflow/global/captured_variables.rb
```diff
@@ -0,0 +1,38 @@
...
+    fn = -> { sink(x) } # $ hasValueFlow=1.1
+    fn.call
+end
+capture_local_call source(1.1)
+
+def capture_escape_return1 x
+    -> {
+        sink(x) # $ MISSING: hasValueFlow=1.2
```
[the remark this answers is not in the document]
asgerf: AFAICT it's an issue with capture steps in SSA

ruby/ql/lib/codeql/ruby/dataflow/internal/DataFlowPublic.qll
```diff
@@ -559,3 +710,609 @@ abstract deprecated class BarrierGuard extends CfgNodes::ExprCfgNode {
...
+      this.getAnOwnInstanceVariableAccess(name).(InstanceVariableReadAccess)
+  }
+
+  /**
+   * Gets the right-hand side of an assignment to the instance variable `name` in this module.
+   *
+   * Does not take inheritance into account.
+   */
+  Node getAnOwnInstanceVariableWriteValue(string name) {
+    exists(Assignment assignment |
```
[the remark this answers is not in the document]
asgerf: Yes, thanks for catching that

ruby/ql/lib/codeql/ruby/dataflow/internal/DataFlowPublic.qll
```diff
@@ -559,3 +710,609 @@ abstract deprecated class BarrierGuard extends CfgNodes::ExprCfgNode {
...
+  }
+
+  /**
+   * Gets the canonical return node from this callable.
+   *
+   * Each callable has exactly one such node, and its location may not correspond
+   * to any particular return site - consider using `getAReturningNode` to get nodes
+   * whose locations correspond to return sites.
+   */
+  Node getReturn() { result.(SynthReturnNode).getCfgScope() = callable }
```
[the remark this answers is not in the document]
asgerf: We do this in JavaScript so I don't see why we'd need to hide it in Ruby.

ruby/ql/lib/codeql/ruby/dataflow/internal/DataFlowPublic.qll
```diff
@@ -559,3 +710,609 @@ abstract deprecated class BarrierGuard extends CfgNodes::ExprCfgNode {
...
+ * into calls to `Array.[]`, so this includes both desugared calls as well as
+ * explicit calls.
+ */
+class ArrayLiteralNode extends LocalSourceNode, ExprNode {
+  ArrayLiteralNode() { super.getExprNode() instanceof CfgNodes::ExprNodes::ArrayLiteralCfgNode }
+
+  /**
+   * Gets an element of the array.
+   */
+  Node getAnElement() { result = this.(CallNode).getPositionalArgument(_) }
```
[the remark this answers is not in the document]
asgerf: Not being a `CallNode` was deliberate. It exposes member predicates that aren't necessarily helpful or do what you expect when thinking about a hash/array literal.

ruby/ql/lib/codeql/ruby/dataflow/internal/DataFlowPublic.qll
asgerf: same

ruby/ql/lib/codeql/ruby/dataflow/internal/DataFlowPublic.qll
```diff
@@ -559,3 +710,609 @@ abstract deprecated class BarrierGuard extends CfgNodes::ExprCfgNode {
...
+   *
+   * class C < B
+   * end
+   *
+   * module E
+   *   include C
+   * end
+   * ```
+   */
+  ModuleNode getADescendentModule() { MkAncestorLookup(result) = this.getATargetScope() }
```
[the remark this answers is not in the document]
asgerf: We already have `Module#getASubClass()` from before this PR, which returns an immediate subclass (not transitive). For that reason I'm trying to avoid adding things called `getSubClass/getSubclass` that don't behave the same way.

asgerf:
> On the naming of `getConst`, is there any reason not to use the full word `getConstant`? 

I preferred a short name given how ubiquitous the corresponding `DataFlow::moduleImport` and `DataFlow::globalVarRef` are in JavaScript. I'm not sure it's a good enough reason though, since it's not that much shorter. I'd be happy to hear if anyone else has thoughts on this as well.

ruby/ql/lib/codeql/ruby/dataflow/internal/DataFlowPublic.qll
```diff
@@ -174,6 +225,82 @@ class LocalSourceNode extends Node {
...
+   * Concretely, this gets the second argument from any call to `[]=` where this node flows to the receiver
+   * and the first argument has constant value `key`.
+   */
+  Node getAnElementWriteValue(ConstantValue key) {
+    exists(CallNode call |
+      call = this.getAMethodCall("[]=") and
+      call.getNumberOfArguments() = 2 and
+      call.getArgument(0).getConstantValue() = key and
+      result = call.getArgument(1)
+    )
```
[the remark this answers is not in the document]
asgerf: Factored out `getAnElementWriteCall` to reduce duplication

ruby/ql/lib/codeql/ruby/dataflow/internal/DataFlowPrivate.qll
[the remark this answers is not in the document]
hvitved: Makes perfect sense; thanks.

asgerf:
[New evaluation](https://github.com/github/codeql-dca-main/issues/8390) shows we gain about 1,600 new call edges due to handling of `proc`, and is otherwise unchanged.

ruby/ql/lib/codeql/ruby/dataflow/internal/DataFlowPrivate.qll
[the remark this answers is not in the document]
asgerf: Merged your fix from https://github.com/asgerf/codeql/pull/1, thanks!

hvitved (approved):
Let's see what DCA says.

asgerf:
[Latest evaluation](https://github.com/github/codeql-dca-main/issues/8423) still looks good

hvitved (approved):

PR #11027, opened 2022-10-28
Swift: WebView JS-native bridge sources

atorralba:
Adds taint sources for JS-native bridges in WebViews. These are objects where JavaScript running in the WebView can execute native code, which can be exploited by malicious websites, or external attackers through XSS vulnerabilities, if it performs sensitive actions.

### DIFF
--- a/swift/ql/lib/codeql/swift/dataflow/internal/DataFlowPrivate.qll
+++ b/swift/ql/lib/codeql/swift/dataflow/internal/DataFlowPrivate.qll
@@ -574,6 +574,14 @@ private module PostUpdateNodes {
 
     override DataFlowCallable getEnclosingCallable() { result = TDataFlowFunc(n.getScope()) }
   }
+
+  class SummaryPostUpdateNode extends SummaryNode, PostUpdateNodeImpl {
+    SummaryPostUpdateNode() { FlowSummaryImpl::Private::summaryPostUpdateNode(this, _) }
+
+    override Node getPreUpdateNode() {
+      FlowSummaryImpl::Private::summaryPostUpdateNode(this, result)
+    }
+  }
 }
 
 private import PostUpdateNodes
--- a/swift/ql/lib/codeql/swift/frameworks/StandardLibrary/WebView.qll
+++ b/swift/ql/lib/codeql/swift/frameworks/StandardLibrary/WebView.qll
@@ -125,7 +125,7 @@ private class JsExportedSource extends RemoteFlowSource {
       base.getEnclosingDecl() instanceof JsExportedProto and
       adopter.getEnclosingDecl() instanceof JsExportedType
     |
-      this.asDefinition().getSourceVariable() = adopter and adopter.getName() = base.getName()
+      this.asExpr().(MemberRefExpr).getMember() = adopter and adopter.getName() = base.getName()
     )
   }
 
--- a/swift/ql/lib/codeql/swift/frameworks/StandardLibrary/WebView.qll
+++ b/swift/ql/lib/codeql/swift/frameworks/StandardLibrary/WebView.qll
@@ -91,39 +91,43 @@ private class JsValueSummaries extends SummaryModelCsv {
   }
 }
 
-/** The class `JSContext`. */
-private class JsContextDecl extends ClassDecl {
-  JsContextDecl() { this.getName() = "JSContext" }
+/** The `JSExport` protocol. */
+private class JsExport extends ProtocolDecl {
+  JsExport() { this.getName() = "JSExport" }
 }
 
-/** The type of an object exposed to JavaScript through `JSContext.setObject`. */
-private class TypeExposedThroughJsContext extends Type {
-  TypeExposedThroughJsContext() {
-    exists(ApplyExpr c, FuncDecl f |
-      c.getStaticTarget() = f and
-      f.getEnclosingDecl() instanceof JsContextDecl and
-      f.getName() = "setObject(_:forKeyedSubscript)"
-    |
-      c.getArgument(0).getExpr().getType() = this
-    )
-  }
+/** A protocol inheriting `JSExport`. */
+private class JsExportedProto extends ProtocolDecl {
+  JsExportedProto() { this.getABaseTypeDecl+() instanceof JsExport }
+}
+
+/** A type that adopts a `JSExport`-inherited protocol. */
+private class JsExportedType extends ClassOrStructDecl {
+  JsExportedType() { this.getABaseTypeDecl*() instanceof JsExportedProto }
 }
 
 /**
- * The members (fields and parameters of functions) of a class or struct
- * an instance of which is exposed to JavaScript through `JSContext.setObject`.
+ * A flow source that models properties and methods defined in a `JSExport`-inherited protocol
+ * and implemented in a type adopting that protcol. These members are accessible from JavaScript
+ * when the object is assigned to a `JSContext`.
  */
-private class ExposedThroughJsContextSource extends RemoteFlowSource {
-  ExposedThroughJsContextSource() {
-    exists(ParamDecl p | this.(DataFlow::ParameterNode).getParameter() = p |
-      p.getDeclaringFunction().getEnclosingDecl().(ClassOrStructDecl).getType() instanceof
-        TypeExposedThroughJsContext
+private class JsExportedSource extends RemoteFlowSource {
+  JsExportedSource() {
+    exists(MethodDecl adopter, MethodDecl base |
+      base.getEnclosingDecl() instanceof JsExportedProto and
+      adopter.getEnclosingDecl() instanceof JsExportedType
+    |
+      this.(DataFlow::ParameterNode).getParameter().getDeclaringFunction() = adopter and
+      adopter.getName() = base.getName()
     )
     or
-    exists(FieldDecl f | this.asDefinition().getSourceVariable() = f |
-      f.getEnclosingDecl().(ClassOrStructDecl).getType() instanceof TypeExposedThroughJsContext
+    exists(FieldDecl adopter, FieldDecl base |
+      base.getEnclosingDecl() instanceof JsExportedProto and
+      adopter.getEnclosingDecl() instanceof JsExportedType
+    |
+      this.asDefinition().getSourceVariable() = adopter and adopter.getName() = base.getName()
     )
   }
 
-  override string getSourceType() { result = "Member of a type exposed through JSContext" }
+  override string getSourceType() { result = "Member of a type exposed through JSExport" }
 }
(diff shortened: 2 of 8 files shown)


swift/ql/test/library-tests/dataflow/taint/webview.swift
```diff
@@ -0,0 +1,99 @@
...
+    v1.defineProperty("", descriptor: s as Any)
+    sink(v1) // $ MISSING: tainted=56
+
+    let v2 = JSValue(object: "", in: context)
+    v2.setValue(s as Any, at: 0)
+    sink(v2) // $ MISSING: tainted=56
+
+    let v3 = JSValue(object: "", in: context)
+    v2.setValue(s as Any, forProperty: "")
+    sink(v3) // $ MISSING: tainted=56
```
atorralba, line 98: I _think_ this `MISSING` cases will be fixed by https://github.com/github/codeql/pull/10967.

atorralba: Fixed it in 2402504a4c095cf9b2218dce80752d17733112a7.

swift/ql/lib/codeql/swift/frameworks/StandardLibrary/WebView.qll
```diff
@@ -27,3 +32,98 @@ private class WKScriptMessageBodyInheritsTaint extends TaintInheritingContent,
...
+ * native code, so any object obtained from it has the potential of being tainted by a malicious
+ * website visited in the WebView.
+ */
+private class JsContextSources extends SourceModelCsv {
+  override predicate row(string row) {
+    row =
+      [
+        ";JSContext;true;globalObject;;;;remote",
+        ";JSContext;true;objectAtIndexedSubscript(_:);;;ReturnValue;remote",
+        ";JSContext;true;objectForKeyedSubscript(_:);;;ReturnValue;remote"
```
geoffw0, line 47: Should these things also be taint sources?
 - the value of `JSContext.exception`
 - the second input to `JSContext.exceptionHandler`?
 - the return value from both variants of `evaluateScript`?

swift/ql/lib/codeql/swift/dataflow/internal/DataFlowPrivate.qll
```diff
@@ -574,6 +574,14 @@ private module PostUpdateNodes {
 
     override DataFlowCallable getEnclosingCallable() { result = TDataFlowFunc(n.getScope()) }
   }
+
+  class SummaryPostUpdateNode extends SummaryNode, PostUpdateNodeImpl {
```
geoffw0, line 578: Please could you explain the purpose of this class?

swift/ql/test/library-tests/dataflow/flowsources/webview.swift
```diff
@@ -23,13 +24,20 @@ class TestMessageHandler: WKScriptMessageHandler {
...
-    context.setObject(Exposed.self, forKeyedSubscript: "exposed" as! NSCopying & NSObjectProtocol)
 }
 
-class Exposed {
+protocol Exported : JSExport {
+    var tainted: Any { get }
+    func tainted(arg1: Any, arg2: Any)
+}
+class ExportedImpl : Exported {
     var tainted: Any { get { return "" } } // SOURCE
```
geoffw0, line 34: We don't appear to catch this one.

swift/ql/lib/codeql/swift/frameworks/StandardLibrary/WebView.qll
```diff
@@ -91,39 +91,43 @@ private class JsValueSummaries extends SummaryModelCsv {
...
-      c.getStaticTarget() = f and
-      f.getEnclosingDecl() instanceof JsContextDecl and
-      f.getName() = "setObject(_:forKeyedSubscript)"
-    |
-      c.getArgument(0).getExpr().getType() = this
-    )
-  }
+/** A protocol inheriting `JSExport`. */
+private class JsExportedProto extends ProtocolDecl {
+  JsExportedProto() { this.getABaseTypeDecl+() instanceof JsExport }
```
geoffw0, line 101: Why is this not `getABaseTypeDecl*()`?  Can `JSExport` not be applied directly?

swift/ql/lib/codeql/swift/frameworks/StandardLibrary/WebView.qll
```diff
@@ -91,39 +91,43 @@ private class JsValueSummaries extends SummaryModelCsv {
...
-      p.getDeclaringFunction().getEnclosingDecl().(ClassOrStructDecl).getType() instanceof
-        TypeExposedThroughJsContext
+private class JsExportedSource extends RemoteFlowSource {
+  JsExportedSource() {
+    exists(MethodDecl adopter, MethodDecl base |
+      base.getEnclosingDecl() instanceof JsExportedProto and
+      adopter.getEnclosingDecl() instanceof JsExportedType
+    |
+      this.(DataFlow::ParameterNode).getParameter().getDeclaringFunction() = adopter and
+      adopter.getName() = base.getName()
```
geoffw0, line 121: What is the purpose of this check?  Is this because some things in the class can't be accessed by JS?

swift/ql/test/library-tests/dataflow/flowsources/webview.swift
```diff
@@ -0,0 +1,43 @@
...
+protocol WKScriptMessageHandler {
+    func userContentController(_ userContentController: WKUserContentController, didReceive message: WKScriptMessage)
+}
+protocol NSCopying {}
+protocol NSObjectProtocol {}
+class JSValue {}
+class JSContext {
+    var globalObject: JSValue { get { return JSValue() } }
+    func objectForKeyedSubscript(_: Any!) -> JSValue! { return JSValue() } 
+    func setObject(_: Any, forKeyedSubscript: (NSCopying & NSObjectProtocol) ) {}
```
geoffw0, line 14: So to make sure I understand: in an early version you were looking for types that are handed to `setObject`, because JS can potentially taint them or anything contained in them.  But now you look for all types with the `JSExport` prototype because presumably that's required to do this, and is more accurate?

geoffw0 (commented):
This is complicated stuff, but they look like fantastic taint sources.  A few comments / questions...

swift/ql/lib/codeql/swift/dataflow/internal/DataFlowPrivate.qll
atorralba: This was a missing piece in the data flow library to catch cases like the this: https://github.com/github/codeql/pull/11027/files#diff-7e1f3d77d83b67247f8f21045e82539532beff203bfccf85a0080314064f3ffaR89

Basically, if a summary introduced a step involving updating the taint of a node (like `arg -> qualifier` flow), we weren't propagating flow correctly through it because of this missing node.

swift/ql/test/library-tests/dataflow/flowsources/webview.swift
atorralba: Good catch, there as an issue in how the source was modeled around fields. Fixed in b62ede1, thanks!

swift/ql/lib/codeql/swift/frameworks/StandardLibrary/WebView.qll
atorralba: Applying `JSExport` directly would theoretically do nothing. The expected use is extending the prototype in another prototype which declares the members you want to export to JavaScript. Then classes applying the prototype will export those members only (by default, if `JSExport` isn't used, no members are exported).

swift/ql/lib/codeql/swift/frameworks/StandardLibrary/WebView.qll
atorralba: Yes, this is to reflect that only the members declared in the protocol are exported to JavaScript, not all members of the adopting class/struct. I couldn't find a better way to model member inheritance, happy to improve this if there's a better alternative.

swift/ql/test/library-tests/dataflow/flowsources/webview.swift
atorralba: Yes, I first thought that an object exposed through `setObject` would automatically export all its members to JS, but turns out it doesn't. You must declare explicitly, through a protocol that extends `JSExport`, the members that need to be exported. All the other members remain unaccessible from JavaScript.

And I thought that modeling the `JSExport` logic was enough. Theoretically the object would still need to be added to the `JSContext`, but I imagine that applications rarely would declare a type adopting a `JSExport`-inherited protocol and then not use it in the context. And still if they do, we probably want to alert if there's a dangerous path starting on them anyway, since it would be a ticking bomb waiting to explode when the object finally got added to the context.

But let me know if you think we should add the `setObject` constraint to the existing logic to be 100% accurate.

swift/ql/lib/codeql/swift/frameworks/StandardLibrary/WebView.qll
atorralba: > the value of `JSContext.exception`

[`JSContext.exception`](https://developer.apple.com/documentation/javascriptcore/jscontext/1451499-exception) is an exception that gets thrown on the JavaScript side once the native code execution has finished. So I don't _think_ JavaScript can use that field to execute native code.

> the second input to `JSContext.exceptionHandler`?

The `exceptionHandler` is a block that executes native code indeed, but I think it can't be set from the JavaScript side. The `JSValue` that is the second argument is the exception that got thrown in a native code callback, which again doesn't seem to be JS-provided nor should affect the native code — it's targeted to JavaScript so that the exception can be handled there (if I understood [all this](https://developer.apple.com/documentation/javascriptcore/jscontext/1451731-exceptionhandler) correctly).

> the return value from both variants of `evaluateScript`?

That indeed comes for JavaScript. The thing is that the JavaScript being executed in this case comes from the app itself, not the website visited in the WebView. So unless the existing JavaScript could somehow tamper the code passed to `evaluateScript`, its return value wouldn't be under the malicious site's control. Of course controlling the script of the `evaluateScript` call itself would be a problem, but that's more akin to a XSS vulnerability.

geoffw0: Thanks for clearing these things up.  It sounds like you've considered these cases in much more detail than I have!

swift/ql/lib/codeql/swift/dataflow/internal/DataFlowPrivate.qll
MathiasVP: Indeed, this change LGTM 👍.

swift/ql/test/library-tests/dataflow/flowsources/webview.swift
geoffw0: This sounds quite reasonable as it is, if someone is marking something to export to JS but we can't see where, on balance I would probably prefer to still report a result.

geoffw0 (approved):
I'm happy with this and I'm really glad to have more remote flow sources!

PR #11029, opened 2022-10-28
Swift: add possibility to run the extractor under an env-specified tool

redsun82:
if `CODEQL_EXTRACTOR_SWIFT_RUN_UNDER` env variable is set, and either
* `CODEQL_EXTRACTOR_SWIFT_RUN_UNDER_FILTER` is not set, or
* it is set to a regexp matching any substring of the extractor call then the extractor process is substituted with the command (and possibly options) stated in `CODEQL_EXTRACTOR_SWIFT_RUN_UNDER`, followed by the system arguments of the extractor itself (which should include the extractor program itself at the start).

Before calling `exec`, `CODEQL_EXTRACTOR_SWIFT_RUN_UNDER` is unset to avoid unpleasant loops.

An example usage is to run the extractor under `gdbserver :1234` when the arguments match a given source file.


swift/extractor/main.cpp
```diff
@@ -51,7 +53,51 @@ static void lockOutputSwiftModuleTraps(const codeql::SwiftExtractorConfiguration
...
+void checkToRunUnderTool(int argc, char* const* argv) {
+  auto runUnder = getenv("CODEQL_EXTRACTOR_SWIFT_RUN_UNDER");
+  if (runUnder == nullptr) {
+    return;
+  }
+  auto runUnderFilter = getenv("CODEQL_EXTRACTOR_SWIFT_RUN_UNDER_FILTER");
+  if (runUnderFilter != nullptr) {
+    assert(argc > 0);
+    std::string call = argv[0];
+    for (auto i = 0; i < argc; ++i) {
```
d10c, line 74: `argv[0]` gets appended twice at the beginning; `i` should start from 1.

d10c (commented):
LGTM, with minor suggestions.

swift/extractor/main.cpp
redsun82: thanks for spotting this! I hesitated between two versions and ended up doing a wrong mix of the two :sweat_smile:

d10c (approved):
Some tests would be welcome, but the code looks good.

d10c (dismissed):
Nice work on the tests!

d10c (approved):

PR #11014, opened 2022-10-27
Kotlin: Resugar `for` loops

tamasvajk:

PR #11032, opened 2022-10-28
Kotlin: exclude loop variables on ranges from 'unused locals' check

tamasvajk:
Based on https://github.com/github/codeql/pull/11014.

### DIFF
--- a/java/ql/src/Violations of Best Practice/Dead Code/UnreadLocal.ql
+++ b/java/ql/src/Violations of Best Practice/Dead Code/UnreadLocal.ql
@@ -26,5 +26,10 @@ where
   not exists(getARead(v)) and
   // Discarded exceptions are covered by another query.
   not exists(CatchClause cc | cc.getVariable().getVariable() = v) and
+  // Exclude common Kotlin pattern to do something n times: `for(i in 1..n) { doSomething() }
+  not exists(EnhancedForStmt f |
+    f.getVariable().getVariable() = v and
+    f.getExpr().getType().(RefType).hasQualifiedName("kotlin.ranges", ["IntRange", "LongRange"])
+  ) and
   not readImplicitly(v)
 select v, "Variable '" + v + "' is never read."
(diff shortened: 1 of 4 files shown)


java/ql/src/Violations of Best Practice/Dead Code/UnreadLocal.ql
```diff
@@ -26,5 +26,10 @@ where
   not exists(getARead(v)) and
   // Discarded exceptions are covered by another query.
   not exists(CatchClause cc | cc.getVariable().getVariable() = v) and
+  // Exclude common Kotlin pattern: `for(i in x..y) { ... }`
```
smowton, line 29: Clarify what we're excluding? I guess not `i`; something synthetic instead?

tamasvajk: We're excluding `i`. I've seen the below pattern in a couple of variations:
```
for (i in 1..5) {
  val frame = input.readFrame(Long.MAX_VALUE, 0)
  ... // not using i
}
```

smowton: Seems like a bad pattern to me considering https://kotlinlang.org/api/latest/jvm/stdlib/kotlin/repeat.html exists, but I guess if it happens often enough it's more likely to be noise.

tamasvajk: I found it a couple of times in `ktor`, that's why I thought I'd exclude it. Also, I don't think there's a real alternative. I haven't seen `repeat` before, but `repeat` takes a lambda `(Int) -> Unit`, where we wouldn't use the `Int` parameter, so it's the same issue. (I guess in a lambda we can use an implicit parameter, or a parameter named `_`, so the useless parameter check would not trigger, but in essence it's the same issue)

igfoo: Might be worth expanding the comment, e.g.
```
// Exclude common Kotlin pattern to do something n times: `for(i in 1..n) { doSomething() }
```

aschackmull (approved):

PR #11033, opened 2022-10-28
C++: Don't create `DataFlow::Node`s for `void`-typed instructions

MathiasVP:
Shamelessly stolen from [Java](https://github.com/github/codeql/blob/main/java/ql/lib/semmle/code/java/dataflow/internal/DataFlowNodes.qll#L13) 😂. These nodes can never contain any data so we don't need dataflow nodes for them.


geoffw0 (approved):
LGTM.  I assume this is intended to have a positive effect on performance?

MathiasVP:
Yep, exactly. Thanks for the review!

PR #11034, opened 2022-10-28
Swift: Add and use AbstractFunctionDecl.hasGlobalName predicate.

geoffw0:
This allows us to concisely match a non-member function.  At the moment its only used in one place but I expect there will be a lot more of this in future.


d10c (approved):
LGTM 👍

I was considering adding a GlobalFunctionDecl class (AbstractFunctionDecl minus MethodDecl), but I guess this predicate will subsume most of the usefulness of that class.

MathiasVP:
Is "global" the right terminology in a language that uses modules? After all, the function isn't global in the same sense that it's global in C/C++.

d10c:
You're right. Another name for non-methods could be "free" functions instead of "global". But `f.hasFreeName(...)` sounds awkward, so another way to phrase it could be `f.(FreeFunctionDecl).hasName(...)`

geoffw0:
> `f.hasFreeName(...)` sounds awkward, so another way to phrase it could be `f.(FreeFunctionDecl).hasName(...)`

I'm happy with either.  @MathiasVP which do you prefer?

MathiasVP:
> I'm happy with either. @MathiasVP which do you prefer?

I think I like `f.(FreeFunctionDecl).hasName(...)` more. We should really have such a function anyway so if it also makes the ergonomics easier in this case I think it's a strong argument for choosing this one 👍.

geoffw0:
Updated.  And fixed the merge conflict.  And I've added a bit more qldoc, though I'm not expecting to win any awards for:
```
/**
 * A function.
 */
```

MathiasVP (approved):
😂. Yeah, we should revisit that QLDoc and provide proper examples. But the code LGTM, at least!

geoffw0:
("A function" is actually not saying nothing in this case; it's saying "Yeah, don't worry about that funky name `AbstractFunctionDecl`, this really is the class you want to represent a function")

MathiasVP:
> ("A function" is actually not saying nothing in this case; it's saying "Yeah, don't worry about that funky name AbstractFunctionDecl, this really is the class you want to represent a function")

Good point. That name _should_ be changed eventually as well, though.

PR #10993, opened 2022-10-26
Swift: detect the use of constant salts

karimhamdanali:
Constant salts should not be used for password hashing. Data hashed using constant salts are vulnerable to dictionary attacks, enabling attackers to recover the original input.

The rule currently supports all ciphers that the CryptoSwift API provides, but we can always extend it further if more APIs are added.

I'd appreciate a review of the query itself, the accompanying tests, and the associated documentation.

PR #11036, opened 2022-10-28
Swift: Add and use ApplyExpr.getArgumentByParamName.

geoffw0:
Add and use a predicate `ApplyExpr.getArgumentByParamName` (please say if you can think of a cleaner name for it).  This makes it easier to get an argument by its parameter name, rather than the argument name (which is often anonymous) or index (usually the preferred method, but sometimes varies across the functions of interest).

This will be used in a couple of upcoming queries in addition to the two I've updated here.

### DIFF
--- a/swift/ql/lib/codeql/swift/elements/expr/ApplyExpr.qll
+++ b/swift/ql/lib/codeql/swift/elements/expr/ApplyExpr.qll
@@ -19,6 +19,17 @@ class ApplyExpr extends Generated::ApplyExpr {
   /** Gets the method qualifier, if this is applying a method */
   Expr getQualifier() { none() }
 
+  /**
+   * Gets the argument that has corresponding parameter name `paramName` (if
+   * any). If this call does not have a static target, there will be no result.
+   */
+  final Argument getArgumentByParamName(string paramName) {
+    exists(int arg |
+      this.getStaticTarget().getParam(pragma[only_bind_into](arg)).getName() = paramName and
+      this.getArgument(pragma[only_bind_into](arg)) = result
+    )
+  }
+
   override string toString() {
     result = "call to " + this.getStaticTarget().toString()
     or
--- a/swift/ql/lib/codeql/swift/elements/expr/ApplyExpr.qll
+++ b/swift/ql/lib/codeql/swift/elements/expr/ApplyExpr.qll
@@ -19,17 +19,6 @@ class ApplyExpr extends Generated::ApplyExpr {
   /** Gets the method qualifier, if this is applying a method */
   Expr getQualifier() { none() }
 
-  /**
-   * Gets the argument that has corresponding parameter name `paramName` (if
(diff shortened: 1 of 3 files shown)


MathiasVP:
Those names are really called "labels". In fact, "parameter names" are really the name used for the parameter inside the function body. i.e., for a function like:
```
func f(x y : Int) -> Int { return y }
```
the parameter name for the first parameter is `y`, but the _label_ used for the first argument when calling `f` is `x`. See https://docs.swift.org/swift-book/LanguageGuide/Functions.html#ID166 for some of the official Swift jargon.

So maybe we should call this new predicate `getArgumentByLabel`?

d10c:
What happens when a parameter has a default value, and an ApplyExpr omits an argument for that parameter? Do you think the parameter's default expression should be the argument in that case?

geoffw0:
Good point, yes, that seems like a reasonable thing to do.  In which case I think I'm going to need to add a test for this...

d10c:
As for naming, getArgumentByParamName sounds fine. Alternatives: getArgumentNamed, getArgumentByName.

MathiasVP:
Hmm. Slightly related comment: why are we using the name of the parameter and not just its index? The internal parameter name seems like something that the developers are free to change, and it's a bit scary that we're depending on this name

geoffw0:
> why are we using the name of the parameter and not just its index?

There have been a couple of cases where there's a reasonably large collection of functions with the same name and similar parameters, but the target parameter is not at a consistent position.  So expressing it by parameter name is more concise and probably more robust.  I think using the index is preferred in most other cases.

geoffw0:
... see also upcoming PRs https://github.com/github/codeql/pull/10947 and https://github.com/github/codeql/pull/10993 .  Now that you mention it the two already in the main branch are not great examples as they would work just as well with indices.

geoffw0:
I've changed the two existing cases to use argument indexes instead.  I'm now unsure whether we want `ApplyExpr.getArgumentByParamName` (probably renamed and with a default values fix) for the upcoming PRs or not.

geoffw0:
OK, in the interests of having fewer balls in the air I've removed the addition of `ApplyExpr.getArgumentByParamName` from this PR, all it does now is convert two places where parameter names were being matches to used to use argument indexes instead.  Lets get this merged + closed.

We can have another go at a `getArgumentByParamName` at a later point if we decide we really do want it - but with a better name + tests including cases with default arguments.

MathiasVP (approved):
LGTM! Since this is changing code that needed some join-order assistance in the past I think we should run DCA on it before merging it.

geoffw0:
DCA run is finished.  Some stage timing results (quite a lot actually - and apparently not in a good way - but I don't know how to read these yet) .  Overall analysis time is unchanged.

MathiasVP:
(I'm currently looking at a bugfix for the stage timing. Don't worry about those for now)

PR #11045, opened 2022-10-31
Kotlin: Excluded compiler generated methods from `java/confusing-method-signature`

tamasvajk:


java/ql/src/Violations of Best Practice/Naming Conventions/ConfusingOverloading.ql
```diff
@@ -137,6 +137,8 @@ private predicate delegate(Method caller, Method callee) {
 
 from Method m, Method n, string messageQualifier
 where
+  // Exclude compiler generated methods, such as Kotlin `$default` methods:
+  not m.isCompilerGenerated() and
```
igfoo, line 141: I think that it would be better to do this in `candidateMethod`.

smowton (approved):

PR #11046, opened 2022-10-31
Swift: Simplify the API for Decl members

d10c:
- [x] Accessors are considered immediate members of nominal types and extensions, not just their directly enclosing `VarDecl`/`SubscriptDecl`.
- [x] `MethodDecl` relies on the above behaviour, not a special-cased `getAMember()` predicate.
- [ ] Rename the user-facing IterableDeclContext to DeclWithMembers, add Decl as a superclass.
- [ ] Maybe somehow move the `DeclWithMembers.getAMember()` (etc) methods to the Decl class so as to be more generally useful. But this would involve interfering with the autogenerated code in `Generated::IterableDeclContext`.
- [ ] New method named `getDeclaringDecl` as an inverse of `getAMember()`.
- [ ] New method named `getDeclaringTypeDecl` that also resolves Extensions to the `NominalTypeDecl` that they extend. Use this to simplify `MethodDecl.hasQualifiedName/2`.


swift/ql/lib/codeql/swift/elements/decl/VarDecl.qll
```diff
@@ -9,5 +9,5 @@ class VarDecl extends Generated::VarDecl {
 }
 
 class FieldDecl extends VarDecl {
```
d10c, line 11: Maybe this should be PropertyDecl, to keep it consistent with the language's preferred name.

swift/ql/lib/codeql/swift/elements/decl/Decl.qll
```diff
@@ -1,4 +1,23 @@
...
+private import codeql.swift.elements.decl.NominalTypeDecl
+private import codeql.swift.elements.decl.ExtensionDecl
 
-class Decl extends Generated::Decl { }
+class Decl extends Generated::Decl {
+  /**
+   * Gets the class, struct, enum, protocol, or extension that declared
+   * this `Decl` as a member.
+   */
+  cached
```
d10c, line 11: Is there a rule of thumb for when I should cache a predicate?

swift/ql/lib/codeql/swift/elements/decl/DeclWithMembers.qll
```diff
@@ -0,0 +1,26 @@
...
+private import codeql.swift.elements.decl.VarDecl
+private import codeql.swift.elements.decl.SubscriptDecl
+private import codeql.swift.elements.decl.Decl
+
+class DeclWithMembers extends Decl, IterableDeclContext {
+  /**
+   * Gets the `index`th member (0-based) of this nominal type or extension `Decl`,
+   * including `AccessorDecl`s of immediate members.
+   */
+  override Decl getImmediateMember(int index) {
```
d10c, line 11: Not sure, but maybe this logic should be part of the extractor or codegen.

d10c (commented):
Code reviews welcome.

swift/ql/lib/codeql/swift/elements/decl/Decl.qll
MathiasVP: It's always a bit difficult to judge when (and how) something should be cached. Generally, if a predicate is (1) expensive to compute, and (2) used in multiple queries, then it should most likely be cached. In this case, I don't think that the predicate should be cached as I don't think `getDeclaringDecl` is expensive to compute.

swift/ql/lib/codeql/swift/elements/decl/DeclWithMembers.qll
```diff
@@ -0,0 +1,26 @@
+private import codeql.swift.elements.decl.IterableDeclContext
+private import codeql.swift.elements.decl.VarDecl
+private import codeql.swift.elements.decl.SubscriptDecl
+private import codeql.swift.elements.decl.Decl
+
+class DeclWithMembers extends Decl, IterableDeclContext {
```
MathiasVP, line 6: I'm still of the opinion that there's not really a need to have a class like this as part of the public AST. Rather, I'd much rather see the importing of the `IterableDeclContext` class disappear from `swift.qll`, and its use purely as an implementation detail in order to get `getImmediateMember`/`getMember` predicates on the [classes](https://github.com/github/codeql/blob/a1edd655420b86a776c7a8c3659fbacd194ca664/swift/ql/lib/codeql/swift/generated/decl/NominalTypeDecl.qll#L9) [that](https://github.com/github/codeql/blob/a1edd655420b86a776c7a8c3659fbacd194ca664/swift/ql/lib/codeql/swift/generated/decl/ExtensionDecl.qll#L10) currently extend `IterableDeclContext`.

I think we should delay this change until we've settled on a decision here.

geoffw0: And I'm not very happy with the name `DeclWithMembers`.  I think we're talking about a declaration of a kind that _can_ have members, not one that necessarily does.  Consider:
```
struct MyStruct {}
```

swift/ql/lib/codeql/swift/elements/decl/DeclWithMembers.qll
geoffw0: We generally prefer to put logic in QL when we have the choice, and there are no good reasons for putting in the extractor.

geoffw0: ... I'd like to tell you why we prefer that.  I think it has to do with QL being easier to see and easier to change.

swift/ql/lib/codeql/swift/elements/decl/MethodDecl.qll
```diff
@@ -29,17 +11,7 @@ class MethodDecl extends AbstractFunctionDecl {
...
-        c.getFullName() = typeName and
-        c.getAMember() = this
-      )
-      or
-      exists(ExtensionDecl e |
-        e.getExtendedTypeDecl().getFullName() = typeName and
-        e.getAMember() = this
-      )
-    )
+    this.getDeclaringTypeDecl().getFullName() = typeName
```
geoffw0, line 14: I do like this, having `getDeclaringDecl` and `getDeclaringTypeDecl` both widely available and clearly defined is a big plus.

swift/ql/lib/codeql/swift/elements/decl/Decl.qll
geoffw0: The simplest rule of thumb is: if in doubt, don't mark predicates  `cached` when you write them (or use `pragma[only_bind_into]` and other hints).  The optimizer doesn't _usually_ need these tags to do a good job, and you can make things worse by using them.  In the case of `cached`, I believe it prevents the optimizer from using the context in which the predicate is called to reduce its size, which can sometimes cost more than any gains.

The time to use these tools IMO is when you've identified a measurable performance problem and you're trying to fix it.  I guess the other situation is when you're implementing or copying a familiar pattern where `cached` has worked out well before.

swift/ql/lib/codeql/swift/elements/decl/IterableDeclContext.qll
```diff
@@ -1,4 +1,26 @@
...
+          or
+          member = immMember.(VarDecl).getImmediateAccessorDecl(accIndex)
+          or
+          member = immMember.(SubscriptDecl).getImmediateAccessorDecl(accIndex)
+        )
+      |
+        member order by immIndex, accIndex
+      )
+  }
+}
```
redsun82: I'm not really sold on this change, as I do think accessors are logically and visually under a `VarDecl` or `SubscriptDecl` (and by visually I mean they are declared in subblocks of variable/subscript declarations). I would rather keep them that way.

Even if we do go with this, this requires taking them out of the children of `VarDecl`/`SubscriptDecl`, as the `ast/no_double_parents.ql` test failure shows.

PR #11048, opened 2022-10-31
C++: Add `strcpy` test for `cpp/non-constant-format`

jketema:


cpp/ql/test/query-tests/Likely Bugs/Format/NonConstantFormat/test.cpp
```diff
@@ -150,3 +151,8 @@ void print_ith_message() {
   set_value_of(&i);
   printf(messages[i], 1U); // GOOD
 }
+
+void fmt_via_strcpy(char *data) {
+    strcpy(data, "some string");
+    printf(data);
```
MathiasVP, line 157:
```suggestion
    printf(data); // BAD
```
right?

jketema: Good point.

MathiasVP (approved):
LGTM!

PR #11049, opened 2022-10-31
QL: prevent some cross-talk between modules

erik-krogh:
This PR makes no difference now, but it will once I introduce the shared ReDoS pack.  

The problem comes e.g. from me importing `RegExpTreeView` (defined by each language) through some of the shared implementations.  
Because we collapse the instances of the parameterized modules there is no way of preventing that cross-talk between the modules.  

The consequence was e.g. that a reference `DataFlow::Configuration` in a Python query resolved too `DataFlow::Configuration` from JavaScript.  

The change is ugly, but it seems to work (the consistency queries are empty).


tausbn (approved):
Not too thrilled about a `noopt` this complicated. (Could it really not be done using directional unbind and `nomagic`?)

However, to unblock things in the short term, I think this is okay.

PR #11055, opened 2022-10-31
Sink endpoint characteristics

tiferet:
This PR adds the class `EndpointCharacteristic` (formerly referred to as `ClassificationReason` in [the design doc](https://docs.google.com/document/d/1PO0kxGCGt07J2ek0vXb1NDL8QAzTrR-hiHAgOiOHrRU/edit#)).

As a first step, it implements only the four characteristics that indicate that an endpoint is a _sink_. Subsequent PRs will add characteristics that indicate an endpoint is not a sink. 

The definition of a known sink can now be written in a generic fashion in the base class ATMConfig.qll without needing each query's config to implement it independently.

The same logic will be used to surface positive training samples in a subsequent PR.

Update: I've written the characteristics that will replace NotASinkReason and verified that I can reproduce the current selection of training examples. I need to clean up that code before opening a PR with it, though.

Timing experiment: https://github.com/github/codeql-dca-main/issues/8273

Closes https://github.com/github/ml-ql-adaptive-threat-modeling/issues/2096

### DIFF
--- a/javascript/ql/experimental/adaptivethreatmodeling/lib/experimental/adaptivethreatmodeling/ATMConfig.qll
+++ b/javascript/ql/experimental/adaptivethreatmodeling/lib/experimental/adaptivethreatmodeling/ATMConfig.qll
@@ -6,7 +6,7 @@
 
 private import javascript as JS
 import EndpointTypes
-import ClassificationReasons
+import EndpointCharacteristics
 
 /**
  * EXPERIMENTAL. This API may change in the future.
@@ -46,11 +46,11 @@ abstract class AtmConfig extends string {
    * Holds if `sink` is a known sink of flow.
    */
   predicate isKnownSink(JS::DataFlow::Node sink) {
-    // If the list of reasons includes positive indicators with maximal confidence for this class, it's a known sink for
-    // the class.
-    exists(ClassificationReason reason |
-      reason.getEndpoints(sink) and
-      reason.getImplications(this.getASinkEndpointType(), true, 1.0)
+    // If the list of characteristics includes positive indicators with maximal confidence for this class, then it's a
+    // known sink for the class.
+    exists(EndpointCharacteristic characteristic |
+      characteristic.getEndpoints(sink) and
+      characteristic.getImplications(this.getASinkEndpointType(), true, 1.0)
     )
   }
 
--- a/javascript/ql/experimental/adaptivethreatmodeling/lib/experimental/adaptivethreatmodeling/ATMConfig.qll
+++ b/javascript/ql/experimental/adaptivethreatmodeling/lib/experimental/adaptivethreatmodeling/ATMConfig.qll
@@ -6,6 +6,7 @@
 
 private import javascript as JS
 import EndpointTypes
+import ClassificationReasons
 
 /**
  * EXPERIMENTAL. This API may change in the future.
@@ -44,7 +45,14 @@ abstract class AtmConfig extends string {
    *
    * Holds if `sink` is a known sink of flow.
    */
-  predicate isKnownSink(JS::DataFlow::Node sink) { none() }
+  predicate isKnownSink(JS::DataFlow::Node sink) {
+    // If the list of reasons includes positive indicators with maximal confidence for this class, it's a known sink for
+    // the class.
+    exists(ClassificationReason reason |
+      reason.getEndpoints(sink) and
+      reason.getImplications(this.getASinkEndpointType(), true, 1.0)
+    )
+  }
 
   /**
    * EXPERIMENTAL. This API may change in the future.
--- a/javascript/ql/experimental/adaptivethreatmodeling/lib/experimental/adaptivethreatmodeling/ATMConfig.qll
+++ b/javascript/ql/experimental/adaptivethreatmodeling/lib/experimental/adaptivethreatmodeling/ATMConfig.qll
@@ -45,7 +45,7 @@ abstract class AtmConfig extends string {
    *
    * Holds if `sink` is a known sink of flow.
    */
-  predicate isKnownSink(JS::DataFlow::Node sink) {
+  final predicate isKnownSink(JS::DataFlow::Node sink) {
     // If the list of characteristics includes positive indicators with maximal confidence for this class, then it's a
     // known sink for the class.
     exists(EndpointCharacteristic characteristic |
--- a/javascript/ql/experimental/adaptivethreatmodeling/lib/experimental/adaptivethreatmodeling/EndpointCharacteristics.qll
+++ b/javascript/ql/experimental/adaptivethreatmodeling/lib/experimental/adaptivethreatmodeling/EndpointCharacteristics.qll
@@ -9,19 +9,22 @@ private import semmle.javascript.security.dataflow.NosqlInjectionCustomizations
 private import semmle.javascript.security.dataflow.TaintedPathCustomizations
 
 /**
- * Defines a set of characteristics that a particular endpoint might have. This set of characteristics is used to make
- * decisions about whether to include the endpoint in the training set and with what label, as well as whether to score
- * the endpoint at inference time.
+ * A set of characteristics that a particular endpoint might have. This set of characteristics is used to make decisions
+ * about whether to include the endpoint in the training set and with what label, as well as whether to score the
+ * endpoint at inference time.
  */
 abstract class EndpointCharacteristic extends string {
   /**
-   * The name of the characteristic, which should describe some characteristic of the endpoint that is meaningful for
-   *  determining whether it's a sink and if so of which type
+   * Holds when the string matches the name of the characteristic, which should describe some characteristic of the
+   * endpoint that is meaningful for determining whether it's a sink and if so of which type
    */
   bindingset[this]
   EndpointCharacteristic() { any() }
 
-  /** The logic to identify which endpoints have this characteristic. */
+  /**
+   * Holds for endpoints that have this characteristic. This predicate contains the logic that applies characteristics
+   * to the appropriate set of dataflow nodes.
+   */
   abstract predicate getEndpoints(DataFlow::Node n);
 
   /**
--- a/javascript/ql/experimental/adaptivethreatmodeling/lib/experimental/adaptivethreatmodeling/EndpointCharacteristics.qll
+++ b/javascript/ql/experimental/adaptivethreatmodeling/lib/experimental/adaptivethreatmodeling/EndpointCharacteristics.qll
@@ -21,12 +21,6 @@ abstract class EndpointCharacteristic extends string {
   bindingset[this]
   EndpointCharacteristic() { any() }
 
-  /** Indicators with confidence at or above this threshold are considered to be high-confidence indicators. */
-  float getHighConfidenceThreshold() { result = 0.8 }
-
-  /** Indicators with confidence at or above this threshold are considered to be medium-confidence indicators. */
-  float getMediumConfidenceThreshold() { result = 0.5 }
-
   /** The logic to identify which endpoints have this characteristic. */
   abstract predicate getEndpoints(DataFlow::Node n);
 
--- a/javascript/ql/experimental/adaptivethreatmodeling/lib/experimental/adaptivethreatmodeling/EndpointCharacteristics.qll
+++ b/javascript/ql/experimental/adaptivethreatmodeling/lib/experimental/adaptivethreatmodeling/EndpointCharacteristics.qll
@@ -1,9 +1,5 @@
 /**
  * For internal use only.
- *
- * Defines a set of characteristics that a particular endpoint might have. This set of characteristics is used to make
- * decisions about whether to include the endpoint in the training set and with what label, as well as whether to score
- * the endpoint at inference time.
  */
 
 import experimental.adaptivethreatmodeling.EndpointTypes
@@ -12,37 +8,45 @@ private import semmle.javascript.security.dataflow.DomBasedXssCustomizations
 private import semmle.javascript.security.dataflow.NosqlInjectionCustomizations
 private import semmle.javascript.security.dataflow.TaintedPathCustomizations
 
+/**
+ * Defines a set of characteristics that a particular endpoint might have. This set of characteristics is used to make
+ * decisions about whether to include the endpoint in the training set and with what label, as well as whether to score
+ * the endpoint at inference time.
+ */
 abstract class EndpointCharacteristic extends string {
-  // The name of the characteristic, which should describe some characteristic of the endpoint that is meaningful for
-  // determining whether it's a sink and if so of which type
+  /**
+   * The name of the characteristic, which should describe some characteristic of the endpoint that is meaningful for
+   *  determining whether it's a sink and if so of which type
+   */
   bindingset[this]
   EndpointCharacteristic() { any() }
 
-  // Indicators with confidence at or above this threshold are considered to be high-confidence indicators.
+  /** Indicators with confidence at or above this threshold are considered to be high-confidence indicators. */
   float getHighConfidenceThreshold() { result = 0.8 }
 
-  // Indicators with confidence at or above this threshold are considered to be medium-confidence indicators.
+  /** Indicators with confidence at or above this threshold are considered to be medium-confidence indicators. */
   float getMediumConfidenceThreshold() { result = 0.5 }
 
-  // The logic to identify which endpoints have this characteristic.
+  /** The logic to identify which endpoints have this characteristic. */
   abstract predicate getEndpoints(DataFlow::Node n);
 
-  // This predicate describes what the characteristic tells us about an endpoint.
-  //
-  // Params:
-  // endpointClass: Class 0 is the negative class. Each positive int corresponds to a single sink type.
-  // isPositiveIndicator: Does this characteristic indicate this endpoint _is_ a member of the class, or that it
-  // _isn't_ a member of the class?
-  // confidence: A number in [0, 1], which tells us how strong an indicator this characteristic is for the endpoint
-  // belonging / not belonging to the given class.
+  /**
+   * This predicate describes what the characteristic tells us about an endpoint.
+   *
+   *  Params:
+   *  endpointClass: Class 0 is the negative class. Each positive int corresponds to a single sink type.
+   *  isPositiveIndicator: Does this characteristic indicate this endpoint _is_ a member of the class, or that it
+   *  _isn't_ a member of the class?
+   *  confidence: A number in [0, 1], which tells us how strong an indicator this characteristic is for the endpoint
+   *  belonging / not belonging to the given class.
+   */
   abstract predicate getImplications(
     EndpointType endpointClass, boolean isPositiveIndicator, float confidence
   );
 }
 
 /**
- * Endpoints that were identified as "DomBasedXssSink" by the standard Javascript library are XSS sinks with maximal
- * confidence.
+ * Endpoints identified as "DomBasedXssSink" by the standard JavaScript libraries are XSS sinks with maximal confidence.
  */
 private class DomBasedXssSinkCharacteristic extends EndpointCharacteristic {
   DomBasedXssSinkCharacteristic() { this = "DomBasedXssSink" }
@@ -57,8 +61,8 @@ private class DomBasedXssSinkCharacteristic extends EndpointCharacteristic {
 }
 
 /**
- * Endpoints that were identified as "TaintedPathSink" by the standard Javascript library are path injection sinks with
- * maximal confidence.
+ * Endpoints identified as "TaintedPathSink" by the standard JavaScript libraries are path injection sinks with maximal
+ * confidence.
  */
 private class TaintedPathSinkCharacteristic extends EndpointCharacteristic {
   TaintedPathSinkCharacteristic() { this = "TaintedPathSink" }
@@ -73,8 +77,8 @@ private class TaintedPathSinkCharacteristic extends EndpointCharacteristic {
 }
 
 /**
- * Endpoints that were identified as "SqlInjectionSink" by the standard Javascript library are SQL injection sinks with
- * maximal confidence.
+ * Endpoints identified as "SqlInjectionSink" by the standard JavaScript libraries are SQL injection sinks with maximal
+ * confidence.
  */
 private class SqlInjectionSinkCharacteristic extends EndpointCharacteristic {
   SqlInjectionSinkCharacteristic() { this = "SqlInjectionSink" }
@@ -91,8 +95,8 @@ private class SqlInjectionSinkCharacteristic extends EndpointCharacteristic {
 }
 
 /**
- * Endpoints that were identified as "NosqlInjectionSink" by the standard Javascript library are NoSQL injection sinks
- * with maximal confidence.
+ * Endpoints identified as "NosqlInjectionSink" by the standard JavaScript libraries are NoSQL injection sinks with
+ * maximal confidence.
  */
 private class NosqlInjectionSinkCharacteristic extends EndpointCharacteristic {
   NosqlInjectionSinkCharacteristic() { this = "NosqlInjectionSink" }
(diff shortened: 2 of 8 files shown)


tiferet:
@kaeluka IIRC you said there's extensive testing in the PR checks, so the fact that those have passed indicates this PR has made no change to the training data or the endpoints that get scored at inference time, right?

javascript/ql/experimental/adaptivethreatmodeling/lib/experimental/adaptivethreatmodeling/EndpointCharacteristics.qll
```diff
@@ -0,0 +1,113 @@
...
+  override predicate getEndpoints(DataFlow::Node n) { n instanceof DomBasedXss::Sink }
+
+  override predicate getImplications(
+    EndpointType endpointClass, boolean isPositiveIndicator, float confidence
+  ) {
+    endpointClass instanceof XssSinkType and isPositiveIndicator = true and confidence = 1.0
+  }
+}
+
+/*
```
kaeluka, line 60: Could you use the `/** ... */` (double asterisk) form of comments for class-documentation (and also drop the newline between comment and class declaration)?

javascript/ql/experimental/adaptivethreatmodeling/lib/experimental/adaptivethreatmodeling/EndpointCharacteristics.qll
```diff
@@ -0,0 +1,113 @@
...
+ * Endpoints that were identified as "TaintedPathSink" by the standard Javascript library are path injection sinks with
+ * maximal confidence.
+ */
+
+class TaintedPathSinkCharacteristic extends EndpointCharacteristic {
```
kaeluka, line 65: I *think* that we should be able to get away with making the implementations `private`.

javascript/ql/experimental/adaptivethreatmodeling/lib/experimental/adaptivethreatmodeling/EndpointCharacteristics.qll
```diff
@@ -0,0 +1,113 @@
...
+ * Defines a set of characteristics that a particular endpoint might have. This set of characteristics is used to make
+ * decisions about whether to include the endpoint in the training set and with what label, as well as whether to score
+ * the endpoint at inference time.
+ */
+
+import experimental.adaptivethreatmodeling.EndpointTypes
+import semmle.javascript.security.dataflow.SqlInjectionCustomizations
+private import semmle.javascript.security.dataflow.DomBasedXssCustomizations
+private import semmle.javascript.security.dataflow.NosqlInjectionCustomizations
+private import semmle.javascript.security.dataflow.TaintedPathCustomizations
```
kaeluka, line 13: At first, I thought importing all these 4 customizations modules from the same location might lead to inconsistent extensions of the class hierarchy being present at the same time, but the `CoreKnowledge` module already does the same thing..

kaeluka:
Yes, but they're no guarantee :+1:. One crucial test is `javascript/ql/experimental/adaptivethreatmodeling/test/endpoint_large_scale/ExtractEndpointDataTraining.qlref` together with the `ExtractEndpointDataTraining.expected` file in the same dir.

tiferet:
> I have left a few comments. Have you also started a performance evaluation for this change that you could link here?

It's [this experiment](https://github.com/github/codeql-dca-main/issues/8243) that I pinged you about yesterday, because I don't understand the failure errors

javascript/ql/experimental/adaptivethreatmodeling/lib/experimental/adaptivethreatmodeling/EndpointCharacteristics.qll
tiferet: My linter makes that change. Are there standard linter settings for the CodeQL repo that I should be using?

javascript/ql/experimental/adaptivethreatmodeling/lib/experimental/adaptivethreatmodeling/ATMConfig.qll
```diff
@@ -44,7 +45,14 @@ abstract class AtmConfig extends string {
    *
    * Holds if `sink` is a known sink of flow.
    */
-  predicate isKnownSink(JS::DataFlow::Node sink) { none() }
+  predicate isKnownSink(JS::DataFlow::Node sink) {
+    // If the list of characteristics includes positive indicators with maximal confidence for this class, then it's a
+    // known sink for the class.
```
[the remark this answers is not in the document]
tiferet: If they're for a different class, then that's fine -- that just means this endpoint isn't a sink for that other class. If they're low confidence that's fine too -- a high confidence indicator overrides a low one. If there's a high confidence characteristic that implies this endpoint is not a sink for this same class, that gets caught by the erroneous characteristics predicate. It means something is inconsistent in our logic, and a human needs to review it. I opened an issue yesterday to figure out where and how we should run that predicate.

javascript/ql/experimental/adaptivethreatmodeling/lib/experimental/adaptivethreatmodeling/EndpointCharacteristics.qll
```diff
@@ -0,0 +1,113 @@
+/**
+ * For internal use only.
+ *
+ * Defines a set of characteristics that a particular endpoint might have. This set of characteristics is used to make
+ * decisions about whether to include the endpoint in the training set and with what label, as well as whether to score
+ * the endpoint at inference time.
+ */
+
+import experimental.adaptivethreatmodeling.EndpointTypes
+import semmle.javascript.security.dataflow.SqlInjectionCustomizations
```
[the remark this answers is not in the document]
tiferet: Again, what's the advantage of a private import? Is the idea to prevent naming clashes down the line if someone imports this module?

javascript/ql/experimental/adaptivethreatmodeling/lib/experimental/adaptivethreatmodeling/EndpointCharacteristics.qll
tiferet: Yeah, I copied it from there, but I don't know if that's the best practice

javascript/ql/experimental/adaptivethreatmodeling/lib/experimental/adaptivethreatmodeling/EndpointCharacteristics.qll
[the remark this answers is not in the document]
jhelie: (regardless of the answer I think the point is about being consistent here)

jhelie:
Thanks @tiferet - I've had a look and LGTM, I don't have anything to add to @kaeluka's comments. My only suggestion would be regarding the below:

> Update: I've written the characteristics that will replace NotASinkReason and verified that I can reproduce the current selection of training examples

I think we should define a canary database (or a couple, to hit all the sink types) that we can all use and refer to ensure we're reproducing the set of endpoints. Concretely I'd like us to define somewhere:

```
Using databases `foo`, `bar` and `baz` we currently extract the following numbers of negative/{sinkType}/Unknown endpoints:... 

With this WIP design we are currently extract the following numbers: ...
```

While a couple of databases won't capture all edge cases the aim would be to have a simple targets so we can track progress and easily identify regressions.

javascript/ql/experimental/adaptivethreatmodeling/lib/experimental/adaptivethreatmodeling/EndpointCharacteristics.qll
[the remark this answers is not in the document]
kaeluka: I private import does not re-export declarations; it makes it so that modules depending on `EndpointCharacteristics.qll` may not depend on (eg.) `SqlInjectionCustomizations`. I'd consider the import of `SqlInjectionCustomizations` as an implementation detail and would therefore default to making it a private import.

javascript/ql/experimental/adaptivethreatmodeling/lib/experimental/adaptivethreatmodeling/EndpointCharacteristics.qll
kaeluka: Currently, outside code can depend on that class specifically; this invites breaking the abstraction (eg., by adding one method to `TaintedPathCharacteristic` only and using that in other modules, you're bound to re-introduce magic that we're trying to hide behind a neat interface).

javascript/ql/experimental/adaptivethreatmodeling/lib/experimental/adaptivethreatmodeling/ATMConfig.qll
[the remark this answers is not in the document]
kaeluka: > I opened an issue yesterday to figure out where and how we should run that predicate.

OK, I think it should be a consistency test like `FilteredTruePositives.ql` 👍

javascript/ql/experimental/adaptivethreatmodeling/lib/experimental/adaptivethreatmodeling/EndpointCharacteristics.qll
kaeluka: What are you using? The `Format document` command in VSCode? Or the `codeql` cli?

None of these make that change for me 😕

javascript/ql/experimental/adaptivethreatmodeling/lib/experimental/adaptivethreatmodeling/EndpointCharacteristics.qll
```diff
@@ -0,0 +1,113 @@
...
+private import semmle.javascript.security.dataflow.TaintedPathCustomizations
+
+abstract class EndpointCharacteristic extends string {
+  // The name of the characteristic, which should describe some characteristic of the endpoint that is meaningful for
+  // determining whether it's a sink and if so of which type
+  bindingset[this]
+  EndpointCharacteristic() { any() }
+
+  // Indicators with confidence at or above this threshold are considered to be high-confidence indicators.
+  float getHighConfidenceThreshold() { result = 0.8 }
```
jhelie, line 22: One question I had actually was what is the role of this predicate? I'm guessing it's a categorical selection wrapper on the more fine grain confidence floats, but why do we need it at this stage? (same question with the `Medium` one below)

adityasharad:
I think we could accomplish Jean's suggestion using QL unit tests. The test case could have sample code with all the possible endpoint types, and you can check in an `.expected` file for the expected output of the endpoint query on that test case.

javascript/ql/experimental/adaptivethreatmodeling/lib/experimental/adaptivethreatmodeling/EndpointCharacteristics.qll
tiferet: I see. So the non-abstract `EndpointCharacteristic`s should preferably be private, right? That way any special logic must live in an abstract base class that neatly encapsulates either all `EndpointCharacteristic`s or some intentional subset of them 👍

tiferet:
Regarding Jean and Aditya's testing suggestions, these QL tests already exist 👍 : e.g. `javascript/ql/experimental/adaptivethreatmodeling/test/endpoint_large_scale/ExtractEndpointDataTraining.qlref` together with the [ExtractEndpointDataTraining.expected](https://raw.githubusercontent.com/github/codeql/main/javascript/ql/experimental/adaptivethreatmodeling/test/endpoint_large_scale/ExtractEndpointDataTraining.expected) file

javascript/ql/experimental/adaptivethreatmodeling/lib/experimental/adaptivethreatmodeling/EndpointCharacteristics.qll
tiferet: You're right, I could have left these for a later PR. I use them in the endpoint selection code. For example, to implement logic such as "if the list of characteristics includes positive indicators with high confidence for this class, select this as a training sample belonging to the class". I put them in `EndpointCharacteristic` because I think the place that sets confidences for various types of endpoints should also define what we mean by "high confidence".

javascript/ql/experimental/adaptivethreatmodeling/lib/experimental/adaptivethreatmodeling/EndpointCharacteristics.qll
tiferet: Done in 99b75766f2295b1c39bd1fdf3c9e174c841ea744

javascript/ql/experimental/adaptivethreatmodeling/lib/experimental/adaptivethreatmodeling/EndpointCharacteristics.qll
tiferet: Oh, nice: It turns out that when I add the second asterisk the linter also gets rid of the pesky newline 😄 37862f77ffe0be8f525a6e1e991004c00441b560

tiferet:
I think I've addressed all the review comments 🏓 

I ran `endpoint_large_scale/ExtractEndpointData` and `endpoint_large_scale/ExtractEndpointDataTraining` locally and they run as fast as on `main`. Excluding DB extraction and query compilation,
- `ExtractEndpointData` takes 3.5-4 seconds on `main` and 3.6-3.9 seconds on this branch ✅ 
- `ExtractEndpointDataTraining` takes 3.6-3.9 seconds on `main` and 3.5-3.7 seconds on this branch ✅ 

What's left is to get the timing DCA tests to run ([latest attempt](https://github.com/github/codeql-dca-main/issues/8273)), plus a (hopefully) final review....

tiferet:
@kaeluka Using the latest CLI seems to have [solved the DCA failures](https://github.com/github/codeql-dca-main/issues/8273). I don't know how to read the resulting report. [The instructions](https://github.com/github/ml-ql-adaptive-threat-modeling/tree/main/docs/codeql_performance#locally-starting-experiment-observing-the-overhead-of-adding-the-atm-to-the-security-extended-query-suite) say `If timing/ATM-Threshold issue summary, when done, contains failures, the respective repositories have exceeded their permissible overhead quota`, but I can't find anything called `timing/ATM-Threshold`. If I understand [this table](https://github.com/github/codeql-dca-main/tree/data/tiferet/timing-sink-endpoint-characteristics5/reports#slow-command-times-per-source-and-command) correctly, though, it seems like we have a problem?

jhelie:
> I can't find anything called timing/ATM-Threshold. If I understand [this table](https://github.com/github/codeql-dca-main/tree/data/tiferet/timing-sink-endpoint-characteristics5/reports#slow-command-times-per-source-and-command) correctly, though, it seems like we have a problem?

@tiferet see the [Note on performance](https://github.com/github/ml-ql-adaptive-threat-modeling/tree/main/docs/codeql_performance#note-on-performance-measurement) we added to the instructions recently: - the table you linked to is not the one used to decide whether we meet our KPI (as it's using `relative` and no `relative_with_overhead`)
- the fact you do not see any ATM-threshold summary in the experiment issue suggests we do not have a problem (I think we don't show anything if there's no problem as real estate is at a premium on the DCA issue)

cc @esbena to confirm

jhelie:
> Regarding Jean and Aditya's testing suggestions, these QL tests already exist 

That's great - to clarify I had in mind something more basic that just checked on the endpoints, without coupling it to format as I thought we might have to break that format during implementation of the design (if only temporarily, as discussed we'll need to write glue-code to minimise disruption to the pipeline inputs). But we can proceed with these tests for now and cross that bridge when we get to it.

javascript/ql/experimental/adaptivethreatmodeling/lib/experimental/adaptivethreatmodeling/EndpointCharacteristics.qll
jhelie: Should we introduce them in the PR that will use them then? I know you have written more code locally, but for us that will make things a little easier to follow along.

tiferet:
> > I can't find anything called timing/ATM-Threshold. If I understand [this table](https://github.com/github/codeql-dca-main/tree/data/tiferet/timing-sink-endpoint-characteristics5/reports#slow-command-times-per-source-and-command) correctly, though, it seems like we have a problem?
> 
> @tiferet see the [Note on performance](https://github.com/github/ml-ql-adaptive-threat-modeling/tree/main/docs/codeql_performance#note-on-performance-measurement) we added to the instructions recently:
> 
> * the table you linked to is not the one used to decide whether we meet our KPI (as it's using `relative` and no `relative_with_overhead`)
> * the fact you do not see any ATM-threshold summary in the experiment issue suggests we do not have a problem (I think we don't show anything if there's no problem as real estate is at a premium on the DCA issue)
> 
> cc @esbena to confirm

But the absolute time difference is sometimes big (e.g. 191.7 seconds). That's why I assumed the `relative_with_overhead` would fail too.

javascript/ql/experimental/adaptivethreatmodeling/lib/experimental/adaptivethreatmodeling/EndpointCharacteristics.qll
tiferet: IMHO it's not worth deleting them from this PR just to add them in the next PR 🤷

jhelie: ok but in the future please do not introduce concepts not discussed beforehand and not used in the PR you open.

jhelie:
> But the absolute time difference is sometimes big (e.g. 191.7 seconds). That's why I assumed the relative_with_overhead would fail too.

I only had a quick look but the diff is only big for one source and from memory even then `relative_with_overhead` would be below the threshold - it's easy enough to do the maths though

esbena:
The ATM thresholds have not been crossed. You can confirm by looking in reports/any.md and observing the ToC:

| Title                                                                                                                                          | Interesting rows | Undecided rows | Bad rows | Good rows | min    | max    |
| ---------------------------------------------------------------------------------------------------------------------------------------------- | ---------------- | -------------- | -------- | --------- | ------ | ------ |
| [Alert count using manual result classifications, per query](#alert-count-using-manual-result-classifications-per-query)                       |                  |                |          |           |        |        |
| [Alert count using manual result classifications, per source and query](#alert-count-using-manual-result-classifications-per-source-and-query) |                  |                |          |           |        |        |
| [Analysis time, per source (ATM threshold)](#analysis-time-per-source-atm-threshold)                                                           |                  |                |          |           |        |        |

^ nothing interesting.

kaeluka (dismissed):

tiferet:
> The ATM thresholds have not been crossed. You can confirm by looking in reports/any.md and observing the ToC

@esbena Thank you for the clarification about where to look! I've updated our instructions to reflect this ❤️

henrymercer (commented):
This looks generally good!  A few comments.

javascript/ql/experimental/adaptivethreatmodeling/lib/experimental/adaptivethreatmodeling/ATMConfig.qll
```diff
@@ -44,7 +45,14 @@ abstract class AtmConfig extends string {
    *
    * Holds if `sink` is a known sink of flow.
    */
-  predicate isKnownSink(JS::DataFlow::Node sink) { none() }
+  predicate isKnownSink(JS::DataFlow::Node sink) {
```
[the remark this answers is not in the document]
tiferet: Fixed in 787352f443452a83de9bc73f9daa926f99499d56

javascript/ql/experimental/adaptivethreatmodeling/lib/experimental/adaptivethreatmodeling/EndpointCharacteristics.qll
```diff
@@ -0,0 +1,109 @@
...
+
+import experimental.adaptivethreatmodeling.EndpointTypes
+private import semmle.javascript.security.dataflow.SqlInjectionCustomizations
+private import semmle.javascript.security.dataflow.DomBasedXssCustomizations
+private import semmle.javascript.security.dataflow.NosqlInjectionCustomizations
+private import semmle.javascript.security.dataflow.TaintedPathCustomizations
+
+abstract class EndpointCharacteristic extends string {
+  // The name of the characteristic, which should describe some characteristic of the endpoint that is meaningful for
+  // determining whether it's a sink and if so of which type
```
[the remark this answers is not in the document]
tiferet: Fixed in 787352f443452a83de9bc73f9daa926f99499d56

javascript/ql/experimental/adaptivethreatmodeling/lib/experimental/adaptivethreatmodeling/EndpointCharacteristics.qll
```diff
@@ -0,0 +1,109 @@
...
+  // _isn't_ a member of the class?
+  // confidence: A number in [0, 1], which tells us how strong an indicator this characteristic is for the endpoint
+  // belonging / not belonging to the given class.
+  abstract predicate getImplications(
+    EndpointType endpointClass, boolean isPositiveIndicator, float confidence
+  );
+}
+
+/**
+ * Endpoints that were identified as "DomBasedXssSink" by the standard Javascript library are XSS sinks with maximal
```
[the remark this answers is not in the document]
tiferet: Fixed in 787352f443452a83de9bc73f9daa926f99499d56

kaeluka (approved):
all new changes are benign

javascript/ql/experimental/adaptivethreatmodeling/lib/experimental/adaptivethreatmodeling/EndpointCharacteristics.qll
```diff
@@ -0,0 +1,110 @@
...
+   * Holds for endpoints that have this characteristic. This predicate contains the logic that applies characteristics
+   * to the appropriate set of dataflow nodes.
+   */
+  abstract predicate getEndpoints(DataFlow::Node n);
+
+  /**
+   * This predicate describes what the characteristic tells us about an endpoint.
+   *
+   *  Params:
+   *  endpointClass: Class 0 is the negative class. Each positive int corresponds to a single sink type.
```
aeisenberg, line 34: I know what you mean here, but it took me a second since 0 is not negative. I don't have any suggestions on improvement, though.

javascript/ql/experimental/adaptivethreatmodeling/lib/experimental/adaptivethreatmodeling/EndpointCharacteristics.qll
```diff
@@ -0,0 +1,110 @@
...
+   *  isPositiveIndicator: Does this characteristic indicate this endpoint _is_ a member of the class, or that it
+   *  _isn't_ a member of the class?
```
aeisenberg, line 36: Minor:

```suggestion
   *  isPositiveIndicator: If true, this endpoint is a member of the class.
```

javascript/ql/experimental/adaptivethreatmodeling/lib/experimental/adaptivethreatmodeling/EndpointCharacteristics.qll
```diff
@@ -0,0 +1,110 @@
...
+   *  confidence: A number in [0, 1], which tells us how strong an indicator this characteristic is for the endpoint
+   *  belonging / not belonging to the given class.
```
aeisenberg, line 38: I'm not confident that this comment change makes things better. It's mostly for my own understanding.

```suggestion
   *  confidence: A float in [0, 1], which tells us how strong an indicator this characteristic is for the endpoint
   *  belonging / not belonging to the given class. 0 means complete confidence that this characteristic _does not_ indicate belonging to this endpoint. And 1 means complete confidence that this characteristic _does_ belong..
```

aeisenberg (approved):
Just some minor comment changes. I'm not too familiar with this area, so I'm not even sure if these suggestions are correct. Feel free to take or leave them.

PR #11058, opened 2022-10-31
Ruby: Model various ActionController methods

hmac:
`ActionController` methods have access to a `logger` method which returns an instance of `ActiveSupport::Logger`.

The first commit moves the existing `ActionController.ql` test to be inside the `action_controller` directory, so that is no longer affected by changes to other test fixtures. I think we should probably do this for the remaining tests (ActionDispatch, ActionView, Core, GraphQL, PosixSpawn) at some point.

Later commits model `send_data` and `body_stream`.

[Evaluation](https://github.com/github/codeql-dca-main/blob/data/hmac/pr-11058-3c0a39__nightly__nightly/reports/alert-comparison.md) shows new TPs for log injection, with one FP due to us not considering `Object#inspect` to be a sanitizer. I've added a commit to fix this FP.

### DIFF
--- a/ruby/ql/lib/codeql/ruby/frameworks/ActionController.qll
+++ b/ruby/ql/lib/codeql/ruby/frameworks/ActionController.qll
@@ -559,11 +559,7 @@ private class SendFile extends FileSystemAccess::Range, Http::Server::HttpRespon
  */
 class SendDataCall extends DataFlow::CallNode, Http::Server::HttpResponse::Range {
   SendDataCall() {
-    this.getMethodName() = "send_data" and
-    (
-      this.asExpr().getExpr() instanceof ActionControllerContextCall or
-      this.getReceiver().asExpr().getExpr() instanceof Response::ResponseCall
-    )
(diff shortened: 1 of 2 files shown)


nickrolfe (dismissed):
Changes LGTM. I've started another DCA run, since the last one had some (presumably transient) failures.

erik-krogh:
> That run shows a stage timing (?) slowdown for `Regexp::regExpSource(rb/incomplete-hostname-regexp[11], 7/7)` on `gitlabhq/gitlabhq`. Should we be worried about that?

That is a fluke that you shouldn't worry about.   
Your PR doesn't affect that predicate (I checked), so there's no way your PR caused a regression.  

But you have merge conflicts beyond the test-files that GitHub is complaining about (e.g. `private class ActionControllerContextCall` has disappeared on `main`), so you need to update the PR to reflect those changes.

erik-krogh (approved):
[A new evaluation](https://github.com/github/codeql-dca-main/tree/data/erik-krogh/pr-11058-ed3270__nightly__code-scanning/reports) still looks good. 

👍

PR #11060, opened 2022-11-01
Data flow: Restrict public `PathNode`s to those that may reach a sink

hvitved:
Just like `PathNode::getASuccessor` [is restricted to nodes that can reach a sink](https://github.com/github/codeql/pull/9316), it also makes sense to only expose nodes that can reach a sink. This means that the exposed `PathNode` class and the `nodes` predicate contain the same elements.

The reason for the updated C++ test output is because `DefaultTaintTracking.qll` did not previously restrict the `nodes` relation to nodes that can reach a sink.

### DIFF
--- a/cpp/ql/lib/experimental/semmle/code/cpp/ir/dataflow/internal/DataFlowImpl.qll
+++ b/cpp/ql/lib/experimental/semmle/code/cpp/ir/dataflow/internal/DataFlowImpl.qll
@@ -2874,54 +2874,16 @@ private class AccessPathCons1 extends AccessPath, TAccessPathCons1 {
   }
 }
 
-/**
- * A `Node` augmented with a call context (except for sinks), an access path, and a configuration.
- * Only those `PathNode`s that are reachable from a source are generated.
- */
-class PathNode extends TPathNode {
-  /** Gets a textual representation of this element. */
-  string toString() { none() }
-
-  /**
-   * Gets a textual representation of this element, including a textual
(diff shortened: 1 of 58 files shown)


jketema (approved):
C++ expected test output changes 👍

cpp/ql/lib/experimental/semmle/code/cpp/ir/dataflow/internal/DataFlowImpl.qll
```diff
@@ -2986,14 +2958,59 @@ private predicate directReach(PathNodeImpl n) {
...
+private predicate pathSucc(PathNodeImpl n1, PathNodeImpl n2) {
   n1.getANonHiddenSuccessor() = n2 and directReach(n2)
 }
 
-private predicate pathSuccPlus(PathNode n1, PathNode n2) = fastTC(pathSucc/2)(n1, n2)
+private predicate pathSuccPlus(PathNodeImpl n1, PathNodeImpl n2) = fastTC(pathSucc/2)(n1, n2)
+
+/**
+ * A `Node` augmented with a call context (except for sinks), an access path, and a configuration.
+ * Only those `PathNode`s that are reachable from a source, and which can reach a sink, are generated.
```
aschackmull, line 2972: This is not entirely accurate. Nodes that are used in a subpath are also included.

aschackmull (approved):
One minor nit re. a qldoc, otherwise LGTM.

PR #11062, opened 2022-11-01
Swift: Rename ECB-Encryption directory

geoffw0:
Rename the "ECB-Encryption" query directory to "CWE-327", for consistency with how the other security query directories are named.

@karimhamdanali do you have any thoughts about this?


karimhamdanali:
I'm ok with that name change. I just followed the same naming convention that we have for C#, but I agree that the suggested change is more consistent with what we have.

MathiasVP (approved):

PR #11064, opened 2022-11-01
Kotlin: handle default parameter values inherited from an overridden function

smowton:


tamasvajk (commented):
Looks okay to me, but some CI tests are failing.

tamasvajk (approved):

PR #11065, opened 2022-11-01
ATM: add PR check running boosted queries

jhelie:
As discussed this clones a hard-coded open-source repo and builds a JavaScript database from the latest commit.

POC run [here](https://github.com/github/codeql/actions/runs/3369358110)

### DIFF
--- a/.github/workflows/atm-check-queries-run.yml
+++ b/.github/workflows/atm-check-queries-run.yml
@@ -1,19 +1,15 @@
 name: ATM Check Queries Run
 
 env:
-  AZURE_STORAGE_URL: "https://atmcodeqldata.blob.core.windows.net"
-  DB_NAME: "AmanSultanBaig/SignIn-SignUp-System-with-Nodejs"
   DB_PATH: test_db
-  MODEL_BULDING_PACK_PATH: javascript/ql/experimental/adaptivethreatmodeling/modelbuilding
-  QUERY_SUITE: javascript/ql/experimental/adaptivethreatmodeling/src/codeql-suites/javascript-atm-code-scanning.qls
+  ATM_MODEL_PACK: javascript/ql/experimental/adaptivethreatmodeling/src
+  QUERY_SUITE: codeql-suites/javascript-atm-code-scanning.qls
 
 on:
   pull_request:
     paths:
-      - "javascript/ql/experimental/adaptivethreatmodeling/modelbuilding/codeql-pack.lock.yml"
-      - "javascript/ql/experimental/adaptivethreatmodeling/modelbuilding/qlpack.yml"
-      - "javascript/experimental/adpativethreatmodeling/src/qlpack.yml"
-      - "javascript/experimental/adpativethreatmodeling/src/codeql-pack.lock.yml"
+      - ".github/workflows/atm-check-queries-run.yml"
+      - "javascript/ql/experimental/adaptivethreatmodeling/**"
   workflow_dispatch:
 
 jobs:
@@ -22,8 +18,6 @@ jobs:
 
     steps:
       - uses: actions/checkout@v3
-        with:
-          path: codeql-lib
 
       - name: Install CodeQL CLI
         env:
@@ -32,48 +26,31 @@ jobs:
           gh extensions install github/gh-codeql
           gh codeql download
 
-      - name: Download model pack
+      - name: Install ATM model pack
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
-          STORAGE_SAS: ${{ secrets.AZURE_BLOB_STORAGE_ATMCODEQLDATA_SAS }}
         run: |
-          echo "::group::Download ATM model pack"
           set -exu
+          
+          # Install ATM model pack
+          gh codeql pack install ${ATM_MODEL_PACK}
 
-          # Get pack version and checksum
-          pack_version=$(yq '.dependencies.codeql/javascript-experimental-atm-model' ./codeql-lib/${MODEL_BULDING_PACK_PATH}/qlpack.yml )
-          model_checksum="${pack_version##*.}"
-          echo "Will use pack model ${pack_version} with model checksum ${model_checksum}."
-
-          # Download the model to the package cache
-          tmp_dir=$(mktemp -d)
-          gh codeql pack download codeql/javascript-experimental-atm-model@${pack_version}
+          # Retrieve model checksum
+          model_checksum=$(gh codeql resolve extensions ${ATM_MODEL_PACK}/${QUERY_SUITE} | jq -r '.models[0].checksum')
 
           # Trust the model so that we can use it in the ATM boosted queries
           mkdir -p "$HOME/.config/codeql"
           echo "--insecurely-execute-ml-model-checksums ${model_checksum}" >> "$HOME/.config/codeql/config"
-          echo "::endgroup::"
 
       - name: Create test DB
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
         run: |
-          echo "::group::Create test database"
-          gh repo clone ${DB_NAME} -- --depth 1
-          gh codeql database create ${DB_PATH} --language javascript 
-          echo "::endgroup::"
+          gh codeql database create ${RUNNER_TEMP}/${DB_PATH} --source-root config/atm/ --language javascript 
 
       - name: Run ATM query suite
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
         run: |
-          echo "::group::Run boosted query suite"
-          gh codeql database run-queries -vv -- ${DB_PATH} codeql-lib/${QUERY_SUITE}
-          if [[ $? -ne 0 ]]; then
-            echo "Failed to run the ATM query suite."
-            exit 1
-          else
-            echo "Successfully run ATM query suite."
-          fi
-          echo "::endgroup::"
+          gh codeql database run-queries -vv -- ${RUNNER_TEMP}/${DB_PATH} ${ATM_MODEL_PACK}/${QUERY_SUITE}
       
--- a/.github/workflows/atm-check-queries-run.yml
+++ b/.github/workflows/atm-check-queries-run.yml
@@ -1,13 +1,79 @@
 name: ATM Check Queries Run
 
-# This check is required, therefore we must run it on all PRs, even if only Markdown has changed.
+env:
+  AZURE_STORAGE_URL: "https://atmcodeqldata.blob.core.windows.net"
+  DB_NAME: "AmanSultanBaig/SignIn-SignUp-System-with-Nodejs"
+  DB_PATH: test_db
+  MODEL_BULDING_PACK_PATH: javascript/ql/experimental/adaptivethreatmodeling/modelbuilding
+  QUERY_SUITE: javascript/ql/experimental/adaptivethreatmodeling/src/codeql-suites/javascript-atm-code-scanning.qls
+
 on:
+  pull_request:
+    paths:
+      - "javascript/ql/experimental/adaptivethreatmodeling/modelbuilding/codeql-pack.lock.yml"
+      - "javascript/ql/experimental/adaptivethreatmodeling/modelbuilding/qlpack.yml"
+      - "javascript/experimental/adpativethreatmodeling/src/qlpack.yml"
+      - "javascript/experimental/adpativethreatmodeling/src/codeql-pack.lock.yml"
   workflow_dispatch:
 
 jobs:
-  hello-world:
+  run-atm-queries:
     runs-on: ubuntu-latest
 
     steps:
-      - name: foo
-        run: echo "Hello world"
+      - uses: actions/checkout@v3
+        with:
+          path: codeql-lib
+
+      - name: Install CodeQL CLI
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: |
+          gh extensions install github/gh-codeql
+          gh codeql download
+
+      - name: Download model pack
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          STORAGE_SAS: ${{ secrets.AZURE_BLOB_STORAGE_ATMCODEQLDATA_SAS }}
+        run: |
+          echo "::group::Download ATM model pack"
+          set -exu
+
+          # Get pack version and checksum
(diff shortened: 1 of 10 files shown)


.github/workflows/atm-check-queries-run.yml
```diff
@@ -1,13 +1,79 @@
...
+          set -exu
+
+          # Get pack version and checksum
+          pack_version=$(yq '.dependencies.codeql/javascript-experimental-atm-model' ./codeql-lib/${MODEL_BULDING_PACK_PATH}/qlpack.yml )
+          model_checksum="${pack_version##*.}"
+          echo "Will use pack model ${pack_version} with model checksum ${model_checksum}."
+
+          # Download the model to the package cache
+          tmp_dir=$(mktemp -d)
+          gh codeql pack download codeql/javascript-experimental-atm-model@${pack_version}
```
henrymercer, line 50: We can use the standard pack installation mechanism `gh codeql pack install javascript/ql/experimental/adaptivethreatmodeling/src` to do all of this :)

.github/workflows/atm-check-queries-run.yml
```diff
@@ -1,13 +1,79 @@
...
+        run: |
+          gh extensions install github/gh-codeql
+          gh codeql download
+
+      - name: Download model pack
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          STORAGE_SAS: ${{ secrets.AZURE_BLOB_STORAGE_ATMCODEQLDATA_SAS }}
+        run: |
+          echo "::group::Download ATM model pack"
```
henrymercer, line 40: Groups are great for Actions, but in workflows each step will already be grouped in the UI, so this and others in the workflow can be removed.

.github/workflows/atm-check-queries-run.yml
```diff
@@ -1,13 +1,79 @@
...
+          mkdir -p "$HOME/.config/codeql"
+          echo "--insecurely-execute-ml-model-checksums ${model_checksum}" >> "$HOME/.config/codeql/config"
+          echo "::endgroup::"
+
+      - name: Create test DB
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: |
+          echo "::group::Create test database"
+          gh repo clone ${DB_NAME} -- --depth 1
```
henrymercer, line 62: Hmm, `--` usually indicates the end of option processing, so I'm confused what that's doing here.

.github/workflows/atm-check-queries-run.yml
```diff
@@ -1,13 +1,79 @@
...
+          gh codeql database create ${DB_PATH} --language javascript
```
henrymercer, line 63: I think you probably want to specify the source root as wherever you downloaded the database, otherwise we'll include everything in the current working directory in the database.

.github/workflows/atm-check-queries-run.yml
```diff
@@ -1,13 +1,79 @@
...
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: |
+          echo "::group::Run boosted query suite"
+          gh codeql database run-queries -vv -- ${DB_PATH} codeql-lib/${QUERY_SUITE}
+          if [[ $? -ne 0 ]]; then
+            echo "Failed to run the ATM query suite."
+            exit 1
+          else
+            echo "Successfully run ATM query suite."
+          fi
```
henrymercer, line 77: Nit: This could probably be removed in the interests of brevity.

henrymercer (commented):
Thanks for doing this, this will be a really useful PR check!

.github/workflows/atm-check-queries-run.yml
```diff
@@ -1,13 +1,79 @@
 name: ATM Check Queries Run
 
-# This check is required, therefore we must run it on all PRs, even if only Markdown has changed.
+env:
+  AZURE_STORAGE_URL: "https://atmcodeqldata.blob.core.windows.net"
```
jhelie, line 4: I should delete this reference too

.github/workflows/atm-check-queries-run.yml
```diff
@@ -1,13 +1,80 @@
...
+  DB_NAME: "AmanSultanBaig/SignIn-SignUp-System-with-Nodejs"
```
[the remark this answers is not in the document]
jhelie: > It's good to avoid making our PR checks depend on repos that we don't control

Well yes I agree and this is exactly the problem I brought up and I thought we had agreed on this method, but I must have misunderstood.

> I would suggest we instead copy the small test project

I'm not sure I followed the suggestion, do you mind pushing a commit?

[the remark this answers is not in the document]
henrymercer: If I add a commit, I won't be able to review this PR any more due to the new branch protection rules :)

I was thinking we could check the files under [this directory](https://github.com/github/codeql-action/tree/main/tests/ml-powered-queries-repo) into this repo under a new `config/atm` directory (within https://github.com/github/codeql/tree/main/config).  Then when creating the database we can run `codeql database create --source-root config/atm`.

.github/workflows/atm-check-queries-run.yml
jhelie: I guess what I'm asking is: will it update our config file to trust the checksum?

.github/workflows/atm-check-queries-run.yml
```diff
@@ -1,13 +1,79 @@
...
+          echo "::group::Download ATM model pack"
+          set -exu
+
+          # Get pack version and checksum
+          pack_version=$(yq '.dependencies.codeql/javascript-experimental-atm-model' ./codeql-lib/${MODEL_BULDING_PACK_PATH}/qlpack.yml )
+          model_checksum="${pack_version##*.}"
+          echo "Will use pack model ${pack_version} with model checksum ${model_checksum}."
+
+          # Download the model to the package cache
+          tmp_dir=$(mktemp -d)
```
jhelie, line 49: (that line needs to go)

.github/workflows/atm-check-queries-run.yml
henrymercer: Using the standard interface for installing the model makes the workflow more robust (and is also more readable).  If the behind-the-scenes mechanism for installing the model changes, then the workflow doesn't break.  The workflow is currently not robust to the name of the model pack changing for instance, or the addition of a new pack dependency.

You're right that we'll still need to trust the checksum.  We should grab the checksum from `model.properties` instead of the version number though, since the version number of a released pack will not contain the checksum (it'll just be `0.4.0` for example).

jhelie: What is the recommended way to find the model.properties file? resolve using the query suite? or assume there's a single ml model?

henrymercer: I'd recommend `gh codeql resolve extensions javascript/ql/experimental/adaptivethreatmodeling/src/NosqlInjectionATM.ql | jq -r '.models[0].checksum'`.  It's fine to make the assumption that there's a single ML model — that is unlikely to change and won't happen underneath us.

jhelie: I think we should avoid using a random query, it's more robust to change behind the scenes to use the suite.

.github/workflows/atm-check-queries-run.yml
jhelie: done

.github/workflows/atm-check-queries-run.yml
jhelie: any idea what I'm missing here? https://github.com/github/codeql/actions/runs/3371044939/jobs/5592734461

jhelie: ugh it seems `actions/checkout` doesn't create a specific folder

.github/workflows/atm-check-queries-run.yml
```diff
@@ -1,13 +1,58 @@
...
-      - name: foo
-        run: echo "Hello world"
+      - uses: actions/checkout@v3
+
+      - name: Install CodeQL CLI
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: |
+          gh extensions install github/gh-codeql
+          gh codeql download
```
aeisenberg, line 27: Is there a reason why we are downloading codeql instead of using the version in the toolcache?

.github/workflows/atm-check-queries-run.yml
```diff
@@ -1,13 +1,80 @@
...
-  hello-world:
+  run-atm-queries:
     runs-on: ubuntu-latest
 
     steps:
-      - name: foo
-        run: echo "Hello world"
+      - uses: actions/checkout@v3
+        with:
+          path: codeql-lib
```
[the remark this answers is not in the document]
jhelie: This will checkout the repo in the current dir. I don't see how this is desirable, are we sure about this?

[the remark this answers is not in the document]
henrymercer: It's desirable for standardisation and because other Actions can make the assumption that they have a checkout of the repo in the expected place.

.github/workflows/atm-check-queries-run.yml
aeisenberg: Is it to ensure we are always running with the latest codeql version?

.github/workflows/atm-check-queries-run.yml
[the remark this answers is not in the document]
jhelie: to make sure I understand: we do ditch this approach for workflows involving several checkouts, right?

.github/workflows/atm-check-queries-run.yml
jhelie: Yes

.github/workflows/atm-check-queries-run.yml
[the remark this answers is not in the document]
jhelie: and does it mean we need to worry about where we create our database? there's no guarantee that `DB_PATH` doesn't collide with a directory from the checkout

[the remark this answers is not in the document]
henrymercer: > to make sure I understand: we do ditch this approach for workflows involving several checkouts, right?

If the workflow needs to check out another repo, we'd typically check out `github/codeql` under the default path and check that the other repo under a different path.

> and does it mean we need to worry about where we create our database? there's no guarantee that DB_PATH doesn't collide with a directory from the checkout

Good point — to avoid this it's good practice to put the database under a temporary directory (see `$RUNNER_TEMP`).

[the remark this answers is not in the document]
jhelie: The same $RUNNER_TEMP logic really ought to apply to the other checked out repos - or any directory we'd like to create for that matter. I don't have enough knowledge to have a strong opinion but, on the basis of the workflows I interact with, this new approach is a net regression.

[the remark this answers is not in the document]
jhelie: (I initially thought that the default path would be the name of the repo, similar to the `gh repo clone` command...)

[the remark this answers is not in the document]
henrymercer: > The same `$RUNNER_TEMP` logic really ought to apply to the other checked out repos - or any directory we'd like to create for that matter. I don't have enough knowledge to have a strong opinion but, on the basis of the workflows I interact with, this new approach is a net regression.

The only difficulty with that is it becomes slightly harder to reference where the repo is, so you'd have to write `working-directory: ${{ env.CODEQL_LIB_PATH }}` for example.

I'll have to disagree with you on the net regression.  The Action is now considerably terser and easier to read in my opinion.

henrymercer (approved):

.github/workflows/atm-check-queries-run.yml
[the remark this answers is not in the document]
jhelie: It's fine to disagree - on this workflow the workflow is actually more verbose and convoluted: instead of checking out in `foo` and then creating a `bar` without worrying we have to use temporary folders. And the more repos we need to checkout / directories to check out the worse it will get. But I admit I don't understand what you mean initially with the expected benefit:

> It's desirable for standardisation and because other Actions can make the assumption that they have a checkout of the repo in the expected place.

Did you mean composite Actions? In the case of this workflow the drawbacks jump out but I'm missing the improvements.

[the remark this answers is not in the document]
jhelie: (having said that `$RUNNER_TEMP` is very convenient - but the inconsistency of the approach remains in my opinion)

jhelie:
New POC run [here](https://github.com/github/codeql/actions/runs/3372419988/jobs/5595820484)

jhelie:
@henrymercer all requested changes are in and the new test run completed but I need a new ✅ since I had to rebase 😞

.github/workflows/atm-check-queries-run.yml
```diff
@@ -1,13 +1,56 @@
...
+
+          # Trust the model so that we can use it in the ATM boosted queries
+          mkdir -p "$HOME/.config/codeql"
+          echo "--insecurely-execute-ml-model-checksums ${model_checksum}" >> "$HOME/.config/codeql/config"
+
+      - name: Create test DB
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: |
+          gh codeql database create ${RUNNER_TEMP}/${DB_PATH} --source-root config/atm/ --language javascript
```
henrymercer, line 49: Minor: It'd be a bit more intuitive to set `DB_PATH=${RUNNER_TEMP}/test_db` here and then add that to the global env, so we don't have to reference the DB as `${RUNNER_TEMP}/${DB_PATH}`.

henrymercer (approved):

.github/workflows/atm-check-queries-run.yml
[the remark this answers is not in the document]
henrymercer: > Did you mean composite Actions? In the case of this workflow the drawbacks jump out but I'm missing the improvements.

The improvements are most noticeable when using composite Actions, but I think there's a small benefit in standardisation having the repo checked out in the same place — it reduces cognitive load a tiny bit.

.github/workflows/atm-check-queries-run.yml
jhelie: Yes, that would require a separate step though. The new branch protection rules mean it's not worth our time at this point.

henrymercer: I think we could do it in this step:

```
DB_PATH=${RUNNER_TEMP}/test_db
gh codeql database create "${DB_PATH}" --source-root config/atm/ --language javascript
echo "DB_PATH=${DB_PATH}" >> "${GITHUB_ENV}"
```

Your call whether to do it or not since it's minor.

PR #11066, opened 2022-11-01
SSA: Deprecate `*NoUncertainReads` predicates

hvitved:
These predicates are currently only used by C#, and since I plan to also expose phi-read nodes, we would have to expose 4 versions of the same predicate (with/without phi-reads, with/without uncertain reads), and one could argue that we should then also have with/without uncertain writes, which just doesn't scale.

Instead, it will be up to the users of the shared SSA library to post-filter the use-use relations.


MathiasVP (approved):
LGTM!

PR #11067, opened 2022-11-01
ATM: add hello world workflow for development of model integration te…

jhelie:
For testing of this workflow via workflow dispatch on a branch as it is implemented.


.github/workflows/atm-run-model-integration-tests.yml
```diff
@@ -0,0 +1,12 @@
+name: ATM Run Model Integration Tests
```
henrymercer, line 1: These should eventually run on PRs, so I think we can drop the `-run` from the name.

henrymercer: Could you drop it from the file name too?  Thanks.

jhelie: done

henrymercer (approved):

PR #5704, opened 2021-04-16
Java: Regex injection

edvraa:

PR #11070, opened 2022-11-01
Java: Promote regex injection query from experimental 

jcogs33:
This PR promotes https://github.com/github/codeql/pull/5704 from experimental.


java/ql/lib/semmle/code/java/security/regexp/RegexInjection.qll
```diff
@@ -0,0 +1,51 @@
...
+/** A sanitizer for untrusted user input used to construct regular expressions. */
+abstract class RegexInjectionSanitizer extends DataFlow::ExprNode { }
+
+/** A method call that takes a regular expression as an argument. */
+private class DefaultRegexInjectionSink extends RegexInjectionSink {
+  DefaultRegexInjectionSink() {
+    exists(string kind |
+      kind.matches(["regex-use[]", "regex-use[f1]", "regex-use[f-1]", "regex-use[-1]", "regex-use"]) and
+      sinkNode(this, kind)
+    )
```
atorralba, line 20: I you're not using wildcards (still curious why `regex-use%` doesn't work), you don't need `matches` (and then the `exists` isn't really needed either):

```suggestion
    sinkNode(this, ["regex-use[]", "regex-use[f1]", "regex-use[f-1]", "regex-use[-1]", "regex-use"])
```

java/ql/lib/semmle/code/java/security/regexp/RegexInjection.qll
```diff
@@ -0,0 +1,51 @@
...
+ * or escape sequences no special meaning.
+ */
+private class PatternLiteralFlag extends RegexInjectionSanitizer {
+  PatternLiteralFlag() {
+    exists(MethodAccess ma, Method m, Field field | m = ma.getMethod() |
+      ma.getArgument(0) = this.asExpr() and
+      m.getDeclaringType() instanceof TypeRegexPattern and
+      m.hasName("compile") and
+      field instanceof PatternLiteralField and
+      ma.getArgument(1) = field.getAnAccess()
```
atorralba, line 48: You can remove the `instanceof` by using the stricter type directly in `exists`:

```suggestion
    exists(MethodAccess ma, Method m, PatternLiteralField field | m = ma.getMethod() |
      ma.getArgument(0) = this.asExpr() and
      m.getDeclaringType() instanceof TypeRegexPattern and
      m.hasName("compile") and
      ma.getArgument(1) = field.getAnAccess()
```

java/ql/test/query-tests/security/CWE-730/RegexInjectionTest.ql
```diff
@@ -0,0 +1,21 @@
+import java
+import TestUtilities.InlineExpectationsTest
+import semmle.code.java.security.regexp.RegexInjectionQuery
+
+//import semmle.code.java.security.regexp.PolynomialReDoSQuery
```
atorralba, line 5: This can be removed.

```suggestion
```

atorralba (commented):
Mostly LGTM, with a couple of minor comments. Curious about why exactly we couldn't use the `regex-use[*]` format for the new sinks so that the existing ReDoS queries could benefit from them.

java/ql/lib/semmle/code/java/security/regexp/RegexInjection.qll
joefarebrother: Some of the regex-use sinks are specific to the ReDoS query; in particular the `Predicate.test` sink is likely to cause many false positives if included. Which is probably the reason regex-use[0] is being excluded here. 

The best solution is probably to exclude that sink specifically - perhaps by giving it a seperate kind that only the ReDoS query checks for.

atorralba: Ah, makes sense, thanks! I would be surprised if more queries used these sinks aside from ReDos and regex injection, but if you think it makes sense we can create a special sink kind for `Predicate.test`. Specially if `regex-use[0]` includes other sinks that are valid for this query.

java/ql/lib/semmle/code/java/security/regexp/RegexInjection.qll
jcogs33: Updated in https://github.com/github/codeql/pull/11070/commits/f64572d432251f62d05755816481e476563d8e1a.

java/ql/test/query-tests/security/CWE-730/RegexInjectionTest.ql
jcogs33: Updated in https://github.com/github/codeql/pull/11070/commits/f64572d432251f62d05755816481e476563d8e1a.

java/ql/lib/semmle/code/java/security/regexp/RegexInjection.qll
jcogs33: Updated in https://github.com/github/codeql/pull/11070/commits/13decd38d90d30449f0a875278c35d8a2d9f3772, and made a follow-up issue as discussed.

hubwriter (approved):
Docs changes look good. 👍

atorralba (approved):
Nicely done!

jcogs33:
> Docs changes look good. 👍

Thanks for the review @hubwriter! Can I confirm if you were able to review the entire contents of the `RegexInjection.qhelp` file and not just the changes in this file? The original parts of this file did not go through a docs review, so I just want to confirm that the full file is good to go.

hubwriter (approved):
Thanks @jcogs33 - you're right, I only reviewed the changes.
I've reviewed the whole file now. Just the one comment.

java/ql/src/Security/CWE/CWE-730/RegexInjection.qhelp
```diff
@@ -15,25 +15,25 @@ perform a Denial of Service attack.
...
 
 <example>
 <p>
-The following example shows a HTTP request parameter that is used to construct a regular expression:
+The following example shows an HTTP request parameter that is used to construct a regular expression.
 </p>
-<sample src="RegexInjection.java" />
 <p>
 In the first case the user-provided regex is not escaped.
 If a malicious user provides a regex that has exponential worst case performance,
```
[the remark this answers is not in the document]
jcogs33: Thanks for the suggestion! I've updated this in commit https://github.com/github/codeql/pull/11070/commits/9e2ec9d12f599408f20a28d036163316b6eb077a.

atorralba (approved):

PR #11081, opened 2022-11-02
QL4QL: Add DependencyPath.ql query

asgerf:
Adds an explorative query for visualising how a given predicate/class/file depends on a given predicate or class.

By "explorative" I mean it's not intended to run in CI. It's a query you have to adapt to your needs and run locally.

Dependencies introduced by magic are not known to the query.

To use the query:
- Build the QL4QL extractor and extract a database for the relevant CodeQL library
- Modify `ql/src/queries/explore/DependencyPath.ql` and run it

---

For example, here I ran it to answer how `RE::RegExpTerm` depends on `isLocalSource`:

dependency-path-sloww


erik-krogh (dismissed):
Nice 👍
I see you're using your new key-bindings.    

LGTM.

erik-krogh (approved):

PR #11083, opened 2022-11-02
C#: Telemetry query updates.

michaelnebel:
It turns out that the Telemetry we emit for API calls in C# needs to be changed.
- We have removed the (assembly) file name as the same method can be located in different assemblies across version, which clutters the picture of the usage.
- We have aligned the counting to only count *calls* and not the possible (runtime) dispatch targets.
- We now only consider callables that are effectively public.

This should align the implemetation a bit better with Java.

With respect to testing of the PR
- The unit tests behaves as expected (the coverage is not optimal, but still a good indication that everything works as expected).
- I have tried to run some of the queries locally against .NET Runtime and .NET EfCore and the results looks reasonable (ie. the results doesn't contain any "noise", but it would be a huge task to verify that the results are correct manually).
- The DCA results looks acceptable (ie. there are no performance regressions when running the full nightly query suite).

The only thing in this PR I am a bit uncertain about is, whether it is a good idea to tag some of the string generating predicates with the `bindingset` pragma.

### DIFF
--- a/csharp/ql/lib/semmle/code/csharp/Type.qll
+++ b/csharp/ql/lib/semmle/code/csharp/Type.qll
@@ -824,6 +824,8 @@ class RecordClass extends RecordType, Class {
  */
 class AnonymousClass extends Class {
   AnonymousClass() { anonymous_types(this) }
+
+  override string getAPrimaryQlClass() { result = "AnonymousClass" }
 }
 
 /**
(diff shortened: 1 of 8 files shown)


sidshank (approved):
Based on the async chats we've had about this PR, looks plausible to me!

michaelnebel:
> Based on the async chats we've had about this PR, looks plausible to me!

Great - thank you for the review! :-)

PR #11085, opened 2022-11-02
Use latest released bundle for QL-for-QL

dbartol:


erik-krogh (approved):
LGTM 👍  

But I'm wondering why you couldn't use `tools:latest`? 
Is it because we're using a pinned version of `codeql-action`?

PR #11088, opened 2022-11-02
Python: Inline query tests

yoff:
Similar to the consistency queries used in javascript, but based on the inline expectations test framework.
Annotations include relative line numbers, which may or may not be appropriate.

### DIFF
--- a/python/ql/test/experimental/dataflow/TestUtil/DataflowQueryTest.qll
+++ b/python/ql/test/experimental/dataflow/TestUtil/DataflowQueryTest.qll
@@ -1,28 +1,43 @@
 import python
-import experimental.dataflow.TestUtil.FlowTest
+import semmle.python.dataflow.new.DataFlow
+import TestUtilities.InlineExpectationsTest
 private import semmle.python.dataflow.new.internal.PrintNode
 
-class DataFlowQueryTest extends FlowTest {
+class DataFlowQueryTest extends InlineExpectationsTest {
   DataFlowQueryTest() { this = "DataFlowQueryTest" }
 
-  override string flowTag() { result = "flow" }
+  override string getARelevantTag() { result = "result" }
 
-  override predicate relevantFlow(DataFlow::Node source, DataFlow::Node sink) {
-    exists(DataFlow::Configuration cfg | cfg.hasFlow(source, sink))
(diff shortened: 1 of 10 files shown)


RasmusWL (changes requested):
I was hoping to get annotations like they do in JS and Ruby with bad/good, or `ok`/`not ok`. Can we get that instead of the `flow` label which leaves more up for interpretation?

You looked at the way JS does it currently, here is an example of how they do it in Ruby:

- https://github.com/github/codeql/blob/2f9f1f73b75bc4a1168c793c93c3fbc5493cbea8/ruby/ql/test/query-tests/experimental/improper-memoization/ImproperMemoization.ql
- https://github.com/github/codeql/blob/2f9f1f73b75bc4a1168c793c93c3fbc5493cbea8/ruby/ql/test/query-tests/experimental/improper-memoization/improper_memoization.rb#L50

I think it's a clever trick to consider the sink of _any_ configuration, but I'm wondering if it might be a bit too clever; it will not behave properly if we need to use multiple configurations for a single query, and come to think of it, we actually do some more logic in the to distinguish between full/partial SSRF... but I guess as long as we're able to override `relevantFlow` to be query specific, it's going to be ok!

Also, I think we should remove `missingAnnotationOnSink`, when adding safe examples, we certainly don't want to add a `MISSING: result=BAD` annotation :flushed: Actually, I remember we got support for optional annotations, so it might be really nice to be able to add `result=safe` (or similar) annotations in the code (but these can't be required, since for an FN we would then have `MISSING: result=bad SPURIOUS: result=safe")

yoff:
> Also, I think we should remove `missingAnnotationOnSink`, when adding safe examples, we certainly don't want to add a `MISSING: result=BAD` annotation 😳 Actually, I remember we got support for optional annotations, so it might be really nice to be able to add `result=safe` (or similar) annotations in the code (but these can't be required, since for an FN we would then have `MISSING: result=bad SPURIOUS: result=safe")

Good point.

yoff:
Hm, just realised that I picked up `OK` while you suggested `safe`. The latter might indeed be clearer..

RasmusWL (approved):
Thanks for making those updates, I think it looks fine to use `OK` :+1:

PR #11089, opened 2022-11-02
Kotlin: Extract missing arguments of enum constructor calls

tamasvajk:
Additionally a consistency check is also added to compare argument count of calls with the parameter count of the call targets.

### DIFF
--- a/java/kotlin-extractor/src/main/kotlin/KotlinFileExtractor.kt
+++ b/java/kotlin-extractor/src/main/kotlin/KotlinFileExtractor.kt
@@ -1868,13 +1868,12 @@ open class KotlinFileExtractor(
             extractStaticTypeAccessQualifierUnchecked(overriddenCallTarget.parent, id, locId, enclosingCallable, enclosingStmt)
         }
 
-        extractDefaultsCallArguments(id, overriddenCallTarget, locId, enclosingCallable, enclosingStmt, valueArguments, dispatchReceiver, extensionReceiver)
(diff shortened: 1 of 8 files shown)


java/kotlin-extractor/src/main/kotlin/KotlinFileExtractor.kt
```diff
@@ -2116,19 +2117,38 @@ open class KotlinFileExtractor(
...
+                    realCallTarget.parentClassOrNull?.fqNameWhenAvailable?.asString() == "kotlin.Enum" &&
+                    realCallTarget.valueParameters.size == 2 &&
+                    i < realCallTarget.valueParameters.size) {
+
+                    if (i == 0 && realCallTarget.valueParameters[i].type == pluginContext.irBuiltIns.stringType) {
+
+                        val id = extractNull(pluginContext.irBuiltIns.stringType, fallbackLocation, callId, (i++) + idxOffset, enclosingCallable, enclosingStmt)
+                        tw.writeCompiler_generated(id, CompilerGeneratedKinds.ENUM_CONSTRUCTOR_ARGUMENT.kind)
+                    } else if (i == 1 && realCallTarget.valueParameters[i].type == pluginContext.irBuiltIns.intType) {
+
```
smowton, line 2147: Can this only happen from the callsite at line 3080? If so recommend moving this there.

tamasvajk: Thanks, looks cleaner.

smowton (approved):

smowton (approved):

PR #11091, opened 2022-11-02
Fix AV Rule 76

JarLob:
The rule wasn't finding anything. The problem was that
```ql
    c.getAMemberFunction() instanceof CopyConstructor and
    c.getAMemberFunction() instanceof CopyAssignmentOperator
```
was returning compiler generated implicit assignment operators and copy constructors too.

### DIFF
--- a/cpp/ql/src/change-notes/2022-11-03-av-rule-76.md
+++ b/cpp/ql/src/change-notes/2022-11-03-av-rule-76.md
@@ -1,0 +1,4 @@
+---
+category: minorAnalysis
+---
+* Fixed a bug in `cpp/jsf/av-rule-76` that caused the query to miss results when an implicitly-defined copy constructor or copy assignment operator was generated.
--- a/cpp/ql/src/jsf/4.10 Classes/AV Rule 76.ql
+++ b/cpp/ql/src/jsf/4.10 Classes/AV Rule 76.ql
@@ -38,9 +38,9 @@ predicate hasNontrivialDestructor(Class c) {
 from Class c
 where
   (hasPointerMember(c) or hasNontrivialDestructor(c)) and
-  not (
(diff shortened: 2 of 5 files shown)


geoffw0 (approved):
Looks good to me.

I'm a bit mystified how this had happened, but it is a very old query - I can only assume once upon a time we didn't extract implicit copy assignment operators and copy constructors at all.

The addition of the test is much appreciated.  This should not happen again!

MathiasVP:
We should probably run DCA on this before we merge it.

geoffw0:
The query doesn't have an `@precision` so I don't think it will appear in any query suites at present.  Maybe that makes a DCA run excessive, but I've started one selecting just this query.

I actually think we should have a change note for this as users of the query may be affected.

geoffw0:
Ha, looks like we get two DCA runs then!

MathiasVP:
I've started a DCA run for the PR. I'm not sure if you can see the backlink though, @JarLob. We will keep you updated if anything else needs to happen with the PR before we merge it!

MathiasVP:
> Ha, looks like we get two DCA runs then!

Sorry 😂.

> The query doesn't have an @precision so I don't think it will appear in any query suites at present. Maybe that makes a DCA run excessive, but I've started one selecting just this query.

Yeah, we actually need to start a DCA run that specifically includes this query (since it's not part of any suite). And I see that we both created a custom suite that does exactly this 😄.

> I actually think we should have a change note for this as users of the query may be affected.

Good point. I'd also be okay with including a change note for this. Maybe something like:
> Fixed a bug in `cpp/jsf/av-rule-76` that caused the query to miss results when an implicitly-defined copy constructor or copy assignment operator was generated.

?

geoffw0:
DCA looks good to me.

geoffw0 (approved):
:+1:

PR #11094, opened 2022-11-03
Swift: refactor visitors to use translations

redsun82:
This refactors all visitors to use translation style implementations with structured C++ classes (as opposed to visit and explicit traps). This was required as a preparation step for further work on making the extractor reliable against incomplete ASTs.

Also most explicit asserts have been removed, as they became obsolete with the assert in `fetchLabel`.

In the process of this there was
* one fixed bug where the guard of `CaseLabelItem` was not being extracted
* some schema changes from `list[...]` to `list[optional[...]]` that were required with the structured C++ classes

Commit by commit review is strongly encouraged.


AlexDenisov (approved):
The change looks reasonable 😄
Did you start the DCA experiment already?

PR #11096, opened 2022-11-03
Swift: Move the Unit class to its own file

atorralba:
This removes the necessity of importing an internal dataflow file to be able to use the `Unit` class.

### DIFF
--- a/swift/ql/lib/codeql/swift/dataflow/internal/DataFlowImplSpecific.qll
+++ b/swift/ql/lib/codeql/swift/dataflow/internal/DataFlowImplSpecific.qll
@@ -1,9 +1,15 @@
 /**
  * Provides Swift-specific definitions for use in the data flow library.
  */
+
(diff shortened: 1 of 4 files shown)


swift/ql/lib/codeql/swift/dataflow/internal/DataFlowImplSpecific.qll
```diff
@@ -1,9 +1,14 @@
 /**
  * Provides Swift-specific definitions for use in the data flow library.
  */
+
+import swift as Swift
```
MathiasVP, line 5: Could this be private?

atorralba: Yep, just realized it 😅 Fixed in 978ed03, thanks for pointing it out.

MathiasVP (approved):
LGTM!

PR #11099, opened 2022-11-03
Add ATM to labeler

henrymercer:
Label PRs touching files matching `javascript/ql/experimental/adaptivethreatmodeling/**/*` with "ATM".


jhelie (approved):
Thanks @henrymercer, TIL about [labeler](https://github.com/actions/labeler).

PR #11101, opened 2022-11-03
Swift: extract diagnostics

AlexDenisov:
For some reason, the diagnostic in test doesn't have a location even though the trap file looks right:
```
#0=*
diagnostics(#0, "Hello, warning", 2)
#1=*
files(#1, "<truncated>/ql/swift/ql/test/extractor-tests/Diagnostics/diags.swift")
#2=*
locations(#2, #1, 1, 10, 1, 10)
locatable_locations(#0, #2)
```


swift/extractor/invocation/CodeQLDiagnosticsConsumer.cpp
```diff
@@ -0,0 +1,97 @@
...
+
+  LocationsTrap location;
+  location.id = trap.createLabel<LocationTag>();
+  location.file = file.id;
+  std::tie(location.start_line, location.start_column) =
+      sourceManager.getLineAndColumnInBuffer(loc);
+  std::tie(location.end_line, location.end_column) = sourceManager.getLineAndColumnInBuffer(loc);
+  trap.emit(location);
+  trap.emit(LocatableLocationsTrap{locatable.id, location.id});
+}
```
redsun82, line 64: that's copy&pasted, right? but I guess is just to test it out for the moment

swift/extractor/invocation/CodeQLDiagnosticsConsumer.cpp
```diff
@@ -0,0 +1,97 @@
...
+  trap.emit(file);
+
+  LocationsTrap location;
+  location.id = trap.createLabel<LocationTag>();
+  location.file = file.id;
+  std::tie(location.start_line, location.start_column) =
+      sourceManager.getLineAndColumnInBuffer(loc);
+  std::tie(location.end_line, location.end_column) = sourceManager.getLineAndColumnInBuffer(loc);
+  trap.emit(location);
+  trap.emit(LocatableLocationsTrap{locatable.id, location.id});
```
redsun82, line 63: you are emitting non-final classes, that's your issue. `File` is now a super class of `DbFile` (should really be named `RawFile` or something) and synthesized `UnknownFile`. Same goes for `Location`. Also, you are not deduplicating with trap keys. I guess you took this code from a previous version of the extractor, prior to the synthesized `UnknownFile` and `UnknownLocation`.
```suggestion
  DbFile file{trap.createLabel<FileTag>(filepath)};
  file.name = filepath;
  trap.emit(file);
  DbLocation loc{
  file.name = filepath;
  trap.emit(file);

  DbLocation entry{{}};
  entry.file = fetchLabel(file);
  std::tie(entry.start_line, entry.start_column) = sourceManager.getLineAndColumnInBuffer(start);
  std::tie(entry.end_line, entry.end_column) = sourceManager.getLineAndColumnInBuffer(end);
  entry.id = trap.createLabel<DbLocationTag>('{', entry.file, "}:", entry.start_line, ':',
                                             entry.start_column, ':', entry.end_line, ':',
                                             entry.end_column);
  emit(entry);
  emit(LocatableLocationsTrap{locatableLabel, entry.id});
```
We should probably find a way to share this with `SwiftDispatcher`. Maybe a separate `SwiftLocationExtractor` for locations and files, with its own cache?

swift/schema.py
```diff
@@ -40,6 +40,10 @@ class Locatable(Element):
 class Comment(Locatable):
     text: string
 
+class Diagnostics(Locatable):
+    text: string
+    kind: int
```
redsun82, line 45: would it maybe make sense to use separate subclasses instead of `kind`, something like

```python
class Diagnostics(Locatable):
    text: string

class CompilerError(Diagnositcs):
    pass

class CompilerWarning(Diagnositcs):
    pass

class CompilerNotice(Diagnositcs):
    pass

class CompilerRemark(Diagnositcs):
    pass
```
to which we could later add our own
```python
class ExtractorError(Diagnostics):
    pass

class ExtractorWarning(Diagnostics):
    pass
```

On the other hand, this could also be done on the QL side.

swift/extractor/invocation/CodeQLDiagnosticsConsumer.cpp
redsun82: the problem is that I think the database will be flooded with duplicate files (and possibly locations, but that's less problematic). It's not a question of performance, but of database consistency :slightly_smiling_face:

redsun82: in any case, we definitely have to share the code for emitting files and locations, so once that is done proper trap keys will be emitted

swift/extractor/infra/SwiftDispatcher.h
```diff
@@ -11,6 +11,7 @@
 #include "swift/extractor/infra/SwiftTagTraits.h"
 #include "swift/extractor/trap/generated/TrapClasses.h"
 #include "swift/extractor/infra/file/PathHash.h"
```
redsun82, line 13: nit: this include is not needed any more I think

swift/extractor/infra/SwiftLocationExtractor.h
```diff
@@ -0,0 +1,72 @@
...
+#include "swift/extractor/trap/TrapDomain.h"
+#include "swift/extractor/trap/generated/TrapEntries.h"
+#include "swift/extractor/trap/generated/TrapClasses.h"
+#include "swift/extractor/infra/file/PathHash.h"
+
+namespace codeql {
+
+class TrapDomain;
+
+class SwiftLocationExtractor {
```
redsun82, line 15: nit: as this class has no template member function (:tada:), we could split all its member functions definitions into a separate cpp file, which is a bit friendlier for the build

swift/extractor/invocation/CodeQLDiagnosticsConsumer.h
```diff
@@ -0,0 +1,24 @@
+#pragma once
+
+#include <swift/AST/DiagnosticConsumer.h>
+#include "swift/extractor/infra/SwiftLocationExtractor.h"
+
+namespace codeql {
+
+class TrapDomain;
+
+class CodeQLDiagnosticsConsumer : public swift::DiagnosticConsumer {
```
redsun82, line 10: nit: shouldn't this be `SwiftDiagnosticsConsumer` to be consistent with other names?

swift/extractor/invocation/CodeQLDiagnosticsConsumer.cpp
```diff
@@ -0,0 +1,59 @@
...
+                                                      const swift::DiagnosticInfo& diagInfo) {
+  // Translate ranges.
+  llvm::SmallVector<llvm::SMRange, 2> ranges;
+  for (auto R : diagInfo.Ranges)
+    ranges.push_back(getRawRange(sourceManager, R));
+
+  // Translate fix-its.
+  llvm::SmallVector<llvm::SMFixIt, 2> fixIts;
+  for (const swift::DiagnosticInfo::FixIt& F : diagInfo.FixIts)
+    fixIts.push_back(getRawFixIt(sourceManager, F));
```
redsun82, line 49: `ranges` and `fixIts` are not used anywhere, what is this code for? If there is some obscure side effect it should be described with comments

swift/extractor/main.cpp
```diff
@@ -96,6 +104,23 @@ static void checkWhetherToRunUnderTool(int argc, char* const* argv) {
...
+  for (auto& option : configuration.frontendOptions) {
+    hasher.process(option.c_str(), option.c_str() + option.size() + 1);
+  }
+  hasher.finish();
+  auto target = "invocation-"s + get_hash_hex_string(hasher);
+  auto maybeFile = codeql::createTargetTrapFile(configuration, target);
+  if (!maybeFile) {
+    std::cerr << "Cannot create invocation trap file: " << target << "\n";
+    abort();
+  }
```
redsun82, line 120: what if for some reason we are calling the swift frontend with exactly the same arguments? Should we really abort in such a case? Also, now that I think about it hashing the command line flags might miss changes to the files provided to the call...

As there is no need to reference this output trap file outside of this very call of the extractor, maybe it'd be better to just use a random new file name every time?

redsun82 (commented):
Looks good! I mostly commented with nits, but I'd also like to have an explanation on the unused `ranges` and `fixIts` :slightly_smiling_face:

swift/extractor/invocation/CodeQLDiagnosticsConsumer.cpp
AlexDenisov: Good catch, they were used originally, but then I tweaked the diagnostics message and they are not needed anymore.

swift/extractor/main.cpp
AlexDenisov: Yeah, you are right, some random name would work better.

redsun82: as discussed on chat, probably timestamp + getpid is a good fit for the trap target. Also, maybe use `invocations` as a directory containing the traps rather than as a filename prefix?

swift/extractor/main.cpp
```diff
@@ -107,12 +107,9 @@ static void checkWhetherToRunUnderTool(int argc, char* const* argv) {
...
 // compilations, diagnostics, etc.
 codeql::TargetFile invocationTargetFile(codeql::SwiftExtractorConfiguration& configuration) {
-  auto hasher = picosha2::hash256_one_by_one();
-  for (auto& option : configuration.frontendOptions) {
-    hasher.process(option.c_str(), option.c_str() + option.size() + 1);
-  }
-  hasher.finish();
-  auto target = "invocation-"s + get_hash_hex_string(hasher);
+  auto timestamp = std::chrono::system_clock::now().time_since_epoch().count();
+  auto filename = std::to_string(timestamp) + '-' + std::to_string(getpid());
```
redsun82, line 111: does this work for sorting? I suppose it wouldn't if the time counter gets one more digit during the build, but that's pretty rare for the time since epoch I guess?

It's so unfortunate easy human readable conversion of time points to strings only hit C++ with C++20
https://en.cppreference.com/w/cpp/chrono/system_clock/operator_ltlt

An alternative is this somewhat clunky code
```suggestion
  using Clock = std::chrono::system_clock;
  auto timestamp = Clock::to_time_t(Clock::now());
  std::string filaneme = "YYYY-MM-DDThh:mm:ss";
  std::strftime(filaneme.data(), filaneme.size() + 1, "%FT%T", std::localtime(&now));
  filename += '-';
  filename += std::to_string(getpid());
```
Resolution is at seconds, which however should not be a problem with the PID besides it. Getting decimal seconds in is possible but even more painful :old-man-yells-at-c++:

Feel free to use it or ignore it :slightly_smiling_face:

redsun82 (approved):
small suggestion on the invocation trap timestamping that you are free to ignore. Great stuff! :shipit:

swift/extractor/main.cpp
AlexDenisov: I don't it's worth dealing with `strftime` 😆

PR #11104, opened 2022-11-03
Swift: fix bad join order in WebView/JsExportedSource

d10c:
The `getName = getName` join was happening too early,
before the methods themselves have been enumerated.


MathiasVP (approved):

atorralba:
Thanks for catching and fixing this! I should've run DCA for that PR.

PR #11105, opened 2022-11-03
Kotlin: use `$default` functions to implement `@JvmOverloads`

smowton:
This avoids extracting the default value expression in more than one place, which causes inconsistencies for e.g. anonymous classes, which expect to have a single `new` expression associated.

### DIFF
--- a/java/kotlin-extractor/src/main/kotlin/KotlinFileExtractor.kt
+++ b/java/kotlin-extractor/src/main/kotlin/KotlinFileExtractor.kt
@@ -21,6 +21,7 @@ import org.jetbrains.kotlin.ir.declarations.*
 import org.jetbrains.kotlin.ir.declarations.lazy.IrLazyFunction
 import org.jetbrains.kotlin.ir.expressions.*
 import org.jetbrains.kotlin.ir.expressions.impl.IrConstImpl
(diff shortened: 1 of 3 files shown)


tamasvajk (commented):
What were the "inconsistencies for e.g. anonymous classes"?

smowton:
@tamasvajk the `isAnonymClass` relation is supposed to be a bijection, meaning a one-to-one `ClassInstanceExpr <-> AnonymousType` relationship. If we extract the default value in both the `$default` function and a `JvmOverloads` function, then you end up with something like

```
fun f(x: SomeInterface = { a -> a }, y: Int)
```
becoming
```
fun f$default(var x: SomeInterface?, var y: Int?, argsMask: Int, ...) {
  if (!(argsMask & 1)) x = { a -> a }
  ...    
}

fun f(y: Int) {
  f({ a -> a }, y)
}
```

Thus the source-level class representing the anonymous class has been duplicated. We can handle that either by allowing that duplication and extracting two classes, or work around the problem by implementing `fun f(y: Int) { f$default(null, y, 3, ...)`, which is the option I've taken here.

tamasvajk:
Thanks for the explanation, makes sense.

tamasvajk (approved):

PR #11111, opened 2022-11-03
Swift: Content flow through tuples

geoffw0:
Content flow through tuples - see the tests for examples.

~Flow through named tuple accesses is not yet supported because this doesn't appear to be extracted yet(?)~ --- the missing results are actually a result of not modelling `DestructureTupleExpr`.

@MathiasVP I would appreciate your review.  I intend to do dictionaries next and perhaps arrays at some point.


MathiasVP (dismissed):
LGTM! We should run a DCA experiment on this before we merge it, though.

geoffw0:
Yep, also I've spotted a couple of minor corrections I want to make...

swift/ql/test/library-tests/dataflow/dataflow/test.swift
```diff
@@ -284,13 +284,13 @@ func testTuples() {
 
     sink(arg: t1)
     sink(arg: t1.0)
-    sink(arg: t1.1) // $ flow=277
+    sink(arg: t1.1)
```
MathiasVP, line 287: Yay for https://github.com/github/codeql/blob/main/swift/ql/lib/codeql/swift/dataflow/internal/DataFlowPrivate.qll#L527 🎉

geoffw0:
I was meaning to ask about the references to `AssignExpr` in `storeStep`.  I don't particularly want to try and change it in this PR, but shouldn't they refer to any write to an LValue - e.g. to a function call 'out' parameter - not just to the LHS of assignment?

MathiasVP:
> I was meaning to ask about the references to AssignExpr in storeStep. I don't particularly want to try and change it in this PR, but shouldn't they refer to any write to an LValue - e.g. to a function call 'out' parameter - not just to the LHS of assignment?

That case should be handled automatically by the shared dataflow library's reverse read mechanism. The TLDR explanation of the reverse-read mechanism is that, given a piece of code like:
```swift
f(&a.x);
```
then will be a post-update node representing the value of `a.x` after leaving the function. The pre-update node of that node is the `ExprNode` attached to `a.x`, and since there's a read step from `a` to `a.x`, the shared dataflow library will generate a store step from the post-update node of `a.x` to the post-update node of `a`.

geoffw0:
Latest DCA run analysis failed with "A fatal error occurred: The CodeQL database is not compatible with a QL library referenced by the query you are trying to run.".  I was getting something similar on my Mac yesterday and I think it was caused by a bad version of ... something.  Most likely something is still on a bad version here now.

I will try again tomorrow.  Maybe merge latest `main` in first in case that matters.

jketema:
Probably the same issue `@karimhamdanali` had earlier, but now showing up in DCA. You probably need to run without caching.

geoffw0:
> You probably need to run without caching.

Oooh, how do I do that?

karimhamdanali:
The solution for me was to rebuild the swift extractor actually.

jketema:
> The solution for me was to rebuild the swift extractor actually.

Yup that happens in DCA when you turn off caching.

geoffw0:
DCA run succeeded
 - apart from one project, where it "lost communication with the server" (doesn't sound like it could be caused by a QL change)
 - some wobbles in things like database build times (which also can't be caused by a QL change)
 - overall analysis was quicker, which is the main reassurance I was hoping for
 - though `signalapp__Signal-iOS` may have regressed; and there's a correlated stage timing report:
```
signalapp__Signal-iOS
ControlFlowGraphImpl::Cached#9#ControlFlowGraphImplShared::Cached#12#Splitting::Cached#2(swift/unsafe-webview-fetch[6], 12/12)
129.3s -> 173.5s
```
I don't think I'd expect changes to control flow performance resulting from data flow changes, but I'm not used to interpretting the stage timing report yet.  I'd like a second opinion on this.

geoffw0:
Can we approve this?  Or would you rather wait until after the release branch is created next week?

MathiasVP:
> I don't think I'd expect changes to control flow performance resulting from data flow changes, but I'm not used to interpretting the stage timing report yet. I'd like a second opinion on this.

Indeed, there's no reason to expect the control-flow stage to be affected by this, so the stage timing is most likely just noise.

The end-to-end timing isn't known to be this noisy, though. So it's fair to assume that there's some performance penalty attached to this, but that's to be expected.

MathiasVP (approved):
LGTM!

PR #11112, opened 2022-11-03
C++: Improve `Buffer.qll` performance

MathiasVP:
Removes this large TC I saw on openjdk when evaluating `BadlyBoundedWrite.ql`:
```
[2022-11-03 20:23:46] (5s) Starting to evaluate predicate #Buffer#61e3d199::localFlowToExprStep#2Plus#ff/2@7c04651e = HOP fastTC(17471850)
[2022-11-03 20:24:09] (28s)  >>> Relation #Buffer#61e3d199::localFlowToExprStep#2Plus#ff: 312912206 rows using 611 MB
```
Instead we get this nice and fast cached predicate:
```
Tuple counts for DataFlowUtil#47741e1f::localExprFlowStep#2#ff/2@ebe6dafb after 3.4s:
  20455857 ~0%      {2} r1 = SCAN DataFlowUtil#47741e1f::simpleLocalFlowStep#2#ff OUTPUT In.1, In.0
  11981993 ~26%     {2} r2 = JOIN r1 WITH project#DataFlowUtil#47741e1f::localStepsToExpr#3#fff ON FIRST 1 OUTPUT Lhs.1, Rhs.1 'e2'
  2480227  ~45%     {2} r3 = JOIN r2 WITH DataFlowUtil#47741e1f::Node::asExpr#0#dispred#ff ON FIRST 1 OUTPUT Rhs.1 'e1', Lhs.1 'e2'
  return r3
```

### DIFF
--- a/cpp/ql/lib/semmle/code/cpp/commons/Buffer.qll
+++ b/cpp/ql/lib/semmle/code/cpp/commons/Buffer.qll
@@ -74,39 +74,32 @@ private int isSource(Expr bufferExpr, Element why) {
 }
 
 /**
- * Holds if data flow steps from `e1` to `e2` without stepping through any
- * other intermediate expressions.
- */
-private predicate localFlowStepToExpr(Expr e1, Expr e2) {
-  getBufferSizeCand0(e1) and
-  DataFlow::localExprFlowStep(e1, e2)
-}
-
-/**
  * Holds if `e2` is an expression that is derived from `e1` such that if `e1[n]` is a
  * well-defined expression for some number `n`, then `e2[n + delta]` is also a well-defined
  * expression.
  */
 private predicate step(Expr e1, Expr e2, int delta) {
   getBufferSizeCand0(e1) and
-  exists(Variable bufferVar, Class parentClass, VariableAccess parentPtr, int bufferSize |
-    e1 = parentPtr
-  |
-    bufferVar = e2.(VariableAccess).getTarget() and
-    // buffer is the parentPtr->bufferVar of a 'variable size struct'
-    memberMayBeVarSize(parentClass, bufferVar) and
-    parentPtr = e2.(VariableAccess).getQualifier() and
-    parentPtr.getTarget().getUnspecifiedType().(PointerType).getBaseType() = parentClass and
-    (
-      if exists(bufferVar.getType().getSize())
-      then bufferSize = bufferVar.getType().getSize()
-      else bufferSize = 0
-    ) and
-    delta = bufferSize - parentClass.getSize()
+  (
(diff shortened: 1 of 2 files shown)


cpp/ql/lib/semmle/code/cpp/ir/dataflow/internal/DataFlowUtil.qll
```diff
@@ -1320,7 +1320,36 @@ predicate localInstructionFlow(Instruction e1, Instruction e2) {
...
+ * dataflow step.
+ */
+private predicate localStepFromNonExpr(Node n1, Node n2) {
+  not exists(n1.asExpr()) and
+  localFlowStep(n1, n2)
+}
+
+/**
+ * Holds if `n1.asExpr()` doesn't have a result, `n2.asExpr() = e2` and
+ * `n2` is the first node reachable from `n2` such that `n2.asExpr()` exists.
```
MathiasVP, line 1336:
```suggestion
 * `n2` is the first node reachable from `n1` such that `n2.asExpr()` exists.
```

rdmarsh2:
Can we do a DCA run to make sure this doesn't have unexpected impacts elsewhere?

MathiasVP:
DCA looks good. 10-12% speedup on a couple of projects. OpenJDK OOMs on the base branch (which is the reason for this PR in the first place), so we never get to see the impact on OpenJDK, sadly. I expect this to fixed by this PR, though.

cpp/ql/lib/semmle/code/cpp/commons/Buffer.qll
```diff
@@ -73,20 +73,13 @@ private int isSource(Expr bufferExpr, Element why) {
...
-  DataFlow::localFlowStep(n1, n2)
-}
-
-/** Holds if `n2 + delta` may be equal to `n1`. */
+/**
+ * Holds if data flow steps from `e1` to `e2` without stepping through any
+ * other intermediate expressions.
+ */
 private predicate localFlowStepToExpr(Expr e1, Expr e2) {
   getBufferSizeCand0(e1) and
```
jketema, line 81: Nitpicking: since we only use this in one place, do we just want to inline this?

jketema (approved):
A bit of nitpicking, but otherwise looks good to me.

cpp/ql/lib/semmle/code/cpp/commons/Buffer.qll
MathiasVP: Good idea. Fixed in 2617e6d7c6d1c4e38b81ba275f09c398c66ec8d2.

jketema (approved):

PR #11114, opened 2022-11-03
Ruby: Add case string comparison barrier guard

hmac:
This recognises barriers of the form

```rb
STRINGS = ["foo", "bar"]

case foo
when "some string literal"
  foo
when *["other", "strings"]
  foo
when *STRINGS
  foo
end
```

where the reads of `foo` inside each `when` are guarded by the comparison
of `foo` with the string literals.

We don't yet recognise this construct:

```rb
case foo
when "foo", "bar"
  foo
end
```

This is due to a limitation in the shared barrier guard logic.

### DIFF
--- a/ruby/ql/lib/codeql/ruby/dataflow/BarrierGuards.qll
+++ b/ruby/ql/lib/codeql/ruby/dataflow/BarrierGuards.qll
@@ -31,15 +31,15 @@ private predicate stringConstCompare(CfgNodes::AstCfgNode guard, CfgNode testedN
   or
   stringConstCaseCompare(guard, testedNode, branch)
   or
-  exists(Ssa::Definition def, CfgNodes::ExprNodes::BinaryOperationCfgNode g |
+  exists(CfgNodes::ExprNodes::BinaryOperationCfgNode g |
     g = guard and
-    stringConstCompareOr(guard, def, branch) and
+    stringConstCompareOr(guard, branch) and
     stringConstCompare(g.getLeftOperand(), testedNode, _)
   )
 }
 
 /**
- * Holds if `guard` is an `or` expression whose operands are string comparison guards that test `def`.
+ * Holds if `guard` is an `or` expression whose operands are string comparison guards.
  * For example:
  *
  * ```rb
@@ -47,12 +47,12 @@ private predicate stringConstCompare(CfgNodes::AstCfgNode guard, CfgNode testedN
  * ```
  */
 private predicate stringConstCompareOr(
-  CfgNodes::ExprNodes::BinaryOperationCfgNode guard, Ssa::Definition def, boolean branch
+  CfgNodes::ExprNodes::BinaryOperationCfgNode guard, boolean branch
 ) {
   guard.getExpr() instanceof LogicalOrExpr and
   branch = true and
   forall(CfgNode innerGuard | innerGuard = guard.getAnOperand() |
-    stringConstCompare(innerGuard, def.getARead(), branch)
+    stringConstCompare(innerGuard, any(Ssa::Definition def).getARead(), branch)
   )
 }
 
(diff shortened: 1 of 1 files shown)


ruby/ql/lib/codeql/ruby/dataflow/BarrierGuards.qll
```diff
@@ -132,3 +136,57 @@ deprecated class StringConstArrayInclusionCall extends DataFlow::BarrierGuard,
...
+      branchNode = case.getBranch(_) and
+      guard = branchNode.getPattern(_) and
+      // For simplicity, consider patterns that contain only string literals or arrays of string literals
+      forall(ExprCfgNode pattern | pattern = branchNode.getPattern(_) |
+        // when "foo"
+        // when "foo", "bar"
+        pattern instanceof ExprNodes::StringLiteralCfgNode
+        or
+        // array literals behave weirdly in the CFG so we need to drop down to the AST level for this bit
+        // specifically: `SplatExprCfgNode.getOperand()` does not return results for array literals
```
aibaars, line 172: I'm a bit surprised by this comment. It could be that there is a bug in the CFG library, but other than that `getOperand` should return an array literal.

ruby/ql/test/library-tests/dataflow/barrier-guards/barrier-guards.rb
```diff
@@ -102,3 +102,97 @@
...
+end
+
+FOO_AND_X = ["foo", x]
+
+case foo
+when *FOO_AND_X # not a guard - includes non-constant element `x`
+    foo
+end
+
+if foo == "foo" or foo == "bar" # not recognised
```
aibaars, line 173: I think this should be fixed by https://github.com/github/codeql/pull/11153

ruby/ql/lib/codeql/ruby/controlflow/CfgNodes.qll
```diff
@@ -866,6 +866,17 @@ module ExprNodes {
...
 
+  /** A control-flow node that wraps a `SplatExpr` AST expression. */
+  class SplatExprCfgNode extends ExprCfgNode {
+    override string getAPrimaryQlClass() { result = "SplatExprCfgNode" }
+
+    SplatExprCfgNode() { e instanceof SplatExpr }
+
+    final override SplatExpr getExpr() { result = super.getExpr() }
+
+    final ExprCfgNode getOperand() { result.getExpr() = e.(SplatExpr).getOperand() }
```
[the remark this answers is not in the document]
aibaars: There may be multiple CFG nodes for the same AST node due to control flow graph splitting. The `e.hasCfgChild` predicate picks the CfgNode that matches the AstNode child node and is connected to the current CfgNode.

aibaars:
@hmac I trivially fixed all merge conflicts in the test cases through the GitHub UI. You probably need to refresh the expected output, though.

ruby/ql/lib/codeql/ruby/dataflow/BarrierGuards.qll
```diff
@@ -7,22 +7,71 @@ private import codeql.ruby.controlflow.CfgNodes
 private import codeql.ruby.dataflow.SSA
 private import codeql.ruby.ast.internal.Constant
 private import codeql.ruby.InclusionTests
+private import codeql.ruby.ast.internal.Literal
 
-private predicate stringConstCompare(CfgNodes::ExprCfgNode g, CfgNode e, boolean branch) {
+private predicate stringConstCompare(CfgNodes::AstCfgNode guard, CfgNode testedNode, boolean branch) {
```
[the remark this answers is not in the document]
hvitved: As well as `stringConstArrayInclusionCall`.

hvitved:
> Another result of this is a strange failure in one of the dataflow tests (see CI). This seems to imply that dataflow through case expressions is no longer working correctly, but I haven't found any obvious problems apart from that test failure, so I don't know what's going on there.

I have pushed some CFG fixes that resolve this.

ruby/ql/lib/codeql/ruby/dataflow/BarrierGuards.qll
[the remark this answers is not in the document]
hvitved: That predicate is only used in `stringConstCompare`, which is already cached.

ruby/ql/lib/codeql/ruby/dataflow/BarrierGuards.qll
```diff
@@ -132,3 +166,67 @@ deprecated class StringConstArrayInclusionCall extends DataFlow::BarrierGuard,
...
+ * ```
+ */
+private predicate stringConstCaseCompare(
+  CfgNodes::AstCfgNode guard, CfgNode testedNode, boolean branch
+) {
+  branch = true and
+  exists(CfgNodes::ExprNodes::CaseExprCfgNode case |
+    case.getValue() = testedNode and
+    (
+      exists(CfgNodes::ExprNodes::WhenClauseCfgNode branchNode |
```
hvitved, line 193: I prefer `guard = any(CfgNodes::ExprNodes::WhenClauseCfgNode branchNode |`

ruby/ql/lib/codeql/ruby/controlflow/internal/ControlFlowGraphImpl.qll
```diff
@@ -1414,9 +1416,6 @@ module Trees {
       last = this and
       c.isValidFor(this) and
       c.(ConditionalCompletion).getValue() = false
-      or
-      last(this.getBody(), last, c) and
-      c instanceof NormalCompletion
```
hvitved: Sorry; I was not clear: What I meant was that the `c instanceof NormalCompletion` should be removed; we still need `last(this.getBody(), last, c)` (I wonder why the CFG tests don't fail).

hvitved: Ah yes, it works because of the (somewhat non-compositional)
```rb
last(this.getABranch().(WhenClause).getBody(), pred, c)
```
I still think we should add this back, though.

hvitved (approved):

PR #11117, opened 2022-11-04
fix merge-base compilation when running directly on main

erik-krogh:


erik-krogh:
@aibaars you reviewed the previous PR, so I thought it would be natural if you took this one.  

The workflow crashes when running on `main`: https://github.com/github/codeql/actions/runs/3393142619/jobs/5640140768 
I think I've fixed that now.  
It's not that critical, because the workflow still works fine on PRs.

.github/workflows/compile-queries.yml
```diff
@@ -18,12 +18,18 @@ jobs:
...
-        id: merge-base
+        if: ${{ github.event_name == 'pull_request' }}
         env:
           BASE_BRANCH: ${{ github.base_ref }}
         run: |
           MERGE_BASE=$(git merge-base --fork-point origin/$BASE_BRANCH)
           echo "merge-base=$MERGE_BASE" >> $GITHUB_ENV
+      - name: Calculate merge-base - main
+        if: ${{ github.event_name != 'pull_request' }}
+        # using github.sha instead, since we're on main
```
aibaars, line 29: We may also be on an `rc` branch. It shouldn't matter too much, but perhaps update the name and comment.

aibaars (approved):
Looks good to me. One minor comment.

.github/workflows/compile-queries.yml
erik-krogh, line 29:
```suggestion
      - name: Calculate merge-base - branch
        if: ${{ github.event_name != 'pull_request' }}
        # using github.sha instead, since we're directly on a branch, and not in a PR
```

aibaars (approved):

PR #11118, opened 2022-11-04
ATM: Miscellaneous improvements for the check queries workflow

henrymercer:
This PR makes some improvements to the ATM check queries workflow that ensures that we can run the ATM query pack on `main` using the latest release of the CodeQL CLI.


aeisenberg (approved):
Looks good.

PR #11120, opened 2022-11-04
Swift: Adds XMLDocument sinks to the XXE query

atorralba:
Adds [`XMLDocument`](https://developer.apple.com/documentation/foundation/xmldocument) sinks.


swift/ql/test/query-tests/Security/CWE-611/testXMLDocumentXXE.swift
```diff
@@ -0,0 +1,86 @@
...
+        static let nodeLoadExternalEntitiesNever = XMLNode.Options(rawValue: 1 << 1)
+    }
+}
+
+class XMLElement {}
+
+class XMLDocument {
+    init(contentsOf: URL, options: XMLNode.Options = []) {}
+    init(data: Data, options: XMLNode.Options = []) {}
+    init(rootElement: XMLElement?) {}
```
atorralba, line 31: This constructor doesn't receive `XMLNode.Options` to enable external entities, so it can't be a sink.

geoffw0 (commented):
Looks good to me.

We do rely on the fact that all `XMLDocument` constructors have the same pattern (data in the first argument, options in the second argument, if any) rather than modelling each one individually.  At present all such constructors do follow that pattern, I don't suppose that's likely to change, and nothing really terrible would happen if it did.

atorralba:
> We do rely on the fact that all XMLDocument constructors have the same pattern (data in the first argument, options in the second argument, if any) rather than modelling each one individually.

Would you prefer to model them by signature? Or what would be your preferred way of approaching this? (regardless of whether we end up doing it in this specific case, it will be useful to know in the future).

geoffw0:
After a bit of thought I'm happy with the way you've done it.  Its actually quite well constrained as, in addition to matching an `XMLDocument` constructor, the first argument must be tainted (so likely a string or URL) and the second has to accept an `XMLNode.Options` value for this query to do recognize the call.  I can't imagine this not working fine.

geoffw0 (approved):
:+1:

PR #11125, opened 2022-11-04
Go: Use `any()` to stub getCallbackParameter/ReturnType

smowton:
This is used in all other languages not currently using dataflow type pruning.


aschackmull (approved):
I assume `DataFlowType` is singleton? If so, this LGTM.

smowton:
It is not: `class DataFlowType = Type;` -- whereas I note other languages use a singleton that don't do type pruning.

aschackmull (changes requested):
Then this change won't work - you're introducing huge cartesian products by assigning _all_ types to synthesised nodes related to callbacks.

smowton:
ok, I hadn't looked in detail -- assumed because other languages were stubbing using `any()` it must be e.g. a type-pruning filter. I'll take a look at whether we should use a singleton `DataFlowType` if we're not providing type pruning at all.

smowton:
Replaced by https://github.com/github/codeql/pull/11706

PR #11127, opened 2022-11-04
ATM: Fix CodeQL pack workspace references

henrymercer:
This fixes the ATM PR checks, which [broke on main](https://github.com/github/codeql/actions/runs/3392995797/jobs/5639827326) as a result of the JavaScript pack being updated to use `${workspace}` in https://github.com/github/codeql/pull/11004.


adityasharad (approved):
Looks sensible, LGTM if tests pass.

dbartol (approved):
This should fix it, although I think I would not have expected it to fail. I've opened https://github.com/github/codeql-core/issues/2937 for further investigation.

Issue #11107, opened 2022-11-03
Ruby: Noisiness of rb/weak-cryptographic-algorithm / MD5 detection


SampsonCrowley:
**Description of the false positive**
https://github.com/github/codeql/blob/a520de3986987baf4c5f846bd82bf68536ae042c/ruby/ql/src/queries/security/cwe-327/BrokenCryptoAlgorithm.ql

This flags every single use of MD5 as a cryptography problem.

MD5 exists for a reason an it's entirely inappropriate to flag any and every usage of it as a cryptographic usage

It is intended to be a lighter weight, simpler algorithm. Using it *at all* should not be a flag. there are plenty of legitimate use cases that have nothing to do with security

### example:

this sorting algorithm has nothing to do with security and absolutely does not need the heavier implementation of an SHA1 hash

![Screenshot 2022-11-03 at 9 54 13 AM](https://user-images.githubusercontent.com/22461638/199770098-faeee28a-c0ba-48e4-8617-33f02da5a703.png)

turbo:
Hi all 👋,

Thank you for reporting this. We are aware of an increased amount of alerts caused by recent changes to this query, and a currently working on a fix. We realize this may be disruptive to your workflow at this time, so this has a high priority for us.

If you need guidance on dismissing the existing alerts or exclusion logic, please let us know and someone from our team will assist here.

Thanks

aibaars:
This PR should address the issue https://github.com/github/codeql/pull/11119

PR #11129, opened 2022-11-04
Ruby: Improve weak crypto query

aibaars:
Weak hash algorithms such as MD5 and SHA1 are often used in non security sensitive contexts and reporting all uses is far too noisy.

Fixes: https://github.com/github/codeql/issues/11107


hvitved (approved):

theveriton:
sorry to comment on closed PR, but could this be done for Java as well?

PR #11130, opened 2022-11-04
Kotlin: always populate the `files` table

smowton:
Previously individual top-level file declarations relied on their corresponding file-class to declare their `File` instance, but this can be scuppered by a Java extractor replacing that file-class and identifying a different file location.


tamasvajk (approved):
Looks reasonable to me. Would it be super cumbersome to add a test to illustrate why this change is needed?

smowton:
Not super cumbersome, but on this one occasion I'll take a shortcut given high confidence that the change is safe :)

PR #11131, opened 2022-11-04
Swift: deal with incomplete ASTs

redsun82:


AlexDenisov (approved):
All in all, looks great.
The only tiny concern is that the extractor is becoming way too template-heavy as to my taste 😅

PR #11132, opened 2022-11-04
QL: fix non-attached annotations for newtype branches

erik-krogh:
Fixes this FP: https://github.com/github/codeql/security/code-scanning/77304

### DIFF
--- a/ql/ql/src/codeql_ql/ast/Ast.qll
+++ b/ql/ql/src/codeql_ql/ast/Ast.qll
@@ -983,6 +983,8 @@ class NewTypeBranch extends TNewTypeBranch, Predicate, TypeDeclaration {
 
   override NewTypeBranchType getReturnType() { result.getDeclaration() = this }
 
+  override Annotation getAnAnnotation() { toQL(this).getAFieldOrChild() = toQL(result) }
+
   override Type getParameterType(int i) { result = this.getField(i).getType() }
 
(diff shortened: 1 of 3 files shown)


ql/ql/src/codeql_ql/ast/Ast.qll
```diff
@@ -983,6 +983,8 @@ class NewTypeBranch extends TNewTypeBranch, Predicate, TypeDeclaration {
 
   override NewTypeBranchType getReturnType() { result.getDeclaration() = this }
 
+  override Annotation getAnAnnotation() { toQL(this).getAFieldOrChild() = toQL(result) }
```
tausbn, line 986: Should `getAChild` be updated, or do we not count annotations as children of the nodes they are attached to?

tausbn (approved):
One question, otherwise this looks good to me. 👍

ql/ql/src/codeql_ql/ast/Ast.qll
erik-krogh: The base definition of `getAChild` (in `AstNode`) includes `getAnAnnotation()`.   

So `getAnAnnotation()` is part of `getAChild` here in `NewTypeBranch` through the super call.  

I think that answers your question.

tausbn: Makes sense, thank you!

PR #11136, opened 2022-11-06
Ruby: JSON flow summaries

hmac:
Add flow summaries for `ActiveSupport::JSON` and various JSON methods. We consider JSON parsing and generation to be taint-preserving.

### DIFF
--- a/ruby/ql/test/library-tests/dataflow/local/TaintStep.expected
+++ b/ruby/ql/test/library-tests/dataflow/local/TaintStep.expected
@@ -23,7 +23,7 @@
 | file://:0:0:0:0 | parameter position 0 of activesupport;;Member[ActionView].Member[SafeBuffer].Method[new] | file://:0:0:0:0 | [summary] to write: return (return) in activesupport;;Member[ActionView].Member[SafeBuffer].Method[new] |
 | file://:0:0:0:0 | parameter position 0 of activesupport;;Member[ActiveSupport].Member[JSON].Method[decode,load] | file://:0:0:0:0 | [summary] to write: return (return) in activesupport;;Member[ActiveSupport].Member[JSON].Method[decode,load] |
 | file://:0:0:0:0 | parameter position 0 of activesupport;;Member[ActiveSupport].Member[JSON].Method[encode,dump] | file://:0:0:0:0 | [summary] to write: return (return) in activesupport;;Member[ActiveSupport].Member[JSON].Method[encode,dump] |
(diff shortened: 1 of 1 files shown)


ruby/ql/test/library-tests/frameworks/json/JsonDataFlow.ql
```diff
@@ -0,0 +1,12 @@
+/**
+ * @kind path-problem
+ */
+
+import codeql.ruby.AST
+import TestUtilities.InlineFlowTest
+import codeql.ruby.Frameworks
+import PathGraph
+
+from DataFlow::PathNode source, DataFlow::PathNode sink, DefaultValueFlowConf conf
```
erik-krogh, line 10:
```suggestion
from DataFlow::PathNode source, DataFlow::PathNode sink, DefaultTaintFlowConf conf
```

Your inline tests are testing for taint-flow, but in this test you are outputting all the data-flows.  
Is that intentional? 
You could add a negative inline expectation about data-flow?

ruby/ql/lib/codeql/ruby/frameworks/Json.qll
```diff
@@ -0,0 +1,22 @@
...
+   * Not all of these methods are strictly defined in the `json` gem.
+   * The `JSON` namespace is heavily overloaded by other JSON parsing gems such as `oj`, `json_pure`, `multi_json` etc.
+   * This summary covers common methods we've seen called on `JSON` in the wild.
+   */
+  private class JsonSummary extends ModelInput::SummaryModelCsv {
+    override predicate row(string row) {
+      row =
+        [
+          "json;;Member[JSON].Method[parse,parse!,load,restore];Argument[0];ReturnValue;taint",
+          "json;;Member[JSON].Method[generate,fast_generate,dump,unparse,fast_unparse];Argument[0];ReturnValue;taint",
```
alexrford, line 18:
```suggestion
          "json;;Member[JSON].Method[generate,fast_generate,pretty_generate,dump,unparse,fast_unparse];Argument[0];ReturnValue;taint",
```

https://rubyapi.org/3.1/o/json#method-i-pretty_generate

alexrford (commented):
There are a few instance methods for `oj` that look relevant for flow from the `self` arg to the return value, e.g. `as_json`, `to_json` etc: https://rubydoc.info/gems/oj/JSON/GenericObject Are these in scope for this PR - I recognize that there's probably a ton of potential methods to model that might not be used much in practice.

alexrford:
> That's a good question. As you say, there's a lot of potential methods we should model. My preference is to tackle the obvious low-hanging fruit here, and then tackle the various JSON gems in full later on, as part of general non-rails library modelling. What do you think?

Sounds like a good approach.

alexrford (dismissed):
The tests need updating, but otherwise LGTM - I'll reapprove when the tests are fixed.

alexrford (approved):

PR #11137, opened 2022-11-07
C++: Small fixes for the dataflow tests

jketema:
This targets `main`, as this is independent of the use-use dataflow work.


d10c (approved):
LGTM 👍

PR #11138, opened 2022-11-07
Swift: Add AEXML sinks to XXE query

atorralba:
Adds sinks for [AEXML](https://github.com/tadija/AEXML).

The caveat of these sinks is that they heavily rely on local data flow, and because of that we might have some FNs when tracking the unsafe option all the way to the actual sink requires global data flow instead.


swift/ql/lib/codeql/swift/security/XXE.qll
```diff
@@ -90,3 +91,73 @@ private class NodeLoadExternalEntitiesAlways extends VarDecl {
...
+      a.getDest() = sree and
+      sree.(MemberRefExpr).getBase() = parserSettings and
+      parserSettings.(MemberRefExpr).getBase() = this
+    )
+  }
+}
+
+/** An expression of type `AEXMLOptions.ParserSettings`. */
+class ParserSettings extends Expr {
+  pragma[inline]
```
atorralba, line 147: These two `pragma[inlines]` are an attempt at avoiding fully computing these classes, which seemed to be potentially too big depending on the codebase. Hopefully inlining them where they are used should add the appropriate context to prevent it, while allowing us to reuse this code, but please let me know if this solution is not desirable performance-wise and/or there's a better alternative. I'll launch DCA in any case to see how this behaves.

geoffw0: Yep, I agree we should do a DCA run to check how this does.  Though I'm not _overly_ concerned about Swift query performance - I think it will be more effective to do a review later on (when we have a full query suite and thus better data).

swift/ql/test/query-tests/Security/CWE-611/testAEXMLDocumentXXE.swift
```diff
@@ -0,0 +1,148 @@
...
+
+    struct ParserSettings {
+        public var shouldResolveExternalEntities = false
+    }
+}
+
+class AEXMLDocument {
+    init(root: AEXMLElement? = nil, options: AEXMLOptions) {}
+    init(xml: Data, options: AEXMLOptions = AEXMLOptions()) {}
+    init(xml: String, encoding: String.Encoding, options: AEXMLOptions) {}
```
geoffw0, line 36: The `AEXML` library itself has some defaults you don't specify here.  Your test cases below obviously don't use them.  Can you imagine it matters?

geoffw0 (commented):
Looks like a good addition. :+1:

swift/ql/test/query-tests/Security/CWE-611/testAEXMLDocumentXXE.swift
atorralba: The empty `AEXMLOptions` and `String.Encoding` = `utf8` would hardly affect the results. At most, we could verify that we correctly have FNs when the default `options:` argument is used, but the test would look very similar to the ones explicitly setting an empty `AEXMLOptions` (i.e. the `*safeImplicit` test cases). I'm open to adding those additional cases nevertheless, if you think they are worth it.

geoffw0: I'm happy how it is.  But I have wondered when implementing tests myself whether I should put effort into modelling and using default values accurately to how the libraries do it.  I think you're right there isn't much difference between using a default and using a non-default value that's the same as it anyway.  The query certainly doesn't care if a default is there / used.

geoffw0:
From the last two DCA runs:
- analysis _might_ be a bit slower overall (by 3%, 11%; evidence unclear).
  - it isn't even consistent which projects are slower or faster; I'm inclined to dismiss this.
- no stage timings results for the query in question (`swift/xxe`) or anything with `xxe` or `aexml` in it..
- overall I think we're just looking at noise; DCA for Swift isn't very reliable or tuned yet.

geoffw0 (approved):

atorralba:
Due to this:

> no stage timings results for the query in question (swift/xxe) or anything with xxe or aexml in it..

I agree with this:

> overall I think we're just looking at noise; DCA for Swift isn't very reliable or tuned yet.

I ran it the second time to discard potential noise, but even though there are slowdowns again, they indeed don't seem related to these additions.

Thanks for the review!

PR #11145, opened 2022-11-07
Ruby: add a couple of missing links to a new article

felicitymay:
When looking at the preview of the new Ruby articles, I realized that we were missing a couple of links in general articles.

This PR adds those links.

@aibaars - this is targeted at the 2.11.3 branch as I don't think this is urgent and the docs updates can wait until the next release of the CLI.


aibaars:
Well spotted @felicitymay . I agree that these updates can wait a few days and be published alongside 2.11.3.

aibaars (approved):

PR #11149, opened 2022-11-07
C++: Clearer messages for the format args queries

geoffw0:
Makes the messages from the `cpp/too-many-format-arguments` and `cpp/wrong-number-format-arguments` queries a bit clearer in cases where macros are hiding the formatting function call in question.  Motivated by a real case that took a bit of investigation before anyone realized it was a TP.

No change note required because results should be the same apart from clarification in the messages.


geoffw0:
I've created an internal PR to update the internal test.

PR #11153, opened 2022-11-08
Ruby: Split basic blocks around constant conditionals

hvitved:
In cases such as
```rb
if x or y
    foo
else
    bar
```

we have a CFG that looks like

```mermaid
graph TD;

xvar(x);
yvar(y);
foo(foo);
bar(bar);
orfalse("or [false]");
ortrue("or [true]");

xvar-- false -->yvar;
xvar-- true -->ortrue;
yvar-- false -->orfalse;
yvar-- true -->ortrue;

ortrue-- true -->foo;
orfalse-- false -->bar;
```

In order to record the fact that `x or y` `true`-controls `foo` (resp. `false`-controls `bar`), we need to make sure that a new basic block is started at both `foo` and `bar`.

### DIFF
--- a/ruby/ql/test/library-tests/dataflow/barrier-guards/barrier-guards.rb
+++ b/ruby/ql/test/library-tests/dataflow/barrier-guards/barrier-guards.rb
@@ -102,3 +102,9 @@ if bars.include?(foo)
 else
     foo
 end
+
+if x or y then
+    foo
+else
+    foo
+end
--- a/ruby/ql/lib/codeql/ruby/controlflow/BasicBlocks.qll
+++ b/ruby/ql/lib/codeql/ruby/controlflow/BasicBlocks.qll
@@ -252,6 +252,30 @@ private module Cached {
     cfn.isJoin()
     or
     cfn.getAPredecessor().isBranch()
+    or
+    /*
+     * In cases such as
+     *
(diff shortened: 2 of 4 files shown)


ruby/ql/test/library-tests/dataflow/barrier-guards/barrier-guards.rb
```diff
@@ -102,3 +102,9 @@
 else
     foo
 end
+
+if x or y then
+    foo
+else
+    foo
```
aibaars, line 109: Let's use a different name in the else clause to make it easier to validate the expected output.

```suggestion
    bar
```

ruby/ql/test/library-tests/dataflow/barrier-guards/barrier-guards.rb
```diff
@@ -102,3 +102,9 @@
 else
     foo
 end
+
+if x or y then
```
aibaars, line 106: Also add a test case for `and` and perhaps `not` as well.

aibaars (commented):
Looks good to me.

hvitved:
@aibaars : Comments addressed, and DCA looks fine.

aibaars (approved):
LGTM

PR #11157, opened 2022-11-08
JS: fix issue with zero-column yaml locations

asgerf:
The alert renderer does not allow non-empty locations to end at the start of a line, but SnakeYaml emits such locations for multiline strings.

When the YAML extractor is about to emit a non-empty location that ends at column zero, it will instead move the end position backwards until it sees a non-line break character.


erik-krogh (approved):
Did you do a DCA run?  
(Remember to build extractor from `github/codeql`).

asgerf:
[Evaluation](https://github.com/github/codeql-dca-main/issues/8457) on nightly looks fine.

erik-krogh (approved):

PR #11159, opened 2022-11-08
C++: Also taint the return value dereference in the `strcat` model

jketema:


MathiasVP (approved):
The code LGTM! As this could create some new alerts it would be nice to run this on DCA before we merge it though.

jketema:
Do we need a change note, you think?

MathiasVP:
> Do we need a change note, you think?

Nah, I don't think we've done this in the past for simple model fixes, at least.

jketema:
DCA looks good. No changes in alert counts.

PR #11161, opened 2022-11-08
Swift: Add LocalFlowSource class and a few sources.

geoffw0:
Adds a `LocalFlowSource` class to complement `RemoteFlowSource` (both under a `FlowSource` parent class, as it is in CPP).

Adds a handful of local (and remote) flow sources for `String`s, demonstrating that this is all working correctly.  At some point we will want local flow sources for user input as well, but this requires a bit more research and is probably not a high priority.


MathiasVP (approved):
LGTM!

PR #11168, opened 2022-11-08
Swift: extract `PostfixUnaryExpr`

redsun82:


redsun82:
> We have an internal test (with a whitelist list) that checks that all source files are ASCII only. I expect the whitelist will need to be updated.

hmm, it's not really important, I can just not use an emoji here :slightly_smiling_face: thanks for the info!

jketema (approved):
LGTM. Do I understand correctly that all other infrastructure for this was already present, i.e., in the dbscheme and in the QL library?

PR #11170, opened 2022-11-08
Ruby: handle knownOrUnkown in default taint step

asgerf:
Fixes a bug in the default taint steps, which caused us to miss some taint steps out of collections.

[Evaluation](https://github.com/github/codeql-dca-main/issues/8429) shows 538 new tainted nodes and 5 new alerts. 1 of the new alerts seem plausible and the other 4 are due to a missed sanitizer which will most likely be fixed by https://github.com/github/codeql/pull/11114.

### DIFF
--- a/ruby/ql/lib/codeql/ruby/dataflow/internal/TaintTrackingPrivate.qll
+++ b/ruby/ql/lib/codeql/ruby/dataflow/internal/TaintTrackingPrivate.qll
@@ -105,7 +105,7 @@ private module Cached {
     exists(DataFlow::ContentSet c | readStep(nodeFrom, c, nodeTo) |
       c.isSingleton(any(DataFlow::Content::ElementContent ec))
       or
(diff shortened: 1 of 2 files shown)


ruby/ql/lib/codeql/ruby/dataflow/internal/TaintTrackingPrivate.qll
```diff
@@ -105,6 +105,8 @@ private module Cached {
     exists(DataFlow::ContentSet c | readStep(nodeFrom, c, nodeTo) |
       c.isSingleton(any(DataFlow::Content::ElementContent ec))
       or
+      c.isKnownOrUnknownElement(any(DataFlow::Content::ElementContent ec))
```
hvitved, line 108:
```suggestion
      c.isKnownOrUnknownElement(_)
```

hvitved (approved):

PR #11175, opened 2022-11-08
JS: extract .erb files as html

erik-krogh:
CVE-2022-3704: TP/TN 

In that CVE the security issue is entirely contained within some JS code, that sits inside a `<script>` tag in an `.erb` file.  

I figured that our HTML parser already handles HTML with various templating languages, so handling `.erb` shouldn't be that different.  
And it wasn't, it just worked.  

---- 

I tried to add an `.erb` file to the xss-through-dom test, but that wouldn't work, and I don't know why.  
That's why I put an `.erb` test in the trap-tests. 

---- 

Evaluations look fine. 
[4 new extracted files in our default.yml suite](https://github.com/github/codeql-dca-main/tree/data/erik-krogh/erb__default__code-scanning__2/reports).  
[Evaluation on the nightly sources used by Ruby shows a bunch of new extracted files, and OK results](https://github.com/github/codeql-dca-main/tree/data/erik-krogh/erb__ruby__code-scanning__1/reports).


javascript/extractor/src/com/semmle/js/extractor/FileExtractor.java
```diff
@@ -104,7 +104,7 @@ private static int skipBOM(byte[] bytes, int length) {
 
   /** Information about supported file types. */
   public static enum FileType {
-    HTML(".htm", ".html", ".xhtm", ".xhtml", ".vue", ".hbs", ".ejs", ".njk") {
+    HTML(".htm", ".html", ".xhtm", ".xhtml", ".vue", ".hbs", ".ejs", ".njk", ".erb") {
```
aibaars, line 107: Perhaps restrict to `.html.erb`. While many `erb` files are html-like templates, they are also used for many other text formats.

```suggestion
    HTML(".htm", ".html", ".xhtm", ".xhtml", ".vue", ".hbs", ".ejs", ".njk", ".html.erb") {
```

aibaars (commented):
LGTM, one suggestion.

aibaars (approved):
Looks good to me. I assume the HTML extractor behaves nicely when parsing invalid HTML code, right? `.html.erb` files tend to be mostly HTML but also have `<% %>`-like tags.

erik-krogh:
> Looks good to me. I assume the HTML extractor behaves nicely when parsing invalid HTML code, right? `.html.erb` files tend to be mostly HTML but also have `<% %>`-like tags.

Yes. The parser already handles that various templating systems does various weird things. 
(It does so by accepting lots and lots of parse-errors).

aibaars (approved):

PR #11177, opened 2022-11-09
Swift: extract `PoundDiagnosticDecl` and `MissingMemberDecl`

redsun82:
The first are directives `#error` and `#warning`. The second is a placeholder that seem to be possible to arise from deserialization.


swift/extractor/translators/DeclTranslator.cpp
```diff
@@ -401,4 +401,23 @@ std::optional<codeql::OpaqueTypeDecl> DeclTranslator::translateOpaqueTypeDecl(
   return std::nullopt;
 }
 
+static int translateDiagnosticsKind(swift::DiagnosticKind kind) {
```
AlexDenisov, line 404: I guess we should find a way to combine this with the other instance https://github.com/github/codeql/pull/11101/files#diff-05ba0db58312ad56f391863b26935ca619b3808897ed517a4cc68b2ea97be111R13

AlexDenisov (approved):
Would it make sense to actually combine the decl with an actual diagnostic instead of having a somewhat duplicate entity in the database?

swift/extractor/translators/DeclTranslator.cpp
redsun82: yes, I was considering it, but we can do it with a later PR if you prefer :slightly_smiling_face:

redsun82: ah, no wait, your PR is merged, so I can do it :slightly_smiling_face:

AlexDenisov (approved):

PR #11180, opened 2022-11-09
CI: compile-queries: use cache when running on main, and support more base-branches

erik-krogh:
Followup to https://github.com/github/codeql/pull/11162

### DIFF
--- a/.github/workflows/compile-queries.yml
+++ b/.github/workflows/compile-queries.yml
@@ -31,8 +31,8 @@ jobs:
           path: '*/ql/src/.cache'
           key: codeql-compile-pr-${{ github.sha }} # deliberately not using the `compile-compile-main` keys here.
           restore-keys: |
-            codeql-compile-${{ github.base_ref }}-${{ env.merge-base }}
-            codeql-compile-${{ github.base_ref }}-
+            codeql-compile-refs/heads/${{ github.base_ref }}-${{ env.merge-base }}
+            codeql-compile-refs/heads/${{ github.base_ref }}-
       - name: Fill CodeQL query compilation cache - main
         if: ${{ github.event_name != 'pull_request' }}
         uses: actions/cache@v3
--- a/.github/workflows/compile-queries.yml
+++ b/.github/workflows/compile-queries.yml
@@ -38,9 +38,9 @@ jobs:
         uses: actions/cache@v3
         with:
           path: '*/ql/src/.cache'
-          key: codeql-compile-main-${{ github.sha }} # just fill on main
+          key: codeql-compile-${{ github.ref }}-${{ github.sha }} # just fill on main
           restore-keys: | # restore from another random commit, to speed up compilation.
-            codeql-compile-main- 
+            codeql-compile-${{ github.ref }}- 
       - name: Setup CodeQL
         uses: ./.github/actions/fetch-codeql
         with:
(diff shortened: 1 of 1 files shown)


.github/workflows/compile-queries.yml
```diff
@@ -38,7 +38,9 @@ jobs:
         uses: actions/cache@v3
         with:
           path: '*/ql/src/.cache'
-          key: codeql-compile-main-${{ github.sha }} # just fill on main
+          key: codeql-compile-${{ github.ref }}-${{ github.sha }} # just fill on main
+          restore-keys: | # restore from another random commit, to speed up compilation.
+            codeql-compile-${{ github.ref }}-
```
aibaars, line 43: Would it make sense to add the `main` branch as a fall back?
```suggestion
            codeql-compile-${{ github.ref }}- 
            codeql-compile-refs/heads/main- 
```

erik-krogh:
The `use base_ref to calculate the cache-keys when running in a PR.` commit was a little late.  
I had forgotten to save the file, so my commit didn't include those changes.

.github/workflows/compile-queries.yml
```diff
@@ -31,8 +31,8 @@ jobs:
           path: '*/ql/src/.cache'
           key: codeql-compile-pr-${{ github.sha }} # deliberately not using the `compile-compile-main` keys here.
           restore-keys: |
-            codeql-compile-${{ github.base_ref }}-${{ env.merge-base }}
-            codeql-compile-${{ github.base_ref }}-
+            codeql-compile-refs/heads/${{ github.base_ref }}-${{ env.merge-base }}
```
aibaars, line 34: `github.base_ref` should already include the `refs/heads` prefix. Doesn't, it?

.github/workflows/compile-queries.yml
```diff
@@ -31,14 +31,17 @@ jobs:
           path: '*/ql/src/.cache'
           key: codeql-compile-pr-${{ github.sha }} # deliberately not using the `compile-compile-main` keys here.
           restore-keys: |
-            codeql-compile-main-${{ env.merge-base }}
-            codeql-compile-main-
+            codeql-compile-refs/heads/${{ github.base_ref }}-${{ env.merge-base }}
+            codeql-compile-refs/heads/${{ github.base_ref }}-
```
aibaars, line 35: `github.base_ref` should already include the refs/heads prefix. Doesn't, it?

```suggestion
            codeql-compile-${{ github.base_ref }}-${{ env.merge-base }}
            codeql-compile-${{ github.base_ref }}-
```

erik-krogh: No, it does not. I was also surprised by that.   

If you open the [latest compile-queries run](https://github.com/github/codeql/actions/runs/3427945326/jobs/5711566587) and expand `Read CodeQL query compilation - PR`, there you can see the cache keys that were computed.

aibaars: Ah that is surprising, so `ref` is fully qualified but `base_ref` and `head_ref` are not.

.github/workflows/compile-queries.yml
```diff
@@ -31,14 +29,17 @@ jobs:
...
```
aibaars, line 33: Perhaps fall back to `main` too, just in case. It won't matter much in practice as most PR will target `main` anyway.

```suggestion
            codeql-compile-refs/heads/${{ github.base_ref }}-
            codeql-compile-refs/heads/main-            
```

aibaars (approved):
Looks good to me!

PR #11182, opened 2022-11-09
Java/C#: Split active configurations for model generator

aschackmull:
Let's undo the configuration footgun firing. We shouldn't have all of the model generation configurations active at the same time.
Also includes a minor strengthening of the Java to-sink source definition (pushed in some magic).

cc: @tamasvajk @michaelnebel

### DIFF
--- a/java/ql/src/utils/model-generator/internal/CaptureModelsSpecific.qll
+++ b/java/ql/src/utils/model-generator/internal/CaptureModelsSpecific.qll
@@ -18,6 +18,8 @@ module TaintTracking = Tt::TaintTracking;
 
 class Type = J::Type;
 
+class Unit = J::Unit;
+
 private J::Method superImpl(J::Method m) {
   result = m.getAnOverride() and
   not exists(result.getAnOverride()) and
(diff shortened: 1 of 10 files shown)


tamasvajk (approved):

java/ql/src/utils/model-generator/internal/CaptureModelsSpecific.qll
```diff
@@ -223,24 +225,21 @@ predicate isOwnInstanceAccessNode(ReturnNode node) {
...
+    source instanceof DataFlow::ParameterNode
+  ) and
+  source.getEnclosingCallable().isPublic() and
+  exists(J::RefType t |
+    t = source.getEnclosingCallable().getDeclaringType().getAnAncestor() and
+    not t instanceof J::TypeObject and
+    t.isPublic()
+  ) and
+  isRelevantForModels(source.getEnclosingCallable()) and
+  exists(asPartialModel(source.getEnclosingCallable()))
```
michaelnebel, line 242: What is the impact of already narrowing the potential sources nodes by our ability to print a partial model for its enclosing callable (I know this is needed for the model - but which potential source nodes does this exclude?)?

michaelnebel:
> Let's undo the configuration footgun firing. We shouldn't have all of the model generation configurations active at the same time. 

Why is that the case (are they all evaluated - and if so - why?)? As far as I can see the predicates relevant to each of the queries only depend on a single of the configurations. 
In any case, it is a need trick with the `Active` override :-)

michaelnebel (commented):
Looks good to me!

java/ql/src/utils/model-generator/internal/CaptureModelsSpecific.qll
aschackmull: Not sure, I didn't check that thoroughly - I only noticed that it made a (somewhat significant) difference.

michaelnebel:
> > As far as I can see the predicates relevant to each of the queries only depend on a single of the configurations.
> 
> There's no such thing as calculating data flow for just a single configuration - data flow is always calculated for all configurations in scope. That's why we have multiple library copies and why we try to keep the configurations as close to the queries as possible, so two unrelated configuration don't end up in the same scope.

Great! Thank you for the explanation!

PR #11183, opened 2022-11-09
Dataflow: Introduce support for src/sink grouping in path results.

aschackmull:
This introduces a new feature in the shared data flow library. The primary use-case is currently intended to be debugging, but the feature may also find other uses.

Sometimes we get a huge number of path results in a query, which can be cumbersome to look through. If, in these case, we mostly care about flow reaching _some_ sink (possibly with some property) rather than needing to know about the specific sink, and we just want to see representative paths reaching arbitrary sinks, then it can be beneficial to reduce the number of source-sink pairs by adding one or more synthesised sink nodes that collect the flow from multiple sinks. E.g. for MaD sink model generation we care only about the sink kind that's reached rather than the specific sink node.
This PR introduces support for grouping several sink nodes into a single sink-group-node (represented by a string) in order to reduce the number of path results. The symmetric support for source grouping is also added.

### DIFF
--- a/java/ql/lib/semmle/code/java/dataflow/internal/DataFlowImpl.qll
+++ b/java/ql/lib/semmle/code/java/dataflow/internal/DataFlowImpl.qll
@@ -147,6 +147,12 @@ abstract class Configuration extends string {
    */
   FlowFeature getAFeature() { none() }
 
+  /** Holds if sources should be grouped in the result of `hasFlowPath`. */
+  predicate sourceGrouping(Node source, string sourceGroup) { none() }
+
+  /** Holds if sinks should be grouped in the result of `hasFlowPath`. */
+  predicate sinkGrouping(Node sink, string sinkGroup) { none() }
+
   /**
    * Holds if data may flow from `source` to `sink` for this configuration.
    */
@@ -158,7 +164,7 @@ abstract class Configuration extends string {
    * The corresponding paths are generated from the end-points and the graph
    * included in the module `PathGraph`.
    */
-  predicate hasFlowPath(PathNode source, PathNode sink) { flowsTo(source, sink, _, _, this) }
+  predicate hasFlowPath(PathNode source, PathNode sink) { hasFlowPath(source, sink, this) }
 
   /**
    * Holds if data may flow from some source to `sink` for this configuration.
@@ -2712,6 +2718,18 @@ private newtype TPathNode =
       state = sink.getState() and
       config = sink.getConfiguration()
     )
+  } or
+  TPathNodeSourceGroup(string sourceGroup, Configuration config) {
+    exists(PathNodeImpl source |
+      sourceGroup = source.getSourceGroup() and
+      config = source.getConfiguration()
+    )
+  } or
+  TPathNodeSinkGroup(string sinkGroup, Configuration config) {
+    exists(PathNodeSink sink |
+      sinkGroup = sink.getSinkGroup() and
+      config = sink.getConfiguration()
+    )
   }
 
 /**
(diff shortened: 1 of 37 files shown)


aschackmull:
> Looks like PrintAst nondeterminism. Feel free to ignore it.

The Swift CI failure can be ignored.

java/ql/lib/semmle/code/java/dataflow/internal/DataFlowImpl.qll
```diff
@@ -147,6 +147,12 @@ abstract class Configuration extends string {
    */
   FlowFeature getAFeature() { none() }
 
+  /** Holds if sources should be grouped in the result of `hasFlowPath`. */
+  predicate sourceGrouping(Node source, string sourceGroup) { none() }
+
+  /** Holds if sinks should be grouped in the result of `hasFlowPath`. */
+  predicate sinkGrouping(Node sink, string sinkGroup) { none() }
+
```
MathiasVP, line 155: So just to make sure: If `conf.sourceGrouping(mySource1, "foo")` and `conf.sourceGrouping(mySource2, "foo")` holds, then any `PathNode` whose `getNode()` is `mySource1` or `mySource2` will be assigned the same `PathNodeSourceGroup`. Is that correct?

Is a given node allowed to be grouped into multiple `sourceGroups` (and similarly for sinkGroup)?

java/ql/lib/semmle/code/java/dataflow/internal/DataFlowImpl.qll
```diff
@@ -3136,9 +3178,66 @@ private class PathNodeSink extends PathNodeImpl, TPathNodeSink {
...
   override Configuration getConfiguration() { result = config }
 
-  override PathNodeImpl getASuccessorImpl() { none() }
+  override PathNodeImpl getASuccessorImpl() {
+    result = TPathNodeSinkGroup(this.getSinkGroup(), config)
+  }
 
   override predicate isSource() { sourceNode(node, state, config) }
+
+  string getSinkGroup() { config.sinkGrouping(node.asNode(), result) }
```
MathiasVP, line 3187: Should this be `getASinkGroup` since there might be 0 or 1 (or maybe even multiple?) results?

java/ql/lib/semmle/code/java/dataflow/internal/DataFlowImpl.qll
```diff
@@ -2920,6 +2938,22 @@ abstract private class PathNodeImpl extends TPathNode {
     )
   }
 
+  string getSourceGroup() {
```
MathiasVP, line 2941: Similarly: Should this be `getASourceGroup`?

MathiasVP (commented):
A couple of clarifying questions, but otherwise this LGTM!

java/ql/lib/semmle/code/java/dataflow/internal/DataFlowImpl.qll
aschackmull: Yes. Indeed there is a certain degree of flexibility here. The expectation is that you'll assign each sink to a single sink group, but assigning any combination of 0, 1, or 2+ groups to a sink is also going to work, but I'm not sure that the 2+ case is going to be useful. Think of this as augmenting the flow graph with some additional sink nodes (the groups) and adding some edges from the original sinks to those new nodes, and then have the flow be reported targeting those new sink nodes, if possible. Grouping several sinks will then allow you to reduce the total number of alerts. If you add multiple groups for a given sink, then you're effectively inflating the number of results, which I assume will be counterproductive.

java/ql/lib/semmle/code/java/dataflow/internal/DataFlowImpl.qll
aschackmull: I don't think so. Expecting 0 or 1 result is, I think, a reasonable expectation, so the name shouldn't indicate that 2+ is expected.

aschackmull:
> This might also warrant a change note.

I was wondering about that. So far I think this is mostly going to be useful for internal debugging, which is why I considered dark-shipping.

java/ql/lib/semmle/code/java/dataflow/internal/DataFlowImpl.qll
MathiasVP: Resolved [here](https://github.com/github/codeql/pull/11183#discussion_r1023853746).

java/ql/lib/semmle/code/java/dataflow/internal/DataFlowImpl.qll
MathiasVP: Thanks for the intuition!

MathiasVP (approved):
LGTM!

PR #11184, opened 2022-11-09
C++: Re-introduce the `ast` annotations in the taint tests and related test infrastructure

jketema:
This makes it easier to fix the library tests for me, and we should probably keep these around anyway until we remove the AST-based dataflow library.


MathiasVP (approved):
LGTM!

PR #11185, opened 2022-11-09
Swift: add `String` taint steps

karimhamdanali:
If a `String` is tainted, then all its fields should be tainted including computed properties such as `utf8`. This support follows a similar solution to what we do with [URLs](https://github.com/github/codeql/pull/10774) using `TaintInheritingContent`.

### DIFF
--- a/swift/ql/lib/codeql/swift/frameworks/StandardLibrary/String.qll
+++ b/swift/ql/lib/codeql/swift/frameworks/StandardLibrary/String.qll
@@ -1,5 +1,7 @@
 import swift
+private import codeql.swift.dataflow.DataFlow
 private import codeql.swift.dataflow.ExternalFlow
+private import codeql.swift.dataflow.FlowSteps
 
 private class StringSource extends SourceModelCsv {
   override predicate row(string row) {
@@ -14,5 +16,17 @@ private class StringSource extends SourceModelCsv {
         ";String;true;init(contentsOfFile:encoding:);(String,String.Encoding);;ReturnValue;local",
         ";String;true;init(contentsOfFile:usedEncoding:);(String,String.Encoding);;ReturnValue;local"
       ]
+  }
+}
+
+/**
+ * A content implying that, if a `String` is tainted, then all its fields are tainted.
+ */
+private class StringFieldsInheritTaint extends TaintInheritingContent,
+  DataFlow::Content::FieldContent {
+  StringFieldsInheritTaint() {
+    this.getField().getEnclosingDecl().(ClassOrStructDecl).getFullName() = "String" or
+    this.getField().getEnclosingDecl().(ExtensionDecl).getExtendedTypeDecl().getFullName() =
+      "String"
   }
 }
--- a/swift/ql/test/library-tests/dataflow/taint/LocalTaint.expected
+++ b/swift/ql/test/library-tests/dataflow/taint/LocalTaint.expected
@@ -600,6 +600,8 @@
 | nsmutabledata.swift:49:15:49:15 | nsMutableDataTainted6 | nsmutabledata.swift:49:15:49:37 | .mutableBytes |
 | string.swift:5:7:5:7 | SSA def(x) | string.swift:7:16:7:16 | x |
 | string.swift:5:11:5:18 | call to source() | string.swift:5:7:5:7 | SSA def(x) |
+| data.swift:20:23:20:23 | 123456 | data.swift:20:23:20:32 | .utf8 |
+| data.swift:21:25:21:32 | call to source() | data.swift:21:25:21:34 | .utf8 |
 | string.swift:7:13:7:13 |  | string.swift:7:13:7:13 | [post]  |
 | string.swift:7:13:7:13 |  | string.swift:7:14:7:14 | [post] &... |
 | string.swift:7:13:7:13 | SSA def($interpolation) | string.swift:7:14:7:14 | SSA phi($interpolation) |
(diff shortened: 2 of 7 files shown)


swift/ql/lib/codeql/swift/frameworks/StandardLibrary/String.qll
```diff
@@ -10,3 +12,15 @@ private class StringSource extends SourceModelCsv {
...
+ * A content implying that, if a `String` is tainted, then all its fields are tainted.
+ */
+private class StringFieldsInheritTaint extends TaintInheritingContent,
+  DataFlow::Content::FieldContent {
+  StringFieldsInheritTaint() {
+    this.getField().getEnclosingDecl().(ClassOrStructDecl).getFullName() = "String" or
+    this.getField().getEnclosingDecl().(ExtensionDecl).getExtendedTypeDecl().getFullName() =
+      "String"
+  }
+}
```
geoffw0, line 26: This appears to be a sensible approach, but I will note a few rough edges in the case of `String`: `String.count`, `String.isEmpty`, `String.first` and `String.last`.  These things can't contain a lot of data (especially `String.isEmpty` which is a `Bool`) so it would be very difficult for an attacker to utilize a theoretical taint path through any of them.  In CPP in particular we've found taint flow through `strlen` (equivalent to `String.count`) usually produces many more false positive results than useful ones.

I'm not sure what the right solution should be.  We could try to exclude the ones we don't want from this `TaintInheritingContent`, or switch to an explicit modelling approach.  We could potentially block all taint through 'small' types everywhere, solving the problem for good (but would this limit some uses of taint?).  Or we could stick to our principles that these *are* taint flows and expect queries to sanitize on an ad-hoc basis if they tend to find false positive results?

d10c: On the one hand, taint flow through small types irrelevant for most "code injection" type queries. On the other hand, there was Heartbleed.

geoffw0: Good point, there have been plenty of real bugs related to tainted integers (malicious image width/height bugs spring to mind as a second example) - and as you suggest our recent focus on code injection queries may be biasing my viewpoint.

I remain somewhat nervous of taint flow through `strlen` / `.count`.  In CPP we've seen a lot of code like this:
```
char *buffer = (char *)malloc(sizeof(char) * strlen(inputString));
```
which is usually a reasonable thing to do, even if `inputString` is tainted.  There were other similar patterns, and in the end we blocked taint through `strlen`.  This particular example ought to be uncommon in Swift, because you don't need to explicitly allocate like this when you're using a proper `String` class.  I suppose we could take a wait-and-see approach on whether flow through `.count` is actually a problem (or a positive) on Swift?

geoffw0 (commented):
Happy with this.  We should keep an eye on how flow through `String.count` works out in practice.  And I've started a DCA run to check any impact on performance.

swift/ql/lib/codeql/swift/frameworks/StandardLibrary/String.qll
MathiasVP: It may also be worth asking other language teams about this. Experience from C/C++ might not apply very well to this judgment call, but experience from Java/C#/JavaScript/Python and related languages might be a better indication of which direction which should go for Swift.

atorralba: In Java, we _usually_ (we aren't 100% consistent) don't consider things like `length` to be propagating taint, because there are few opportunities in a memory-safe language where a tainted integer could be a problem. Although IMHO, I think we _should_ have those models as well, because some queries actually look for tainted integers. Other queries not interested in that (like code injection) can always add a type-based sanitizer that filters out uninteresting types. 

In Swift, where memory management _can_ be a thing, I think modeling these kind of methods makes even more sense. Maybe `String.count` should be a specific sanitizer for overflow-like vulnerabilities instead of not modeling it at all?

geoffw0:
The DCA run failed spectacularly.  Before we start another, it might be a good idea to merge latest main into this PR so that everything is up-to-date.

MathiasVP:
Indeed, merging in `main` (or anything that includes https://github.com/github/codeql/pull/11181) should fix this.

swift/ql/lib/codeql/swift/frameworks/StandardLibrary/String.qll
d10c: If we opt for leaving `.count` et al as taint propagating steps, maybe we could have a library predicate like `isConversionToSmallType(DataFlow::Node)`, so that queries that don't want this behavior can sanitize it out without reinventing the wheel.

karimhamdanali: I'd say we keep all those fields/members as potential sources of taint until proven otherwise. As folks have mentioned here, even integers can be sources of taint in some cases. I'm wondering though if there's a way to implement @d10c suggestion without the query writer doing too much work here.

karimhamdanali:
Sounds good. Let me do that then, because my won DCA run also failed on all levels! I'll start another DCA run once I merge/rebase with main.

geoffw0 (dismissed):
My DCA run succeeded and I don't see anything concerning in the results. :tada:

geoffw0 (approved):
:tada:

PR #11187, opened 2022-11-09
Ruby: add ActionCable channel RPC params as remote flow sources

nickrolfe:


ruby/ql/lib/codeql/ruby/frameworks/ActionCable.qll
```diff
@@ -26,4 +28,33 @@ module ActionCable {
...
+        // Any method on a subclass of `ActionCable::Channel::Base`
+        // automatically becomes an RPC endpoint
+        m =
+          DataFlow::getConstant("ActionCable")
+              .getConstant("Channel")
+              .getConstant("Base")
+              .getADescendentModule()
+              .getAnOwnInstanceMethod() and
+        // as long as it's public
+        m.asCallableAstNode().isPublic() and
```
asgerf, line 48: Would you mind adding `isVisible()` to `DataFlow::MethodNode`?

From JS my experience was that we should have added a lot of these trivial things sooner rather than later.

ruby/ql/lib/codeql/ruby/frameworks/ActionCable.qll
```diff
@@ -26,4 +28,33 @@ module ActionCable {
...
+    ActionCableChannelRpcParam() {
+      exists(DataFlow::MethodNode m |
+        // Any method on a subclass of `ActionCable::Channel::Base`
+        // automatically becomes an RPC endpoint
+        m =
+          DataFlow::getConstant("ActionCable")
+              .getConstant("Channel")
+              .getConstant("Base")
+              .getADescendentModule()
+              .getAnOwnInstanceMethod() and
```
asgerf, line 46: If the subclass `include`s a another module, the methods from that module do not become RPC endpoints?

ruby/ql/lib/codeql/ruby/frameworks/ActionCable.qll
nickrolfe: Sure, but wouldn't it be better to add `isPublic`, `isPrivate`, and `isProtected`?

ruby/ql/lib/codeql/ruby/frameworks/ActionCable.qll
nickrolfe: I'm not sure, but I'll try and find out.

I must admit I did it this way because I couldn't find a reliable way of excluding methods on the base class when it exists in the database (i.e. in `rails/rails`).

ruby/ql/lib/codeql/ruby/frameworks/ActionCable.qll
asgerf: Yes, you're right, I don't know where I got `isVisible()` from 🤦

ruby/ql/lib/codeql/ruby/frameworks/ActionCable.qll
asgerf: The docs mention that 
> the subscribed and unsubscribed methods are not callable methods, as they were already declared in [ActionCable::Channel::Base](https://api.rubyonrails.org/v6.1.4/classes/ActionCable/Channel/Base.html)

If being declared in `ActionCable::Channel::Base` is truly the criterion for not being an RPC endpoint, then I guess we can expand the list `["subscribed", "unsubscribed"]` to include the names of all such methods. That should serve a double purpose of excluding the methods actually declared in `ActionCable::Channel::Base` as well as those that override them.

nickrolfe: That sounds pretty reasonable, I'll do that.

nickrolfe: And now I've pushed that change. I think I realise now the problem I had originally. The following change seems like it should be equivalent, but it's not:

```
diff --git a/ruby/ql/lib/codeql/ruby/frameworks/ActionCable.qll b/ruby/ql/lib/codeql/ruby/frameworks/ActionCable.qll
index a3f9b720c3..7a032be71b 100644
--- a/ruby/ql/lib/codeql/ruby/frameworks/ActionCable.qll
+++ b/ruby/ql/lib/codeql/ruby/frameworks/ActionCable.qll
@@ -35,24 +35,15 @@ module ActionCable {
    */
   class ActionCableChannelRpcParam extends RemoteFlowSource::Range {
     ActionCableChannelRpcParam() {
-      exists(DataFlow::MethodNode m |
+      exists(DataFlow::MethodNode m, DataFlow::ConstRef base |
+        base = DataFlow::getConstant("ActionCable").getConstant("Channel").getConstant("Base") and
         // Any method on a subclass of `ActionCable::Channel::Base`
         // automatically becomes an RPC endpoint
-        m =
-          DataFlow::getConstant("ActionCable")
-              .getConstant("Channel")
-              .getConstant("Base")
-              .getADescendentModule()
-              .getAnInstanceMethod() and
+        m = base.getADescendentModule().getAnInstanceMethod() and
         // as long as it's not an instance method of
         // `ActionCable::Channel::Base` itself, which might exist in the
         // database
-        not m =
-          DataFlow::getConstant("ActionCable")
-              .getConstant("Channel")
-              .getConstant("Base")
-              .asModule()
-              .getAnInstanceMethod() and
+        not m = base.asModule().getAnInstanceMethod() and
         // and as long as it's public
         m.asCallableAstNode().isPublic() and
         // and is not called `subscribed` or `unsubscribed`.
```

asgerf (approved):
👍

PR #11188, opened 2022-11-09
Java: Improve sink model generation precision by excluding variable capture.

aschackmull:
Variable-capture-as-jump-steps discards the call context, which is crucial to sink model generation precision.

### DIFF
--- a/csharp/ql/src/utils/model-generator/internal/CaptureModels.qll
+++ b/csharp/ql/src/utils/model-generator/internal/CaptureModels.qll
@@ -272,6 +272,8 @@ private class PropagateToSinkConfiguration extends TaintTracking::Configuration 
 
   override predicate isSink(DataFlow::Node sink) { ExternalFlow::sinkNode(sink, _) }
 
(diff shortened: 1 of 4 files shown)


tamasvajk (approved):

michaelnebel:
> Variable-capture-as-jump-steps discards the call context, which is crucial to sink model generation precision.

Could you provide a concrete example to help me understand?

Furthermore, does this only hold for Java or can we make a similar sanitizer for C#?

aschackmull:
> Could you provide a concrete example to help me understand?

```
void apiEntry(x) {
  lambda = wrapInLambda(x)
}

Producer<T> wrapInLambda(T x) {
  return () -> x;
}

void unrelated() {
  lambda = wrapInLambda(safeValue);
  sink(lambda.apply());
}
```
This code yields a FP flow path from the `apiEntry` parameter `x` to `sink`, and thus wrongly constructs `apiEntry` as an additional sink model.

I believe variable capture happens slightly differently in C#, so the corresponding FP paths might look a bit different, but as long as the capture happens via a jump-step then there's a similar FP risk in C#. But it probably makes sense to look at some paths to see how widespread it is.

PR #11191, opened 2022-11-09
RB: add join(" ") calls as a sink for rb/shell-command-constructed-from-input

erik-krogh:
Inspired by CVE-2021-33473 (recognizes a sink there). 

[Evaluation looks OK](https://github.com/github/codeql-dca-main/tree/data/erik-krogh/PR-11191-0-ruby__2/reports).  
There is one new result, which looks benign (and there was already an alert on the line below).


aibaars (approved):
LGTM

p-:
@erik-krogh Very cool! I manually found such an instance and wanted to write an additional taint step this week. Now I don't have to 🎉 
(I'll check if it can be found with these changes)

PR #11192, opened 2022-11-09
Share encryption key sizes between Java and Python

jcogs33:
This PR shares encryption key sizes between Java and Python.
It is based on discussion with @RasmusWL.


python/ql/lib/semmle/python/security/internal/EncryptionKeySizes.qll
```diff
@@ -0,0 +1,15 @@
+/**
+ * INTERNAL: Do not use.
+ *
+ * Provides predicates for recommended encryption key sizes.
+ * Such that we can share this logic across our CodeQL analysis of different languages.
+ */
+
+/** Returns the minimum recommended key size for asymmetric algorithms (RSA, DSA, and DH). */
+int minSecureKeySizeAsymmetricNonEc() { result = 2048 }
```
RasmusWL, line 9: Only part I'm concerned about it whether minimum secure keysize for RSA/DSA/DH will always be the same. Personally, I would favor a predicate for each, but if you're more informed on this area than me, I'm happy to listen to reason :wink:

RasmusWL (changes requested):
Overall looks good to me, really nice to have this shared! 

Thanks for fixing up our .qhelp file as well :+1:

python/ql/lib/semmle/python/security/internal/EncryptionKeySizes.qll
RasmusWL: This is my thinking:

In a hypothetical scenario, I read an article showing that DSA needs a keysize of 4096 to be secure. I'll go fix our query, and need to bump this number. If there is only ONE shared number, I need to consider whether this change is correct for RSA/DH as well. The encryption community might not have this figured out yet, so giving alerts for RSA/DH key sizes of 2048 would all be considered FPs.

A worse scenario of the above is that DSA is determined to need 4096, and we merge that change without checking if it applies to RSA/DH, which it then turns out it didn't :flushed: 

With a minimum secure keysize for each algorithm, it will be super easy to adjust keysizes individually, without fear of unintended side effects.

My hypothetical scenario above might be considered paranoid, but in terms of creating an easily maintainable codebase that will set us up for success in the future, I would like to see this change. _(I'm probably at 7/10 on [The Sliding Scale of Giving a Fuck](http://capwatkins.com/blog/the-sliding-scale-of-giving-a-fuck))_

RasmusWL: I obviously don't want to make a huge deal out of this, and this request can absolutely be overruled... if we do leave it like it is right now, I do want a comment on _why_ this approach is acceptable, since it's not obvious without deep knowledge of crypto algorihtms.

jcogs33: > in terms of creating an easily maintainable codebase that will set us up for success in the future, I would like to see this change (I'm probably at 7/10 on [The Sliding Scale of Giving a Fuck](http://capwatkins.com/blog/the-sliding-scale-of-giving-a-fuck))

I don't mind making the change if you feel strongly about it (I'm probably only at a 4/10 for keeping it as is 🙂). As discussed elsewhere, based on my knowledge of the math behind these algorithms, I doubt that the minimum key sizes for these will diverge, but I don't mind making the change anyways.
@atorralba Any additional thoughts on this?

atorralba: As mentioned elsewhere, I neither have a strong opinion about this. I felt inclined to leave it as one predicate if the alternative implied too many changes in the existing libraries, since according to [NIST SP 800-57 Part 1 ](https://nvlpubs.nist.gov/nistpubs/SpecialPublications/NIST.SP.800-57pt1r5.pdf)(see Table 2) equal key sizes produce equal security strength for all three algorithms.

But if the change is simple enough I guess it doesn't hurt to have a little more flexibility in our libraries, just in case the cryptography standards do change at some point in the future 🤷‍♂️ .

RasmusWL: I see all the info on how this should just work, but personally I've seen my own assumptions on how the world works fail too many times, so I'm starting to favor simple/stupid but obvious solutions :sweat_smile: 

I'll try to make a PR for this change tomorrow (ran out of time today). I've looked over the Java code, and have a good idea of how to adopt it to what I'm asking for :blush:

atorralba: Please coordinate with @jcogs33 if that's the case because AFAIK she's also thinking about how to do the change you suggested.

jcogs33: Yep, I experimented with splitting up the algorithms to see how complicated it would make the Java code. I've put a naive approach in commit https://github.com/github/codeql/pull/11192/commits/0744e5af11cf5b3da081b4d82c1ac8bb3693fedf. I don't particularly like this approach because it feels like there's a lot of code duplication across the Rsa/Dsa/Dh modules. I can maybe reduce the code duplication, but my initial quick attempt at doing so resulted in test case failures, so I'll need to spend more time looking into that. @rasmuswl, let me know if you were thinking of a different approach.

RasmusWL: https://github.com/RasmusWL/codeql/commit/79ade1160a830a4b6538407b5e9f7e0d1eaf807a is an approach that has less duplication, and still gets the job done. Feel free to adapt the code in this PR if you like it :blush:

jcogs33: Thanks @RasmusWL! I like it, and I've added it to this PR in commit https://github.com/github/codeql/pull/11192/commits/e64ca97fd65144598485d7ac16ca37a737151865. I changed the source slightly so that it also uses `algoName` as an argument to `getMinKeySize` and so that its `hasState` predicate still matches the sink's (necessary for the sources and sinks to be matched correctly). Let me know if these changes good to you.

RasmusWL (approved):
Python :+1:

RasmusWL (approved):
actually, we needed a change-note for Python, but I've just added one myself :+1:

java/ql/lib/semmle/code/java/security/InsufficientKeySize.qll
```diff
@@ -21,39 +22,67 @@ private module Asymmetric {
...
 
     /** A sink for an insufficient key size used in RSA, DSA, and DH algorithms. */
     private class Sink extends InsufficientKeySizeSink {
+      string algoName;
+
       Sink() {
         exists(KeyPairGenInit kpgInit, KeyPairGen kpg |
-          kpg.getAlgoName().matches(["RSA", "DSA", "DH"]) and
+          algoName in ["RSA", "DSA", "DH"] and
+          kpg.getAlgoName().matches(algoName) and
```
atorralba, line 41: Is `matches` needed here? Or could we just use equality?

atorralba (approved):
One minor nitpick, otherwise Java LGTM 👍.

java/ql/lib/semmle/code/java/security/InsufficientKeySize.qll
jcogs33: Good question. @RasmusWL, is there a reason you used `matches` here instead of equality? 
Using equality instead would be consistent with `algoName = spec.getAlgoName()` below.

RasmusWL: I think the original code used `matches` :shrug: I agree that using equality seems more appropriate.

```suggestion
          kpg.getAlgoName() = algoName and
```

jcogs33: Thanks! I've applied the change.
Can one of you re-approve when you have a chance? 🙏

atorralba (dismissed):
Always happy to get more code shared between languages! 🎉

atorralba (approved):

atorralba:
@jcogs33 the failing CI check can be ignored (not related to this PR), so feel free to merge.

PR #11198, opened 2022-11-10
SSA: Expose phi-read nodes

hvitved:
This PR follows up on https://github.com/github/codeql/pull/10035 by exposing "phi read" definitions in the shared SSA library.

The motivation for exposing phi reads is to be able to avoid the same potential combinatorial explosion that we also avoid by including normal phi nodes in the data flow graph:

```mermaid
  graph TD;
      A-->read_phi;
      B-->read_phi;
      C-->read_phi;
      read_phi-->D;
      read_phi-->E;
      read_phi-->F;
```

Assuming that `A,...,F` read the same underlying variable as `read_phi`, one can reduce the number of data flow edges from 9 (quadratic) to 6 (linear) by including the `read_phi` join-point.

This PR does not actually include read-phis in the data flow graph for any languages (follow-up work, e.g. https://github.com/github/codeql/pull/10927), it merely exposes them.

The old API (which ignores phi-reads) is maintained, and if one wants to take phi-reads into account, the new class `DefinitionExt` should be used instead of the existing `Definition` class. Existing predicates for `Definition` now exist in both the old version and a new `Ext`-suffixed version for `DefinitionExt`.

### DIFF
--- a/csharp/ql/consistency-queries/SsaConsistency.ql
+++ b/csharp/ql/consistency-queries/SsaConsistency.ql
@@ -1,8 +1,17 @@
 import csharp
-import semmle.code.csharp.dataflow.internal.SsaImpl::Consistency
+import semmle.code.csharp.dataflow.internal.SsaImpl as Impl
+import Impl::Consistency
 import Ssa
 
 class MyRelevantDefinition extends RelevantDefinition, Ssa::Definition {
+  override predicate hasLocationInfo(
+    string filepath, int startline, int startcolumn, int endline, int endcolumn
+  ) {
+    this.getLocation().hasLocationInfo(filepath, startline, startcolumn, endline, endcolumn)
+  }
+}
+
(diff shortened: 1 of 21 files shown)


hvitved (commented):
Note to self: Remember to revert the last "Trigger CI" commit.

MathiasVP:
I've only reviewed 81a1fa167ab6569bb0956ede441af3e1b31bf63b, but that one certainly LGTM! The scary C++ CI failures can be ignored, although they _might_ be hiding some changes to `.expected` that need to be accepted.

MathiasVP (commented):
LGTM :shipit:! The Swift .expected files need to be updated. The scary C++ CI failures can be ignored, although they _might_ be hiding some changes to `.expected` that need to be accepted.

MathiasVP (approved):
LGTM!

PR #11201, opened 2022-11-10
QL-for-QL: don't use the deprecated set-output feature in github-actions

erik-krogh:
`set-output` is deprecated and will stop working soon: https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/ 

CodeQL-action was updated to the latest `main` (with the patch that adds `ql` as another language).


aibaars (approved):
Looks good to me.

PR #11203, opened 2022-11-10
C#: update cs/assembly-path-injection cs/hardcoded-key to path-problems

erik-krogh:
Something I randomly stumbled upon a while ago.  

The two queries look very much like `@kind path-problems` queries, and can easily be converted to that format, but they were `@kind problem` queries for some reason? 

[This blackbird search](https://cs.github.com/github/codeql?q=path%3A%2F.*%5C%2Fsrc%5C%2F.*%5C.ql%2F+%2Fwhere+.*%5C.hasFlow%5C%28%2F+) helped me find them. 

Does this need a change-note?


hvitved (approved):
Thanks!

PR #11204, opened 2022-11-10
ATM: add XSSThroughDOM boosted query

jhelie:


tiferet:
This is exciting! Since this is a [risky change](https://github.com/github/ml-ql-adaptive-threat-modeling/blob/main/ADRs/001_ATM_testing_and_updating.md#appendix-a-glossary), we do [need end-to-end testing and an ATM update process](https://github.com/github/ml-ql-adaptive-threat-modeling/blob/main/ADRs/001_ATM_testing_and_updating.md#when-is-an-update-process-needed). Hopefully this will be easier now that we can use the orchestrator at least through the endpoint pipeline stage 🎉

PR #11205, opened 2022-11-10
Swift: db up/downgrade scripts

AlexDenisov:
This is inspired by the [ruby checks](https://github.com/github/codeql/blob/main/.github/workflows/ruby-qltest.yml#L51-L66).

<s>The scripts check is running on macOS since the first command yields no results and breaks the whole invocation.
Perhaps, xargs works differently on macOS and Linux?
```
codeql resolve upgrades --format=lines --allow-downgrades --additional-packs downgrades \
         --dbscheme=ql/lib/swift.dbscheme --target-dbscheme=downgrades/initial/swift.dbscheme |
         xargs codeql execute upgrades testdb
```
</s>


swift/actions/database-upgrade-scripts/action.yml
```diff
@@ -0,0 +1,23 @@
...
+    - name: Check downgrade scripts
+      shell: bash
+      working-directory: swift
+      run: |
+        echo > empty.trap
+        rm -rf testdb
+        codeql dataset import -S ql/lib/swift.dbscheme testdb empty.trap
+        codeql resolve upgrades --format=lines --allow-downgrades --additional-packs downgrades \
+         --dbscheme=ql/lib/swift.dbscheme --target-dbscheme=downgrades/initial/swift.dbscheme |
+         xargs codeql execute upgrades testdb
```
redsun82, line 22: This will make `xargs` not run on empty
```suggestion
         xargs -r codeql execute upgrades testdb
```

redsun82 (commented):
> This is inspired by the [ruby checks](https://github.com/github/codeql/blob/main/.github/workflows/ruby-qltest.yml#L51-L66).
> 
> The scripts check is running on macOS since the first command yields no results and breaks the whole invocation. Perhaps, xargs works differently on macOS and Linux?
> 
> ```
> codeql resolve upgrades --format=lines --allow-downgrades --additional-packs downgrades \
>          --dbscheme=ql/lib/swift.dbscheme --target-dbscheme=downgrades/initial/swift.dbscheme |
>          xargs codeql execute upgrades testdb
> ```

I think the problem is the behaviour of `xargs` when the input is empty (which is currently the case because we have no scripts :sweat_smile:). By default on Linux the command is still run once with no input, and I'm guessing macOS does not run it at all (which by the way tbh seems the most sensible thing to do :slightly_smiling_face:).

So either we wait to add at least one upgrade/downgrade script, or we add `-r` to skip `xargs` on empty.

AlexDenisov:
Right, from the `man` page:

```
     -r      Compatibility with GNU xargs.  The GNU version of xargs runs the utility argument at least once, even if xargs input is empty, and it supports a -r option to
             inhibit this behavior.  The FreeBSD version of xargs does not run the utility argument on empty input, but it supports the -r option for command-line
             compatibility with GNU xargs, but the -r option does nothing in the FreeBSD version of xargs.
```

.github/workflows/swift.yml
```diff
@@ -111,4 +111,19 @@ jobs:
       - uses: actions/upload-artifact@v3
         with:
           name: swift-generated-cpp-files
-          path: swift/generated-cpp-files/**
\ No newline at end of file
+          path: swift/generated-cpp-files/**
+  qlformat:
```
AlexDenisov, line 115: Hmm, this is most likely a rebase artifacts, not sure where it comes 🤔

jketema: It's just a new line missing at the end of the file before this change. This adds a new line, so 114 changes.

jketema (approved):

.github/workflows/swift.yml
jketema: Assuming you also wanted to add a formatting check.

AlexDenisov: Yeah, this is indeed a rebase artifact, the check was removed here https://github.com/github/codeql/pull/11214

redsun82 (approved):
:shipit: But please also update the obsolete PR description :slightly_smiling_face:

PR #11209, opened 2022-11-10
Swift: drop impossible nodes from schema

redsun82:
There were a bunch of nodes that were not really possible to get on a sensible extractor run. See comments to `SwiftTagTraits.h` for specific reasons.

In order to gracefully treat them, we can now alias the relevant tag as `void` in `SwiftTagTraits.h`, which will cause those swift AST classes to be ignored when encountered.

If such a class does indeed appear in an extraction run, then
* in debug mode, [this assertion](https://github.com/github/codeql/blob/e18b2cfa3921dbc93e6c92b593bc4b004a7eacab/swift/extractor/infra/SwiftDispatcher.h#L161) will fire
* in release mode, an `undefined_label` will be fetched for that entity, which will result in an `UnspecifiedElement` being emitted in its place.


AlexDenisov (approved):
That's a great improvement! :shipit:

PR #11210, opened 2022-11-10
Swift: Add Alamofire model to swift/cleartext-transmission

geoffw0:
Add [Alamofire](https://github.com/Alamofire/Alamofire) library model to `swift/cleartext-transmission`.  Results are disappointing at the moment, I've spent quite a bit of time trying to figure out why and I believe we need:
 - content flow through dictionaries and perhaps tuples (in progress https://github.com/github/codeql-c-team/issues/1344).
 - we may need an `allowImplicitRead` predicate on this taint configuration. (issue: https://github.com/github/codeql-c-team/issues/1386)
 - flow through the `HTTPHeaders` class (straightforward but partially depends on the previous points). (issue: https://github.com/github/codeql-c-team/issues/1386)
 - in addition I believe something is going wrong specifically with the version of `request` that declares `<Parameters: Encodable>`. (issue: https://github.com/github/codeql-c-team/issues/1386)
 - we may also want to model `AF.upload`. (issue: https://github.com/github/codeql-c-team/issues/1385)

That's a lot, so for now I'd like to get the tests and basic coverage in.


geoffw0:
Tests pass for me locally.  I will retry / investigate...

geoffw0:
> geoffw0 requested review from github/codeql-kotlin and github/codeql-java as code owners 1 minute ago

No, no I didn't.  :(

geoffw0:
DCA LGTM!  Would be good to finally merge this.

swift/ql/src/queries/Security/CWE-311/CleartextTransmission.ql
```diff
@@ -54,6 +54,25 @@ class Url extends Transmitted {
...
+  AlamofireTransmitted() {
+    // sinks are the first argument containing the URL, and the `parameters`
+    // and `headers` arguments to appropriate methods of `Session`.
+    exists(CallExpr call, string fName |
+      call.getStaticTarget().(MethodDecl).hasQualifiedName("Session", fName) and
+      fName.regexpMatch("(request|streamRequest|download)\\(.*") and
+      (
+        call.getArgument(0).getExpr() = this or
+        call.getArgumentWithLabel("parameters").getExpr() = this or
+        call.getArgumentWithLabel("headers").getExpr() = this
```
atorralba, line 70: We could simplify:

```suggestion
        call.getArgumentWithLabel(["headers", "parameters"]).getExpr() = this
```

swift/ql/src/queries/Security/CWE-311/CleartextTransmission.ql
```diff
@@ -54,6 +54,25 @@ class Url extends Transmitted {
...
+/**
+ * An `Expr` that transmitted through the Alamofire library.
+ */
+class AlamofireTransmitted extends Transmitted {
+  AlamofireTransmitted() {
+    // sinks are the first argument containing the URL, and the `parameters`
+    // and `headers` arguments to appropriate methods of `Session`.
+    exists(CallExpr call, string fName |
+      call.getStaticTarget().(MethodDecl).hasQualifiedName("Session", fName) and
+      fName.regexpMatch("(request|streamRequest|download)\\(.*") and
```
atorralba, line 66: I'm not sure if `matches` is more performant than `regexpMatch` (I think it is). If that's the case, this might be better:

```suggestion
      fName.matches(["request", "streamRequest", "download"] + "(%") and
```

atorralba (approved):
This LGTM and I'm happy for it to be merged as-is (it will be nice to remove all those `[NOT DETECTED]` once we model all the missing flow). Regardless, I added a couple of NITs in case you want to address them.

swift/ql/src/queries/Security/CWE-311/CleartextTransmission.ql
geoffw0: For me this was just over the threshold where its cleaner (and more efficient???) to just use a regexp.  I'm open to persuasion though (particularly on the 'more efficient' part because its only a weak intuition).

atorralba (approved):

swift/ql/src/queries/Security/CWE-311/CleartextTransmission.ql
geoffw0: True.  I am slightly curious which is faster (`match` is generally faster than `regexpMatch` but there's a sense in which this is three uses of `match`).  But lets not worry about it now.

PR #11212, opened 2022-11-10
C++: Improve handling of `std::string::insert` with iterator return type and do some cleanup

jketema:


cpp/ql/lib/semmle/code/cpp/models/implementations/StdString.qll
```diff
@@ -187,26 +177,30 @@ private class StdStringAppend extends TaintFunction {
...
+    // flow from string and parameter to string (qualifier) and return value
+    (
+      input.isQualifierObject() or
+      input.isParameterDeref(this.getAStringParameterIndex()) or
+      input.isParameter(this.getAnIteratorParameterIndex())
+    ) and
+    (
+      output.isQualifierObject() or
+      output.isReturnValue()
+    )
```
geoffw0, line 195: If I'm understanding correctly, this is the only change that will affect behaviour - `std::string::insert` no longer has 'reverse taint flow' from the return value to the qualifier.  What is the reason for this change?  Could it be easily demonstrated in a test?

geoffw0 (commented):
Cleanup looks excellent, and there is already good test coverage for this stuff.

cpp/ql/lib/semmle/code/cpp/models/implementations/StdString.qll
jketema: This also replaces `output.isReturnValueDeref()` by `output.isReturnValue()`. The reason for both changes is that insert returns a fresh iterator instead of a reference to the string that is being modified (like `+=`, `append` and `replace`). So, `output.isReturnValueDeref()` is incorrect and reverse flow seems inappropriate (that's assuming I understand reverse flow correctly).

jketema: Fixed in https://github.com/github/codeql/pull/11212/commits/bfda1f013ebbdecad3b6400d18a28708b2bd5b6f

jketema: > What is the reason for this change?

The reason is that the "return iterator" variants are used in the dataflow taint tests, and the tests are failing on the use-use dataflow feature branch. We are missing flow because of the use of `output.isReturnValueDeref()` even when an iterator is returned.

geoffw0 (dismissed):
LGTM assuming DCA doesn't show us anything unexpected.

jketema:
The version with `insert` merged back into the class it was in showed some DCA performance issues. I pulled it back out, which probably also communicates more clearly that `insert` is special. While here, I reworked the commits a bit.

Re-running DCA.

geoffw0 (approved):
Still LGTM (subject to DCA).

jketema:
I honestly cannot tell whether the DCA results are bad or good. @geoffw0 could you have a look at the last linked one?

MathiasVP:
> I honestly cannot tell whether the DCA results are bad or good. @geoffw0 could you have a look at the last linked one?

This LGTM. I wouldn't read too much into the stage timing if there are no end-to-end timing differences.

PR #11214, opened 2022-11-10
CI: remove langauge specific format checks

erik-krogh:
We now have [this format check](https://github.com/github/codeql/blob/main/.github/workflows/compile-queries.yml#L49) which checks all the languages all the time, so there's no need for language specific format checks.

### DIFF
--- a/.github/workflows/compile-queries.yml
+++ b/.github/workflows/compile-queries.yml
@@ -46,7 +46,7 @@ jobs:
         with:
           channel: 'release'
       - name: check formatting
-        run: codeql query format */ql/{src,lib,test}/**/*.{qll,ql} --check-only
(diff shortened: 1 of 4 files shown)


.github/workflows/compile-queries.yml
```diff
@@ -46,7 +46,7 @@ jobs:
         with:
           channel: 'release'
       - name: check formatting
-        run: codeql query format */ql/{src,lib,test}/**/*.{qll,ql} --check-only
+        run: codeql query format */ql/**/*.{qll,ql} --check-only
```
aeisenberg, line 49: So, this expands the format check to _every_ ql and qll file in the repo. I think that's fine.

aeisenberg (approved):

aibaars (approved):
Nice!

PR #11217, opened 2022-11-10
Java/Kotlin: Write Kotlin version information to the database

igfoo:
We get rows like:
```
compilation_info(#compilation, "Kotlin Compiler Version", "1.7.10")
compilation_info(#compilation, "Kotlin Extractor Name", "codeql-extractor-kotlin-embeddable-1.7.0.jar")
```


java/kotlin-extractor/src/main/kotlin/KotlinExtractorExtension.kt
```diff
@@ -131,6 +131,8 @@ class KotlinExtractorExtension(
             // The interceptor has already defined #compilation = *
             val compilation: Label<DbCompilation> = StringLabel("compilation")
             tw.writeCompilation_started(compilation)
+            tw.writeCompilation_info(compilation, "Kotlin Compiler Version", KotlinCompilerVersion.getVersion() ?: "<unknown>")
```
tamasvajk, line 134: Does `"<unknown>"` give us any value? `compilation_compiler_times` is written conditionally below. Shouldn't we do the same here, and omit the `"Kotlin Compiler Version"` is the version is null?

java/kotlin-extractor/build.py
```diff
@@ -185,6 +185,10 @@ def compile(jars, java_jars, dependency_folder, transform_to_embeddable, output,
     include_version_folder = tmp_src_dir + '/main/kotlin/utils/versions/to_include'
     os.makedirs(include_version_folder)
 
+    with open(tmp_src_dir + '/main/kotlin/utils/ExtractorName.kt', 'w') as f:
+        f.write('package com.github.codeql\n')
+        f.write('val extractor_name: String = "' + output + '"\n')
```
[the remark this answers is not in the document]
igfoo: I've moved it to a resource

java/kotlin-extractor/src/main/kotlin/KotlinExtractorExtension.kt
igfoo: I plan to add it to the telemetry query, and I think that explicitly returning `<unknown>` will be better than just not returning anything. Especially as a single build might have one compilation for which we have a value, and one for which we don't. This will let us easily see if `<unknown>` is common (I expect that it will be very rare, so that would be something to look into).

Putting the `<unknown>` into the database means that we'll be able to do something like
```
predicate compilationInfo(string key, int value) {
  exists(Compilation c, string infoKey |
    key = infoKey + ": " + c.getInfo(infoKey) and
    value = 1
  )
}
```
without having to worry about unknown values specially in the QL. And that also means that the QL doesn't need to know what keys are valid.

tamasvajk (approved):

PR #11226, opened 2022-11-11
CI: use `find` in the format check to fix it

erik-krogh:
The previous approach worked on my local machine, but apparently that didn't translate to the cloud.  
So now I'm using the `find` command instead, which checks formatting on all files.  

For some reason the previous approach skipped a lot of files, and I'm not sure why.


nickrolfe (approved):
Looks right to me, and it at least works on my Mac.

PR #11227, opened 2022-11-11
Remove issue template for LGTM.com false positive reports

sj:
[LGTM.com is in the process of being deprecated](https://github.blog/2022-08-15-the-next-step-for-lgtm-com-github-code-scanning/). This PR removes the GitHub Issue template that is dedicated to false positive reports on LGTM.com.

Fixes: https://github.com/github/code-scanning/issues/7888


aibaars:
@sj Shouldn't we add an alternative template for reporting false positives. Otherwise they would all end up being reported as "general issues".

turbo (approved):
LGTM, replaced by: https://github.com/github/codeql/pull/11276#pullrequestreview-1182406362

PR #11232, opened 2022-11-11
Go: Optimize trap.Writer by buffering gzip writes

grddev:
The TRAP writer already buffers writes before emitting to file, but running gzip compression is also fairly costly (especially if you only do it a couple of bytes at a time). Thus, this injects another buffer that collects the emitted tuples in string form, and only triggers gzip compression once the buffer is full. In my local testing, this buffering was actually more beneficial than the one between gzip and file (likely because the gzip writer already emits data in chunks), but that one is still beneficial.

### DIFF
--- a/go/extractor/trap/trapwriter.go
+++ b/go/extractor/trap/trapwriter.go
@@ -19,7 +19,8 @@ import (
 // A Writer provides methods for writing data to a TRAP file
 type Writer struct {
 	zip             *gzip.Writer
-	w               *bufio.Writer
+	wzip            *bufio.Writer
+	wfile           *bufio.Writer
 	file            *os.File
 	Labeler         *Labeler
 	path            string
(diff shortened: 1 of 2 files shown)


go/extractor/trap/trapwriter.go
```diff
@@ -88,13 +91,18 @@ func trapFolder() (string, error) {
 
 // Close the underlying file writer
 func (tw *Writer) Close() error {
-	err := tw.zip.Close()
+	err := tw.wzip.Flush()
+	if err != nil {
+		// throw away file close error
+		tw.file.Close()
+	}
```
smowton, line 98:
```suggestion
		tw.file.Close()
		return err
	}
```

smowton (dismissed):
Looks sensible; one minor tweak to avoid a double close of `tw.file` in event of a write error.

grddev:
@smowton Thanks for taking a look. Applied your fix, but also noticed that the tests were failing. I realized I had missed the direct references to `tw.zip` in `labels.go`. Hopefully, the test should pass now.

smowton (approved):

PR #11234, opened 2022-11-11
C++: Recognize `basic_string::iterator` as an iterator

jketema:
Follow-up to https://github.com/github/codeql/pull/11212. I played around a bit in that PR with:
```codeql
int getAnIteratorParameterIndex() { this.getParameter(result).getType() instanceof Iterator }
```
and
```codeql
int getAnIteratorParameterIndex() { this.getParameter(result).getUnspecifiedType() instanceof Iterator }
```
I eventually went for the first, but never tested that on the use-use dataflow feature branch. Turns out that doesn't work, as the iterator is typedef'ed in `basic_string` in out taint tests. So, we either need `getUnspecifiedType` or make sure the typedef is recoginzed as an iterator. The latter seems safer, as the C++ standard does not guarantee that `basic_string::iterator` needs to be a typedef.

### DIFF
--- a/cpp/ql/lib/semmle/code/cpp/models/implementations/StdString.qll
+++ b/cpp/ql/lib/semmle/code/cpp/models/implementations/StdString.qll
@@ -16,6 +16,15 @@ private class StdBasicString extends ClassTemplateInstantiation {
 }
 
 /**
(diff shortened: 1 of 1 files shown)


MathiasVP (approved):
LGTM is DCA is happy!

PR #11241, opened 2022-11-12
Java: Query to detect Android Webview file access

egregius313:
Added a query for detecting settings of Android WebViews which enable access to the file system.

### DIFF
--- a/java/ql/src/Security/CWE/CWE-200/AndroidWebViewSettingsFileAccess.qhelp
+++ b/java/ql/src/Security/CWE/CWE-200/AndroidWebViewSettingsFileAccess.qhelp
@@ -12,7 +12,9 @@
   </overview>
 
   <recommendation>
-    <p>When possible, you should disallow file access by setting the following settings to <code>false</code>:</p>
+    <p>When possible, you should not allow file access. The file access settings
+    are disabled by default. You can explicitly disbale them by setting the
+    following settings to <code>false</code>:</p>
 
     <ul>
       <li><code>setAllowFileAccess</code></li>
--- a/java/ql/src/Security/CWE/CWE-200/AndroidWebViewSettingsFileAccess.qhelp
+++ b/java/ql/src/Security/CWE/CWE-200/AndroidWebViewSettingsFileAccess.qhelp
@@ -4,16 +4,16 @@
 <qhelp>
   <overview>
     <p>
-      File access in an Android WebView can expose the device's file system to
-      the JavaScript running in the WebView. If there are vulnerabilities in the
-      JavaScript or untrusted content is loaded in the WebView, file access may
-      allow an attacker to access or steal the user's data.
+      Allowing file access in an Android WebView can expose a device's file system to
+      the JavaScript running in that WebView. If the JavaScript contains
+      vulnerabilities or the WebView loads untrusted content, file access
+      allows an attacker to steal the user's data.
     </p>
   </overview>
 
   <recommendation>
-    <p>When possible, you should not allow file access. The file access settings
-    are disabled by default. You can explicitly disbale them by setting the
+    <p>When possible, do not allow file access. The file access settings
+    are disabled by default. You can explicitly disable file access by setting the
     following settings to <code>false</code>:</p>
 
     <ul>
@@ -23,14 +23,14 @@
     </ul>
 
     <p>If your application requires access to the file system, it is best to
-    avoid using <code>file://</code> urls, and instead use an alternative that
-    allows loading files via https, such
+    avoid using <code>file://</code> URLs. Instead, use an alternative that
+    loads files via HTTPS, such
     as <code>androidx.webkit.WebViewAssetLoader</code>.</p>
   </recommendation>
 
   <example>
-    <p>In the following (bad) example, the WebView is configured with the settings
-    which would allow local file access.</p>
+    <p>In the following (bad) example, the WebView is configured with settings
+    that allow local file access.</p>
 
     <sample src="WebViewFileAccessUnsafe.java"/>
 
@@ -39,8 +39,9 @@
     <sample src="WebViewFileAccessSafe.java"/>
 
     <p>
-      In the following example, an asset loader is configured so that assets can
-      be access over <code>https</code>, and not require file system access.
+      As mentioned previously, asset loaders can load files without file system 
+      access. In the following (good) example, an asset loader is configured to 
+      load assets over HTTPS.
     </p>
 
     <sample src="AssetLoaderExample.java"/>
--- a/java/ql/src/Security/CWE/CWE-200/AndroidWebViewSettingsFileAccess.qhelp
+++ b/java/ql/src/Security/CWE/CWE-200/AndroidWebViewSettingsFileAccess.qhelp
@@ -21,6 +21,11 @@
       <li><code>setAllowFileAccessFromFileURLs</code></li>
       <li><code>setAllowUniversalAccessFromFileURLs</code></li>
     </ul>
+
+    <p>If your application requires access to the file system, it is best to
+    avoid using <code>file://</code> urls, and instead use an alternative that
+    allows loading files via https, such
+    as <code>androidx.webkit.WebViewAssetLoader</code>.</p>
   </recommendation>
 
   <example>
@@ -44,6 +49,9 @@
     </li>
     <li>
       Android documentation: <a href="https://developer.android.com/reference/android/webkit/WebSettings#setAllowUniversalAccessFromFileURLs(boolean)">WebSettings.setAllowUniversalAccessFromFileURLs</a>.
+    </li>
+    <li>
+      Android documentation: <a href="https://developer.android.com/reference/androidx/webkit/WebViewAssetLoader">WebViewAssetLoader</a>.
     </li>
   </references>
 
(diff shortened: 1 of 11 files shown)


java/ql/src/Security/CWE/CWE-200/AndroidWebViewSettingsFileAccess.qhelp
```diff
@@ -0,0 +1,57 @@
...
+    <p>
+      File access in an Android WebView can expose the device's file system to
+      the JavaScript running in the WebView. If there are vulnerabilities in the
+      JavaScript, file access may allow an attacker to access or steal the
+      user's data.
+    </p>
+  </overview>
+
+  <recommendation>
+    <p>When possible, you should disallow file access by setting the following settings to <code>false</code>:</p>
```
atorralba, line 15: I would mention this is `false` by default, so leaving it untouched would also suffice to prevent the issue. Also, we should probably recommend the use of [`AssetLoader`](https://developer.android.com/reference/androidx/webkit/WebViewAssetLoader) instead as well (see the note in https://developer.android.com/reference/android/webkit/WebSettings#setAllowFileAccess(boolean)).

java/ql/src/Security/CWE/CWE-200/AndroidWebViewSettingsFileAccess.qhelp
```diff
@@ -0,0 +1,57 @@
...
+    </ul>
+  </recommendation>
+
+  <example>
+    <p>In the following (bad) example, the WebView is configured with the settings
+    which would allow local file access.</p>
+
+    <sample src="WebViewFileAccessUnsafe.java"/>
+
+    <p>In the following (good) example, the WebView is configured to disallow file access.</p>
```
atorralba, line 30: I think an example of how to use [`AssetLoader`](https://developer.android.com/reference/androidx/webkit/WebViewAssetLoader) as a safe alternative would be interesting as well.

java/ql/src/Security/CWE/CWE-200/AndroidWebViewSettingsFileAccess.ql
```diff
@@ -0,0 +1,19 @@
...
+ * @tags security
+ *       external/cwe/cwe-200
+ */
+
+import java
+import semmle.code.java.frameworks.android.WebView
+
+from MethodAccess ma
+where ma.getMethod() instanceof CrossOriginAccessMethod
+select ma, "WebView setting $@ may allow for unauthorized access of sensitive information.", ma,
```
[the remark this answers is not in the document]
atorralba: Please address this.

java/ql/src/Security/CWE/CWE-200/AndroidWebViewSettingsFileAccess.qhelp
```diff
@@ -40,18 +40,6 @@
     <li>
       Android documentation: <a href="https://developer.android.com/reference/android/webkit/WebSettings#setAllowFileAccessFromFileURLs(boolean)">WebSettings.setAllowFileAccessFromFileURLs</a>.
     </li>
-    <li>
-      Android documentation: <a href="https://developer.android.com/reference/android/webkit/WebSettings#setAllowUniversalAccessFromFileURLs(boolean)">WebSettings.setAllowUniversalAccessFromFileURLs</a>.
```
atorralba: I think this could stay.

java/ql/test/query-tests/security/CWE-200/semmle/tests/options
```diff
@@ -1 +1 @@
-//semmle-extractor-options: --javac-args -cp ${testdir}/../../../../../stubs/apache-commons-lang3-3.7/
\ No newline at end of file
+//semmle-extractor-options: --javac-args -cp ${testdir}/../../../../../stubs/apache-commons-lang3-3.7/:${testdir}/../../../../../stubs/google-android-9.0.0
```
atorralba, line 1: Is the `apache-commons-lang` stub needed?

java/ql/test/query-tests/security/CWE-200/semmle/tests/WebViewFileAccess.java
```diff
@@ -0,0 +1,14 @@
+import android.webkit.WebView;
+import android.webkit.WebSettings;
+
+class WebViewFileAccess {
+    void configure(WebView view) {
```
atorralba, line 5: Maybe add a negative test case? (i.e. `false` being set instead)

java/ql/src/Security/CWE/CWE-200/AndroidWebViewSettingsFileAccess.ql
```diff
@@ -0,0 +1,18 @@
+/**
+ * @name Android WebSettings file access
+ * @kind problem
+ * @description Enabling access to the file system in a WebView can enable access to sensitive information.
```
atorralba, line 4: This sentence seems like it needs a little rewriting (repetition of "enable access"). We can wait for docs review though, if you prefer.

java/ql/test/query-tests/security/CWE-200/semmle/tests/options
egregius313: Yes, but not for these tests. That's for a pre-existing test for another CWE-200 test: `TempDirLocalInformationDisclosure.qlref`

java/ql/src/Security/CWE/CWE-200/AndroidWebViewSettingsFileAccess.qhelp
egregius313: Oh that was a mistake. I'll fix that.

java/ql/test/query-tests/security/CWE-200/semmle/tests/options
atorralba: Ah makes sense, thanks!

java/ql/src/Security/CWE/CWE-200/AndroidWebViewSettingsFileAccess.ql
```diff
@@ -0,0 +1,22 @@
...
+ * @id java/android-websettings-file-access
+ * @problem.severity warning
+ * @security-severity 6.5
+ * @precision high
```
atorralba, line 8: Since this won't always be a security problem, let's get this down to medium so that the query becomes opt-in for users that want to see more results.

```suggestion
 * @precision medium
```

java/ql/src/Security/CWE/CWE-200/AndroidWebViewSettingsFileAccess.ql
```diff
@@ -0,0 +1,22 @@
+/**
+ * @name Android WebSettings file access
+ * @kind problem
+ * @description Enabling access to the file system in a WebView can enable access to sensitive information.
```
sabrowning1, line 4:
```suggestion
 * @description Enabling access to the file system in a WebView allows attackers to view sensitive information.
```

Small edit for word choice

java/ql/src/Security/CWE/CWE-200/AndroidWebViewSettingsFileAccess.qhelp
```diff
@@ -0,0 +1,64 @@
...
+    <p>When possible, you should not allow file access. The file access settings
+    are disabled by default. You can explicitly disbale them by setting the
+    following settings to <code>false</code>:</p>
+
+    <ul>
+      <li><code>setAllowFileAccess</code></li>
+      <li><code>setAllowFileAccessFromFileURLs</code></li>
+      <li><code>setAllowUniversalAccessFromFileURLs</code></li>
+    </ul>
+
```
sabrowning1, line 24: There's some weird spacing/indentation here in the [preview](https://github.com/github/codeql/pull/11241#issuecomment-1312502915) generated by the bot and I'm not sure why 🤔 will that go away in production? Or does that somehow need to be addressed here?

sabrowning1 (commented):
👋🏼 @egregius313, thanks for your work on the documentation! I've suggested some small changes for concision/to bring things in line with our style guide; please let me know if you have any questions.

java/ql/src/Security/CWE/CWE-200/AndroidWebViewSettingsFileAccess.qhelp
egregius313: I believe this is an issue from the way the Markdown is being rendered. But I will ask others if this is a common issue.

felicitymay: I just had a look at other Qhelp files. It seems as if this is a bug in the Markdown generator that we haven't noticed before. (🎖️ for Sam)

If you look at: https://github.com/github/codeql/blob/1bf82cecfa7b7f8365c58d08583985ce84df78e4/python/ql/src/Security/CWE-327/InsecureProtocol.qhelp#L46, it seems as if this should be a new, unindented paragraph. But it's currently being generated as wrapped into the bullet above: https://codeql.github.com/codeql-query-help/python/py-insecure-default-protocol/.

atorralba (approved):
🎉

egregius313:
> Thanks for applying those suggestions @egregius313, the documentation looks good to me! Do you mind creating an issue for that bug in the Markdown generator? Otherwise, I can open one 👍🏼

@sabrowning1 I have asked some of the other CodeQL engineers and they referred me to the appropriate issue. I'm going to mention this PR as an example there.

PR #11243, opened 2022-11-14
Java: Use data extensions for MaD models.

michaelnebel:
## Description
In this PR we start making use of data extensions for MaD models instead of inlining the in the code.
There is a high level of re-use of the scripts originally committed as a part of doing the same exercise for C#: https://github.com/github/codeql/pull/10777

### Details
More concretely in this PR we
- Introduce a script for converting the existing inlined MaD models to data extensions.
- Introduce a script for invoking the model generator queries and emit data extension files.
- Include all manual models as data extensions and remove all the inlined models (the manual models are now grouped by their package name (one package name pr. file) and located in `lib/ext`).
- Re-generate the models for `commons-io` as data extensions and include these and delete the inlined models.
- Define all the relevant extensible predicates and new framework functionality in `ExternalFlow.qll` and include the data extensions in the `lib/qlpack.yml` file.
- Convert all experimental queries to use data extensions. This requires that we introduce some new extensible predicates that allows filtering such that models can be applied on query basis. It also means that the format for the models used by experimental queries has one extra column, which provides a *name* that we can use for filtering.
- Convert all integration tests (and update their qlpacks) to use data extensions.
- Convert all unit tests to use data extensions (and update the qlpack).

### Prerequisites
The pre-requisites are also a part of this PR (the first two commits).
- Update the commons-io models using the *old* Model generator script.
- Add a missing bi-directional import of `FragmentInjection.qll` in `ExternalFlow.qll` before doing any conversion (this might deliver a performance improvement in itself).

### Difference in the number of produced models
The pre-requisites above are made on the first two commits. It is worth noting that from the second execution of the Models as Data framework coverage report
- The number of summaries is increased by 4, which are the summary models that the re-generation of the commons-io introduced.
- The number of sinks is increased by 33, which are caused by the bi-directional import of FragmentInjection (these were not counted before due to the missing import).
- The last execution of the framework coverage report yields the same numeric difference in the number of produced models. That is, we can within reason assume that no models has disappeared as a part of the conversion.

### Other notes
 - The PR can be reviewed on a commit pr commit basis. 
 - Some MaD models comments have been moved manually on the commit that deletes all the manual models, but not everything. Furthermore, some "new" models have been included here (for rebases required after the deletion of the manual inlined models)

### Missing parts for follow up
- Improve the test generator script and queries to emit data extension files instead of inlining the summaries in the unit test Code QL file.
- Convert Kotlin standard library models to data extensions.
- Deprecate (and later delete) `[Summary|Sink|Source]ModelCsv` classes and cleanup related predicates.
- Cleanup `ExternalFlow.qll` imports.

### Performance

DCA reports a significant performance improvement of around 6% (for the projects where the workflow doesn't fail).
image


java/ql/lib/ext/generated/org.apache.commons.io.model.yml
```diff
@@ -0,0 +1,1433 @@
+ 
+# THIS FILE IS AN AUTO-GENERATED MODELS AS DATA FILE. DO NOT EDIT.
+# Definitions of taint steps in the org.apache.commons.io framework.
```
aschackmull, line 3: Not entirely accurate - there are also e.g. sinks.

aschackmull (approved):
Minor comments, otherwise LGTM.

java/ql/lib/ext/generated/org.apache.commons.io.model.yml
michaelnebel: Yes, you are right. I will fix this in one of the follow ups.

PR #11244, opened 2022-11-14
Python: support grouped exceptions

yoff:
ql part of https://github.com/github/semmle-code/pull/44008


python/ql/test/experimental/dataflow/exceptions/test.py
```diff
@@ -0,0 +1,46 @@
+# This should cover all the syntactical constructs that we hope to support.
+# Headings refer to https://docs.python.org/3/reference/expressions.html,
+# and are selected whenever they incur dataflow.
+# Intended sources should be the variable `SOURCE` and intended sinks should be
+# arguments to the function `SINK` (see python/ql/test/experimental/dataflow/testConfig.qll).
+#
+# Functions whose name ends with "_with_local_flow" will also be tested for local flow.
+#
+# All functions starting with "test_" should run and execute `print("OK")` exactly once.
+# This can be checked by running validTest.py.
```
RasmusWL, line 10: Some of this text is not valid, and should be removed. I would probably delete all of it, but I think it's also acceptable to have the last two lines... up to you :shrug:

python/ql/test/experimental/dataflow/exceptions/test_group.py
```diff
@@ -0,0 +1,46 @@
+# This should cover all the syntactical constructs that we hope to support.
+# Headings refer to https://docs.python.org/3/reference/expressions.html,
+# and are selected whenever they incur dataflow.
+# Intended sources should be the variable `SOURCE` and intended sinks should be
+# arguments to the function `SINK` (see python/ql/test/experimental/dataflow/testConfig.qll).
+#
+# Functions whose name ends with "_with_local_flow" will also be tested for local flow.
+#
+# All functions starting with "test_" should run and execute `print("OK")` exactly once.
+# This can be checked by running validTest.py.
```
RasmusWL, line 10: Some of this text is not valid, and should be removed. I would probably delete all of it, but I think it's also acceptable to have the last two lines... up to you :shrug:

python/ql/test/experimental/dataflow/exceptions/test_group.py
```diff
@@ -0,0 +1,46 @@
...
+    else:
+        print("OK")
+
+def test_as_binding():
+    try:
+        e_with_source = Exception()
+        e_with_source.a = SOURCE
+        raise e_with_source
+    except* Exception as e:
+        SINK(e.a) # $ MISSING: flow
```
RasmusWL, line 46: Reading over https://peps.python.org/pep-0654/ it seems like we are not currently testing the grouping feature. Can you add a test that uses nested groups like the example below? (

```
>>> eg = ExceptionGroup(
...     "one",
...     [
...         TypeError(1),
...         ExceptionGroup(
...             "two",
...              [TypeError(2), ValueError(3)]
...         ),
...         ExceptionGroup(
...              "three",
...               [OSError(4)]
...         )
...     ]
... )
```

python/ql/test/experimental/dataflow/validTest.py
```diff
@@ -51,6 +51,14 @@ def check_tests_valid(testFile):
                 check_async_test_function(item)
 
 
+def check_tests_valid_after_version(testFile, version):
+
+    if sys.version_info[:2] >= version:
+        print("INFO: Will run tests in", testFile, "since we're running Python", version, "or newer")
+        check_tests_valid(testFile)
+    else:
+        print("WARN: Will not run tests in", testFile, "since we're running Python", sys.version_info[:2], "and need", version, "or newer")
```
RasmusWL, line 60: Nice :muscle:

python/ql/lib/semmle/python/types/Exceptions.qll
```diff
@@ -439,6 +447,29 @@ class ExceptFlowNode extends ControlFlowNode {
...
+   * `eg` in `except* eg as e:`
+   */
+  ControlFlowNode getType() {
+    this.getBasicBlock().dominates(result.getBasicBlock()) and
+    result = this.getNode().(ExceptGroupStmt).getType().getAFlowNode()
+  }
+
+  /**
+   * Gets the name assigned to the handled exception, if any.
+   * `e` in `except* eg as e:`
```
RasmusWL, line 465: NIT: same as above.

```suggestion
   * `e` in `except* ExceptionType as e:`
```

python/ql/lib/semmle/python/Stmts.qll
```diff
@@ -364,10 +387,13 @@ class Try extends Try_ {
     result = this.getAnOrelse()
   }
 
-  override ExceptStmt getHandler(int i) { result = Try_.super.getHandler(i) }
+  override Stmt getHandler(int i) { result = Try_.super.getHandler(i) }
+
```
RasmusWL, line 391: QL4QL suggests that you can also remove this override.

```suggestion
```

RasmusWL (dismissed):
Overall looks good, although I have some things that I would like changes

python/ql/src/Exceptions/UnguardedNextInGenerator.ql
```diff
@@ -52,7 +52,7 @@ predicate iter_not_exhausted(EssaVariable iterator) {
 predicate stop_iteration_handled(CallNode call) {
   exists(Try t |
     t.containsInScope(call.getNode()) and
-    t.getAHandler().getType() = stopIteration().getAValueReachableFromSource().asExpr()
+    t.getANormalHandler().getType() = stopIteration().getAValueReachableFromSource().asExpr()
```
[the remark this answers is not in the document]
yoff: Yeah, it does seem the only reasonable way to do it, but it still annoys me that we get a user facing breakage here..

python/ql/lib/semmle/python/essa/Essa.qll
```diff
@@ -511,14 +511,18 @@ class AssignmentDefinition extends EssaNodeDefinition {
...
     SsaSource::exception_capture(this.getSourceVariable(), this.getDefiningNode())
   }
 
+  /**
+   * Gets the type handled by this exception handler
+   * `ExceptionType` in `except ExceptionType as ex:`.
+   */
   ControlFlowNode getType() {
-    exists(ExceptFlowNode ex |
+    exists(ExceptGroupFlowNode ex |
```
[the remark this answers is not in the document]
yoff: Huh..no.

python/ql/lib/semmle/python/Stmts.qll
```diff
@@ -364,10 +387,10 @@ class Try extends Try_ {
     result = this.getAnOrelse()
   }
 
-  override ExceptStmt getHandler(int i) { result = Try_.super.getHandler(i) }
+  override Stmt getHandler(int i) { result = Try_.super.getHandler(i) }
 
   /** Gets an exception handler of this try statement. */
-  override ExceptStmt getAHandler() { result = Try_.super.getAHandler() }
+  override Stmt getAHandler() { result = Try_.super.getAHandler() }
```
tausbn: It's probably not too serious, but strictly speaking both this and the change on line 390 are breaking changes, as they generalise the result type (and so code that calls a method that's available on `ExceptStmt` but not `Stmt` now needs an inline cast.

tausbn (changes requested):
(NB: I reviewed commit-by-commit, so some of my comments may have been addressed already.)

Nice work! A few high-level notes:

- I wonder if `except* A as b` should become a `DefinitionNode` as well. We _are_ assigning a value to `b`, but it's not really `A` that gets assigned this way, just something with the type of `A`.
- Regarding breaking the API: Since `ExceptGroupStmt` and `ExceptStmt` expose exactly the same methods, perhaps we could create a common superclass of these two classes, and use _that_ as the result type for `getHandler`? That way, we're not breaking the API.

yoff:
> * I wonder if `except* A as b` should become a `DefinitionNode` as well. We _are_ assigning a value to `b`, but it's not really `A` that gets assigned this way, just something with the type of `A`.

Indeed, `DefinitionNode` has a `getValue` member predicate that we cannot really implement here.

> * Regarding breaking the API: Since `ExceptGroupStmt` and `ExceptStmt` expose exactly the same methods, perhaps we could create a common superclass of these two classes, and use _that_ as the result type for `getHandler`? That way, we're not breaking the API.

I did consider this, but the we also change the existing `ExceptStmt`. Still, it might be worth it. Then there is the question of whether it should expose a `handlesGroups` member predicate, so people can write
```codeql
and try.getAHandler().handlesGroups()
```
instead of
```codeql
and try.getAHandler() instanceof `ExceptGroupStmt`
```
In any case, I will have a go at not breaking normal usage, it does annoy me..

python/ql/test/experimental/dataflow/exceptions/test_group.py
yoff: Sheesh, my one test was not even valid. Thanks for pointing me in this direction.

python/ql/src/Exceptions/UnguardedNextInGenerator.ql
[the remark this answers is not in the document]
yoff: I am trying the common super class route.

python/ql/lib/semmle/python/Stmts.qll
yoff: I am trying the common superclass idea that you suggested.

python/ql/lib/semmle/python/Stmts.qll
```diff
@@ -143,12 +143,46 @@ class Exec extends Exec_ {
...
+ * class is `ExceptionHandler`.
+ */
+abstract private class ExceptionHandlerImpl extends Stmt {
+  /** Gets the name of this except group block. */
+  abstract Expr getName();
+
+  /** Gets the type of this except group block. */
+  abstract Expr getType();
+}
+
```
RasmusWL, line 161: When we already know of all the things that should extend this abstract class, I think it would be better to just list them explicitly, and not have an abstract class at all -- at least that's the advice we've written up here: inhttps://github.com/github/codeql/blob/main/docs/ql-design-patterns.md#abstract-classes-as-open-or-closed-unions

```suggestion
```

I guess the `getName`/`getType` can be defined as abstract, such that they **must** be overriden in the subclasses.

python/ql/lib/change-notes/2022-11-14-grouped-exceptions-breaking.md
```diff
@@ -0,0 +1,4 @@
+---
+category: breaking
+---
+* The result of `Try::getAHandler` and `Try.getHandler(<index>)` is no longer of type `ExceptStmt`, as handlers may also be `ExceptGroupStmt`s (After Python 3.11 introduced PEP 654). Instead, it is of the new type `ExceptionHandler` of which `ExceptStmt` and `ExceptGroupStmt` are subtypes. To support selecting only one type of handler, `try.getANormalHandler` and `try.getAGroupHandler` have been added. Existing uses of `Try::getAHandler` for which it is important to select only normal handlers, will need to be updated to `try.getANormalHandler`.
```
RasmusWL, line 4: NIT: `Try::getAHandler` refers to a predicate `getAHandler` of a module named `Try`

(also used consistent casing of Try)

```suggestion
* The result of `Try.getAHandler` and `Try.getHandler(<index>)` is no longer of type `ExceptStmt`, as handlers may also be `ExceptGroupStmt`s (After Python 3.11 introduced PEP 654). Instead, it is of the new type `ExceptionHandler` of which `ExceptStmt` and `ExceptGroupStmt` are subtypes. To support selecting only one type of handler, `Try.getANormalHandler` and `Try.getAGroupHandler` have been added. Existing uses of `Try.getAHandler` for which it is important to select only normal handlers, will need to be updated to `Try.getANormalHandler`.
```

python/ql/lib/change-notes/2022-11-14-grouped-exceptions-breaking.md
```diff
@@ -0,0 +1,4 @@
+---
+category: breaking
```
RasmusWL, line 2: With the introduction of `ExceptionHandler`, I guess this change is no longer breaking -- from my point of view `fix` would be sufficient

RasmusWL (commented):
One thing about how ExceptionHandler is implemented, otherwise just some NIT

tausbn (approved):
Apart from the issues that rasmuswl raised, I am happy with the current rewrite. 👍

python/ql/lib/change-notes/2022-11-14-grouped-exceptions-breaking.md
yoff: I think you corrected this before, but it got lost in a bad merge. Thanks for catching it again!

PR #11245, opened 2022-11-14
Ruby: use the shared regex pack

erik-krogh:

[Evaluation looks fine](https://github.com/github/codeql-dca-main/tree/data/erik-krogh/rb-redosMod__nightly__code-scanning/reports).


ruby/ql/lib/codeql/ruby/security/BadTagFilterQuery.qll
```diff
@@ -2,155 +2,7 @@
...
-    msg = "This regular expression does not match script end tags like </script >."
-    or
-    not regexp.matches("<script src=\"foo\">foo</script foo=\"bar\">") and
-    msg = "This regular expression does not match script end tags like </script foo=\"bar\">."
-    or
-    not regexp.matches("<script src=\"foo\">foo</script\t\n bar>") and
-    msg = "This regular expression does not match script end tags like </script\\t\\n bar>."
-  )
-}
+private import codeql.ruby.regexp.RegExpTreeView::RegexTreeView as TreeView
```
aibaars, line 5: I don't like the repetitive `RegExpTreeView::RegexTreeView` . Is it really needed to have a  `RegexTreeView` module in the `RegexTreeView` module?

aibaars (commented):
This looks good to me, thanks for splitting the changes in easy to review commits!

I don't really like the double `import Impl` statements. Is there a way to avoid exporting the same things twice under different names?
```
// exporting as RegexTreeView, and in the top-level scope.
import Impl as RegexTreeView
import Impl
```

erik-krogh:
> This looks good to me, thanks for splitting the changes in easy to review commits!
> I don't really like the double import Impl statements. Is there a way to avoid exporting the same things twice under different names?

I don't think there is another way, if we want to keep backwards compatibility.  
My reasoning is:   
- A file cannot implement a module signature, only a module can do that. 
  So I need the `RegexTreeView` module in order to have a module that implements the signature.  
- If I called the module `RegexTreeView` instead of `Impl`, then I would keep an ambiguous import when I did `import RegexTreeView` (because it can both reference the file and the module.).  
 So I instead name the module `Impl`, and just re-export it as `RegexTreeView`, because I don't want the exported name to be `Impl` (or is that OK?).  
- Previously the `RegExpTreeView` file exported all the regex classes directly, so I need the `import Impl` in order to keep backwards compatibility.

aibaars:
> > This looks good to me, thanks for splitting the changes in easy to review commits!
> > I don't really like the double import Impl statements. Is there a way to avoid exporting the same things twice under different names?
> 
> I don't think there is another way, if we want to keep backwards compatibility. My reasoning is:
> 
> * A file cannot implement a module signature, only a module can do that.
>   So I need the `RegexTreeView` module in order to have a module that implements the signature.
> * If I called the module `RegexTreeView` instead of `Impl`, then I would keep an ambiguous import when I did `import RegexTreeView` (because it can both reference the file and the module.).
>   So I instead name the module `Impl`, and just re-export it as `RegexTreeView`, because I don't want the exported name to be `Impl` (or is that OK?).
> * Previously the `RegExpTreeView` file exported all the regex classes directly, so I need the `import Impl` in order to keep backwards compatibility.

Yeah, I was afraid of that. Thanks for explaining. Sounds like we don't have much choice.

ruby/ql/lib/codeql/ruby/security/BadTagFilterQuery.qll
erik-krogh: Yes, as explained below I need a reference to a module instead of a file.  

I could perhaps just do `private import codeql.ruby.regexp.RegExpTreeView`, but this was the thing I ended up with.

PR #11246, opened 2022-11-14
Java: use the shared regex pack

erik-krogh:

[Evaluation looks OK](https://github.com/github/codeql-dca-main/tree/data/erik-krogh/java-redosMod__nightly__code-scanning/reports).  
(I tried to replicate the biggest slowdown locally, but didn't get anything).


aschackmull:
> I'm pretty sure the QLDoc check failure is spurious, but I don't know that happens.

Yeah, that looks weird. Could we perhaps just delete `SuperlinearBackTracking.qll`? It's just a couple of imports now, so it doesn't look like it serves any purpose.

aschackmull:
Ah, wait, I see the problem. It's in the shared file - the qldoc is missing an `*` so it's just a regular comment instead of proper qldoc.

erik-krogh:
> Ah, wait, I see the problem. It's in the shared file - the qldoc is missing an `*` so it's just a regular comment instead of proper qldoc.

:doh:  
I've pushed a fix that add the missing `*`.

aschackmull:
Looks like the shared regex pack isn't triggering CI. I've retriggered the two failing jobs.

aschackmull (approved):

PR #11248, opened 2022-11-14
JS: use the shared regex pack

erik-krogh:

[An evaluation looks good](https://github.com/github/codeql-dca-main/blob/data/erik-krogh/js-redosMod__default__code-scanning__3/reports/).  
The extra result is just another instance in the same regex, but I'm not sure why that happened (I haven't been able to make a nice test-case).  

I'm pretty sure the QLDoc check failure is spurious, but I don't know that happens.
I've tried to add all the QLDoc I could in those files, and I've even tried to add extra modules into the file (just to have something to attach a QLDoc too), but it keeps failing.


javascript/ql/lib/semmle/javascript/security/BadTagFilterQuery.qll
```diff
@@ -2,155 +2,7 @@
...
-    not regexp.matches("<script src=\"foo\">foo</script foo=\"bar\">") and
-    msg = "This regular expression does not match script end tags like </script foo=\"bar\">."
-    or
-    not regexp.matches("<script src=\"foo\">foo</script\t\n bar>") and
-    msg = "This regular expression does not match script end tags like </script\\t\\n bar>."
-  )
-}
+private import regexp.RegexTreeView::RegexTreeView as TreeView
+// BadTagFilterQuery should be used directly from the shared pack, and not from this file.
+deprecated import codeql.regex.nfa.BadTagFilterQuery::Make<TreeView> as Dep
```
asgerf, line 7: What does `deprecated import` do exactly? We don't get a deprecation warning for referencing `Dep` on the next line? Also shouldn't the import be private, so `Dep` is not exposed?

javascript/ql/lib/semmle/javascript/security/regexp/RegexTreeView.qll
```diff
@@ -0,0 +1,73 @@
+/**
+ * Provides JavaScript-specific definitions for use in the NfaUtils module.
+ */
+
+private import codeql.regex.nfa.NfaUtils as NfaUtils
+private import codeql.regex.RegexTreeView
+
+/** An implementation that parses a regular expression into a tree of `RegExpTerm`s. */
+module RegexTreeView implements RegexTreeViewSig {
```
asgerf, line 9: The lack of consistency between `Regex` and `RegExp` is a bit unfortunate. I understand this differs between languages, but at least in JS-specific code, could we keep it to `RegExp`?

javascript/ql/lib/semmle/javascript/security/BadTagFilterQuery.qll
erik-krogh: It's only possible to use `deprecated` in an import if you also add a `as Y` in the end.  
So it's essentially a shorthand for `import X as Tmp; deprecated module Y = Tmp;`.  

If the language supported to do `deprecated import codeql.regex.nfa.BadTagFilterQuery::Make<TreeView>` directly, then I would have done that.  

> Also shouldn't the import be private, so Dep is not exposed? 

Yes 👍  
I'll fix that.

asgerf (approved):

PR #11251, opened 2022-11-14
Kotlin: Add total number of diagnostics to telemetry

igfoo:
This isn't as simple as counting them, as `CODEQL_EXTRACTOR_KOTLIN_DIAGNOSTIC_LIMIT` means that they don't all get written. Also discovered that we weren't writing the message about having hit the limit to the database, so that's now done too.

### DIFF
--- a/java/ql/src/Telemetry/ExtractorInformation.ql
+++ b/java/ql/src/Telemetry/ExtractorInformation.ql
@@ -53,6 +53,34 @@ predicate extractorDiagnostics(string key, int value) {
   )
 }
 
+/*
+ * Just counting the diagnostics doesn't give the full picture, as
+ * CODEQL_EXTRACTOR_KOTLIN_DIAGNOSTIC_LIMIT means that some diagnostics
+ * will be suppressed. In that case, we need to look for the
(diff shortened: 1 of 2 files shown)


java/ql/src/Telemetry/ExtractorInformation.ql
```diff
@@ -53,6 +53,34 @@ predicate extractorDiagnostics(string key, int value) {
...
+predicate extractorTotalDiagnostics(string key, int value) {
+  exists(string extractor |
+    key = "Total number of diagnostics from " + extractor and
+    value =
+      strictcount(Diagnostic d | d.getGeneratedBy() = extractor) +
+        sum(Diagnostic d |
+          d.getGeneratedBy() = extractor
+        |
+          d.getMessage()
+                  .regexpCapture("Total of ([0-9]+) diagnostics \\(reached limit of ([0-9]+)\\).*",
```
smowton, line 73: Lift this into a utility predicate to save both repeating the regex and weird autoformatting? Otherwise lgtm

smowton (approved):

PR #11252, opened 2022-11-14
Ruby: add flow summary for Enumerable#index_by

nickrolfe:


ruby/ql/lib/codeql/ruby/frameworks/ActiveSupport.qll
```diff
@@ -284,7 +284,17 @@ module ActiveSupport {
...
         }
       }
-      // TODO: index_by, index_with, pick, pluck (they require Hash dataflow)
+
+      private class IndexBySummary extends SimpleSummarizedCallable {
+        IndexBySummary() { this = "index_by" }
+
+        override predicate propagatesFlowExt(string input, string output, boolean preservesValue) {
+          input = "Argument[self].Element[any]" and
+          output = ["Argument[block].Parameter[0]", "ReturnValue.Element[?]"] and
```
erik-krogh, line 293: Is there a difference between `Element[any]` and `Element[?]`?

nickrolfe: `any` means any known or unknown index, while `?` means an unknown index.

hvitved: Note that `Element[any]` in an output spec means the same as `Element[?]`.

hvitved: One would never be interested to have an output spec that writes to _any_ content, as that would result in a huge fan-out. The reason why `Element[any]` in an output is still allowed, is because we need to support writes to `Element[any]` in "reverse stores" as well.

Assume we have a flow summary
```
input = Argument[0].Element[any]
output = ReturnValue
```
for a method `get`. Then when we have a call like
```rb
get(x).foo = taint
```
the `get(x)` needs to be translated into a reverse store, such that data is stored inside field `foo`, inside an element of `[post update] x`. And here we interpret the store into `Element[any]` as a store into the unknown element content (`Element[?]`); again to avoid the fan-out.

hvitved: I would absolutely also use `Element[?]` instead of `Element[any]` in output specs of summaries (we should update the summary that you mention); I was just mentioning, that `Element[any]` would in fact mean the same thing.

PR #11253, opened 2022-11-14
Dynamic: Merge package and type columns

asgerf:
This merges the `package;type` columns into a single `type` column. This would be a breaking change if we do it later, so I'd prefer to make the change ASAP.

Concretely, for each language this means:
- In JS, `package;type` is now written as `package.type`, or just `package` if `type` was the empty string.
  - If the `package` contains a `.` character, it must be surrounded with single-quotes: `'package'.type`. The single quote style reflects how the TypeScript language server renders type names.
- In Python, `python;type` is now written as `package.type` or just `package` if `type` was the empty string.
- In Ruby, `package;type` is now just `type` as the `package` had no effect.
  - Previously the `package` was the name of a gem, but this information wasn't used for anything. When and if we care about this information in the future, we can store the mapping between gems and constants in a separate table.

### Additional changes in Ruby
Moreover, the built-in interpretation of Ruby's `type` column has been changed so that:
- The type `A::B` refers to an instance of `A::B` (or more specifically, an instance of a class that has `A::B` on its ancestor chain).
- The type `A::B!` refers to the `A::B` class/module itself.

These types are automatically populated based on language semantics, so the following trivial type definitions can be omitted from the models:
- `A::B!;;Member[A].Member[B]` (i.e. the expression `A::B` has type `A::B!`)
- `A::B;A::B!;Method[new].ReturnValue` (i.e. an expression like `A::B.new` has type `A::B`)

Additionally, `A::B` is intended to match `self` on instance methods on subclasses of `A::B` (but some API graph changes need to be merged before this works). For example, `ActionController::Base;Method[params].ReturnValue` would then match calls to `params` inside subclasses of `ActionController::Base`.

It also removes some boiler plate and standardises the name of a type, so models can contribute to it without having to look up what it's called in other models.


javascript/ql/lib/semmle/javascript/frameworks/SQL.qll
```diff
@@ -368,10 +368,10 @@ private module SpannerCsv {
       // package; type; path; kind
```
erik-krogh, line 368: This `// package; type; path; kind` comment should be updated (and similar comments, if there are any).

asgerf:
@hmac The original intent behind the ‘package’ column didn’t work for Ruby or Python, since neither refer to the package name in the source code. So it ended up meaning very different things in each language. For Ruby the package column is even ignored. One of the two columns was often left empty. Type-definition rows are much easier to read now that we have 2 fewer columns in there. It ended up being a really awkward split, and we're approaching a point after which we won't be able to make this change, so I'd like to get it in while we can.

javascript/ql/lib/semmle/javascript/frameworks/data/internal/ApiGraphModelsSpecific.qll
```diff
@@ -31,6 +31,15 @@ import semmle.javascript.frameworks.data.internal.AccessPathSyntax as AccessPath
...
 
+bindingset[rawType]
+predicate parseTypeString(string rawType, string package, string qualifiedName) {
+  exists(string regexp |
+    regexp = "('[^']+'|[^.]+)(.*)" and
+    package = rawType.regexpCapture(regexp, 1).regexpReplaceAll("^'|'$", "") and
+    qualifiedName = rawType.regexpCapture(regexp, 2).regexpReplaceAll("^\\.", "")
+  )
+}
+
```
erik-krogh, line 42: I understand this, but I basically dream in regular expressions at this point. 
A documentation string could be nice.  

```suggestion
/**
 * Holds if `rawType` represents the JavaScript npm `package` with type `qualifiedName`.  
 * The `qualifiedName` is everything after the first ".".   
 * So e.g. `rawType = "foo.bar"` will parse to `package = "foo" and qualifiedName = "bar"`.  
 * `package´ can be enclosed in single-quotes if the package name includes dots, 
 * e.g. `'my.package'.type` parses to `package = "my.package" and qualifiedName = "type"`.  
 */ 
bindingset[rawType]
predicate parseTypeString(string rawType, string package, string qualifiedName) {
  exists(string regexp |
    regexp = "('[^']+'|[^.]+)(.*)" and
    package = rawType.regexpCapture(regexp, 1).regexpReplaceAll("^'|'$", "") and
    qualifiedName = rawType.regexpCapture(regexp, 2).regexpReplaceAll("^\\.", "")
  )
}

```

erik-krogh: Also, is this syntax documented anywhere?

RasmusWL (commented):
I think this PR contains two changes, and would like to review them seperately:

> This merges the package;type columns into a single type column.

Nice, Python :+1:

> Additional changes in Ruby

I would like to learn more about this change before approving.

I'm curious what motivated this change? -- was it to more easily map to the optional type annotations for Ruby?

I'm also considering whether this is something we should do for Python :thinking:

RasmusWL (approved):
Thanks for talking things through, all looks good to me :+1:

javascript/ql/lib/semmle/javascript/frameworks/data/internal/ApiGraphModelsSpecific.qll
asgerf: > Also, is this syntax documented anywhere?

Not yet, but I'll be working on documentation for the MaD format

erik-krogh (dismissed):

hvitved:
> * The type `A::B!` refers to the `A::B` class/module itself.

And not an exact instance of the class itself?

ruby/ql/lib/codeql/ruby/frameworks/data/internal/ApiGraphModelsSpecific.qll
```diff
@@ -66,9 +109,27 @@ API::Node getExtraNodeFromPath(string package, string type, AccessPath path, int
...
-/** Gets a Ruby-specific interpretation of the `(package, type)` tuple. */
-API::Node getExtraNodeFromType(string package, string type) {
-  isRelevantFullPath(package, type, _) and // Allow any package name, see `isPackageUsed`.
+/** Gets a Ruby-specific interpretation of the given `type`. */
+API::Node getExtraNodeFromType(string type) {
+  exists(string consts, string suffix, DataFlow::ConstRef constRef |
+    parseRelevantType(type, consts, suffix) and
+    constRef = getConstantFromConstPath(consts)
+  |
+    suffix = "!" and
```
hvitved, line 118: I wonder if we could use some different syntax, as `A::B!` would make me think _exatly_ type `A::B` (as opposed to `A::B` or any descendant thereof).

asgerf:
> > The type `A::B!` refers to the `A::B` class/module itself.
>
> And not an exact instance of the class itself?

No, it really just refers to the class/module itself. Is there another syntax you'd prefer for this?

ruby/ql/test/library-tests/dataflow/local/TaintStep.expected
```diff
@@ -1,9 +1,6 @@
 | file://:0:0:0:0 | [summary] read: argument position 0.any element in Hash[] | file://:0:0:0:0 | [summary] read: argument position 0.any element.element 1 or unknown in Hash[] |
-| file://:0:0:0:0 | parameter any of ;Pathname;Method[join] | file://:0:0:0:0 | [summary] to write: return (return) in ;Pathname;Method[join] |
```
hvitved, line 2: Why did all these lines disappear?

ruby/ql/test/library-tests/dataflow/summaries/Summaries.ql
```diff
@@ -101,23 +101,21 @@ private class TypeFromModel extends ModelInput::TypeModelCsv {
   override predicate row(string row) {
     row =
       [
-        "test;FooOrBar;;;Member[Foo].Instance", //
-        "test;FooOrBar;;;Member[Bar].Instance", //
-        "test;FooOrBar;test;FooOrBar;Method[next].ReturnValue",
+        "~FooOrBar;Foo;", //
```
hvitved, line 104: Could you remind me what the `~` does?

hvitved (commented):
LGTM. Only a couple of questions.

ruby/ql/lib/codeql/ruby/frameworks/data/internal/ApiGraphModelsSpecific.qll
asgerf: Some other alternatives I've thought about
- `A::B.module`
- `A::B.class`
- `A::B.self`
- `A::B.instance` for instances and `A::B` for the class/module itself

For completeness there's also another alternative, which is to merge the `type` and `path` columns as well. So we'd have tables of form:
- `sourceModel(path, kind)`
- `sinkModel(path, kind)`
- `summaryModel(path, input, output, kind)`
- `typeModel(path1, path2)`

The "types" would then be written out as access paths, exactly as before
- `Member[A].Member[B]` for the `A::B` class/module itself
- `Member[A].Member[B].Instance` for instances of `A::B`

This means type model rows no longer just contribute to a named type, but instead contribute to an access path. I believe this is what @max-schaefer called an alias edge in some earlier discussions about this. If there isn't an intuitive syntax for the types you typically care about in Ruby, it might be worth trying to switch to that system instead.

ruby/ql/test/library-tests/dataflow/local/TaintStep.expected
asgerf: Sorry I should have mentioned that. This test case doesn't reference `Pathname` at all, so the summaries with `Pathname` in the type column get pruned.

Pruning was previously based entirely on the package column, and was disabled for Ruby since the package column couldn't be used for anything.

ruby/ql/test/library-tests/dataflow/summaries/Summaries.ql
asgerf: It's the convention for ensuring that we don't clash with a built-in type. The `~` character itself doesn't do anything special. If we simply called the type `ForOrBar` it would be also interpreted to mean instances of `FooOrBar`.

ruby/ql/lib/codeql/ruby/frameworks/data/internal/ApiGraphModelsSpecific.qll
asgerf: `A::B.static` maybe?

hvitved: If we want to use the same syntax across languages, then perhaps `A::B.static` is better.

hvitved: I think I perhaps favor the

> `A::B.instance` for instances and `A::B` for the class/module itself

suggestion, though that probably means having to write `.instance` a lot more often than having to write `.static` as per the suggestion above.

RasmusWL: Following along on the sidelines, two thoughts:

1) I thought it sounded like a fine idea to merge `type` and `path`, but how will we allow "fake" types used for aliasing, such as `~foo.Bar` (what we currently write as `foo;~Bar`)? (I know it's not going to happen right now, but just wanted to point out another potential problem)

2) I didn't personally have anything against using `!`. If we need to go with one of your alternatives, I would personally prefer `.class`. (I'm assuming that we'll do the same change in Python at some point, so that's why I dare participate in the coloring of this bikeshed :wink:) -- with my Python eyes:
    - `A::B.self` could easily be a reference to self parameters on methods of the `A::B` class
    - `A::B.module` could easily be a reference to the module containing `A::B`
    - I think `A::B.instance` is a natural idea, but wouldn't 99% of our type rows would be filled with `.instance`? I'm guessing it would, so don't think it's worth it.
    - `A::B.static` doesn't have any special meaning

hvitved: Or we we want to keep it closer to Ruby, `A::B.new` instead of `A::B.instance`.

asgerf: We could go with both `A::B.static` and `A::B.instance` to eliminate the ambiguity.

hvitved: > We could go with both `A::B.static` and `A::B.instance` to eliminate the ambiguity.

I like that compromise, for now.

asgerf: After implementing `.static` and `.instance` and seeing [what it looks like in practice](https://github.com/asgerf/codeql/commit/cb6d4c7a80e5ada0d0010851176e5bf7781df1fb) I don't really want to go ahead with that. I'd prefer to just stick with `!`.

> I thought it sounded like a fine idea to merge type and path, but how will we allow "fake" types used for aliasing, such as ~foo.Bar (what we currently write as foo;~Bar)? (I know it's not going to happen right now, but just wanted to point out another potential problem)

There would be a `Type[x]` token to support named types still, both for model-authored types, and type annotations whose meaning can't be known without seeing its definition (which is the case for TypeScript).

hvitved (approved):

PR #11254, opened 2022-11-14
C++: Fix spurious reference flow

MathiasVP:
On the use-use-flow feature-branch we were getting spurious flow like:
```cpp
void local_references(int &source) {
  source = 0;
  sink(source); // clean, but was marked as tainted
}
```
when the `source` dataflow node was defined as `source.asParameter().hasName("source")`. This turns out to be because `source.asParameter()` defaults to `source.asParameter(0)`, which means that the value of `source` was marked as the source instead of the value of `*source` (if we pretend that the source is of a pointer type instead of a reference type).

Ideally, the fix in 5e9e7da7a4e36a9603c5a7e1d769b639908bd2e3 should be enough to fix this. However, the [first commit](7408931565ecd6b817cd738626eeafd1c936dea0) is a necessary precondition for this. This first commit fixes two things:
- It removes the ad-hoc dataflow we added between reference addresses and their dereferences. That's pretty simple to remove ✔️.
- The second issue is a subtle bug in `adjacentDefRead`: Previously, when we were flowing from one def/use to an adjacent use we did the following:
```ql
defOrUse.asDefOrUse().hasIndexInBlock(bb1, i1, v)
adjacentDefRead(_, bb1, i1, bb2, i2) and
use.asDefOrUse().(UseImpl).hasIndexInBlock(bb2, i2, v)
```
which amounts to saying:
1. A def/use of variable `v` is at position `(bb1, i1)`
2. There's an adjacent-def/use-use relationship between the locations `(bb1, i1)` and `(bb2, i2)`
3. A use of variable `v` is at position `(bb2, i2)`.

However, note that point 2 doesn't talk about which variable is participating in the adjacent-def/use-use relationship. So in our example above, we'd have a use-use relation between `source` (note: _not_ `*source`) in `source = 0` and `source` in `sink(source)`. And since `source` and `*source` are located at the same index in the same block we'd get use-use flow between `*source` in `source = 0` and `*source` in `sink(source)`.

The fix is very simple: instead of ignoring the first column of `adjacentDefRead` we simply check that the variable in question is actually `v`.


cpp/ql/lib/semmle/code/cpp/ir/dataflow/internal/SsaInternals.qll
```diff
@@ -265,10 +265,10 @@ predicate adjacentDefRead(DefOrUse defOrUse1, UseOrPhi use) {
...
     defOrUse1.asDefOrUse().hasIndexInBlock(bb1, i1, v)
   |
-    exists(IRBlock bb2, int i2 |
-      adjacentDefRead(_, pragma[only_bind_into](bb1), pragma[only_bind_into](i1),
-        pragma[only_bind_into](bb2), pragma[only_bind_into](i2))
-    |
+    exists(IRBlock bb2, int i2, Definition def |
+      adjacentDefRead(pragma[only_bind_into](def), pragma[only_bind_into](bb1),
+        pragma[only_bind_into](i1), pragma[only_bind_into](bb2), pragma[only_bind_into](i2)) and
+      def.getSourceVariable() = v and
```
jketema, line 271: Why wasn't this here before?

MathiasVP: I thought it wasn't needed because of the conditions on the `def/use` and the `use` (i.e., they are both attached to the same variable). This was actually pointed out by Tom in the initial PR [here](https://github.com/github/codeql/pull/10190#discussion_r960524653), but I managed to convince him that this was actually okay.

cpp/ql/test/library-tests/dataflow/dataflow-tests/test.cpp
```diff
@@ -423,25 +423,32 @@ void intPointerSourceCaller() {
   intPointerSource(&local);
   sink(local); // $ ast,ir=422:7 ast,ir=423:20
 }
-
+// The IR results for this test _is_ equivalent to the AST ones.
+// The IR annotation is just "ir" because the sink of the unitialized source at
+// 428:7 is value of `local`, but the sink of the source from `intPointerSource`
+// value of `*local` (i.e., the indirection node of `local`). So unlike AST dataflow,
+// each of these two sinks correspond to a unique source, and thus we don't need to
+// attach a location annotation to it.
```
rdmarsh2, line 431: Should we still provide the location annotations to guarantee that if these get conflated we'll have a test failure?

MathiasVP: I think that's a good idea. I _think_ this requires adding another tag though, right (since I don't think we can have tags with optional values)?

MathiasVP:
I just force-pushed away a commit I added by mistake. Sorry about that!

cpp/ql/test/library-tests/dataflow/dataflow-tests/test.cpp
MathiasVP: I would prefer to add this in another PR if that's okay with you, though.

cpp/ql/test/library-tests/dataflow/dataflow-tests/test.cpp
```diff
@@ -423,25 +423,32 @@ void intPointerSourceCaller() {
   intPointerSource(&local);
   sink(local); // $ ast,ir=422:7 ast,ir=423:20
 }
-
+// The IR results for this test _is_ equivalent to the AST ones.
```
jketema, line 426:
```suggestion
// The IR results for this test _are_ equivalent to the AST ones.
```

cpp/ql/test/library-tests/dataflow/dataflow-tests/test.cpp
```diff
@@ -423,25 +423,32 @@ void intPointerSourceCaller() {
...
 }
 
 void intArraySourceCaller() {
   int local;
   intArraySource(&local, 1);
-  sink(local); // $ ast,ir=435:7 ast,ir=436:18
+  sink(local); // $ ast,ir=440:7 ast,ir=441:18
 }
 
+// The IR results for this test _is_ equivalent to the AST ones.
```
jketema, line 445:
```suggestion
// The IR results for this test _are_ equivalent to the AST ones.
```

jketema (commented):
Two typos.

cpp/ql/lib/semmle/code/cpp/ir/dataflow/internal/DataFlowUtil.qll
```diff
@@ -235,7 +235,13 @@ class Node extends TIRDataFlowNode {
   Expr asIndirectArgument() { result = this.asIndirectArgument(_) }
 
   /** Gets the positional parameter corresponding to this node, if any. */
-  Parameter asParameter() { result = this.asParameter(0) }
+  Parameter asParameter() {
+    exists(int indirectionIndex | result = this.asParameter(indirectionIndex) |
+      if result.getUnspecifiedType() instanceof ReferenceType
+      then indirectionIndex = 1
+      else indirectionIndex = 0
+    )
```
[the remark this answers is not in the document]
MathiasVP: I added that code to this PR, and force-pushed it away when I added this commit. So I think we're all good? Or am I misunderstanding the what piece of code you're referring to?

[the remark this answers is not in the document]
jketema: Then we're all good. Just hard for me to keep track of all the PRs.

cpp/ql/test/library-tests/dataflow/dataflow-tests/test.cpp
jketema: I think the `428:7` on l. 428 is incorrect now, because this comment was added.

I do have some trouble wrapping my head around the reasoning here. My expectation for `int local[1]` would be that the array is initialised, but the elements of the array are not, but that's apparently not the way to think about this?

MathiasVP: It looks like the AST library does:
1. Flow from an initialized node to `sink(local)`, _and_ to `sink(*local)`
2. Flow from the elements of the array after leaving `intPointerSource` to `sink(local)`, _and_ to `sink(*local)`.

jketema: > Correct?

Yes.

> It looks like the AST library does:

I was assuming this is due to conflation problems the AST library has. Or does the AST dataflow library indicate that there may be usability reasons for doing this? I can imagine that it might make sense to consider `int local[1]` to be uninitialised for ease writing some of the sources/sinks that involve arrays.

MathiasVP: I agree that `UninitializedNode` should probably be handled differently to work for array variables. I think this should be solved in a separate PR, though.

PR #11257, opened 2022-11-14
C++: Reduce path duplication

MathiasVP:
This fixes some issues uncovered in https://github.com/github/codeql/pull/11254.

### DIFF
--- a/cpp/ql/lib/semmle/code/cpp/ir/dataflow/internal/DataFlowUtil.qll
+++ b/cpp/ql/lib/semmle/code/cpp/ir/dataflow/internal/DataFlowUtil.qll
@@ -752,15 +752,26 @@ private predicate exprNodeShouldBeIndirectOperand(IndirectOperand node, Expr e, 
   not convertedExprMustBeOperand(e)
 }
 
+private predicate indirectExprNodeShouldBeIndirectOperand0(
+  VariableAddressInstruction instr, IndirectOperand node, Expr e
+) {
+  instr = node.getOperand().getDef() and
+  not node instanceof ExprNode and
+  e = instr.getAst().(Expr).getUnconverted()
+}
+
 /** Holds if `node` should be an `IndirectOperand` that maps `node.asIndirectExpr()` to `e`. */
 private predicate indirectExprNodeShouldBeIndirectOperand(IndirectOperand node, Expr e) {
   exists(Instruction instr |
     instr = node.getOperand().getDef() and
     not node instanceof ExprNode
   |
-    e = instr.(VariableAddressInstruction).getAst().(Expr).getFullyConverted()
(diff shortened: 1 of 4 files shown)


cpp/ql/src/Security/CWE/CWE-311/CleartextTransmission.ql
```diff
@@ -261,8 +261,7 @@ class ToEncryptionConfiguration extends TaintTracking2::Configuration {
   ToEncryptionConfiguration() { this = "ToEncryptionConfiguration" }
 
   override predicate isSource(DataFlow::Node source) {
-    any(FromSensitiveConfiguration config).hasFlow(source, _) and
-    isSourceImpl(source)
```
MathiasVP, line 265: Drive-by fix. This condition follows from `any(FromSensitiveConfiguration config).hasFlow(source, _)`.

cpp/ql/lib/semmle/code/cpp/ir/dataflow/internal/DataFlowUtil.qll
```diff
@@ -752,15 +752,26 @@ private predicate exprNodeShouldBeIndirectOperand(IndirectOperand node, Expr e,
...
+  instr = node.getOperand().getDef() and
+  not node instanceof ExprNode and
+  e = instr.getAst().(Expr).getUnconverted()
+}
+
 /** Holds if `node` should be an `IndirectOperand` that maps `node.asIndirectExpr()` to `e`. */
 private predicate indirectExprNodeShouldBeIndirectOperand(IndirectOperand node, Expr e) {
   exists(Instruction instr |
     instr = node.getOperand().getDef() and
     not node instanceof ExprNode
```
jketema, line 767: This now seems duplicated in the `indirectExprNodeShouldBeIndirectOperand0` predicate?

MathiasVP: True. I pulled it in there to reduce the size of `indirectExprNodeShouldBeIndirectOperand0`. Maybe that was a bit of a premature optimization 🤔. We want that condition to apply to all of `indirectExprNodeShouldBeIndirectOperand`, but we can't just have it in this new predicate because then it'll only apply to one side of the disjunct.

jketema: After some out-of-band discussion: we do need this in two places for the second disjunct, as `instr` needs to be restricted, and any other instruction that we capture `not indirectExprNodeShouldBeIndirectOperand0(_, _, e.getUnconverted())` also needs to satisfy it.

jketema (approved):
LGTM

PR #11266, opened 2022-11-15
C#: Update all nuget packages

hvitved:
Update all nuget packages to the latest version.

### DIFF
--- a/csharp/extractor/Semmle.Extraction.CSharp/Entities/UserOperator.cs
+++ b/csharp/extractor/Semmle.Extraction.CSharp/Entities/UserOperator.cs
@@ -162,10 +162,14 @@ namespace Semmle.Extraction.CSharp.Entities
                 case "op_False":
                     operatorName = "false";
                     break;
-                case var @checked when Regex.IsMatch(@checked, "^op_Checked.*"):
-                    operatorName = @checked;
-                    break;
                 default:
+                    var match = Regex.Match(methodName, "^op_Checked(.*)$");
+                    if (match.Success)
+                    {
+                        OperatorSymbol("op_" + match.Groups[1], out var uncheckedName);
+                        operatorName = "checked " + uncheckedName;
+                        break;
+                    }
                     operatorName = methodName;
                     success = false;
                     break;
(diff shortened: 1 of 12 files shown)


csharp/extractor/Semmle.Extraction.CSharp/Entities/UserOperator.cs
```diff
@@ -161,6 +162,9 @@ public static bool OperatorSymbol(string methodName, out string operatorName)
                 case "op_False":
                     operatorName = "false";
                     break;
+                case var @checked when Regex.IsMatch(@checked, "^op_Checked.*"):
+                    operatorName = @checked;
+                    break;
```
tamasvajk, line 167: Why is this change needed? Are there any tests that cover it? `checked +` might be another operator name that we could use. That would match better the unchecked `+`.

hvitved: Indeed; we should use `checked +` instead. Apparently some existing tests exercise this; probably because checked operators exist in library code.

csharp/extractor/Semmle.Extraction.CSharp/Entities/UserOperator.cs
```diff
@@ -166,6 +166,13 @@ public static bool OperatorSymbol(string methodName, out string operatorName)
                     operatorName = @checked;
                     break;
                 default:
+                    var match = Regex.Match(methodName, "^op_Checked(.*)$");
+                    if (match.Success)
+                    {
+                        OperatorSymbol("op_" + match.Groups[1], out var uncheckedName);
+                        operatorName = "checked " + uncheckedName;
+                        break;
+                    }
```
tamasvajk, line 175: Do we need this in the `default` branch? Why wasn't this added in the `var @checked when Regex.IsMatch(@checked, "^op_Checked.*"):` case? Or maybe the `var @checked when ...` could be removed now.

tamasvajk: But isn't the below redundant?
```
case var @checked when Regex.IsMatch(@checked, "^op_Checked.*"):
                    operatorName = @checked;
                    break;
```

hvitved: Ah 🤦 I thought I had removed it; is removed now.

tamasvajk (approved):

PR #11267, opened 2022-11-15
ATM: remove superfluous class in EndpointCharacteristics hierarchy

kaeluka:
In response to my own comment: https://github.com/github/codeql/pull/11174#discussion_r1018036893


henrymercer (approved):
This looks right.  I don't think `NotASinkCharacteristic` should be a subtype of `OtherModeledArgumentCharacteristic`.

kaeluka:
Thank you, Henry! If we ever need `OtherModeledArgumentCharacteristic` we can of course think about re-introducing it at a fitting location in the hierarchy 👍

kaeluka:
@tiferet I sent this PR to allow you to stay focus on what matters. As I said above, I'm not against having `OtherModeledArgumentCharacteristic` — but it wasn't used anywhere, so removing was the obvious way to address the linked hierarchy-comment :)

Issue #9304, opened 2021-09-16
Newly-broken builds


smowton:
Likely a consequence of https://github.com/github/codeql-go/pull/554 there are two new classes of failure: segfaults and a no-source-code-found error. Lists of affected projects posted on Slack.

owen-mc:
This issue was solved by reverting https://github.com/github/codeql-go/pull/554 in https://github.com/github/codeql-go/pull/572. There is a follow-up task to see if the problems can be fixed and this feature can be re-enabled: https://github.com/github/codeql-go-team/issues/282.

PR #11268, opened 2022-11-15
Enable accelerated go-extractor opt-in using 'go list -deps'

grddev:
Resurrect https://github.com/github/codeql-go/pull/554, but behind an environment variable as to avoid the broken builds noted in https://github.com/github/codeql/issues/9304, but still allowing some people to opt in to the much faster approach.

Ideally, the broken builds should be resolved, but that is difficult to accomplish with the limited information in https://github.com/github/codeql/issues/9304


go/extractor/util/util.go
```diff
@@ -54,6 +54,21 @@ func runGoListWithEnv(format string, pkgpath string, additionalEnv []string, fla
...
+	PkgDir string // the directory directly containing source code of this package
+	ModDir string // the module directory containing this package, empty if not a module
+}
+
+// GetPkgInfo fills the package info structure for the specified package path.
+// It passes the `go list` the flags specified by `flags`.
+func GetPkgInfo(pkgpath string, flags ...string) PkgInfo {
+	return PkgInfo{
+		PkgDir: GetModDir(pkgpath, flags...),
+		ModDir: GetPkgDir(pkgpath, flags...),
```
smowton, line 127: This looks suspicious

grddev: Indeed 😅

smowton:
Sorry, forgot about this. The changes aren't totally simple to the point I'd want to just deploy this -- we should do some internal testing to verify that the builds that newly failed with the resurrected PR now work with this one (with the env var unset, so in theory it's supposed to be behaving the same as without this PR)

smowton:
I pushed a commit to autoformat the edited `.go` files

go/extractor/extractor.go
```diff
@@ -174,16 +177,17 @@ func ExtractWithFlags(buildFlags []string, patterns []string) error {
...
 			if err != nil || noExtractRe.MatchString(relDir) {
 				// if the path can't be made relative or matches the noExtract regexp skip it
 				continue
 			}
 
 			extraction.extractPackage(pkg)
 
-			if pkgRoots[pkg.PkgPath] != "" {
-				modPath := filepath.Join(pkgRoots[pkg.PkgPath], "go.mod")
+			if pkgInfo.ModDir != "" {
```
smowton, line 189: Because this doesn't fall back to `PkgDir` as before, we don't extract `go.mod` files in subdirectories when we couldn't work out their `ModDir`. Would it be simplest to just populate `ModDir` with `PkgDir` as the old code did, to avoid needlessly confusing users? It's fair to say if there's a `go.mod` and we couldn't work out the `ModDir` we're probably building the project wrongly anyhow, but we may as well limit confusion on top of that.

grddev: Yeah, this was an oversight on my part. Updated.

smowton (approved):

PR #11269, opened 2022-11-15
Java: Stub generator: Exclude invalid identifiers from generated stubs

joefarebrother:
It appears that compiled `.class` files are able to contain identifiers that are not valid in Java source code. 
For instance, when stubing `okhttp`, I ran into several identifiers beginning with `-deprecated_`.
This PR excludes these identifiers from being stubbed.


tamasvajk (approved):
LGTM

PR #11271, opened 2022-11-15
C++: Reintroduce the AST testing configuration for the smart pointer test too

jketema:
I had previously omitted this, as I had not looked at the smart pointer test yet.


MathiasVP (approved):
LGTM!

PR #11272, opened 2022-11-15
CI: clean up the cache when compiling on main

erik-krogh:
The cache was append-only, so it would keep monotonically increasing in size.  
It started at ~60mb, but it had grown to ~250mb.  

This PR makes sure that only the cache objects needed are preserved.  

The CodeQL CLI accepts a `--compilation-cache` option where it additionally reads cache objects from.  
But the CLI will still populate the relevant cache folders. 
Thereby effectively moving the relevant cache from the directory specified `--compilation-cache` into the relevant folders.  

In a [test run](https://github.com/github/codeql/actions/runs/3480558055/jobs/5820513882) this decreased the cache size from 251mb to 62mb.   
(See the cache read and cache write steps).  

This was inspired by DCA that uses `--compilation-cache` to achieve the same effect.

### DIFF
--- a/.github/workflows/compile-queries.yml
+++ b/.github/workflows/compile-queries.yml
@@ -60,14 +60,10 @@ jobs:
           # Move all the existing cache into another folder, so we only preserve the cache for the current queries.
           mkdir -p ${COMBINED_CACHE_DIR}
           rm */ql/src/.cache/{lock,size}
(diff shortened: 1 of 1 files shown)


erik-krogh:
/cc @aibaars  
Yes, the simpler approach worked 🎉  

New test run: https://github.com/github/codeql/actions/runs/3480558055/jobs/5820513882

aibaars (approved):

PR #11276, opened 2022-11-15
CodeQL: add 'False positive' issue template

aibaars:


.github/ISSUE_TEMPLATE/ql--false-positive.md
```diff
@@ -0,0 +1,36 @@
...
+    return cp.spawn(getShell(), ["-c", command], options) // <- command line injection
+};
+```
+-->
+
+**URL to the alert on GitHub CodeScanning (optional)**
+
+<!--
+1. Open the project on GitHub.com.
+2. Switch to the `Security` tab. For example, https://lgtm.com/projects/g/pallets/click/alerts/.
```
igfoo, line 33: I think that this URL should be updated to a code-scanning URL

aibaars: Good catch. I removed that URL. Such URLs are inaccessible for others, so a real example would just be a dead link for most viewers.

.github/ISSUE_TEMPLATE/ql--false-positive.md
```diff
@@ -0,0 +1,36 @@
+---
+name: False positive
+about: Tell us about an alert that shouldn't be reported
```
turbo, line 3:
```suggestion
about: Report code scanning (CodeQL) alerts that you think should not have been detected (not applicable, not exploitable, etc.)
```

.github/ISSUE_TEMPLATE/ql--false-positive.md
```diff
@@ -0,0 +1,36 @@
+---
+name: False positive
```
turbo, line 2:
```suggestion
name: CodeQL False positive
```

turbo (changes requested):

.github/ISSUE_TEMPLATE/ql--false-positive.md
aibaars: I wondered about having a `CodeQL` prefix in the title, but decided against it because the issue template is for the `github/codeql` repository, so `codeql` is sort of implied. Happy to add the prefix though, if you think it is important here.

.github/ISSUE_TEMPLATE/ql--false-positive.md
aibaars: Shall we replace `code scanning (CodeQL) alerts` with `CodeQL alerts` ? The template is not for arbitrary code scanning alerts; in addition alerts found directly by the CodeQL CLI are also in scope.

turbo: Yep, makes sense

.github/ISSUE_TEMPLATE/ql--false-positive.md
turbo: I wanted to re-emphasize this here because it's the bold text users are likely to read. Sometimes general code scanning users end up here and it helps to highlight that this is about CodeQL again I think

.github/ISSUE_TEMPLATE/ql--false-positive.md
```diff
@@ -0,0 +1,36 @@
+---
+name: CodeQL False positive
+about: Report CodeQL alerts that you think should not have been detected (not applicable, not exploitable, etc.)
+title: False positive
+labels: false-positive
+assignees: ''
```
aeisenberg, line 6: Do we want to assign this to the first responder automatically? @github/codeql-first-responder

adityasharad: I don't think that will work - we can't assign teams to issues, only individuals. If we put the handle in the issue template body, the issue creator would need to be a member of our org to actually notify a team in our org.

aeisenberg: Ah...didn't know that. I am happy with this template, though. @turbo should probably approve since he had some suggestions earlier.

turbo (approved):

PR #11282, opened 2022-11-16
Java: Query for detecting addJavascriptInterface method calls

egregius313:
Added a query to detect invocations of `addJavascriptInterface`, which can lead to cross-site scripting.

### DIFF
--- a/java/ql/src/Security/CWE/CWE-079/AndroidWebViewAddJavascriptInterface.ql
+++ b/java/ql/src/Security/CWE/CWE-079/AndroidWebViewAddJavascriptInterface.ql
@@ -1,7 +1,7 @@
 /**
  * @name Access Java object methods through JavaScript exposure
  * @id java/android-webview-addjavascriptinterface
- * @description Exposing a Javascript interface to a Java object in a WebView can lead to malicious JavaScript controlling the application.
+ * @description Exposing a Java object in a WebView with a JavaScript interface can lead to malicious JavaScript controlling the application.
  * @kind problem
  * @problem.severity warning
  * @security-severity 6.1
--- a/java/ql/src/Security/CWE/CWE-079/AndroidWebViewAddJavascriptInterface.ql
+++ b/java/ql/src/Security/CWE/CWE-079/AndroidWebViewAddJavascriptInterface.ql
@@ -1,4 +1,5 @@
 /**
+ * @name Access Java object methods through JavaScript exposure
  * @id java/android-webview-addjavascriptinterface
  * @description Exposing a Javascript interface to a Java object in a WebView can lead to malicious JavaScript controlling the application.
  * @kind problem
--- a/java/ql/src/Security/CWE/CWE-079/AndroidWebViewAddJavascriptInterface.ql
+++ b/java/ql/src/Security/CWE/CWE-079/AndroidWebViewAddJavascriptInterface.ql
@@ -4,7 +4,7 @@
  * @kind problem
  * @problem.severity warning
  * @security-severity 6.1
- * @precision high
+ * @precision medium
  * @tags security
  *       external/cwe/cwe-079
  */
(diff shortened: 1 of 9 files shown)


java/ql/src/Security/CWE/CWE-079/AndroidWebViewAddJavascriptInterfaceExample.java
```diff
@@ -0,0 +1,11 @@
...
+    @JavascriptInterface
+    public String example() {
+        return "String from Java";
+    }
+}
+
+webview.getSettings().setJavaScriptEnabled(true);
+webview.addJavaScriptInterface(new ExposedObject(), "exposedObject");
+webview.loadData("", "text/html", null);
+webview.loadUrl("javascript:alert(exposedObject.example())");
```
atorralba, line 11: I would like to see a more obvious example of why this could be a security problem. Instead of the trivial "return a String" method, you could show a more interesting case, e.g. one that takes parameters from the JavaScript and runs an OS command or builds a SQL statement with them.

java/ql/src/Security/CWE/CWE-079/AndroidWebViewAddJavascriptInterface.ql
```diff
@@ -0,0 +1,17 @@
+/**
+ * @id java/android-webview-addjavascriptinterface
+ * @description Exposing a Javascript interface to a Java object in a WebView can lead to malicious JavaScript controlling the application.
+ * @kind problem
+ * @problem.severity warning
+ * @security-severity 6.1
+ * @precision high
```
atorralba, line 7: Since this won't always be a security problem, let's get the precision down to `medium`. That way this query won't run by default, i.e. it will be opt-in for users that want to see more results.

java/ql/src/Security/CWE/CWE-079/AndroidWebViewAddJavascriptInterface.ql
```diff
@@ -0,0 +1,17 @@
...
+ * @tags security
+ *       external/cwe/cwe-079
+ */
+
+import java
+import semmle.code.java.frameworks.android.WebView
+
+from MethodAccess ma
+where ma.getMethod() instanceof WebViewAddJavascriptInterfaceMethod
```
atorralba, line 16: I think we could make this query slightly smarter by discarding methods that check the URL origin, see https://github.com/OWASP/owasp-mastg/blob/28116d4ac1b1187893687f416ffc5d0fc6898bd9/Document/0x05h-Testing-Platform-Interaction.md#determining-whether-java-objects-are-exposed-through-webviews-mstg-platform-7.

> If `addJavascriptInterface` is necessary, take the following considerations:
> * Only JavaScript provided with the APK should be allowed to use the bridges, e.g. by verifying the URL on each bridged Java method (via WebView.getUrl).

atorralba (commented):
Hey Ed, I added some early comments. Let's review this again once those are applied and the qhelp is finished.

java/ql/src/Security/CWE/CWE-079/AndroidWebViewAddJavascriptInterfaceExample.java
egregius313: Ok good idea. I'll come up with something for that

sabrowning1 (commented):
@egregius313 thanks for working on another round of docs! I left some small comments, let me know if you have any questions.

java/ql/src/Security/CWE/CWE-079/AndroidWebViewAddJavascriptInterface.ql
```diff
@@ -0,0 +1,17 @@
+/**
+ * @id java/android-webview-addjavascriptinterface
+ * @description Exposing a Javascript interface to a Java object in a WebView can lead to malicious JavaScript controlling the application.
```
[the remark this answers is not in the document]
egregius313: Do you think "with a JavaScript interface" would work? Because `addJavascriptInterface` is the means for exposing an interface.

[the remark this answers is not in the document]
sabrowning1: Ah, in that case that definitely makes more sense than "to a JavaScript interface"! Sounds good to me 👍🏼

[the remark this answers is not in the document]
egregius313: Ok, I changed the wording to "with a JavaScript interface".

sabrowning1 (approved):
Thanks for making those edits @egregius313, looks good for docs! 🚀

atorralba (approved):

PR #11287, opened 2022-11-16
Swift: expand definition of `ipa` in `qlgen`

redsun82:
This will now cover non-final classes that even though are not marked with `@synth` explicitly, they only have `@synth` final descendants.

We currently do not have such example, hence the absence of changes in generated code.

There are currently no unit tests covering the synth part of `qlgen`, as it is planned to split this script into smaller ones, and it would require more effort to port unit tests rather than write them anew when the split happens.


redsun82:
hmm, I changed my mind, I will move this directly to the schema loading, as otherwise I'm more or less duplicating the same code in `cppgen`, `dbschemegen` and `qlgen`

PR #11288, opened 2022-11-16
Golang: add `rsync` as a program capable of arbitrary shell command execution

pwntester:
Add rsync since both --rsh and --rsync-path admit commands


smowton:
@pwntester please rebase so we only get one commit instead of a big merge. I think this would also be sensible to add to `override predicate doubleDashIsSanitizing()` in `SystemCommandExecutors.qll`, since I think `"rsync ... -- " + userData` would be as safe as letting someone control an rsync could ever be?

smowton:
Please add a test case to https://github.com/github/codeql/tree/main/go/ql/test/query-tests/Security/CWE-078 too

pwntester:
@smowton let me know if it looks better now. Would it make sense to share these lists of commands with other languages in a shared qlpack or similar?

smowton (dismissed):

smowton:
Wants a change-note, otherwise looks good

owen-mc:
@pwntester The tests are failing because it doesn't like your declaration of  a function named`handler` - there's already one in another file in the same folder. Also you need to rebase your branch onto `main`.

owen-mc (approved):

pwntester:
@owen-mc sure! I cant merge it though

owen-mc:
@pwntester Ah, I didn't realise. I've done it now. Feel free to prod us if it looks like we've forgotten a PR, especially if it's ready to be merged.

PR #11294, opened 2022-11-16
QL: improve the "this block-comment should have been a QLDoc"-query

erik-krogh:
This would have caught a recent QLDoc check that I spent way too long trying to fix, where I didn't realise that the top-level comment was not a QLDoc.

I also fixed a whole bunch of the issues found by the query.   
Not all, some results are benign.  

And also some improvements to how we match up QLDoc in QL-for-QL.

### DIFF
--- a/ql/ql/src/queries/style/NonDocBlock.ql
+++ b/ql/ql/src/queries/style/NonDocBlock.ql
@@ -30,23 +30,43 @@ int getLineAboveNodeThatCouldHaveDoc(File file) {
 pragma[noinline]
 BlockComment getACommentThatCouldBeQLDoc(File file) {
   file = result.getLocation().getFile() and
-  result.getLocation().getEndLine() = getLineAboveNodeThatCouldHaveDoc(file) and
   result.getLocation().getFile().getExtension() = "qll" and
-  not result.getContents().matches("/**%")
+  not result.getContents().matches("/**%") and
+  (
+    // above something that can be commented.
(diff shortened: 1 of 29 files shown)


michaelnebel (approved):
C# LGTM

ql/ql/src/queries/style/NonDocBlock.ql
```diff
@@ -29,24 +29,59 @@ int getLineAboveNodeThatCouldHaveDoc(File file) {
...
-  canHaveQLDoc(node) and
+  (
+    canHaveQLDoc(node) and
+    comment = getCommentAtEnd(node.getLocation().getFile(), node.getLocation().getStartLine() - 1) and
+    nodeDescrip = "the below code"
+    or
+    node instanceof TopLevel and
+    comment = getCommentAtStart(node.getLocation().getFile(), 1) and
+    nodeDescrip = "the file"
+  ) and
```
geoffw0, line 82: It feels like we could remove some redundant logic here by adding a `reason` parameter to `getACommentThatCouldBeQLDoc` returning `"the below code"` or `"the file"`.

geoffw0 (approved):
C++ differences LGTM.

ql/ql/src/queries/style/NonDocBlock.ql
erik-krogh: Yeah, that could save some duplication.  
But that sounds like something for a followup, this seems good enough to merge.

PR #11296, opened 2022-11-16
Kotlin: format string literals like the Java annotaton extractor

smowton:
Java's regular strings are formatted as they appear in source, but we don't easily have this information available in Kotlin. During annotation extraction however it guesses a source rendering because the source is not necessarily available. By formatting to match the annotation extractor, we prepare to ensure consistency with a Java database when extracting annotations as seen by Kotlin.


aschackmull (approved):
Looks plausible to me.

tamasvajk (dismissed):

cklin:
We should update this file: https://github.com/github/semmle-code/blob/main/odasa-buildutils/kotlin_basic_functionality_test/exprs.ql.expected

smowton:
@cklin done in the corresponding PR

cklin (approved):

PR #11311, opened 2022-11-16
C++: Repair `MustFlow` library for use-use flow

MathiasVP:
The `MustFlow` library was previously defined in terms of the IR-based dataflow nodes, and implicitly relied on the assumption that all relevant instructions had dataflow nodes.

On the feature branch, this is no longer the case. Thus, we cannot use the dataflow nodes, and we have to use the IR directly.

Sadly, this is a breaking change. I think this is okay, however, because:
1. I haven't seen any uses of it outside the two queries we have in Code Scanning that use it.
2. It's really easy to transition from the dataflow nodes to the instructions (since the library was already using the IR-based dataflow library).

### DIFF
--- a/cpp/ql/lib/change-notes/2022-11-16-must-flow.md
+++ b/cpp/ql/lib/change-notes/2022-11-16-must-flow.md
@@ -1,0 +1,4 @@
+---
+category: breaking
+---
+The predicates in the `MustFlow::Configuration` class used by the `MustFlow` library (`semmle.code.cpp.ir.dataflow.MustFlow`) has changed to be defined directly in terms of the C++ IR instead of IR dataflow nodes.
--- a/cpp/ql/lib/change-notes/2022-11-16-must-flow.md
+++ b/cpp/ql/lib/change-notes/2022-11-16-must-flow.md
@@ -1,4 +1,4 @@
 ---
 category: breaking
 ---
-The predicates in the `MustFlow::Configuration` class used by the `MustFlow` library (`semmle.code.cpp.ir.dataflow.MustFlow`) has changed to be defined directly in terms of the C++ IR instead of IR dataflow nodes.
+The predicates in the `MustFlow::Configuration` class used by the `MustFlow` library (`semmle.code.cpp.ir.dataflow.MustFlow`) have changed to be defined directly in terms of the C++ IR instead of IR dataflow nodes.
--- a/cpp/ql/lib/semmle/code/cpp/ir/dataflow/MustFlow.qll
+++ b/cpp/ql/lib/semmle/code/cpp/ir/dataflow/MustFlow.qll
@@ -5,7 +5,6 @@
  */
 
 private import cpp
-import semmle.code.cpp.ir.dataflow.DataFlow
 private import semmle.code.cpp.ir.IR
 
 /**
@@ -25,18 +24,18 @@ abstract class MustFlowConfiguration extends string {
   /**
    * Holds if `source` is a relevant data flow source.
    */
-  abstract predicate isSource(DataFlow::Node source);
+  abstract predicate isSource(Instruction source);
 
   /**
    * Holds if `sink` is a relevant data flow sink.
    */
-  abstract predicate isSink(DataFlow::Node sink);
+  abstract predicate isSink(Operand sink);
 
   /**
    * Holds if the additional flow step from `node1` to `node2` must be taken
    * into account in the analysis.
    */
-  predicate isAdditionalFlowStep(DataFlow::Node node1, DataFlow::Node node2) { none() }
+  predicate isAdditionalFlowStep(Operand node1, Instruction node2) { none() }
 
   /** Holds if this configuration allows flow from arguments to parameters. */
   predicate allowInterproceduralFlow() { any() }
@@ -48,17 +47,17 @@ abstract class MustFlowConfiguration extends string {
    * included in the module `PathGraph`.
    */
   final predicate hasFlowPath(MustFlowPathNode source, MustFlowPathSink sink) {
-    this.isSource(source.getNode()) and
+    this.isSource(source.getInstruction()) and
     source.getASuccessor+() = sink
   }
 }
 
 /** Holds if `node` flows from a source. */
 pragma[nomagic]
-private predicate flowsFromSource(DataFlow::Node node, MustFlowConfiguration config) {
+private predicate flowsFromSource(Instruction node, MustFlowConfiguration config) {
   config.isSource(node)
   or
-  exists(DataFlow::Node mid |
+  exists(Instruction mid |
     step(mid, node, config) and
     flowsFromSource(mid, pragma[only_bind_into](config))
   )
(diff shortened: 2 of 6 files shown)


jketema:
I wonder if this PR should just target `main` instead of the feature branch, as it's pretty independent of the use-use dataflow changes.

MathiasVP:
> I wonder if this PR should just target main instead of the feature branch, as it's pretty independent of the use-use dataflow changes.

Indeed, I'd also be okay with this.

MathiasVP:
I've retargeted the PR to target `main` now. I managed to do so _without_ pinging all teams 🎉, but I didn't manage to stop the bot from adding a bunch of unnecessary labels 😭. Oh well!

cpp/ql/lib/change-notes/2022-11-16-must-flow.md
```diff
@@ -0,0 +1,4 @@
+---
+category: breaking
+---
+The predicates in the `MustFlow::Configuration` class used by the `MustFlow` library (`semmle.code.cpp.ir.dataflow.MustFlow`) has changed to be defined directly in terms of the C++ IR instead of IR dataflow nodes.
```
jketema, line 4:
```suggestion
The predicates in the `MustFlow::Configuration` class used by the `MustFlow` library (`semmle.code.cpp.ir.dataflow.MustFlow`) have changed to be defined directly in terms of the C++ IR instead of IR dataflow nodes.
```

cpp/ql/lib/semmle/code/cpp/ir/dataflow/MustFlow.qll
```diff
@@ -25,18 +24,18 @@ abstract class MustFlowConfiguration extends string {
...
    * Holds if `source` is a relevant data flow source.
    */
-  abstract predicate isSource(DataFlow::Node source);
+  abstract predicate isSource(Instruction source);
 
   /**
    * Holds if `sink` is a relevant data flow sink.
    */
-  abstract predicate isSink(DataFlow::Node sink);
+  abstract predicate isSink(Operand sink);
```
jketema, line 32: This looks somewhat asymmetric now. Is this what we want, or should we introduce a `MustFlowNode` that can either be an instruction or an operand (as we also have a `MustFlowPathNode`)?

cpp/ql/lib/semmle/code/cpp/ir/dataflow/MustFlow.qll
```diff
@@ -48,30 +47,30 @@ abstract class MustFlowConfiguration extends string {
...
 }
 
 /** Holds if `node` flows to a sink. */
 pragma[nomagic]
-private predicate flowsToSink(DataFlow::Node node, MustFlowConfiguration config) {
+private predicate flowsToSink(Instruction node, MustFlowConfiguration config) {
   flowsFromSource(node, pragma[only_bind_into](config)) and
   (
-    config.isSink(node)
+    config.isSink(node.getAUse())
```
jketema, line 71: Is it guaranteed that there still will be a use?

cpp/ql/lib/semmle/code/cpp/ir/dataflow/MustFlow.qll
```diff
@@ -198,12 +197,13 @@ private module Cached {
...
 
   cached
-  predicate step(DataFlow::Node nodeFrom, DataFlow::Node nodeTo) {
-    instructionToOperandStep(nodeFrom.asInstruction(), nodeTo.asOperand())
-    or
-    flowThroughCallable(nodeFrom.asInstruction(), nodeTo.asInstruction())
+  predicate step(Instruction nodeFrom, Instruction nodeTo) {
+    exists(Operand mid |
+      instructionToOperandStep(nodeFrom, mid) and
+      operandToInstructionStep(mid, nodeTo)
```
jketema, line 203: Why is this now needed, while it wasn't before?

jketema (commented):
One typo and a number of related questions. The latter boil down to: why can we get away with just `Instruction`s as sources and `Operand`s as sinks?

jketema:
We should probably also do a DCA experiment once my questions have been resolved.

cpp/ql/lib/semmle/code/cpp/ir/dataflow/MustFlow.qll
MathiasVP: The sink is defined to be an operand. So if there's a sink, then it's guaranteed to be an operand, and that operand will be a use of some instruction (i.e., the instruction currently named `node`). Does that make sense?

cpp/ql/lib/semmle/code/cpp/ir/dataflow/MustFlow.qll
MathiasVP: The `step` function must be a binary relation between two columns of the same type (otherwise a transitive closure over the step relation doesn't make any sense). So if it has to be a binary relation between instructions, then we need to do two things in order to step from one instruction to another:
1. An `Instruction` to `Operand` step.
2. An `Operand` to `Instruction` step.

Previously, since the columns were `DataFlow::Node`s (which contains both instructions and operands), this was done in two iterations of the `step` predicate (i.e., the `instructionToOperandStep(nodeFrom.asInstruction(), nodeTo.asOperand())` disjunct and `operandToInstructionStep(nodeFrom.asOperand(), nodeTo.asInstruction())` disjunct), but we can't do that now since the columns are both instructions. So instead we do two "old" steps in one "new" step.

cpp/ql/lib/semmle/code/cpp/ir/dataflow/MustFlow.qll
MathiasVP: True. I think this is okay since I consider `Instruction`s to be "producers" and `Operand`s to be "consumers", and thus I think it makes sense for an instruction to be a source and an operand to be a sink.

I'd also be totally okay with adding a `MustFlowNode` instead (and having a branch for instructions and a branch for operands).

cpp/ql/lib/semmle/code/cpp/ir/dataflow/MustFlow.qll
jketema: I guess my question is: is the result of an instruction (if it has one) always guaranteed to be used?

MathiasVP: I don't think so, no.

jketema: But those cases weren't captured before either?

MathiasVP: Correct. If the def-use isn't captured explicitly in the IR, then it was never captured by the `MustFlow` library.

jketema: That makes sense. Thanks.

MathiasVP:
> We should probably also do a DCA experiment once my questions have been resolved.

Good point. I've started a run now.

jketema (approved):
I'm happy if DCA is happy.

PR #11315, opened 2022-11-16
C++: Accept test changes on the use-use flow branch

MathiasVP:
I forgot to accept these test changes in https://github.com/github/codeql/pull/11218.

There are no result changes. This is just changing some paths.


geoffw0 (approved):
LGTM.

PR #11316, opened 2022-11-17
CI: Also compile the `examples` folder

erik-krogh:
The cache read doesn't hit anything, but I think that's from the `path:` changing.  
The cache write looks fine, and is slightly larger than before, so it seems to be working. 

Test run simulating what happens when running on `main`: https://github.com/github/codeql/actions/runs/3492067983/jobs/5845431720

### DIFF
--- a/.github/workflows/compile-queries.yml
+++ b/.github/workflows/compile-queries.yml
@@ -26,7 +26,7 @@ jobs:
         if: ${{ github.event_name == 'pull_request' }}
         uses: actions/cache@v3
         with:
-          path: '*/ql/src/.cache'
+          path: '**/.cache'
           key: codeql-compile-pr-${{ github.sha }} # deliberately not using the `compile-compile-main` keys here.
           restore-keys: |
             codeql-compile-${{ github.base_ref }}-${{ env.merge-base }}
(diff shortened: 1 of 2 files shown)


smowton (approved):

.github/workflows/compile-queries.yml
```diff
@@ -51,21 +51,21 @@ jobs:
...
         if : ${{ github.event_name != 'pull_request' }}
         shell: bash
         run: |
           # Move all the existing cache into another folder, so we only preserve the cache for the current queries.
           mkdir -p ${COMBINED_CACHE_DIR}
-          rm */ql/src/.cache/{lock,size}
+          rm -f */ql/{src,examples}/.cache/{lock,size} # -f to avoid errors if the cache is empty.
           # copy the contents of the .cache folders into the combined cache folder.
-          cp -r */ql/src/.cache/* ${COMBINED_CACHE_DIR}/
+          cp -r */ql/{src,examples}/.cache/* ${COMBINED_CACHE_DIR}/ || : # ignore missing files
```
aibaars, line 64: Looks good to me, although I find `true` to be more readable than`:` .

aibaars (approved):

PR #11328, opened 2022-11-18
QL/RB: delete language specific codeql query compile checks

erik-krogh:
We now have a combined workflow that does the same thing for all languages in one go, and which does it faster (thanks to caching).  

ATM still have their specific compilation check as their queries aren't covered by the combined workflow. 
ATM queries have compilation warnings by design, so the combined workflow cannot work for their queries.

### DIFF
--- a/.github/workflows/ruby-qltest.yml
+++ b/.github/workflows/ruby-qltest.yml
@@ -33,11 +33,6 @@ jobs:
     steps:
       - uses: actions/checkout@v3
       - uses: ./.github/actions/fetch-codeql
-      - name: Check QL compilation
-        run: |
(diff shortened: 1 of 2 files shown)


tausbn (dismissed):
Nice!

.github/workflows/ruby-qltest.yml
```diff
@@ -33,11 +33,6 @@ jobs:
     steps:
       - uses: actions/checkout@v3
       - uses: ./.github/actions/fetch-codeql
-      - name: Check QL compilation
```
aibaars, line 36: Let's delete the entire `qlcompile` job.

aibaars (approved):

PR #11329, opened 2022-11-18
C#: Add workflow for running QL tests

hvitved:
This PR adds a job for executing C# QL tests directly in this repo, like we do for Ruby. That is, we
- Fetch a nightly `codeql` CLI.
- Build the C# extractor.
- Execute tests with the nightly CLI, using the freshly built C# extractor and the bundled (legacy) ASP extractor.

Additionally, we make use of Actions cache to cache the compilation cache (like Ruby), and we dispatch the workload onto two `ubuntu-latest-xl` machines (Ruby only uses one).

Comparison:
- The internal `C# Language Tests`, which uses one `ubuntu-latest-xl` machine, typically takes 1h20min.
- The tests introduced by this PR typically take 20 min on a cold cache, and 10 min on a warm cache.

In addition to achieving a very significant speedup, running the tests in this repo also means that external contributors will have access to the results, so we will no longer have to communicate test failures by proxy.

### DIFF
--- a/.github/workflows/csharp-qltest.yml
+++ b/.github/workflows/csharp-qltest.yml
@@ -1,0 +1,72 @@
+name: "C#: Run QL Tests"
+
+on:
+  push:
+    paths:
+      - "csharp/**"
+      - "shared/**"
+      - .github/actions/fetch-codeql/action.yml
+      - codeql-workspace.yml
+    branches:
+      - main
+      - "rc/*"
+  pull_request:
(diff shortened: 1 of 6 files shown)


erik-krogh (approved):
Test workflow and caching 👍

henti (approved):
LGTM

michaelnebel (commented):
Looks plausible to me!

PR #11334, opened 2022-11-18
Make model generator emit trailing comma for set literals

Marcono1234:
Adjusts the model generator to emit trailing commas for set literals (see also https://github.com/github/codeql-cli-binaries/issues/76). This reduces the Git diff when new lines are added at the end.

Note that I am not very familiar with the model generator scripts, but I did not notice any issues locally with this change.


aschackmull:
We plan to change the output of the model generator to a completely different format in the near future, so this change is somewhat irrelevant.

PR #11337, opened 2022-11-21
Ruby: Model ActionMailbox

hmac:
Recognise various calls on `mail` and `inbound_mail.mail` in ActionMailbox classes as remote flow sources. For example:

```rb
class MyMailbox < ActionMailbox::Base

  def process
    mail.body # remote flow
  end
end
```

### DIFF
--- a/ruby/ql/lib/change-notes/2022-11-30-actionmailbox.md
+++ b/ruby/ql/lib/change-notes/2022-11-30-actionmailbox.md
@@ -1,0 +1,4 @@
+---
+ category: minorAnalysis
+---
+ * Calls to `mail` and `inbound_mail` in `ActionMailbox` controllers are now considered sources of remote input.


ruby/ql/lib/codeql/ruby/frameworks/ActionMailbox.qll
```diff
@@ -0,0 +1,54 @@
...
+    result = DataFlow::getConstant("ActionMailbox").getConstant("Base").getADescendentModule()
+  }
+
+  /**
+   * A call to `mail` on the return value of
+   * `ActionMailbox::Base#inbound_email`, or a direct call to
+   * `ActionMailbox::Base#mail`, which is equivalent. The returned object
+   * contains data from the incoming email.
+   */
+  class Mail extends DataFlow::CallNode {
```
alexrford, line 26: This is kind of nitpicky, but as I understand it the `Mail` module is defined in a separate gem https://rubygems.org/gems/mail that isn't part of the rails project. Would it make sense to move this and the `RemoteContent` class below into a separate `Mail.qll` library?

alexrford (commented):
One comment/question about the structure, but LGTM otherwise.

ruby/ql/lib/codeql/ruby/frameworks/ActionMailbox.qll
alexrford: Cool, that seems fine.

alexrford (approved):

PR #11338, opened 2022-11-21
Swift: set @github/codeql-swift as owner

andersfugmann:


andersfugmann:
Members of `@github/codeql-swift` are alexdenisov, redsun82, geoffw0, mathiasvp and karimhamdanali. If we need to make changes, we should do that using the [epd](https://github.com/github/epd) tool.

PR #11339, opened 2022-11-21
Ruby: Active support enumerable

aibaars:


ruby/ql/lib/codeql/ruby/frameworks/ActiveSupport.qll
```diff
@@ -294,7 +294,143 @@ module ActiveSupport {
...
+                ) + ")"
+          )
+        }
+
+        override MethodCall getACall() { result = mc }
+
+        override predicate propagatesFlowExt(string input, string output, boolean preservesValue) {
+          exists(string s, int i |
+            s = getKeyArgument(mc, i) and
+            input = "Argument[self].Element[0].Element[" + s + "!]" and
```
hvitved, line 364: I don't think we want the `!` here.

hvitved (approved):

PR #11341, opened 2022-11-21
CFG: Workaround in test output for origin/target pairs with multiple edges

hvitved:
It turns out that `@kind graph` queries do not currently support multiple edges with the same origin and target (but distinguished by label), so we were missing CFG edges in the test output for C#, Ruby, and Swift.


ruby/ql/test/library-tests/controlflow/graph/Cfg.expected
```diff
@@ -811,7 +811,7 @@ case.rb:
 #-----|  -> [ ..., * ]
 
 #   35| [ ..., * ]
-#-----| match -> x
+#-----| false, match, true -> x
```
aibaars, line 814: Is this actually right? I wonder where the `true` and `false` edges come from.

hvitved: No, this is a bug. It may be addressed already by @hmac's [CFG changes](https://github.com/github/codeql/pull/11114), but in all cases, I think we should handle it follow-up.

aibaars: Perfect!

aibaars (approved):

PR #11343, opened 2022-11-21
QL: add redundant-assignment query

erik-krogh:
A quick idea I had.  
Assigning the same value to a variable twice is redundant, so the second assignment can be removed.  
There weren't a lot of results.


owen-mc (approved):
Go changes approved

hvitved (approved):
C# 👍  Nice!

ql/ql/src/queries/style/RedundantAssignment.ql
```diff
@@ -0,0 +1,110 @@
...
+ * Holds if `assigned1` and `assigned2` assigns the same value to `var`.
+ * The assignments may be on different branches of a disjunction.
+ */
+predicate candidateRedundantAssignment(AssignedVariable var, Expr assigned1, Expr assigned2) {
+  assigned1 = var.getAnAssignedExpr() and
+  assigned2 = var.getAnAssignedExpr() and
+  (
+    GVN::valueNumber(assigned1) = GVN::valueNumber(assigned2)
+    or
+    // because GVN skips large strings, we need to check for equality manually
```
MathiasVP, line 55: I really should investigate why this ugly hack is necessary 🤔.

erik-krogh: I think it's the MaD rows that go wrong in some way.
There are some massive disjunctions in there.

PR #11344, opened 2022-11-21
Ruby: Use compilation cache for the qltest CI workflow

erik-krogh:


.github/workflows/ruby-qltest.yml
```diff
@@ -48,17 +44,20 @@ jobs:
...
       - uses: ./ruby/actions/create-extractor-pack
+      - name: Cache compilation cache
+        id: query-cache
+        uses: ./.github/actions/cache-query-compilation
+        with: 
+          key: ruby-qltest
       - name: Run QL tests
         run: |
-          codeql test run --threads=0 --ram 5000 --slice ${{ matrix.slice }} --search-path "${{ github.workspace }}/ruby/extractor-pack" --check-databases --check-undefined-labels --check-unused-labels --check-repeated-labels --check-redefined-labels --check-use-before-definition --consistency-queries ql/consistency-queries ql/test
+          codeql test run --threads=0 --search-path "${{ github.workspace }}/ruby/extractor-pack" --check-databases --check-undefined-labels --check-unused-labels --check-repeated-labels --check-redefined-labels --check-use-before-definition --consistency-queries ql/consistency-queries ql/test --compilation-cache "${{ steps.query-cache.outputs.cache-dir }}"
```
hvitved, line 61: AFAIK, it is still a good idea to set `--ram`. I believe these machines have 56 GB or RAM, so setting to e.g. 52 GB should work.

hvitved (commented):
Looks plausible to me.

.github/actions/cache-query-compilation/action.yml
```diff
@@ -0,0 +1,60 @@
...
+    # Cache the query compilation caches.
+    # calculate the merge-base with main, in a way that works both on PRs and pushes to main. 
+    - name: Calculate merge-base
+      shell: bash
+      if: ${{ github.event_name == 'pull_request' }}
+      env:
+        BASE_BRANCH: ${{ github.base_ref }}
+      run: |
+        MERGE_BASE=$(git cat-file commit $GITHUB_SHA | grep '^parent ' | head -1 | cut -f 2 -d " ")
+        echo "merge-base=$MERGE_BASE" >> $GITHUB_ENV
```
aibaars, line 26: `merge-base` is not a valid environment variable in `bash`. You cant write `${merge-base}` in a script. I'd recommend using `merge_base`  or `MERGE_BASE`.

aibaars: Is there a reason for using `GITHUB_ENV` here instead of `GITHUB_OUTPUT` ?

aibaars (approved):

.github/actions/cache-query-compilation/action.yml
erik-krogh: > Is there a reason for using GITHUB_ENV here instead of GITHUB_OUTPUT ?

Nope, it was just the first data-sharing mechanism I found.  
But now that I look at it, I still like it. I don't have to do `${{ steps.step-name.outputs.output-name }}`, but can just do `${{ env.merge_base }}`.

erik-krogh: > merge-base is not a valid environment variable in bash. You cant write ${merge-base} in a script. I'd recommend using merge_base or MERGE_BASE.

It works with `merge-base` 🤷  
I changed it anyway.

aibaars: Thanks! It's true that `merge-base` works, however, things get annoying if you want to print the value in a bash script for debugging reasons. It's just to prevent our future selves from scratching our heads why `${merge-base} == "base"` ;-)

aibaars (changes requested):
Perhaps move caching the extractor to a follow-up PR.

.github/workflows/ruby-qltest.yml
```diff
@@ -48,17 +44,20 @@ jobs:
...
       - uses: ./ruby/actions/create-extractor-pack
+      - name: Cache compilation cache
+        id: query-cache
+        uses: ./.github/actions/cache-query-compilation
+        with: 
+          key: ruby-qltest
       - name: Run QL tests
         run: |
-          codeql test run --threads=0 --ram 5000 --slice ${{ matrix.slice }} --search-path "${{ github.workspace }}/ruby/extractor-pack" --check-databases --check-undefined-labels --check-unused-labels --check-repeated-labels --check-redefined-labels --check-use-before-definition --consistency-queries ql/consistency-queries ql/test
+          codeql test run --threads=0 --ram 52000 --search-path "${{ github.workspace }}/ruby/extractor-pack" --check-databases --check-undefined-labels --check-unused-labels --check-repeated-labels --check-redefined-labels --check-use-before-definition --consistency-queries ql/consistency-queries ql/test --compilation-cache "${{ steps.query-cache.outputs.cache-dir }}"
```
p0, line 61: @erik-krogh Did you mean to raise the `--ram` argument to 52G here?

erik-krogh: Yes.  
That change was based on [this comment by Tom](https://github.com/github/codeql/pull/11344#discussion_r1028071285) further up. 

I didn't observe any difference from changing `--ram`, so it seemed fine to me.

PR #11347, opened 2022-11-21
Python: Add change note for module resolution

tausbn:
Also adapts the version-specific tests to support results specific to Python 2 (though at the moment there are no such tests).

### DIFF
--- a/python/ql/lib/change-notes/2022-11-21-module-resolution-rewrite.md
+++ b/python/ql/lib/change-notes/2022-11-21-module-resolution-rewrite.md
@@ -1,0 +1,5 @@
+---
+ category: minorAnalysis
+---
+ * The data-flow library has been rewritten to no longer rely on the points-to analysis in order to
+   resolve references to modules. This should result in more results for data-flow queries.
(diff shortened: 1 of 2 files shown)


python/ql/lib/change-notes/2022-11-21-module-resolution-rewrite.md
```diff
@@ -0,0 +1,5 @@
+---
+ category: minorAnalysis
+---
+ * The data-flow library has been rewritten to no longer rely on the points-to analysis in order to
+   resolve references to modules. This should result in more results for data-flow queries.
```
RasmusWL, line 5: I was going to make this suggestion, since non data-flow queries that just rely on type-tracking will also be improved. 

```suggestion
   resolve references to modules. This should result in more results.
```
The phrasing of the last paragraph makes it sounds a bit like if you _don't_ see more results, that indicates a problem (since you _should_ see more results). So I'd also like to come with this rephrased suggestion:

```suggestion
   resolve references to modules. Improvements in the module resolution can lead to more results.
```

RasmusWL (changes requested):
A minor thing about the wording of the change-note, otherwise LGTM :+1:

PR #11348, opened 2022-11-21
Ruby: cache the entire extractor

erik-krogh:


.github/workflows/ruby-build.yml
```diff
@@ -86,20 +82,23 @@ jobs:
...
         run: |
-          codeql pack create ../shared/ssa --output target/packs
-          codeql pack create ../misc/suite-helpers --output target/packs
-          codeql pack create ../shared/regex --output target/packs
-          codeql pack create ql/lib --output target/packs
-          codeql pack create ql/src --output target/packs
+          codeql pack create -j0 ../shared/ssa --output target/packs --compilation-cache "${{ steps.query-cache.outputs.cache-dir }}"
+          codeql pack create -j0 ../misc/suite-helpers --output target/packs --compilation-cache "${{ steps.query-cache.outputs.cache-dir }}"
+          codeql pack create -j0 ../shared/regex --output target/packs --compilation-cache "${{ steps.query-cache.outputs.cache-dir }}"
+          codeql pack create -j0 ql/lib --output target/packs --compilation-cache "${{ steps.query-cache.outputs.cache-dir }}"
```
aibaars, line 100: Multi-threading and compilation caching is probably overkill for library pack creation. I believe library packs are not compiled at all. It still makes sense to use the cache for the `ql/src` pack though.
```suggestion
          codeql pack create ../shared/ssa --output target/packs
          codeql pack create ../misc/suite-helpers --output target/packs
          codeql pack create ../shared/regex --output target/packs
          codeql pack create ql/lib --output target/packs
```

aibaars (approved):
Looks good to me.  You might want to simplify the `codeql pack create` commands for the library packs.

.github/workflows/ruby-build.yml
```diff
@@ -86,20 +86,23 @@ jobs:
             ruby/target/release/ruby-extractor.exe
           retention-days: 1
   compile-queries:
-    runs-on: ubuntu-latest
-    env:
-      CODEQL_THREADS: 4 # TODO: remove this once it's set by the CLI
+    runs-on: ubuntu-latest-xl
```
aibaars, line 89: I guess XL runners are a bit overkill , this job only compiles some Ruby queries, and most of them should be cached already.

erik-krogh: I still want to keep it.  
If the core libraries are changed the cache won't help much. 
And even when we get a full cache hit the XL worker is significantly faster.

PR #11350, opened 2022-11-21
Add macOS 13 to supported platforms

mbg:
I noticed that the list of supported platforms in the documentation does not yet include macOS 13. As far as I can tell, everything runs fine on it, so this change adds macOS 13 to the list of supported platforms.

### DIFF
--- a/docs/codeql/support/reusables/platforms.rst
+++ b/docs/codeql/support/reusables/platforms.rst
@@ -18,7 +18,11 @@
 
    macOS 11 Big Sur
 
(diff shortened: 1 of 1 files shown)


smowton:
Rebased and chased as https://github.com/github/codeql/pull/11350

intrigus-lgtm:
I guess #12097 is the fix commit and not #11350 :)

smowton:
Hah, yes, I won't try to stand by the claim that this PR is its own resurrection

PR #11355, opened 2022-11-21
C++: Reduce `readStep` fan-in

MathiasVP:
On the feature branch, every `IndirectOperand` whose operand is the qualifier of a field address will generate a `readStep`. This is very wasteful, as the only nodes produced by `storeStep` are the ones at indirection index 1.

Other than producing too many `readStep`s, this also generated some dataflow-consistency issues since nodes that were going into `readStep`s couldn't didn't have corresponding `PostUpdateNode`s (i.e., the nodes usually directly produced by `storeStep`s).

This PR is a one-character fix for this: only the `IndirectOperand`s with an indirection index of 1 should generate a `readStep`.


geoffw0 (approved):
Seems reasonable.

PR #11356, opened 2022-11-21
 C++: replace Guards with IRGuards

rdmarsh2:


MathiasVP:
@rdmarsh2 I think the test changes are all good, but I'd appreciate a second pair of 👀s on it.

jketema:
Happily have a look, but please update the tests, so I don't have to look at the workflow output.

jketema (approved):
Test changes make sense to me.

MathiasVP:
I guess the remaining task is to run a DCA on this 🎉?

rdmarsh2:
DCA run: https://github.com/github/codeql-dca-main/issues/8858#issuecomment-1329821671

MathiasVP:
Looks like there are a couple of results on `cpp/toctou-race-condition` that we need to investigate. It's probably because we're catching more sanitizers now 🤞.

rdmarsh2:
I think it's related to non-returning functions that aren't modeled as such in the IR...

MathiasVP:
> I think it's related to non-returning functions that aren't modeled as such in the IR...

Indeed. Looking at the very first lost result on SAMATE, It looks like we're no longer raising an alert on this pattern (from `C/testcases/CWE367_TOC_TOU/CWE367_TOC_TOU__stat_08.c`):
```cpp
if (STAT(filename, &statBuffer) == -1)
{
    exit(1);
}
fileDesc  = OPEN(filename, O_RDWR);
```

Do we need to teach the IR guards library about non-returning functions before we can merge this PR?

rdmarsh2:
I think we'd be better off modeling it properly in the IR than handling it in the Guards library specifically. A jump from the call to the `Unreachable` block should make the guards library work as expected.

cpp/ql/lib/semmle/code/cpp/ir/implementation/raw/internal/IRConstruction.qll
```diff
@@ -386,6 +386,16 @@ Instruction getPrimaryInstructionForSideEffect(SideEffectInstruction instruction
         .getPrimaryInstructionForSideEffect(getInstructionTag(instruction))
 }
 
+predicate hasUnreachedInstruction(IRFunction func) {
+  exists(Call c | 
+    c.getEnclosingFunction() = func.getFunction() and
+    (
+      c.getTarget().hasSpecifier("_Noreturn") or
```
jketema, line 393: As far as I remember `_Noreturn` is represented as `noreturn` in the database.

Wouldn't it be better to use the `exits(Function f)` predicate from `DefaultOptions.qll`? And, `exprExits(Expr e)` if you want to do something similar for `assume(0)` expressions?

jketema: For reference: https://github.com/github/codeql/commit/1a1c34e1bebf558142af6305c144a88345059714

jketema:
I seems some files need to be sync'ed and some formatting is required.

MathiasVP:
It seems like `cpp/toctou-race-condition` is an excellent test for the Guards library 😂. I like the fact that the set of result changes is very small! It would be interesting to see if these are good or bad changes, though.

jketema:
> It seems like `cpp/toctou-race-condition` is an excellent test for the Guards library 😂. I like the fact that the set of result changes is very small! It would be interesting to see if these are good or bad changes, though.

The unclassified results that disappeared are all of the form:
```
if (file check fails)
  exit()
do something with the file
```
So that looks ok.

The new unclassified results look a bit funny, though not wrong.
```
		filemode = (!chmod(path, st1.st_mode ^ S_IXUSR) &&  // file being operated on and two checks
				!lstat(path, &st2) &&               // another check
				st1.st_mode != st2.st_mode &&       // another check
				!chmod(path, st1.st_mode));         // yet another check
```
and
```
    return (chmod((char *)                                       // file being operated on
		    name,                                        // check
		    (mode_t)perm) == 0 ? OK : FAIL);
```

jketema:
Three of the syntax zoo consistency check tests are currently failing by the way.

jketema:
Should we update this PR now https://github.com/github/codeql/pull/12982 is merged?

MathiasVP:
Yeah, let's at least fix up the merge conflicts. I guess we need to schedule when to do the last part of this work at the next iteration planning.

geoffw0:
Superceded by https://github.com/github/codeql/pull/14992 ?

PR #11357, opened 2022-11-21
C++: Ignore more instructions in dataflow

MathiasVP:
These instructions aren't used for dataflow. So there's no need to allocate `DataFlow::Node`s for them.


jketema:
The expected results of `cpp/ql/test/library-tests/syntax-zoo/dataflow-ir-consistency.ql` need to be updated.

jketema (approved):
LGTM

MathiasVP:
Thanks! Done in a2ce51fdf77f5e8e3941410e4624bccab858fab0. Always nice to see unexpected consistency fixes 😂.

PR #11362, opened 2022-11-22
CI: use read-only-cache when running on a PR

erik-krogh:
[The actions cache is currently filling up with loads of useless `codeql-compile-*` caches that will never be read](https://github.com/github/codeql/actions/caches).  
Currently our oldest cache is 22 hours old, everything older has been evicted from the actions cache.   

My solution is to not save the compilation cache when running on PRs, but only do it on pushes to `main`.  
That way we only save the caches that we expect to read.  

Read-only caches is not a feature currently supported by `actions/cache`, so I'm using my own fork.  
We should go back to `actions/cache` once https://github.com/actions/cache/pull/995 (or similar) has been merged.

### DIFF
--- a/.github/actions/cache-query-compilation/action.yml
+++ b/.github/actions/cache-query-compilation/action.yml
@@ -26,9 +26,10 @@ runs:
         echo "merge_base=$MERGE_BASE" >> $GITHUB_ENV
     - name: Read CodeQL query compilation - PR
       if: ${{ github.event_name == 'pull_request' }}
-      uses: actions/cache@v3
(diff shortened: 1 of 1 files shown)


aibaars (approved):
Looks good to me. Shall we wait until https://github.com/actions/cache/pull/995 (or similar) is merged? Or is it better to merge this right away to avoid filling up the cache storage space?

erik-krogh:
> Looks good to me. Shall we wait until [actions/cache#995](https://github.com/actions/cache/pull/995) (or similar) is merged? Or is it better to merge this right away to avoid filling up the cache storage space?

I suspect that the PR will take a while to merge, if it gets merged at all.   
So I don't want to wait.

PR #11365, opened 2022-11-22
Swift: fix extractor tests pack

redsun82:
Prior to this, `codeql query compile` on the integration tests queries was complaining that it could not find the `codeql/swift-all` pack (even though `codeql test run` managed it just fine). After using the `${workspace}` variable, it was also necessary to include the integration tests pack into the workspace, but then it became possible to compile the queries without running the test.


dbartol (approved):

geoffw0:
🎉

PR #11366, opened 2022-11-22
Ruby: Add additional sinks to the `rb/kernel-open` query

p-:
Hi!

This PR adds following sinks to the `rb/kernel-open` query:
* `IO.write`
* `IO.binread`
* `IO.binwrite`
* `IO.foreach`
* `IO.readlines`
* `URI.open`

These sinks all belong to the [same family](https://cheatsheetseries.owasp.org/cheatsheets/Ruby_on_Rails_Cheat_Sheet.html#command-injection) as the already implemented `Kernel.open` and `IO.read`. (Command injection with a starting pipe (`|`).

In addition to that I added a sanitizer I encountered in the wild: `File.join` sanitizes an untrusted user input if the tainted data is not passed in as the first argument (command execution is only possible if the first sign character passed to those sinks is a pipe (`|`).


erik-krogh (approved):
I had my doubt abouts `URI.open`, but yes that's also vulnerable 🤦  

```bash
irb(main):001:0> require "open-uri"
=> true
irb(main):002:0> URI
=> URI
irb(main):003:0> URI.open "https://google.com"
=> #<File:/var/folders/qp/rmn5qy357vb70d2sgqv1tt_c0000gn/T/open-uri20221122-8505-omzyyf>
irb(main):004:0> URI.open "| touch pwned"
=> #<IO:fd 11>
```

LGTM 👍

erik-krogh (changes requested):
I'm sorry, I'm going to need changes.  
Can you see [the evaluation](https://github.com/github/codeql-dca-main/blob/data/erik-krogh/PR-11366-0-ruby/reports/alert-comparison.md)?   

There are bunch of results in `opal/opal` where the alert is on a safe call to e.g. `File.write`, but according to the alert message the call is to `IO.write`, so you'll have to do something about that.  
You can see in `KernelOpenQuery.qll` that I've already done that for `IO.read`.   

I think that happens due to some aliasing 🤷, but I don't know precisely how.  
(You can ask the more senior CodeQL Ruby devs on Slack about that).

erik-krogh:
I found some more FPs, I just pushed a fix directly to this PR.   
I've triggered another evaluation.

calumgrant:
I looked into the in-the-wild alerts generated by the existing query - it's quite a low-volume query generating only 2 alerts in the past 90 days. This is good for this PR because it means we're not too noisy yet.

ruby/ql/lib/codeql/ruby/security/KernelOpenQuery.qll
```diff
@@ -16,21 +16,70 @@ class AmbiguousPathCall extends DataFlow::CallNode {
...
+/**
+ * A sanitizer for kernel open vulnerabilities.
+ */
+abstract class Sanitizer extends DataFlow::Node { }
+
+/**
+ * If a File.join is performed the resulting string will not start with a pipe |
+ * This is true as long the tainted data doesn't flow into the first argument.
+ */
+private class FileJoinCall extends Sanitizer, DataFlow::ExprNode {
```
alexrford, line 81: Marking the arguments here ends up potentially sanitizing nodes on a dataflow path that's not actually sanitized, e.g. in 
```
 file = params[:file]
 IO.read(File.join("", file)) # this call is safe
 IO.read(file)                # this call is unsafe, but marking the `file` argument to `File.join` as a sanitizer means that we think it's safe
```
With the `SummarizedCallable::getParameter` predicate from https://github.com/github/codeql/pull/11621, we can redefine this sanitizer as:
```
private class FileJoinSanitizer extends Sanitizer {
  FileJoinSanitizer() { this = any(File::FileJoinSummary s).getParameter("1..") }
}
```
after changing https://github.com/github/codeql/blob/6fa2fe6c864d522769fa3132b0b9c869775a53b8/ruby/ql/lib/codeql/ruby/frameworks/Files.qll#L136 to `input = "Argument[0,1..]"`.

This lets us use the synthesized parameter node (for any parameter with index 1 or higher) as a sanitizer, meaning that only flow into the `File.join` call is sanitized rather than all flow from the argument.

I recommend trying this out with a new test case to check that I haven't made a mistake here.

alexrford (changes requested):
Just a suggestion on the `File.join` sanitizer and a nitpick.

ruby/ql/lib/codeql/ruby/security/KernelOpenQuery.qll
```diff
@@ -16,21 +16,70 @@ class AmbiguousPathCall extends DataFlow::CallNode {
...
+    name = "IO.binread"
+    or
+    methodCallOnlyOnIO(this, "binwrite") and
+    name = "IO.binwrite"
+    or
+    methodCallOnlyOnIO(this, "foreach") and
+    name = "IO.foreach"
+    or
+    methodCallOnlyOnIO(this, "readlines") and
+    name = "IO.readlines"
```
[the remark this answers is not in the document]
p-: Thanks!

ruby/ql/lib/codeql/ruby/security/KernelOpenQuery.qll
alexrford: The new tests look good 👍

alexrford (commented):
Started a new DCA run, I'm happy with this as long as the results are okay.

alexrford (approved):
DCA looks good 👍

erik-krogh (approved):
LGTM 👍  

I'll merge if no one complains at our monday meeting.

PR #11367, opened 2022-11-22
Java: Consider taint through bitwise operations on PendingIntent flags

atorralba:
This fixes spurious results when `PendingIntent.FLAG_IMMUTABLE` is correctly used on a `PendingIntent` but it goes through a bitwise operation, which previously broke taint flow.


aschackmull (approved):

atorralba:
DCA looks acceptable (aside from a couple errors caused by running out of disk space, which doesn't seem related to this change).

PR #11370, opened 2022-11-22
Swift: upgrade to Swift 5.7.1

AlexDenisov:


swift/extractor/infra/SwiftDispatcher.h
```diff
@@ -215,6 +215,17 @@ class SwiftDispatcher {
...
+    // getPatternList() when getting the source location.
+    // So far this only observed with the entities coming from precompiled modules, so they
+    // should not have locations anyway.
+    if constexpr (std::is_base_of<swift::Decl, Locatable>::value) {
+      if (auto* decl = llvm::dyn_cast<swift::PatternBindingDecl>(locatable)) {
+        if (decl->getPatternList().empty()) {
+          return;
+        }
+      }
+    }
```
redsun82, line 228: I don't think you need a `dyn_cast` here because `Locatable` is already `const swift::PatternBindingDecl` here. And also, rather than not giving any location, we can use the (existing) start location as and location as well.
So maybe the following works as well:
```suggestion
    if constexpr (std::is_same_v<const swift::PatternBindingDecl, Locatable>) {
      if (locatable->getPatternList().empty()) {
        attachLocation(locatable->getStartLoc(), locatableLabel);
        return;
      }
    }
```
Apart from that, this is a bit ewww :sweat_smile: 

Shouldn't we maybe consider patching this on `PatternBindingDecl::getSourceRange()` before prebuilding? We already have patches on the swift frontend library, although for the moment it's just for the build system, right? If we go for that we could also move there the patch I already introduced for headers.

AlexDenisov: Yeah, patching the source code makes more sense, agreed, I'll clean it up when we resolve all the rest of the issues if it doesn't cause too much friction.

The suggested change won't work unfortunately as the crash is coming right from `Decl::getStart/EndLoc` which  subsequently calls `getSourceRange`.

redsun82: ah, you are right, this code is called in `fetchLabel` that might be called with a less specific type than the translation actually uses. If we moved the location attachment to `createEntry` instead, we would always get the exact specific type and we would be able to do this kind of tweaks to locations on a per-type basis. But we can leave that for another PR

redsun82:
@AlexDenisov could you try applying
```diff
diff --git a/swift/actions/run-integration-tests/action.yml b/swift/actions/run-integration-tests/action.yml
index 9325fb8b11..2ff09ae9c0 100644
--- a/swift/actions/run-integration-tests/action.yml
+++ b/swift/actions/run-integration-tests/action.yml
@@ -13,7 +13,7 @@ runs:
     - uses: actions/setup-python@v4
       with:
         python-version-file: 'swift/.python-version'
-    - uses: swift-actions/setup-swift@v1
+    - uses: redsun82/setup-swift@63ad25fa169bb1e89d8aa67036007d6f3078e9f6
       with:
         swift-version: "${{steps.get_swift_version.outputs.version}}"
     - uses: ./.github/actions/fetch-codeql
```
This might make the setup swift work with 5.7.1 in the integration tests.

AlexDenisov:
Started an experiment https://github.com/github/codeql-dca-main/issues/8901

AlexDenisov:
GitHub mobile app also extracts without any issues.

swift/downgrades/e02469b183f72c50d72f0c6926fdd8366ae45aec/upgrade.properties
```diff
@@ -0,0 +1,4 @@
+description: Remove new implicit conversion
+compatibility: partial
+unspecified_elements.rel: run unspecified_elements.ql
+implicit_conversion_exprs.rel: run implicit_conversion_exprs.ql
```
redsun82, line 4: forgot about this when we chatted, we also need to delete the now non-existing table
```suggestion
implicit_conversion_exprs.rel: run implicit_conversion_exprs.ql
abi_safe_conversion_exprs.rel: delete
```

AlexDenisov:
The DCA run looks good to me, but any extra 👀 would be appreciated https://github.com/github/codeql-dca-main/issues/8901#issuecomment-1331855649

MathiasVP:
The DCA looks fine 👍. There are some weird speedups and slowdowns for various performance summaries, but this can probably all be attributed to the macOS runners.

redsun82 (approved):
That's great!

PR #11372, opened 2022-11-22
Python: Model `getpass.getpass` as source of passwords

RasmusWL:


tausbn (approved):
Looks good to me. I liked the fact that making space for the test happened in a separate commit. ❤️ 

The way you exclude results internal to `getpass` is interesting. If we need to do this a lot, I think we'll need to figure out a more modular solution. 

Ideally, I feel like the fact that `getpass` is a source should enable us to prune away any results that arise from inside the `getpass` call itself. (But of course, the obvious way of doing this would cause negative recursion.)

PR #11374, opened 2022-11-22
C++: Field flow through reference-returning functions

MathiasVP:
This PR adds use-use flow through examples such as:
```cpp
int source();
void sink(int);

struct MyInt {
  int i;
  int* get() { return &i; }
};

void test()
{
  MyInt mi;
  *mi.get() = source();
  sink(mi.i);
}
```
It turns out we _almost_ had this flow on the feature branch already. The shared dataflow library has a concept of flow-through [here](https://github.com/github/codeql/blob/main/java/ql/lib/semmle/code/java/dataflow/internal/DataFlowImplCommon.qll#L804). This disjunct is responsible for adding a store step from `c [post update]` to `getB() [post update]` in the following example:
```cpp
struct B { int c; };
struct A {
  B b;
  B* getB() { return b; }
};
...
a.getB()->c = v;
```
the rule used by the shared dataflow library is as follows:

Generate a store step from `node1` to `node2` (storing `f`) if:
1. `node1` is a `PostUpdateNode` with a pre-update node that is an output of a function (that's `getB() [post update]` in our example above node above with a pre-update node of `getB()`)
2. `node2` is a `PostUpdateNode` with a pre-update node that is an argument to a function (that's `a [post update]` in our example above with a pre-update node of `a`).
3. There's flow from the argument node, into the function, and to a read step that reads `f`, which then flows to a return node of the function.

Step 2 and 3 already worked for our example, but there was no `PostUpdateNode` with a pre-update note that was the output of the function. So this PR simply marks the appropriate `OutNode` as a `PostUpdateNode` and assigns the pre-update node to `this`.

Having the pre-update node equal the post-update node _is_ a consistency violation (as demonstrated by the tests). I'm not aware of anything bad happening due to this, however. I'll run DCA to make sure, though 🤞.

### DIFF
--- a/cpp/ql/lib/semmle/code/cpp/ir/dataflow/internal/DataFlowUtil.qll
+++ b/cpp/ql/lib/semmle/code/cpp/ir/dataflow/internal/DataFlowUtil.qll
@@ -596,9 +596,32 @@ predicate indirectReturnOutNodeInstruction0(
 }
 
 /**
+ * Holds if `node` is an indirect operand with columns `(operand, indirectionIndex)`, and
+ * `operand` represents a use of the fully converted value of `call`.
+ */
+private predicate hasOperand(Node node, CallInstruction call, int indirectionIndex, Operand operand) {
+  indirectReturnOutNodeOperand0(call, operand, indirectionIndex) and
+  hasOperandAndIndex(node, operand, indirectionIndex)
+}
+
+/**
+ * Holds if `node` is an indirect instruction with columns `(instr, indirectionIndex)`, and
+ * `instr` represents a use of the fully converted value of `call`.
+ *
+ * Note that `hasOperand(node, _, _, _)` implies `not hasInstruction(node, _, _, _)`.
+ */
+private predicate hasInstruction(
+  Node node, CallInstruction call, int indirectionIndex, Instruction instr
+) {
+  indirectReturnOutNodeInstruction0(call, instr, indirectionIndex) and
+  hasInstructionAndIndex(node, instr, indirectionIndex)
+}
+
+/**
  * INTERNAL: do not use.
  *
- * A node representing the value of a function call.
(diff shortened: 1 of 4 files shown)


MathiasVP:
DCA looks good 🎉. No result changes, and no performance changes.

rdmarsh2:
> marks the appropriate OutNode as a PostUpdateNode and assigns the pre-update node to `this`.

What do other languages do here? Separate post-update and pre-update nodes for each return?

aschackmull:
> I'm not aware of anything bad happening due to this, however

The thing that can happen is weird FP flow in certain cases. If the store step that targets the result of the wrapped read isn't written directly, as it is in your examples, but instead occurs as a side-effect of a call then you'll get spurious re-entry flow into that setter. Let me demonstrate with an example:
```
Foo getFoo() { return this.foo; }
...
void setBar(x) {
  sink(this.Bar);
  this.Bar = x;
}
...
x.getFoo().setBar(source())
```
If the `OutNode` associated with the result of `getFoo` is its own `PostUpdateNode` then the post-update will also be an argument to `setBar` and the result of the side-effect from `setBar` will then be able to re-enter `setBar` and cause FP flow from `source` to `sink`.

MathiasVP:
Thanks for your input, @aschackmull! Luckily, the above example doesn't apply in this case since we're doing the Right Thing™ for constructs such as:
```
x.getFoo().setBar(source())
```
where `x.getFoo()` will have a post-update node `x.getFoo() [post update]` since it's the argument of a function call (and its pre-update node will be the expression node for `x.getFoo()`).

This PR only does the pre-update = post-update on constructs like `*mi.get() = source();` (where the destination address is an address returned by a call to a function that returns a reference or a pointer).

rdmarsh2 (approved):

PR #11378, opened 2022-11-22
Swift: Add models for NSData and NSMutableData

atorralba:
Note that models requiring flow through code blocks have been left out (see `TODO` comments).

### DIFF
--- a/swift/ql/lib/codeql/swift/frameworks/StandardLibrary/NSData.qll
+++ b/swift/ql/lib/codeql/swift/frameworks/StandardLibrary/NSData.qll
@@ -1,0 +1,92 @@
+/** Provides classes and models to work with `NSData`-related objects. */
+
+import swift
+private import codeql.swift.dataflow.DataFlow
+private import codeql.swift.dataflow.ExternalFlow
+private import codeql.swift.dataflow.FlowSteps
+
+/** The class `NSData`. */
+class NsData extends ClassDecl {
+  NsData() { this.getFullName() = "NSData" }
+}
+
+/** The class `NSMutableData`. */
+class NsMutableData extends ClassDecl {
+  NsMutableData() { this.getFullName() = "NSMutableData" }
+}
+
+private class NsDataSources extends SourceModelCsv {
+  override predicate row(string row) {
+    row =
+      [
+        ";NSData;true;init(contentsOf:);;;ReturnValue;remote",
+        ";NSData;true;init(contentsOf:options:);;;ReturnValue;remote"
+      ]
+  }
+}
+
+private class NsDataSummaries extends SummaryModelCsv {
+  override predicate row(string row) {
+    row =
+      [
+        ";NSData;true;init(bytes:length:);;;Argument[0];ReturnValue;taint",
+        ";NSData;true;init(bytesNoCopy:length:);;;Argument[0];ReturnValue;taint",
+        ";NSData;true;init(bytesNoCopy:length:deallocator:);;;Argument[0];ReturnValue;taint",
+        ";NSData;true;init(bytesNoCopy:length:freeWhenDone:);;;Argument[0];ReturnValue;taint",
+        ";NSData;true;init(data:);;;Argument[0];ReturnValue;taint",
+        ";NSData;true;init(contentsOfFile:);;;Argument[0];ReturnValue;taint",
+        ";NSData;true;init(contentsOfFile:options:);;;Argument[0];ReturnValue;taint",
+        ";NSData;true;init(contentsOf:);;;Argument[0];ReturnValue;taint",
+        ";NSData;true;init(contentsOf:options:);;;Argument[0];ReturnValue;taint",
+        ";NSData;true;init(contentsOfMappedFile:);;;Argument[0];ReturnValue;taint",
+        ";NSData;true;init(base64Encoded:options:);;;Argument[0];ReturnValue;taint",
+        ";NSData;true;init(base64Encoding:);;;Argument[0];ReturnValue;taint",
+        ";NSData;true;base64EncodedData(options:);;;Argument[-1];ReturnValue;taint",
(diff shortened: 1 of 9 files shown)


swift/ql/test/library-tests/dataflow/taint/nsmutabledata.swift
```diff
@@ -0,0 +1,50 @@
...
+    nsMutableDataTainted4.replaceBytes(in: NSRange(), withBytes: source() as? UnsafeRawPointer, length: 0)
+    sink(arg: nsMutableDataTainted4) // $ tainted=40
+    // ";NSMutableData;true;setData(_:);;;Argument[1];Argument[-1];taint",
+    let nsMutableDataTainted5 = NSMutableData()
+    nsMutableDataTainted5.setData(source() as! Data)
+    sink(arg: nsMutableDataTainted5) // $ tainted=44
+
+    // Fields
+    let nsMutableDataTainted6 = source() as! NSMutableData
+    sink(arg: nsMutableDataTainted6.mutableBytes) // $ tainted=48
```
geoffw0, line 49: Should we test coverage of writes to `mutableBytes` as well?  As I understand it we don't model this at the moment, but it would still be good to have a test documenting that.

geoffw0 (approved):
Looks good to me.  Very thorough.  This should have a big effect on our taint flow!

swift/ql/lib/codeql/swift/frameworks/StandardLibrary/NSData.qll
```diff
@@ -0,0 +1,92 @@
...
+        ";NSData;true;init(contentsOf:options:);;;Argument[0];ReturnValue;taint",
+        ";NSData;true;init(contentsOfMappedFile:);;;Argument[0];ReturnValue;taint",
+        ";NSData;true;init(base64Encoded:options:);;;Argument[0];ReturnValue;taint",
+        ";NSData;true;init(base64Encoding:);;;Argument[0];ReturnValue;taint",
+        ";NSData;true;base64EncodedData(options:);;;Argument[-1];ReturnValue;taint",
+        ";NSData;true;base64EncodedString(options:);;;Argument[-1];ReturnValue;taint",
+        ";NSData;true;base64Encoding();;;Argument[-1];ReturnValue;taint",
+        ";NSData;true;dataWithContentsOfMappedFile(_:);;;Argument[0];ReturnValue;taint",
+        // TODO: Needs block flow
+        // ";NSData;true;enumerateBytes(_:);;;Argument[-1];Argument[0].Parameter[0];taint"
```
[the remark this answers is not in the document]
atorralba: I think flow through [closures](https://docs.swift.org/swift-book/LanguageGuide/Closures.html) (which I called "blocks" here) is a general data flow feature that should be handled separately (should be equivalent to lambda flow in other languages). Probably related to "flow through `withUnsafeBytes` and similar methods", which we already have an issue for. /cc @MathiasVP in case you have some plans for this already.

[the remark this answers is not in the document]
MathiasVP: I _thought_ we had support for flow-through blocks (see [here](https://github.com/github/codeql/blob/1eec067474c45397a2453925422786c21ca27dac/swift/ql/lib/codeql/swift/dataflow/internal/DataFlowPrivate.qll#L644)). But we might be missing something to connect this to MaD 🤔? The taint source I added [here](https://github.com/github/codeql/pull/10699/files#diff-6a9b9f57af2b6dc32149efa7893028ee8aa080b7f5eae35d3c797d30af604d14R67) convinced me it was working for MaD, though.

swift/ql/test/library-tests/dataflow/taint/nsmutabledata.swift
atorralba: If you're referring to writing to the field with a tainted value, then reading from it and using it in a sink, that should be picked up by normal field flow without requiring a specific model.

If you mean that writing to the field with tainted data should taint the whole object, then we should add some code for that.

In any case, I don't think you can actually write to that field, [it seems to be read-only](https://developer.apple.com/documentation/foundation/nsmutabledata/1410770-mutablebytes) (it only defines a getter, not a setter).

swift/ql/lib/codeql/swift/frameworks/StandardLibrary/NSData.qll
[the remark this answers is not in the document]
atorralba: Huh, I don't know how I previously tested that but it indeed works after uncommenting the CSV row (see 2ac06b8). That's great, we have better MaD support than I thought! Sorry for unnecessarily bothering you @MathiasVP, and thanks for pointing this out @geoffw0 🙏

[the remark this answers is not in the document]
geoffw0: Great, I'm just happy it's working!

swift/ql/test/library-tests/dataflow/taint/nsmutabledata.swift
geoffw0: I meant the latter, but if it's a read-only field we don't need it!

geoffw0 (approved):

swift/ql/test/library-tests/dataflow/taint/TaintInline.expected
```diff
@@ -0,0 +1 @@
+| nsdata.swift:110:45:110:45 | bytes | Unexpected result: tainted=108 |
```
MathiasVP, line 1: We're missing a `$` on the `// tainted=108` comment.

atorralba: 🤦‍♂️ sorry. Fixed here https://github.com/github/codeql/pull/11411.

PR #11383, opened 2022-11-22
C#: Also include extractor unit tests in `csharp-qltest.yml`

hvitved:
These unit tests were part of the old `C# Language Tests` run; so we should include them in the new setup.


michaelnebel (approved):
Looks good to me!

PR #11384, opened 2022-11-22
Adds Kotlin (beta) content

subatoi:
Note that the content design plan noted possible additions to https://codeql.github.com/docs/codeql-cli/extractor-options/; I could not find a suitable place

### DIFF
--- a/docs/codeql/codeql-cli/creating-codeql-databases.rst
+++ b/docs/codeql/codeql-cli/creating-codeql-databases.rst
@@ -168,7 +168,9 @@ generate a database, therefore the build method must be available to the CLI.
 Detecting the build system
 ~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-The CodeQL CLI includes autobuilders for C/C++, C#, Go, and Java code. CodeQL
+.. include:: ../reusables/kotlin-beta-note.rst
+
+The CodeQL CLI includes autobuilders for C/C++, C#, Go, Java and Kotlin code. CodeQL
 autobuilders allow you to build projects for compiled languages without
 specifying any build commands. When an autobuilder is invoked, CodeQL examines
 the source for evidence of a build system and attempts to run the optimal set of
--- a/docs/codeql/codeql-language-guides/abstract-syntax-tree-classes-for-working-with-java-programs.rst
+++ b/docs/codeql/codeql-language-guides/abstract-syntax-tree-classes-for-working-with-java-programs.rst
@@ -7,6 +7,8 @@ CodeQL has a large selection of classes for representing the abstract syntax tre
 
 .. include:: ../reusables/abstract-syntax-tree.rst
 
+.. include:: ../reusables/kotlin-java-differences.rst
+
 Statement classes
 -----------------
 
--- a/docs/codeql/codeql-language-guides/abstract-syntax-tree-classes-for-working-with-java-programs.rst
+++ b/docs/codeql/codeql-language-guides/abstract-syntax-tree-classes-for-working-with-java-programs.rst
@@ -7,6 +7,8 @@ CodeQL has a large selection of classes for representing the abstract syntax tre
 
 .. include:: ../reusables/abstract-syntax-tree.rst
 
+.. include:: ../reusables/kotlin-beta-note.rst
+
 .. include:: ../reusables/kotlin-java-differences.rst
 
 Statement classes
--- a/docs/codeql/codeql-language-guides/analyzing-data-flow-in-java.rst
+++ b/docs/codeql/codeql-language-guides/analyzing-data-flow-in-java.rst
@@ -5,6 +5,8 @@ Analyzing data flow in Java
 
 You can use CodeQL to track the flow of data through a Java program to its use. 
 
+.. include:: ../reusables/kotlin-java-differences.rst
+
 About this article
 ------------------
 
--- a/docs/codeql/codeql-language-guides/analyzing-data-flow-in-java.rst
+++ b/docs/codeql/codeql-language-guides/analyzing-data-flow-in-java.rst
@@ -5,6 +5,8 @@ Analyzing data flow in Java
 
 You can use CodeQL to track the flow of data through a Java program to its use. 
 
+.. include:: ../reusables/kotlin-beta-note.rst
+
 .. include:: ../reusables/kotlin-java-differences.rst
 
 About this article
--- a/docs/codeql/codeql-language-guides/codeql-for-java.rst
+++ b/docs/codeql/codeql-language-guides/codeql-for-java.rst
@@ -1,9 +1,17 @@
 .. _codeql-for-java:
 
-CodeQL for Java
-===============
+CodeQL for Java and Kotlin
+==========================
 
-Experiment and learn how to write effective and efficient queries for CodeQL databases generated from Java codebases.
+Experiment and learn how to write effective and efficient queries for CodeQL databases generated from Java and Kotlin codebases.
+
+.. include:: ../reusables/kotlin-beta-note.rst
+
+.. include:: ../reusables/kotlin-java-differences.rst
+
+.. pull-quote:: Enabling Kotlin support
+
+   To enable Kotlin support, you should enable `java` as a language.
 
 .. toctree::
    :hidden:
--- a/docs/codeql/codeql-language-guides/codeql-for-java.rst
+++ b/docs/codeql/codeql-language-guides/codeql-for-java.rst
@@ -11,7 +11,7 @@ Experiment and learn how to write effective and efficient queries for CodeQL dat
 
 .. pull-quote:: Enabling Kotlin support
 
-   To enable Kotlin support, you should enable `java` as a language.
+   To enable Kotlin support, you should enable ``java`` as a language.
 
 .. toctree::
    :hidden:
--- a/docs/codeql/codeql-language-guides/codeql-for-java.rst
+++ b/docs/codeql/codeql-language-guides/codeql-for-java.rst
@@ -7,7 +7,6 @@ Experiment and learn how to write effective and efficient queries for CodeQL dat
 
 .. include:: ../reusables/kotlin-beta-note.rst
 
-.. include:: ../reusables/kotlin-java-differences.rst
 
 .. pull-quote:: Enabling Kotlin support
 
--- a/docs/codeql/codeql-language-guides/codeql-for-java.rst
+++ b/docs/codeql/codeql-language-guides/codeql-for-java.rst
@@ -10,7 +10,7 @@ Experiment and learn how to write effective and efficient queries for CodeQL dat
 
 .. pull-quote:: Enabling Kotlin support
 
-   To enable Kotlin support, you should enable ``java`` as a language.
+   CodeQL treats Java and Kotlin as parts of the same language, so to enable Kotlin support you should enable ``java`` as a language.
 
 .. toctree::
    :hidden:
--- a/docs/codeql/reusables/kotlin-java-differences.rst
+++ b/docs/codeql/reusables/kotlin-java-differences.rst
@@ -1,5 +1,5 @@
-Writing CodeQL queries in Kotlin versus Java
---------------------------------------------
+Writing CodeQL queries for Kotlin versus Java analysis
+------------------------------------------------------
 
 When writing Kotlin-specific elements (such as a `WhenExpr`) you’ll need to use Kotlin-specific CodeQL classes, but writing queries for Kotlin and Java is largely the same. The two make use of the same libraries such as DataFlow, TaintTracking, or SSA, and the same classes such as `MethodAccess` or `Class`. 
 
--- a/docs/codeql/reusables/kotlin-java-differences.rst
+++ b/docs/codeql/reusables/kotlin-java-differences.rst
@@ -5,7 +5,7 @@ When writing Kotlin-specific elements (such as a `WhenExpr`) you’ll need to us
 
 There are however some important cases where writing queries for Kotlin can produce surprising results compared to writing Java queries, as CodeQL works with the JVM bytecode representation of the source code. 
 
-Be careful when trying to model code elements that don’t exist in Java, such as `NotNullExpr (expr!!)`, because they could interact in unexpected ways with common predicates. For example, `MethodAccess.getQualifier()` gets a `NotNullExpr `instead of a `VarAccess`` in the following Kotlin code:
+Be careful when you model code elements that don’t exist in Java, such as ``NotNullExpr (expr!!)``, because they could interact in unexpected ways with common predicates. For example, ``MethodAccess.getQualifier()`` returns a ``NotNullExpr`` instead of a ``VarAccess`` in the following Kotlin code:
 
 .. code-block:: kotlin
    someVar!!.someMethodCall()
--- a/docs/codeql/reusables/kotlin-java-differences.rst
+++ b/docs/codeql/reusables/kotlin-java-differences.rst
@@ -8,6 +8,7 @@ There are however some important cases where writing queries for Kotlin can prod
 Be careful when you model code elements that don’t exist in Java, such as ``NotNullExpr (expr!!)``, because they could interact in unexpected ways with common predicates. For example, ``MethodAccess.getQualifier()`` returns a ``NotNullExpr`` instead of a ``VarAccess`` in the following Kotlin code:
 
 .. code-block:: kotlin
+   
    someVar!!.someMethodCall()
 
 In that specific case, you can use the predicate `Expr.getUnderlyingExpr()`. This goes directly to the underlying `VarAccess`` to produce a more similar behavior to that in Java.
--- a/docs/codeql/reusables/kotlin-java-differences.rst
+++ b/docs/codeql/reusables/kotlin-java-differences.rst
@@ -16,4 +16,4 @@ Nullable elements (`?`) can also produce unexpected behavior. To avoid a `NullPo
 
 Another example is that if-else expressions are translated into `WhenExprs` in CodeQL, instead of the more typical `IfStmt` in Java.
 
-In general, you can debug these issues with the AST (you can use the `CodeQL: View AST`` command from Visual Studio Code’s CodeQL extension, or run the `PrintAst.ql`` query) and checking what exactly CodeQL is extracting from your code.
+In general, you can debug these issues with the AST (you can use the ``CodeQL: View AST`` command from Visual Studio Code’s CodeQL extension, or run the ``PrintAst.ql`` query) and see exactly what CodeQL is extracting from your code.
--- a/docs/codeql/reusables/kotlin-java-differences.rst
+++ b/docs/codeql/reusables/kotlin-java-differences.rst
@@ -11,7 +11,7 @@ Be careful when you model code elements that don’t exist in Java, such as ``No
    
    someVar!!.someMethodCall()
 
-In that specific case, you can use the predicate `Expr.getUnderlyingExpr()`. This goes directly to the underlying `VarAccess`` to produce a more similar behavior to that in Java.
+In that specific case, you can use the predicate ``Expr.getUnderlyingExpr()``. This goes directly to the underlying ``VarAccess`` to produce a more similar behavior to that in Java.
 
 Nullable elements (`?`) can also produce unexpected behavior. To avoid a `NullPointerException`, Kotlin may inline calls like `expr.toString()` to `String.valueOf(expr)` when `expr` is nullable. Make sure that you write CodeQL around the extracted code, and do not directly modify the source code in the codebase.
 
--- a/docs/codeql/reusables/kotlin-java-differences.rst
+++ b/docs/codeql/reusables/kotlin-java-differences.rst
@@ -1,7 +1,7 @@
 Writing CodeQL queries for Kotlin versus Java analysis
 ------------------------------------------------------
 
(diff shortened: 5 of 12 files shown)


docs/codeql/codeql-cli/creating-codeql-databases.rst
```diff
@@ -168,7 +168,9 @@ generate a database, therefore the build method must be available to the CLI.
 Detecting the build system
 ~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-The CodeQL CLI includes autobuilders for C/C++, C#, Go, and Java code. CodeQL
+.. include:: ../reusables/kotlin-beta-note.rst
+
+The CodeQL CLI includes autobuilders for C/C++, C#, Go, Java and Kotlin code. CodeQL
```
subatoi, line 173: Do we also need an example build command for Kotlin for this section?

docs/codeql/reusables/kotlin-java-differences.rst
```diff
@@ -0,0 +1,18 @@
+Writing CodeQL queries in Kotlin versus Java
+--------------------------------------------
+
+When writing Kotlin-specific elements (such as a `WhenExpr`) you’ll need to use Kotlin-specific CodeQL classes, but writing queries for Kotlin and Java is largely the same. The two make use of the same libraries such as DataFlow, TaintTracking, or SSA, and the same classes such as `MethodAccess` or `Class`.
```
subatoi, line 4: Just to check: we shouldn't mention any specific libraries for Kotlin?

docs/codeql/codeql-language-guides/analyzing-data-flow-in-java.rst
```diff
@@ -5,6 +5,8 @@ Analyzing data flow in Java
 
 You can use CodeQL to track the flow of data through a Java program to its use. 
 
+.. include:: ../reusables/kotlin-java-differences.rst
```
subatoi, line 8:
```suggestion
.. include:: ../reusables/kotlin-beta-note.rst

.. include:: ../reusables/kotlin-java-differences.rst
```

docs/codeql/codeql-language-guides/abstract-syntax-tree-classes-for-working-with-java-programs.rst
```diff
@@ -7,6 +7,8 @@ CodeQL has a large selection of classes for representing the abstract syntax tre
 
 .. include:: ../reusables/abstract-syntax-tree.rst
 
+.. include:: ../reusables/kotlin-java-differences.rst
```
subatoi, line 10:
```suggestion
.. include:: ../reusables/kotlin-beta-note.rst

.. include:: ../reusables/kotlin-java-differences.rst
```

docs/codeql/codeql-language-guides/codeql-for-java.rst
```diff
@@ -1,9 +1,17 @@
...
-Experiment and learn how to write effective and efficient queries for CodeQL databases generated from Java codebases.
+Experiment and learn how to write effective and efficient queries for CodeQL databases generated from Java and Kotlin codebases.
+
+.. include:: ../reusables/kotlin-beta-note.rst
+
+.. include:: ../reusables/kotlin-java-differences.rst
+
+.. pull-quote:: Enabling Kotlin support
+
+   To enable Kotlin support, you should enable `java` as a language.
```
intrigus-lgtm, line 14: Is this correct?
If yes, it sounds a bit unintuitive that you have to enable `java` to enable `Kotlin` IMHO.

intrigus-lgtm (commented):
Not an official committer, so feel free to ignore my review.

docs/codeql/codeql-language-guides/codeql-for-java.rst
igfoo: Yes: Due to how common it is for Kotlin code to call Java code and vice versa within a single project, we treat them as essentially 2 compilers for the same language. This also means that the CodeQL libraries and queries can be shared between the two which, as they are very similar underneath, saves a lot of duplication. This is similar to how javascript and typescript are treated by CodeQL as a single language, as well as the C family (standard C and C++ in their various versions, as well as Microsoft and GNU extensions to those languages).

docs/codeql/reusables/kotlin-java-differences.rst
```diff
@@ -0,0 +1,18 @@
...
+
+`someVar!!.someMethodCall()`
+
+In that specific case, you can use the predicate `Expr.getUnderlyingExpr()`. This goes directly to the underlying `VarAccess`` to produce a more similar behavior to that in Java.
+
+Nullable elements (`?`) can also produce unexpected behavior. To avoid a `NullPointerException`, Kotlin may inline calls like `expr.toString()` to `String.valueOf(expr)` when `expr` is nullable. Make sure that you write CodeQL around the extracted code, and do not directly modify the source code in the codebase.
+
+Another example is that if-else expressions are translated into `WhenExprs` in CodeQL, instead of the more typical `IfStmt` in Java.
+
+In general, you can debug these issues with the AST (you can use the `CodeQL: View AST`` command from Visual Studio Code’s CodeQL extension, or run the `PrintAst.ql`` query) and checking what exactly CodeQL is extracting from your code.
```
felicitymay, line 18: Add some missing backticks and a minor text change.

```suggestion
In general, you can debug these issues with the AST (you can use the ``CodeQL: View AST`` command from Visual Studio Code’s CodeQL extension, or run the ``PrintAst.ql`` query) and see exactly what CodeQL is extracting from your code.
```

docs/codeql/reusables/kotlin-java-differences.rst
felicitymay, line 4: Suggest leading with the fact that writing queries for the two languages is mostly the same. Maybe:

```suggestion
Generally you use the same classes to write queries for Kotlin and for Java. You use the same libraries such as DataFlow, TaintTracking, or SSA, and the same classes such as `MethodAccess` or `Class` for both languages. When you want to access Kotlin-specific elements (such as a ``WhenExpr``) you’ll need to use Kotlin-specific CodeQL classes.
```

docs/codeql/reusables/kotlin-java-differences.rst
```diff
@@ -0,0 +1,18 @@
+Writing CodeQL queries in Kotlin versus Java
+--------------------------------------------
+
+When writing Kotlin-specific elements (such as a `WhenExpr`) you’ll need to use Kotlin-specific CodeQL classes, but writing queries for Kotlin and Java is largely the same. The two make use of the same libraries such as DataFlow, TaintTracking, or SSA, and the same classes such as `MethodAccess` or `Class`. 
+
+There are however some important cases where writing queries for Kotlin can produce surprising results compared to writing Java queries, as CodeQL works with the JVM bytecode representation of the source code.
```
felicitymay, line 6:
```suggestion
There are however some important cases where writing queries for Kotlin can produce surprising results compared to writing queries for Java, as CodeQL works with the JVM bytecode representation of the Kotlin source code. 
```

docs/codeql/reusables/kotlin-java-differences.rst
```diff
@@ -0,0 +1,18 @@
+Writing CodeQL queries in Kotlin versus Java
+--------------------------------------------
+
+When writing Kotlin-specific elements (such as a `WhenExpr`) you’ll need to use Kotlin-specific CodeQL classes, but writing queries for Kotlin and Java is largely the same. The two make use of the same libraries such as DataFlow, TaintTracking, or SSA, and the same classes such as `MethodAccess` or `Class`. 
+
+There are however some important cases where writing queries for Kotlin can produce surprising results compared to writing Java queries, as CodeQL works with the JVM bytecode representation of the source code. 
+
+Be careful when trying to model code elements that don’t exist in Java, such as `NotNullExpr (expr!!)`, because they could interact in unexpected ways with common predicates. For example, `MethodAccess.getQualifier()` gets a `NotNullExpr `instead of a `VarAccess`` in the following Kotlin code:
```
felicitymay, line 8: I think we should be more positive here:

```suggestion
Be careful when you model code elements that don’t exist in Java, such as ``NotNullExpr (expr!!)``, because they could interact in unexpected ways with common predicates. For example, ``MethodAccess.getQualifier()`` returns a ``NotNullExpr`` instead of a ``VarAccess`` in the following Kotlin code:
```

docs/codeql/reusables/kotlin-java-differences.rst
```diff
@@ -0,0 +1,18 @@
...
+
+`someVar!!.someMethodCall()`
```
felicitymay, line 10: If this is Kotlin code, it should perhaps be in a code block:

```suggestion
.. code-block:: kotlin
   someVar!!.someMethodCall()
```

docs/codeql/reusables/kotlin-java-differences.rst
```diff
@@ -0,0 +1,18 @@
+Writing CodeQL queries in Kotlin versus Java
+--------------------------------------------
```
felicitymay, line 2:
```suggestion
Writing CodeQL queries for Kotlin versus Java analysis
------------------------------------------------------
```

docs/codeql/codeql-language-guides/codeql-for-java.rst
felicitymay, line 14:
```suggestion
   To enable Kotlin support, you should enable ``java`` as a language.
```

docs/codeql/codeql-language-guides/codeql-for-java.rst
```diff
@@ -1,9 +1,17 @@
...
-===============
+CodeQL for Java and Kotlin
+==========================
 
-Experiment and learn how to write effective and efficient queries for CodeQL databases generated from Java codebases.
+Experiment and learn how to write effective and efficient queries for CodeQL databases generated from Java and Kotlin codebases.
+
+.. include:: ../reusables/kotlin-beta-note.rst
+
+.. include:: ../reusables/kotlin-java-differences.rst
```
felicitymay, line 10: This note ended up longer than I expected. I'm wondering if it's needed on this page 🤔 

If we're going to include it, I'd tend to put both it and the note about enabling Kotlin support after the list of articles.

felicitymay: I'm also wondering if including it here is somehow causing the heading to appear in the TOC for the parent language guides article:
image

I haven't seen this behavior before and my Sphinx knowledge is rather rusty. Happy to experiment locally or to look at this together.

felicitymay (commented):
@subatoi - thank you for pitching in and getting these changes sorted out so quickly 💖 

Generally this looks good to me. I've added a few comments but nothing major.

One difference between RST and MD which trips us all up is the use of double backticks for monospaced font. A single backtick appears as italic text. I've fixed some of the backtick errors where I was commenting anyway, but it's worth checking for any that I missed.

docs/codeql/codeql-language-guides/codeql-for-java.rst
subatoi: Hmm, I'm not sure. Maybe it'd be easier to just take it out, per your first comment? 🤔

docs/codeql/codeql-cli/creating-codeql-databases.rst
subatoi: No probs!

> I guess we could update "Java project built using Gradle" to "Java and Kotlin project built using Gradle".

This seems like a possibility, but possibly not needed?

felicitymay: > This seems like a possibility, but possibly not needed?

I agree. I think it's fine to leave it as is.

docs/codeql/codeql-language-guides/codeql-for-java.rst
felicitymay: Sounds like a good plan

docs/codeql/reusables/kotlin-java-differences.rst
```diff
@@ -0,0 +1,20 @@
+Writing CodeQL queries for Kotlin versus Java analysis
+------------------------------------------------------
+
+Generally you use the same classes to write queries for Kotlin and for Java. You use the same libraries such as DataFlow, TaintTracking, or SSA, and the same classes such as `MethodAccess` or `Class` for both languages. When you want to access Kotlin-specific elements (such as a ``WhenExpr``) you’ll need to use Kotlin-specific CodeQL classes.
```
felicitymay, line 4:
```suggestion
Generally you use the same classes to write queries for Kotlin and for Java. You use the same libraries such as DataFlow, TaintTracking, or SSA, and the same classes such as ``MethodAccess`` or ``Class`` for both languages. When you want to access Kotlin-specific elements (such as a ``WhenExpr``) you’ll need to use Kotlin-specific CodeQL classes.
```

docs/codeql/reusables/kotlin-java-differences.rst
```diff
@@ -0,0 +1,20 @@
...
+
+There are however some important cases where writing queries for Kotlin can produce surprising results compared to writing queries for Java, as CodeQL works with the JVM bytecode representation of the Kotlin source code. 
+
+Be careful when you model code elements that don’t exist in Java, such as ``NotNullExpr (expr!!)``, because they could interact in unexpected ways with common predicates. For example, ``MethodAccess.getQualifier()`` returns a ``NotNullExpr`` instead of a ``VarAccess`` in the following Kotlin code:
+
+.. code-block:: kotlin
+   
+   someVar!!.someMethodCall()
+
+In that specific case, you can use the predicate `Expr.getUnderlyingExpr()`. This goes directly to the underlying `VarAccess`` to produce a more similar behavior to that in Java.
```
felicitymay, line 14:
```suggestion
In that specific case, you can use the predicate ``Expr.getUnderlyingExpr()``. This goes directly to the underlying ``VarAccess`` to produce a more similar behavior to that in Java.
```

docs/codeql/reusables/kotlin-java-differences.rst
```diff
@@ -0,0 +1,20 @@
...
+
+Nullable elements (`?`) can also produce unexpected behavior. To avoid a `NullPointerException`, Kotlin may inline calls like `expr.toString()` to `String.valueOf(expr)` when `expr` is nullable. Make sure that you write CodeQL around the extracted code, and do not directly modify the source code in the codebase.
+
+Another example is that if-else expressions are translated into `WhenExprs` in CodeQL, instead of the more typical `IfStmt` in Java.
```
felicitymay, line 18:
```suggestion
Nullable elements (``?``) can also produce unexpected behavior. To avoid a ``NullPointerException``, Kotlin may inline calls like ``expr.toString()`` to ``String.valueOf(expr)`` when ``expr`` is nullable. Make sure that you write CodeQL around the extracted code, and do not directly modify the source code in the codebase.

Another example is that if-else expressions in Kotlin are translated into ``WhenExprs`` in CodeQL, instead of the more typical ``IfStmt`` seen in Java.
```

felicitymay (approved):
I spotted a few more issues with backticks, otherwise this looks fine from my point of view.

docs/codeql/codeql-language-guides/codeql-for-java.rst
subatoi, line 10:
```suggestion
```

docs/codeql/codeql-language-guides/codeql-for-java.rst
```diff
@@ -1,9 +1,16 @@
...
 
-Experiment and learn how to write effective and efficient queries for CodeQL databases generated from Java codebases.
+Experiment and learn how to write effective and efficient queries for CodeQL databases generated from Java and Kotlin codebases.
+
+.. include:: ../reusables/kotlin-beta-note.rst
+
+
+.. pull-quote:: Enabling Kotlin support
+
+   To enable Kotlin support, you should enable ``java`` as a language.
```
igfoo, line 13: I think it would be worth expanding on this, e.g.:
```
CodeQL treats Java and Kotlin as parts of the same language, so to enable Kotlin support you should enable ``java`` as a language.

File: docs/codeql/reusables/kotlin-java-differences.rst [hunk 16]
```diff
@@ -0,0 +1,20 @@
...
+
+Be careful when you model code elements that don’t exist in Java, such as ``NotNullExpr (expr!!)``, because they could interact in unexpected ways with common predicates. For example, ``MethodAccess.getQualifier()`` returns a ``NotNullExpr`` instead of a ``VarAccess`` in the following Kotlin code:
+
+.. code-block:: kotlin
+   
+   someVar!!.someMethodCall()
+
+In that specific case, you can use the predicate ``Expr.getUnderlyingExpr()``. This goes directly to the underlying ``VarAccess`` to produce a more similar behavior to that in Java.
+
+Nullable elements (``?``) can also produce unexpected behavior. To avoid a ``NullPointerException``, Kotlin may inline calls like ``expr.toString()`` to ``String.valueOf(expr)`` when ``expr`` is nullable. Make sure that you write CodeQL around the extracted code, and do not directly modify the source code in the codebase.
```
igfoo, line 16: I think the last clause is meant to say something like `which may not exactly match the code as written in the codebase.`?

docs/codeql/codeql-language-guides/codeql-for-java.rst
subatoi: Thanks @igfoo I've added this and your other comment in a single commit 👍

docs/codeql/reusables/kotlin-java-differences.rst
```diff
@@ -0,0 +1,20 @@
...
+
+Be careful when you model code elements that don’t exist in Java, such as ``NotNullExpr (expr!!)``, because they could interact in unexpected ways with common predicates. For example, ``MethodAccess.getQualifier()`` returns a ``NotNullExpr`` instead of a ``VarAccess`` in the following Kotlin code:
+
+.. code-block:: kotlin
+   
+   someVar!!.someMethodCall()
+
+In that specific case, you can use the predicate ``Expr.getUnderlyingExpr()``. This goes directly to the underlying ``VarAccess`` to produce a more similar behavior to that in Java.
+
+Nullable elements (`?`) can also produce unexpected behavior. To avoid a `NullPointerException`, Kotlin may inline calls like `expr.toString()` to `String.valueOf(expr)` when `expr` is nullable. Make sure that you write CodeQL around the extracted code, which may not exactly match the code as written in the codebase.
```
felicitymay, line 16: @subatoi - it looks as if you may have an extension that's removing the double backticks because they've vanished again from this paragraph: https://github.com/github/codeql/pull/11384/commits/357c823b92b9f2dc7d2f16db1f523c1f269bdd53#diff-8e08ecaf9b0928a502efdea7c357815043c20b9d0e11aa9a303b7ed07ddf2267.

subatoi:
```suggestion
Nullable elements (``?``) can also produce unexpected behavior. To avoid a ``NullPointerException``, Kotlin may inline calls like ``expr.toString()`` to ``String.valueOf(expr)`` when ``expr`` is nullable. Make sure that you write CodeQL around the extracted code, which may not exactly match the code as written in the codebase.
```

docs/codeql/reusables/kotlin-java-differences.rst
```diff
@@ -0,0 +1,20 @@
...
+
+Be careful when you model code elements that don’t exist in Java, such as ``NotNullExpr (expr!!)``, because they could interact in unexpected ways with common predicates. For example, ``MethodAccess.getQualifier()`` returns a ``NotNullExpr`` instead of a ``VarAccess`` in the following Kotlin code:
+
+.. code-block:: kotlin
+   
+   someVar!!.someMethodCall()
+
+In that specific case, you can use the predicate ``Expr.getUnderlyingExpr()``. This goes directly to the underlying ``VarAccess`` to produce a more similar behavior to that in Java.
+
+Nullable elements (``?``) can also produce unexpected behavior. To avoid a ``NullPointerException``, Kotlin may inline calls like ``expr.toString()`` to ``String.valueOf(expr)`` when ``expr`` is nullable. Make sure that you write CodeQL around the extracted code, which may not exactly match the code as written in the codebase.
```
subatoi, line 16: Good catch with this one @felicitymay I'll have a look into that 👍

docs/codeql/reusables/kotlin-java-differences.rst
```diff
@@ -0,0 +1,20 @@
...
+
+There are however some important cases where writing queries for Kotlin can produce surprising results compared to writing queries for Java, as CodeQL works with the JVM bytecode representation of the Kotlin source code. 
+
+Be careful when you model code elements that don’t exist in Java, such as ``NotNullExpr (expr!!)``, because they could interact in unexpected ways with common predicates. For example, ``MethodAccess.getQualifier()`` returns a ``NotNullExpr`` instead of a ``VarAccess`` in the following Kotlin code:
+
+.. code-block:: kotlin
+   
+   someVar!!.someMethodCall()
+
+In that specific case, you can use the predicate ``Expr.getUnderlyingExpr()``. This goes directly to the underlying ``VarAccess`` to produce a more similar behavior to that in Java.
```
atorralba, line 14: > to produce a more similar behavior to that in Java.

The message here was that if you want to replicate the Java behavior you're used to when calling `getQualifier()` in Kotlin, you should call `getUnderlyingExpr()` on the result. Not sure if the current phrasing reflects that.

PR #11386, opened 2022-11-23
Ruby/QL: only create dbscheme case-splits for columns on defining tables

nickrolfe:
Dbscheme case-splits are only valid for `int` columns on defining tables.

This fix doesn't actually affect Ruby/QL (so the dbschemes haven't changed), but for other tree-sitter grammars it could generate invalid dbschemes by creating case-splits for columns in auxiliary tables.

### DIFF
--- a/ql/node-types/src/lib.rs
+++ b/ql/node-types/src/lib.rs
@@ -81,6 +81,14 @@ pub enum Storage {
     },
 }
 
+impl Storage {
+    pub fn is_column(&self) -> bool {
+        match self {
+            Storage::Column { .. } => true,
+            _ => false,
+        }
+    }
+}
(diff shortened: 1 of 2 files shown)


aibaars:
This seems to work on https://github.com/JoranHonig/tree-sitter-solidity . @nickrolfe Have you checked the differences to the dbscheme and QL library for some sample grammars?

aibaars (approved):
Looks good to me. It seems to do the right thing on one example I tried and the changes look sensible to me.

erik-krogh (approved):
QL-for-QL 👍  (because nothing changed).

nickrolfe:
Yes, I checked that it did the right thing (and previously did the wrong thing) for the C# grammar.

PR #11387, opened 2022-11-23
C#: Split `AutobuildOptions` into C#/C++ specific classes

hvitved:
I wanted to include the C++ Windows Autobuilder tests in the CI PR check. It turned out that those tests are not working, and I realized that we need to distinguish between C# autobuilder options and C++ autobuilder options (the C++ tests were failing due to a missing `CODEQL_EXTRACTOR_CSHARP_OPTION_BUILDLESS` environment variable, which has nothing to do with C++).

So this PR make such a splitting, and adds the appropriate `dotnet test` CI invocation.

### DIFF
--- a/csharp/autobuilder/Semmle.Autobuild.Shared/BuildCommandAutoRule.cs
+++ b/csharp/autobuilder/Semmle.Autobuild.Shared/BuildCommandAutoRule.cs
@@ -7,11 +7,11 @@ namespace Semmle.Autobuild.Shared
     /// <summary>
     /// Auto-detection of build scripts.
     /// </summary>
-    public class BuildCommandAutoRule : IBuildRule
+    public class BuildCommandAutoRule : IBuildRule<AutobuildOptionsShared>
     {
-        private readonly WithDotNet withDotNet;
+        private readonly WithDotNet<AutobuildOptionsShared> withDotNet;
 
-        public BuildCommandAutoRule(WithDotNet withDotNet)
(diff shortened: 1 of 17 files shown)


michaelnebel (approved):
LGTM

csharp/autobuilder/Semmle.Autobuild.Shared/BuildCommandAutoRule.cs
```diff
@@ -7,11 +7,11 @@ namespace Semmle.Autobuild.Shared
     /// <summary>
     /// Auto-detection of build scripts.
     /// </summary>
-    public class BuildCommandAutoRule : IBuildRule
+    public class BuildCommandAutoRule : IBuildRule<AutobuildOptionsShared>
```
aeisenberg, line 10: I'm not really familiar with C#, but in java we'd do something like this since I think we want to handle subclasses of `AutobuildOptionsShared`:

```suggestion
    public class BuildCommandAutoRule<T> : IBuildRule<T extends AutobuildOptionsShared>
```

This is more a question than a suggestion.

aeisenberg (approved):
My question is more out of ignorance than anything else. So, this is probably fine to be merged. The CI part looks fine.

csharp/autobuilder/Semmle.Autobuild.Shared/BuildCommandAutoRule.cs
hvitved: In this case `BuildCommandAutoRule` only needs access to the common options, so there is no need to make it more general.

PR #11389, opened 2022-11-23
Java: Add JDK sinks

tamasvajk:
This PR is adding sinks that were generated on JDK databases. I generated the results by running the below command, where `[JDK_DBs_FOLDER]` is the path to a folder containing JDK databases `jdk7`, `jdk8`, ... `jdk18`.
```
python3 java/ql/lib/ext/generated/jdk/generate.py [JDK_DBs_FOLDER]
```

Some remarks on the generated sinks:
- The `create-file` sinks are removed from [the output](https://github.com/github/codeql/pull/11389/files#diff-7de22582d24c76d8a93c3d236d53779435d0dabf57443ac33fd5b3804aaa38a0). 
- The generated sinks contained paths that were incorrect or not exploitable, these sinks were added to [an exclude list](https://github.com/github/codeql/pull/11389/files#diff-5d994ab1b06b483be058bc8d558228e7b03128cabdf218ffcf2f7754ab53dfa8), so that they're removed from `java/ql/lib/ext/generated/jdk/jdk[X].generated.notincluded.yml` on subsequent runs. 
- Questionable sinks were also excluded with the same mechanism. We should investigate, rework, and add these.
- The sinks are using the new data extension format. 
- Generated files have the suffix `generated.notincluded.yml`, these are not matching extension data patterns, thus not included in the dataflow. 
- Diff files (`java/ql/lib/ext/generated/jdk/jdk[X].diff.model.yml`) are computed and included, and 
- the initial JDK7 model (`java/ql/lib/ext/generated/jdk/jdk7.generated.model.yml`) is also included.

Additionally a [debug path-problem query](https://github.com/github/codeql/pull/11389/files#diff-3ea7a15b006ab8670dfe213a94000d5db3efd2f4e62ff9920daa65b400f70f46) is also added to make it easier to investigate the discovered paths that resulted in the generated sinks.


java/ql/src/utils/model-generator/SinkFlows.ql
```diff
@@ -0,0 +1,44 @@
+/**
+ * @name Capture sink models.
+ * @description Finds public methods that act as sinks as they flow into a known sink.
+ * @kind path-problem
+ * @id java/utils/model-generator/sink-models
+ * @tags model-generator
```
tamasvajk, line 6: These metadata were not changed from the capture sink query.

java/ql/src/utils/model-generator/SinkFlows.ql
```diff
@@ -0,0 +1,40 @@
...
+
+class PropagateToSinkConfiguration extends CaptureModels::PropagateToSinkConfiguration {
+  PropagateToSinkConfiguration() { this = "parameters or fields flowing into sinks" }
+
+  override predicate sinkGrouping(DataFlow::Node sink, string sinkGroup) {
+    ExternalFlow::sinkNode(sink, sinkGroup)
+  }
+}
+
+from
```
michaelnebel, line 29: Is this query for debugging purposes?

java/ql/src/utils/model-generator/CaptureSinkModels.ql
```diff
@@ -7,11 +7,13 @@
...
+import Excludes.Sinks
 
 class Activate extends ActiveConfiguration {
   override predicate activateToSinkConfig() { any() }
 }
 
-from DataFlowTargetApi api, string sink
-where sink = captureSink(api)
+from DataFlowTargetApi api, string kind, string sink
+where sink = captureSink(api, kind) and not hasSink(api, kind, false) and not isExcludedSink(sink)
```
michaelnebel, line 18: Yes, this is exactly what I had in mind!

java/ql/src/utils/model-generator/SinkFlows.ql
tamasvajk: Yes it is. Do you think we should move it somewhere else? I have no clue what side effects it might have if we have it in `java/ql/src`.

tamasvajk: I guess I could move it to `ql/java/ql/test/utils/model-generator/`

michaelnebel: It should be fine to keep it here, we just need to make sure that it is excluded from the query suites, which should be the case when the `model-generator` tag is used.
In any case, maybe we should consider 
- Introducing a new tag named `debug` and also tag the query with that to signal that it is *only* used for debugging purposes (this requires that we add it to the exclusion list in `security-and-quality-selectors.yml` AND the other filters).
- Make a debug folder in the model-generator folder and put the queries there.

Also, we might get an issue with these queries (I have a draft PR somewhere doing approximately the same as you are doing here, but this might not be compatible with the field summary generation).

tamasvajk: I pushed some changes with the `debug` tag. (I also squashed and rebased the PR to fix merge conflicts) Let me know if something is missing.

michaelnebel: That part looks good to me!

michaelnebel:
I just merged https://github.com/github/codeql/pull/11243
It would be great, if you could use data extensions for JDK generated models.
Maybe it is worth considering to use data extensions for the exclusion list as well.

PR #11390, opened 2022-11-23
ATM: Add descriptions to ML-powered packs

henrymercer:
This will allow us to populate a description in the UI (example: https://github.com/orgs/codeql/packages/container/javascript-experimental-atm-queries/49184568?tag=0.4.0), making it easier to see at a glance what the pack does.  The UI will be updated as of the next release of the pack.

### DIFF
--- a/javascript/ql/experimental/adaptivethreatmodeling/lib/qlpack.yml
+++ b/javascript/ql/experimental/adaptivethreatmodeling/lib/qlpack.yml
@@ -1,4 +1,5 @@
 name: codeql/javascript-experimental-atm-lib
+description: CodeQL libraries for the experimental ML-powered queries
 version: 0.4.2
 extractor: javascript
 library: true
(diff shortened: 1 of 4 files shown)


javascript/ql/experimental/adaptivethreatmodeling/lib/qlpack.yml
```diff
@@ -1,4 +1,5 @@
 name: codeql/javascript-experimental-atm-lib
+description: CodeQL libraries for the experimental ML-powered queries
```
aeisenberg, line 2: Do we want to include the term "Adaptive Threat Modeling (ATM)" in all of these descriptions?

Also, I wonder if the descriptions support html (or at least a subset of it).

henrymercer: My understanding's that ATM is an internal-only term, and the customer-facing term is ML-powered queries.  Admittedly we're a little inconsistent about this given that we include "atm" in the pack name.

aeisenberg (approved):

PR #11395, opened 2022-11-23
C#: ExternalFlow.qll cleanup.

michaelnebel:
In this PR we delete the bi-directional import of `ExternalFlow` that was needed to include all models defined in the framework specific files.
It turned out that some other parts of the code relied on the SCC that was generated from the bi-directional import. Most cases were fixed on https://github.com/github/codeql/pull/10777 and this PR fixes the remaining one and deletes all the un-needed import code.


michaelnebel:
Will trigger DCA to see, if we experience any performance issues by un-tying the SCC knot.

michaelnebel:
DCA Looks good!

hvitved (approved):
Nice!

PR #11397, opened 2022-11-23
Rb: use `instanceof` instead of `extends` on `DataFlow::CallNode` in some case

erik-krogh:
I noticed an `override DataFlow::Node getAnArgument` in classes that `extends DataFlow::CallNode`.   
That is just a foot-gun waiting to happen if `DataFlow::CallNode` gets an `getAnArgument` predicate, so I changed those classes to `instanceof DataFlow::CallNode`.  

Drive-by cleanup of imports in `Open3.qll` based on feedback from QL-for-QL.


hvitved (approved):
LGTM. Let's do a DCA run before merging.

erik-krogh:
[Evaluation looks good. ](https://github.com/github/codeql-dca-main/tree/data/erik-krogh/PR-11397-0-ruby/reports)

PR #11398, opened 2022-11-23
Rb: add some more flow through splat parameters 

erik-krogh:
This was something I needed in my [second-order-command-injection WIP branch](https://github.com/github/codeql/pull/11236).  

Precise flow through splat parameters is hard in the general case.  
But if we special-case to the situration where both the argument and the parameter are in the first position, then it's easy.  

[Evaluation looks OK](https://github.com/github/codeql-dca-main/tree/data/erik-krogh/pr-11398-c95432__nightly__code-scanning/reports).  
The easiest way to see the impact is to look at the new call-edges. Those new edges mostly appear from the receiver being tracked more precisely.  
E.g. [here](https://github.com/github/codeql-dca-main/blob/data/erik-krogh/pr-11398-c95432__nightly__code-scanning/reports/alert-meta-comparison.md#-rbmetacall-graph-for-fastlane__fastlane-2) where `args` is a splat parameter that is tracked more precisely with this change.  

This is my first venture into the inner workings of the dataflow library, so I hope I got it right.

### DIFF
--- a/ruby/ql/test/library-tests/dataflow/params/params_flow.rb
+++ b/ruby/ql/test/library-tests/dataflow/params/params_flow.rb
@@ -61,6 +61,6 @@ posargs(*args)
 
 args = taint(26)
 def splatstuff(*x)
-    sink x[0] # $ MISSING: hasValueFlow=26
+    sink x[0] # $ hasValueFlow=26
 end
 splatstuff(*args)
--- a/ruby/ql/test/library-tests/dataflow/params/params_flow.rb
+++ b/ruby/ql/test/library-tests/dataflow/params/params_flow.rb
@@ -58,3 +58,9 @@ posargs(taint(23), *args)
 
 args = [taint(24), taint(25)]
 posargs(*args)
+
+args = taint(26)
+def splatstuff(*x)
+    sink x[0] # $ MISSING: hasValueFlow=26
+end
+splatstuff(*args)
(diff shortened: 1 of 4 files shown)


ruby/ql/test/library-tests/dataflow/params/params_flow.rb
```diff
@@ -57,4 +57,10 @@ def posargs(p1, *posargs)
 posargs(taint(23), *args)
 
 args = [taint(24), taint(25)]
-posargs(*args)
\ No newline at end of file
+posargs(*args)
+
+args = taint(26)
+def splatstuff(*x)
+    sink x # $ hasValueFlow=26
```
hvitved, line 64: I don't think we want flow in this case; we only want flow to `x[n]`.

hvitved (commented):
I am not convinced that this adds the type of flow that we want; I think it would be better to add support a la how we do it for hash splats.

ruby/ql/test/library-tests/dataflow/params/params_flow.rb
erik-krogh: I've made it similar to how we handle hash-splat, except I still restrict it to the first argument/parameter.  

It still gives me the flow I want in [the second-order-command-injection PR](https://github.com/github/codeql/pull/11236).

hvitved (approved):
LGTM, good to have some initial support at least for splat parameters. Will you do a final DCA run?

erik-krogh:
> LGTM, good to have some initial support at least for splat parameters. Will you do a final DCA run?

[New evaluation](https://github.com/github/codeql-dca-main/tree/data/erik-krogh/pr-11398-0c2ff9__nightly__code-scanning/reports) looks about the same as the old evaluation.   
Some new call-edges, and a few new results that look good.   
I'm also running on `rails-projects.yml` to be sure.

PR #11400, opened 2022-11-23
C++: Fix CWE-611 XXE query to work with use-use dataflow

jketema:
I'm not completely sure if this is the optimal fix, so I happily discuss that. I does restore all the missing results.

### DIFF
--- a/cpp/ql/src/Security/CWE/CWE-611/Xerces.qll
+++ b/cpp/ql/src/Security/CWE/CWE-611/Xerces.qll
@@ -67,8 +67,7 @@ class XercesDomParserLibrary extends XmlLibrary {
     // constructor.
     exists(CallInstruction call |
       call.getStaticCallTarget() = any(XercesDomParserClass c).getAConstructor() and
-      node.asInstruction().(WriteSideEffectInstruction).getDestinationAddress() =
-        call.getThisArgument() and
+      node.asInstruction().(StoreInstruction).getSourceValue() = call.getThisArgument() and
       encodeXercesFlowState(flowstate, 0, 1) // default configuration
     )
   }
@@ -153,8 +152,7 @@ class SaxParserLibrary extends XmlLibrary {
     // constructor.
     exists(CallInstruction call |
       call.getStaticCallTarget() = any(SaxParserClass c).getAConstructor() and
-      node.asInstruction().(WriteSideEffectInstruction).getDestinationAddress() =
-        call.getThisArgument() and
+      node.asInstruction().(StoreInstruction).getSourceValue() = call.getThisArgument() and
       encodeXercesFlowState(flowstate, 0, 1) // default configuration
     )
   }
--- a/cpp/ql/test/query-tests/Security/CWE/CWE-611/XXE.expected
+++ b/cpp/ql/test/query-tests/Security/CWE/CWE-611/XXE.expected
@@ -1,4 +1,6 @@
 edges
+| tests2.cpp:20:17:20:31 | Store | tests2.cpp:22:2:22:2 | p |
+| tests2.cpp:33:17:33:31 | Store | tests2.cpp:37:2:37:2 | p |
 | tests3.cpp:23:21:23:53 | call to createXMLReader | tests3.cpp:25:2:25:2 | p |
 | tests3.cpp:35:16:35:20 | p_3_3 | tests3.cpp:38:2:38:6 | Load |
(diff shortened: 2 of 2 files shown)


cpp/ql/src/Security/CWE/CWE-611/Xerces.qll
```diff
@@ -67,8 +67,7 @@ class XercesDomParserLibrary extends XmlLibrary {
     // constructor.
     exists(CallInstruction call |
       call.getStaticCallTarget() = any(XercesDomParserClass c).getAConstructor() and
-      node.asInstruction().(WriteSideEffectInstruction).getDestinationAddress() =
-        call.getThisArgument() and
+      node.asInstruction().(StoreInstruction).getSourceValue() = call.getThisArgument() and
```
MathiasVP, line 70: Well spotted! The original (i.e., the `node.asInstruction().(WriteSideEffectInstruction)`) definitely wouldn't work since write side effects don't have [instruction nodes](https://github.com/github/codeql/blob/mathiasvp/replace-ast-with-ir-use-usedataflow/cpp/ql/lib/experimental/semmle/code/cpp/ir/dataflow/internal/SsaInternalsCommon.qll#L27).

I think this fix is okay. A couple of points are worth mentioning, though:
1. It slightly transforms what is being tracked: the original code tracks the allocation (by picking the write side effect as the source), but the new code tracks the pointer pointing to the new allocation.
2. Sicne the code is specifically looking for `StoreInstruction`s, it works on instances such as:
```cpp
p = new XercesDOMParser();
```
but I don't think it works on things like:
```cpp
void useDomParser(XercesDOMParser* p) { ... }
...
useDomParser(new XercesDOMParser());
```
In order to fix this, one could use a more literal translation of the old code such as:
```ql
override predicate configurationSource(DataFlow::Node node, string flowstate) {
  // source is the write on `this` of a call to the `XercesDOMParser`
  // constructor.
  exists(Call call |
    call.getTarget() = any(XercesDomParserClass c).getAConstructor() and
    node.asExpr() = call and
    encodeXercesFlowState(flowstate, 0, 1) // default configuration
  )
}
```
Because `node.asExpr()` maps a constructor call to the underlying `DataFlow::IndirectArgumentOutNode` (this is special-cased [here](https://github.com/github/codeql/blob/mathiasvp/replace-ast-with-ir-use-usedataflow/cpp/ql/lib/semmle/code/cpp/ir/dataflow/internal/DataFlowUtil.qll#L778)), which represents the value of `*pointer` after leaving the constructor, this will track the value of the value that's pointed to by the constructor. Then, changing `configurationSink` to use `asIndirectConvertedExpr` instead of `asConvertedExpr` should hopefully give the same flow.

Does that make sense?

MathiasVP: Hm! It seems to work locally for me 🤔. Did you also change the sink as I suggested here:
> Then, changing `configurationSink` to use `asIndirectConvertedExpr` instead of `asConvertedExpr` should hopefully give the same flow.
?

jketema: Ah, I missed that bit.

MathiasVP: > I'm not sure if we were only supposed to capture cases where the allocation happened using new? I could imagine a stack allocation.

Indeed, I don't see why we shouldn't also support stack allocation.

jketema: > Hm! It seems to work locally for me 🤔. Did you also change the sink as I suggested here:

That gives me too many results (things that are clearly marked as `GOOD`). Also changing the barrier gives me too few results.

MathiasVP: Oh, interesting. In that case, I think we should get this PR in since it's clearly better than what we currently have on the feature branch.

MathiasVP (approved):
LGTM given the discussion we've had here. I'll create an internal issue for potential follow-ups.

cpp/ql/src/Security/CWE/CWE-611/Xerces.qll
jketema: I think I'll go for the solution with `NewExpr` though, that makes slightly more sense given the above. Also just stack allocating the parser variable, doesn't currently work either.

PR #11402, opened 2022-11-23
Python: port `py/super-not-enclosing-class`

yoff:


python/ql/src/Expressions/CallToSuperWrongClass.ql
```diff
@@ -13,17 +13,14 @@
...
-  exists(GlobalVariable gv, ControlFlowNode cn |
-    call_to_super = ClassValue::super_().getACall() and
-    gv.getId() = "super" and
-    cn = call_to_super.getArg(0) and
-    name = call_to_super.getScope().getScope().(Class).getName() and
-    exists(ClassValue other |
-      cn.pointsTo(other) and
-      not other.getScope().getName() = name
-    )
+  call_to_super.getFunction().getALocalSource().asExpr().(Name).getId() = "super" and
```
tausbn, line 20:
```suggestion
  call_to_super = API::builtin("super").getACall() and
```
Untested (and needs an appropriate import of the API graphs), but seeing as `super ` is a built-in, it would be nicer to use this interface, I think.

python/ql/src/Expressions/CallToSuperWrongClass.ql
```diff
@@ -13,17 +13,14 @@
...
+  name = call_to_super.getScope().getScope().(Class).getName() and
+  exists(DataFlow::Node arg |
+    arg = call_to_super.getArg(0) and
+    not arg.getALocalSource().asExpr().(Name).getId() = name
```
tausbn, line 24:
```suggestion
    arg.getALocalSource().asExpr().(Name).getId() != name
```
What you wrote is very broad, since it yields a result if, say, the local source for `arg` is not an instance of `Name`. I think it makes more sense to limit ourselves to the case where it _is_ a `Name`.

tausbn (changes requested):
Two small suggestions, otherwise this looks good to me. 👍

python/ql/src/Expressions/CallToSuperWrongClass.ql
yoff: Agreed.

python/ql/src/Expressions/CallToSuperWrongClass.ql
yoff: Thanks! That may account for the many new results 😅

yoff: That made quite a difference. The evaluation is now reasonable...

tausbn (approved):
Great work! 🚀

PR #11405, opened 2022-11-23
C++: Fix CWE-611 XXE query to work with use-use dataflow - take 2

jketema:
This commit ensures stack allocated parsers are also handled. Currently missing are the cases where the parser is assigned to global variable, as this requires indirect flow through global variables, which we know is currently missing from use-use dataflow.


MathiasVP (approved):
Awesome!

PR #11409, opened 2022-11-24
Ruby: cache the compiled extractor in the build tests

erik-krogh:
I've been looking at bit at [how we use cache in our CI](https://github.com/github/codeql/actions/caches), and I've noticed that the cargo caches are big 
(e.g. [here](https://github.com/github/codeql/actions/runs/3532847226/jobs/5927713764) where just the Windows cargo cache is ~900MB).   
There is a limit of 10GB cache pr. repo, and when the cache goes over the 10GB limit then the oldest caches are evicted, and there is lot of cache eviction right now.  

With this PR I'm caching the compiled extractor, and if there's a cache hit on the extractor, then the cargo cache is skipped completely.   

This should ensure less cache evictions throughout `github/codeql`, which can potentially speed up all workflows that use caches.

### DIFF
--- a/.github/workflows/ruby-build.yml
+++ b/.github/workflows/ruby-build.yml
@@ -48,7 +48,19 @@ jobs:
         run: |
           brew install gnu-tar
           echo "/usr/local/opt/gnu-tar/libexec/gnubin" >> $GITHUB_PATH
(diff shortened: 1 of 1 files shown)


aibaars (approved):
Good idea!

PR #11410, opened 2022-11-24
Go: Add query to check for deferred calls to functions which may return errors

mbg:
A simple query which checks that there are no deferred calls to functions which may return errors, since it prevents errors from being handled.

### DIFF
--- a/go/ql/src/InconsistentCode/DeferredErrorCall.go
+++ b/go/ql/src/InconsistentCode/DeferredErrorCall.go
@@ -1,0 +1,15 @@
+package main
+
+import (
+	"io"
+	"log"
+	"os"
+)
+
+func example() {
+	defer io.WriteString(os.Stdout, "All done!")
+
+	if _, err := io.WriteString(os.Stdout, "Hello"); err != nil {
+		log.Fatal(err)
+	}
+}
--- a/go/ql/src/InconsistentCode/DeferredErrorCall.ql
+++ b/go/ql/src/InconsistentCode/DeferredErrorCall.ql
@@ -15,7 +15,7 @@ import go
 from DeferStmt defer, SignatureType sig
 where
   // match all deferred function calls and obtain their type signatures
-  sig = defer.getCall().getCalleeExpr().getType().(SignatureType) and
+  sig = defer.getCall().getCalleeExpr().getType() and
   // check that one of the results is an error
   sig.getResultType(_).implements(Builtin::error().getType().getUnderlyingType())
 select defer,
--- a/go/ql/src/InconsistentCode/DeferredErrorCall.ql
+++ b/go/ql/src/InconsistentCode/DeferredErrorCall.ql
@@ -1,0 +1,22 @@
+/**
+ * @name Deferred call may return an error
+ * @description Deferring a call to a function which may return an error means that an error may not be handled.
+ * @kind problem
+ * @problem.severity warning
+ * @id go/deferred-error-call
+ * @tags maintainability
+ *  correctness
+ *  call
+ *  defer
+ */
+
+import go
+
+from DeferStmt defer, SignatureType sig
+where
+  // match all deferred function calls and obtain their type signatures
+  sig = defer.getCall().getCalleeExpr().getType().(SignatureType) and
+  // check that one of the results is an error
+  sig.getResultType(_).implements(Builtin::error().getType().getUnderlyingType())
+select defer,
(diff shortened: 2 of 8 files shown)


go/ql/src/InconsistentCode/DeferredErrorCall.ql
```diff
@@ -0,0 +1,22 @@
...
+ *  call
+ *  defer
+ */
+
+import go
+
+from DeferStmt defer, SignatureType sig
+where
+  // match all deferred function calls and obtain their type signatures
+  sig = defer.getCall().getCalleeExpr().getType().(SignatureType) and
```
owen-mc, line 18:
```suggestion
  sig = defer.getCall().getCalleeExpr().getType() and
```
(I see ql-for-ql has also picked this up.)

go/ql/src/InconsistentCode/DeferredErrorCall.ql
```diff
@@ -0,0 +1,22 @@
...
+  // check that one of the results is an error
+  sig.getResultType(_).implements(Builtin::error().getType().getUnderlyingType())
+select defer,
+  "Deferred a call to " + defer.getCall().getCalleeName() + ", which may return an error."
```
owen-mc, line 22: `getCalleeName` isn't always defined, so we would lose any alerts for those cases, though I can't remember exactly what they are at the moment. Is it worth trying to provide a different string when `getCalleeName` doesn't exist, to cover those cases? People defer anonymous functions quite a bit, though it would be weird to write an anonymous function with return values and then defer it.

go/ql/test/query-tests/InconsistentCode/DeferredErrorCall/tests.go
```diff
@@ -0,0 +1,23 @@
...
+	return "", errors.New("Test error")
+}
+
+func checkSomething() error {
+	return nil
+}
+
+func deferredCalls() {
+	defer returnsPair()
+	defer checkSomething()
```
owen-mc, line 17: If you want to be thorough, you can cover some more cases: (1) assign a function to a variable and use the variable in the defer statement; (2) defer a function literal.

owen-mc (approved):

go/ql/src/InconsistentCode/DeferredErrorCall.ql
mbg: Fixed in https://github.com/github/codeql/pull/11410/commits/c907a9e70719cd4521a43275b5585ddd5751eae0

go/ql/test/query-tests/InconsistentCode/DeferredErrorCall/tests.go
mbg: Added in https://github.com/github/codeql/pull/11410/commits/5620d247fd0d1763fdf8fe62bb3683481bc4ea25

go/ql/src/InconsistentCode/DeferredErrorCall.ql
mbg: Fixed in [c907a9e](https://github.com/github/codeql/pull/11410/commits/c907a9e70719cd4521a43275b5585ddd5751eae0)

owen-mc (approved):
Looks great.

owen-mc:
Oh, I just realised that it needs a change note. Luckily there's an automatic check for that. Oh, and it needs docs review. I think you have to add the `ready-for-doc-review` label and they will come and do it.

hubwriter (approved):
Docs review: one typo, otherwise LGTM 👍

go/ql/src/InconsistentCode/DeferredErrorCall.go
```diff
@@ -0,0 +1,15 @@
...
+	"log"
+	"os"
+)
+
+func example() {
+	defer io.WriteString(os.Stdout, "All done!")
+
+	if _, err := io.WriteString(os.Stdout, "Hello"); err != nil {
+		log.Fatal(err)
+	}
```
sidshank, line 14: Would it be possible to present a code example that preserves the behaviour we get from the usage of `defer`, while still handling any errors that that callee may produce i.e. what if I wanted to `defer` and also handle errors?

mbg: Great question! It seems like there are techniques for doing this by deferring an anonymous function which captures and writes to an error variable, that is then returned by the parent function. I will modify the example to make use of that technique.

PR #11422, opened 2022-11-24
LGTM deprecation: updates to CodeQL for Python articles

felicitymay:
This pull request updates the language guides for Python to remove references to LGTM.

As far as I can tell, most of the references can be removed without substitution. If I've misjudged this, please let me know.

This is one of many pull requests. I've raised a separate pull request to update the "Basic query" articles so that they guide users through get started running queries in VS Code. This should help offset the removal of the links to the query console: https://github.com/github/codeql/pull/11423.


hubwriter (dismissed):
🚀

tausbn (approved):
Slightly sad to see all these useful query console links go away, but in the absence of a replacement I guess it's the best we can do.

PR #11423, opened 2022-11-24
LGTM deprecation: Update basic queries to use VS Code

felicitymay:
This pull request updates the basic query introduction article for Python to use VS Code. I've raised this for discussion. Once we've agreed the updates, I'll apply them to the other basic query introductions.


docs/codeql/codeql-language-guides/basic-query-for-cpp-code.rst
```diff
@@ -3,7 +3,9 @@
 Basic query for C and C++ code
 ==============================
 
-Learn to write and run a simple CodeQL query using LGTM.
+Learn to write and run a simple CodeQL query using Visual Studio Code with the CodeQL extension.
+
+.. include:: ../reusables/setup-to-run-queries.rst
```
felicitymay, line 8: I've added this placeholder here so that it's easy to update this as we steadily make it easier to get started with running CodeQL queries.

docs/codeql/codeql-language-guides/basic-query-for-cpp-code.rst
```diff
@@ -14,62 +16,65 @@ The query we're going to run performs a basic search of the code for ``if`` stat
...
+Before you start writing queries, you need a CodeQL database to run them against. The simplest way to do this is to download a database directly from GitHub.com.
 
-#. Click **Query this project**.
+#. In Visual Studio Code, click the **QL** icon |codeql-ext-icon| in the left sidebar to display the CodeQL extension. 
 
-   This opens the query console. (For information about using this, see `Using the query console <https://lgtm.com/help/lgtm/using-query-console>`__.)
+#. Click **From GitHub** or the GitHub logo |github-db| at the top of the CodeQL extension to open an entry field.
 
-   .. pull-quote::
+#. Copy the URL for the repository into the field and press the keyboard **Enter** key.
```
felicitymay, line 28: ❓ Do we want to suggest an example repository here? For example, https://github.com/protocolbuffers/protobuf.

docs/codeql/codeql-language-guides/basic-query-for-cpp-code.rst
```diff
@@ -14,62 +16,65 @@ The query we're going to run performs a basic search of the code for ``if`` stat
...
 
-   .. image:: ../images/query-progress.png
-       :align: center
+      name: github-owner/cpp-experiments
+      version: 0.0.1
+      dependencies:
+         codeql/cpp-all: ^0.1.2
 
-   .. pull-quote::
+#. Create a second new file in your experiements folder with the ``.ql`` file extension. You will write your first query in this file.
```
jf205, line 53: For simplicity, we could point users to the quick query command here: https://codeql.github.com/docs/codeql-for-visual-studio-code/analyzing-your-projects/#running-a-quick-query

It would save folks having to create a new qlpack.yml file. However, they would have to copy the contents of the synthetic CodeQL pack to somewhere permanent if they wanted to retain their test queries, but that could be an optional extra step at the end of this procedure (and we could simply link to the docs).

docs/codeql/codeql-language-guides/basic-query-for-cpp-code.rst
jf205: Yes, I think that's quite helpful 😄

felicitymay:
@jf205 - thanks for your review comments ✨ 

I've updated the article to use quick query and added a note that they should move the whole directory if they want to store the query somewhere more permanent. Could you take a second look? I'd like to be sure that we're both happy before applying this approach to the other similar articles.

docs/codeql/codeql-language-guides/basic-query-for-cpp-code.rst
```diff
@@ -134,14 +141,38 @@ To exclude ``if`` statements that have an ``else`` branch:
...
+#. In Visual Studio Code, create a new folder to store your experimental queries for C and C++ CodeQL databases. For example, ``cpp-experiments``.
+
+#. Create a ``qlpack.yml`` file in your experiments folder with the contents shown below. This tells the CodeQL extension that any queries you create in the folder are intended to run on C or C++ CodeQL databases.
+
+   .. code-block:: yaml
+
+      name: github-owner/cpp-experiments
+      version: 0.0.1
+      dependencies:
+         codeql/cpp-all: ^0.1.2
```
jf205, line 161: I'm not up to date withe core library pack versioning, but do you know if would `codeql/cpp-all: *` (latest version) suffice?

docs/codeql/codeql-language-guides/basic-query-for-cpp-code.rst
```diff
@@ -134,14 +141,38 @@ To exclude ``if`` statements that have an ``else`` branch:
...
+
+#. Create a second new file in your experiements folder with the ``.ql`` file extension. You will write your first query in this file.
```
jf205, line 163:
```suggestion
#. Create a second new file in your experiments folder with the ``.ql`` file extension. You can write your next query in this file. You can also copy any of your quick queries from the current session over to the ``cpp-experiments`` folder for future development. 
```

docs/codeql/codeql-language-guides/basic-query-for-cpp-code.rst
```diff
@@ -134,14 +141,38 @@ To exclude ``if`` statements that have an ``else`` branch:
         block.getNumStmt() = 0 and
         not ifstmt.hasElse()
 
-#. Click **Run**.
+#. Re-run the query.
 
    There are now fewer results because ``if`` statements with an ``else`` branch are no longer reported.
 
-➤ `See this in the query console <https://lgtm.com/query/1899933116489579248/>`__
+Saving your quick query for future development
```
jf205, line 148: If we can get away with specifying the latest version of the library packs, then maybe this could be turned into a reusable snippet (with a placeholder <lang> instead of cpp as appropriate)?

jf205 (commented):
Thanks for the updates @felicitymay ✨ . I have a few questions but otherwise this is looking in good shape.

docs/codeql/codeql-language-guides/basic-query-for-cpp-code.rst
felicitymay: Many thanks for the review.

🤦🏻‍♀️ I copied this here when I started updating the procedure, and then realized that people could just move/copy the `Quick Query` folder created by the extension to a permanent location, as described on line 79.

I meant to delete this section from line 148 to 164 before committing the changes. Unless you think this would be useful, I suggest that we delete this.

jf205: 🔥  (please delete away!)

felicitymay:
My apologies for all the pings. I clearly managed to mess something up when I updated my local branch. Closing.

PR #11424, opened 2022-11-25
Swift: Alamofire taint sources

geoffw0:
Model taint sources from the Alamofire networking library.  I didn't get as far as I'd hoped this iteration, but I have stuff working that I want to get merged now.  Things you might expect that are _not_ in this PR:
 - the `.upload` methods (which was planned as a future extension)
 - the `stream` variant methods (the tests are here, the modelling has hit some complications I haven't had time to figure out)
 - taint sinks (they're already in for [`swift/cleartext-transmission`](https://github.com/github/codeql/pull/11210), though improvements are planned for that as well)


swift/ql/lib/codeql/swift/frameworks/Alamofire/Alamofire.qll
```diff
@@ -0,0 +1,39 @@
...
+private import codeql.swift.dataflow.ExternalFlow
+private import codeql.swift.dataflow.FlowSources
+
+/**
+ * An Alamofire response handler type.
+ */
+private class AlamofireResponseType extends NominalTypeDecl {
+  AlamofireResponseType() {
+    this.getFullName() = ["DataResponse", "DownloadResponse"] or
+    this.getABaseTypeDecl() instanceof AlamofireResponseType
```
MathiasVP, line 16: Would this be equivalent to the following

```suggestion
    this.getABaseTypeDecl*().getFullName() = ["DataResponse", "DownloadResponse"]
```
? If so, I think that's more clear.

swift/ql/test/library-tests/dataflow/flowsources/alamofire.swift
```diff
@@ -0,0 +1,551 @@
...
+}
+
+struct MyStreamSerializer: DataStreamSerializer {
+    func serialize(_ data: Data) throws -> String { return "" }
+}
+
+struct MyDecodable: Decodable {
+}
+
+func doSomething(_ p: Any) {}
```
MathiasVP, line 335: The syntax highlighting is awful from here and onwards in this file. Did a non-printable character sneak in here?

swift/ql/lib/codeql/swift/frameworks/Alamofire/Alamofire.qll
```diff
@@ -0,0 +1,39 @@
...
+
+/**
+ * A remote flow source that is an access to remote data from an Alamofire response handler.
+ */
+private class AlamofireResponseSource extends RemoteFlowSource {
+  AlamofireResponseSource() {
+    exists(AlamofireResponseType responseType |
+      this.asExpr().(MemberRefExpr).getMember() = responseType.getADataField()
+    )
+  }
```
MathiasVP, line 36: Could this not be done with the following MaD row:
```qll
";" + ["DataResponse","DownloadResponse"] + ";true;" + ["data", "value", "result"] + ";;;;remote"
```
?

MathiasVP (commented):
A couple of small comments.

swift/ql/lib/codeql/swift/frameworks/Alamofire/Alamofire.qll
geoffw0: I agree.

swift/ql/test/library-tests/dataflow/flowsources/alamofire.swift
geoffw0: Yeah, I noticed that, couldn't figure out what's causing it.  There don't seem to be any hidden characters (I tried deleting and re-typing several lines including this one).  My best guess is its a bug in the syntax highlighting???

swift/ql/lib/codeql/swift/frameworks/Alamofire/Alamofire.qll
geoffw0: I'll try it.  I haven't tried MaD on member variables but I'm guessing it should work...

MathiasVP: We have a couple of cases of that already in Swift if you need to lookup the syntax

geoffw0: Update following a bit more investigation today - for the line:
```
struct DataResponse<Success, Failure: Error> { ...
```
the `Decl` for `DataResponse` is a `StructDecl` and thus a `NominalTypeDecl`.  On the flip side, the corresponding `Type` I get for the `DataResponse` is an `UnboundGenericType` and not a `NominalType`.  This is probably reasonable (I think), but is currently tripping up `interpretElement0`.

As I said before I can fix this, but I'm currently working on the *right* fix that is clean and well understood.  Then I'll make some tests to prove we're really doing the right thing.

geoffw0: I decided to put the fix in a separate PR - https://github.com/github/codeql/pull/11514 - as it probably deserves separate review and discussion.  This PR is now blocked by that one.

swift/ql/test/library-tests/dataflow/flowsources/alamofire.swift
geoffw0: On another PR I saw red highlighting due to inconsistent indentation.  That doesn't seem to be exactly what's going on here, but a clue perhaps.

geoffw0: I've adjusted this bit to try and make the red highlighting go away.  There's nothing telling us what the problem is, it could be a highlighting but, so I'm not inclined to take this too seriously if it persists.

geoffw0:
Converted to MaD, merged main (we need the fix from https://github.com/github/codeql/pull/11514).

swift/ql/lib/codeql/swift/frameworks/Alamofire/Alamofire.qll
geoffw0: Done.

MathiasVP (approved):
LGTM!

swift/ql/test/library-tests/dataflow/flowsources/alamofire.swift
geoffw0: (this issue was reported to the appropriate team)

PR #11428, opened 2022-11-25
C++: Add more tests that exercise the default taint barrier implementation

jketema:
These demonstrate that the barrier on the sue-use dataflow feature branch is broken.


MathiasVP (approved):
LGTM! We could always argue whether or not that ideally should be a sanitizer, but from the current query's point of view, it should definitely be regarded as a sanitizer.

jketema:
Definitely. Point is mostly for us to be able to notice regressions/changes on real code once we start changing this.

PR #11432, opened 2022-11-25
Java: Flow test case generator

michaelnebel:
In this PR we
- Introduce a small script including data to allow executing the flow test case generator script locally.
- Update the access path printing for CSV format.

This is a warmup PR before migrating the flow test case generator to emit data extensions.


java/ql/test/utils/flowtestcasegenerator/pom.xml
```diff
@@ -0,0 +1,133 @@
...
+      <name>Google Maven</name>
+      <url>https://maven.google.com/</url>
+    </repository>
+  </repositories>
+  <properties>
+    <android.platform>31</android.platform>
+    <maven.compiler.source>1.8</maven.compiler.source>
+    <maven.compiler.target>1.8</maven.compiler.target>
+  </properties>
+  <dependencies>
```
atorralba, line 20: Probably most of these dependencies aren't needed for this test case, but I get that it's a bit of a pain going through all of them to determine which ones can be removed 😄

atorralba (approved):
LGTM!

java/ql/test/utils/flowtestcasegenerator/pom.xml
michaelnebel: Yeah - you are right.
Note that this is not anything that is executed by the CI but merely some internal tooling for doing developer testing of the flow test case generator.
Thank you for the review!

atorralba (approved):

PR #11435, opened 2022-11-25
C++: Rewrite `cpp/path-injection` to not use `DefaultTaintTracking`

jketema:
This is the most straightforward rewrite that made both the test here and the ones in the coding standards repo pass. This might not be sufficient to get parity with (or improvement over the old query). Will look at DCA and MRVA results (in that order).

Only annoying thing is that there's now some duplication in the results. I'm not sure if that's easily removed? There's both duplication on the source side and on the sink side, and this is the cause for the bulk of the increase in DCA results.


geoffw0 (commented):
Looks good to me, though I definitely agree about the need to look at differences on MRVA.

On the topic of result duplication, two approaches spring to mind.  If duplicates are caused by sources / sinks of the form:
```
result = a or
result = b
```
try changing them to something like:
```
result = a
or
(
  not exists(a) and
  result = b
)
```
The other thought is that if there are two sinks at slightly different `DataFlow::Node`s, we might try making sinks a `SanitizerOut` so that we only get flow to the first of them.

cpp/ql/src/Security/CWE/CWE-022/TaintedPath.ql
```diff
@@ -46,18 +47,44 @@ class FileFunction extends FunctionWithWrappers {
...
+
+  override predicate isSource(DataFlow::Node node) { isUserInput(asSourceExpr(node), _) }
+
+  override predicate isSink(DataFlow::Node node) {
+    exists(FileFunction fileFunction |
+      fileFunction.outermostWrapperFunctionCall(asSinkExpr(node), _)
+    )
   }
+
+  override predicate isSanitizerIn(DataFlow::Node node) { this.isSource(node) }
```
MathiasVP, line 77: Indeed, now that we have use-use flow this will be a common pattern to use whenever we can have flow into a source 👍.

MathiasVP (commented):
This looks great! We need to validate that the new results we're getting are mostly sensible, but once that's done I think this LGTM!

jketema:
Regarding duplication: neither of those solutions work, as the source nodes and the sink nodes of the duplicated results might have no other solution than that the underlying expression is the same.There's no flow between the two alternatives. The only thing that is giving me something close to the old result is dropping in something like:
```codeql
  predicate hasFilteredFlowPath(DataFlow::PathNode source, DataFlow::PathNode sink) {
    this.hasFlowPath(source, sink) and
    not exists(DataFlow::PathNode source2, DataFlow::PathNode sink2 |
      this.hasFlowPath(source2, sink2) and
      asSourceExpr(source.getNode()) = asSourceExpr(source2.getNode()) and
      asSinkExpr(sink.getNode()) = asSinkExpr(sink2.getNode())
    |
      not exists(source.getNode().asConvertedExpr()) and exists(source2.getNode().asConvertedExpr())
      or
      not exists(sink.getNode().asConvertedExpr()) and exists(sink2.getNode().asConvertedExpr())
    )
  }
```

Also it turns out the DCA results are very deceptive here although DCA only reports 19 more results for vim, the result count (even with the above predicate dropped in) goes from slightly over 200 to slightly over 600(!).

cpp/ql/src/Security/CWE/CWE-022/TaintedPath.ql
jketema: Note that this PR is against main.

MathiasVP: Good point. I just wanted to point out that this pattern will be even more important once we merge these changes into the feature branch (although I do see that saying "now that we have use-use flow" was a bit confusing!).

cpp/ql/src/Security/CWE/CWE-022/TaintedPath.ql
```diff
@@ -46,18 +47,44 @@ class FileFunction extends FunctionWithWrappers {
...
+Expr asSinkExpr(DataFlow::Node node) {
+  result = node.asConvertedExpr()
+  or
+  result =
+    node.asOperand()
+        .(SideEffectOperand)
+        .getUse()
+        .(ReadSideEffectInstruction)
+        .getArgumentDef()
+        .getUnconvertedResultExpression()
```
MathiasVP, line 63: Any reason why this disjunct get the unconverted expression, while the first disjunct (i.e., the `node.asConvertedExpr()`) gets the potentially _converted_ expression?

jketema: `node.asConvertedExpr()` removed some of the duplication. I didn't see a similar effect here.

jketema: That might just have been because tests coverage is not very good though.

jketema: Actually changing the last line to `.getConvertedResultExpression()` removes all the duplication for the tests, but I also lose the result for https://github.com/github/codeql/blob/165c882be54eeeb412fe5c2a532753799a0fba92/cpp/ql/test/query-tests/Security/CWE/CWE-022/semmle/tests/test.c#L35-L39

MathiasVP: Hmm. Is that because `fileFunction.outermostWrapperFunctionCall(expr, _)` only holds when `expr` is an unconverted expression in that case?

geoffw0:
> Regarding duplication: neither of those solutions work, as the source nodes and the sink nodes of the duplicated results might have no other solution than that the underlying expression is the same. There's no flow between the two alternatives.

That's unfortunate.  Where there are two alternatives with no flow between them, can we identify a common parent perhaps?

jketema:
> can we identify a common parent perhaps?

What do you mean by parent?

cpp/ql/src/Security/CWE/CWE-022/TaintedPath.ql
jketema: So it seems we need `getUnconvertedResultExpression()`?

MathiasVP: Either that, or we need to change `isSink` so that it:
- Marks the converted node as the source (so we don't get path duplication), and
- Calls `fileFunction.outermostWrapperFunctionCall` with the converted node's unconverted expression (so that we still find the right sinks).

jketema:
> I was thinking on the AST

Well, that's the problem right: it's the same expression twice in each case. Hence the code in https://github.com/github/codeql/pull/11435#issuecomment-1328876266, where is just pick one at "random".

cpp/ql/src/Security/CWE/CWE-022/TaintedPath.ql
MathiasVP: I was thinking of having `asSinkExpr` do:
```ql
result =
  node.asOperand()
      .(SideEffectOperand)
      .getUse()
      .(ReadSideEffectInstruction)
      .getArgumentDef()
      .getConvertedResultExpression()
```
(i.e., have the node be the converted one to remove the path duplication), and then implement `isSink` as something like:

```ql
  exists(FileFunction fileFunction |
    fileFunction.outermostWrapperFunctionCall(asSinkExpr(node).getUnconverted(), _)
  )
}
```
(i.e., we still mark the converted expression as a source, but we use the unconverted expression for the call to `outermostWrapperFunctionCall`).

Note that this last change _might_ mean that we get all the result duplications again (since any converted node whose _unconverted_ expression satisfies `outermostWrapperFunctionCall` will be a sink).

jketema: > Note that this last change might mean that we get all the result duplications again 

Still the same disappearing result 😄

MathiasVP: :wat:

geoffw0:
It looks like you're way ahead of me in the discussion thread with @MathiasVP , so I'll leave this to the two of you now.  Let me know if you get stuck / would like my opinion on anything.

jketema:
As far as I can currently tell, most new results I'm seeing in DCA are coming from the rewritten version not having a sanitizer mimicking the following one from default taint tracking https://github.com/github/codeql/blob/b8c11de89cedbe3686c6f145a77fc7d721bbde3f/cpp/ql/lib/semmle/code/cpp/ir/dataflow/DefaultTaintTracking.qll#L182-L196 Is this something we want to have?

MathiasVP:
> As far as I can currently tell, most new results I'm seeing in DCA are coming from the rewritten version not having a sanitizer mimicking the following one from default taint tracking

In general, I really don't like that sanitizer in DefaultTaintTracking (since a node may be marked as a sanitizer when the equality comparison is totally unrelated to the dataflow path). It's a good hammer for when a query has an unreasonable number of FPs, but my guess is that it's a lot more useful in query that does taint through integral expressions (where I'd expect a lot more equality comparisons).

jketema:
Turns out the real difference in number of results was coming from `hasUpperBoundsCheck`. Having that is not ideal either, but introducing it in this rewrite gives as very similar before and after results in DCA.

I briefly discussed with @geoffw0 and I agree with him that we should aim here for a very similar set of results when comparing with the version we had before the rewrite. Improving the query is a separate task, which we can do at a later moment, and which should now be easier, as we no longer depend on `DefaultTaintTracking`.

jketema:
MRVA results on 996 projects (one workflow timed out): https://gist.github.com/jketema/b2da9ff4d0939ca3a63e15bbcd05a556 There are 188 result changes in total (both lost and gained). This seems acceptable to me.

jketema:
As briefly discussed last week. Could you have a look at this @rdmarsh2? Especially the `hasFilteredFlowPath` predicate.

rdmarsh2:
Could we use `RemoteFlowSource`/`LocalFlowSource` rather than `isUserInput` to get the "right thing" for sources? For sinks, I'm surprised we can't just use the operand of `ReadSideEffect`.

jketema:
> Could we use `RemoteFlowSource`/`LocalFlowSource` rather than `isUserInput` to get the "right thing" for sources?

If I replace `isSink` by:
```codeql
  override predicate isSource(DataFlow::Node node) {
    node instanceof LocalFlowSource
    or
    node instanceof RemoteFlowSource
  }
```
I'm starting to lose results. In particular `scanf` related ones like:
```cpp
    char fileName[20];
    scanf("%s", fileName);
    fopen(fileName, "wb+"); // BAD
```

> For sinks, I'm surprised we can't just use the operand of `ReadSideEffect`. 

That seems to work for the test cases we have. I frankly never tried this.

MathiasVP:
> I'm starting to lose results. In particular scanf related ones like:

I hope we can do such a rewrite as a follow-up. We need to check that the FlowSource class covers all the cases that isUserInput covers, though.

jketema:
> We need to check that the FlowSource class covers all the cases that isUserInput covers, though.

Yeah, it currently doesn't, but let's not pull that into this PR.

jketema:
With the sink changes suggested by @rdmarsh2 there are [65 additional results (all losses in this case)](https://gist.github.com/jketema/68e6feb18b3416a56e609f71c3e047c7) in MRVA. Given the number of results this query gives, that's probably acceptable. So, if the query looks fine in its current form, I'd like to move on.

I'll file an internal issue for the `scanf`/`LocalFlowSource` issue identified above.

jketema:
Does this need a change note?

MathiasVP:
> Does this need a change note?

I'd say no.

geoffw0:
> I'll file an internal issue for the scanf/LocalFlowSource issue identified above.

And `fscanf`, I think in CPP we consider file reads to be local flow sources as well.  These sources will benefit other queries as well.

MathiasVP (approved):
LGTM!

